Issue #4, opened 2018-09-26
sync-catalog: Support `ClusterIP` endpoints


mitchellh:
Context: https://github.com/hashicorp/consul/pull/4710#discussion_r220575428

We should support ClusterIP by using endpoints since we already assume that pods are routable given how we treat NodePort.

This is a good first issue because we already have the machinery to sync endpoints (for NodePort) and we already have test examples showing how this would work, so this is a great way to dive in and enhance the functionality.

/cc @roboll

jackpizarro:
This would be great to have. In a setup where the HAProxy farm runs itself on pods and can reach any endpoint, services need to be created as NodePort so they get picked up.

Kudos for the work done so far, very useful!

civik:
Just curious why you assume the pods are routable? Typically pods are only routable when using a cloud providers CNI plugins.  Almost anything on-prem, most commonly Flannel, will have non-routable pod IPs.  

I'm assuming this is why when I try to deploy the client only, the pod goes into CLBO with the log message `[WARN ] to-consul/sink: error querying services, will retry: err="Get http://<host ip>:8500/v1/catalog/services?index=1&stale=&wait=60000ms: dial tcp <host ip>:8500: connect: connection refused"`

mitchellh:
> Just curious why you assume the pods are routable? Typically pods are only routable when using a cloud providers CNI plugins. Almost anything on-prem, most commonly Flannel, will have non-routable pod IPs.

There are two reasons!

First, with any new release ("0.1"-type software), we focus on getting started experience as a  primary use case. Most getting started experiences with K8S are either minikube or a hosted solution. All hosted solutions have easily routable pod IPs as you stated. 

Second, we also of course survey our customers (which are generally enterprises, global 2000 size) and get their use case so that we're designing with that in mind even if we don't solve their problem day 1. A surprisingly large number have routable pod IPs! 

Given both of these reasons, it felt like a good starting point. The architecture we built does not limit us to this, however.

> 
> I'm assuming this is why when I try to deploy the client only, the pod goes into CLBO with the log message `[WARN ] to-consul/sink: error querying services, will retry: err="Get http://<host ip>:8500/v1/catalog/services?index=1&stale=&wait=60000ms: dial tcp <host ip>:8500: connect: connection refused"`

[...]

Issue #8, opened 2018-10-03
sync-catalog: K8s to Consul: NodePort register service with POD IP address


descrepes:
Hi,

I'm deploying Consul with Helm on an AKS cluster.
Everything is working fine except the NodePort service sync.
Every K8S service is registered by default with the POD IP address and the NodePort.
Shouldn't all the services register with the Node IP address instead of the POD one ?

From now on i'm bypassing this with the annotation `consul.hashicorp.com/service-port` to force the registration with the port from the endpoint.
So a service is registered with all the endpoints.

Best regards

dikderoy:
facing same issue on bare-metal

lindenle:
@descrepes Can you explain the workaround you used in more detail? I tried adding the port but I am still seeing errors when it tries to add the nodes.

descrepes:
@lindenle In fact consul register the service with POD_IP:NodePort which is wrong.
You have to use the annotation in the service definition to use the container port instead of the NodePort. Here is a small example using helm:

> service.yaml

```
apiVersion: v1
kind: Service
metadata:
  name: {{ template "morpho.fullname" . }}
  labels:
    app: {{ template "morpho.name" . }}
    heritage: {{ .Release.Service }}
    release: {{ .Release.Name }}
    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
  annotations:
    {{- range $key, $value := .Values.service.annotations }}
      {{ $key }}: {{ $value | quote }}
    {{- end }}
spec:
  type: {{ .Values.service.type }}
  ports:
  - port: {{ .Values.service.port }}
    targetPort: {{ .Values.service.port }}
    protocol: TCP
  selector:
    app: {{ template "morpho.name" . }}
    release: {{ .Release.Name }}
```

> values.yaml

```
service:
  type: NodePort
  port: 5000
  annotations:
    consul.hashicorp.com/service-tags: "traefik.enable=true,traefik.frontend.rule=Host: morpho.mycompany.com,traefik.frontend.passHostHeader=true"
    consul.hashicorp.com/service-port: "5000"
```
Hope it will help !

lorierp:
Please help
How can I get in the consul node ip?

hidaen:
> Please help
> How can I get in the consul node ip?

see PR: #45 
You can wait for PR #45  to merge, or you can merge yourself

adilyse:
[...]

PR #18, opened 2018-10-18
catalog/from-k8s: adds support for ClusterIP Service

jipperinbham:
addresses #4 

In order to get the correct port, it has to loop through the endpoint ports while looping through the subsets. Also, in order to locate the correct "main" port, I moved the `main` variable outside the if ports condition to be referenced in the `switch`. 

I believe this covers all the cases for ClusterIP services:
- 1 port w/ no name
- multiple ports (take first port defined in the service)
- multiple ports with annotated override

### DIFF
--- a/catalog/from-k8s/resource.go
+++ b/catalog/from-k8s/resource.go
@@ -193,7 +193,11 @@ func (t *ServiceResource) shouldTrackEndpoints(key string) bool {
 		return false
 	}
 
-	return svc.Spec.Type == apiv1.ServiceTypeNodePort
+	switch svc.Spec.Type {
+	case apiv1.ServiceTypeNodePort, apiv1.ServiceTypeClusterIP:
+		return true
+	}
+	return false
 }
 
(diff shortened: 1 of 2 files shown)


jipperinbham:
I _think_ this will also address https://github.com/hashicorp/consul-k8s/issues/8 as the current release will use the NodePort for the Service port which doesn't work.

leosunmo:
This would be immensely useful for me, I hope it doesn't fall through the cracks!

darnould:
Can confirm that this made the from-k8s catalog sync functional for my use-case with ClusterIP services.

Well done @jipperinbham and thank you for your efforts.  Very keen to see this merged/resolved in a release, HashiCorp! cc/ @mitchellh

adilyse:
Huge thanks to @jipperinbham for doing the work to support ClusterIP services! A version of this has now been merged with the addition of an option to turn off ClusterIp syncing if anyone should so desire.

PR #31, opened 2018-11-02
connect-inject: use Deployment/ReplicaSet name for service registration if available

epd:
When creating Pods via Deployments or ReplicaSets, the Pod name is generated and therefore the `Name` attribute is not explicitly set. This PR allows service registration to use the generated prefix in lieu of the Pod name if the Pod name is not available.

For example, a Pod named `mypod` will generate the service name:
`mypod-<service-name-or-first-container>-proxy`.

For example, a Deployment named `mydeploy` will generate the service name:
`mydeploy-<random-hash>-<service-name-or-first-container>-proxy`.

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -23,8 +23,13 @@ type initContainerCommandUpstreamData struct {
 // containerInit returns the init container spec for registering the Consul
 // service, setting up the Envoy bootstrap, etc.
 func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
+	var podName string
+	if podName = pod.Name; podName == "" {
+		podName = strings.Trim(pod.GenerateName, "-")
+	}
+
 	data := initContainerCommandData{
-		PodName:     pod.Name,
+		PodName:     podName,
 		ServiceName: pod.Annotations[annotationService],
 	}
 	if data.ServiceName == "" {


adilyse:
Could you tell me a bit about what problem this solves for you? I see that we're not generating the proxy id correctly in the case of Deployments and ReplicaSets, but it would help to have some context to make sure that this is the correct solution.

Also, after taking a quick look at the code, I haven't been able to find any information about `GenerateName` in the docs for [Pod](https://godoc.org/k8s.io/api/core/v1#Pod). Should there be some additional code here to support this?

evan-eb:
Hi @adilyse! Forgive me for not providing enough context. I'm using the auto-inject feature, with the following Deployment:
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  annotations:
  labels:
    app: some-app
  name: some-app
spec:
  replicas: 3
  selector:
    matchLabels:
      app: some-app
  template:
    metadata:
      annotations:
        consul.hashicorp.com/connect-service: some-app
      labels:
        app: some-app
    spec:
      containers:
      ...
```
I'm observing through logging to stdout in the handler, that for Deployments, the `Name` property for the Pod is not set. I believe this is because of the reasoning mentioned here in the comments for `GenerateName`: https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta. This works as expected when annotating a simple Pod.

From the template that the init container generates:
```hcl
services {
   id   = "{{ .PodName }}-{{ .ServiceName }}-proxy"
   name = "{{ .ServiceName }}-proxy"
   kind = "connect-proxy"
```
This would register a service with the id of `some-app-<random-hash>-some-app-proxy" and a name of `some-app-proxy` in the case of the Deployment above. The `some-app-<random-hash>` coming from the `GenerateName` of the Deployment.

Looking forward to your advice - thank you for taking the time to review!

evan-eb:
@adilyse no problem - thank you for your time on this project!

PR #34, opened 2018-11-21
catalog/from-k8s: Use correct ServicePollPeriod

mattrobenolt:
I think this is correct, but it seems otherwise, this value isn't used
anywhere, and based on the corresponding doc strings, it seems we should
be using ServicePollPeriod instead.

### DIFF
--- a/catalog/from-k8s/syncer.go
+++ b/catalog/from-k8s/syncer.go
@@ -212,7 +212,7 @@ func (s *ConsulSyncer) watchService(ctx context.Context, name string) {
 			return
 
 		// Wait for our poll period
-		case <-time.After(s.SyncPeriod):
+		case <-time.After(s.ServicePollPeriod):
 		default:
 		}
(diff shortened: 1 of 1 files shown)


adilyse:
Hi @mattrobenolt,

This is a great find! You're completely right, we currently don't use the `ServicePollPeriod` at all and don't even allow users to set the value. There doesn't seem to be a compelling reason to keep this as a distinct value, so I think the correct solution at this point is to remove `ServicePollPeriod` entirely. 

Given that, I'm going to go ahead and close this PR, but I would be happy to merge one that removes `ServicePollPeriod` if you'd like to help us clean this up!

mattrobenolt:
Users are allowed to set this value. It is derived from `-consul-write-interval`. Now a user can't set it explicitly, but it's set to `syncInterval * 2`, so there is apparently some reason to use a different value here.

PR #38, opened 2018-11-26
Use stale flag

i0rek:
I noticed that there is a `-stale` flag provided by `flags.HTTPFlags` but it is not actually used. This PR changes it so that the value of the flag is being pushed down to the request.

This however changes the behaviour since before `stale` was always enabled and now it is disabled by default.

### DIFF
--- a/catalog/from-consul/source.go
+++ b/catalog/from-consul/source.go
@@ -19,13 +19,14 @@ type Source struct {
 	Sink   Sink         // Sink is the sink to update with services
 	Prefix string       // Prefix is a prefix to prepend to services
 	Log    hclog.Logger // Logger
+	Stale  bool         // Stale enables stale reads from Consul servers
 }
 
 // Run is the long-running runloop for watching Consul services and
 // updating the Sink.
 func (s *Source) Run(ctx context.Context) {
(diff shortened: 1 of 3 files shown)


mitchellh:
I think we should prefer stale being default. The syncers in particular can put some pretty heavy load on the servers and we'd like them to not have to route back to the leader every time for reads. Not sure if there is an easy way to change this default. I'd be happy to allow this to be configurable though.

i0rek:
I think changing it in Consul is not an option since `false` is the sensible default. That makes me think a workaround is the better choice. We could do something like the this:

```go
func (c *Command) getStaleWithDefaultTrue() bool {
	stale := true
	c.flags.Visit(func(f *flag.Flag) {
		if f.Name == "stale" {
			stale = c.http.Stale()
			return
		}
	})
	return stale
}
```

adilyse:
Thanks for putting this together!

At the moment, there's not a lot of value to being able to set `stale=false` as there can be up to a 30 second delay in the catalog sync already. At this point in time, I don't think it's worth making this configurable as it's more likely to trip people up than help them.

We can revisit this in the future if we find a compelling reason for it.

PR #39, opened 2018-11-29
Customizable Consul sync tag (fixes #42)

### LINKED ISSUE #42: Multi-Cluster Deploy
This may be a duplicate of #6, if so please feel free to close this, but currently running more than one consul-k8s deployment across different clusters results in the deployments fighting one another:

```
2018-12-07T05:39:41.560Z [INFO ] to-consul/sink: invalid service found, scheduling for delete: service-name=SERVICE_NAME
```

I'd ideally like to be able to run multiple deployments across clusters, and have a particular deployment only manage what it synced. Would adding a `-svc-tag` option solve this I wonder? I'm thinking if set consul-k8s would only manage the services, and more specifically the nodes, with said tag value? Something like this would solve our use case as we do have multiple clusters with the same service defined in both.

r2bit:
Customizable Consul catalog sync tag makes it possible to share a single Consul cluster with more than one K8S cluster. With the hardcoded tag ("k8s"), all catalog-sync processes would delete each other's services from Consul.
Controlled via "-consul-k8s-tag" flag.

### DIFF
--- a/catalog/from-consul/source.go
+++ b/catalog/from-consul/source.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff"
-	fromk8s "github.com/hashicorp/consul-k8s/catalog/from-k8s"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
 )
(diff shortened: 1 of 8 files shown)


tvi:
@mitchellh would it be possible to get review on this? and potentially get it released? Thanks

jacksontj:
:+1: from me, this is a requirement before we could use this.

jacksontj:
cc @adilyse as you seem to be the most recently active :)

adilyse:
Merged the rebased version with attribution in #51. Thanks again for a great contribution!

PR #41, opened 2018-11-30
catalog/from-consul: datacenter flag

egoldblum:
Allow configuration of datacenter for service sync from consul


egoldblum:
A naive attempt to support my use case as described here: https://groups.google.com/forum/#!topic/consul-tool/3-14opoD5jk

egoldblum:
Turns out this already possible using the standard http `-datacenter` flag!

PR #43, opened 2018-12-07
Add support for ClusterIP Services

adilyse:
This takes the great [PR](https://github.com/hashicorp/consul-k8s/pull/18) from @jipperinbham that adds ClusterIP service sync support and adds a flag to make it optional. Since there are many Kubernetes environments where ClusterIPs aren't externally routeable, this allows people in those environments an easier way to only sync routeable services.

### DIFF
--- a/catalog/from-k8s/resource.go
+++ b/catalog/from-k8s/resource.go
@@ -43,6 +43,10 @@ type ServiceResource struct {
 	// enabled (aka default enabled).
 	ExplicitEnable bool
 
+	// ClusterIPSync set to true (the default) syncs ClusterIP-type services.
+	// Setting this to false will ignore ClusterIP services during the sync.
+	ClusterIPSync bool
+
 	// serviceMap is a mapping of unique key (given by controller) to
 	// the service structure. endpointsMap is the mapping of the same
 	// uniqueKey to a set of endpoints.
@@ -157,6 +161,13 @@ func (t *ServiceResource) shouldSync(svc *apiv1.Service) bool {
 		return false
 	}
 
+	// Ignore ClusterIP services if ClusterIP sync is disabled
+	if svc.Spec.Type == apiv1.ServiceTypeClusterIP {
+		if !t.ClusterIPSync {
+			return false
+		}
+	}
+
 	raw, ok := svc.Annotations[annotationServiceSync]
 	if !ok {
 		// If there is no explicit value, then set it to our current default.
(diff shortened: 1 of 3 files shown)


catalog/from-k8s/resource.go
```diff
@@ -43,6 +43,10 @@ type ServiceResource struct {
 	// enabled (aka default enabled).
 	ExplicitEnable bool
 
+	// ClusterIPSync set to true (the default) syncs ClusterIP-type services.
```
rboyer: why default to true?

catalog/from-k8s/resource.go
```diff
@@ -157,6 +161,13 @@ func (t *ServiceResource) shouldSync(svc *apiv1.Service) bool {
 		return false
 	}
 
+	// Ignore ClusterIP services if ClusterIP sync is disabled
+	if svc.Spec.Type == apiv1.ServiceTypeClusterIP {
+		if !t.ClusterIPSync {
```
banks: nit: I'd probably make this `if svc.Spec.Type == apiv1.ServiceTypeClusterIP &&  !t.ClusterIPSync {`

catalog/from-k8s/resource.go
adilyse: This makes it easier for first time users to see catalog sync working since ClusterIP is the default service type.

catalog/from-k8s/resource.go
```diff
@@ -372,6 +380,13 @@ func (t *ServiceResource) generateRegistrations(key string) {
 				}
 				seen[addr] = struct{}{}
 
+				for _, port := range subset.Ports {
```
rboyer: Will this still sync the NodePort value properly for NodePort services as in the now-deleted line 264?

catalog/from-k8s/resource.go
```diff
@@ -157,6 +161,13 @@ func (t *ServiceResource) shouldSync(svc *apiv1.Service) bool {
 		return false
 	}
 
+	// Ignore ClusterIP services if ClusterIP sync is disabled
+	if svc.Spec.Type == apiv1.ServiceTypeClusterIP {
```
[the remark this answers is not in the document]
adilyse: I think the filtering previously was a side effect of the service type switch statement in `generateRegistrations`. Putting it in the `shouldSync` method seems much cleaner, since we now support `ClusterIP` in `generateRegistrations`.

catalog/from-k8s/resource.go
adilyse: Good catch. Reverting to previous port calculations until we can look at this further.

adilyse:
Fixes #4 .

adilyse:
I will make sure to follow up and see that we fix that. For now, it made sense to separate these to make it a little cleaner. Thanks again for your changes-- there's been a lot of enthusiasm for this feature and it's great to be able to get it merged!

PR #45, opened 2018-12-20
Nodeport syncing updates

adilyse:
This updates the NodePort sync logic to use the node's ip address, rather than the pod's ip. This puts a fully routable address into the sync, making it possible to externally sync with these services.

Additionally, it updates the port determination logic to take the first defined NodePort when there are multiples. It also now takes the first address when the ports are unnamed, as well as adding tests for this case.

### DIFF
--- a/catalog/from-k8s/resource.go
+++ b/catalog/from-k8s/resource.go
@@ -349,36 +349,55 @@ func (t *ServiceResource) generateRegistrations(key string) {
 			t.consulMap[key] = append(t.consulMap[key], &r)
 		}
 
-	// For NodePort services, we register each K8S
+	// For NodePort services, we create a service instance for each
+	// endpoint of the service, which corresponds to the nodes the service's
+	// pods are running on. This way we don't register _every_ K8S
 	// node as part of the service.
 	case apiv1.ServiceTypeNodePort:
-		// Get all nodes to be able to reference their ip addresses
-		nodes, err := t.Client.CoreV1().Nodes().List(metav1.ListOptions{})
-		if err != nil || len(nodes.Items) == 0 {
-			t.Log.Warn("error getting nodes", "error", err)
+		if t.endpointsMap == nil {
 			return
 		}
 
-		// Create a service instance for each node
-		for _, node := range nodes.Items {
-			for _, address := range node.Status.Addresses {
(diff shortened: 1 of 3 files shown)


hidaen:
Hi @adilyse I found some problems while testing the nodeport-syncing branch:
When I tested nodeport, I found that some nodeport service address.Type is NodeInternalIP. This situation is not registered. Will this be considered for entry? Or why there is only reason to register NodeExternalIP.

adilyse:
@hidaen 

Thanks for taking a look! The InternalIP issue is a good find. I was working from the assumption that people were wanting to access these services from outside of their Kubernetes cluster, which would make the ExternalIP more appropriate. 

For your InternalIP use case, would using an InternalIP if an ExternalIP isn't found be enough? Or do you need more control?

adilyse:
Thinking about this further, adding an option seems to support a wider array of use cases, so I've opted for that.

hidaen:
> Thinking about this further, adding an option seems to support a wider array of use cases, so I've opted for that.

@adilyse 
I think there is a problem with my description, I think adding the sync-node-external-ip parameter is not a good choice.
External access to the Kubernetes service, using ExternalIP is indeed more appropriate.
But Kubernetes NodePort service should register ExternalIP or Node IP whether it is ExternalIP or InternalIP
I am going to try to write this piece of code, so I will ask again.

catalog/from-k8s/resource.go
```diff
@@ -47,6 +47,11 @@ type ServiceResource struct {
 	// Setting this to false will ignore ClusterIP services during the sync.
 	ClusterIPSync bool
 
+	// NodeExternalIPSync set to true (the default) syncs NodePort services
+	// using the node's external ip address. When false, the node's internal
+	// ip address will be used instead.
+	NodeExternalIPSync bool
```
mitchellh: We don't have to, but I wonder if we should change this to take an enum of values so we can support the various ways people do NodePort syncs. I'm not sure if its necessary, but don't want to end up with boolean-hell one day with a bunch of modes. 😄

adilyse: That's a good thought, and allows me to address the case where someone might want the ExternalIP if it's available, but wants to fall back to the InternalIP case.

adilyse:
Fixes #8

PR #49, opened 2019-01-08
Add a flag to specify Kubernetes NodePort service register IP address

hidaen:
Kubernetes NodePort service, whether using ExternalIP or InternalIP, Node IP should be registered.

NodePort service the default exposed port is on the Node IP
NodeHostName, NodeExternalDNS, NodeInternalDNS for the time being, I don't know how to use it. I will use it again. :D

k8s.io/api/core/v1/types.go:
const (
    NodeHostName    NodeAddressType = "Hostname"                                                                                                                           NodeExternalIP  NodeAddressType = "ExternalIP"
    NodeInternalIP  NodeAddressType = "InternalIP"                                                                                                                         NodeExternalDNS NodeAddressType = "ExternalDNS"
    NodeInternalDNS NodeAddressType = "InternalDNS"                                                                                                                    )

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 /bin
 /pkg
+vendor/
--- a/catalog/from-k8s/resource.go
+++ b/catalog/from-k8s/resource.go
@@ -406,7 +406,7 @@ func (t *ServiceResource) generateRegistrations(key string) {
 				// Find the external ip address for the node and
 				// create the Consul service using it
 				for _, address := range node.Status.Addresses {
-					if address.Type == apiv1.NodeExternalIP {
+					if address.Type == apiv1.NodeExternalIP || address.Type == apiv1.NodeInternalIP {
 						r := baseNode
 						rs := baseService
 						r.Service = &rs


adilyse:
The problem I see with this implementation is that a node that has an `ExternalIP` also has an `InternalIP`, so this will give a non-deterministic IP address back for service registration. 

Can you tell me a bit more about the Kubernetes setup that would give some nodes that have external ips and some that don't? Or do you need per-service choice about external vs internal ip usage?

(Closing this PR for the above reason, but I'd still like to hear more about your situation!)

hidaen:
> The problem I see with this implementation is that a node that has an `ExternalIP` also has an `InternalIP`, so this will give a non-deterministic IP address back for service registration.
> 
> Can you tell me a bit more about the Kubernetes setup that would give some nodes that have external ips and some that don't? Or do you need per-service choice about external vs internal ip usage?
> 
> (Closing this PR for the above reason, but I'd still like to hear more about your situation!)

@adilyse 

When we use k8s service type: NodePort to expose the service, when we do not set ExternalIP, Nodeport is bound to the Node IP, so when not setting ExternalIP, register NodeIP: Nodeport

adilyse:
@hidaen Thanks for the follow up information!

I ended up changing the flag to allow a variety of syncing options- ExternalOnly, InternalOnly and ExternalFirst. I believe your use case is covered by the last option. It will sync with an ExternalIP if it exists, but if it doesn't, it will use the InternalIP instead.

Thank you for the feedback to help make this a better feature!

hidaen:
> @hidaen Thanks for the follow up information!
> 
> I ended up changing the flag to allow a variety of syncing options- ExternalOnly, InternalOnly and ExternalFirst. I believe your use case is covered by the last option. It will sync with an ExternalIP if it exists, but if it doesn't, it will use the InternalIP instead.
> 
> Thank you for the feedback to help make this a better feature!

👍 Thank you

PR #51, opened 2019-01-08
Customizable Consul sync tag

adilyse:
Rebased version of #39 .

Notes from that PR:
Customizable Consul catalog sync tag makes it possible to share a single Consul cluster with more than one K8S cluster. With the hardcoded tag ("k8s"), all catalog-sync processes would delete each other's services from Consul.
Controlled via "-consul-k8s-tag" flag.


adilyse:
Fixes #42.

PR #62, opened 2019-02-01
Add datacenter support to inject annotation

nicholasjackson:
This PR adds support to specify the datacenter for a consul connect upstream from the K8s annotation.
https://www.consul.io/docs/connect/proxies.html#datacenter

The new format has the form of the following example:
```
annotations:
  "consul.hashicorp.com/connect-service-upstreams": "[service]:[port]:[datacenter]"
```

`datacenter` is an optional parameter 

If this is all OK I will raise a separate PR for the consul.io docs

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -15,8 +15,9 @@ type initContainerCommandData struct {
 }
 
 type initContainerCommandUpstreamData struct {
-	Name      string
-	LocalPort int32
+	Name       string
+	LocalPort  int32
+	Datacenter string
 }
 
 // containerInit returns the init container spec for registering the Consul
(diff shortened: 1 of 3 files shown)


connect-inject/container_init.go
```diff
@@ -124,6 +133,9 @@ services {
     upstreams {
       destination_name = "{{ .Name }}"
       local_bind_port = {{ .LocalPort }}
+			{{ if .Datacenter }}
```
adilyse:
```suggestion
      {{- if .Datacenter }}
```
This removes the white space before the command, so the datacenter lines up correctly with `local_bind_port`.

connect-inject/container_init.go
```diff
@@ -124,6 +133,9 @@ services {
...
+			datacenter = "{{ .Datacenter }}"
```
adilyse:
```suggestion
      datacenter = "{{ .Datacenter }}"
```
Aligns the indentations.

connect-inject/container_init.go
```diff
@@ -124,6 +133,9 @@ services {
...
+			{{ end }}
```
adilyse:
```suggestion
      {{- end }}
```
Similarly removes the blank line that would otherwise be inserted here and updates indentation.

connect-inject/container_init_test.go
```diff
@@ -70,6 +70,28 @@ func TestHandlerContainerInit(t *testing.T) {
...
+
+		{
+			"No Upstream datacenter specified",
+			func(pod *corev1.Pod) *corev1.Pod {
+				pod.Annotations[annotationService] = "web"
+				pod.Annotations[annotationUpstreams] = "db:1234"
+				return pod
+			},
+			"",
+			`datacenter = "dc1"`,
```
adilyse: In this case, it would be good to make sure that no datacenter info is included rather than making sure a specific value is not present.
```suggestion
			`datacenter`,
```

nicholasjackson:
Hey @adilyse, thanks so much for your help with this, I have updated the code with your suggestions.

Many thanks,

Nic

adilyse:
Many thanks for adding this! We'll need to make sure the docs get updated as well.

Issue #59, opened 2019-01-29
services de-registering randomly 


tsmgeek:
Kubernetes Platform Version: 1.11.5 (eks.1)

I installed the consul-0.5.0 helm chart and enabled sync k8s->consul.
In the default setup services appear and disappear randomly, I am unable to see any pattern to what is happening.

There are logs such as these on startup but nothing else causing concern other than `registering services` for the most part.
```
2019-01-29T09:59:38.933Z [WARN ] to-consul/source: error loading initial endpoints: key=kafka/confluent-1-nodeport err="an empty namespace may not be set when a resource name is provided"
2019-01-29T10:00:13.564Z [INFO ] to-consul/sink: invalid service found, scheduling for delete: service-name=sentry
2019-01-29T10:32:17.364Z [WARN ] to-consul/source: error loading initial endpoints: key=testing/rocketchat err="an empty namespace may not be set when a resource name is provided"
2019-01-29T10:32:17.368Z [INFO ] to-consul/source: upsert: key=testing/rocketchat
```

I have done some checks and this is what i can see.
- services randomly disappear between "registering services" log entries
- services that have disappeared instantly show up again when "registering services" is shown
- seems it just re-registers the missing services
- happens with services that have any number of pods
- ive tested 0.3.0 and 0.4.0

adilyse:
More details about this issue are available in #40, and it should be fixed by #63.

josephschadlick:
I am experiencing this same issue even after upgrading to consul-k8s v0.5.0 (consul-helm v0.6.0). I can't nail down the cause. Let me know if you require additional information or if I should open a new  issue around this.

tsmgeek:
Confirmed that I am still seeing the issue, its not as bad as before but there are still some services that are disappearing.

PR #63, opened 2019-02-01
Register all kubernetes services to the `k8s-sync` node

### LINKED ISSUE #40: Services deregistering - shouldn't the consul agent service api be used rather than the catalog's?
Hello,

I have consul running on my kubernetes nodes and I got the consul-k8s sync so work, no problems there. (Well there are, as it uses the POD IP and not the NODE IP /which also makes no sense to me, as it supports only loadbalancer/nodeport, but I see there is issue opened for this/.
What am I seeing is that on each of my kubernetes nodes, after certain amount of time, the consul agent auto deregisters the services, registered by consul-sync daemon. I experience this issue both running consul as a pod on each kubernetes server and as running it as external to the kubernetes client.
I see that the Catalog registration is used in the from-k8s source code. This triggers the  Anti-Entropy  feature of consul agent for services registered this way.
Even in the documentation of the API documentation it says 

> **"It is usually preferable to instead use the agent endpoints for registration as they are simpler and perform anti-entropy."**

Also this here from the documentation

>  If any services or checks exist in the catalog that the agent is not aware of, they will be automatically removed to make the catalog reflect the proper set of services and health information for that agent. Consul treats the state of the agent as authoritative; if there are any differences between the agent and catalog view, the agent-local view will always be used.

I see that this can be disabled using enable_tag_override, but did not have the time today to compile the syncing daemon with this option to test it.  Also this would imho be a dirty workaround.

Am I doing/understanding something wrong here and this is desired behavior? Because the constant registration/deregistration of services triggers multiple consul-template events.

Shouldn't the consul-sync daemon be run as daemonset on each node and contact and update the client api of that kubernetes node? Or something similar...

adilyse:
Registering the services with their node information causes issues with
anti-entropy on the clients because the kubernetes services are registered directly with
the servers rather than the clients. So, the clients don't have a record
of the service and deregister them. This causes a register/deregister loop
because the sync process automatically re-registers them.

Given that the sync doesn't register things directly to the clients, it
is more correct to register them to a synthetic node (`k8s-sync` in this
case) to indicate that they didn't come from a client registration.

Fixes #40 and #59.


adilyse:
Yes, this should fix the agents deregistering Kubernetes services.

PR #70, opened 2019-02-21
Add prepared_query annotation support

s-christoff:
This PR adds support to specify the [prepared query](https://www.consul.io/docs/connect/proxies.html#upstreams) for consul connect upstream from the K8s annotation.

```
annotations:
  "consul.hashicorp.com/connect-service-upstreams": "prepared_query:[query name]:[port]"
```
Services named `prepared_query` will caught in the prepared query logic, if that does occur.

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -133,9 +133,9 @@ services {
     upstreams {
       destination_name = "{{ .Name }}"
       local_bind_port = {{ .LocalPort }}
-			{{- if .Datacenter }}
+      {{- if .Datacenter }}
       datacenter = "{{ .Datacenter }}"
-			{{- end}}
+      {{- end}}
     }
     {{ end }}
   }
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -18,6 +18,7 @@ type initContainerCommandUpstreamData struct {
 	Name       string
 	LocalPort  int32
 	Datacenter string
+	Query	   string
 }
 
 // containerInit returns the init container spec for registering the Consul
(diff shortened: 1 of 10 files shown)


connect-inject/container_init.go
```diff
@@ -131,7 +141,14 @@ services {
 
     {{ range .Upstreams -}}
     upstreams {
+		{{- if .Name }}
```
adilyse:
```suggestion
      {{- if .Name }}
```

Looks like the whitespace beast got us here too!

connect-inject/container_init_test.go
```diff
@@ -91,6 +91,27 @@ func TestHandlerContainerInit(t *testing.T) {
 			"",
 			`datacenter`,
 		},
+		{
+			"Prepared Query specified",
```
adilyse: Since we have two of these checks, it would be good to specify that this one checks that the `destination_type` is set correctly for a query annotation.

connect-inject/container_init_test.go
```diff
@@ -91,6 +91,27 @@ func TestHandlerContainerInit(t *testing.T) {
...
+			func(pod *corev1.Pod) *corev1.Pod {
+				pod.Annotations[annotationService] = "web"
+				pod.Annotations[annotationUpstreams] = "prepared_query:handle:1234"
+				return pod
+			},
+			`destination_type = "prepared_query"`,
+			"",
```
adilyse:
```suggestion
			`destination_type = "service"`,
```

Ooo, we could make this test do more work for us by checking that `destination_type = "service"` is not included as part of it!

connect-inject/container_init_test.go
```diff
@@ -91,6 +91,27 @@ func TestHandlerContainerInit(t *testing.T) {
...
+		},
+
+		{
+			"Prepared Query specified",
```
adilyse: As with the other test case, mentioning in the test title that this one checks that the `destination_name` is set correctly from a query annotation would help make its purpose more explicit.

s-christoff:
Ran into some Git Issues, Fixed in #71

PR #75, opened 2019-03-01
govet and gofmt cleanup

alvin-huang:
These are just a few small changes while working on putting this repo into CircleCI. Govet had the following complaints:
```
# github.com/hashicorp/consul-k8s/helper/cert
helper/cert/notify.go:31: the cancelFunc function is not used on all paths (possible context leak)
helper/cert/notify.go:38: this return statement may be reached without using the cancelFunc var defined on line 31
# github.com/hashicorp/consul-k8s/helper/coalesce
helper/coalesce/coalesce.go:54: the curCancel function is not used on all paths (possible context leak)
helper/coalesce/coalesce.go:63: this return statement may be reached without using the curCancel var defined on line 54
```
Instead of deferring `curCancel()` after the context creation which could bubble up an _infinite_ number of `cancelFunc()` calls or placing it in the error check which is what `go vet` is complaining about (since I don't think it knows about the defer call outside the loop), I opted to refactor the for loop a bit to both make the code more readble while also cancelling the contexts in a way that is more straightforward.

### DIFF
--- a/helper/coalesce/coalesce.go
+++ b/helper/coalesce/coalesce.go
@@ -52,6 +52,7 @@ func Coalesce(ctx context.Context, quiet, max time.Duration, f func(context.Cont
 
 		// Create a context with our quiet period
 		curCtx, curCancel = context.WithTimeout(ctx, quiet)
+		defer curCancel()
 
 		// Call the function
 		f(curCtx)
--- a/helper/coalesce/coalesce.go
+++ b/helper/coalesce/coalesce.go
@@ -33,35 +33,22 @@ func Coalesce(ctx context.Context, quiet, max time.Duration, f func(context.Cont
 	ctx, maxCloser := context.WithTimeout(ctx, max)
 	defer maxCloser()
 
-	// Setup the variables to store the context and cancellation function
-	// for each iteration. We do this outside of the loop so that we can
-	// setup a defer to ensure we cleanup any resources associated with the
-	// context.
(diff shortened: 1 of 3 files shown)


helper/coalesce/coalesce.go
```diff
@@ -33,34 +33,22 @@ func Coalesce(ctx context.Context, quiet, max time.Duration, f func(context.Cont
...
-		// end, then the function processed successfully and we continue.
-		err := curCtx.Err()
+		err := doCoalesce(ctx, quiet, f)
 		if err != nil {
 			return
 		}
 	}
 }
+
+func doCoalesce(ctx context.Context, quiet time.Duration, f func(context.Context)) error {
```
alvin-huang: I couldn't figure out a good name for this.... 🤔

mitchellh:
The `go vet` output for the `curCancel` variable seems wrong to me. I've been wrong about that before though but it just seems incorrect. The way it was before is sort of complex but it keeps all the context (no pun intended) in one location with heavy commenting, which was an explicit intention so that as you try to understand it you can read it.

I'd be curious if someone else can take a look but this seems like a bug in `go vet` to me.

Unfortunately I don't think there is a way to silence `go vet` on a file or line basis, you'd have to ignore the whole package. In light of that, your change seems like it'd work fine. I think we have unit tests covering the coalescence behavior and it seems to do the right thing.

PR #81, opened 2019-03-15
Register Connect services along with their proxies

adilyse:
This allows both services and their proxies to show up in the Consul
catalog. It makes it easier to find services in the Intentions UI.

These services are registered through the client pod on their local node,
like the proxies are, so unlike services in the catalog sync, they
are not connected to the `k8s-sync` synthetic node.

Note: it looks like there are a couple of whitespace gofmt changes
included as well. They do not change any functionality.

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -18,7 +18,7 @@ type initContainerCommandUpstreamData struct {
 	Name       string
 	LocalPort  int32
 	Datacenter string
-	Query	   string
+	Query      string
 }
 
 // containerInit returns the init container spec for registering the Consul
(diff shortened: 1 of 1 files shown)


connect-inject/container_init.go
```diff
@@ -169,6 +169,13 @@ services {
     alias_service = "{{ .ServiceName }}"
   }
 }
+
+services {
+  id   = "${POD_NAME}-{{ .ServiceName }}"
```
banks: nit: would it be nicer to manage if the service name were first? I'm assuming pod name is long and random looking so it would possibly be cleaner for humans to scan if the readable service name was first?

adilyse: These will all get grouped together under the service `name` already. At the service level though, it's more important to be able to identify the specific instance first, rather than the service. I'm going to leave this for now, but I'm open to changing it later based on community feedback.

PR #89, opened 2019-04-24
Support Prefixes #140

s-christoff:
Add `K8sConsulPrefix` flag to be used to prepend K8s services in Consul with a prefix

### DIFF
--- a/catalog/from-k8s/resource.go
+++ b/catalog/from-k8s/resource.go
@@ -1,6 +1,7 @@
 package catalog
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 	"sync"
@@ -52,6 +53,9 @@ type ServiceResource struct {
 
 	// ConsulK8STag is the tag value for services registered.
 	ConsulK8STag string
+
+	//ConsulServicePrefix prepends K8s services in Consul with a prefix
+	ConsulServicePrefix string
 
 	// ExplictEnable should be set to true to require explicit enabling
 	// using annotations. If this is false, then services are implicitly
@@ -177,7 +181,7 @@ func (t *ServiceResource) shouldSync(svc *apiv1.Service) bool {
 	// a sync for that namespace.
 	if t.namespace() == metav1.NamespaceAll && svc.Namespace == metav1.NamespaceSystem {
 		t.Log.Debug("ignoring system service since we're listening on all namespaces",
-			"service-name", svc.Name)
+			"service-name", t.prefixServiceName(svc.Name))
 		return false
 	}
 
@@ -195,7 +199,7 @@ func (t *ServiceResource) shouldSync(svc *apiv1.Service) bool {
 	v, err := strconv.ParseBool(raw)
 	if err != nil {
 		t.Log.Warn("error parsing service-sync annotation",
-			"service-name", svc.Name,
+			"service-name", t.prefixServiceName(svc.Name),
 			"err", err)
 
 		// Fallback to default
@@ -259,7 +263,7 @@ func (t *ServiceResource) generateRegistrations(key string) {
 	}
 
 	baseService := consulapi.AgentService{
-		Service: svc.Name,
+		Service: t.prefixServiceName(svc.Name),
 		Tags:    []string{t.ConsulK8STag},
 		Meta: map[string]string{
 			ConsulSourceKey: ConsulSourceValue,
@@ -608,4 +612,11 @@ func (t *serviceEndpointsResource) Delete(key string) error {
 
 	t.Service.Log.Info("delete endpoint", "key", key)
 	return nil
+}
+
+func (t *ServiceResource) prefixServiceName(name string) string {
+	if t.ConsulServicePrefix != "" {
+		return fmt.Sprintf("%s%s", t.ConsulServicePrefix, name)
+	}
+	return name
 }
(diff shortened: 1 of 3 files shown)


catalog/from-k8s/resource_test.go
```diff
@@ -1770,3 +1770,297 @@ func testService(name string) *apiv1.Service {
...
+	require.NoError(err)
+
+	// Wait a bit
+	time.Sleep(300 * time.Millisecond)
+
+	// Verify what we got
+	syncer.Lock()
+	defer syncer.Unlock()
+	actual := syncer.Registrations
+	require.Len(actual, 2)
```
freddygv: To cut down on the amount of checks one thing you could do is declare a struct with the output you expect and let `require.Equal` do all the comparing:
```go
expected := []*api.CatalogRegistration{
	&api.CatalogRegistration{
		Service: &api.AgentService{
			Service: "prefixfoo",
			Address: "1.2.3.4",
			Port:    30000},
		Node: "k8s-sync",
	},
	&api.CatalogRegistration{
		Service: &api.AgentService{
			Service: "prefixfoo",
			Address: "2.3.4.5",
			Port:    30000},
		Node: "k8s-sync",
	},
}
require.Equal(t, expected, actual)
```

I find that it can help clarify the values and structure of what you're expecting.

Also, if you declare a struct like this in the beginning of the test, you can substitute hardcoded values like the addresses in lines 1935 and 1945 for `expected[0].Service.Address` and `expected[1].Service.Address`

freddygv: Actually, nevermind, this wouldn't really work here because not all the fields will match. (like ID)

catalog/from-k8s/syncer.go
```diff
@@ -62,6 +62,9 @@ type ConsulSyncer struct {
 	// ConsulK8STag is the tag value for services registered.
 	ConsulK8STag string
 
+	//K8sConsulPrefix prepends K8s services in Consul with a prefix
+	K8sConsulPrefix string
```
adilyse: Is this additional struct field necessary? It doesn't seem to be used anywhere in the syncer.

subcommand/sync-catalog/command.go
```diff
@@ -157,21 +161,23 @@ func (c *Command) Run(args []string) int {
 			SyncPeriod:        syncInterval,
 			ServicePollPeriod: syncInterval * 2,
 			ConsulK8STag:      c.flagConsulK8STag,
+			K8sConsulPrefix:   c.flagK8sConsulPrefix,
```
adilyse: If we're able to remove the above `syncer` struct field, this also wouldn't be necessary.

catalog/from-k8s/resource.go
```diff
@@ -609,3 +613,10 @@ func (t *serviceEndpointsResource) Delete(key string) error {
 	t.Service.Log.Info("delete endpoint", "key", key)
 	return nil
 }
+
+func (t *ServiceResource) serviceName(name string) string {
```
adilyse: To be just a bit more explicit about what this method does to make it easier to read in-line:

```suggestion
func (t *ServiceResource) prefixServiceName(name string) string {
```

catalog/from-k8s/resource_test.go
```diff
@@ -1770,3 +1770,297 @@ func testService(name string) *apiv1.Service {
...
+	// Verify what we got
+	syncer.Lock()
+	defer syncer.Unlock()
+	actual := syncer.Registrations
+	require.Len(actual, 1)
+	require.Equal("prefixfoo", actual[0].Service.Service)
+	require.Equal("1.2.3.4", actual[0].Service.Address)
+}
+
+//Test externalIP with Prefix
```
adilyse: Nitpick:
```suggestion
// Test externalIP with Prefix
```

catalog/from-k8s/resource_test.go
```diff
@@ -1770,3 +1770,297 @@ func testService(name string) *apiv1.Service {
...
+
+	// Insert an LB service
+	_, err := client.CoreV1().Services(metav1.NamespaceDefault).Create(&apiv1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "foo",
+		},
+
+		Spec: apiv1.ServiceSpec{
+			Type:        apiv1.ServiceTypeLoadBalancer,
+			ExternalIPs: []string{"a", "b"},
```
adilyse: It would probably be good to have the ExternalIPs resemble what someone might usually find in that field (i.e. "2.3.4.5") rather than an arbitrary string.

catalog/from-k8s/resource_test.go
```diff
@@ -1770,3 +1770,297 @@ func testService(name string) *apiv1.Service {
 		},
 	}
 }
+
```
adilyse: The `testService` function above this is a helper function rather than a test. Usually helper functions are at the bottom of the file, so I'd suggest moving the tests around a bit. Additionally, the standard in this file is to have all of the permutations of `loadBalancer`, `nodePort`, etc. tests are grouped together, so it would make more sense to insert these tests near their service type.

adilyse:
Shoot, I still need you to update `resource_test.go` with go fmt. Then I can merge!

s-christoff:
Everything is 💚  🤞 Thank you for all the help.

PR #90, opened 2019-05-01
Add auth method registration in bootstrapping and login on injection

adilyse:
Supports the AuthMethod introduced in Consul 1.5. Does some extra work
to copy the service account volume mount into the init and sidecar
containers since they're not automatically created through Kubernetes.


connect-inject/container_init.go
```diff
@@ -103,14 +124,15 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
 					FieldRef: &corev1.ObjectFieldSelector{FieldPath: "metadata.name"},
 				},
 			},
-		},
-		VolumeMounts: []corev1.VolumeMount{
-			corev1.VolumeMount{
-				Name:      volumeName,
-				MountPath: "/consul/connect-inject",
+			{
+				Name: "NAMESPACE",
```
rboyer:
```suggestion
				Name: "POD_NAMESPACE",
```

connect-inject/container_sidecar.go
```diff
@@ -39,11 +64,19 @@ func (h *Handler) containerSidecar(pod *corev1.Pod) corev1.Container {
...
-const sidecarPreStopCommand = `
+const sidecarPreStopCommandTpl = `
 export CONSUL_HTTP_ADDR="${HOST_IP}:8500"
 /consul/connect-inject/consul services deregister \
+  {{- if . }}
+  -token-file="/consul/connect-inject/acl-token" \
+  {{- end }}
   /consul/connect-inject/service.hcl
+{{- if . }}
+&& /consul/connect-inject/consul \
```
rboyer:
```suggestion
&& /consul/connect-inject/consul logout \
```

connect-inject/container_sidecar.go
```diff
@@ -39,11 +64,19 @@ func (h *Handler) containerSidecar(pod *corev1.Pod) corev1.Container {
...
+  -token-file="/consul/connect-inject/acl-token" \
```
rboyer:
```suggestion
  -token-file="/consul/connect-inject/acl-token"
```

connect-inject/container_sidecar.go
```diff
@@ -1,12 +1,42 @@
...
+func (h *Handler) containerSidecar(pod *corev1.Pod) (corev1.Container, error) {
+	// Create expected volume mounts
+	volMounts := []corev1.VolumeMount{
+		corev1.VolumeMount{
+			Name:      volumeName,
+			MountPath: "/consul/connect-inject",
+		},
+	}
+
+	if h.AuthMethod != "" {
```
rboyer: the only thing that needs the service account mount point is the init container doing `consul login`. I think you can omit this explicit mount here.

subcommand/server-acl-init/command.go
```diff
@@ -332,6 +338,65 @@ service_prefix "" {
 		}
 	}
 
+	// Support ConnectInject using Kuberentes as an auth method
```
rboyer:
```suggestion
	// Support ConnectInject using Kubernetes as an auth method
```

subcommand/server-acl-init/command.go
```diff
@@ -332,6 +338,65 @@ service_prefix "" {
...
+		// Assume the jwt is the first secret attached to the service account
+		saSecret, err := clientset.CoreV1().Secrets(c.flagNamespace).Get(amServiceAccount.Secrets[0].Name, metav1.GetOptions{})
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error getting service account JWT secret: %s", err))
+			return 1
+		}
+
+		// Set up auth method
+		aam := api.ACLAuthMethod{
+			Name: fmt.Sprintf("%s-consul-k8s-auth-method", c.flagReleaseName),
```
rboyer: You may want to add in a `Description` here like you do on the binding rule to indicate that consul-k8s seeded this for you.

subcommand/server-acl-init/command.go
```diff
@@ -332,6 +338,65 @@ service_prefix "" {
...
+			Type: "kubernetes",
+			Config: map[string]interface{}{
+				"Host":              fmt.Sprintf("https://%s:443", k8sService.Spec.ClusterIP),
+				"CACert":            string(saSecret.Data["ca.crt"]),
+				"ServiceAccountJWT": string(saSecret.Data["token"]),
+			},
+		}
+
+		authMethod, _, err := consulClient.ACL().AuthMethodCreate(&aam, &api.WriteOptions{})
```
rboyer: And you said that it fails when you provided `nil` as the second argument here?

connect-inject/container_sidecar.go
adilyse: With the other commands, it would not correctly read the `-token-file` flag coming after the main part of the command. Is that not the case for this one?

subcommand/server-acl-init/command.go
adilyse: No, I was having issues with the binding rule call, line 393. If I use `&api.WriteOptions{}` instead of `nil` for the second argument, it tells me I have an invalid BindName.

connect-inject/container_sidecar.go
rboyer: flags always have to come after the subcommand from what i can tell

subcommand/server-acl-init/command.go
adilyse: Added.

connect-inject/container_init.go
adilyse: Good catch! I was wondering why I wasn't seeing this info...

connect-inject/container_sidecar.go
adilyse: Getting things confused. I need a nap.

connect-inject/container_sidecar.go
adilyse: Removed the sidecar volume mount copying.

subcommand/server-acl-init/command.go
adilyse: Fixed.

PR #96, opened 2019-05-09
Handle error output from central config write command

adilyse:
The addition of the `-cas` flag to the `consul config write` keeps the
system from doing the additional work of writing the central config again
if it already exists. However, if the config does exist, this call
returns an error code. This update handles that case.

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -225,7 +225,7 @@ EOF
   {{- if .AuthMethod }}
   -token-file="/consul/connect-inject/acl-token" \
   {{- end }}
-  /consul/connect-inject/central-config.hcl
+  /consul/connect-inject/central-config.hcl || true
 {{- end }}
 
 /bin/consul services register \


eikenb:
`|| true` is the common pattern for a shell script using `set -e` when you want a command  error to not halt the script. The only caveat with this pattern is that it will ignore all errors  when you may only want to ignore a particular one. Generally this is not really an issue though, but something to keep in mind.

I'd +1 it if I could.

eikenb:
`true` is a required shell builtin in POSIX shells. So the path shouldn't be necessary.

adilyse:
I've tested this out in the docker container where it will be run, and the path works fine. It's a good callout that we'll swallow all errors with this. I'm okay with it for now, but we should keep an eye on it in the future.

mkeeler:
@eikenb Totally. All reasonable systems this would work on. I don't think we needed to change it at all. It was more of a note about some issues I have ran into with different platforms in the past.

PR #98, opened 2019-05-14
Migrate tests to retry.Run dynamic sleep intervals

freddygv:
Currently many tests in the `catalog` package use sleeps with hardcoded durations to wait for registrations. 

This leads to a couple issues:
- Tests can become flaky in CI if registrations take longer than expected to show up
- Hard-coded sleep durations lead to longer test run times

This PR adds retries from `consul/sdk/testutil` over hard-coded sleeps where appropriate. 

Note: This conversion was not made in tests that verify that a sync was **not** performed, such as: TestServiceResource_clusterIPSyncDisabled. 

In those cases sleeps were preserved, since we need to allow for the time typically needed to register the services.

### DIFF
--- a/catalog/from-k8s/resource_test.go
+++ b/catalog/from-k8s/resource_test.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/consul-k8s/helper/controller"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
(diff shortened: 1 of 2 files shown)


lkysow:
@ishustava we should merge this, just need to fix the conflicts.

lkysow:
Closed by #290

PR #100, opened 2019-06-07
Add the option to create an ACL token for snapshot agents

adilyse:
Additionally refactors the create ACL logic out in a reusable method
for most of the group tokens.

Adds a flag to allow the client ACL token to be optional, though it
still defaults to true.

### DIFF
--- a/subcommand/server-acl-init/command.go
+++ b/subcommand/server-acl-init/command.go
@@ -1,6 +1,7 @@
 package serveraclinit
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -20,17 +21,21 @@ import (
 type Command struct {
 	UI cli.Ui
 
-	flags                      *flag.FlagSet
-	k8s                        *k8sflags.K8SFlags
-	flagReleaseName            string
-	flagReplicas               int
-	flagNamespace              string
-	flagAllowDNS               bool
-	flagCreateSyncToken        bool
-	flagCreateInjectAuthMethod bool
-	flagBindingRuleSelector    string
-	flagCreateEntLicenseToken  bool
-	flagLogLevel               string
+	flags                        *flag.FlagSet
+	k8s                          *k8sflags.K8SFlags
+	flagReleaseName              string
+	flagReplicas                 int
+	flagNamespace                string
+	flagAllowDNS                 bool
+	flagCreateClientToken        bool
+	flagCreateSyncToken          bool
+	flagCreateInjectAuthMethod   bool
+	flagBindingRuleSelector      string
+	flagCreateEntLicenseToken    bool
+	flagCreateSnapshotAgentToken bool
+	flagLogLevel                 string
+
+	clientset *kubernetes.Clientset
 
 	once sync.Once
 	help string
@@ -46,6 +51,8 @@ func (c *Command) init() {
 		"Name of Kubernetes namespace where the servers are deployed")
 	c.flags.BoolVar(&c.flagAllowDNS, "allow-dns", false,
 		"Toggle for updating the anonymous token to allow DNS queries to work")
+	c.flags.BoolVar(&c.flagCreateClientToken, "create-client-token", true,
+		"Toggle for creating a client agent token")
 	c.flags.BoolVar(&c.flagCreateSyncToken, "create-sync-token", false,
 		"Toggle for creating a catalog sync token")
 	c.flags.BoolVar(&c.flagCreateInjectAuthMethod, "create-inject-token", false,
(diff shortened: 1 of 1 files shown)


subcommand/server-acl-init/command.go
```diff
@@ -54,6 +61,8 @@ func (c *Command) init() {
 		"Selector string for connectInject ACL Binding Rule")
 	c.flags.BoolVar(&c.flagCreateEntLicenseToken, "create-enterprise-license-token", false,
 		"Toggle for creating a token for the enterprise license job")
+	c.flags.BoolVar(&c.flagCreateSnapshotAgentToken, "create-snapshot-agent-token", false,
+		"Toggle for creating a token for the Consul snapshot agent deployment")
```
lkysow:
```suggestion
		"Toggle for creating a token for the Consul snapshot agent deployment (enterprise only)")
```

subcommand/server-acl-init/command.go
```diff
@@ -291,77 +268,47 @@ service_prefix "" {
...
-		sToken := api.ACLToken{
-			Description: "Catalog Sync Token",
-			Policies:    []*api.ACLTokenPolicyLink{&api.ACLTokenPolicyLink{Name: aclSyncPolicy.Name}},
-		}
-
-		// Create catalog sync token
-		syncToken, _, err := consulClient.ACL().TokenCreate(&sToken, &api.WriteOptions{})
-		if err != nil {
-			c.UI.Error(fmt.Sprintf("Error creating catalog sync token: %s", err))
+	// Create enterprise license token if necessary
```
lkysow: Don't think this comment is necessary since the below if statement says the same thing.

subcommand/server-acl-init/command.go
```diff
@@ -291,77 +268,47 @@ service_prefix "" {
...
-			ObjectMeta: metav1.ObjectMeta{
-				Name: fmt.Sprintf("%s-consul-catalog-sync-acl-token", c.flagReleaseName),
-			},
-			StringData: map[string]string{
-				"token": syncToken.SecretID,
-			},
-		})
-		if err != nil {
-			c.UI.Error(fmt.Sprintf("Error creating catalog sync token secret: %s", err))
+	// Create snapshot agent token if necessary
```
lkysow: same for this comment

subcommand/server-acl-init/command.go
```diff
@@ -401,51 +348,47 @@ service_prefix "" {
...
+	}
 
-		aclEntLicensePolicy, _, err := consulClient.ACL().PolicyCreate(&entLicensePolicy, &api.WriteOptions{})
-		if err != nil {
-			c.UI.Error(fmt.Sprintf("Error creating enterprise license policy: %s", err))
-			return 1
-		}
+	createdPolicy, _, err := consulClient.ACL().PolicyCreate(&policy, &api.WriteOptions{})
+	if err != nil {
+		return errors.New(fmt.Sprintf("Error creating %s policy: %s", name, err))
```
lkysow:
```suggestion
                return fmt.Errorf("Error creating %s policy: %s", name, err)
```

lkysow: If the policy is already created, does this fail? Just thinking if one of these steps fail and the job is re-run, if we'll never get past this step.

subcommand/server-acl-init/command.go
```diff
@@ -401,51 +348,47 @@ service_prefix "" {
...
-		// Create catalog sync token
-		entLicenseToken, _, err := consulClient.ACL().TokenCreate(&eToken, &api.WriteOptions{})
-		if err != nil {
-			c.UI.Error(fmt.Sprintf("Error creating enterprise license token: %s", err))
-			return 1
-		}
+	// Create token
+	createdToken, _, err := consulClient.ACL().TokenCreate(&token, &api.WriteOptions{})
+	if err != nil {
+		return errors.New(fmt.Sprintf("Error creating %s token: %s", name, err))
```
lkysow: use fmt.Errorf

subcommand/server-acl-init/command.go
```diff
@@ -401,51 +348,47 @@ service_prefix "" {
...
+	_, err = c.clientset.CoreV1().Secrets(c.flagNamespace).Create(&apiv1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-consul-%s-acl-token", c.flagReleaseName, name),
+		},
+		StringData: map[string]string{
+			"token": createdToken.SecretID,
+		},
+	})
+	if err != nil {
+		return errors.New(fmt.Sprintf("Error creating %s token secret: %s", name, err))
```
lkysow: fmt.Errorf

subcommand/server-acl-init/command.go
adilyse: Probably. But we currently only create this on server creation, so if a policy already exists with that name, someone manually changed things as everything was getting created which has larger security implications and should be investigated.

subcommand/server-acl-init/command.go
adilyse: Not strictly necessary, but I like using comment lines to break up chunks of code into logical blocks.

subcommand/server-acl-init/command.go
adilyse: See above.

Issue #101, opened 2019-06-10
Syncing services from many k8s clusters to single consul


mgrzybowski:
Hi
   I have consul servers outside k8s clusters and i'm traing to register services from many k8s cluster to this single central consul directory. It seems that the current sync service from k8s to consul is limited to single k8s cluster (in the same consul datacenter ) . 

From what i understand consul node name for sync is hardcoded:
https://github.com/hashicorp/consul-k8s/blob/aa7d80a934a1c2faf5ffe9b3bd0abba4e8754d14/catalog/from-k8s/resource.go#L258

What if node name for sync could be changed in config ( each k8s have its own unique node for service sync ) ? Will this allow to sync services from many k8s to single consul ( to the same consul datacenter )   ?

Kind Regards
    Marek Grzybowski

lkysow:
Hi mgrzybowski, why does this cause you to not be able to sync from multiple k8s clusters?

mgrzybowski:
Hi lkysow 
  The same consul node name  "k8s-sync" exists in each of  k8s cluster. You can not register the same consul node multiple times in the consul directory.

lkysow:
Ahh I see, thanks for clarifying.

praymann:
I'm not sure this is an issue.

I'm in the process of setting up this exact same situation:

> [...] consul servers outside k8s clusters and [...] register services from many k8s cluster to this single central consul directory

Given a unique tag via `consul-k8s-tag` as mentioned in https://github.com/hashicorp/consul-k8s/pull/103#issuecomment-505426930 the sync appears to "just work", no conflict of Nodes.

![image](https://user-images.githubusercontent.com/9577818/66351003-55411800-e922-11e9-8cc8-9f142abb13bf.png)

This issues calls out [L258](https://github.com/hashicorp/consul-k8s/blob/aa7d80a934a1c2faf5ffe9b3bd0abba4e8754d14/catalog/from-k8s/resource.go#L258) as the root of the issue, but I think [L257](https://github.com/hashicorp/consul-k8s/blob/aa7d80a934a1c2faf5ffe9b3bd0abba4e8754d14/catalog/from-k8s/resource.go#L257) makes this "just work". Document states:

[...]

PR #103, opened 2019-06-19
Add option to name K8s Node in Consul

Mikulas:
This change adds an Consul node name option to sync-catalog, which resolves https://github.com/hashicorp/consul-k8s/issues/101

I'm not sure if the changes I made make much sense, but I will try the build tomorrow to at least see if it has the desired functionality.

Next up would be adding tests and updating the helm chart with an option to specify the node name.

As @lkysow mentioned in #101 it would be nice to default to kubernetes cluster name, but the closest concept there is to a cluster name in kubernetes is `cluster.local`, which is afaik rarely changed anyway. I've instead kept the original `k8s-sync` as the default.

### DIFF
--- a/catalog/from-k8s/resource.go
+++ b/catalog/from-k8s/resource.go
@@ -255,7 +255,7 @@ func (t *ServiceResource) generateRegistrations(key string) {
 	// shallow copied for each instance.
 	baseNode := consulapi.CatalogRegistration{
 		SkipNodeUpdate: true,
-		Node:           "k8s-sync",
+		Node:           t.Syncer.Node(),
 		Address:        "127.0.0.1",
 		NodeMeta: map[string]string{
 			ConsulSourceKey: ConsulSourceValue,
--- a/catalog/from-k8s/resource_test.go
+++ b/catalog/from-k8s/resource_test.go
@@ -27,7 +27,7 @@ func TestServiceResource_createDelete(t *testing.T) {
 	t.Parallel()
 	require := require.New(t)
 	client := fake.NewSimpleClientset()
-	syncer := &TestSyncer{}
+	syncer := NewTestSyncer()
 
 	// Start the controller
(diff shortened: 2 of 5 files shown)


Mikulas:
Hi @adilyse, do those changes make sense for the project? Please let me know if there is anything I can do to get this merged eventually. :)

Mikulas:
I've tested a build of this in our dev clusters and it mostly works as expected.

Unfortunately even if each cluster creates unique node names (as I've done in #103) consul-k8s still does not support multiple clusters by default. This logic https://github.com/hashicorp/consul-k8s/blob/master/catalog/from-k8s/syncer.go#L181 takes all services labeled `k8s` and removes all not known to the current cluster. This means services are flapping up and down, depending on which operator ran last.

It may make sense to warn users when also not setting the `consul-k8s-tag` flag to a unique value.

nerzhul:
@Mikulas i would like to see it merged too.
An interesting point to have it named is running multiple consul-k8s on a cluster for syncing multiple namespaces with differents agents depending on the client consuming the k8s. 

In my company we have a common consul for the whole organization and we have multiple business units on the k8s and they may need consul sync or not, then we can see multiple syncer depending on the business unit populating the consul

pnovotnak:
Anyone home?

Mikulas:
I still think this would be a nifty improvement and I am quite upset hashicorp never acknowledged this merge request.

However, we have since migrated to a different patch (not on github) that exposes kubernetes pod names instead of the hardcoded `k8s-sync` or the per-cluster configurable string from this merge request.

lkysow:
Thanks! This was merged as part of https://github.com/hashicorp/consul-k8s/pull/312. The node name can be set in Helm:
```yaml
syncCatalog:
    consulNodeName: "k8s-sync"
```

PR #108, opened 2019-06-25
Make sure the kubernetes service is looked up in the default namespace

HansK-p:
During the bootstrap and initialization process, the kubernetes service is looked up in the same namespace as the Consul servers are running. This works well as long as they actually are in the same namespace (the default namespace), but fails when Consul servers are placed in a custom namespace.
 
This patch makes sure that the kubernetes service is always looked up in the default namespace. The code is only run with ACLs are enabled and automatic inject using Kubernetes as Auth method - as far as I can see and I've only validated that the patch works with the Helm chart.


HansK-p:
It turned out that the previous pull request was related to exactly the same issue. There is no reason to have to pull requests with (close to) exactly the same change.

PR #109, opened 2019-06-26
add resources to injector for sidecar/init container

wuvs:
if resource quotas for namespace are set, init container and sidecar pod require to set cpu and memory resources, else they are not able to start. the init container uses the same resources as are specified for the sidecar pod. the resources for the injector are disabled by default.

related PR https://github.com/hashicorp/consul-helm/pull/200


lkysow:
Merged as part of #267

PR #113, opened 2019-07-02
Add -create-mesh-gateway-token flag

lkysow:
This flag will be used when bootstrapping Connect mesh gateways.


subcommand/server-acl-init/command.go
```diff
@@ -441,4 +451,8 @@ service "consul-snapshot" {
    policy = "write"
 }`
 
+const meshGatewayRules = `service_prefix "" {
```
lkysow: @mkeeler is this the correct ACL?

lkysow:
He took a look already actually! I've documented the requirement in the
helm chart about the name as well.

PR #115, opened 2019-07-08
Add tag annotation to Connect service registration

adilyse:
Allows users to add Consul tags to their services via a Kubernetes
annotation.

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -2,6 +2,7 @@ package connectinject
 
 import (
 	"bytes"
+	"encoding/json"
 	"strings"
 	"text/template"
 
@@ -15,6 +16,7 @@ type initContainerCommandData struct {
 	AuthMethod      string
 	CentralConfig   bool
 	Upstreams       []initContainerCommandUpstreamData
+	Tags            string
 }
 
 type initContainerCommandUpstreamData struct {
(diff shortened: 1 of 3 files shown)


connect-inject/handler.go
```diff
@@ -52,6 +52,10 @@ const (
 	// is the local port in the pod that the listener will bind to. It can
 	// be a named port.
 	annotationUpstreams = "consul.hashicorp.com/connect-service-upstreams"
+
+	// annotationTags is a list of tags to register with the service
+	// this is specified as a comma separated list e.g. abc,123
+	annotationTags = "consul.hashicorp.com/connect-service-tags"
```
lkysow: is it specific to connect? Why not `service-tags`?

connect-inject/container_init.go
```diff
@@ -47,6 +49,17 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
 		}
 	}
 
+	// If tags are specified split the string into an array and create
+	// the tags string
+	if raw, ok := pod.Annotations[annotationTags]; ok && raw != "" {
+		tags := strings.Split(raw, ",")
+
+		// Create json array from the annotations
+		jsonTags, _ := json.Marshal(tags)
```
lkysow: It's a JSON array but then it's being templated into HCL. Are there any json marshaling vs. HCL issues we might run  into here?

lkysow: I don't think you should swallow the error here. Would be useful to surface instead I think.

connect-inject/handler.go
adilyse: This is specific for Connect, since it's only part of the connect injector.

connect-inject/container_init.go
adilyse: I've added a log message for this, but I don't think a bad annotation should stop connect injection. This will at least make it possible for folks to see what's happening if it's not what they expect in this situation.

adilyse: From the [hcl readme](https://github.com/hashicorp/hcl): "Arrays can be made by wrapping it in []. Example: ["foo", "bar", 42]", so we should be good for formatting.

Issue #106, opened 2019-06-25
Allow installation of consul-k8s into a non-default namespace


paulbouwer:
# Issue 

When installing Consul into a Kubernetes cluster via the [Consul Helm Chart](https://github.com/hashicorp/consul-helm) and targeting a namespace that is not **default**, there is the following issue.

The `consul-k8s server-acl-init` command executed by the `server-acl-init-job` attempts to get the Kubernetes api-server service `kubernetes.default.svc`. 

See https://github.com/hashicorp/consul-k8s/blob/master/subcommand/server-acl-init/command.go#L347

This code attempts to obtain the service from the namespace specified to the helm chart. If this namespace value is **default**, then everything just works. If this namespace value is something like **consul**, then the code fails. 

> While running in a Pod, the Kubernetes apiserver is accessible via a Service named `kubernetes` in the `default` namespace. Therefore, Pods can use the `kubernetes.default.svc` hostname to query the API server. Official client libraries do this automatically.
> 
> https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#accessing-the-api-from-a-pod

# Solution

Hardcode the namespace searched by this specific piece of code to `default`, since that is always where the service will be.

lkysow:
Fixed by #107

PR #117, opened 2019-07-09
Add a flag for the kubernetes service namespace name

joe-pll:
This solves #106 

The `kubernetes` service is present only in the `default` namespace. When deploying consul-k8s in a different namespace the *acl-init-server* job fails because it can't find the endpoint.

With the merge request the namespace used by default for the service is `default` but it can be overwritten by another namespace name.


lkysow:
Hi Joe,
Thanks for the PR! This was fixed by https://github.com/hashicorp/consul-k8s/pull/107 (similar solution to yours) so I'm going to close this PR.

PR #118, opened 2019-07-12
Service annotation

nicholasjackson:
This PR adds the ability to add meta data to services registered with consul connect injector. metadata is a common way of configuring service splitting.

```
      annotations:
        consul.hashicorp.com/connect-service-meta: "version:1, foo:bar"
        consul.hashicorp.com/connect-service-protocol: "http"
        consul.hashicorp.com/connect-service-tags: "v1"
        prometheus.io/port: "9102"
        prometheus.io/scrape: "true"
```

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -17,6 +17,7 @@ type initContainerCommandData struct {
 	CentralConfig   bool
 	Upstreams       []initContainerCommandUpstreamData
 	Tags            string
+	Meta            map[string]string
 }
 
 type initContainerCommandUpstreamData struct {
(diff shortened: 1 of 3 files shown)


lkysow:
I'd get @adilyse to take a look too since I'm just getting familiar with this.

banks:
@nicholasjackson would it be cleaner to use separate annotations per meta label?

E.g. 

```
  annotations:
    consul.hashicorp.com/service-meta-version: 1
    consul.hashicorp.com/service-meta-foo: bar
```

Note I also propose dropping connect prefix for meta and tags since they are not really connect specific and are properties of any service in Consul whether using connect or now. I guess in practice the will only be picked up if you have connect injector enabled though so I'm not super sure which is the cleanest?

lkysow:
What about
```
 annotations:
    consul.hashicorp.com/service-meta/version: 1
    consul.hashicorp.com/service-meta/foo: bar
```
Looks a bit cleaner

banks:
I don't mind that but i have a nagging thought that Kube expectes them to be in the form `namespace/name` - does adding more slashes violate some assumption? Is that idiomatic?

lkysow:
> I don't mind that but i have a nagging thought that Kube expectes them to be in the form `namespace/name` - does adding more slashes violate some assumption? Is that idiomatic?

Yeah looking through things I think you're right

nicholasjackson:
Superseded by https://github.com/hashicorp/consul-k8s/pull/129

PR #121, opened 2019-07-24
Fix service id generation

lohmag:
Fix issue when you have multiple services endpoints under one service.  
For example:  
```
 postgres-db --
              |--  postgres-db-sessions
              |--  postgres-db-users
```
with different consul.hashicorp.com/service-meta-XXX, but with the same ip address.  
Because hash now calculated only by ip address - the service id is the same all the time and they overwrite each other. I added tags to the equation.


lohmag:
Added better algoritm..
Bump

lkysow:
Hi Nikita, thanks for updating this PR. Can you explain in more detail what your setup looks like? What do you mean by having multiple services endpoints under one service?

lohmag:
> Hi Nikita, thanks for updating this PR. Can you explain in more detail what your setup looks like? What do you mean by having multiple services endpoints under one service?

I have multiple databases on different hosts in prod. Needed service discovers by servicename and tag.
But in dev it is only one database in k8s which includes all out bases and so IP is same for each servicename. For this case we use consul-k8s.
I have such structure:
```
        "Service": {
            "ID": "postgres-db-01f18b3819b8",
            "Service": "postgres-db",
            "Tags": [
                "deffault"
            ]
...
        "Service": {
            "ID": "postgres-db-4e51454a4c91",
            "Service": "postgres-db",
            "Tags": [
                "messages"
            ]
```
Main target to have different Meta, becouse it includes login/password.
Multiple services could be created with use of headless k8s service, like that:
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    consul.hashicorp.com/service-meta-dbname: messages
    consul.hashicorp.com/service-meta-password: password
    consul.hashicorp.com/service-meta-user: username
    consul.hashicorp.com/service-name: postgres-db
    consul.hashicorp.com/service-sync: "true"
    consul.hashicorp.com/service-tags: 'messages'
spec:
  clusterIP: None
  ports:
  - name: postgresql
    port: 5432
    protocol: TCP
    targetPort: postgresql
  selector:
    app: postgresql
    release: dev-postgresql
    role: master
  type: ClusterIP
```

david-yu:
Closing as this PR is now quite old. Please re-open a new PR if you would like us to review.

david-yu:
Closing as this PR is now quite old. Please re-open a new PR if you would like us to review.

PR #123, opened 2019-07-26
Add the —max-obj-name-len parameter to the injected Envoy container

### LINKED ISSUE #120: Run Envoy with `--max-obj-name-len 256`
This will be necessary for Consul 1.6 as the cluster names in use are now likely to exceed the 60 char default.

mkeeler:
Fixes #120

I think I added this properly but let me know if it should have been somewhere else.

### DIFF
--- a/connect-inject/container_sidecar.go
+++ b/connect-inject/container_sidecar.go
@@ -49,6 +49,7 @@ func (h *Handler) containerSidecar(pod *corev1.Pod) (corev1.Container, error) {
 		},
 		Command: []string{
 			"envoy",
+			"--max-obj-name-len", "256",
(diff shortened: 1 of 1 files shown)


lkysow:
Since this is deprecated in Envoy 1.11.0 (https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated.html?highlight=max%20obj%20name%20len#version-1-11-0-july-11-2019) I was thinking about adding some version detection here via a new `-envoy-version` flag. What do you think?

mkeeler:
Yeah, version detection would be good. Having to ever set that seemed bad so I am glad they made it useless.

PR #125, opened 2019-07-30
Sync changes as they occur in k8s and reduce full sync default interval

ShimmerGlass:
In addition to registering all services periodically, now sync new
services as they appear, allowing to increase the full sync interval and
reduce the overall number of catalog register call to Consul servers.
This makes a difference for kube clusters with hundreds of instances
without delaying updates.

### DIFF
--- a/catalog/from-k8s/syncer.go
+++ b/catalog/from-k8s/syncer.go
@@ -2,6 +2,7 @@ package catalog
 
 import (
 	"context"
+	"reflect"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pierresouchay:
@alvin-huang @lkysow would it be possible to get a review? it really helps on large clusters improving performance

david-yu:
Closing as there is a duplicate PR that has attempted to address this https://github.com/hashicorp/consul-k8s/pull/228

PR #128, opened 2019-08-16
Fixes #127 - Allow ClusterIP services with ExternalIPs

### LINKED ISSUE #127: syncClusterIPService false does not allow me to use something like kuberouter for externalip
This line:
https://github.com/hashicorp/consul-k8s/blob/3c0cf6872a839a753fa7a0775dca9a7c3fa10704/catalog/from-k8s/resource.go#L189

Should be moved to the case for ClusterIP: 
https://github.com/hashicorp/consul-k8s/blob/3c0cf6872a839a753fa7a0775dca9a7c3fa10704/catalog/from-k8s/resource.go#L474

In my case we create service VIPs with kube-router and use type ClusterIP. I don't want to sync any cluster IP services, only the ones with externalIPs.

I'll add a PR and tests for this.

filintod:
This fixes the issue where I cannot use false to not sync clusterIPs 
but still want to export the ones with ExternalIPs like when using kube-router

### DIFF
--- a/catalog/from-k8s/resource.go
+++ b/catalog/from-k8s/resource.go
@@ -185,8 +185,8 @@ func (t *ServiceResource) shouldSync(svc *apiv1.Service) bool {
 		return false
 	}
 
-	// Ignore ClusterIP services if ClusterIP sync is disabled
-	if svc.Spec.Type == apiv1.ServiceTypeClusterIP && !t.ClusterIPSync {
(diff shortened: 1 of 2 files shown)


filintod:
@lkysow could you take a look at this change, it's been there for a while and I would not like to keep a separate repo for it if possible. Thanks

PR #135, opened 2019-09-18
All creation of metadata and tags from k8s annotations

nicholasjackson:
This PR is a roll up of previous PRs as I found a bug. The original code was not setting the tags and service metadata for the proxy. While the tests were passing for this as the test only searches for the existence of tags in the command. Since this data was present on the service config the test passed however for Consul L7 config tags and metadata need to be present on the proxy.

**Content:**
Add annotation to allow the configuration of service tags in Consul

`consul.hashicorp.com/service-tags`, value is a comma separated string of tags to add to the service

```
      annotations:
        "consul.hashicorp.com/service-tags": "v1, k8s"
```

Add annotation to allow the configuration of service metadata in Consul service catalog

`consul.hashicorp.com/service-meta-<key>` individual annotations are specified with the metadata key in the annotation name. Key is the name of the metadata key to store in the Consul service catalog. Value of the annotation is the value set against the metadata.

```
      annotations:
        "consul.hashicorp.com/service-meta-version": "1"
        "consul.hashicorp.com/service-meta-foo": "bar"
```

A test docker container for this branch can be found at `nicholasjackson/consul-k8s-dev:beta`

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -181,6 +181,16 @@ services {
   kind = "connect-proxy"
   address = "${POD_IP}"
   port = 20000
+  {{- if .Tags}}
+  tags = {{.Tags}}
+  {{- end}}
+  {{- if .Meta}}
+  meta = {
+    {{- range $key, $value := .Meta }}
+    {{$key}} = "{{$value}}"
+    {{- end }}
+  }
+  {{- end}}
 
   proxy {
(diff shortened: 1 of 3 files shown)


anubhavmishra:
@lkysow excited for this to be merged! I am using Nic's branch right now and it works well.

lkysow:
Closed by #141

PR #143, opened 2019-10-03
Idempotent server-acl-init command

### LINKED ISSUE #138: server-acl-init should be idempotent
Currently, if any step in the `server-acl-init` command fails, then when the command is re-run it will never succeed.

This is because it first tries to bootstrap the ACL system. If it's already been bootstrapped then this will fail and the command will exit without performing the rest of its tasks (creating other acl tokens).

Not only should it be idempotent but it should continue executing until its completed its job. If it can't connect to Consul servers or it gets an error it should just retry.

lkysow:
This PR fixes #138 and https://github.com/hashicorp/consul-helm/issues/186

* The `server-acl-init` command would previously exit on any failure. Now it runs indefinitely, retrying each of its tasks until they are complete. It will retry every 1s.
* It can be run before Consul servers are up and will wait for a leader to be elected.
* It is now idempotent and will handle objects in Consul and Kubernetes that have already been created, either by bypassing those steps or updating the values (e.g. if the bootstrap Secret already exists it will update its data) 
* I've added tests for this behaviour using the Consul testagent and also a simple httpserver mock

### DIFF
--- a/subcommand/server-acl-init/command_test.go
+++ b/subcommand/server-acl-init/command_test.go
@@ -1,0 +1,964 @@
+package serveraclinit
+
+import (
+	"encoding/base64"
+	"fmt"
+	"github.com/hashicorp/consul/agent"
+	"github.com/hashicorp/consul/api"
+	"github.com/mitchellh/cli"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	"math/rand"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strconv"
+	"testing"
+	"time"
+)
+
+var ns = "default"
+var releaseName = "release-name"
+
+// Set up test consul agent and kubernetes clusters with
+func completeSetup(t *testing.T) (*fake.Clientset, *agent.TestAgent) {
+	require := require.New(t)
+	k8s := fake.NewSimpleClientset()
+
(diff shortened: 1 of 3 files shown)


subcommand/server-acl-init/command_test.go
```diff
@@ -0,0 +1,984 @@
...
+			})
+			require.Equal(0, responseCode, ui.ErrorWriter.String())
+
+			// Get the bootstrap token from the kube secret so we can make API calls
+			// to Consul.
+			bootstrapSecret, err := k8s.CoreV1().Secrets(ns).Get(fmt.Sprintf("%s-consul-bootstrap-acl-token", releaseName), metav1.GetOptions{})
+			require.NoError(err)
+			require.NotNil(bootstrapSecret)
+			bootToken, ok := bootstrapSecret.StringData["token"]
+			require.True(ok)
```
ishustava: I noticed that this gets repeated a few times. Although I generally don't think that tests should be DRY, it might be worth extracting it into a helper.

subcommand/server-acl-init/command_test.go
```diff
@@ -0,0 +1,984 @@
...
+
+	consulURL, err := url.Parse("http://" + a.HTTPAddr())
+	require.NoError(err)
+	port, err := strconv.Atoi(consulURL.Port())
+	require.NoError(err)
+
+	// Create Consul server Pod.
+	_, err = k8s.CoreV1().Pods(ns).Create(&v1.Pod{
+		ObjectMeta: v12.ObjectMeta{
+			Name: "release-name-consul-server-0",
```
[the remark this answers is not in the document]
lkysow: Fixed

subcommand/server-acl-init/command_test.go
lkysow: Good call, updated.

PR #144, opened 2019-10-07
Fix bug looking for boot token in StringData field

lkysow:
Previously, we would re-use an existing bootstrap token if the Secret
already existed, by looking for the token in the StringData field of the
Secret.

Unfortunately, the StringData field of a Kubernetes Secret is only used as a
convenience when writing a Kubernetes Secret. When reading the secret it
is not populated and is converted to a []byte and moved to the Data
field. This conversion happens in a real Kubernetes cluster but not with
the testing 'fakeclient', hence not being caught by our tests.

This change removes all use of StringData to avoid confusion.

### DIFF
--- a/subcommand/server-acl-init/command.go
+++ b/subcommand/server-acl-init/command.go
@@ -161,7 +161,7 @@ func (c *Command) Run(args []string) int {
 	consulClient, err = api.NewClient(&api.Config{
 		Address: firstServerAddr,
 		Scheme:  "http",
-		Token:   bootstrapToken,
+		Token:   string(bootstrapToken),
 	})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lkysow:
@ishustava this is why you were getting that error about the secret not being found in the token field.

ishustava:
@lkysow 

My error was actually because the secret didn't exist in kubernetes due to user error 😆  

This change makes sense, but I can't come up with steps to reproduce.

When upgrading from 0.9.0 to 0.10.0, the upgrade fails because helm generates a new yaml for the ACL init job, but the update of the job resource fails in k8s because the job spec is immutable.

Trying to upgrade from 0.8.0 to 0.10.0 results in the ACL init job discovering the old server IP. After that it runs infinitely trying to connect to the old server IP and getting a timeout error.

PR #148, opened 2019-10-08
Fix #146 add annotation for add http check

### LINKED ISSUE #146: connect sidecar add healthcheck in annotation
Now, consul sidecar check services through ip:port. If service listening port, but don't have success probe, consul direct traffic in the dead pod. 
Add some annotations such as: 
```"consul.hashicorp.com/service-check": "http://127.0.0.1:8080/ping/serice/alive':10s"     ```

May be here add checks

https://github.com/hashicorp/consul-k8s/blob/8b83fd98f2bbaeed810954bc9a7298db1efe4d14/connect-inject/container_init.go#L245-L261

example:
```
  checks {
    name = "{{ .ServiceName }} check"
    http = "{{ .checkPath }}"
    interval = "{{ .checkIntervas }}"
  }
```

### PR DESCRIPTION
**Pull request by Veres72**
I try to resolve #146 
Annotation, that add service HTTP check with POD_IP and ServicePort. 
It allows adding HTTP checks because by default services check only by TCP check.

### CODE REVIEW

**Comment by lkysow** [comment]
Might solve https://github.com/hashicorp/consul-k8s/issues/155

**Comment by omerlh** [comment]
Any update regarding this PR? Will it be merged?

File: connect-inject/container_init.go [hunk 1]
```diff
@@ -257,6 +259,13 @@ services {
     {{- end }}
   }
   {{- end}}
+  {{- if .ServiceCheck}}
+  checks {
```
anasokan: Is this 'check' or 'checks'?

pbusko:
Any chance that this PR will be merged?

lkysow:
Hi @pbusko and others, we need to do a broad review of our approach to health checks so I don't think this will be merged until we get to that. The things we need to think about are:
* should we support consul-specific health checks or do we want to hook into the kube readiness or liveliness checks so users don't need to configure both
* if we hook into readiness/liveness probes which ones do we use and how does this effect connect traffic
* how do we integrate with the external paths support for exposing services through the envoy proxy

lkysow:
Hi all, with the release of consul-helm 0.26.0 we support syncing kube readiness status to Consul. So instead of setting checks via annotations in the PR can you try creating kube readiness probes? That way Consul and Kube will agree on the readiness status.

lkysow:
Closing because setting kube readiness probes is the recommended way.

Issue #110, opened 2019-06-26
Rewrite services being injected to k8s to be lowercase because K8S DNS implementation sucks


ejmarten:
I have my external consul doing a two way sync with my k8s (v1.15.0) service registry/coreDNS, 
On inspecting the log for the consul-sync-catalog pod, I see an entry like the following for each service registered in consul with uppercase characters.  

I jumped in the external-dns channel in k8s slack, and was told that kubernetes does not support uppercase names.  

Well crud.  Now I have to either write a wrapper for k8s-consul, create a docker image with a non-standard coredns binary that supports the rewrite plugin, or hotfix hundreds of consul services...

Guess which one I refuse to do?

2019-06-26T19:55:09.033Z [WARN ] to-k8s/sink: error creating service: name=consul-ORDER-API error="Service "consul-ORDER-API" is invalid: [metadata.name: Invalid value: "consul-ORDER-API": a DNS-1035 label must consist of lower case alphanumeric characters or '-', st
art with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?'), spec.externalName: Invalid value: "ORDER-API.service.cluster.local": a DNS-1123 subdomain must consist
 of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')]"

I'm working on a workaround, but this is most definitely a bug, I'll share my results if I figure it out.  

![image](https://user-images.githubusercontent.com/29284062/60211235-424b8180-9824-11e9-8538-4531ba568986.png)

lkysow:
I think we'll have to register them as lower case then. Note that in Consul registering services is case sensitive (you can have two services `foo` and `FOO` https://github.com/hashicorp/consul/issues/5707) but then most of the endpoints like consul DNS will only return one of them.

ejmarten:
This might work.  I'll test and submit a PR if it does.  

[root@dev]# grep -rl ToLower catalog/from-consul/source.go
catalog/from-consul/source.go
[root@dev]# grep -rn ToLower catalog/from-consul/source.go
[...]

PR #149, opened 2019-10-08
sync-catalog tok8s lowercase service names

lkysow:
When syncing from Consul to Kubernetes, lowercase the service names so
they're valid Kubernetes services.
Fixes https://github.com/hashicorp/consul-k8s/issues/110

### DIFF
--- a/catalog/from-consul/sink.go
+++ b/catalog/from-consul/sink.go
@@ -2,6 +2,7 @@ package catalog
 
 import (
 	"context"
+	"strings"
 	"sync"
 	"time"
 
@@ -56,6 +57,8 @@ type K8SSink struct {
 	// sourceServices holds Consul services that should be synced to Kube.
 	// It maps from Consul service names to Consul DNS entry, e.g.
 	// foo => foo.service.consul. It's populated from the Consul API.
+	// We lowercase the Consul service names and DNS entries
+	// because Kube names must be lowercase.
 	sourceServices map[string]string
 
 	// keyToName maps from Kube controller keys to Kube service names.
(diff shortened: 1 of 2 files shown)


catalog/from-consul/source.go
```diff
@@ -67,10 +69,40 @@ func (s *Source) Run(ctx context.Context) {
 					break
 				}
 			}
+			if k8s {
+				continue
+			}
 
-			if !k8s {
-				services[s.Prefix+name] = fmt.Sprintf("%s.service.%s", name, s.Domain)
+			// Consul service names have no constraints, they can have any
```
lkysow: I'm not sure if this should be done here or when we're about to create the service in Kubernetes in sink.go.

I put it here because in sink.go, all the maps use the service name as keys so I was worried they'd get out of sync

ishustava: I think sink.go is a more appropriate place to do it since it seems to be more "k8s-aware".

lkysow:
@ishustava after looking at this again I've greatly simplified the implementation:
1. `source.go` doesn't know anything about the lowercasing, I moved that logic into sink (as suggested)
1. in sink.go, all we do is intercept the services in `SetServices()` and lowercase them. This way we're only ever dealing with lowercase names (and keys) inside sink.go
1. I removed the logging that we're lowercasing the services because it would be logged every sync period and also because it will be obvious that they have been lowercased simply by looking at Kubernetes
1. I removed the detection and logging that the name is invalid (in the case of special characters that are allowed in Consul but not in Kube for example) because during the Create() call to Kubernetes, we'll just get the error there and the registration will fail.

PR #152, opened 2019-10-10
Delete completed job command and fix server-acl-init

lkysow:
Add new `delete-completed-job` command that is used to delete the `server-acl-init` Kubernetes Job once it's completed. This enables us to keep the `server-acl-init` job as **not** a Helm hook. When it was a helm hook it got deleted by Helm which was great, however as a helm hook it caused deadlock when users ran helm with `--wait` because it needs to run concurrently with everything coming up.

This command will run in its own Job and clean up the `server-acl-init` job.

Also fixes server-acl-init as follows:
- previously, if already bootstrapped it would still try to pass out the bootstrap token and reload all the servers. Now it assumes that if the bootstrap Secret is created, all the servers are bootstrapped. I think it's a safe assumption because the Job won't complete until it performs all those actions. If we were to validate that each step in that chain had completed, the complexity of the job would increase immensely. So we're just shortcutting and assuming that if the bootstrap token exists then we can skip bootstrapping
- fixes a bug where even if the ACL Tokens for the other components existed (e.g. client or sync-catalog) we'd try to generate new tokens and update the secrets. This is unnecessary if the Secret already exists and was previously failing because the job didn't have permissions to update secrets.
- adds retry tests to all the tests where we re-run the command and expect it to be successful

### DIFF
--- a/subcommand/delete-completed-job/command.go
+++ b/subcommand/delete-completed-job/command.go
@@ -25,6 +25,7 @@ type Command struct {
 	flags         *flag.FlagSet
 	k8s           *k8sflags.K8SFlags
 	flagNamespace string
+	flagTimeout string
 
 	once      sync.Once
 	help      string
@@ -40,6 +41,8 @@ func (c *Command) init() {
 	c.k8s = &k8sflags.K8SFlags{}
 	c.flags.StringVar(&c.flagNamespace, "k8s-namespace", "",
 		"Name of Kubernetes namespace where the job is deployed")
+	c.flags.StringVar(&c.flagTimeout, "timeout", "10m",
+		"How long we'll wait for the job to complete before timing out, e.g. 1ms, 2s, 3m")
 	flags.Merge(c.flags, c.k8s.Flags())
 	c.help = flags.Usage(help, c.flags)
 
@@ -59,10 +62,20 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 	if len(c.flags.Args()) != 1 {
-		c.UI.Error(fmt.Sprintf("Must have one arg: the job name to delete."))
+		c.UI.Error("Must have one arg: the job name to delete.")
 		return 1
 	}
 	jobName := c.flags.Args()[0]
+	if c.flagNamespace == "" {
+		c.UI.Error("Must set flag -k8s-namespace")
+		return 1
+	}
+	timeout, err := time.ParseDuration(c.flagTimeout)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("%q is not a valid timeout: %s", c.flagTimeout, err))
+		return 1
+	}
+	start := time.Now()
 
 	// c.k8sclient might already be set in a test.
 	if c.k8sClient == nil {
@@ -102,13 +115,18 @@ func (c *Command) Run(args []string) int {
 			break
 		}
 
-		// If its reached its backoff limit then it will never complete,
-		// exit and delete ourselves.
+		// If its reached its backoff limit then it will never complete.
 		for _, condition := range job.Status.Conditions {
 			if condition.Type == v1.JobFailed && condition.Reason == "BackoffLimitExceeded" {
 				logger.Warn(fmt.Sprintf("job %q has reached its backoff limit and will never complete", jobName))
 				return 0
 			}
+		}
+
+		// If we've reached our timeout, exit.
+		if time.Now().After(start.Add(timeout)) {
+			logger.Warn(fmt.Sprintf("timeout %q has been reached, exiting without deleting job", timeout))
+			return 0
 		}
 
 		logger.Info(fmt.Sprintf("job %q has not yet succeeded, waiting %v", jobName, c.retryDuration))
@@ -119,7 +137,7 @@ func (c *Command) Run(args []string) int {
 	// ourselves.
 	logger.Info(fmt.Sprintf("job %q has succeeded, deleting", jobName))
 	propagationPolicy := metav1.DeletePropagationForeground
-	err := c.k8sClient.BatchV1().Jobs(c.flagNamespace).Delete(jobName, &metav1.DeleteOptions{
+	err = c.k8sClient.BatchV1().Jobs(c.flagNamespace).Delete(jobName, &metav1.DeleteOptions{
 		// Needed so that the underlying pods are also deleted.
 		PropagationPolicy: &propagationPolicy,
 	})
--- a/subcommand/delete-completed-job/command.go
+++ b/subcommand/delete-completed-job/command.go
@@ -1,0 +1,147 @@
+package deletecompletedjob
+
+import (
+	"flag"
+	"fmt"
+	"github.com/hashicorp/consul-k8s/subcommand"
+	k8sflags "github.com/hashicorp/consul-k8s/subcommand/flags"
+	"github.com/hashicorp/consul/command/flags"
+	"github.com/hashicorp/go-hclog"
+	"github.com/mitchellh/cli"
+	v1 "k8s.io/api/batch/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"os"
+	"sync"
+	"time"
+)
+
+// Command is the command for deleting completed jobs.
+type Command struct {
+	UI cli.Ui
+
+	flags         *flag.FlagSet
+	k8s           *k8sflags.K8SFlags
+	flagNamespace string
+
+	once      sync.Once
+	help      string
+	k8sClient kubernetes.Interface
+
+	// retryDuration is how often we'll retry deletion.
+	retryDuration time.Duration
+}
+
+func (c *Command) init() {
+	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
+
+	c.k8s = &k8sflags.K8SFlags{}
+	c.flags.StringVar(&c.flagNamespace, "k8s-namespace", "",
+		"Name of Kubernetes namespace where the job is deployed")
+	flags.Merge(c.flags, c.k8s.Flags())
+	c.help = flags.Usage(help, c.flags)
+
+	// Default retry to 5s. This is exposed for setting in tests.
+	if c.retryDuration == 0 {
+		c.retryDuration = 5 * time.Second
+	}
+}
+
+// Run will attempt to delete the job once it succeeds. If the job hits its
+// backoff limit, it will give up deleting it.
+func (c *Command) Run(args []string) int {
+	c.once.Do(c.init)
+
+	// Validate command.
+	if err := c.flags.Parse(args); err != nil {
+		return 1
+	}
+	if len(c.flags.Args()) != 1 {
+		c.UI.Error(fmt.Sprintf("Must have one arg: the job name to delete."))
+		return 1
+	}
+	jobName := c.flags.Args()[0]
+
+	// c.k8sclient might already be set in a test.
+	if c.k8sClient == nil {
+		config, err := subcommand.K8SConfig(c.k8s.KubeConfig())
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error retrieving Kubernetes auth: %s", err))
+			return 1
+		}
+
+		c.k8sClient, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error initializing Kubernetes client: %s", err))
+			return 1
+		}
+	}
+
+	logger := hclog.New(&hclog.LoggerOptions{
+		Level:  hclog.Info,
+		Output: os.Stderr,
+	})
+
+	// Wait for job to complete.
+	logger.Info(fmt.Sprintf("waiting for job %q to complete successfully", jobName))
+	for {
+		job, err := c.k8sClient.BatchV1().Jobs(c.flagNamespace).Get(jobName, metav1.GetOptions{})
+		if k8serrors.IsNotFound(err) {
+			logger.Info(fmt.Sprintf("job %q does not exist, no need to delete", jobName))
+			return 0
+		}
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error getting job %q: %s", jobName, err))
+			return 1
+		}
+
+		// If its succeeded we're done.
+		if job.Status.Succeeded > 0 {
+			break
+		}
+
+		// If its reached its backoff limit then it will never complete,
+		// exit and delete ourselves.
+		for _, condition := range job.Status.Conditions {
+			if condition.Type == v1.JobFailed && condition.Reason == "BackoffLimitExceeded" {
+				logger.Warn(fmt.Sprintf("job %q has reached its backoff limit and will never complete", jobName))
+				return 0
+			}
+		}
+
+		logger.Info(fmt.Sprintf("job %q has not yet succeeded, waiting %v", jobName, c.retryDuration))
+		time.Sleep(c.retryDuration)
+	}
+
+	// Here we know the job has succeeded. We can delete it and then delete
+	// ourselves.
+	logger.Info(fmt.Sprintf("job %q has succeeded, deleting", jobName))
+	propagationPolicy := metav1.DeletePropagationForeground
+	err := c.k8sClient.BatchV1().Jobs(c.flagNamespace).Delete(jobName, &metav1.DeleteOptions{
+		// Needed so that the underlying pods are also deleted.
+		PropagationPolicy: &propagationPolicy,
+	})
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("unable to delete job %q: %s", jobName, err))
+		return 1
+	}
+
+	logger.Info(fmt.Sprintf("Deleted job %q successfully", jobName))
+	return 0
+}
+
+func (c *Command) Synopsis() string { return synopsis }
+func (c *Command) Help() string {
+	c.once.Do(c.init)
+	return c.help
+}
+
+const synopsis = "Delete Kubernetes Job when complete."
+const help = `
+Usage: consul-k8s delete-completed-job [name] [options]
+
+  Waits for job to complete, then deletes it. If the job reaches its
+  backoff limit then the command will exit.
+`
--- a/subcommand/delete-completed-job/command.go
+++ b/subcommand/delete-completed-job/command.go
@@ -47,9 +47,9 @@ func (c *Command) init() {
 	flags.Merge(c.flags, c.k8s.Flags())
 	c.help = flags.Usage(help, c.flags)
 
-	// Default retry to 30s. This is exposed for setting in tests.
+	// Default retry to 1s. This is exposed for setting in tests.
 	if c.retryDuration == 0 {
-		c.retryDuration = 30 * time.Second
+		c.retryDuration = 1 * time.Second
 	}
 }
 
--- a/subcommand/delete-completed-job/command.go
+++ b/subcommand/delete-completed-job/command.go
@@ -25,7 +25,7 @@ type Command struct {
 	flags         *flag.FlagSet
 	k8s           *k8sflags.K8SFlags
 	flagNamespace string
-	flagTimeout string
+	flagTimeout   string
 
 	once      sync.Once
 	help      string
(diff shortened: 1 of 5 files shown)


lkysow:
Needs #144 merged first.

subcommand/delete-completed-job/command.go
```diff
@@ -0,0 +1,147 @@
...
+		for _, condition := range job.Status.Conditions {
+			if condition.Type == v1.JobFailed && condition.Reason == "BackoffLimitExceeded" {
+				logger.Warn(fmt.Sprintf("job %q has reached its backoff limit and will never complete", jobName))
+				return 0
+			}
+		}
+
+		logger.Info(fmt.Sprintf("job %q has not yet succeeded, waiting %v", jobName, c.retryDuration))
+		time.Sleep(c.retryDuration)
+	}
```
ishustava: I feel like there could be cases when the ACL init job will keep running forever and not exit. There are few times I ran into this. The first time was when the ACL init job was trying to connect to the consul server on the wrong IP. The second time was when the job didn't have the right permissions to write a secret. Although we can fix these issues (I believe this PR fixes the second one), it seems like there could be other edge cases when this could happen. 

In case the ACL init job runs forever, we should prevent the delete task from running forever as well.

subcommand/delete-completed-job/command_test.go
```diff
@@ -0,0 +1,170 @@
...
+			// Check job deletion.
+			_, err = k8s.BatchV1().Jobs(ns).Get(jobName, meta.GetOptions{})
+			if c.ExpDelete {
+				require.True(k8serrors.IsNotFound(err))
+			} else {
+				require.NoError(err)
+			}
+		})
+	}
+}
```
ishustava: Adding to my comment above, it'd be good to test the case when the job is currently active.

subcommand/server-acl-init/command.go
```diff
@@ -233,7 +233,7 @@ func (c *Command) getConsulServers(logger hclog.Logger) *apiv1.PodList {
 				return fmt.Errorf("no server pods with labels %q found", labelSelector)
 			}
 
-			if len(serverPods.Items) < c.flagReplicas {
+			if len(serverPods.Items) < n {
 				return fmt.Errorf("found %d servers, require %d", len(serverPods.Items), c.flagReplicas)
```
ishustava: Should this error message use `n` instead of `c.flagReplicas` as well?

subcommand/server-acl-init/command.go
```diff
@@ -244,13 +244,48 @@ func (c *Command) getConsulServers(logger hclog.Logger) *apiv1.PodList {
...
+	firstServerAddr := serverPods[0].Addr
+	consulClient, err := api.NewClient(&api.Config{
+		Address: firstServerAddr,
+		Scheme:  "http",
+	})
+	if err != nil {
+		return "", fmt.Errorf("creating Consul client for address %s: %s", firstServerAddr, err)
+	}
+
+	// Call bootstrap ACLs API.
```
ishustava: I think we need to move this into the `c.untilSucceeds` to avoid a race condition when the ACL init job discovers the server that is about to be terminated.

subcommand/delete-completed-job/command.go
lkysow: Should we have a timeout then, you think? Maybe 10 minutes?

subcommand/delete-completed-job/command_test.go
lkysow: 👍 will add

subcommand/server-acl-init/command.go
lkysow: 🍺

subcommand/server-acl-init/command.go
lkysow: Hmm, that case should now not be possible but in another way it could happen. Previously, if this job was run again (ex. during a helm upgrade), it would try to bootstrap servers again, even if there was the Kubernetes bootstrap secret.

Now, during a helm upgrade it would not try to boostrap again because it will detect the secret.

However, say they added the meshgateway=true value and ran an upgrade. Then [here](https://github.com/hashicorp/consul-k8s/blob/5a1b0928f789e50b99d6c9492e6e0c6bf7c9d8e4/subcommand/server-acl-init/command.go#L141) we grab the first consul server and use it to generate tokens. If that server was lost then we'd get into the same situation.

So maybe we need to loop back up and try a different server in that situation. I'll look into how best to do that in the code.

ishustava: Yep, you're right! I keep coming back to this. Although I think this could happen during initial installation too. For example, suppose your servers didn't come up successfully and you had to do something to restart them. It would be nice if the acl init job (which would be already running) could recover on its own. 

Your example with enabling a feature that requires an ACL change is a great example!

subcommand/delete-completed-job/command.go
lkysow: I've added a timeout that defaults to 10m to the delete command. I also added a timeout to the acl-init command. For the latter, I used the context package. This way if it gets into a weird state it will eventually exit. It added a lot of code though.

subcommand/server-acl-init/command.go
lkysow: I've added code that checks if there's an in-progress rollout and waits for it to finish before proceeding. I've also added an overall timeout so if we get into a pathological situation where we're looping forever we'll eventually exit and the job will retry.

subcommand/server-acl-init/command.go
```diff
@@ -117,95 +136,138 @@ func (c *Command) Run(args []string) int {
...
-	// Get Consul Server Pods and construct their addresses.
-	serverPods := c.getConsulServers(logger)
-	var serverPodAddrs []string
-	for _, pod := range serverPods.Items {
-		var httpPort int32
-		for _, p := range pod.Spec.Containers[0].Ports {
-			if p.Name == "http" {
-				httpPort = p.ContainerPort
-			}
+	// Wait if there's a rollout of servers.
```
lkysow: @ishustava I added rollout checks here.

subcommand/server-acl-init/command.go
```diff
@@ -522,42 +681,68 @@ func (c *Command) configureConnectInject(logger hclog.Logger, consulClient *api.
...
-		logger.Info("Retrying in 1s")
-		time.Sleep(1 * time.Second)
+		logger.Info("Retrying in " + c.retryDuration.String())
+		// Wait on either the retry duration (in which case we continue) or the
+		// overall command timeout.
+		select {
+		case <-time.After(c.retryDuration):
+			continue
+		case <-c.cmdTimeout.Done():
+			return errors.New("reached command timeout")
```
lkysow: @ishustava here I'm checking the overall command timeout

subcommand/server-acl-init/command_test.go
```diff
@@ -408,6 +400,21 @@ func TestRun_DelayedServerPods(t *testing.T) {
 			},
 		})
 		require.NoError(err)
+		_, err = k8s.AppsV1().StatefulSets(ns).Create(&appv1.StatefulSet{
```
lkysow: @ishustava I'd like to factor this out into some helper functions but I'll do that after a review so it's easier to read

subcommand/server-acl-init/command.go
lkysow: FYI if you're testing this use my latest PR of consul-helm because the acl-init job needs rbac for listing the statefulsets

lkysow: > It would be nice if the acl init job (which would be already running) could recover on its own.

I forgot to address this point. I agree that this would be nice but I think there's too much complexity to make it happen. We'd need to be able to re-lookup consul server pod ips and recreate a consul client given a certain error condition. Since the API calls happen so fast once we get a consul IP that works I think it's sufficient to wait for a rollout to be complete and then continue.

However I do agree that there are cases that we might get into and rather than add a layer of complexity for relooping back to create the consul client I added an overall timeout mechanism for the command which should then get retried because its a job.

Not saying that's the only way of doing things though.

subcommand/delete-completed-job/command.go
```diff
@@ -0,0 +1,165 @@
...
+			if condition.Type == v1.JobFailed && condition.Reason == "BackoffLimitExceeded" {
+				logger.Warn(fmt.Sprintf("job %q has reached its backoff limit and will never complete", jobName))
+				return 0
+			}
+		}
+
+		// If we've reached our timeout, exit.
+		if time.Now().After(start.Add(timeout)) {
+			logger.Warn(fmt.Sprintf("timeout %q has been reached, exiting without deleting job", timeout))
+			return 0
```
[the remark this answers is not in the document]
lkysow: Yeah, I was doing that because I wanted helm to delete the job regardless and I wasn't sure if you could set `    "helm.sh/hook-delete-policy": hook-succeeded` and `    "helm.sh/hook-delete-policy": hook-failed`

Let me test that first though. And regardless, I should comment explaining why.

subcommand/server-acl-init/command.go
```diff
@@ -117,95 +139,138 @@ func (c *Command) Run(args []string) int {
...
+		if ss.Status.CurrentRevision == ss.Status.UpdateRevision {
+			return nil
 		}
-		addr := fmt.Sprintf("%s:%d", pod.Status.PodIP, httpPort)
-		serverPodAddrs = append(serverPodAddrs, addr)
+		return fmt.Errorf("rollout is in progress (CurrentRevision=%s UpdateRevision=%s)",
+			ss.Status.CurrentRevision, ss.Status.UpdateRevision)
+	}, logger)
+	if err != nil {
+		c.UI.Error(err.Error())
```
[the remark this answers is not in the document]
lkysow: I was following the pattern of using `c.UI` before exiting but I agree it makes more sense to just use log and then `return 1`

subcommand/delete-completed-job/command.go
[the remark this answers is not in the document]
lkysow: Okay, you can do that :D thanks for keeping me honest

subcommand/delete-completed-job/command.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	logger := hclog.New(&hclog.LoggerOptions{
+		Level:  hclog.Info,
+		Output: os.Stderr,
+	})
+
+	// Wait for job to complete.
+	logger.Info(fmt.Sprintf("waiting for job %q to complete successfully", jobName))
+	for {
```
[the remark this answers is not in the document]
lkysow: Hmm, I think I'd like to wait until we have 3 different uses before generalizing it. The use here is slightly different because the `op` that we'd pass in wouldn't need to complete without error unlike in server-acl-init. So we'd need to parameterize it in some way and I'm not sure it's worth that effort yet.

subcommand/delete-completed-job/command.go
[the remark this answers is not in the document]
lkysow: done

subcommand/server-acl-init/command.go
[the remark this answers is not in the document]
lkysow: done

subcommand/delete-completed-job/command.go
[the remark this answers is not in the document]
ishustava: Ok, that makes sense.

subcommand/delete-completed-job/command.go
ishustava: @lkysow I think we need to change default timeouts for both commands to be less than 5min (helm's default timeout). 

I had a deployment where servers never became healthy. The ACL init job has timed out eventually and finished with error, as expected. But the cleanup container got deleted by helm before it had a chance to delete the ACL init job. This happened because `helm upgrade` or `helm install` will wait until the hook job will finish or until a timeout is reached. The default helm timeout of 5min is smaller than the default timeouts for both commands, so helm considered the hook "failed" and deleted the job.

subcommand/delete-completed-job/command.go
```diff
@@ -0,0 +1,170 @@
...
+
+		// If its reached its backoff limit then it will never complete.
+		for _, condition := range job.Status.Conditions {
+			if condition.Type == v1.JobFailed && condition.Reason == "BackoffLimitExceeded" {
+				logger.Warn(fmt.Sprintf("job %q has reached its backoff limit and will never complete", jobName))
+				return 1
+			}
+		}
+
+		logger.Info(fmt.Sprintf("job %q has not yet succeeded, waiting %v", jobName, c.retryDuration))
```
ishustava: Might be worth changing this to `has not yet finished`, since it could also fail with a timeout now.

subcommand/delete-completed-job/command.go
ishustava: Alternatively, we could change `hook-delete-policy` to `before-hook-creation`, so that helm will delete it only on the next install or upgrade.

ishustava: Another issue is that in the worst case it'll take the ACL init job to run 60min by default (the default backoff limit of 6 x job's default timeout of 10min). The default timeout for the ACL cleanup job should be more than that.

lkysow:
So the timeout of the delete job doesn't really matter right? Because
either way if it hits its own or helm's timeout it's still gonna get deleted

But for the acl init job it's kinda nice it tries past the helm timeout no?

subcommand/server-acl-init/command.go
```diff
@@ -244,14 +311,52 @@ func (c *Command) getConsulServers(logger hclog.Logger) *apiv1.PodList {
...
-	return serverPods
+	if err != nil {
+		return nil, err
+	}
+
+	var podAddrs []podAddr
+	for _, pod := range serverPods.Items {
+		var httpPort int32
+		for _, p := range pod.Spec.Containers[0].Ports {
+			if p.Name == "http" {
```
filintod: why no option for `https` port here for TLS only environments?

subcommand/server-acl-init/command.go
```diff
@@ -522,42 +684,68 @@ func (c *Command) configureConnectInject(logger hclog.Logger, consulClient *api.
...
 
-const synopsis = "Initialize ACLs on Consul servers."
+// podAddr is a convenience struct for passing around pod names and
+// addresses for Consul servers.
+type podAddr struct {
+	// Name is the name of the pod.
+	Name string
+	// Addr is in the form "<ip>:<port>".
+	Addr string
+}
```
filintod: Add `Scheme` attribute to allow for https scheme consul client to be assigned on the getConsulServers above.

subcommand/server-acl-init/command.go
```diff
@@ -117,95 +131,146 @@ func (c *Command) Run(args []string) int {
...
+
+	// For all of the next operations we'll need a Consul client.
+	serverPods, err := c.getConsulServers(logger, 1)
+	if err != nil {
+		logger.Error(err.Error())
+		return 1
+	}
+	serverAddr := serverPods[0].Addr
+	consulClient, err := api.NewClient(&api.Config{
+		Address: serverAddr,
```
filintod: I think if you allow my comment above of also checking for TLS `https` port and add attribute `Scheme` to podAddr struct, we could use:

```Scheme: serverPods[0].Scheme```

So this can be use for both `http` and `https` environments

File: subcommand/server-acl-init/command.go [hunk 15]
```diff
@@ -244,14 +311,52 @@ func (c *Command) getConsulServers(logger hclog.Logger) *apiv1.PodList {
...
+	if err != nil {
+		return "", err
+	}
+	logger.Info(fmt.Sprintf("Found %d Consul server Pods", len(serverPods)))
+
+	// Pick the first pod to connect to for bootstrapping and set up connection.
+	firstServerAddr := serverPods[0].Addr
+	consulClient, err := api.NewClient(&api.Config{
+		Address: firstServerAddr,
+		Scheme:  "http",
```
filintod: Same comment about TLS with added `Scheme` attribute to podAddr

subcommand/server-acl-init/command.go
```diff
@@ -244,14 +311,52 @@ func (c *Command) getConsulServers(logger hclog.Logger) *apiv1.PodList {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var podAddrs []podAddr
+	for _, pod := range serverPods.Items {
+		var httpPort int32
+		for _, p := range pod.Spec.Containers[0].Ports {
+			if p.Name == "http" {
+				httpPort = p.ContainerPort
```
filintod: why no option for https port here in addition to http in case for TLS only environments?

subcommand/server-acl-init/command.go
```diff
@@ -273,59 +379,60 @@ func (c *Command) bootstrapACLs(logger hclog.Logger, consulClient *api.Client) (
...
+	if err != nil {
+		return "", err
 	}
-	return bootstrapToken, nil
+
+	// Override our original client with a new one that has the bootstrap token
+	// set.
+	consulClient, err = api.NewClient(&api.Config{
+		Address: firstServerAddr,
+		Scheme:  "http",
```
filintod: same related to TLS and new podAddr `Scheme` attribute

subcommand/server-acl-init/command.go
```diff
@@ -522,42 +684,68 @@ func (c *Command) configureConnectInject(logger hclog.Logger, consulClient *api.
...
 }
 
-const synopsis = "Initialize ACLs on Consul servers."
+// podAddr is a convenience struct for passing around pod names and
+// addresses for Consul servers.
+type podAddr struct {
+	// Name is the name of the pod.
+	Name string
+	// Addr is in the form "<ip>:<port>".
+	Addr string
```
filintod: Why not add `Scheme` attribute to allow for TLS environments

subcommand/server-acl-init/command.go
```diff
@@ -334,58 +441,75 @@ func (c *Command) setServerTokens(logger hclog.Logger, consulClient *api.Client,
...
 	}
 
 	// Pass out agent tokens to servers.
-	for i := 0; i < len(serverPodAddrs); i++ {
+	for i, pod := range serverPods {
 		// We create a new client for each server because we need to call each
 		// server specifically.
 		serverClient, err := api.NewClient(&api.Config{
-			Address: serverPodAddrs[i],
+			Address: pod.Addr,
```
filintod: same use pod.Scheme instead of hardcoded `http`

subcommand/server-acl-init/command.go
```diff
@@ -244,14 +311,52 @@ func (c *Command) getConsulServers(logger hclog.Logger) *apiv1.PodList {
...
+				httpPort = p.ContainerPort
+			}
+		}
+		if httpPort == 0 {
+			return nil, fmt.Errorf("pod %s has no port labeled 'http'", pod.Name)
+		}
+		addr := fmt.Sprintf("%s:%d", pod.Status.PodIP, httpPort)
+		podAddrs = append(podAddrs, podAddr{
+			Name: pod.Name,
+			Addr: addr,
```
filintod: add `Scheme` attribute if `https` port is available

subcommand/server-acl-init/command.go
ishustava: Hey @filintod, thanks for pointing this out. Currently, consul servers deployed by the helm chart [do not support](https://github.com/hashicorp/consul-helm/blob/24d3115a2f04e8b5058f43b1d67afe6eff20df19/templates/server-statefulset.yaml#L142-L156) https, that's why the code here doesn't have that option. This is definitely something we're looking to add soon.

filintod: @ishustava I know about the helm chart as I had to go and customize it and this command.go file to support it and looking at how substantial this change is I though it would be good to also add it.  Without this change the helm chart with TLS will not work, there is a PR https://github.com/hashicorp/consul-helm/pull/137 that really does not work without changes in parallel with this consul-k8s tool.  But you can resolve it, maybe adding a bug to add TLS support to keep track and work to be released with helm tls support.

ishustava: Yes, definitely. We didn't have an issue for this, but it was certainly something we were aware of. I've created #154 to track this separately.

lkysow: Thanks for the review Filinto 🙏. Getting TLS working is an important goal for us but I think we'll tackle in a separate PR as our priority is getting these fixes out.

When we start on TLS work we'll tag you.

subcommand/delete-completed-job/command.go
```diff
@@ -0,0 +1,170 @@
...
+	c.flags.StringVar(&c.flagNamespace, "k8s-namespace", "",
+		"Name of Kubernetes namespace where the job is deployed")
+	c.flags.StringVar(&c.flagTimeout, "timeout", "30m",
+		"How long we'll wait for the job to complete before timing out, e.g. 1ms, 2s, 3m")
+	flags.Merge(c.flags, c.k8s.Flags())
+	c.help = flags.Usage(help, c.flags)
+
+	// Default retry to 30s. This is exposed for setting in tests.
+	if c.retryDuration == 0 {
+		c.retryDuration = 30 * time.Second
```
lkysow: I realized this should be 1s because helm upgrade (and install I think) will wait until the job exits before the helm command exits. So if it waits 30s when the acl-init completes in 2s then it's just sitting unnecessarily.

PR #153, opened 2019-10-11
Fix: changing sync annotation=false should delete

### LINKED ISSUE #76: Disabling or removing 'consul.hashicorp.com/service-sync' annotation does nothing
### What i'm trying to do
* Consul-k8s is running with `-k8s-default-sync=false`
* Starting with a synced service :
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    consul.hashicorp.com/service-meta-OWNER: foobar
    consul.hashicorp.com/service-name: my-kubernetes-hosted-svc
    consul.hashicorp.com/service-sync: "true"
    consul.hashicorp.com/service-tags: foo,bar
[...]
```

### What is the problem
When we set `consul.hashicorp.com/service-sync` to `"false"` or if we remove it (using kubectl edit) :
* Consul-k8s does nothing
* Service is still synchronized

### What it should be doing (IMO)
* Removing all registered instances for this service when we remove (or disable) this annotation

The only way I find to do so is to delete this service and create it again without this annotation.

Did I missed something or is it the intended behavior ?
Thx a lot :) !

lkysow:
Previously, if consul.hashicorp.com/service-sync was set to true and
then set to false, the service wouldn't get deleted.

Fixes #76

### DIFF
--- a/catalog/from-k8s/resource.go
+++ b/catalog/from-k8s/resource.go
@@ -71,15 +71,21 @@ type ServiceResource struct {
 	// ip address will be used instead.
 	NodePortSync NodePortSyncType
 
-	// serviceMap is a mapping of unique key (given by controller) to
-	// the service structure. endpointsMap is the mapping of the same
-	// uniqueKey to a set of endpoints.
-	//
 	// serviceLock must be held for any read/write to these maps.
-	serviceLock  sync.RWMutex
-	serviceMap   map[string]*apiv1.Service
+	serviceLock sync.RWMutex
+
+	// serviceMap holds services we should sync to Consul. Keys are the
+	// in the form <kube namespace>/<kube svc name>.
+	serviceMap map[string]*apiv1.Service
+
+	// endpointsMap uses the same keys as serviceMap but maps to the endpoints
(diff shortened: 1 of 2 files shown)


catalog/from-k8s/resource.go
```diff
@@ -151,18 +164,23 @@ func (t *ServiceResource) Upsert(key string, raw interface{}) error {
 func (t *ServiceResource) Delete(key string) error {
 	t.serviceLock.Lock()
 	defer t.serviceLock.Unlock()
+	t.doDelete(key)
+	t.Log.Info("delete", "key", key)
+	return nil
+}
```
ishustava: This function never returns an actual error.

lkysow: Yes but it's implementing the `controller.Resource` interface:
https://github.com/hashicorp/consul-k8s/blob/81026eb1a4cb2c7c3c75b03ff36eb8328a0300e1/helper/controller/resource.go#L12-L27

ishustava: got it!

PR #169, opened 2019-11-30
Only write service-defaults if protocol set

### LINKED ISSUE #168: Don't write service-defaults during connect-inject unless protocol set explicitly
## Background
The init container we inject for Consul connect writes a [service-defaults](https://www.consul.io/docs/agent/config-entries/service-defaults.html) config for the service if central config is enabled (the default in the helm chart):
https://github.com/hashicorp/consul-k8s/blob/8a2c7cb13e1de058bea79cb1cf2b34749a7cb6aa/connect-inject/container_init.go#L263-L270

The protocol can be set passing the `-default-protocol` flag which is controlled by the Helm chart's `connectInject.centralConfig.defaultProtocol` value. This protocol can be overridden by an annotation on the pod (`consul.hashicorp.com/connect-service-protocol`). If there is no annotation or Helm value, it will be set to an empty string which Consul will treat as TCP.

## Problem
**By default we write TCP** - When the Pod starts, the `service-defaults` config will be written. By default, this will set the protocol to TCP. If the protocol is TCP, none of the L7 routing rules will work. A user isn't expecting this config to be written because they haven't configured it anywhere. If they had written a `global` `proxy-defaults` config and set the protocol to `http` so their L7 rules work, the `service-defaults` config will override the protocol to tcp. This is unexpected. If a user hasn't explicitly configured a protocol, we should not write a service-defaults config.

## Proposal
Only write `service-defaults` if the protocol is explicitly configured, i.e. in `connectInject.centralConfig.defaultProtocol` or the annotation.

lkysow:
Previously, we would write a service-defaults config regardless of the
value of the -default-protocol flag or the Pod's protocol annotation. If
neither of these were set, we would write a config with protocol set to
the empty string. This is the same as setting it to tcp. So for every
service, we were by default setting its protocol to tcp.

If a user explicitly created a global proxy-defaults config and set the
protocol to, say, http. Then our service-defaults config would override
that protocol. This behaviour was unexpected because users had never
explicitly told us to write a service-defaults config and it wasn't
helping them.

This change will cause us to only write a service-defaults config if the
protocol is explicitly set–either via the -default-protocol flag or via
the Pod annotation.

I've also renamed the CentralConfig field to WriteServiceDefaults
because there are many types of central config now so it's best to be
explicit.

Fixes #168

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -13,14 +13,15 @@ type initContainerCommandData struct {
 	ServiceName string
 	ServicePort int32
 	// ServiceProtocol is the protocol for the service-defaults config
-	// that will be written if CentralConfig is true. If empty, Consul
-	// will default to "tcp".
+	// that will be written if WriteServiceDefaults is true.
 	ServiceProtocol string
 	AuthMethod      string
-	CentralConfig   bool
-	Upstreams       []initContainerCommandUpstreamData
-	Tags            string
-	Meta            map[string]string
+	// WriteServiceDefaults controls whether a service-defaults config is
+	// written for this service.
+	WriteServiceDefaults bool
+	Upstreams            []initContainerCommandUpstreamData
+	Tags                 string
+	Meta                 map[string]string
 }
 
 type initContainerCommandUpstreamData struct {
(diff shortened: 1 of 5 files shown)


connect-inject/container_init.go
```diff
@@ -37,11 +38,17 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
 	if annoProtocol, ok := pod.Annotations[annotationProtocol]; ok {
 		protocol = annoProtocol
 	}
+	// We only write a service-defaults config if central config is enabled
+	// and a protocol is specified. Previously, we would write a config when
+	// the protocol was empty. This is the same as setting it to tcp. This
+	// would then override any global proxy-defaults config. Now, we only
+	// write the config if a protocol is explicitly set.
+	writeServiceDefaults := h.WriteServiceDefaults && protocol != ""
```
lkysow: Here's the only real change. Previously this was just the value of `h.WriteServiceDefaults`

connect-inject/handler.go
```diff
@@ -364,9 +364,9 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
 		}
 	}
 
-	if h.CentralConfig {
+	if h.WriteServiceDefaults {
 		// Default protocol is specified by a flag if not explicitly annotated
-		if _, ok := pod.ObjectMeta.Annotations[annotationProtocol]; !ok {
+		if _, ok := pod.ObjectMeta.Annotations[annotationProtocol]; !ok && h.DefaultProtocol != "" {
```
lkysow: Previously we would write an annotation even if the protocol was empty (i.e. we'd set it to the empty string).

Issue #163, opened 2019-11-08
Connect proxy uses wrong service id


lkysow:
The service ID is set here:
https://github.com/hashicorp/consul-k8s/blob/8a2c7cb13e1de058bea79cb1cf2b34749a7cb6aa/connect-inject/container_init.go#L245-L247

to `"${POD_NAME}-{{ .ServiceName }}"`. However the proxy config doesn't use the correct `destination_service_id` and just uses `{{.ServiceName}}`:

https://github.com/hashicorp/consul-k8s/blob/8a2c7cb13e1de058bea79cb1cf2b34749a7cb6aa/connect-inject/container_init.go#L207-L209

This doesn't seem to break Connect, however it breaks the links in the UI where the Dest. service instance results in a 404 since it links to `static-client` instead of `static-client-static-client`:
![image](https://user-images.githubusercontent.com/1034429/68438695-ccfc9f80-0179-11ea-879a-faef2c050922.png)

PR #172, opened 2019-12-07
fix: proxy's destination service id is not service id

lkysow:
See https://github.com/hashicorp/consul-k8s/issues/163 for bug description.
The fix is to make the service id and proxy service id environment variables (since they're dynamic based on the pod name) so that they can be interpolated in the bash script without copy paste errors.

* Carries https://github.com/hashicorp/consul-k8s/pull/164. Just needed a rebase and a test update.
* Fixes https://github.com/hashicorp/consul-k8s/issues/163

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -3,6 +3,7 @@ package connectinject
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


connect-inject/container_init.go
```diff
@@ -179,6 +182,14 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
 					FieldRef: &corev1.ObjectFieldSelector{FieldPath: "metadata.namespace"},
 				},
 			},
+			{
+				Name:  "SERVICE_ID",
+				Value: fmt.Sprintf("$(POD_NAME)-%s", data.ServiceName),
```
lkysow: Kube interpolates these

PR #173, opened 2019-12-09
Add consul-k8s agent hostname meta to default AgentService

ashangit:
We had an issue with consul services ,managed by consul-k8s sync-catalog agent, flapping.
This was due to 2 agents running on 2 different kubernetes clusters with same configuration (same 
<consul-k8s-tag>) exposing same services.
As no information from the agent exposing those service is available in the service it took us some times to identify the issue.
This PR is to add the hostname of the agent to the service definition so we can fastly identify which agents manage it

### DIFF
--- a/catalog/to-consul/resource.go
+++ b/catalog/to-consul/resource.go
@@ -26,6 +26,10 @@ const (
 	// ConsulK8SNS is the key used in the meta to record the namespace
 	// of the service/node registration.
 	ConsulK8SNS = "external-k8s-ns"
+
+	// SyncCatalogPodNameMeta is the key used in the meta to track hostname
+	// of the consul-k8s agent syncing.
+	SyncCatalogPodNameMeta = "sync-catalog-pod-name"
 )
 
 type NodePortSyncType string
@@ -92,6 +96,9 @@ type ServiceResource struct {
 	// It's populated via Consul's API and lets us diff what is actually in
 	// Consul vs. what we expect to be there.
 	consulMap map[string][]*consulapi.CatalogRegistration
+
+	// SyncCatalogPodName is the name of the pod running the sync catalog.
+	SyncCatalogPodName string
 }
 
 // Informer implements the controller.Resource interface.
@@ -290,8 +297,9 @@ func (t *ServiceResource) generateRegistrations(key string) {
 		Service: t.addPrefixAndK8SNamespace(svc.Name, svc.Namespace),
 		Tags:    []string{t.ConsulK8STag},
 		Meta: map[string]string{
-			ConsulSourceKey: ConsulSourceValue,
-			ConsulK8SNS:     t.namespace(),
+			ConsulSourceKey:        ConsulSourceValue,
+			ConsulK8SNS:            t.namespace(),
+			SyncCatalogPodNameMeta: t.SyncCatalogPodName,
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


lkysow:
Thanks for the PR. One problem with this is that the hostname is just the pod id which will change if the sync-catalog pod is restarted. Then you'd have a hostname that doesn't correspond with any running pod in your cluster which would be confusing.

What about setting the
```yaml
syncCatalog:
  k8sTag: my-tag
```

value in your Helm chart? If this was set separately for each syncer then your services in Consul will have separate tags and you can tell them apart.

ashangit:
@lkysow we are already setting the k8sTagSuffix which also set consul-k8s-tag and in fact it works very well.
The issue there was due to a bad configuration leading to 2 sync agents from 2 different clusters running with same consul-k8s-tag config.
The idea behind adding the hostname tag is not to solve that issue but to have more insights when trying to identify such a bad config
From my understanding of the code the service will be updated when the sync-catalog will be restarted so you should not keep a 'bad' hostname tag.

lkysow:
> From my understanding of the code the service will be updated when the sync-catalog will be restarted so you should not keep a 'bad' hostname tag.

Ahh, good point!

lkysow:
Hi Nicolas,
Would this [PR](https://github.com/hashicorp/consul-k8s/pull/103) that lets you name the `k8s-sync` node something else also work for you? Then in each cluster you could name the node something different (like the cluster id) and then in the UI you could easily trace which service is registered by which sync process.

catalog/to-consul/resource.go
```diff
@@ -26,6 +27,10 @@ const (
 	// ConsulK8SNS is the key used in the meta to record the namespace
 	// of the service/node registration.
 	ConsulK8SNS = "external-k8s-ns"
+
+	// ConsulK8SHostNameKey is the key used in the meta to track hostname
+	// of the consul-k8s agent syncing.
+	ConsulK8SHostNameKey = "agent-k8s-hostname"
```
lkysow:
```suggestion
	SyncCatalogPodNameMeta = "sync-catalog-pod-name"
```

I think better to be explicit about what the value it

catalog/to-consul/resource.go
```diff
@@ -289,6 +294,11 @@ func (t *ServiceResource) generateRegistrations(key string) {
 		},
 	}
 
+	hostname, err := os.Hostname()
```
lkysow: rather than doing this every time here, lets make this a field on the `ServiceResource` struct:
```
// SyncCatalogPodName is the name of the pod running the sync catalog.
SyncCatalogPodName string
```

You can add the field on line 183 of the `sync-catalog/command.go` file.

catalog/to-consul/resource.go
```diff
@@ -289,6 +294,11 @@ func (t *ServiceResource) generateRegistrations(key string) {
...
+	if err == nil {
```
lkysow: If there was an error, we'll need to log it and then we can set the `hostname` to `<unknown>`.

ashangit:
@lkysow this in fact an interesting PR but not exactly what I try to achieve there.
Idea is more to have a tag auto-generated which is not based on some manual configuration so even if an agent is badly configured this tag will still be different from multiple agent

david-yu:
Closing this PR as its been a long time since this was last reviewed. Please re-open a new PR if you are still looking for such an enhancement.

david-yu:
Closing this PR as its been a long time since this was last reviewed. Please re-open a new PR if you are still looking for such an enhancement.

Issue #165, opened 2019-11-16
Incorrect assumption of StatefulSet name in server-acl-init


sbuzonas:
Wasn't sure if this was better reported in hashicorp/consul-helm or here

The `server-acl-init` subcommand assumes that the name of the StatefulSet will be `{release_name}-consul-server`

https://github.com/hashicorp/consul-k8s/blob/master/subcommand/server-acl-init/command.go#L143

The helm chart allows a user to specify a fullname override.
https://github.com/hashicorp/consul-helm/blob/v0.12.0/templates/_helpers.tpl#L8

Given the following configuration values.yaml provided to helm:
```
fullnameOverride: consul
global:
  bootstrapACLs: true
```

The StatefulSet will be named `consul-server` and the bootstrap job will fail

sbuzonas:
Created hashicorp/consul-helm#286 as a fix likely require updates to both

lkysow:
This is fixed in consul-helm v0.15.0. Thanks!

PR #174, opened 2019-12-09
Support fullnameOverride

lkysow:
In the Helm chart there is an undocumented value that allows users to
override the prefix we usually apply to the resources. The default
prefix is <helm release name>-consul. This change adds two flags to the
server-acl-init command (the only command that is affected by the
prefix): `-resource-prefix` and `-server-label-selector`.

If set, we will use the correct prefix,
both for finding resources that we expect to be created by the helm
chart, e.g. Consul servers, and for creating our own resources, e.g.
Kubernetes secrets.

The flags are backwards compatible. Users can upgrade consul-k8s and not require any changes to their consul-helm version. There is a corresponding consul-helm update that uses these flags. That is only required if users want to take advantage of this bugfix (which now fully supports the name override values).

* Fixes https://github.com/hashicorp/consul-k8s/issues/165
* Corresponding consul-helm PR: https://github.com/hashicorp/consul-helm/pull/307

### DIFF
--- a/subcommand/server-acl-init/command.go
+++ b/subcommand/server-acl-init/command.go
@@ -56,7 +56,7 @@ func (c *Command) init() {
 	c.flags.StringVar(&c.flagReleaseName, "release-name", "",
 		"Name of Consul Helm release")
 	c.flags.StringVar(&c.flagResourcePrefix, "resource-prefix", "",
-		"Prefix to use for Kubernetes resources. If not set, will default to ${release-name}-consul.")
+		"Prefix to use for Kubernetes resources. If not set, the \"<release-name>-consul\" prefix is used, where <release-name> is the value set by the -release-name flag.")
 	c.flags.IntVar(&c.flagReplicas, "expected-replicas", 1,
 		"Number of expected Consul server replicas")
 	c.flags.StringVar(&c.flagNamespace, "k8s-namespace", "",
--- a/subcommand/server-acl-init/command.go
+++ b/subcommand/server-acl-init/command.go
@@ -28,6 +28,7 @@ type Command struct {
 	flags                        *flag.FlagSet
 	k8s                          *k8sflags.K8SFlags
 	flagReleaseName              string
+	flagResourcePrefix           string
 	flagReplicas                 int
 	flagNamespace                string
 	flagAllowDNS                 bool
@@ -54,6 +55,8 @@ func (c *Command) init() {
 	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
 	c.flags.StringVar(&c.flagReleaseName, "release-name", "",
 		"Name of Consul Helm release")
+	c.flags.StringVar(&c.flagResourcePrefix, "resource-prefix", "",
+		"Prefix to use for Kubernetes resources. If not set, will default to ${release-name}-consul.")
 	c.flags.IntVar(&c.flagReplicas, "expected-replicas", 1,
 		"Number of expected Consul server replicas")
 	c.flags.StringVar(&c.flagNamespace, "k8s-namespace", "",
@@ -115,6 +118,10 @@ func (c *Command) Run(args []string) int {
 		c.UI.Error(fmt.Sprintf("%q is not a valid timeout: %s", c.flagTimeout, err))
 		return 1
 	}
+	if c.flagReleaseName == "" {
+		c.UI.Error("-release-name must be set")
+		return 1
+	}
 	var cancel context.CancelFunc
 	c.cmdTimeout, cancel = context.WithTimeout(context.Background(), timeout)
 	// The context will only ever be intentionally ended by the timeout.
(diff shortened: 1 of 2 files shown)


subcommand/server-acl-init/command_test.go
```diff
@@ -72,7 +72,8 @@ func TestRun_Defaults(t *testing.T) {
 }
 
 // Test the different flags that should create tokens and save them as
-// Kubernetes secrets.
+// Kubernetes secrets. We also test using with and without using the
+// -name-override to ensure the secrets are created with the right prefix.
 func TestRun_Tokens(t *testing.T) {
```
lkysow: I added another case in this test for running with/without the -name-override flag.

subcommand/server-acl-init/command.go
```diff
@@ -115,6 +118,10 @@ func (c *Command) Run(args []string) int {
 		c.UI.Error(fmt.Sprintf("%q is not a valid timeout: %s", c.flagTimeout, err))
 		return 1
 	}
+	if c.flagReleaseName == "" {
+		c.UI.Error("-release-name must be set")
```
lkysow: If this isn't set, we won't find the Consul servers since we do a lookup by label so best to give an error.

subcommand/server-acl-init/command_test.go
```diff
@@ -22,10 +22,22 @@ import (
 
 var ns = "default"
 var releaseName = "release-name"
+var resourcePrefix = "release-name-consul"
```
lkysow: I've updated the tests to use both the `-release-name` and `-resource-prefix` flags since the corresponding change to the helm chart will use both flags and I thought it best to match the Helm chart usage.

I've also updated one test to not use the `-resource-prefix` flag, to test that behaviour (should default to release-name-consul). That's important because if someone upgrades consul-k8s and not the Helm chart, they won't be setting that -resource-prefix flag.

subcommand/server-acl-init/command_test.go
```diff
@@ -72,38 +85,73 @@ func TestRun_Defaults(t *testing.T) {
...
 func TestRun_Tokens(t *testing.T) {
 	t.Parallel()
 
 	cases := map[string]struct {
-		Flag      string
-		TokenName string
+		TokenFlag          string
+		ResourcePrefixFlag string
+		TokenName          string
+		SecretName         string
```
lkysow: I expanded these cases to include whether we set -resource-prefix or not

subcommand/server-acl-init/command.go
```diff
@@ -28,6 +28,7 @@ type Command struct {
 	flags                        *flag.FlagSet
 	k8s                          *k8sflags.K8SFlags
 	flagReleaseName              string
```
ishustava: If we always set the resource prefix flag through the helm chart, then the only thing this flag is used for is selecting the server pods. Would it make sense to always require -resource-prefix and change this flag to take a label selector? 

The reason I'm suggesting it is that I think we are making a few hidden assumptions in this command about how resources are named in the helm chart, and it seems like a leaky abstraction.

subcommand/server-acl-init/command.go
```diff
@@ -140,7 +147,7 @@ func (c *Command) Run(args []string) int {
 	}
 
 	// Wait if there's a rollout of servers.
-	ssName := c.flagReleaseName + "-consul-server"
+	ssName := c.withPrefix("server")
```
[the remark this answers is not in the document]
lkysow: yes, great idea

subcommand/server-acl-init/command.go
```diff
@@ -54,6 +55,8 @@ func (c *Command) init() {
 	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
 	c.flags.StringVar(&c.flagReleaseName, "release-name", "",
 		"Name of Consul Helm release")
+	c.flags.StringVar(&c.flagResourcePrefix, "resource-prefix", "",
+		"Prefix to use for Kubernetes resources. If not set, will default to ${release-name}-consul.")
```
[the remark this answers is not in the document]
lkysow: 💯

subcommand/server-acl-init/command.go
lkysow: Ooh, yeah I like that.

lkysow:
I'm gonna have to re-jig some of the tests to take account for the label-selector.

lkysow:
@ishustava I've updated the PR by adding the new `-server-label-selector` flag as suggested. I like how it removes that implicit reliance on the server name and its labels.

As a result, I've had to stop putting the release name in the description of the different tokens. This doesn't cause any issues with duplication because the token names themselves in Consul require uniqueness not the descriptions and the names are the same as they were previously.

subcommand/server-acl-init/command.go
```diff
@@ -140,25 +163,29 @@ func (c *Command) Run(args []string) int {
 	}
 
 	// Wait if there's a rollout of servers.
-	ssName := c.flagReleaseName + "-consul-server"
+	ssName := c.withPrefix("server")
 	err = c.untilSucceeds(fmt.Sprintf("waiting for rollout of statefulset %s", ssName), func() error {
-		ss, err := c.clientset.AppsV1().StatefulSets(c.flagNamespace).Get(ssName, metav1.GetOptions{})
+		statefulsets, err := c.clientset.AppsV1().StatefulSets(c.flagNamespace).List(metav1.ListOptions{LabelSelector: c.flagServerLabelSelector})
```
ishustava: @lkysow I realized that this will require changes to the helm chart. Currently the labels for the server pods and the server stateful set are different, so we can't use the same label selector.

lkysow: Good catch, I've moved back to using the name of the statefulset.

PR #179, opened 2019-12-13
Add support for https server discovery via -scheme flag

behaviour:
Added flag to server-acl-init to support https by specifying '--scheme=https' (defaults to http as per current scheme)


lkysow:
Hi @behaviour I think this functionality was added by https://github.com/hashicorp/consul-k8s/pull/183 so this can be closed.

lkysow:
Closed by #183.

PR #180, opened 2019-12-17
Support namespaces in the Connect injector

adilyse:
This adds namespace config options for registering injected
services into a single namespace as well as mirroring k8s
namespaces in Consul with an optional prefix.

It adds functionality to check for Consul namespace existence
and create new namespaces. If mirroring namespaces with ACLs
enabled, binding rules are also checked and created if necessary.

Service and proxy registration as well as service-defaults
have been updated to be namespace aware.

Note: this currently doesn't work because the lifecycle sidecar is using
a non-namespace aware function to register services.

Open issues:
- Meta on namespaces is not yet supported in OSS api
- `consul config write` does not take a `-namespace` flag yet
- `consul logout` does not take a `-namespace` flag yet
- `service-defaults` may or may not need a namespace field
- Consul api import will need to be updated to a released version
- Add tests for `command`


connect-inject/container_init.go
```diff
@@ -286,29 +295,40 @@ name = "{{ .ServiceName }}"
...
+  {{- end }}
   /consul/connect-inject/service-defaults.hcl || true
 {{- end }}
 
 /bin/consul services register \
   {{- if .AuthMethod }}
   -token-file="/consul/connect-inject/acl-token" \
   {{- end }}
+  {{- if .ConsulNamespace }}
+  -namespace={{ .ConsulNamespace }} \
```
lkysow: it's a different flag in this case? namespace/ns?

connect-inject/envoy_sidecar.go
```diff
@@ -8,13 +8,23 @@ import (
...
 )
 
+type sidecarContainerCommandData struct {
+	AuthMethod      string
+	ConsulNamespace string
+}
+
 func (h *Handler) envoySidecar(pod *corev1.Pod) (corev1.Container, error) {
+	// Inputs to the sidecar template rendering
+	data := sidecarContainerCommandData{
```
lkysow: maybe call this `templateData` and then you don't need the comment

connect-inject/handler.go
```diff
@@ -119,6 +119,45 @@ type Handler struct {
...
+	// MirroringPrefix is an optional prefix that can be added to the Consul
+	// namespaces created while mirroring. For example, if it is set to "k8s-",
+	// then the k8s `default` namespace will be mirrored in Consul's
+	// `k8s-default` namespace.
+	MirroringPrefix string
+
+	// BindingRuleSelector is only used when mirroring. Because binding rules
+	// are namespaced, a new one is required to be created in each Consul
+	// namespace that is created when mirroring.
+	BindingRuleSelector string
```
lkysow: 💯 comments

connect-inject/handler.go
```diff
@@ -119,6 +119,45 @@ type Handler struct {
...
+	// DenyK8sNamespacesList is a list of k8s namespaces to explicitly deny
+	// injection and thus service registration with Consul. An empty list
+	// means that no namespaces are removed from consideration. This filter
+	// takes precedence over AllowK8sNamespacesList.
+	DenyK8sNamespacesSet mapset.Set
+
+	// ConsulNSName is the name of the Consul namespace to register all
+	// injected services into if Consul namespaces are enabled and mirroring
+	// is disabled. This will not be used if mirroring is enabled.
+	ConsulNSName string
```
lkysow: I'd use `Namespace` here instead of NS since you're using namespace elsewhere in these fields

connect-inject/handler.go
```diff
@@ -181,10 +220,10 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
 	// Decode the pod from the request
 	var pod corev1.Pod
 	if err := json.Unmarshal(req.Object.Raw, &pod); err != nil {
-		log.Printf("Could not unmarshal request to pod: %s", err)
```
lkysow: :beer

connect-inject/handler.go
```diff
@@ -181,10 +220,10 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
+		h.Log.Error("Could not unmarshal request to pod", "Error", err)
```
lkysow: do we use `Error` or `err` as the key elsewhere? `err` seems more idiomatic to me

connect-inject/handler.go
```diff
@@ -296,14 +339,56 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
 		resp.PatchType = &patchType
 	}
 
+	// Check and potentially create Consul resources. This is done after
+	// all patches are created to guarantee the mutation succeeded before
```
lkysow: technically this doesn't guarantee the mutation succeeded since we still need to pass it back kube right? So maybe add a little here about how we know there's no errors on our side

connect-inject/handler.go
```diff
@@ -393,6 +478,76 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+	namespaceInfo, _, err := h.ConsulClient.Namespaces().Read(ns, nil)
+	if err != nil {
+		return err
+	}
+
+	// If not, create it
+	if namespaceInfo == nil {
+		consulNamespace := api.Namespace{
+			Name:        ns,
+			Description: "Auto-generated by a Connect Injector",
```
lkysow:
```suggestion
			Description: "Auto-generated by Connect Injector",
```

connect-inject/handler.go
```diff
@@ -393,6 +478,76 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (h *Handler) checkAndCreateBindingRule(ns string) error {
+	// Check if the binding rule exists in this namespace
```
lkysow: Was a bit confused by this comment because my immediate thought was "what if there's more than one binding rule in this namespace". I realized after a bit that you're only checking for binding rules for our auth method so maybe you can clarify that.

connect-inject/handler.go
```diff
@@ -393,6 +478,76 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+	existingRules, _, err := h.ConsulClient.ACL().BindingRuleList(h.AuthMethod, &api.QueryOptions{Namespace: ns})
+	if err != nil {
+		return err
+	}
+
+	// When there are no binding rules for the current namespace, one needs to be created
```
lkysow:
```suggestion
	// When there are no binding rules for the current namespace and our auth method, one needs to be created
```

connect-inject/handler.go
```diff
@@ -393,6 +478,76 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+	if len(existingRules) == 0 {
+		// Note: This description uses the AuthMethod name which differs from binding rules
+		// create in the ACL bootstrap method where the release name is used. It didn't seem
```
lkysow: This was changed due to the fullname support PR. It's now just `Consul default binding rule` I think

lkysow: I agree it's not super important due to the name.

connect-inject/handler.go
```diff
@@ -393,6 +478,76 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+		abr := api.ACLBindingRule{
+			Description: fmt.Sprintf("Consul %s default binding rule", h.AuthMethod),
+			AuthMethod:  h.AuthMethod,
+			BindType:    api.BindingRuleBindTypeService,
+			BindName:    "${serviceaccount.name}",
+			Selector:    h.BindingRuleSelector,
+			Namespace:   ns,
+		}
+		_, _, err = h.ConsulClient.ACL().BindingRuleCreate(&abr, nil)
+		if err != nil {
```
lkysow: could just return `err` here ;)

subcommand/inject-connect/command.go
```diff
@@ -57,17 +80,35 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagKeyFile, "tls-key-file", "",
 		"PEM-encoded TLS private key to serve. If blank, will generate random cert.")
 	c.flagSet.StringVar(&c.flagConsulImage, "consul-image", connectinject.DefaultConsulImage,
-		"Docker image for Consul. Defaults to an Consul 1.3.0.")
+		"Docker image for Consul. Defaults to an Consul 1.5.0.")
```
```suggestion
		"Docker image for Consul. Defaults to Consul 1.5.0.")
```

subcommand/inject-connect/command.go
```diff
@@ -57,17 +80,35 @@ func (c *Command) init() {
...
 		"Write a service-defaults config for every Connect service using protocol from -default-protocol or Pod annotation.")
 	c.flagSet.StringVar(&c.flagDefaultProtocol, "default-protocol", "",
 		"The default protocol to use in central config registrations.")
+	c.flagSet.BoolVar(&c.flagEnableNamespaces, "enable-namespaces", false,
+		"Enables namespaces, in either a single Consul namespace or mirrored [Enterprise only feature]")
+	c.flagSet.StringVar(&c.flagConsulNamespace, "consul-namespace", "default",
+		"Defines which Consul namespace to register all injected services into. If `enable-namespace-mirroring` "+
+			"is true, this is not used.")
+	c.flagSet.Var((*flags.AppendSliceValue)(&c.flagAllowK8sNamespacesList), "allow-namespace",
+		"K8s namespaces to explicitly allow. May be specified multiple times.")
```
lkysow: Why not do a comma separated list? Might be nicer to use for users

connect-inject/handler.go
```diff
@@ -201,16 +240,18 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
-				Message: err.Error(),
+				Message: fmt.Sprintf("Error creating default annotations: %s", err),
 			},
 		}
 	}
 
 	// Check if we should inject, for example we don't inject in the
 	// system namespaces.
 	if shouldInject, err := h.shouldInject(&pod, req.Namespace); err != nil {
+		h.Log.Error("Error checking if should inject", "Error", err, "Pod Name", pod.Name)
```
lkysow: nice work adding all this logging, especially with the pod name

connect-inject/handler.go
```diff
@@ -296,14 +339,56 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
+				"Namespace", h.consulNamespace(pod.Namespace), "Pod Name", pod.Name)
+			return &v1beta1.AdmissionResponse{
+				Result: &metav1.Status{
+					Message: fmt.Sprintf("Error checking or creating namespace: %s", err),
+				},
+			}
+		}
+
+		// For mirroring, make sure a binding rule exists if ACLs are enabled
+		if h.EnableNSMirroring {
```
lkysow:
```suggestion
		if h.EnableNSMirroring && h.AuthMethod != "" {
```

connect-inject/handler.go
```diff
@@ -393,6 +478,76 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
 	return nil
 }
 
+func (h *Handler) consulNamespace(ns string) string {
```
lkysow: can you add docs for this

connect-inject/container_init.go
```diff
@@ -21,6 +21,7 @@ type initContainerCommandData struct {
 	// WriteServiceDefaults controls whether a service-defaults config is
 	// written for this service.
 	WriteServiceDefaults bool
+	ConsulNamespace      string
```
lkysow: Can you add a doc for this and also how if it's empty that means we're not using namespaces

connect-inject/container_init.go
adilyse: The code currently is implemented as `-ns`, but will be switched to `-namespace` for the final release.

connect-inject/envoy_sidecar.go
adilyse: Updated.

connect-inject/handler.go
adilyse: Updated.

connect-inject/handler.go
adilyse: The existing hclogging in the connect injector was already using `Error`, so I matched it. But you're right, it looks like the other components are using `err` so I'll switch it.

adilyse:
The lifecycle sidecar shouldn't need updating. The service registration that it is using (that was written by the connect injector) will include the namespace if it needs one. Consul will correctly interpret this without any additional config/flags being necessary.

subcommand/inject-connect/command.go
adilyse: We can expose this any way we like in the Helm chart, then do the translation there. This also fits existing patterns, as demonstrated by the server's extra volumes: https://github.com/hashicorp/consul-helm/blob/master/templates/server-statefulset.yaml#L105-L109

lkysow: Okay makes sense. I guess in the end the UX is more about how it's set in the helm chart.

connect-inject/handler.go
adilyse: Heh, whoops. Updated.

connect-inject/handler.go
adilyse: Updated and pulled out the first comment.

connect-inject/container_init.go
adilyse: Added.

subcommand/inject-connect/command.go
adilyse: Updating the default image to be 1.7.0 as it will be needed for namespaces.

connect-inject/handler.go
adilyse: Definitely could! I'm going to leave it because I like the fall through `return nil` to apply to the whole function.

connect-inject/handler.go
adilyse: Updated.

connect-inject/handler.go
adilyse: Updated.

connect-inject/handler.go
adilyse: I phrased it this way because namespaces cut across datacenters, so if someone is running multiple k8s clusters with various injectors, only one of them will have created it. Maybe that's pedantic and potentially confusing, but technically more accurate.

connect-inject/handler.go
adilyse: Updated to be clearer that it's just our patch creation we're waiting on. If kube fails this afterwards, we will have already modified Consul. Not sure there's a way around that.

subcommand/inject-connect/command_test.go
```diff
@@ -1,4 +1,4 @@
-package subcommand
+package connectinject
 
 import (
 	"github.com/mitchellh/cli"
```
ishustava: Do we need command tests too?

adilyse:
It turns out that we won't need to create binding rules after all, as having it defined in the Consul `default` namespace will be enough. Pulling all of that out.

subcommand/inject-connect/command.go
```diff
@@ -59,19 +82,38 @@ func (c *Command) init() {
...
 		"The name of the Kubernetes Auth Method to use for connectInjection if ACLs are enabled.")
-	c.flagSet.BoolVar(&c.flagCentralConfig, "enable-central-config", false,
+	c.flagSet.BoolVar(&c.flagWriteServiceDefaults, "enable-central-config", false,
 		"Write a service-defaults config for every Connect service using protocol from -default-protocol or Pod annotation.")
 	c.flagSet.StringVar(&c.flagDefaultProtocol, "default-protocol", "",
 		"The default protocol to use in central config registrations.")
 	c.flagSet.StringVar(&c.flagConsulCACert, "consul-ca-cert", "",
 		"Path to CA certificate to use if communicating with Consul clients over HTTPS.")
+	c.flagSet.BoolVar(&c.flagEnableNamespaces, "enable-namespaces", false,
+		"Enables namespaces, in either a single Consul namespace or mirrored [Enterprise only feature]")
```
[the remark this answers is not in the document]
adilyse: If we put namespace info into anything in OSS where they're not supported, things will fail and fail hard. Someone can very easily set the Consul namespace to a value (and in fact I want it to default to `default` in the Helm chart) and then get into a lot of trouble.

We could add all sorts of extra validation in the Helm chart to make sure they're running enterprise, but that gets complicated quickly as there are 5 different places that could be specified in the values file. What happens if one of them is enterprise but the others aren't? It seems more straightforward to provide a single on/off switch that can be clearly labeled as an enterprise-only feature.

connect-inject/handler.go
```diff
@@ -124,6 +124,45 @@ type Handler struct {
...
+	// injection. It supports the special character `*` which indicates that
+	// all k8s namespaces are eligible unless explicitly denied. This filter
+	// is applied before checking pod annotations.
+	AllowK8sNamespacesSet mapset.Set
+
+	// DenyK8sNamespacesList is a list of k8s namespaces to explicitly deny
+	// injection and thus service registration with Consul. An empty list
+	// means that no namespaces are removed from consideration. This filter
+	// takes precedence over AllowK8sNamespacesList.
+	DenyK8sNamespacesSet mapset.Set
```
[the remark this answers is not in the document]
adilyse: The deny list takes precedence, so if it had a `*`, everything would be denied. An empty set for this indicates that nothing is a hard denial and lets the allow list do the work.

connect-inject/handler.go
```diff
@@ -301,14 +344,53 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
 		resp.PatchType = &patchType
 	}
 
+	// Check and potentially create Consul resources. This is done after
+	// all patches are created to guarantee no errors were encountered in
+	// that process before modifying the Consul cluster.
+	if h.EnableNamespaces {
+		// Check if the namespace exists. If not, create it.
+		if err := h.checkAndCreateNamespace(h.consulNamespace(pod.Namespace)); err != nil {
```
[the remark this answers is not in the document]
adilyse: Creating a namespace requires `operator: write` ACLs privileges. If we created the namespace in the init container, every pod would need this and so they would be incredibly over-privileged. This keeps those permissions only in the connect injector.

connect-inject/handler.go
```diff
@@ -301,14 +344,53 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
-	for _, ns := range kubeSystemNamespaces {
-		if namespace == ns {
+	if kubeSystemNamespaces.Contains(namespace) {
+		return false, nil
+	}
+
+	// Namespace logic
+	if h.EnableNamespaces {
+		// If in deny list, don't inject
+		if h.DenyK8sNamespacesSet.Contains(namespace) {
```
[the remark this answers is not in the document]
adilyse: Short circuiting to a `true` return value for any of the checks in this method would mean that we're not checking any of the other requirements for injection. Currently, if a namespace is not in the deny list, it still might not be in the allow list and we shouldn't inject. Or it might specifically is annotated to not be injected. The deny list info is not enough to make that determination on its own.

subcommand/inject-connect/command.go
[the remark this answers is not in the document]
adilyse: I described the reasoning for this in another comment. Basically, I'm trying to save us some nasty `ent` image validation in the Helm chart and allow us to default to `default` for the Consul namespace name flag.

connect-inject/handler.go
```diff
@@ -124,6 +124,45 @@ type Handler struct {
 	// If not set, will use HTTP.
 	ConsulCACert string
 
+	// EnableNamespaces indicates that a user is running Consul Enterprise
+	// with version 1.7+ which is namespace aware. It enables Consul namespaces,
+	// with injection into either a single Consul namespace or mirrored from
+	// k8s namespaces.
+	EnableNamespaces bool
```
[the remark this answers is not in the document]
adilyse: This is an indicator that marks that someone is running enterprise and can support namespaces. The Consul namespace name flag defaults to `default`, so it sets up injecting into a single namespace. (Also addressed in other comments)

subcommand/inject-connect/command.go
```diff
@@ -1,4 +1,4 @@
-package subcommand
+package connectinject
```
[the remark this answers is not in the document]
lkysow: It's easy to add another block to `commands.go` for `inject-connect` and then rename everything.

connect-inject/container_init.go
```diff
@@ -298,31 +310,45 @@ cat <<EOF >/consul/connect-inject/service-defaults.hcl
...
   /consul/connect-inject/service-defaults.hcl || true
 {{- end }}
 
 /bin/consul services register \
   {{- if .AuthMethod }}
   -token-file="/consul/connect-inject/acl-token" \
   {{- end }}
+  {{- if .ConsulNamespace }}
+  -namespace={{ .ConsulNamespace }} \
+  {{- end }}
```
ishustava: Also noticed the other day that `/bin/consul connect envoy` command needs the namespace flag (GH doesn't let me comment on the code that wasn't changed). One thing we can do is just provide the `CONSUL_NAMESPACE` env variable to the container and remove the flag from all commands to simplify it.

connect-inject/container_init.go
```diff
@@ -111,14 +112,25 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
...
 			if parts[0] == "prepared_query" {
 				port, _ = portValue(pod, strings.TrimSpace(parts[2]))
 				prepared_query = strings.TrimSpace(parts[1])
 			} else {
 				port, _ = portValue(pod, strings.TrimSpace(parts[1]))
-				service_name = strings.TrimSpace(parts[0])
+
+				// Parse the namespace if provided
+				if data.ConsulNamespace != "" {
+					pieces := strings.SplitN(parts[0], ".", 2)
```
ishustava: If somebody provides an invalid string `service.namespace.dc`, then the result in `pieces` will be `[service namespace.dc]`. We should probably detect it and either log it and ignore everything past the second dot, or error.

connect-inject/container_init.go
```diff
@@ -127,12 +139,19 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
...
 					LocalPort:  port,
 					Datacenter: datacenter,
 					Query:      prepared_query,
-				})
+				}
+
+				// Add namespace to upstream
+				if namespace != "" {
+					upstream.ConsulUpstreamNamespace = namespace
+				}
```
ishustava: Do we need this `if`? If `ConsulUpstreamNamespace` is an empty string, then the template ` {{- if .ConsulUpstreamNamespace }}` below will evaluate to `false` I believe.

connect-inject/container_init_test.go
```diff
@@ -494,6 +495,233 @@ services {
...
+# Copy the Consul binary
+cp /bin/consul /consul/connect-inject/consul`,
+			"",
+		},
+
+		{
+			"Upstream namespace",
+			func(pod *corev1.Pod) *corev1.Pod {
+				pod.Annotations[annotationService] = "web"
+				pod.Annotations[annotationUpstreams] = "db.namespace:1234"
```
ishustava: It'd be good to add some invalid cases too.

subcommand/inject-connect/command.go
```diff
@@ -23,23 +25,43 @@ import (
...
+	// Flags to support namespaces
+	flagEnableNamespaces       bool     // Use namespacing on all components
+	flagConsulNamespace        string   // Consul namespace to register everything if not mirroring
+	flagAllowK8sNamespacesList []string // K8s namespaces to explicitly inject
+	flagDenyK8sNamespacesList  []string // K8s namespaces to deny injection (has precedence)
+	flagEnableNSMirroring      bool     // Enables mirroring of k8s namespaces into Consul
+	flagMirroringPrefix        string   // Prefix added to Consul namespaces created when mirroring
+
+	flagSet *flag.FlagSet
+	http    *flags.HTTPFlags
```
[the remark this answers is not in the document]
adilyse: Let's file a ticket and address it later.

ishustava:
I tested this out and so far everything looks pretty good! 

A couple of issues I've found so far:
1. [Must Have] When mirroring is enabled, injector doesn’t create the namespace if using type deployment rather than pod. It fails with: 
    ```
     2020-02-06T02:21:32.820Z [ERROR] handler: Error checking or creating namespace: err="Unexpected response code: 400 (Bad request: Missing namespace name)" Namespace= Pod Name=
    ```
1. [Nice to Have] We don’t do any validation of the namespace prefix or the destination namespace. Consul will fail anyway but we should fail at deploy time.  
    1. The namespace in Consul must be a valid DNS name
        1. Only alphanumeric chars and a hyphen
        3. Mirroring prefix can’t start with a `-`
     1. We should limit the size of the prefix to 5-10 chars or so since both Kubernetes and consul namespaces have a limit of 63 chars

connect-inject/container_init.go
```diff
@@ -298,31 +332,45 @@ cat <<EOF >/consul/connect-inject/service-defaults.hcl
...
+{{- end }}
 EOF
 {{- end }}
+
 {{- if .AuthMethod }}
 /bin/consul login -method="{{ .AuthMethod }}" \
   -bearer-token-file="/var/run/secrets/kubernetes.io/serviceaccount/token" \
   -token-sink-file="/consul/connect-inject/acl-token" \
+  {{- if .ConsulNamespace }}
+  -namespace="{{ .ConsulNamespace }}" \
```
ishustava: If mirroring is enabled, then the auth method is created in the default namespace. In that case, using `-namespace` flag results in this error:
```
Error logging in: Unexpected response code: 403 (rpc error making call: rpc error making call: ACL not found)
```

lkysow: Which is weird because I thought it was supposed to work globally when set in the default namespace...

ishustava: FWIW I've tried running it locally against a Consul agent on Kubernetes, and running the same command without `-namespace` works as expected (Consul puts the token in the token sink file) 🤷‍♀ .

adilyse: Adding the `-namespace` flag to this means that it's explicitly looking for an auth method in the given namespace. Because we've defined the auth method for mirroring in `default`, it doesn't find anything.

Removing the namespace flag means that the call falls back to running against the `default` namespace, which means that it works in this case because that's where the auth method is defined.

To be explicit, I'm going to set `-namespace="default"` for the mirroring case rather than relying on the fallback behavior. Just sorting out the tests around it now.

connect-inject/handler.go
```diff
@@ -124,6 +124,45 @@ type Handler struct {
...
+
+	// AllowK8sNamespacesList is a list of k8s namespaces to explicitly allow for
```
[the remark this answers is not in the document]
adilyse: Updated.

subcommand/inject-connect/command.go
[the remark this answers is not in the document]
adilyse: Yes, later!

subcommand/inject-connect/command_test.go
adilyse: Tests to be addressed separately.

connect-inject/handler.go
```diff
@@ -206,16 +240,18 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
-				Message: err.Error(),
+				Message: fmt.Sprintf("Error creating default annotations: %s", err),
 			},
 		}
 	}
 
 	// Check if we should inject, for example we don't inject in the
 	// system namespaces.
 	if shouldInject, err := h.shouldInject(&pod, req.Namespace); err != nil {
+		h.Log.Error("Error checking if should inject", "err", err, "Pod Name", pod.Name)
```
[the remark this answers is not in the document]
adilyse: Switching this to the request name. I'm not sure that there's a guarantee that we'll always be able to pull the correct useful field out of things, but hopefully this helps.

connect-inject/handler.go
```diff
@@ -398,6 +462,47 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+		return err
+	}
+
+	// If not, create it
+	if namespaceInfo == nil {
+		consulNamespace := api.Namespace{
+			Name:        ns,
+			Description: "Auto-generated by a Connect Injector",
+			// TODO: when metadata is added to the api
+			// Meta:        map[string]string{"external-source": "kubernetes"},
```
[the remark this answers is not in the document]
adilyse: Updated.

connect-inject/handler_test.go
```diff
@@ -640,7 +641,305 @@ func TestHandlerPortValue(t *testing.T) {
...
+				},
+			},
+			"default",
+			true,
+			mapset.NewSetWith("*", "default"),
+			mapset.NewSet(),
+			true,
+		},
+		{
+			"namespaces enabled, allow only non-default",
```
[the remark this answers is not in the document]
adilyse: Updated.

subcommand/inject-connect/command.go
```diff
@@ -59,19 +81,37 @@ func (c *Command) init() {
...
+		"Docker image for Consul. Defaults to Consul 1.7.0.")
 	c.flagSet.StringVar(&c.flagEnvoyImage, "envoy-image", connectinject.DefaultEnvoyImage,
-		"Docker image for Envoy. Defaults to Envoy 1.8.0.")
+		"Docker image for Envoy. Defaults to Envoy 1.9.1.")
 	c.flagSet.StringVar(&c.flagConsulK8sImage, "consul-k8s-image", "",
 		"Docker image for consul-k8s. Used for the connect sidecar.")
 	c.flagSet.StringVar(&c.flagACLAuthMethod, "acl-auth-method", "",
 		"The name of the Kubernetes Auth Method to use for connectInjection if ACLs are enabled.")
-	c.flagSet.BoolVar(&c.flagCentralConfig, "enable-central-config", false,
+	c.flagSet.BoolVar(&c.flagWriteServiceDefaults, "enable-central-config", false,
```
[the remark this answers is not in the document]
adilyse: Everywhere else in the code, this setting was referred to as `WriteServiceDefaults`, so the previous use of the flag was `WriteServiceDefaults: c.flagCentralConfig,`. It didn't make sense, so I went ahead and updated the flag field name to match how it was used elsewhere. 

It is a non-breaking change this way, but going all the way to renaming the flag is more invasive, so I didn't update that at this time. It's something we should consider in the future.

subcommand/inject-connect/command.go
```diff
@@ -59,19 +81,37 @@ func (c *Command) init() {
...
+		"[Enterprise Only] Defines which Consul namespace to register all injected services into. If '-enable-namespace-mirroring' "+
+			"is true, this is not used.")
+	c.flagSet.BoolVar(&c.flagEnableK8SNSMirroring, "enable-k8s-namespace-mirroring", false, "[Enterprise Only] Enables "+
+		"k8s namespace mirroring")
+	c.flagSet.StringVar(&c.flagK8SNSMirroringPrefix, "k8s-namespace-mirroring-prefix", "",
+		"[Enterprise Only] Prefix that will be added to all k8s namespaces mirrored into Consul if mirroring is enabled.")
+
+	c.http = &flags.HTTPFlags{}
+	flags.Merge(c.flagSet, c.http.ClientFlags())
+	flags.Merge(c.flagSet, c.http.ServerFlags())
```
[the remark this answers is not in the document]
adilyse: I was mirroring the sync catalog client flag setup, but maybe it's not needed there either? https://github.com/hashicorp/consul-k8s/blob/master/subcommand/sync-catalog/command.go#L108

connect-inject/container_init.go
adilyse: Updated.

connect-inject/handler_test.go
```diff
@@ -640,7 +641,305 @@ func TestHandlerPortValue(t *testing.T) {
...
+			false,
+			"default",
+			true,
+			"test-",
+			"namespace",
+			"",
+		},
+
+		{
+			"namespaces enabled, mirroring diabled",
```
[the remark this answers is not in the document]
adilyse: Updated.

connect-inject/handler.go
```diff
@@ -124,6 +124,40 @@ type Handler struct {
...
+	// takes precedence over AllowK8sNamespacesSet.
+	DenyK8sNamespacesSet mapset.Set
+
+	// ConsulDestinationNamespace is the name of the Consul namespace to register all
+	// injected services into if Consul namespaces are enabled and mirroring
+	// is disabled. This will not be used if mirroring is enabled.
+	ConsulDestinationNamespace string
+
+	// EnableK8SNSMirroring causes Consul namespaces to be created to match the
+	// organization within k8s. Services are registered into the Consul
```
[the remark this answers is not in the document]
adilyse: Updated the comment.

connect-inject/handler.go
```diff
@@ -124,6 +124,40 @@ type Handler struct {
...
+
+	// DenyK8sNamespacesSet is a set of k8s namespaces to explicitly deny
+	// injection and thus service registration with Consul. An empty set
+	// means that no namespaces are removed from consideration. This filter
+	// takes precedence over AllowK8sNamespacesSet.
+	DenyK8sNamespacesSet mapset.Set
+
+	// ConsulDestinationNamespace is the name of the Consul namespace to register all
+	// injected services into if Consul namespaces are enabled and mirroring
+	// is disabled. This will not be used if mirroring is enabled.
```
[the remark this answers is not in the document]
adilyse: Updated to clarify.

connect-inject/container_init_test.go
```diff
@@ -494,6 +495,233 @@ services {
...
+			require.NoError(err)
+			actual := strings.Join(container.Command, " ")
+			require.Contains(actual, tt.Cmd)
+			if tt.CmdNot != "" {
+				require.NotContains(actual, tt.CmdNot)
+			}
+		})
+	}
+}
+
```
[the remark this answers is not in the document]
adilyse: Added.

connect-inject/envoy_sidecar.go
```diff
@@ -74,11 +83,15 @@ func (h *Handler) envoySidecar(pod *corev1.Pod) (corev1.Container, error) {
...
+  {{- if .ConsulNamespace }}
+  -namespace="{{ .ConsulNamespace }}" \
+  {{- end }}
   /consul/connect-inject/service.hcl
-{{- if . }}
+
+{{- if .AuthMethod }}
 && /consul/connect-inject/consul logout \
   -token-file="/consul/connect-inject/acl-token"
 {{- end}}
```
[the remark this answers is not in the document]
adilyse: There's an auth method test, a namespace test and one that addresses both.

connect-inject/handler_test.go
```diff
@@ -640,7 +641,305 @@ func TestHandlerPortValue(t *testing.T) {
...
+					},
+				},
+			},
+			"kube-public",
+			false,
+			mapset.NewSet(),
+			mapset.NewSet(),
+			false,
+		},
+		{
```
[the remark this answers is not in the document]
adilyse: The important part of the tests are whether the namespace we're coming from is contained in either the allow or deny list and if that gives us the correct response about whether we should inject. These tests would cover the same cases even if I used a namespace other than default.

subcommand/inject-connect/command.go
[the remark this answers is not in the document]
ishustava: Hmm that's interesting. Those add `datacenter` and `stale` flags. I suppose it doesn't hurt to have them in case someone wants to point their injector at the servers, but if we need to switch away from using http flags, that means duplicating them in our own library to keep it backward compatible. My personal preference is to keep it minimal.

connect-inject/handler_test.go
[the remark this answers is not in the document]
ishustava: What I was trying to say was that using only one namespace is insufficient. For example, for the test case I mentioned there, asserting only on `shouldInject(..., default) == false` is not fully covering that case because it doesn't check that the namespace in the allow list is still allowed. We should be checking both the allow and deny behaviors, but using a single namespace in assertions only allows us to check one. Hope this makes sense 😀

connect-inject/envoy_sidecar.go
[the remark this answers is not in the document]
ishustava: Great, thanks!

PR #30, opened 2018-11-02
connect-inject: adds TLS support for communicating with Consul

epd:
This pull request in general adds support for TLS communication to Consul with the connect-inject web hook. I have a cluster that enforces TLS for both HTTP and GRPC and uses an internal certificate authority. For this reason, it was necessary to augment some of the ways in which both the init containers and sidecar containers are created. 

It adds the following flags to the `connect-inject` command:
- `-consul-cacert`: Kubernetes secret name and mount path for the CA certificate to use for Consul communication.
  - e.g. `-consul-cacert=my-secret-name:/path/to/ca.crt`
- `-consul-tls-server-name`: SNI hostname to use for Consul communication over TLS.
  - e.g. `-consul-tls-server-name=client.us-east-1.consul`
- `-consul-http-ssl`: Whether or not to enable TLS for Consul communication over HTTP.
- `-consul-grpc-ssl`: Whether or not to enable TLS for Consul communication over GRPC.

PR #181, opened 2019-12-17
Enable TLS for connect-inject

### LINKED ISSUE #79: consul-k8s Connect inject does not support consul with TLS enabled and HTTP disabled.
I spent a bunch of time getting TLS set up by modifying the helm chart to suite my needs, and everything seems to be working with consul itself. 

I would like to take advantage of connect-inject, because it is a cool feature, and makes my life a lot easier. Just specify an annotation, and get service level TLS without additional config... that's awesome. 

However, after getting consul servers and agents talking with TLS, and gossip encryption this fails to work anymore. 

```
{
  "verify_incoming": true,
  "verify_outgoing": true,
  "verify_server_hostname": true,
  "ca_file": "consul-agent-ca.pem",
  "cert_file": "dc1-client-consul-0.pem",
  "key_file": "dc1-client-consul-0-key.pem",
  "ports": {
    "http": -1,
    "https": 8501
  }
}
```

Using this config for my clients to enable tls, and disable the HTTP endpoint breaks the connect-inject sidecar.

1) The injection for the sidecar assumes HTTP, as seen below: 

```

# Register the service. The HCL is stored in the volume so that
# the preStop hook can access it to deregister the service.
cat <<EOF >/consul/connect-inject/service.hcl
services {
  id   = "${POD_NAME}-{{ .ServiceName }}-proxy"
  name = "{{ .ServiceName }}-proxy"
  kind = "connect-proxy"
  address = "${POD_IP}"
  port = 20000
  proxy {
    destination_service_name = "{{ .ServiceName }}"
    destination_service_id = "{{ .ServiceName}}"
    {{ if (gt .ServicePort 0) -}}
    local_service_address = "127.0.0.1"
    local_service_port = {{ .ServicePort }}
    {{ end -}}
    {{ range .Upstreams -}}
    upstreams {
      {{- if .Name }}
      destination_type = "service" 
      destination_name = "{{ .Name }}"
      {{- end}}
      {{- if .Query }}
      destination_type = "prepared_query" 
      destination_name = "{{ .Query}}"
      {{- end}}
      local_bind_port = {{ .LocalPort }}
      {{- if .Datacenter }}
      datacenter = "{{ .Datacenter }}"
      {{- end}}
    }
    {{ end }}
  }
  checks {
    name = "Proxy Public Listener"
    tcp = "${POD_IP}:20000"
    interval = "10s"
    deregister_critical_service_after = "10m"
  }
  checks {
    name = "Destination Alias"
    alias_service = "{{ .ServiceName }}"
  }
}
EOF
/bin/consul services register /consul/connect-inject/service.hcl
# Generate the envoy bootstrap code
/bin/consul connect envoy \
  -proxy-id="${POD_NAME}-{{ .ServiceName }}-proxy" \
  -bootstrap > /consul/connect-inject/envoy-bootstrap.yaml
# Copy the Consul binary
cp /bin/consul /consul/connect-inject/consul
```

2) There is no way for me to override that configuration, or modify it without just writing my own init container. 

See error below: 

```
NAME                            READY   STATUS                  RESTARTS   AGE
sample-nginx-68c7d9bdd7-p6fxt   0/2     Init:CrashLoopBackOff   9          25m

$ kubectl logs sample-nginx-68c7d9bdd7-p6fxt consul-connect-inject-init
Error registering service "nginx-proxy": Put http://<ip-redacted>:8500/v1/agent/service/register: dial tcp <ip redacted>:8500: connect: connection refused
```

I would like the ability to tell connect inject that it needs to use a cert, key, and ca from a secret, like I can do for the consul servers, and clients.  That way, I can set up some certificates that the init container can use when registering the service, without having to enable HTTP connections on my consul agents. 

At minimum, it would be more secure if all I had to do was disable "verify_incoming_https" on my consul agents, and specify that I want an HTTPS connection, and which port to use. 

Maybe there is a better solution, but I figured I would put this here so it can be discussed. 

Edit: 

If it was not obvious from above, I am running consul in Kubernetes, using the helm chart to deploy.    
I am running - kubernetes 1.11.  
Consul 1.4.2
Consul-k8s 0.6

The above init config can be seen hardcoded here  : https://github.com/hashicorp/consul-k8s/blob/master/connect-inject/container_init.go

ishustava:
* Add new flag -consul-ca-cert
* Make Consul addresses use HTTPS if CA is provided
* Provide CA certificate to the init and both sidecar containers,
  so that service registration and envoy bootstrapping
  can use TLS.

This PR has a slightly different approach to what was proposed in #30 and assumes the implementation in the hashicorp/consul-helm#313.

Fixes #79

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -24,6 +24,10 @@ type initContainerCommandData struct {
 	Upstreams            []initContainerCommandUpstreamData
 	Tags                 string
 	Meta                 map[string]string
+
+	// The PEM-encoded CA certificate to use when
+	// communicating with Consul clients
+	ConsulCACert string
 }
 
 type initContainerCommandUpstreamData struct {
@@ -52,6 +56,7 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
 		ServiceProtocol:      protocol,
 		AuthMethod:           h.AuthMethod,
 		WriteServiceDefaults: writeServiceDefaults,
+		ConsulCACert:         h.ConsulCACert,
 	}
 	if data.ServiceName == "" {
 		// Assertion, since we call defaultAnnotations above and do
@@ -199,8 +204,17 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
 // initContainerCommandTpl is the template for the command executed by
 // the init container.
 const initContainerCommandTpl = `
+{{- if .ConsulCACert}}
+export CONSUL_HTTP_ADDR="https://${HOST_IP}:8501"
+export CONSUL_GRPC_ADDR="https://${HOST_IP}:8502"
+export CONSUL_CACERT=/consul/connect-inject/consul-ca.pem
+cat <<EOF >/consul/connect-inject/consul-ca.pem
+{{ .ConsulCACert }}
+EOF
+{{- else}}
 export CONSUL_HTTP_ADDR="${HOST_IP}:8500"
 export CONSUL_GRPC_ADDR="${HOST_IP}:8502"
+{{- end}}
 
 # Register the service. The HCL is stored in the volume so that
 # the preStop hook can access it to deregister the service.
--- a/connect-inject/container_sidecar_test.go
+++ b/connect-inject/container_sidecar_test.go
@@ -1,0 +1,137 @@
+package connectinject
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestHandlerContainerSidecar(t *testing.T) {
+	require := require.New(t)
+	h := Handler{}
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{
+				annotationService: "foo",
+			},
+		},
+
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name: "web",
+				},
+			},
+		},
+	}
+	container, err := h.envoySidecar(pod)
+	require.NoError(err)
+	require.Equal(container.Command, []string{
+		"envoy",
+		"--max-obj-name-len", "256",
+		"--config-path", "/consul/connect-inject/envoy-bootstrap.yaml",
+	})
+
+	preStopCommand := strings.Join(container.Lifecycle.PreStop.Exec.Command, " ")
+	require.Equal(preStopCommand, `/bin/sh -ec /consul/connect-inject/consul services deregister \
+  /consul/connect-inject/service.hcl`)
+
+	require.Equal(container.VolumeMounts, []corev1.VolumeMount{
+		{
+			Name:      volumeName,
+			MountPath: "/consul/connect-inject",
+		},
+	})
+
+	require.Equal(container.Env, []corev1.EnvVar{
+		{
+			Name: "HOST_IP",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{FieldPath: "status.hostIP"},
+			},
+		},
+		{
+			Name:  "CONSUL_HTTP_ADDR",
+			Value: "$(HOST_IP):8500",
+		},
+	})
+}
+
+// Test that if AuthMethod is set
+// the preStop command includes a token
+func TestHandlerContainerSidecar_AuthMethod(t *testing.T) {
+	require := require.New(t)
+	h := Handler{
+		AuthMethod: "test-auth-method",
+	}
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{
+				annotationService: "foo",
+			},
+		},
+
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name: "web",
+				},
+			},
+		},
+	}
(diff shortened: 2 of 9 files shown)


subcommand/inject-connect/command.go
```diff
@@ -117,6 +121,16 @@ func (c *Command) Run(args []string) int {
 	defer cancelFunc()
 	go c.certWatcher(ctx, certCh, clientset)
 
+	var consulCACert []byte
+	if c.flagConsulCACert != "" {
+		var err error
+		consulCACert, err = ioutil.ReadFile(c.flagConsulCACert)
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error reading Consul's CA cert file: %s", err))
```
```suggestion
			c.UI.Error(fmt.Sprintf("Error reading Consul's CA cert file %s: %s", c.flagConsulCACert, err))
```

Might be nice for the error to have the path of the file it's reading

connect-inject/container_init.go
```diff
@@ -199,8 +204,17 @@ func (h *Handler) containerInit(pod *corev1.Pod) (corev1.Container, error) {
 // initContainerCommandTpl is the template for the command executed by
 // the init container.
 const initContainerCommandTpl = `
+{{- if .ConsulCACert}}
+export CONSUL_HTTP_ADDR="https://${HOST_IP}:8501"
+export CONSUL_GRPC_ADDR="https://${HOST_IP}:8502"
+export CONSUL_CACERT=/consul/connect-inject/consul-ca.pem
+cat <<EOF >/consul/connect-inject/consul-ca.pem
+{{ .ConsulCACert }}
```
lkysow: Why do this instead of mounting in the same volume in the initContainer?

lkysow: We talked offline and this makes sense to me now based on the tradeoffs. It's nicer than mounting in a secret and fits with the pattern we're following where everything lives in that directory (like the acl token file)

connect-inject/container_sidecar_test.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestHandlerContainerSidecar(t *testing.T) {
```
lkysow: this is now "envoySidecar" I thnk

connect-inject/envoy_sidecar.go
```diff
@@ -52,11 +52,27 @@ func (h *Handler) envoySidecar(pod *corev1.Pod) (corev1.Container, error) {
 			"--max-obj-name-len", "256",
 			"--config-path", "/consul/connect-inject/envoy-bootstrap.yaml",
 		},
-	}, nil
+	}
+	if h.ConsulCACert != "" {
+		caCertEnvVar := corev1.EnvVar{
```
lkysow: I don't personally see the benefit of these extra vars, but personal preference

connect-inject/handler.go
```diff
@@ -119,6 +119,10 @@ type Handler struct {
 	// registrations. It will be overridden by a specific annotation.
 	DefaultProtocol string
 
+	// The PEM-encoded CA certificate string
+	// to use when communicating with Consul clients
```
lkysow:
```suggestion
	// to use when communicating with Consul clients over HTTPS.
	// If not set, will use HTTP.
```

subcommand/inject-connect/command.go
```diff
@@ -68,6 +70,8 @@ func (c *Command) init() {
 		"Write a service-defaults config for every Connect service using protocol from -default-protocol or Pod annotation.")
 	c.flagSet.StringVar(&c.flagDefaultProtocol, "default-protocol", "",
 		"The default protocol to use in central config registrations.")
+	c.flagSet.StringVar(&c.flagConsulCACert, "consul-ca-cert", "",
+		"Path to CA certificate to use when communicating with Consul clients.")
```
```suggestion
		"Path to CA certificate to use if communicating with Consul clients over HTTPS.")
```

connect-inject/container_sidecar_test.go
lkysow: and this file should be renamed IIRC

connect-inject/lifecycle_sidecar.go
```diff
@@ -14,6 +14,12 @@ func (h *Handler) lifecycleSidecar(pod *corev1.Pod) corev1.Container {
 	if h.AuthMethod != "" {
 		command = append(command, "-token-file=/consul/connect-inject/acl-token")
 	}
+	if h.ConsulCACert != "" {
+		command = append(command, "-http-addr", "https://${HOST_IP}:8501")
+		command = append(command, "-ca-file", "/consul/connect-inject/consul-ca.pem")
```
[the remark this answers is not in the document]
ishustava: Hmm ok, I think we should make acl token an env variable too to keep it consistent.

connect-inject/envoy_sidecar.go
ishustava: This is needed so that `preStop` command will succeed.

connect-inject/container_sidecar_test.go
ishustava: yep yep! thank you!

connect-inject/lifecycle_sidecar.go
```diff
@@ -28,12 +34,6 @@ func (h *Handler) lifecycleSidecar(pod *corev1.Pod) corev1.Container {
 					FieldRef: &corev1.ObjectFieldSelector{FieldPath: "status.hostIP"},
 				},
 			},
-			// Kubernetes will interpolate HOST_IP when creating this environment
```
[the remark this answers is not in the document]
ishustava: I've added this back!

PR #183, opened 2019-12-20
Support TLS-enabled servers for the server-acl-init command

ishustava:
This PR makes the following changes
- Adds new flags: `-use-https`, `-consul-ca-cert`, and `-consul-tls-server-name`. The last one is specifically useful for the Helm chart since our certificates don't have Pod IPs as SANs. This allows us to set an SNI header with a custom server name.
- Bumps Consul version to 1.6.2 to use `tlsutil` package for certificate generation in tests.

Ideally, I wanted to change this job to use DNS names of the servers for bootstrapping rather than IPs. But I didn't know how to write tests for it without requiring root privileges to run them. More specifically, we could programmatically add Kubernetes DNS names to `etc/hosts` but that will require `sudo` on the machine where the tests are run. This seemed a bit awkward.

### DIFF
--- a/subcommand/server-acl-init/command.go
+++ b/subcommand/server-acl-init/command.go
@@ -42,6 +42,9 @@ type Command struct {
 	flagCreateMeshGatewayToken   bool
 	flagLogLevel                 string
 	flagTimeout                  string
+	flagConsulCACert             string
+	flagConsulTLSServerName      string
+	flagUseHTTPS                 bool
 
 	clientset kubernetes.Interface
 	// cmdTimeout is cancelled when the command timeout is reached.
@@ -82,6 +85,12 @@ func (c *Command) init() {
 		"Toggle for creating a token for a Connect mesh gateway")
 	c.flags.StringVar(&c.flagTimeout, "timeout", "10m",
 		"How long we'll try to bootstrap ACLs for before timing out, e.g. 1ms, 2s, 3m")
+	c.flags.StringVar(&c.flagConsulCACert, "consul-ca-cert", "",
+		"Path to the PEM-encoded CA certificate of the Consul cluster.")
+	c.flags.StringVar(&c.flagConsulTLSServerName, "consul-tls-server-name", "",
+		"The server name to set as the SNI header when sending HTTPS requests to Consul.")
+	c.flags.BoolVar(&c.flagUseHTTPS, "use-https", false,
+		"Toggle for using HTTPS for all API calls to Consul.")
 	c.flags.StringVar(&c.flagLogLevel, "log-level", "info",
 		"Log verbosity level. Supported values (in order of detail) are \"trace\", "+
 			"\"debug\", \"info\", \"warn\", and \"error\".")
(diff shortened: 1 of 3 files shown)


subcommand/server-acl-init/command.go
```diff
@@ -82,6 +85,12 @@ func (c *Command) init() {
 		"Toggle for creating a token for a Connect mesh gateway")
 	c.flags.StringVar(&c.flagTimeout, "timeout", "10m",
 		"How long we'll try to bootstrap ACLs for before timing out, e.g. 1ms, 2s, 3m")
+	c.flags.StringVar(&c.flagConsulCACert, "consul-ca-cert", "",
+		"Path to the PEM-encoded CA certificate of the Consul cluster.")
+	c.flags.StringVar(&c.flagConsulTLSServerName, "consul-tls-server-name", "",
+		"The server name to set as the SNI header when sending HTTPS requests to Consul.")
+	c.flags.BoolVar(&c.flagUseHTTPS, "use-https", false,
```
lkysow: If we set `consul-ca-cert` or `consul-tls-server-name` with `use-https=false`, does it work? What about vice versa? Wondering if we should error out if not all the flags are set.

subcommand/server-acl-init/command_test.go
```diff
@@ -1146,9 +1151,69 @@ func TestRun_Timeout(t *testing.T) {
 	require.Equal(1, responseCode, ui.ErrorWriter.String())
 }
 
+// Test that the bootstrapping process can make calls to Consul API over HTTPs
```
lkysow:
```suggestion
// Test that the bootstrapping process can make calls to Consul API over HTTPS
```

subcommand/server-acl-init/command_test.go
```diff
@@ -1221,5 +1293,57 @@ func getBootToken(t *testing.T, k8s *fake.Clientset, prefix string) string {
 	return string(bootToken)
 }
 
+// generateServerCerts generates Consul CA
+// and a server certificate and saves them to temp files.
+// It returns file names in this order:
+// CA certificate, server certificate, and server key.
+// Note that it's the responsibility of the caller to
+// remove the temporary files created by this function.
```
lkysow: you could also return a function that the callers can run in a defer that cleans up the files:
```go
func() a (string, func()) {
  return "a", func() {
    os.Remove("a")
  }
}

file, cleanup := a()
defer cleanup()
```

subcommand/server-acl-init/command.go
ishustava: Yeah, good question. If you set `use-https` but choose to add your CA to the system trusted CAs, then this should still work. If you set ca cert and tls server name, but don't enable https, it won't have any effect because the job will just use HTTP for all connections, so I don't want to error here.

subcommand/server-acl-init/command_test.go
ishustava: Oh, that's an awesome idea. I'll do that!

PR #184, opened 2019-12-20
Update ACL bootstrapping job to support namespaces

adilyse:
When namespaces are enabled, all policies and tokens are being created
within the `Consul` default namespace. This is required for any
cross-namespace permissions.


subcommand/server-acl-init/policies.go
```diff
@@ -0,0 +1,208 @@
...
+    policy = "write"
+  }
+{{- if .EnableNamespaces }}
+namespace_prefix "" {
+{{- end }}
+  service_prefix "" {
+    policy = "read"
+  }
+{{- if .EnableNamespaces }}
+}
```
lkysow: I'm finding this pretty hard to read and I'm worried it's fragile to try and get the braces matching with the conditionals. What do you think about duplicating it for clarity:
```
node_prefix "" {
  policy = "write"
}
{{- if .EnableNamespaces }}
namespace_prefix "" {
  service_prefix "" {
    policy = "read"
  }
}
{{- else }}
service_prefix "" {
  policy = "read"
}
{{- end }}

```

subcommand/server-acl-init/policies.go
```diff
@@ -0,0 +1,208 @@
...
+{{- end }}
+`
+
+	// Render the command
```
lkysow: I don't understand this comment. What command?

subcommand/server-acl-init/policies.go
lkysow: There's a couple of methods here where I think this method would make it easier to read.

subcommand/server-acl-init/policies.go
```diff
@@ -0,0 +1,208 @@
...
+	var buf bytes.Buffer
+	tpl := template.Must(template.New("root").Parse(strings.TrimSpace(
+		injectorRulesTpl)))
+	err := tpl.Execute(&buf, &templateData)
+	if err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
+}
```
lkysow: There's a lot of duplication in the rendering code. I like how you've created a common struct for the data `rulesData`. What about creating a helper function that uses it:
```go
func (c *Command) render(tmpl string) (string, error) {
	compiled, err := template.New("root").Parse(strings.TrimSpace(tmpl))
	if err != nil {
		return "", err
	}
	data := rulesData{
		EnableNamespaces:    c.flagEnableNamespaces,
		ConsulSyncNamespace: c.flagConsulSyncNamespace,
		EnableNSMirroring:   c.flagEnableNSMirroring,
		MirroringPrefix:     c.flagMirroringPrefix,
	}
	var buf bytes.Buffer
	err = compiled.Execute(&buf, &data)
	return buf.String(), err
}
```

And then your rules methods will be:
```go
func (c *Command) injectorRules() (string, error) {
	// The Connect injector only needs permissions to create namespaces
	injectorRulesTpl := `
{{- if .EnableNamespaces }}
operator = "write"
{{- end }}
`
	return c.render(injectorRulesTpl)
}
```

subcommand/server-acl-init/command.go
```diff
@@ -83,14 +93,22 @@ func (c *Command) init() {
...
 	c.flags.StringVar(&c.flagConsulCACert, "consul-ca-cert", "",
 		"Path to the PEM-encoded CA certificate of the Consul cluster.")
 	c.flags.StringVar(&c.flagConsulTLSServerName, "consul-tls-server-name", "",
 		"The server name to set as the SNI header when sending HTTPS requests to Consul.")
 	c.flags.BoolVar(&c.flagUseHTTPS, "use-https", false,
 		"Toggle for using HTTPS for all API calls to Consul.")
+	c.flags.BoolVar(&c.flagEnableNamespaces, "enable-namespaces", false,
+		"Enables namespaces, in either a single Consul namespace or mirrored [Enterprise only feature]")
+	c.flags.StringVar(&c.flagConsulSyncNamespace, "consul-sync-namespace", "default",
+		"Defines which Consul namespace to have catalog sync register services into. If `enable-namespace-mirroring` "+
```
lkysow:
```suggestion
		"Defines which Consul namespace to have catalog sync register services into. If `-enable-namespace-mirroring` "+
```

subcommand/server-acl-init/command.go
```diff
@@ -73,8 +81,10 @@ func (c *Command) init() {
 		"Toggle for creating a client agent token")
 	c.flags.BoolVar(&c.flagCreateSyncToken, "create-sync-token", false,
 		"Toggle for creating a catalog sync token")
+	c.flags.BoolVar(&c.flagCreateInjectorToken, "create-injector-namespace-token", false,
```
lkysow: why are you using injector instead of inject here to match the `-create-inject-token` flag?

subcommand/server-acl-init/command.go
```diff
@@ -73,8 +81,10 @@ func (c *Command) init() {
...
+		"Toggle for creating a connect injector token. Only required when namespaces are enabled")
 	c.flags.BoolVar(&c.flagCreateInjectAuthMethod, "create-inject-token", false,
```
lkysow: now that we have two inject flags, what do you think about creating a new `-create-inject-auth-method` flag and deprecating this one? That will make it clearer.
```go
	c.flags.BoolVar(&c.flagCreateInjectorToken, "create-inject-namespace-token", false,
		"Toggle for creating a connect injector token. Only required when namespaces are enabled")
	c.flags.BoolVar(&c.flagCreateInjectAuthMethod, "create-inject-auth-method", false,
		"Toggle for creating a connect inject auth method")
	c.flags.BoolVar(&c.flagCreateInjectAuthMethod, "create-inject-token", false,
		"Toggle for creating a connect inject auth method. Deprecated: use -create-inject-auth-method instead.")

```

subcommand/server-acl-init/policies_test.go
```diff
@@ -0,0 +1,289 @@
+package serveraclinit
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
```
lkysow: maybe make a note that you're only testing the policy methods that have templates so people (like me just now!) aren't wondering why some of the policy methods aren't being tested

subcommand/server-acl-init/policies_test.go
```diff
@@ -0,0 +1,289 @@
...
+func TestAgentPolicy(t *testing.T) {
```
lkysow: the function you're testing is `rules`. I think `policy` is a better name but either way they should be consistent between the tests and the method.

subcommand/server-acl-init/command.go
adilyse: This is fine if folks only define one of the flags. If they define both, they may get unexpected results.

lkysow: True. You could then have a new variable and then error out if they're both set. Luckily it's unlikely because 99.9% of people are using the helm chart and would never get in that situation.

subcommand/server-acl-init/policies.go
adilyse: Whitespace is ignored in these rules. It ends up being a lot of duplication and makes the templates be overkill except for catalog sync.

subcommand/server-acl-init/policies_test.go
adilyse: Switching the two unchanging methods to constants should make this more clear.

subcommand/server-acl-init/policies_test.go
adilyse: Switched to rules.

subcommand/server-acl-init/command.go
adilyse: Leftover from before I added `-namespace-` into the flag name.

subcommand/server-acl-init/policies.go
adilyse: Updated.

subcommand/server-acl-init/command.go
```diff
@@ -720,11 +801,32 @@ func (c *Command) configureConnectInject(logger hclog.Logger, consulClient *api.
...
+
+	// Set up the auth method in the specific namespace if not mirroring
+	// If namespaces and mirroring are enabled, this is not necessary because
+	// the auth method will fall back to being created in the Consul `default`
+	// namespace automatically, as is necessary for mirroring.
+	// Note: if the config changes, an auth method will be created in the
+	// correct namespace, but the old auth method will not be removed.
+	writeOptions := api.WriteOptions{}
+	if c.flagEnableNamespaces && !c.flagEnableSyncK8SNSMirroring {
+		writeOptions.Namespace = c.flagConsulSyncDestinationNamespace
```
lkysow: If the namespace doesn't exist, we need to create it. Otherwise we get these errors:
```
2020-02-05T01:11:13.727Z [ERROR] Failure: creating auth method consul-k8s-auth-method: err="Unexpected response code: 500 (Namespace "luke" does not exist)"
```

subcommand/server-acl-init/command.go
```diff
@@ -739,11 +841,85 @@ func (c *Command) configureConnectInject(logger hclog.Logger, consulClient *api.
...
+		}
+
+		// This will only happen if there are existing policies
+		// for this auth method, but none that match the binding
+		// rule set up here in the bootstrap method.
+		if abr.ID == "" {
+			return errors.New("Unable to find a matching ACL binding rule to update")
+		}
+
+		err = c.untilSucceeds(fmt.Sprintf("updating acl binding rule for %s", authMethodTmpl.Name),
```
lkysow: This doesn't work if upgrading from oss to enterprise, we get this error:
```
2020-02-05T01:13:18.160Z [ERROR] Failure: updating acl binding rule for consul-k8s-auth-method: err="Unexpected response code: 500 (Changing an ACL binding rule's namespace is not permitted)"
```

subcommand/server-acl-init/command.go
```diff
@@ -237,8 +266,26 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 
+	// With the addition of namespaces, the ACL policies associated
```
lkysow: these comments are great

subcommand/server-acl-init/command.go
```diff
@@ -664,7 +756,9 @@ func (c *Command) configureConnectInject(logger hclog.Logger, consulClient *api.
 	if err != nil {
 		return err
 	}
-	if len(existingRules) > 0 {
+	// If Consul namespaces are enabled, someone may have changed their namespace
+	// configuration, so auth methods and binding rules should be updated accordingly
+	if len(existingRules) > 0 && !c.flagEnableNamespaces {
```
lkysow: This is actually resulting in a new binding rule being created each time if a non-default namespace is used:
```
consul acl binding-rule list -namespace luke
25c500b4-433f-78f1-9a1f-a73e23e3c49b:
   Namespace:    luke
   AuthMethod:   consul-k8s-auth-method
   Description:  Kubernetes binding rule
   BindType:     service
   BindName:     ${serviceaccount.name}
   Selector:     serviceaccount.name!=default
479039a1-cfe5-61db-6bde-45a922ee272c:
   Namespace:    luke
   AuthMethod:   consul-k8s-auth-method
   Description:  Kubernetes binding rule
   BindType:     service
   BindName:     ${serviceaccount.name}
   Selector:     serviceaccount.name!=default
8e68f7db-a25f-6c71-76b3-013c74394834:
   Namespace:    luke
   AuthMethod:   consul-k8s-auth-method
   Description:  Kubernetes binding rule
   BindType:     service
   BindName:     ${serviceaccount.name}
   Selector:     serviceaccount.name!=default
bf01cb84-8a80-d7ca-d8a6-74c39cff642a:
   Namespace:    luke
   AuthMethod:   consul-k8s-auth-method
   Description:  Kubernetes binding rule
   BindType:     service
   BindName:     ${serviceaccount.name}
   Selector:     serviceaccount.name!=default
```

This is because the call to list the rules isn't namespaced and so `existingRules` is `[]`. Then when we iterate through it to check if the rule already exists, it never does and so we create a new one.

I think the logic needs to be: if namespaces are not enabled, get the existing rules, check if they're empty, if they are then exit early. Else, continue on and when we get to the code below where the existingRules slice was previously used, instead make another call with the correct namespace.

subcommand/server-acl-init/command.go
```diff
@@ -739,11 +854,85 @@ func (c *Command) configureConnectInject(logger hclog.Logger, consulClient *api.
...
+	// Add a namespace if appropriate
+	// If namespaces and mirroring are enabled, this is not necessary because
+	// the binding rule will fall back to being created in the Consul `default`
+	// namespace automatically, as is necessary for mirroring.
+	if c.flagEnableNamespaces && !c.flagEnableInjectK8SNSMirroring {
+		abr.Namespace = c.flagConsulInjectDestinationNamespace
+	}
+
+	// If the binding rule already exists and namespaces are enabled, update it
+	if len(existingRules) > 0 && c.flagEnableNamespaces {
```
lkysow: We need to make another call here to populate `existingRules`, this time with the namespace set to `c.flagConsulInjectDestinationNamespace`

subcommand/server-acl-init/command_test.go
```diff
@@ -160,6 +160,19 @@ func TestRun_Tokens(t *testing.T) {
...
+			TokenName:          "connect-inject",
+			SecretName:         "release-name-consul-connect-inject-acl-token",
+		},
+		"connect-inject token -resource-prefix": {
+			TokenFlag:          "-create-inject-namespace-token",
+			ResourcePrefixFlag: "my-prefix",
+			TokenName:          "connect-inject",
+			SecretName:         "my-prefix-connect-inject-acl-token",
+		},
 		"enterprise-license token -release-name": {
```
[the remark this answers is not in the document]
lkysow: Nevermind, these tests are only for kube secret creation.

subcommand/server-acl-init/policies.go
```diff
@@ -0,0 +1,158 @@
...
+	return c.renderRules(dnsRulesTpl)
+}
+
+// This assumes users are using the default name for the service, i.e.
+// "mesh-gateway".
+func (c *Command) meshGatewayRules() (string, error) {
+	// Mesh gateways can only act as a proxy for services
+	// that its ACL token has access to. So, in the case of
+	// Consul namespaces, it needs access to all namespaces.
+	meshGatewayRulesTpl := `
```
lkysow: I was testing something else and realized these acls are wrong. We need:
```
service "mesh-gateway" {
   policy = "write"
}
service_prefix "" {
   policy = "read"
}
node_prefix "" {
   policy = "read"
}
agent_prefix "" {
   policy = "read"
}
```

The mesh gateway is always registered only in the default namespace, so if using namespaces we need:
```
service_prefix "gateway" {
   policy = "write"
}
namespace_prefix "" {
  service_prefix "" {
     policy = "read"
  }
  node_prefix "" {
     policy = "read"
  }
  agent_prefix "" {
     policy = "read"
  }
}
```

lkysow: NOTE: I'm not sure if agent_prefix needs to be namespaced.

subcommand/server-acl-init/connect_inject.go
```diff
@@ -12,36 +12,158 @@ import (
...
 		return err
 	}
-	// If Consul namespaces are enabled, someone may have changed their namespace
-	// configuration, so auth methods and binding rules should be updated accordingly
-	if len(existingRules) > 0 && !c.flagEnableNamespaces {
-		c.Log.Info(fmt.Sprintf("Binding rule for %s already exists", authMethodName))
-		return nil
+
+	// If the binding rule already exists, update it
+	// This
```
lkysow:
```suggestion
```

subcommand/server-acl-init/command.go
```diff
@@ -547,37 +607,57 @@ func (c *Command) setServerTokens(logger hclog.Logger, consulClient *api.Client,
 	return nil
 }
 
+func (c *Command) setServerPolicy(consulClient *api.Client, logger hclog.Logger) (api.ACLPolicy, error) {
+	agentRules, err := c.agentRules()
+	if err != nil {
+		logger.Error("Error templating server agent rules", "err", err)
+		return api.ACLPolicy{}, err
```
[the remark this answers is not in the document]
adilyse: There is a whole lot of inconsistency in the whole system about where logging happens. It would be good to file an issue and take a comprehensive look at in the future.

subcommand/server-acl-init/command.go
```diff
@@ -739,11 +841,85 @@ func (c *Command) configureConnectInject(logger hclog.Logger, consulClient *api.
...
+	// With the introduction of Consul namespaces, if someone upgrades into a
+	// Consul version with namespace support or changes any of their namespace
+	// settings, the policies associated with their ACL tokens will need to be
+	// updated to be namespace aware.
+	if isPolicyExistsErr(err, policy.Name) {
+		logger.Info(fmt.Sprintf("Policy %q already exists, updating", policy.Name))
+
+		// The policy ID is required in any PolicyUpdate call, so first we need to
+		// get the existing policy to extract its ID.
+		existingPolicies, _, err := consulClient.ACL().PolicyList(&api.QueryOptions{})
```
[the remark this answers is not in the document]
adilyse: Yes, it's possible to pass nil, but the pattern established elsewhere is to explicitly pass the empty struct.

subcommand/server-acl-init/policies.go
```diff
@@ -0,0 +1,155 @@
...
+}
+service "consul-snapshot" {
+   policy = "write"
+}`
+
+const entLicenseRules = `operator = "write"`
+
+func (c *Command) agentRules() (string, error) {
+	agentRulesTpl := `
+  node_prefix "" {
```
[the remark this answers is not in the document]
adilyse: The different levels of indentation do not affect the functionality because it doesn't matter to the hcl parser being used (see example output below). I created the indentation so that the OSS use case would be consistent. The UI seems to treat it differently, but there isn't a good way to have the indentation perfect for both namespaces and OSS.

ACL parsing with mixed whitespace:

subcommand/server-acl-init/policies.go
```diff
@@ -0,0 +1,155 @@
...
+`
+
+	return c.renderRules(meshGatewayRulesTpl)
+}
+
+func (c *Command) syncRules() (string, error) {
+	syncRulesTpl := `
+  node "k8s-sync" {
+    policy = "write"
+  }
```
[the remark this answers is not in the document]
adilyse: This is a UI oddity. See the other indentation comment for further explanation.

subcommand/server-acl-init/policies.go
```diff
@@ -0,0 +1,155 @@
...
+{{- if .EnableSyncK8SNSMirroring }}
+namespace_prefix "{{ .SyncK8SNSMirroringPrefix }}" {
+{{- else }}
+namespace "{{ .ConsulSyncDestinationNamespace }}" {
+{{- end }}
+{{- end }}
+  node_prefix "" {
+    policy = "read"
+  }
+  service_prefix "" {
```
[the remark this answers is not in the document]
adilyse: Reduced the permissions to only allow service write on the default namespace.

subcommand/server-acl-init/command.go
```diff
@@ -477,21 +550,8 @@ func (c *Command) bootstrapServers(logger hclog.Logger, bootTokenSecretName, sch
 // and then provides the token to the server.
 func (c *Command) setServerTokens(logger hclog.Logger, consulClient *api.Client,
 	serverPods []podAddr, bootstrapToken, scheme string) error {
-	// Create agent policy.
-	agentPolicy := api.ACLPolicy{
-		Name:        "agent-token",
```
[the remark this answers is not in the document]
adilyse: They share the same underlying rules. This allows different naming and description for client vs. server tokens in case their underlying rules end up changing in the future.

subcommand/server-acl-init/command.go
adilyse: Updated.

subcommand/server-acl-init/command.go
adilyse: Fixed.

subcommand/server-acl-init/command.go
adilyse: Fixed, as well as the issue that a user's config change to binding rules would never update the rule.

subcommand/server-acl-init/policies.go
adilyse: It does not. Updated the policy definition.

subcommand/server-acl-init/command.go
adilyse: Updated the logic for the auth method and binding rule checking.

subcommand/server-acl-init/createOrUpdate.go
```diff
@@ -0,0 +1,143 @@
+package serveraclinit
```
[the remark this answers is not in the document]
adilyse: Updated.

subcommand/server-acl-init/connect_inject.go
```diff
@@ -0,0 +1,182 @@
...
+	// If namespaces and mirroring are enabled, this is not necessary because
+	// the auth method will fall back to being created in the Consul `default`
+	// namespace automatically, as is necessary for mirroring.
+	// Note: if the config changes, an auth method will be created in the
+	// correct namespace, but the old auth method will not be removed.
+	writeOptions := api.WriteOptions{}
+	if c.flagEnableNamespaces && !c.flagEnableInjectK8SNSMirroring {
+		writeOptions.Namespace = c.flagConsulInjectDestinationNamespace
+
+		if c.flagConsulInjectDestinationNamespace != "default" {
```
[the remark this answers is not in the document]
adilyse: If this showed up a number of places, I would be happy to. However, this is the only place where it is used. It doesn't justify the switch, especially since we don't do the same for the k8s namespace. It is also highly unlikely to ever change, either in Consul or k8s.

subcommand/server-acl-init/connect_inject.go
adilyse: Finished my thought.

subcommand/server-acl-init/createOrUpdate.go
```diff
@@ -0,0 +1,143 @@
...
+			Name:        ns,
+			Description: "Auto-generated by the ACL bootstrapping process",
+			Meta:        map[string]string{"external-source": "kubernetes"},
+		}
+
+		_, _, err = consulClient.Namespaces().Create(&consulNamespace, nil)
+		if err != nil {
+			return err
+		}
+		c.Log.Info("creating consul namespace", "name", consulNamespace.Name)
```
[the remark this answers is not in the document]
adilyse: Updated the comment.

subcommand/server-acl-init/connect_inject.go
lkysow: In another commit?

lkysow: Ahh I see it in the other branch.

PR #188, opened 2020-01-11
Update catalog sync to support namespaces

adilyse:
This allows the catalog sync process to support Consul namespaces,
an Enterprise feature. It supports no namespaces (OSS), syncing k8s
services into a single Consul namespace and mirroring k8s namespaces
in Consul.

It also updates the settings for which k8s namespaces to sync. These
are now based on allow and deny lists, rather than the two previous
options of (1) a single k8s namespace, or (2) all k8s namespaces
except `kube-system`. This change is backwards compatible, however
if a user upgrades consul-k8s without upgrading the Helm chart as well,
there will be a slight difference in behavior for (2) in that it won't
automatically exclude `kube-system` on its own.

Still to do:
- [ ] Test everything
- [ ] Add namespace specific tests


lkysow:
Played around with this. I tested:
- allow/deny ns's
- `consulNamespaceName: default`
- `consulNamespaceName: other`
- `mirrorK8s`
- `mirrorPrefix`
- adding a ns to the deny list deletes svcs
- remove a ns from the deny list adds the svcs back
- changing namespace to other after sync has started
- deleting a namespace that's in use
- mirrorK8s create new ns
- mirrork8s delete ns

Some bugs I found:

1. If I start the sync-catalog with consulNamespaceName: default and then I do a helm upgrade with consulNamespaceName: other,
  it doesn't delete the old services from the namespace: default. It does register the old services in namespace other (after 30s).

    Logs
    ```
    2020-01-20T22:56:33.245Z [INFO]  K8s namespace syncing configuration: k8s namespaces allowed to be synced=Set{*} k8s namespaces denied from syncing="Set{kube-system, kube-public}"
    2020-01-20T22:56:33.247Z [INFO]  to-consul/source: starting runner for endpoints
    2020-01-20T22:56:33.251Z [INFO]  to-k8s/sink: starting runner for syncing
    Listening on ":8080"...
    2020-01-20T22:56:33.254Z [INFO]  to-k8s/source: received services from Consul: count=1
    2020-01-20T22:56:33.347Z [INFO]  to-consul/controller: initial cache sync complete
    2020-01-20T22:56:33.348Z [INFO]  to-k8s/controller: initial cache sync complete
    2020-01-20T22:56:33.348Z [INFO]  to-k8s/sink: upsert: key=jan20/consul
    2020-01-20T22:56:33.348Z [INFO]  to-k8s/sink: upsert: key=jan20/consul-dns
    2020-01-20T22:56:33.348Z [INFO]  to-k8s/sink: upsert: key=jan20/consul-server
    2020-01-20T22:56:33.348Z [INFO]  to-k8s/sink: upsert: key=jan20/consul-ui
    2020-01-20T22:56:33.352Z [INFO]  to-consul/source.controller/endpoints: initial cache sync complete
    2020-01-20T22:56:33.352Z [INFO]  to-consul/source: upsert: key=jan20/consul-dns
    2020-01-20T22:56:33.355Z [INFO]  to-consul/source: upsert: key=jan20/consul-server
    2020-01-20T22:56:33.358Z [INFO]  to-consul/source: upsert: key=default/kubernetes
    2020-01-20T22:56:33.358Z [INFO]  to-consul/source: upsert endpoint: key=jan20/consul-dns
    2020-01-20T22:56:33.362Z [INFO]  to-consul/source: upsert: key=jan20/consul-ui
    2020-01-20T22:56:33.362Z [INFO]  to-consul/source: upsert endpoint: key=jan20/consul-server
    2020-01-20T22:56:33.362Z [INFO]  to-consul/source: upsert endpoint: key=default/kubernetes
    2020-01-20T22:56:33.362Z [INFO]  to-consul/source: upsert endpoint: key=jan20/consul-ui
    2020-01-20T22:57:03.247Z [INFO]  to-consul/sink: registering services
    2020-01-20T22:57:03.247Z [INFO]  to-consul/sink: starting service watcher: service-name=kubernetes-default service-consul-namespace=other
    2020-01-20T22:57:03.247Z [INFO]  to-consul/sink: starting service watcher: service-name=consul-server-jan20 service-consul-namespace=other
    2020-01-20T22:57:03.248Z [INFO]  to-consul/sink: starting service watcher: service-name=consul-dns-jan20 service-consul-namespace=other
    2020-01-20T22:57:03.248Z [INFO]  to-consul/sink: starting service watcher: service-name=consul-ui-jan20 service-consul-namespace=other
    2020-01-20T22:57:33.440Z [INFO]  to-consul/sink: registering services
    2020-01-20T22:57:34.750Z [INFO]  to-k8s/source: received services from Consul: count=1
    2020-01-20T22:58:03.600Z [INFO]  to-consul/sink: registering services
    ```

1. When deleting a kube namespace
    ```
    2020-01-20T23:22:23.008Z [INFO]  to-consul/source: delete endpoint: key=test/my-nginx
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x1672aef]

    goroutine 28 [running]:
    github.com/hashicorp/consul-k8s/catalog/to-consul.(*ConsulSyncer).watchReapableServices(0xc0000b2280, 0x1e44580, 0xc000086900)
      /Users/lkysow/code/hashicorp/consul-k8s/catalog/to-consul/syncer.go:206 +0x79f
    created by github.com/hashicorp/consul-k8s/catalog/to-consul.(*ConsulSyncer).Run
      /Users/lkysow/code/hashicorp/consul-k8s/catalog/to-consul/syncer.go:129 +0x9d
    ```
    The Pod restarts fine but the namespace and service in consul never get deleted.


1. Setting `mirroringPrefix` works but it doesn't delete the old services in the non-prefixed namespaces.

1. Adding a new namespace to `k8sDenyNamespace` doesn't delete services in that namespace. Probably okay.
1. Other thoughts: Would be nice to print the full config at the beginning. Right now it says "K8s namespace syncing configuration:" but doesn't
  say if mirrorK8S is true or the mirroring prefix or the consul dst namespace.

catalog/to-consul/syncer.go
```diff
@@ -169,31 +194,31 @@ func (s *ConsulSyncer) watchReapableServices(ctx context.Context) {
 			return
 		}
 
-		// Update our blocking index
-		opts.WaitIndex = meta.LastIndex
```
lkysow: So I *think* understand what's happening here. Previously, this function (after the first run through) would block when it called `s.Client.Catalog().Services(&opts)` and wait for updates due to this WaitIndex here. Now that you've broken up that call into one per namespace, we can't do the blocking because it's different per namespace.

Is there not a way to make the catalog services call for all namespaces? Then we could keep the previous blocking behaviour.

If not, can you explain why this change in behaviour is okay.

catalog/to-consul/syncer.go
```diff
@@ -169,31 +194,31 @@ func (s *ConsulSyncer) watchReapableServices(ctx context.Context) {
...
+			for name, tags := range services {
+				for _, tag := range tags {
+					if tag == s.ConsulK8STag {
+						// We only care if we don't know about this service at all.
+						if s.serviceNames[ns].Contains(name) {
+							continue
+						}
+
+						s.Log.Info("invalid service found, scheduling for delete",
+							"service-name", name, "service-consul-namespace", ns)
```
lkysow:
```suggestion
							"name", name, "consul-ns", ns)
```

That the name is of the service is implied since it says "invalid service found". Probably best to keep these short if possible so logs are more concise

catalog/to-consul/syncer.go
```diff
@@ -218,18 +243,26 @@ func (s *ConsulSyncer) watchService(ctx context.Context, name string) {
 		case <-time.After(s.SyncPeriod):
 		}
 
+		// Set up query options
+		queryOpts := &api.QueryOptions{
+			AllowStale: true,
+		}
+		if s.EnableNamespaces {
+			// Sets the Consul namespace to query the catalog
+			queryOpts.Namespace = namespace
```
lkysow: again, what happens if this is set to "default" when there are no namespaces enabled?

catalog/to-consul/syncer.go
```diff
@@ -218,18 +243,26 @@ func (s *ConsulSyncer) watchService(ctx context.Context, name string) {
...
-			services, _, err = s.Client.Catalog().Service(name, s.ConsulK8STag, &api.QueryOptions{
-				AllowStale: true,
-			})
+			services, _, err = s.Client.Catalog().Service(name, s.ConsulK8STag, queryOpts)
 			return err
 		}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
 		if err != nil {
 			s.Log.Warn("error querying service, will retry",
 				"service-name", name,
+				"service-namespace", namespace, // will be "" if namespaces aren't enabled
```
lkysow: Can we use `default` in every case and remove this special-case?

catalog/to-consul/syncer.go
```diff
@@ -238,27 +271,14 @@ func (s *ConsulSyncer) watchService(ctx context.Context, name string) {
 		s.lock.Lock()
 
 		for _, svc := range services {
-			// If we have a namespace set and the key exactly matches this
-			// namespace, then we skip it.
-			if s.Namespace != "" &&
```
lkysow: can you explain why checking the ns annotations isn't necessary anymore?

catalog/to-consul/syncer.go
```diff
@@ -270,28 +290,28 @@ func (s *ConsulSyncer) watchService(ctx context.Context, name string) {
...
+	services, _, err := s.Client.Catalog().Service(name, s.ConsulK8STag, &opts)
 	if err != nil {
 		return err
 	}
 
+	// Create deregistrations for all of these
 	for _, svc := range services {
-		// If we have a namespace set and the key exactly matches this
-		// namespace, then we skip it.
-		if s.Namespace != "" &&
```
lkysow: again please explain why it's okay to delete

catalog/to-consul/syncer.go
```diff
@@ -306,31 +326,48 @@ func (s *ConsulSyncer) syncFull(ctx context.Context) {
...
 	s.Log.Info("registering services")
 
-	// Start the service watchers
-	for k, cf := range s.watchers {
-		if _, ok := s.services[k]; !ok {
-			cf()
-			delete(s.watchers, k)
+	// Update the service watchers
+	for ns, watchers := range s.watchers {
+		// If the service the watcher is watching is no longer valid,
```
lkysow: 🙏 thanks for these docs

catalog/to-consul/syncer.go
```diff
@@ -340,36 +377,54 @@ func (s *ConsulSyncer) syncFull(ctx context.Context) {
...
+				}
+			}
+
+			// Register the service
 			_, err := s.Client.Catalog().Register(r, nil)
 			if err != nil {
 				s.Log.Warn("error registering service",
 					"node-name", r.Node,
 					"service-name", r.Service.Service,
+					"service", r.Service,
```
lkysow: what does this log line look like. How does it serialize the service object? Wondering if this is usable.

catalog/to-consul/syncer.go
```diff
@@ -133,28 +150,36 @@ func (s *ConsulSyncer) Run(ctx context.Context) {
...
+		var err error
+		for ns, _ := range s.namespaces {
+			// Set up default query options
+			opts := api.QueryOptions{
+				AllowStale: true,
+				WaitTime:   15 * time.Second,
+			}
+
+			// Add the namespace to the query options if namespaces are enabled
+			if s.EnableNamespaces {
```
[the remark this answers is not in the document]
adilyse: OSS has no knowledge of namespaces and any requests with namespaces attached will fail with a message to that effect. So unfortunately no way to remove this and still have OSS function properly.

catalog/to-consul/syncer.go
adilyse: Answered above, the call fails.

catalog/to-consul/syncer.go
adilyse: Unfortunately not, as mentioned elsewhere.

adilyse:
The current approach that makes a call to Consul per namespace does not scale well to watching everything, so it would be inefficient to watch the world to remove old items when configuration changes. However, Matt is exposing a cross-namespace endpoint that will keep the Consul query logic for sync the same with a different call. Once that is available, I will update to make use of it which will solve most of the concerns with old services being left registered.

adilyse:
Panic fixed. Removing services after config changes is waiting on [Consul PR 7115](https://github.com/hashicorp/consul/pull/7115).

lkysow:
image

When I click into the node I don't see any services even though they're registered on that node. I'm using the bootstrap token. Any idea why?

adilyse:
Updated to use the cross-namespace call. Services from previous configurations are removed. As per the RFC, any no longer used namespace is not removed as we can't guarantee that there were only sync resources in it.

catalog/to-consul/syncer.go
lkysow: For my reference, it looks like:
```
2020-01-29T12:39:17.015-0700 [DEBUG] [Sync] adding service to serviceNames set: service="&{ k8s-sync-632edf7f9889 bar [k8s] map[external-k8s-ns:namespace external-source:k8s] 0  map[] {0 0} false 0 0  <nil> <nil> }" service name=bar
```

catalog/to-consul/syncer.go
lkysow: from rebecca:
> with the addition of the allow and deny lists, this functionality has to change. before, it could be assumed that a sync pod was responsible for everything or only a single namespace. that is no longer the case, since now they are responsible for some arbitrary set of namespaces. because of this, the annotation no longer provides enough information to determine whether the sync pod should delete a specific service or not

catalog/to-consul/syncer.go
lkysow: from rebecca:
> with the addition of the allow and deny lists, this functionality has to change. before, it could be assumed that a sync pod was responsible for everything or only a single namespace. that is no longer the case, since now they are responsible for some arbitrary set of namespaces. because of this, the annotation no longer provides enough information to determine whether the sync pod should delete a specific service or not

catalog/to-consul/syncer_test.go
```diff
@@ -150,44 +151,47 @@ func TestConsulSyncer_reapServiceInstance(t *testing.T) {
...
-
-	// Sleep for a bit
-	time.Sleep(500 * time.Millisecond)
-
-	// Valid service should exist
-	services, _, err := client.Catalog().Service("baz", "", nil)
-	require.NoError(err)
-	require.Len(services, 1)
-}
+// func TestConsulSyncer_reapServiceOtherNamespace(t *testing.T) {
```
lkysow: After explaining why this doesn't work I think we should delete this test then.

Issue #99, opened 2019-05-31
Get proper clusterIP on service type: ClusterIP


daniego:
Hi,
I'm not sure if I missed something in my configuration but looks like that when a service is "ClusterIP" the sync-catalog synchronises the endpoint addresses rather than the actual clusterIP.
Is that by design or there is a way to get the cluster IP and the ports?

petro-matic:
Hi,
I'm seeing the same behavior as @daniego.
Here is an example:
```
$ kubectl describe svc staging-consul-dns -n consul
Name:              staging-consul-dns
Namespace:         consul
Labels:            app=consul
                   chart=consul-helm
                   heritage=Tiller
                   release=staging
Annotations:       <none>
Selector:          app=consul,hasDNS=true,release=staging
Type:              ClusterIP
IP:                100.69.141.243
Port:              dns-tcp  53/TCP
TargetPort:        dns-tcp/TCP
Endpoints:         100.106.229.106:8600,100.114.45.206:8600,100.114.45.208:8600 + 6 more...
Port:              dns-udp  53/UDP
TargetPort:        dns-udp/UDP
Endpoints:         100.106.229.106:8600,100.114.45.206:8600,100.114.45.208:8600 + 6 more...
Session Affinity:  None
Events:            <none>

# curl localhost:8500/v1/catalog/service/staging-consul-dns
[
    {
        "ID": "",
        "Node": "k8s-sync",
        "Address": "127.0.0.1",
        "Datacenter": "dc-staging",
        "TaggedAddresses": null,
        "NodeMeta": {
            "external-source": "kubernetes"
        },
        "ServiceKind": "",
        "ServiceID": "staging-consul-dns-4968a295cbbb",
        "ServiceName": "staging-consul-dns",
        "ServiceTags": [
            "k8s"
        ],
        "ServiceAddress": "100.116.16.124",
        "ServiceWeights": {
            "Passing": 1,
            "Warning": 1
        },
        "ServiceMeta": {
            "external-k8s-ns": "",
            "external-source": "kubernetes",
            "port-dns-tcp": "53",
            "port-dns-udp": "53"
        },
        "ServicePort": 53,
        "ServiceEnableTagOverride": false,
        "ServiceProxyDestination": "",
        "ServiceProxy": {},
        "ServiceConnect": {},
        "CreateIndex": 17130,
        "ModifyIndex": 17130
    },
    {
        "ID": "",
        "Node": "k8s-sync",
[...]

PR #192, opened 2020-01-17
Allow ClusterIP Services to sync as Endpoints or as ClusterIP

praymann:
Fixes/allows for the use cases outlined in https://github.com/hashicorp/consul-k8s/issues/99.

A new configuration option is added, which details to true, that defines whether or not ClusterIP services are sync'd as Endpoints (the Pod addresses). An annotation is also allowed to override the configuration option on a per Service basis.

Should this option or annotation be false, then the service is sync'd as a single instance with the address of the actual in-cluster ClusterIP.

The only open question is if Headless Services always be sync'd as Endpoints given their lack of an in-cluster ClusterIP. Right now if the option or annotation is false for a Headless Service it is just not sync'd at all.

### DIFF
--- a/catalog/to-consul/annotation.go
+++ b/catalog/to-consul/annotation.go
@@ -15,6 +15,12 @@ const (
 	// service or an integer value.
 	annotationServicePort = "consul.hashicorp.com/service-port"
 
+	// annotationServiceAsEndpoints is valid only on ClusterIP services.
+	// It specifics whether or not to sync the service with each individual endpoint or
+	// as the singular IP of the service. If this isn't set then the
+	// default based on the syncer configuration is chosen.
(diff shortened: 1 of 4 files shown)


daniego:
There is any ETA for this to be merged and released?

ltagliamonte-dd:
I'm running multiple k8s clusters on a flat network (using AWS cni) and i would love to have pods IP in Consul rather than endpoints that are valid only within the same cluster.

aackerman:
I would also love to have this feature, we're working on a migration from running no containers on EC2 to containers and AWS EKS. This feature would help us a lot to be able to test our existing haproxy load balancers sending traffic to both EC2 instance backends and pod backends.

@praymann it looks like there are conflicts would you have time to clean that up?

PR #193, opened 2020-01-30
WIP: Namespaces tests

lkysow:
This PR switches from using `agent.TestAgent` which is imported from the `consul/api` package to using `testutil.NewTestServer` which is imported from `consul/sdk/testutil`. Under the hood, the testutil server uses the `consul` binary in `PATH`. This is unlike the `TestAgent` which was actually running the core Consul code from the `api` package.

This switch will allow us to test enterprise feature like namespaces. We add a build tag to our enterprise tests so that they don't run by default. We add a new build step that uses the Consul enterprise docker image and runs the tests with that tag. Since the `consul` binary in the path of the enterprise docker image is enterprise, the tests will be testing against consul enterprise.

The change to the test code looks like:
```diff
-	a := agent.NewTestAgent(t, t.Name(), ``)
-	defer a.Shutdown()
-	testrpc.WaitForTestAgent(t, a.RPC, "dc1")
-	client := a.Client()

+	svr, err := testutil.NewTestServerT(t)
+	require.NoError(err)
+	defer svr.Stop()
+	client, err := api.NewClient(&api.Config{
+		Address:    svr.HTTPAddr,
+	})
+	require.NoError(err)
```


lkysow:
I'm closing this because it was just created as a guide. We're using this technique in real tests now.

PR #196, opened 2020-02-04
Update the lifecycle sidecar to shell out to Consul

adilyse:
For namespaces, the lifecycle sidecar needs access to Enterprise
features to be able to register and re-register services and proxies
correctly. This switches it to directly call the consul cli with the
consul version specified in the connect injector. If a user is running
enterprise, this is the enterprise version which allows namespaces.

Note: Several tests were deleted because their approach no longer
works with this registration strategy. They will need to be rewritten
in the future.

### DIFF
--- a/subcommand/lifecycle-sidecar/command.go
+++ b/subcommand/lifecycle-sidecar/command.go
@@ -4,26 +4,33 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"github.com/hashicorp/consul/api"
-	"github.com/hashicorp/consul/command/flags"
-	"github.com/hashicorp/consul/command/services"
-	"github.com/hashicorp/go-hclog"
-	"github.com/mitchellh/cli"
-	"github.com/prometheus/common/log"
 	"os"
+	"os/exec"
 	"os/signal"
 	"sync"
 	"time"
+
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/command/flags"
+	"github.com/hashicorp/go-hclog"
+	"github.com/mitchellh/cli"
+	"github.com/prometheus/common/log"
 )
+
+type lifecycleCommandData struct {
+	AuthMethod      string
+	SyncPeriodInSec int
+}
 
 type Command struct {
 	UI cli.Ui
 
-	http              *flags.HTTPFlags
-	flagServiceConfig string
-	flagSyncPeriod    string
-	flagSet           *flag.FlagSet
-	flagLogLevel      string
+	http               *flags.HTTPFlags
+	flagServiceConfig  string
+	flagConsulLocation string
+	flagSyncPeriod     string
+	flagSet            *flag.FlagSet
+	flagLogLevel       string
 
 	once         sync.Once
 	help         string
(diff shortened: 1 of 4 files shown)


subcommand/lifecycle-sidecar/command.go
```diff
@@ -56,16 +64,24 @@ func (c *Command) Run(args []string) int {
...
 	})
 
+	// Log initial configuration
+	logger.Info("Command configuration", "service-config", c.flagServiceConfig,
+		"consul-location", c.flagConsulLocation,
+		"sync-period", syncPeriod,
+		"log-level", logLevel)
+
 	// Set up Consul client (may already exist in tests).
 	if c.consulClient == nil {
```
ishustava: Do we still need the consul client now that we're shelling out?

subcommand/lifecycle-sidecar/command.go
```diff
@@ -100,13 +103,12 @@ func (c *Command) Run(args []string) int {
...
 	// The loop will only exit when the Pod is shut down and we receive a SIGINT.
 	for {
-		for _, svc := range svcs {
-			err := c.consulClient.Agent().ServiceRegister(svc)
-			if err != nil {
-				logger.Error("failed to sync service", "id", svc.ID, "err", err)
-			} else {
-				logger.Info("successfully synced service", "id", svc.ID)
-			}
+		cmd := exec.Command(c.flagConsulLocation, "services", "register", c.flagServiceConfig)
```
ishustava: What about the ACL token? It'd be great if we could change the `lifecycle_sidecar.go` to use an env variable instead of the flag. That way, this command will just work™.

subcommand/lifecycle-sidecar/command.go
```diff
@@ -100,13 +103,12 @@ func (c *Command) Run(args []string) int {
...
+		err := cmd.Run()
```
ishustava: We should also capture stdout and strerr here and include it in the logs. Otherwise, if command fails, the error message will just be something like "exit code 1"
```suggestion
		output, err := cmd.CombinedOutput()
```

subcommand/lifecycle-sidecar/command.go
```diff
@@ -34,6 +41,7 @@ type Command struct {
 func (c *Command) init() {
 	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
 	c.flagSet.StringVar(&c.flagServiceConfig, "service-config", "", "Path to the service config file")
+	c.flagSet.StringVar(&c.flagConsulLocation, "consul-location", "", "Path to a consul binary")
```
lkysow: I think this should be `-consul-binary` to match the flag from `consul connect envoy`:
```
  -envoy-binary=<string>
     The full path to the envoy binary to run. By default will just
     search $PATH. Ignored if -bootstrap is used.
```

subcommand/lifecycle-sidecar/command.go
adilyse: Hmm, I didn't remove anything related to the ACL token. Looks like it might have been missing from before.

An environment variable would be cleaner for the lifecycle sidecar, but the connect injector deals exclusively in a token file, so it's an odd switch to make.

ishustava: Yeah, this was handled for us for free by the consul client because we instantiated it from the HTTP flags. We could set the `CONSUL_HTTP_TOKEN_FILE` env variable pointing to the file location instead.

adilyse: It would be easy to set the `CONSUL_HTTP_TOKEN_FILE` env variable in the lifecycle sidecar definition in the connect injector, but that means that the lifecycle sidecar process itself will not be able to independently handle cases where ACL tokens are enabled. Are we willing to make that trade-off? All of the other design decisions in the lifecycle sidecar process seem to indicate that it should be able to be run by itself.

subcommand/lifecycle-sidecar/command.go
adilyse: Removed.

subcommand/lifecycle-sidecar/command.go
adilyse: Updated.

subcommand/lifecycle-sidecar/command.go
adilyse: Renamed.

PR #198, opened 2020-02-05
Fixes to the lifecycle sidecar cmd, plus tests

ishustava:
This PR makes the following changes:
- Previously, we were calling `consul services register service.hcl -token-file ...`, but it needed to be `consul services register -token-file ... service.hcl`
- Additionally, we weren't passing any other HTTP flags to the underlying Consul command
- Adds tests back in
- Changes CI test job to download the Consul binary

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ executors:
       - image: circleci/golang:1.13
     environment:
       - TEST_RESULTS: /tmp/test-results # path to where test results are saved
+      - CONSUL_VERSION: 1.6.3 # this can be OSS or enterprise, e.g., 1.7.0+ent-beta4
 
 jobs:
   go-fmt-and-vet:
@@ -54,6 +55,10 @@ jobs:
 
       # run go tests with gotestsum
       - run: |
+          # download and install the consul binary
+          wget https://releases.hashicorp.com/consul/"${CONSUL_VERSION}"/consul_"${CONSUL_VERSION}"_linux_amd64.zip && \
+               unzip consul_"${CONSUL_VERSION}"_linux_amd64.zip -d /home/circleci/bin &&
+               rm consul_"${CONSUL_VERSION}"_linux_amd64.zip
           PACKAGE_NAMES=$(go list ./...)
           gotestsum --junitfile $TEST_RESULTS/gotestsum-report.xml -- -p 4 $PACKAGE_NAMES
 
--- a/subcommand/lifecycle-sidecar/command.go
+++ b/subcommand/lifecycle-sidecar/command.go
@@ -10,7 +10,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/command/flags"
 	"github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/cli"
@@ -23,21 +22,20 @@ type Command struct {
 	http              *flags.HTTPFlags
 	flagServiceConfig string
 	flagConsulBinary  string
-	flagSyncPeriod    string
+	flagSyncPeriod    time.Duration
 	flagSet           *flag.FlagSet
 	flagLogLevel      string
 
-	once         sync.Once
-	help         string
-	consulClient *api.Client
-	sigCh        chan os.Signal
+	once  sync.Once
+	help  string
+	sigCh chan os.Signal
 }
 
 func (c *Command) init() {
 	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
 	c.flagSet.StringVar(&c.flagServiceConfig, "service-config", "", "Path to the service config file")
-	c.flagSet.StringVar(&c.flagConsulBinary, "consul-binary", "", "Path to a consul binary")
-	c.flagSet.StringVar(&c.flagSyncPeriod, "sync-period", "10s", "Time between syncing the service registration. Defaults to 10s.")
+	c.flagSet.StringVar(&c.flagConsulBinary, "consul-binary", "consul", "Path to a consul binary")
+	c.flagSet.DurationVar(&c.flagSyncPeriod, "sync-period", 10*time.Second, "Time between syncing the service registration. Defaults to 10s.")
 	c.flagSet.StringVar(&c.flagLogLevel, "log-level", "info",
 		"Log verbosity level. Supported values (in order of detail) are \"trace\", "+
 			"\"debug\", \"info\", \"warn\", and \"error\". Defaults to info.")
@@ -80,6 +78,10 @@ func (c *Command) Run(args []string) int {
 	// Set up channel for graceful SIGINT shutdown.
 	signal.Notify(c.sigCh, os.Interrupt)
 
+	consulCommand := []string{"services", "register"}
+	consulCommand = append(consulCommand, c.parseConsulFlags()...)
+	consulCommand = append(consulCommand, c.flagServiceConfig)
+
 	// The main work loop. We continually re-register our service every
 	// syncPeriod. Consul is smart enough to know when the service hasn't changed
 	// and so won't update any indices. This means we won't be causing a lot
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -54,6 +55,10 @@ jobs:
 
       # run go tests with gotestsum
       - run: |
+          # download and install the consul binary
+          wget https://releases.hashicorp.com/consul/"${CONSUL_VERSION}"/consul_"${CONSUL_VERSION}"_linux_amd64.zip && \
+               unzip consul_"${CONSUL_VERSION}"_linux_amd64.zip -d /home/circleci/bin &&
+               rm consul_"${CONSUL_VERSION}"_linux_amd64.zip
```
ishustava: This is a bit hacky, but I didn't want to break the CI.

subcommand/lifecycle-sidecar/command.go
```diff
@@ -27,16 +26,15 @@ type Command struct {
...
+	once  sync.Once
+	help  string
+	sigCh chan os.Signal
 }
 
 func (c *Command) init() {
 	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
 	c.flagSet.StringVar(&c.flagServiceConfig, "service-config", "", "Path to the service config file")
-	c.flagSet.StringVar(&c.flagConsulBinary, "consul-binary", "", "Path to a consul binary")
+	c.flagSet.StringVar(&c.flagConsulBinary, "consul-binary", "consul", "Path to a consul binary")
```
ishustava: Changing the default here because this seems more intuitive.

.circleci/config.yml
lkysow: ❤️ it!

subcommand/lifecycle-sidecar/command.go
```diff
@@ -89,19 +91,14 @@ func (c *Command) Run(args []string) int {
 	// The loop will only exit when the Pod is shut down and we receive a SIGINT.
 	for {
 		var cmd *exec.Cmd
-		if c.http.TokenFile() == "" {
-			cmd = exec.Command(c.flagConsulBinary, "services", "register", c.flagServiceConfig)
-		} else {
-			cmd = exec.Command(c.flagConsulBinary, "services", "register", c.flagServiceConfig,
-				fmt.Sprintf("-token-file=%s", c.http.TokenFile()))
-		}
+		cmd = exec.Command(c.flagConsulBinary, consulCommand...)
```
adilyse: This could be defined with `:=` and the var definition removed now that it's not needed outside of an if statement.

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -1,11 +1,16 @@
 package subcommand
 
 import (
+	"fmt"
+	"github.com/hashicorp/consul/agent"
```
adilyse: Imports are usually grouped as standard library imports, line break, other imports. It would be good to do the same here.

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -84,6 +89,115 @@ func TestRun_ConsulBinaryMissing(t *testing.T) {
...
+	t.Parallel()
+	tmpDir, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	defer func() { os.RemoveAll(tmpDir) }()
+
+	configFile := filepath.Join(tmpDir, "svc.hcl")
+	err = ioutil.WriteFile(configFile, []byte(servicesRegistration), 0600)
+	require.NoError(t, err)
+
+	a := agent.NewTestAgent(t, t.Name(), `primary_datacenter = "dc1"`)
```
adilyse: Using the TestAgent means we won't be able to add any namespace tests to the sidecar, which it would be good to do. What about using the Consul test server in the sdk instead?

subcommand/lifecycle-sidecar/command.go
```diff
@@ -27,16 +26,15 @@ type Command struct {
 	flagSet           *flag.FlagSet
```
adilyse: Since we're updating things, it would be good to switch the `flagSyncPeriod` to a `flags.DurationValue` instead. The sync process has an [example](https://github.com/hashicorp/consul-k8s/blob/master/subcommand/sync-catalog/command.go#L45).

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -84,6 +89,115 @@ func TestRun_ConsulBinaryMissing(t *testing.T) {
 	require.Contains(t, ui.ErrorWriter.String(), "-consul-binary \"/not/a/valid/path\" not found")
 }
 
+// Test that we register the services.
+func TestRun_ServicesRegistration(t *testing.T) {
+	t.Parallel()
+	tmpDir, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	defer func() { os.RemoveAll(tmpDir) }()
```
adilyse: Does this need the `func` wrapper?

subcommand/lifecycle-sidecar/command.go
```diff
@@ -149,6 +146,19 @@ func (c *Command) validateFlags() (syncPeriod time.Duration, logLevel hclog.Leve
 	return
 }
 
+// parseConsulFlags creates Consul client command flags
+// from command's HTTP flags and returns them as an array of strings.
+func (c *Command) parseConsulFlags() []string {
+	var consulCommandFlags []string
+	c.http.ClientFlags().VisitAll(func(f *flag.Flag) {
+		if f.Value.String() != "" {
+			consulCommandFlags = append(consulCommandFlags, fmt.Sprintf("-%s", f.Name))
```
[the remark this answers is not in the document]
ishustava: Sounds good to me. I'll update it.

subcommand/lifecycle-sidecar/command.go
ishustava: Good catch!

subcommand/lifecycle-sidecar/command_test.go
ishustava: I don't think it does, good catch. I'll remove it.

subcommand/lifecycle-sidecar/command.go
ishustava: Yeah, good call. I've updated it to use `time.Duration` type and parse it with `DurationVar` function instead. It made it easier to handle the default, plus I didn't see the need for a pointer here. Let me know what you think.

subcommand/lifecycle-sidecar/command_test.go
ishustava: Yep, good point. I think I assumed that for namespaces we'd need to create a new test file anyways and would deal with it there. I've updated it.

ishustava:
@adilyse thanks so much for the review! I've made the changes, let me know you think.

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -27,14 +42,6 @@ func TestRun_FlagValidation(t *testing.T) {
 			},
 			ExpErr: "-consul-binary must be set",
 		},
-		{
```
lkysow: does this not error anymore?

ishustava: Yeah, with Rebecca's suggestion to use time.Duration, the flags library handles it for us. Since we don't have any code for validation and parsing of this value anymore, we don't need this test.

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -61,27 +68,164 @@ func TestRun_ServiceConfigFileMissing(t *testing.T) {
...
+	require.Contains(t, ui.ErrorWriter.String(), "unknown log level: foo")
+}
+
+// Test that we register the services.
+func TestRun_ServicesRegistration(t *testing.T) {
+	t.Parallel()
+	tmpDir, configFile := createServicesTmpFile(t)
+	defer os.RemoveAll(tmpDir)
+
+	a, err := testutil.NewTestServerConfigT(t, func(c *testutil.TestServerConfig) {
```
adilyse: For my own education, I'd like to know the trade-offs between using `NewTestServerConfigT` vs. the other test server setups in the sdk. Also, is the datacenter necessary? I would have thought the defaults would be sufficient.

ishustava: There are 2 functions for setting up the server that we want to use:
- `NewTestServerT` - to use when you want to use just the default values
- `NewTestServerConfigT` - to use when you need to tweak some config values; it takes a callback that allows you to overwrite config values as I did in this test.

There is also the `NewTestServerConfig`, but I don't think we need to use it because it doesn't take `t`.

Yeah, you're right. I don't think we need the primary dc setting in this case.

PR #199, opened 2020-02-06
Add tests for namespaces and acl bootstrapping

lkysow:
Add tests for the namespace changes of server-acl-init.

The tests that will fail are due to the binding rule not being able to have its namespace updated.

Tests:
* binding rule
  * single dest namespace
  * mirroring
* auth method
  * single dest namespace
  * mirroring
* updates (i.e. when the command is re-run it updates the binding rule/auth method)
  * no ns => ns mirroring
  * no ns => single dest namespace <= FAILS
  * ns mirroring => single dest namespace <= FAILS
  * single dest namespace => ns mirroing

### DIFF
--- a/subcommand/server-acl-init/command_ent_test.go
+++ b/subcommand/server-acl-init/command_ent_test.go
@@ -1,0 +1,533 @@
+// +build enterprise
+
+package serveraclinit
+
+import (
+	"testing"
+
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/sdk/testutil"
+	"github.com/mitchellh/cli"
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+// Test the auth method and acl binding rule created when namespaces are enabled
+// and there's a single consul destination namespace.
+func TestRun_ConnectInject_SingleDestinationNamespace(t *testing.T) {
+	t.Parallel()
+
+	consulDestNamespaces := []string{"default", "destination"}
+	for _, consulDestNamespace := range consulDestNamespaces {
+		t.Run(consulDestNamespace, func(tt *testing.T) {
+			k8s, testAgent := completeEnterpriseSetup(tt, resourcePrefix, ns)
+			defer testAgent.Stop()
+			setUpK8sServiceAccount(tt, k8s)
+			require := require.New(tt)
+
+			ui := cli.NewMockUi()
+			cmd := Command{
+				UI:        ui,
+				clientset: k8s,
+			}
+			cmd.init()
+			args := []string{
+				"-server-label-selector=component=server,app=consul,release=" + releaseName,
+				"-resource-prefix=" + resourcePrefix,
+				"-k8s-namespace=" + ns,
+				"-expected-replicas=1",
+				"-create-inject-auth-method",
+				"-enable-namespaces",
+				"-consul-inject-destination-namespace", consulDestNamespace,
+				"-acl-binding-rule-selector=serviceaccount.name!=default",
+			}
+
+			responseCode := cmd.Run(args)
+			require.Equal(0, responseCode, ui.ErrorWriter.String())
+
+			bootToken := getBootToken(t, k8s, resourcePrefix, ns)
+			consul, err := api.NewClient(&api.Config{
+				Address: testAgent.HTTPAddr,
+				Token:   bootToken,
+			})
+			require.NoError(err)
+
+			// Ensure there's only one auth method.
+			namespaceQuery := &api.QueryOptions{
+				Namespace: consulDestNamespace,
+			}
+			methods, _, err := consul.ACL().AuthMethodList(namespaceQuery)
+			require.NoError(err)
+			require.Len(methods, 1)
+
+			// Check the ACL auth method is created in the expected namespace.
+			authMethodName := releaseName + "-consul-k8s-auth-method"
+			actMethod, _, err := consul.ACL().AuthMethodRead(authMethodName, namespaceQuery)
+			require.NoError(err)
+			require.NotNil(actMethod)
+			require.Equal("kubernetes", actMethod.Type)
+			require.Equal("Kubernetes AuthMethod", actMethod.Description)
+			require.NotContains(actMethod.Config, "MapNamespaces")
+			require.NotContains(actMethod.Config, "ConsulNamespacePrefix")
+
+			// Check the binding rule is as expected.
+			rules, _, err := consul.ACL().BindingRuleList(authMethodName, namespaceQuery)
+			require.NoError(err)
+			require.Len(rules, 1)
+			actRule, _, err := consul.ACL().BindingRuleRead(rules[0].ID, namespaceQuery)
+			require.NoError(err)
+			require.NotNil(actRule)
+			require.Equal("Kubernetes binding rule", actRule.Description)
+			require.Equal(api.BindingRuleBindTypeService, actRule.BindType)
+			require.Equal("${serviceaccount.name}", actRule.BindName)
+			require.Equal("serviceaccount.name!=default", actRule.Selector)
+		})
+	}
+}
+
+// Test the auth method and acl binding rule created when namespaces are enabled
+// and we're mirroring namespaces.
+func TestRun_ConnectInject_NamespaceMirroring(t *testing.T) {
+	t.Parallel()
+
+	cases := map[string]struct {
+		MirroringPrefix string
+		ExtraFlags      []string
+	}{
+		"no prefix": {
+			MirroringPrefix: "",
+			ExtraFlags:      nil,
+		},
+		"with prefix": {
+			MirroringPrefix: "prefix-",
+			ExtraFlags:      nil,
+		},
+		"with destination namespace flag": {
+			MirroringPrefix: "",
+			// Mirroring takes precedence over this flag so it should have no
+			// effect.
+			ExtraFlags: []string{"-consul-inject-destination-namespace=dest"},
+		},
+	}
+
+	for name, c := range cases {
+		t.Run(name, func(tt *testing.T) {
+			k8s, testAgent := completeEnterpriseSetup(t, resourcePrefix, ns)
+			defer testAgent.Stop()
+			setUpK8sServiceAccount(tt, k8s)
+			require := require.New(tt)
+
+			ui := cli.NewMockUi()
+			cmd := Command{
+				UI:        ui,
+				clientset: k8s,
+			}
+			cmd.init()
+			args := []string{
+				"-server-label-selector=component=server,app=consul,release=" + releaseName,
+				"-resource-prefix=" + resourcePrefix,
+				"-k8s-namespace=" + ns,
+				"-expected-replicas=1",
+				"-create-inject-auth-method",
+				"-enable-namespaces",
+				"-enable-inject-k8s-namespace-mirroring",
+				"-inject-k8s-namespace-mirroring-prefix", c.MirroringPrefix,
+				"-acl-binding-rule-selector=serviceaccount.name!=default",
+			}
+			args = append(args, c.ExtraFlags...)
+			responseCode := cmd.Run(args)
+			require.Equal(0, responseCode, ui.ErrorWriter.String())
+
+			bootToken := getBootToken(t, k8s, resourcePrefix, ns)
+			consul, err := api.NewClient(&api.Config{
+				Address: testAgent.HTTPAddr,
+				Token:   bootToken,
+			})
+			require.NoError(err)
+
+			// Check the ACL auth method is as expected.
+			authMethodName := releaseName + "-consul-k8s-auth-method"
+			method, _, err := consul.ACL().AuthMethodRead(authMethodName, nil)
+			require.NoError(err)
+			require.NotNil(method, authMethodName+" not found")
+			require.Equal("kubernetes", method.Type)
+			require.Equal("Kubernetes AuthMethod", method.Description)
+			require.Contains(method.Config, "MapNamespaces")
+			require.Contains(method.Config, "ConsulNamespacePrefix")
+			require.Equal(true, method.Config["MapNamespaces"])
+			require.Equal(c.MirroringPrefix, method.Config["ConsulNamespacePrefix"])
+
+			// Check the binding rule is as expected.
+			rules, _, err := consul.ACL().BindingRuleList(authMethodName, nil)
+			require.NoError(err)
+			require.Len(rules, 1)
+			actRule, _, err := consul.ACL().BindingRuleRead(rules[0].ID, nil)
+			require.NoError(err)
+			require.NotNil(actRule)
+			require.Equal("Kubernetes binding rule", actRule.Description)
+			require.Equal(api.BindingRuleBindTypeService, actRule.BindType)
+			require.Equal("${serviceaccount.name}", actRule.BindName)
+			require.Equal("serviceaccount.name!=default", actRule.Selector)
+		})
+	}
+}
+
+// Test that ACL policies get updated if namespaces config changes.
+func TestRun_ACLPolicyUpdates(t *testing.T) {
+	t.Parallel()
+
+	k8sNamespaceFlags := []string{"default", "other"}
+	for _, k8sNamespaceFlag := range k8sNamespaceFlags {
+		t.Run(k8sNamespaceFlag, func(t *testing.T) {
+			k8s, testAgent := completeEnterpriseSetup(t, resourcePrefix, k8sNamespaceFlag)
+			defer testAgent.Stop()
+			require := require.New(t)
+
+			ui := cli.NewMockUi()
+			firstRunArgs := []string{
+				"-server-label-selector=component=server,app=consul,release=" + releaseName,
+				"-resource-prefix=" + resourcePrefix,
+				"-k8s-namespace", k8sNamespaceFlag,
+				"-create-client-token",
+				"-allow-dns",
+				"-create-mesh-gateway-token",
+				"-create-sync-token",
+				"-create-inject-namespace-token",
+				"-create-snapshot-agent-token",
+				"-create-enterprise-license-token",
+				"-expected-replicas=1",
+			}
+			// Our second run, we're going to update from namespaces disabled to
+			// namespaces enabled with a single destination ns.
+			secondRunArgs := append(firstRunArgs,
+				"-enable-namespaces",
+				"-consul-sync-destination-namespace=sync",
+				"-consul-inject-destination-namespace=dest")
+
+			// Run the command first to populate the policies.
+			cmd := Command{
+				UI:        ui,
+				clientset: k8s,
+			}
+			responseCode := cmd.Run(firstRunArgs)
+			require.Equal(0, responseCode, ui.ErrorWriter.String())
+
+			bootToken := getBootToken(t, k8s, resourcePrefix, k8sNamespaceFlag)
+			consul, err := api.NewClient(&api.Config{
+				Address: testAgent.HTTPAddr,
+				Token:   bootToken,
+			})
+			require.NoError(err)
+
+			// Check that the expected policies were created.
+			expectedPolicies := []string{
+				"dns-policy",
+				"client-token",
+				"catalog-sync-token",
+				"connect-inject-token",
+				"mesh-gateway-token",
+				"client-snapshot-agent-token",
+				"enterprise-license-token",
+			}
+			policies, _, err := consul.ACL().PolicyList(nil)
+			require.NoError(err)
+
+			// Collect the actual policies into a map to make it easier to assert
+			// on their existence and contents.
+			actualPolicies := make(map[string]string)
+			for _, p := range policies {
+				policy, _, err := consul.ACL().PolicyRead(p.ID, nil)
+				require.NoError(err)
+				actualPolicies[p.Name] = policy.Rules
+			}
+			for _, expected := range expectedPolicies {
+				actRules, ok := actualPolicies[expected]
+				require.True(ok, "Did not find policy %s", expected)
+				// We assert that the policy doesn't have any namespace config
+				// in it because later that's what we're using to test that it
+				// got updated.
+				require.NotContains(actRules, "namespace")
+			}
+
+			// Re-run the command with namespace flags. The policies should be updated.
+			// NOTE: We're redefining the command so that the old flag values are
+			// reset.
+			cmd = Command{
+				UI:        ui,
+				clientset: k8s,
+			}
+			responseCode = cmd.Run(secondRunArgs)
+			require.Equal(0, responseCode, ui.ErrorWriter.String())
+
+			// Check that the policies have all been updated.
+			policies, _, err = consul.ACL().PolicyList(nil)
+			require.NoError(err)
+
+			// Collect the actual policies into a map to make it easier to assert
+			// on their existence and contents.
+			actualPolicies = make(map[string]string)
+			for _, p := range policies {
+				policy, _, err := consul.ACL().PolicyRead(p.ID, nil)
+				require.NoError(err)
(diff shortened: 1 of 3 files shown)


subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+
+	bootToken := getBootToken(t, k8s, resourcePrefix)
+	consul, err := api.NewClient(&api.Config{
+		Address: testAgent.HTTPAddr,
+		Token:   bootToken,
+	})
+	require.NoError(err)
+
+	// Check the ACL auth method is created in the expected namespace.
+	authMethodName := "release-name-consul-k8s-auth-method"
```
adilyse: It would be good to use the `releaseName` variable to construct this so it works if `releaseName` changes.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+	methods, _, err := consul.ACL().AuthMethodList(&api.QueryOptions{
+		Namespace: "dest",
+	})
+	require.NoError(err)
+	found := false
+	for _, m := range methods {
```
adilyse: Alternatively, there should only be one auth method in the `dest` namespace, so you could check that the length of the list is 1 and that the single item is the correct auth method.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+		if m.Name == authMethodName {
+			found = true
+		}
+	}
+	require.True(found, fmt.Sprintf("no auth method with name %s found", authMethodName))
```
adilyse: There should be some validation about the properties of the auth method here, since they matter and can vary with namespace enabled.

Specifically, it would be good to check its Config: `MapNamespaces` and `ConsulNamespacePrefix` fields.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+	cmd.init()
+	args := []string{
+		"-server-label-selector=component=server,app=consul,release=" + releaseName,
+		"-resource-prefix=" + resourcePrefix,
+		"-k8s-namespace=" + ns,
+		"-expected-replicas=1",
+		"-create-inject-auth-method",
+		"-enable-namespaces",
+		"-enable-inject-k8s-namespace-mirroring",
+		"-inject-k8s-namespace-mirroring-prefix=prefix-",
```
adilyse: There should be another test case with an empty prefix. Also, it would be good to test that providing a `-consul-inject-destination-namespace` field gets overridden by the mirroring.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+
+	bootToken := getBootToken(t, k8s, resourcePrefix)
+	consul, err := api.NewClient(&api.Config{
+		Address: testAgent.HTTPAddr,
+		Token:   bootToken,
+	})
+	require.NoError(err)
+
+	// Check the ACL auth method is as expected.
+	authMethodName := "release-name-consul-k8s-auth-method"
```
adilyse: Same comment about using the `releaseName` variable to construct this.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+	cases := map[string]struct {
+		// Args for first run of command.
+		FirstRunArgs []string
+		// Args for second run of command.
+		SecondRunArgs []string
+		// Expected namespace for the auth method.
+		AuthMethodExpectedNS string
+		// Whether to expect the MapNamespaces and ConsulNamespacePrefix
+		// keys to be set on the config.
+		AuthMethodExpectMirroringConfig bool
```
adilyse: The mirroring and prefix checking should be separate and the prefix fields validated.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+		// Expected namespace for the binding rule.
+		BindingRuleExpectedNS string
+	}{
```
adilyse: Additional test cases should check that changing the mirroring prefix gets updated. That includes empty string to non-empty, non-empty to empty and non-empty to non-empty with different values.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+				"-server-label-selector=component=server,app=consul,release=" + releaseName,
+				"-resource-prefix=" + resourcePrefix,
+				"-k8s-namespace=" + ns,
+				"-expected-replicas=1",
+				"-create-inject-auth-method",
+			}
+
+			// First run. NOTE: we don't assert anything here since we've
+			// tested these results in other tests. What we care about here
+			// is the result after the second fun.
```
adilyse: I like fun, but I don't think that's what you meant :).
```suggestion
			// is the result after the second run.
```

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+			cmd := Command{
+				UI:        ui,
+				clientset: k8s,
+			}
+			responseCode := cmd.Run(append(defaultArgs, c.FirstRunArgs...))
+			require.Equal(0, responseCode, ui.ErrorWriter.String())
+
+			// Second run.
+			cmd = Command{
```
adilyse: Does this need to be declared again?

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+			// Now check that everything is as expected.
+			bootToken := getBootToken(t, k8s, resourcePrefix)
+			consul, err := api.NewClient(&api.Config{
+				Address: testAgent.HTTPAddr,
+				Token:   bootToken,
+			})
+			require.NoError(err)
+
+			// Check the ACL auth method is as expected.
+			authMethodName := "release-name-consul-k8s-auth-method"
```
adilyse: `releaseName` again.

subcommand/server-acl-init/command_ent_test.go
adilyse: It would also be good to have one that tests a single destination namespace that is the Consul `default` namespace.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+	t.Parallel()
+	k8s, testAgent := completeEnterpriseSetup(t, resourcePrefix)
+	defer testAgent.Stop()
+	require := require.New(t)
+
+	ui := cli.NewMockUi()
+	firstRunArgs := []string{
+		"-server-label-selector=component=server,app=consul,release=" + releaseName,
+		"-resource-prefix=" + resourcePrefix,
+		"-k8s-namespace=" + ns,
```
adilyse: Besides adding an update scenario to going to mirroring, it would also be good to test the case where the k8s namespace is not `default`.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+		"-create-client-token",
+		"-allow-dns",
+		"-create-mesh-gateway-token",
+		"-create-sync-token",
+		"-create-inject-namespace-token",
+		"-expected-replicas=1",
+		"-allow-dns",
```
adilyse: This is declared twice.
```suggestion
```

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+				require.NoError(err)
+				require.NotContains(policy.Rules, "namespace")
+				break
+			}
+		}
+		require.True(found, "Did not find "+expected)
+	}
+
+	// Re-run the command with namespace flags. The policies should be updated.
+	cmd = Command{
```
adilyse: Does this need to be defined again?

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+
+	ui := cli.NewMockUi()
+	firstRunArgs := []string{
+		"-server-label-selector=component=server,app=consul,release=" + releaseName,
+		"-resource-prefix=" + resourcePrefix,
+		"-k8s-namespace=" + ns,
+		"-create-client-token",
+		"-allow-dns",
+		"-create-mesh-gateway-token",
+		"-create-sync-token",
```
adilyse: It would be good to include the snapshot agent and ent license policy creation as well so that we can validate that they don't change with namespaces.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+		"client-token",
+		"catalog-sync-token",
+		"connect-inject-token",
+		"mesh-gateway-token",
+	}
+	policies, _, err := consul.ACL().PolicyList(nil)
+	require.NoError(err)
+	for _, expected := range expectedPolicies {
+		found := false
+		for _, p := range policies {
```
adilyse: What about doing a single loop through `policies` first to put them in a map of policy name -> policy struct? Then when looping through `expectedPolicies`, you can look them up directly.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+	"github.com/mitchellh/cli"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	"testing"
+)
+
+// Test that ACL policies get updated if namespaces config changes.
+func TestRun_ACLPolicyUpdates(t *testing.T) {
```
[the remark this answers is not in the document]
adilyse: Also, it might be good to test the `createOrUpdateACLPolicy` method directly.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
+// +build enterprise
+
+package serveraclinit
+
+import (
+	"encoding/base64"
+	"fmt"
+	"github.com/hashicorp/consul/api"
```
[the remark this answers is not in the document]
lkysow: I think we should get some ci goodness for this

subcommand/server-acl-init/command_ent_test.go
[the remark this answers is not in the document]
lkysow: > Also, it might be good to test the createOrUpdateACLPolicy method directly.

It looks like the only thing we're not testing is this:
```
		// This shouldn't happen, because we're looking for a policy
		// only after we've hit a `Policy already exists` error.
		if policy.ID == "" {
			return errors.New("Unable to find existing ACL policy")
		}
```

If that's true then not sure this warrants a separate test unless there's a case I'm missing.

subcommand/server-acl-init/command_ent_test.go
lkysow: Do you mean testing the policy updates from when the kubernetes namespace is not default or another type of update? AFAIK the k8s namespace doesn't affect the policies at all right?

subcommand/server-acl-init/command_ent_test.go
lkysow: 🍺

subcommand/server-acl-init/command_ent_test.go
lkysow: Good call, switched to:

```go
	// Collect the actual polices into a map to make it easier to assert
	// on their existence and contents.
	actualPolicies := make(map[string]string)
	for _, p := range policies {
		policy, _, err := consul.ACL().PolicyRead(p.ID, nil)
		require.NoError(err)
		actualPolicies[p.Name] = policy.Rules
	}
	for _, expected := range expectedPolicies {
		actPolicy, ok := actualPolicies[expected]
		require.True(ok, "Did not find policy %s", expected)
		// We assert that the policy doesn't have any namespace config
		// in it because later that's what we're using to test that it
		// got updated.
		require.NotContains(actPolicy, "namespace")
	}
```

subcommand/server-acl-init/command_ent_test.go
lkysow: Yes 😄found out about this the hard way. If we re-use the old command then old flags are still set. In this case it won't make a difference but if the first-run had set `-destination-consul-ns` then even though you set `-mirror-k8s` in the second run it would never take effect since the old flag would still be set.

In this case, it won't have effect but I think it's better to have it re-defined for two reasons:
1. If the test changes in the future it's very confusing to debug
1. Its a better representation of the real-world use-case. If a user upgrades with helm the command won't be re-used, it will be redefined since it's a brand new job. So we wouldn't want a side-effect in this case to cause the test to pass.

subcommand/server-acl-init/command_ent_test.go
lkysow: 🍺

subcommand/server-acl-init/command_ent_test.go
lkysow: See comment above. In this case it has to be otherwise the old flags stay set.

subcommand/server-acl-init/command_ent_test.go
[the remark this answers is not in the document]
lkysow: > It makes more sense to me to have the initial functionality tests before the update tests.

Down to move it but I'll do it after you have a second look so the changes are easier to review.

subcommand/server-acl-init/command_ent_test.go
lkysow: Done.

subcommand/server-acl-init/command_ent_test.go
lkysow: Switched to
```go
			// Check the ACL auth method is created in the expected namespace.
			authMethodName := releaseName + "-consul-k8s-auth-method"
			namespaceQuery := &api.QueryOptions{
				Namespace: consulDestNamespace,
			}
			actMethod, _, err := consul.ACL().AuthMethodRead(authMethodName, namespaceQuery)
			require.NotNil(actMethod)
			require.Equal("kubernetes", actMethod.Type)
			require.Equal("Kubernetes AuthMethod", actMethod.Description)
			require.NotContains(actMethod.Config, "MapNamespaces")
			require.NotContains(actMethod.Config, "ConsulNamespacePrefix")

			// Check the binding rule is as expected.
			rules, _, err := consul.ACL().BindingRuleList(authMethodName, namespaceQuery)
			require.NoError(err)
			require.Len(rules, 1)
			actRule, _, err := consul.ACL().BindingRuleRead(rules[0].ID, namespaceQuery)
			require.NoError(err)
			require.NotNil(actRule)
			require.Equal("Kubernetes binding rule", actRule.Description)
			require.Equal(api.BindingRuleBindTypeService, actRule.BindType)
			require.Equal("${serviceaccount.name}", actRule.BindName)
			require.Equal("serviceaccount.name!=default", actRule.Selector)

```

subcommand/server-acl-init/command_ent_test.go
lkysow: Yep, done!

subcommand/server-acl-init/command_ent_test.go
lkysow: Added cases:
```go
	cases := map[string]struct {
		MirroringPrefix string
		ExtraFlags      []string
	}{
		"no prefix": {
			MirroringPrefix: "",
			ExtraFlags:      nil,
		},
		"with prefix": {
			MirroringPrefix: "prefix-",
			ExtraFlags:      nil,
		},
		"with destination namespace flag": {
			MirroringPrefix: "",
			// Mirroring takes precedence over this flag so it should have no
			// effect.
			ExtraFlags: []string{"-consul-inject-destination-namespace=dest"},
		},
	}

```

subcommand/server-acl-init/command_ent_test.go
lkysow: Changed to
```
		// If true, we expect MapNamespaces to be set on the auth method
		// config.
		AuthMethodExpectMapNamespacesConfig bool
		// If AuthMethodExpectMapNamespacesConfig is true, we will assert
		// that the ConsulNamespacePrefix field on the auth method config
		// is set to this.
		AuthMethodExpectedNamespacePrefixConfig string

```

subcommand/server-acl-init/command_ent_test.go
lkysow: > Additional test cases should check that changing the mirroring prefix gets updated. That includes empty string to non-empty, non-empty to empty and non-empty to non-empty with different values.

Added
* mirroring ns (no prefix) => mirroring ns (no prefix)
* mirroring ns => mirroring ns (same prefix)
* mirroring ns (no prefix) => mirroring ns (prefix)
* mirroring ns (prefix) => mirroring ns (no prefix)

> It would also be good to have one that tests a single destination namespace that is the Consul default namespace.

I added cases to `TestRun_ConnectInject_SingleDestinationNamespace`: `consulDestNamespaces := []string{"default", "destination"}`. How's that?

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,436 @@
...
+			rules, _, err := consul.ACL().BindingRuleList(authMethodName, &api.QueryOptions{
+				Namespace: c.BindingRuleExpectedNS,
+			})
+			require.NoError(err)
+			require.Len(rules, 1)
+		})
+	}
+}
+
+func setupK8sServiceAccount(t *testing.T, k8s *fake.Clientset) {
```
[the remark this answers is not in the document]
lkysow: Done! Good catch. This method had to move to `command_test` though because in this file it won't get picked up by the OSS tests due to the build tag.

subcommand/server-acl-init/command_ent_test.go
adilyse: Looks good!

subcommand/server-acl-init/command_ent_test.go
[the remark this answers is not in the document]
lkysow: Yeah for sure, I did change it but it reverted. I've set my editor now so should be good in the future.

subcommand/server-acl-init/command_ent_test.go
lkysow: So we both agree that this flag has no effect on whether the policies are updated and we've already tested the templating of the policies in the rules_test.go. It sounds like you're saying the reason we should test this flag varying is because it's related to namespaces. I don't think that's a good enough reason to add an additional case here. You could make the same argument for the -allow-k8s-namespace flag. In the end I think that would complicate the test without a corresponding gain in meaningful testing coverage.

subcommand/server-acl-init/command_ent_test.go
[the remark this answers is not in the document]
lkysow: I think those cases are already covered through the tests. I ran with code coverage:
image

subcommand/server-acl-init/command_ent_test.go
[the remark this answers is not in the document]
lkysow: Okay I double checked this time, I think we're good.

subcommand/server-acl-init/command_ent_test.go
lkysow: done.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -0,0 +1,516 @@
...
+				Token:   bootToken,
+			})
+			require.NoError(err)
+
+			// Check the ACL auth method is created in the expected namespace.
+			authMethodName := releaseName + "-consul-k8s-auth-method"
+			namespaceQuery := &api.QueryOptions{
+				Namespace: consulDestNamespace,
+			}
+			actMethod, _, err := consul.ACL().AuthMethodRead(authMethodName, namespaceQuery)
```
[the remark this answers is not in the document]
lkysow: done (lines 174-178)

lkysow:
> Did you get a chance to add a test for being able to update the value of the binding rule? It's not enterprise specific, but that functionality changed as part of the namespace work so it would be good to add one. 

Sorry I totally forgot that from your original review. I've added `TestRun_BindingRuleUpdates` to `command_test.go`.

I also forgot your comment from before:
> Targeted tests for the new checkAndCreateNamespace method

I looked at code coverage and it's fully tested except for the simple error cases:

image

I like that we're mostly testing the public interfaces here while getting full coverage. In the case of a particularly complicated helper method I'm down to write separate tests but looking at the existing code coverage this seems okay to me.

subcommand/server-acl-init/command_ent_test.go
adilyse: It's a test against future human error, where someone modifies the wrong namespace flag.

subcommand/server-acl-init/command_ent_test.go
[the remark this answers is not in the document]
adilyse: This is an argument for doing comprehensive testing at the lowest possible level rather than up the call tree somewhere. Test organization rather than code coverage.

subcommand/server-acl-init/command_test.go
```diff
@@ -406,6 +407,63 @@ func TestRun_ConnectInjectAuthMethod(t *testing.T) {
...
+		"-acl-binding-rule-selector=serviceaccount.name!=changed",
+	)
+
+	// Run the command first to populate the binding rule.
+	cmd := Command{
+		UI:        ui,
+		clientset: k8s,
+	}
+	responseCode := cmd.Run(firstRunArgs)
+	require.Equal(0, responseCode, ui.ErrorWriter.String())
```
adilyse: This needs a check that the binding rule matched the first argument after the first run.

lkysow: In these update tests the pattern has been to only check that it changed. The first run creation is tested in other tests. It was done this way to reduce the amount of test code. If it breaks for the first run, it will be detected in the other tests. I've added it though.

subcommand/server-acl-init/command_test.go
```diff
@@ -406,6 +407,63 @@ func TestRun_ConnectInjectAuthMethod(t *testing.T) {
...
+	firstRunArgs := []string{
+		"-server-label-selector=component=server,app=consul,release=" + releaseName,
+		"-resource-prefix=" + resourcePrefix,
+		"-k8s-namespace=" + ns,
+		"-expected-replicas=1",
+		"-create-inject-auth-method",
+		"-acl-binding-rule-selector=serviceaccount.name!=default",
+	}
+	// Our second run, we change the binding rule selector.
+	secondRunArgs := append(firstRunArgs,
```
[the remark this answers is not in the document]
lkysow: Yeah good call, moved to

```go
	commonArgs := []string{
		"-server-label-selector=component=server,app=consul,release=" + releaseName,
		"-k8s-namespace=" + ns,
		"-expected-replicas=1",
		"-create-inject-auth-method",
	}
	firstRunArgs := append(commonArgs,
		"-acl-binding-rule-selector=serviceaccount.name!=default",
	)
	// Our second run, we change the binding rule selector.
	secondRunArgs := append(commonArgs,
		"-acl-binding-rule-selector=serviceaccount.name!=changed",
	)
```

subcommand/server-acl-init/command_ent_test.go
lkysow: Okay I added it.

subcommand/server-acl-init/command_ent_test.go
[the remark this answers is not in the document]
lkysow: Yes, I prefer testing the public interfaces and allowing the private implementation to vary. In my experience it makes the code easier to work with because you can refactor without having to update a ton of tests. It's a balance tradeoff in case your public interface tests get super complicated to exercise corner cases in helper methods. In my judgement it's not worth it in this case.

PR #200, opened 2020-02-07
Add namespace tests to the lifecycle sidecar command

ishustava:
This PR adds enterprise namespaces tests to the lifecycle command.  

It also fixes a bug with sync period. After switching to `time.Duration`,
we didn't need to parse it anymore, however, validateFlags function
always instantiated and returned `syncPeriod` variable, which was 0 by
default.

Additionally, we no longer allow sync period to be equal to 0. If it is,
then it's impossible to stop command gracefully because the select loop
will always choose 'case <-time.After(c.flagSyncPeriod)' even if
the `SIGINT` signal has been sent.

### DIFF
--- a/subcommand/lifecycle-sidecar/command.go
+++ b/subcommand/lifecycle-sidecar/command.go
@@ -26,6 +26,8 @@ type Command struct {
 	flagSet           *flag.FlagSet
 	flagLogLevel      string
 
+	consulCommand []string
+
 	once  sync.Once
 	help  string
 	sigCh chan os.Signal
@@ -58,7 +60,7 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 
-	syncPeriod, logLevel, err := c.validateFlags()
+	logLevel, err := c.validateFlags()
 	if err != nil {
 		c.UI.Error("Error: " + err.Error())
 		return 1
@@ -72,15 +74,15 @@ func (c *Command) Run(args []string) int {
 	// Log initial configuration
 	logger.Info("Command configuration", "service-config", c.flagServiceConfig,
 		"consul-binary", c.flagConsulBinary,
-		"sync-period", syncPeriod,
-		"log-level", logLevel)
+		"sync-period", c.flagSyncPeriod,
+		"log-level", c.flagLogLevel)
 
 	// Set up channel for graceful SIGINT shutdown.
 	signal.Notify(c.sigCh, os.Interrupt)
 
-	consulCommand := []string{"services", "register"}
-	consulCommand = append(consulCommand, c.parseConsulFlags()...)
-	consulCommand = append(consulCommand, c.flagServiceConfig)
+	c.consulCommand = []string{"services", "register"}
+	c.consulCommand = append(c.consulCommand, c.parseConsulFlags()...)
+	c.consulCommand = append(c.consulCommand, c.flagServiceConfig)
 
 	// The main work loop. We continually re-register our service every
 	// syncPeriod. Consul is smart enough to know when the service hasn't changed
@@ -90,7 +92,7 @@ func (c *Command) Run(args []string) int {
 	//
 	// The loop will only exit when the Pod is shut down and we receive a SIGINT.
 	for {
-		cmd := exec.Command(c.flagConsulBinary, consulCommand...)
+		cmd := exec.Command(c.flagConsulBinary, c.consulCommand...)
 
 		// Run the command and record the stdout and stderr output
 		output, err := cmd.CombinedOutput()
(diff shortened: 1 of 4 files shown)


subcommand/lifecycle-sidecar/command_ent_test.go
```diff
@@ -0,0 +1,90 @@
...
+		require.Equal(r, svcProxy.Proxy.Upstreams[0].DestinationNamespace, "dest-namespace")
+	})
+}
+
+const servicesRegistrationWithNamespaces = `
+services {
+	id   = "service-id"
+	name = "service"
+	port = 80
+    namespace = "namespace"
```
lkysow: indentation looks off here

ishustava: done!

subcommand/lifecycle-sidecar/command.go
```diff
@@ -113,7 +115,7 @@ func (c *Command) Run(args []string) int {
 
 // validateFlags validates the flags and returns the parsed syncPeriod and
 // logLevel.
-func (c *Command) validateFlags() (syncPeriod time.Duration, logLevel hclog.Level, err error) {
+func (c *Command) validateFlags() (logLevel hclog.Level, err error) {
```
adilyse: I just realized this, but this is the only place in all of `consul-k8s` that we use named return values. Is there a compelling reason to do so? It doesn't match any of the other code style in the project.

subcommand/lifecycle-sidecar/command_ent_test.go
```diff
@@ -0,0 +1,90 @@
...
+	cmd := Command{
+		UI: ui,
+	}
+
+	// Run async because we need to kill it when the test is over.
+	exitChan := runCommandAsynchronously(&cmd, []string{
+		"-http-addr", a.HTTPAddr,
+		"-service-config", configFile,
+		"-sync-period", "100ms",
+		"-consul-binary", "consul",
```
adilyse: This should be set by default, right?

subcommand/lifecycle-sidecar/command.go
ishustava: I really like the named return values because it makes the code shorter (Go will auto declare and initialize them to zero values). Another nice thing about them is that you can see the meaning of return values. For example, rather than having a function that returns `(string, string)` you could have `(name string, namespace string)`. 

In this case, though it was definitely a gotcha for me because it's easy to overlook the "magic". Plus, having `syncPeriod` be 0 wasn't actually a valid value because then you can't terminate the command at all. 

My overall feeling is that they're good to use but I don't think it makes sense to require that all functions use it in the code.

subcommand/lifecycle-sidecar/command_ent_test.go
ishustava: Yep, updated!

ishustava:
Thanks so much for the review @adilyse. I have added a [generic test](https://github.com/hashicorp/consul-k8s/blob/1347ee122d5eb2c6011f5b959338b2eebae2b0e0/subcommand/lifecycle-sidecar/command_test.go#L202-L250) checking that all command flags passed to the `lifecycle-sidecar` command are passed down to the `consul services register` command too. This test should cover ACLs and TLS. Do you think we need another test?

subcommand/lifecycle-sidecar/command.go
lkysow: I've also seen a pattern from consul:
```go
	returnErr := func(e error) (time.Duration, hclog.Level, error) {
		return 1 * time.Second, hclog.Info, err
	}
	if c.flagServiceConfig == "" {
		return returnErr(errors.New("-service-config must be set"))
	}
```

It's important to match code style but named returns are pretty idiomatic golang so I think either way is fine, as in we haven't set precedence elsewhere for dealing with this situation (afaik) so I don't think we're going against our code style, more that we're encountering a new situation.

adilyse: I can understand why named return values might be useful, but I don't see that those arguments apply in this case, especially with the gotchas you mention @ishustava . In this case, it's pretty clear what the return values are (hclog.Level and error) and without naming, it would be possible to have a single line return instead of having to define `err` and then return.

I'm happy to revisit the pattern elsewhere when it adds additional value, but this doesn't seem like that place.

adilyse:
Looks great! I'd still prefer we didn't use named returns in `validateFlags`, but I won't hold this up for it.

subcommand/lifecycle-sidecar/command.go
ishustava: Sure, I'm happy to remove them.

Since we've started the code style conversation, I realized that parsing and returning logLevel in a validation function is not very common. I've updated it to just return `error`, and when we need to pass it to the logger we can just call the parse function again. It's safe to do because the parsing function does not return an error.

PR #204, opened 2020-02-12
Don't retry when context cancelled.

lkysow:
Previously, when the context was cancelled we would enter the error
handling clause and the loop would be re-executed. Then the
backoff.Retry would exit immediately because the context was cancelled
and we'd run in a tight loop until the program was killed.

With this change, when we exit the backoff.Retry with an error, we check
if the context is done. If so, we exit immediately.

This functionality already existed in `syncer.go`:

```go
		// Get all services with tags.
		var serviceMap map[string][]string
		var meta *api.QueryMeta
			var err error
			serviceMap, meta, err = s.Client.Catalog().Services(opts)
			return err
		}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))

		// If the context is ended, then we end
		if ctx.Err() != nil {
			return
		}
```

### DIFF
--- a/catalog/to-consul/syncer.go
+++ b/catalog/to-consul/syncer.go
@@ -176,10 +176,10 @@ func (s *ConsulSyncer) watchReapableServices(ctx context.Context) {
 		}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
 		if err != nil {
 			s.Log.Warn("error querying services, will retry", "err", err)
-			continue
+		} else {
+			s.Log.Debug("[watchReapableServices] services returned from NodeServiceList",
+				"services", nodeCatalog.Services)
 		}
-		s.Log.Debug("[watchReapableServices] services returned from NodeServiceList",
-			"services", nodeCatalog.Services)
 
 		// Wait our minimum time before continuing or retrying
 		select {
--- a/catalog/to-consul/syncer_test.go
+++ b/catalog/to-consul/syncer_test.go
@@ -2,6 +2,8 @@ package catalog
 
 import (
 	"context"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


catalog/to-consul/syncer.go
```diff
@@ -176,10 +176,10 @@ func (s *ConsulSyncer) watchReapableServices(ctx context.Context) {
 		}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
 		if err != nil {
 			s.Log.Warn("error querying services, will retry", "err", err)
-			continue
```
lkysow: Previously, this `continue` would have us immediately restarting the `for` loop. Instead we should fall down into the `select` statement where we retry after a defined period of time

catalog/to-consul/syncer.go
```diff
@@ -267,6 +267,10 @@ func (s *ConsulSyncer) watchService(ctx context.Context, name, namespace string)
 			return err
 		}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
 		if err != nil {
+			// If the context is ended, then we end.
+			if ctx.Err() != nil {
+				return
+			}
```
ishustava: The docs for this function say that the `Err()` returns non-nil only if Done is closed. So falling through to continue here would mean that on the next iteration of the loop, it'll fall to `ctx.Done()` case, right? Just want to understand what's happening better 😃

lkysow: Yeah exactly so it will go back but immediately exit (after making the api call) and so loop over and over.

lkysow: Oh wait, good catch yeah we don't need this here.

lkysow:
> Looks good! I added a question inline, but nothing that should block this. You've mentioned that this is a bug; is it a bug that we could add tests for?

It's more of a bug that we see during our unit tests because we're stopping the app many times and so it continues to loop.

Maybe we can add a test, let me think about it.

catalog/to-consul/syncer.go
ishustava: Wouldn't it just hit a `return` statement in the ctx.Done case? So if we don't have the `if` statement and there is an error (i.e. context is done), it will:
1. call `continue`
1. enter the loop again
1. ctx is now done, so it will go to the `ctx.Done()` case, which calls `return`

I'm just curious if the if is necessary here if on the next iteration of the loop the function returns anyways. I could be misunderstanding this though!

lkysow: No you're correct. It doesn't need the early exit in this case.

lkysow:
@ishustava I added a test

catalog/to-consul/syncer_test.go
```diff
@@ -246,6 +248,42 @@ func TestConsulSyncer_reapServiceSameNamespace(t *testing.T) {
...
+	s.Sync([]*api.CatalogRegistration{
+		testRegistration("k8s-sync", "bar", "default"),
+	})
+
+	// Compare the call count before and after stopping the server.
+	beforeStopAPICount := callCount
+	closer()
+	time.Sleep(100 * time.Millisecond)
+	// Before the bugfix, the count would be around 500.
+	require.LessOrEqual(t, callCount-beforeStopAPICount, 2)
```
ishustava: Mine was about `1000`; it could be better not to mention it since it probably depends on your machine etc.

PR #206, opened 2020-02-12
Create a cross-consul-namespace acl policy

adilyse:
Additionally, this adds that policy to any namespace that may be
created during the bootstrapping process (only applicable to connect
injector registering into a single, non-default Consul namespace).

Tests included.


Makefile
```diff
@@ -86,6 +86,10 @@ dev-tree:
 test:
 	go test ./...
 
+# requires a consul enterprise binary on the path
+ent-test:
```
ishustava: Nice, thanks for adding this!

subcommand/server-acl-init/create_or_update.go
```diff
@@ -118,9 +118,18 @@ func (c *Command) checkAndCreateNamespace(ns string, consulClient *api.Client) e
 
 	// If not, create it
 	if namespaceInfo == nil {
+		// Create the ACLs config for the cross-Consul-namespace
+		// default policy that needs to be attached
+		aclConfig := api.NamespaceACLConfig{
+			PolicyDefaults: []api.ACLLink{
+				{Name: "cross-namespace-policy-for-k8s"},
```
ishustava: A slight nit: do we need `-for-k8s` in the name? There's no way we could limit this policy to only-k8s services (nor should we), and I think the description is enough to show that the helm chart created it.

lkysow:
Do we not need to register the policy in the default namespace as well?

Makefile
adilyse: You're welcome!

subcommand/server-acl-init/create_or_update.go
adilyse: Switched to `cross-namespace-policy`.

adilyse:
Updated to also attach the cross namespace policy to the `default` namespace's PolicyDefaults.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -232,6 +260,12 @@ func TestRun_ACLPolicyUpdates(t *testing.T) {
 			policies, _, err := consul.ACL().PolicyList(nil)
 			require.NoError(err)
 
+			// Check that we the right number of policies. The actual
```
ishustava:
```suggestion
			// Check that we attach the right number of policies. The actual
```

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -260,9 +294,25 @@ func TestRun_ACLPolicyUpdates(t *testing.T) {
...
+				"connect-inject-token",
+				"mesh-gateway-token",
+				"client-snapshot-agent-token",
+				"enterprise-license-token",
+				"cross-namespace-policy",
+			}
 			policies, _, err = consul.ACL().PolicyList(nil)
 			require.NoError(err)
 
+			// Check that we the right number of policies. The actual
```
ishustava:
```suggestion
			// Check that we attach the right number of policies. The actual
```

adilyse: Went with "have" instead, since only the `cross-namespace-policy` is attached to the namespace.

PR #207, opened 2020-02-13
Attach an acl policy to created namespaces

adilyse:
This is necessary if ACLs are enabled to grant cross Consul namespace
communication.


subcommand/sync-catalog/command.go
```diff
@@ -123,6 +124,9 @@ func (c *Command) init() {
 		"namespace mirroring")
 	c.flags.StringVar(&c.flagK8SNSMirroringPrefix, "k8s-namespace-mirroring-prefix", "",
 		"[Enterprise Only] Prefix that will be added to all k8s namespaces mirrored into Consul if mirroring is enabled.")
+	c.flags.StringVar(&c.flagCrossNamespaceACLPolicy, "cross-consul-namespace-acl-policy", "",
```
ishustava: Would it make more sense to name it `consul-cross-namespace-acl-policy` to match other names?

adilyse:
These methods are functionally to the one in the acl bootstrapper that is tested extensively. For the sync one, it doesn't make sense to add tests before the rest of the tests are merged. For the injector, this requires a bunch of additional test setup that we felt okay skipping when we chatted earlier today.

All doable, but a lot of extra work.

subcommand/sync-catalog/command.go
adilyse: I was trying to specify that it was for crossing consul namespaces, but it does end up being a little awkward. I'll update it.

adilyse:
Tests for the connect injector have been added. The sync tests will be added once the sync tests have been merged.

PR #208, opened 2020-02-13
Fix initial sync race

lkysow:
A race condition was occurring where we would delete services via the
`watchReapableServices` function if `Sync` wasn't called fast enough by
`resource.go`. This occurred because `watchReapableServices` makes a
call to the catalog and then deletes everything with a k8s tag that
isn't in its `s.services` map. If `Sync` isn't called, then this map
is empty and so it tries to delete everything.
I've solved this by blocking until `Sync` is first called.

Closes https://github.com/hashicorp/consul-k8s/pull/191

### DIFF
--- a/catalog/to-consul/syncer_test.go
+++ b/catalog/to-consul/syncer_test.go
@@ -155,6 +155,51 @@ func TestConsulSyncer_reapService(t *testing.T) {
 	}
 }
 
+// Test that the syncer doesn't reap any services until the initial sync has
+// been performed.
+func TestConsulSyncer_noReapingUntilInitialSync(t *testing.T) {
+	t.Parallel()
+
+	a, err := testutil.NewTestServerT(t)
+	require.NoError(t, err)
+	defer a.Stop()
+	client, err := api.NewClient(&api.Config{
+		Address: a.HTTPAddr,
+	})
+	require.NoError(t, err)
+	s, closer := testConsulSyncer(client)
+	// Set the sync period to 5ms so we know it will have run at least once
+	// after we wait 100ms.
+	s.SyncPeriod = 5 * time.Millisecond
+	defer closer()
+
+	// Create a service directly in Consul. Since it was created on the
+	// synthetic sync node and has the sync-associated tag, we expect
+	// it to be deleted but not until the initial sync is performed.
+	svc := testRegistration(ConsulSyncNodeName, "baz", "default")
+	_, err = client.Catalog().Register(svc, nil)
+	require.NoError(t, err)
+
+	// We wait until the syncer has had the time to delete the service.
+	// Since we set the sync period to 5ms we know that it's run syncFull at
+	// least once.
+	time.Sleep(100 * time.Millisecond)
+
+	// Invalid service should not be deleted.
+	bazInstances, _, err := client.Catalog().Service("baz", "", nil)
+	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


lkysow:
@ashangit sorry I didn't see your PR https://github.com/hashicorp/consul-k8s/pull/191 before putting this one up. I think this PR accomplishes the same fix but instead of adding a new `GetAllServices` method it waits for `Sync` to be called. Does this look like it should work for you?

ashangit:
thks @lkysow looks good to me

catalog/to-consul/syncer_test.go
```diff
@@ -155,6 +155,45 @@ func TestConsulSyncer_reapService(t *testing.T) {
...
+	defer a.Stop()
+	client, err := api.NewClient(&api.Config{
+		Address: a.HTTPAddr,
+	})
+	require.NoError(t, err)
+	s, closer := testConsulSyncer(client)
+	defer closer()
+
+	// Create a service directly in Consul. Since it was created directly we
+	// expect it to be deleted but not until the initial sync is performed.
```
adilyse:
```suggestion
	// Create a service directly in Consul. Since it was created on the 
	// synthetic sync node and has the sync-associated tag, we expect
	// it to be deleted but not until the initial sync is performed.
```

A service needs the sync tag and to be registered on the sync node or the sync process would ignore it.

catalog/to-consul/syncer_test.go
```diff
@@ -155,6 +155,45 @@ func TestConsulSyncer_reapService(t *testing.T) {
...
+	svc := testRegistration(ConsulSyncNodeName, "baz", "default")
+	_, err = client.Catalog().Register(svc, nil)
+	require.NoError(t, err)
+
+	// We wait until the syncer has had the time to delete the service.
+	time.Sleep(200 * time.Millisecond)
```
adilyse: It seems like we might run into edge cases having this value exactly equal the `SyncPeriod` in the test ConsulSyncer. What about setting up the ConsulSyncer with an even smaller interval?

lkysow: Yes! Great idea. I didn't like this sleep but a retry loop doesn't work because it might run through before syncFull has run :D

catalog/to-consul/syncer_test.go
```diff
@@ -155,6 +155,51 @@ func TestConsulSyncer_reapService(t *testing.T) {
...
+	t.Parallel()
+
+	a, err := testutil.NewTestServerT(t)
+	require.NoError(t, err)
+	defer a.Stop()
+	client, err := api.NewClient(&api.Config{
+		Address: a.HTTPAddr,
+	})
+	require.NoError(t, err)
+	s, closer := testConsulSyncer(client)
```
adilyse: Suggestion:
```
s, closer := testConsulSyncerWithConfig(client, func(s *ConsulSyncer) {
		s.SyncPeriod = 5 * time.Millisecond
	})
```

Since the `testConsulSyncer` also `Runs` the syncer, the sync period change may not get picked up if it's made afterwards.

lkysow: Good catch. PR coming.

lkysow: https://github.com/hashicorp/consul-k8s/pull/227

PR #214, opened 2020-02-20
Only update acl policies on rerun when namespaces are enabled

adilyse:


subcommand/server-acl-init/create_or_update.go
```diff
@@ -79,32 +79,37 @@ func (c *Command) createOrUpdateACLPolicy(policy api.ACLPolicy, consulClient *ap
...
 
-		// This shouldn't happen, because we're looking for a policy
-		// only after we've hit a `Policy already exists` error.
-		if policy.ID == "" {
-			return errors.New("Unable to find existing ACL policy")
+			// Update the policy now that we've found its ID
+			_, _, err = consulClient.ACL().PolicyUpdate(&policy, &api.WriteOptions{})
+			return err
+		} else {
+			c.Log.Info(fmt.Sprintf("Policy %q already exists", policy.Name))
```
ishustava:
```suggestion
			c.Log.Info(fmt.Sprintf("Policy %q already exists, skipping update", policy.Name))
```

PR #215, opened 2020-02-21
Update changelog with namespace details

adilyse:


CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+Notes:
+
+* Sync: The underlying call to Consul to retrieve services has been switched to do so by the synthetic node. This causes a slight behavior change in that we will no longer remove services with the `ConsulK8STag` if it's not attached to the `k8s-sync` node.
+
+* Sync: The allow/deny list setting change is backwards compatible for syncing all namespaces, however if if a user upgrades consul-k8s without upgrading the Helm chart as well, there will be a slight difference in behavior for syncing all namespaces in that it won't automatically exclude `kube-system` on its own.
+
+* Sync: Two sync processes running from separate namespaces in the same k8s cluster will now require different tags or they will interfere will each other's service registrations.
+
+* Bootstrap ACLs: Previously, ACL policies were not updated after creation. Now, if namespaces are enabled, they are updated every time the ACL bootstrapper is run so that any namespace config changes can be adjusted. This is only an issue if you are manually updating ACL policies after creation.
```
ishustava: We should clarify what "This" refers to in the last sentence. I also think we shouldn't call it a manual update bacause someone might have written API automation specific to their use case.

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+* Connect: Adds additional parsing of the upstream annotation to support namespaces. The format of the annotation becomes:
+
+    `service_name.namespace:port:optional_datacenter`
```
ishustava: Should it be:
```suggestion
    `service_name.optional_namespace:port:optional_datacenter`
```

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+  The `service_name.namespace` is only parsed if namespaces are enabled. If they are not enabled and someone has added a `.namespace`, the upstream will not work correctly, as is the case where someone has put in an incorrect service name, port or datacenter. If namespaces are enabled and the `.namespace` is not defined, this will automatically fallback to assuming the service is in the same namespace as the service defining the upstream.
```
ishustava:
```suggestion
  The `service_name.namespace` is only parsed if namespaces are enabled. If they are not enabled and someone has added a `.namespace`, the upstream will not work correctly, as is the case when someone has put in an incorrect service name, port or datacenter. If namespaces are enabled and the `.namespace` is not defined, Consul will automatically fallback to assuming the service is in the same namespace as the service defining the upstream.
```

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
 ## UNRELEASED
 
+Improvements:
```
lkysow: I think we need a BREAKING CHANGES section and some of your notes need to be moved here:
```
BREAKING CHANGES:

* Connect Injector
  * New flags `-allow-k8s-namespace` and `-deny-k8s-namespace` were added. Previously the injector would inject all namespaces. To maintain that behaviour, set `-allow-k8s-namespace='*'`. If no `-allow-k8s-namespace` flag is set, no namespaces will be injected.
* Catalog Sync
  * `kube-system` and `kube-public` namespaces are now synced from **unless** `-deny-k8s-namespace=kube-system -deny-k8s-namespace=kube-public` are passed to the `sync-catalog` command.
  * Previously, multiple sync processes could be run in the same Kubernetes cluster with different source Kubernetes namespaces and the same `-consul-k8s-tag`. This is no longer possible. The sync processes will now delete one-another's registrations. To continue running multiple sync processes, each process must be passed a different `-consul-k8s-tag` flag.
  * Previously, catalog sync would delete services tagged with `-consul-k8s-tag` (defaults to `k8s`) that were registered out-of-band, i.e. not by the sync process itself. It would delete services regardless of which node they were registered on. Now the sync process will only delete those services not registered by itself if they are on the `k8s-sync` node (the synthetic node created by the catalog sync process)`.
```

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+* Support Consul namespaces [Enterprise feature] in all consul-k8s components
+* Create allow and deny lists of k8s namespaces for catalog sync and Connect inject
+* Significantly increases the test coverage of all components
+
+Bug Fixes:
+
+* Bootstrap ACLs: Allow users to update their Connect ACL binding rule definition on upgrade
+* Bootstrap ACLs: Fixes mesh gateway ACL policies to have the correct permissions
+* Sync: Fixes a hot loop bug when getting an error from Consul when retrieving service information
```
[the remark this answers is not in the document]
lkysow: it's this one: `https://github.com/hashicorp/consul-k8s/pull/204`

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
 ## UNRELEASED
 
+Improvements:
+
+* Support Consul namespaces [Enterprise feature] in all consul-k8s components
+* Create allow and deny lists of k8s namespaces for catalog sync and Connect inject
+* Significantly increases the test coverage of all components
+
```
lkysow: Missing:
* Connect Inject
  * Changes default Consul Docker image (`-consul-image`) to `consul:1.7.1`
  * Changes default Envoy Docker image (`-envoy-image`) to `envoyproxy/envoy-alpine:v1.13.0`

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+* Bootstrap ACLs: Fixes mesh gateway ACL policies to have the correct permissions
+* Sync: Fixes a hot loop bug when getting an error from Consul when retrieving service information
+
+Notes:
+
+* Sync: The underlying call to Consul to retrieve services has been switched to do so by the synthetic node. This causes a slight behavior change in that we will no longer remove services with the `ConsulK8STag` if it's not attached to the `k8s-sync` node.
+
+* Sync: The allow/deny list setting change is backwards compatible for syncing all namespaces, however if if a user upgrades consul-k8s without upgrading the Helm chart as well, there will be a slight difference in behavior for syncing all namespaces in that it won't automatically exclude `kube-system` on its own.
+
+* Sync: Two sync processes running from separate namespaces in the same k8s cluster will now require different tags or they will interfere will each other's service registrations.
```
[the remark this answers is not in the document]
adilyse: The case I was referring to was each sync process only being responsible for one namespace. Before, you either ran one sync process for everything, or had to run a separate sync process for each namespace you wanted synced if you didn't want them all. That is apparently not clear here in the comment.

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
 ## UNRELEASED
 
+Improvements:
+
+* Support Consul namespaces [Enterprise feature] in all consul-k8s components
+* Create allow and deny lists of k8s namespaces for catalog sync and Connect inject
```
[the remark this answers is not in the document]
adilyse: There isn't a separate PR for the allow/deny list work. It was included as part of the initial namespace PRs.

CHANGELOG.md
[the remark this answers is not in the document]
ishustava: @adilyse have you decided not to add that link?

CHANGELOG.md
```diff
@@ -1,5 +1,46 @@
...
+
+* Bootstrap ACLs: Previously, ACL policies were not updated after creation. Now, if namespaces are enabled, they are updated every time the ACL bootstrapper is run so that any namespace config changes can be adjusted. This change is only an issue if you are updating ACL policies after creation.
+
+* Connect: Adds additional parsing of the upstream annotation to support namespaces. The format of the annotation becomes:
+
+    `service_name.optional_namespace:port:optional_datacenter`
+
+  The `service_name.namespace` is only parsed if namespaces are enabled. If they are not enabled and someone has added a `.namespace`, the upstream will not work correctly, as is the case when someone has put in an incorrect service name, port or datacenter. If namespaces are enabled and the `.namespace` is not defined, Consul will automatically fallback to assuming the service is in the same namespace as the service defining the upstream.
+
+* Connect: Due to a breaking config change in Envoy, if using Consul 1.7+, Envoy version 1.13+ is required.
```
ishustava: I think it's a change in Consul that changed the minimum version of Envoy from 1.8 to 1.10. I'm also not sure it's required, since the docs mention that consul 1.7 is compatible with 1.10: https://www.consul.io/docs/connect/proxies/envoy.html#supported-versions.

adilyse: It's only compatible with an additional flag added to the `consul connect envoy` command that creates the bootstrap config. We put off adding that flag, so 1.10 is not currently compatible with our setup. Do you have a suggestion for how to phrase that nuance?

Also, didn't you mention that 1.13 was needed for namespaces?

CHANGELOG.md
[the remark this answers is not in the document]
ishustava: Got ya. Thanks for clarifying that.

CHANGELOG.md
ishustava: Ah, ok. You're right. I haven't tested all other versions. What I saw was that our default (1.9.1) didn't work with consul 1.7. It wasn't namespace specific. 

Also, this requirement applies to connect and mesh gateways.

Maybe something like:
```
Connect and Mesh Gateways: Consul 1.7+ now requires that we pass `-envoy-version` flag if using a version other than the default (1.13.0) so that it can generate correct bootstrap configuration. This is not yet supported in the Helm chart and consul-k8s, and as such, we require Envoy version 1.13.0. 
```

ishustava: Is this a breaking change in the helm chart? If someone is overwriting their envoy image, they'll see it break.

adilyse: Added this language into the breaking change section.

PR #217, opened 2020-02-21
Inject-connect command: deprecate the -consul-ca-cert flag

ishustava:
Now that the inject-connect command has Consul's HTTP flags, we should use only one flag to control TLS communications with the local agents.

Currently, we set both flags. The `-consul-ca-cert` flag sets the CA cert for the init container, the envoy proxy sidecar, and the lifecycle sidecar, and they use that CA to talk to their local agent.

The `-ca-file` flag controls the CA cert that is used by the handler to talk to its local agent (this happens only if namespaces are enabled).

Although these agents could be on different hosts, the CA that singed their certificates must be the same, so there is no need to have two different flags.

### DIFF
--- a/subcommand/inject-connect/command.go
+++ b/subcommand/inject-connect/command.go
@@ -47,7 +47,7 @@ type Command struct {
 	flagACLAuthMethod        string // Auth Method to use for ACLs, if enabled
 	flagWriteServiceDefaults bool   // True to enable central config injection
 	flagDefaultProtocol      string // Default protocol for use with central config
-	flagConsulCACert         string // Path to CA Certificate to use when communicating with Consul clients
+	flagConsulCACert         string // [Deprecated] Path to CA Certificate to use when communicating with Consul clients
 
 	// Flags to support namespaces
 	flagEnableNamespaces           bool     // Use namespacing on all components
@@ -62,7 +62,7 @@ type Command struct {
 	http    *flags.HTTPFlags
 
 	consulClient *api.Client
-	clientset    *kubernetes.Clientset
+	clientset    kubernetes.Interface
 
 	once sync.Once
 	help string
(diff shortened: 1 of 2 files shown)


subcommand/inject-connect/command.go
```diff
@@ -62,7 +62,7 @@ type Command struct {
 	http    *flags.HTTPFlags
 
 	consulClient *api.Client
-	clientset    *kubernetes.Clientset
+	clientset    kubernetes.Interface
```
ishustava: Changing this for tests. Also we should be using interfaces rather than structs.

subcommand/inject-connect/command.go
```diff
@@ -282,7 +290,7 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 			// The CA Bundle value must be base64 encoded
 			value := base64.StdEncoding.EncodeToString(bundle.CACert)
 
-			_, err := clientset.Admissionregistration().
+			_, err := clientset.AdmissionregistrationV1beta1().
```
ishustava: `Admissionregistration()` is deprecated. Under the hood, both functions return the same client: https://github.com/kubernetes/client-go/blob/7d04d0e2a0a1a4d4a1cd6baa432a2301492e4e65/kubernetes/clientset.go#L161-L170. I'll try to update the client version in a separate PR.

subcommand/inject-connect/command_test.go
```diff
@@ -1,31 +1,40 @@
...
 	}{
 		{
-			Flags:  []string{},
-			ExpErr: "-consul-k8s-image must be set",
+			flags:  []string{},
+			expErr: "-consul-k8s-image must be set",
+		},
+		{
+			flags:  []string{"-consul-k8s-image", "foo", "-ca-file", "bar"},
+			expErr: "Error reading Consul's CA cert file",
```
lkysow:
```suggestion
			expErr: "Error reading Consul's CA cert file \"bar\"",
```

subcommand/inject-connect/command.go
```diff
@@ -145,10 +145,28 @@ func (c *Command) Run(args []string) int {
...
+		cfg.TLSConfig.CAFile = c.flagConsulCACert
+	}
+
+	// load CA file contents
+	var consulCACert []byte
+	if cfg.TLSConfig.CAFile != "" {
+		var err error
+		consulCACert, err = ioutil.ReadFile(cfg.TLSConfig.CAFile)
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error reading Consul's CA cert file %s: %s", c.flagConsulCACert, err))
```
```suggestion
			c.UI.Error(fmt.Sprintf("Error reading Consul's CA cert file %q: %s", c.flagConsulCACert, err))
```
I think having the filename quoted might make the error clearer that the filename is coming from an arg.

PR #220, opened 2020-02-28
allow to customize agent down_policy

ltagliamonte-dd:
Hi there!
Small contribution: allowing to customize agent down_policy

### DIFF
--- a/subcommand/acl-init/command.go
+++ b/subcommand/acl-init/command.go
@@ -22,17 +22,23 @@ import (
 type Command struct {
 	UI cli.Ui
 
-	flags          *flag.FlagSet
-	k8s            *k8sflags.K8SFlags
-	flagSecretName string
-	flagInitType   string
-	flagNamespace  string
-	flagACLDir     string
+	flags             *flag.FlagSet
+	k8s               *k8sflags.K8SFlags
+	flagSecretName    string
+	flagInitType      string
+	flagACLDownPolicy string
+	flagNamespace     string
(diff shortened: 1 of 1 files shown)


ltagliamonte-dd:
@ishustava do you mind have also a look at this?

lkysow:
I wonder if we change the ordering of the flags we're passing in client-daemonset to pass the extra config last.

Change
```
                {{- range .Values.client.extraVolumes }}
                {{- if .load }}
                -config-dir=/consul/userconfig/{{ .name }} \
                {{- end }}
                {{- end }}
                {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}
                -config-dir=/consul/aclconfig \
                {{- end }}
```

to
```
                {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}
                -config-dir=/consul/aclconfig \
                {{- end }}
                {{- range .Values.client.extraVolumes }}
                {{- if .load }}
                -config-dir=/consul/userconfig/{{ .name }} \
                {{- end }}
                {{- end }}
```

Then set
```yaml
client:
  extraConfig: |
     {
       "acl": {"down_policy": "your-policy"}
     }
```

And then I think you'll override what we set automatically. Can you try this?

lkysow:
Hi, we've merged https://github.com/hashicorp/consul-helm/pull/531 which should allow you to override this via the client.extraVolumes setting.

PR #225, opened 2020-03-05
Switch to testutil.TestServer

lkysow:


lkysow:
Yeah for sure I don't like leaving it like that but it looked like a pretty big job to get it working so just wanted to clear out any potential merge conflicts for mesh federation.

I think it might be best treated as a bug with the TestServer implementation. Maybe we can pair on possible solutions?

PR #227, opened 2020-03-09
Fix test configuration race.

lkysow:
Needed to set the config in the closure instead of after the syncer
started. (https://github.com/hashicorp/consul-k8s/pull/208#discussion_r389070766)


lkysow:
Good catch on your part!

PR #232, opened 2020-03-11
Add -token-sink-file flag to acl-init

lkysow:
Add flag `-token-sink-file` to `acl-init` command. If `init-type=sync` (which is used for everything except consul agents) then if `-token-sink-file` is set, we will write the token to that file.

This is needed for the mesh-gateway lifecycle sidecar (https://github.com/hashicorp/consul-helm/pull/380)

### DIFF
--- a/subcommand/acl-init/command.go
+++ b/subcommand/acl-init/command.go
@@ -22,14 +22,15 @@ import (
 type Command struct {
 	UI cli.Ui
 
-	flags          *flag.FlagSet
-	k8s            *k8sflags.K8SFlags
-	flagSecretName string
-	flagInitType   string
-	flagNamespace  string
-	flagACLDir     string
+	flags             *flag.FlagSet
+	k8s               *k8sflags.K8SFlags
+	flagSecretName    string
+	flagInitType      string
+	flagNamespace     string
+	flagACLDir        string
+	flagTokenSinkFile string
 
-	k8sClient *kubernetes.Clientset
+	k8sClient kubernetes.Interface
 
 	once sync.Once
 	help string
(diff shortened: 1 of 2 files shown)


subcommand/acl-init/command.go
```diff
@@ -45,6 +46,8 @@ func (c *Command) init() {
 		"Name of Kubernetes namespace where the servers are deployed")
 	c.flags.StringVar(&c.flagACLDir, "acl-dir", "/consul/aclconfig",
 		"Directory name of shared volume where acl config will be output")
+	c.flags.StringVar(&c.flagTokenSinkFile, "token-sink-file", "",
+		"Filepath to write acl token to when -init-type is sync")
```
ishustava: I was always confused about why we're calling the non-client type `sync`. It doesn't look like it's used anywhere in the command (except for the new condition you're adding). Do you think it's worth removing it?

subcommand/acl-init/command_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+// Test that we write the secret data to a file.
+func TestRun_TokenSinkFile(t *testing.T) {
+	t.Parallel()
+	require := require.New(t)
+	tmpDir, err := ioutil.TempDir("", "")
+	require.NoError(err)
+	defer os.Remove(tmpDir)
```
ishustava: need `os.RemoveAll` here, as well

subcommand/acl-init/command.go
lkysow: From commit history I think it's because this was previously only used by catalog sync:
```
2) acl-init This is run by the init container of clients and (optionally) the catalog sync pod. It waits for a Kubernetes secret to be populated with an ACL token that will be consumed and applied to the pod.
```

Options:
1. Ignore `-init-type` unless it's `client`. Change Helm to not pass `-init-type=sync`. This will be backwards compatible.
1. Add new `-init-type=wait`?

ishustava: I like option 1. it seems like init-type client just adds to whatever command does by default (i.e. write client config), so to me that option seems the most intuitive.

PR #238, opened 2020-03-30
server-acl-init: Add -server-address and -server-port

ishustava:
This PR proposes to change the server-acl-init command to use user-provided server-address and server-port instead of discovering servers' Pod IPs and ports via Kubernetes API.

## Why 
Currently, the `server-acl-init` command knows a lot about how the servers are deployed. It makes some assumptions about the resource names of the stateful set and uses Kubernetes API as its discovery mechanism for server addresses. This creates a fair amount of extra logic to correctly discover the server IPs and ports. In addition, we have some logic that checks if the stateful set rollout is currently in progress and waits until it's complete. The Statefulset logic was added specifically to prevent a situation when we would discover the IPs of the servers right before Kubernetes would kill and restart the container, causing its IP to change and the `server-acl-init` job to run forever trying to talk to the pod IP that no longer exists.

This PR was motivated by the idea that we don't really need to use Kube API to discover the IPs of the servers. Kubernetes already has a built-in mechanism for discovering IPs of the stateful set because each pod in a stateful set has a ["well-known" DNS name](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id). Using these DNS names would allow us to not only simplify the logic of this command but also avoid potential edge cases where the pod IP we've discovered has changed while the job is running. Additionally, this would allow us to eventually run this command against external servers.

### Command Flags
* ❗️ ❗️ **[Breaking change]** Add the required `-server-address` command flag
* Add `-server-port` command flag and default it to `8500`.
* ❗️ ❗️ **[Breaking change]** Remove `-expected-replicas`, `-release-name` (already deprecated), and `-server-label-selector` flags because we no longer need them.

### Other changes
* Because we're requiring server addresses be provided, we're removing the logic that checks whether stateful set rollout is in progress. In the case when the server addresses are DNS names of the stateful set, the reasons for this logic go away since even if pod IPs change, the Kube DNS should handle that for us.

## Testing
All testing was done on AKS with the `ishustava/consul-k8s-dev:04-07-2020-497214b` image (or equivalent).
Cases tested:
- Clean install
- Upgrade of servers
- Upgrade when there's a config change causing an update to ACL policies
- Upgrade of servers when servers were in an unhealthy state (split-brain)

## Feedback
I'm looking for feedback in the following areas:
- Am I missing any test cases?
- Flag names

### DIFF
--- a/subcommand/server-acl-init/command.go
+++ b/subcommand/server-acl-init/command.go
@@ -46,6 +46,8 @@ type Command struct {
 	flagConsulCACert              string
 	flagConsulTLSServerName       string
 	flagUseHTTPS                  bool
+	flagServerHosts               []string
+	flagServerPort                uint
 
 	// Flags to support namespaces
 	flagEnableNamespaces                 bool   // Use namespacing on all components
@@ -57,7 +59,7 @@ type Command struct {
 	flagInjectK8SNSMirroringPrefix       string // Prefix added to Consul namespaces created when mirroring injected services
 
 	flagLogLevel string
-	flagTimeout  string
+	flagTimeout  time.Duration
 
 	clientset kubernetes.Interface
 	// cmdTimeout is cancelled when the command timeout is reached.
@@ -73,14 +75,11 @@ type Command struct {
 
 func (c *Command) init() {
 	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
-	c.flags.StringVar(&c.flagReleaseName, "release-name", "",
-		"Name of Consul Helm release. Deprecated: Use -server-label-selector=component=server,app=consul,release=<release-name> instead")
-	c.flags.StringVar(&c.flagServerLabelSelector, "server-label-selector", "",
-		"Selector (label query) to select Consul server statefulset pods, supports '=', '==', and '!='. (e.g. -l key1=value1,key2=value2)")
 	c.flags.StringVar(&c.flagResourcePrefix, "resource-prefix", "",
 		"Prefix to use for Kubernetes resources. If not set, the \"<release-name>-consul\" prefix is used, where <release-name> is the value set by the -release-name flag.")
-	c.flags.IntVar(&c.flagReplicas, "expected-replicas", 1,
-		"Number of expected Consul server replicas")
+	c.flags.Var((*flags.AppendSliceValue)(&c.flagServerHosts), "server-address",
+		"The address of the Consul server(s), may be provided multiple times. At least one value is required.")
+	c.flags.UintVar(&c.flagServerPort, "server-port", 8500, "The HTTP or HTTPS port of the Consul server.")
 	c.flags.StringVar(&c.flagK8sNamespace, "k8s-namespace", "",
 		"Name of Kubernetes namespace where the servers are deployed")
 	c.flags.BoolVar(&c.flagAllowDNS, "allow-dns", false,
@@ -131,7 +130,7 @@ func (c *Command) init() {
 			"if mirroring is enabled.")
 	c.flags.StringVar(&c.flagACLReplicationTokenFile, "acl-replication-token-file", "",
 		"Path to file containing ACL token to be used for ACL replication. If set, ACL replication is enabled.")
-	c.flags.StringVar(&c.flagTimeout, "timeout", "10m",
+	c.flags.DurationVar(&c.flagTimeout, "timeout", 10*time.Minute,
 		"How long we'll try to bootstrap ACLs for before timing out, e.g. 1ms, 2s, 3m")
 	c.flags.StringVar(&c.flagLogLevel, "log-level", "info",
 		"Log verbosity level. Supported values (in order of detail) are \"trace\", "+
@@ -164,29 +163,15 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 	if len(c.flags.Args()) > 0 {
-		c.UI.Error(fmt.Sprintf("Should have no non-flag arguments."))
+		c.UI.Error("Should have no non-flag arguments.")
 		return 1
 	}
-	timeout, err := time.ParseDuration(c.flagTimeout)
-	if err != nil {
-		c.UI.Error(fmt.Sprintf("%q is not a valid timeout: %s", c.flagTimeout, err))
+	if len(c.flagServerHosts) == 0 {
+		c.UI.Error("-server-address must be set at least once")
 		return 1
 	}
-	if c.flagReleaseName != "" && c.flagServerLabelSelector != "" {
-		c.UI.Error("-release-name and -server-label-selector cannot both be set")
-		return 1
-	}
-	if c.flagServerLabelSelector != "" && c.flagResourcePrefix == "" {
-		c.UI.Error("if -server-label-selector is set -resource-prefix must also be set")
-		return 1
-	}
-	if c.flagReleaseName == "" && c.flagServerLabelSelector == "" {
-		c.UI.Error("-release-name or -server-label-selector must be set")
-		return 1
-	}
-	// If only the -release-name is set, we use it as the label selector.
-	if c.flagReleaseName != "" {
-		c.flagServerLabelSelector = fmt.Sprintf("app=consul,component=server,release=%s", c.flagReleaseName)
+	if c.flagResourcePrefix == "" {
+		c.UI.Error("-resource-prefix must be set")
 	}
 	var aclReplicationToken string
 	if c.flagACLReplicationTokenFile != "" {
@@ -204,7 +189,7 @@ func (c *Command) Run(args []string) int {
 	}
 
 	var cancel context.CancelFunc
-	c.cmdTimeout, cancel = context.WithTimeout(context.Background(), timeout)
+	c.cmdTimeout, cancel = context.WithTimeout(context.Background(), c.flagTimeout)
 	// The context will only ever be intentionally ended by the timeout.
 	defer cancel()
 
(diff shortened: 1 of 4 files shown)


subcommand/server-acl-init/command.go
```diff
@@ -123,7 +119,7 @@ func (c *Command) init() {
 	c.flags.StringVar(&c.flagInjectK8SNSMirroringPrefix, "inject-k8s-namespace-mirroring-prefix", "",
 		"[Enterprise Only] Prefix that will be added to all k8s namespaces mirrored into Consul by Connect inject "+
 			"if mirroring is enabled.")
-	c.flags.StringVar(&c.flagTimeout, "timeout", "10m",
+	c.flags.DurationVar(&c.flagTimeout, "timeout", 10*time.Minute,
```
ishustava, line 122: This change is not related necessarily, but it removes unnecessary parsing.

subcommand/server-acl-init/command_test.go
```diff
@@ -70,57 +66,48 @@ func TestRun_FlagValidation(t *testing.T) {
...
-		{
-			"-server-label-selector=component=server,app=consul,release=" + releaseName,
-			"-resource-prefix=" + resourcePrefix,
-		},
-	} {
-		t.Run(flags[0], func(t *testing.T) {
-			k8s, testSvr := completeSetup(t, resourcePrefix)
-			defer testSvr.Stop()
-			require := require.New(t)
+	k8s, testSvr := completeSetup(t, resourcePrefix)
```
lkysow, line 69: Needs `t.Parallel`

subcommand/server-acl-init/command_test.go
```diff
@@ -129,7 +116,7 @@ func TestRun_Defaults(t *testing.T) {
 func TestRun_TokensPrimaryDC(t *testing.T) {
 	t.Parallel()
 
-	cases := map[string]struct {
+	cases := []struct {
 		TokenFlag          string
 		ResourcePrefixFlag string
```
lkysow, line 121: I think remove this `ResourcePrefixFlag` and use the global `resourcePrefix` variable internally in the test. Then use `SecretName: resourcePrefix + "-client-acl-token"` (for example) for the cases

subcommand/server-acl-init/command_test.go
```diff
@@ -129,7 +116,7 @@ func TestRun_Defaults(t *testing.T) {
 func TestRun_TokensPrimaryDC(t *testing.T) {
```
lkysow, line 116: the doc block needs to be updated. Remove the "We test using the -release-name flag vs using the
// -resource-prefix flag."

subcommand/server-acl-init/command_test.go
```diff
@@ -392,7 +310,7 @@ func TestRun_TokensReplicatedDC(t *testing.T) {
 			tokenFile, fileCleanup := writeTempFile(t, bootToken)
```
lkysow, line 310: Above here we have a lot of `SecretName: "release-name-consul-client-snapshot-agent-acl-token"` but I think it should be `SecretName: resourcePrefix + "-client-snapshot-agent-acl-token"`

subcommand/server-acl-init/command_test.go
```diff
@@ -721,318 +638,100 @@ func TestRun_BindingRuleUpdates(t *testing.T) {
...
+	randomPorts := freeport.MustTake(6)
 
 	ui := cli.NewMockUi()
 	cmd := Command{
 		UI:        ui,
 		clientset: k8s,
 	}
-	cmd.init()
 
 	// Start the command before the Pod exist.
```
lkysow, line 656:
```suggestion
	// Start the command before the server is up.
```

subcommand/server-acl-init/command_test.go
```diff
@@ -721,318 +638,100 @@ func TestRun_BindingRuleUpdates(t *testing.T) {
...
 	// Run in a goroutine so we can create the Pods asynchronously
```
lkysow, line 657:
```suggestion
	// Run in a goroutine so we can start the server asynchronously
```

subcommand/server-acl-init/command_test.go
```diff
@@ -721,318 +638,100 @@ func TestRun_BindingRuleUpdates(t *testing.T) {
...
-		require.NoError(err)
-	}()
-
-	// Wait for the command to exit.
-	select {
-	case <-done:
-		require.Equal(0, responseCode, ui.ErrorWriter.String())
-	case <-time.After(2 * time.Second):
-		require.FailNow("command did not exit after 2s")
+	// Check that the agent policy was created.
```
lkysow, line 724: might be able to use the `policyExists` helper

subcommand/server-acl-init/command_test.go
```diff
@@ -721,318 +638,100 @@ func TestRun_BindingRuleUpdates(t *testing.T) {
...
-		{
-			"PUT",
-			"/v1/acl/policy",
-		},
-		{
-			"PUT",
-			"/v1/acl/token",
-		},
-	}, consulAPICalls)
+	require.True(found, "agent-token policy was not found")
```
lkysow, line 734: 💯 wow quite the refactor. The one thing we're not testing anymore is that the servers get the ACL token applied to themselves (the `PUT /v1/agent/token/agent` assertion above). That's why I used an http server here instead of the consul test server.

This is still tested like that in the noleader and clienttokensretry tests so we're okay to not have it here.

subcommand/server-acl-init/command_test.go
```diff
@@ -1505,43 +1064,31 @@ func TestRun_HTTPS(t *testing.T) {
...
+
 	responseCode := cmd.Run([]string{
-		"-server-label-selector=component=server,app=consul,release=" + releaseName,
 		"-resource-prefix=" + resourcePrefix,
 		"-k8s-namespace=" + ns,
 		"-use-https",
 		"-consul-tls-server-name", "server.dc1.consul",
 		"-consul-ca-cert", caFile,
-		"-expected-replicas=1",
+		"-server-address=127.0.0.1",
```
lkysow, line 1090: why aren't you using `strings.Split(srv.HTTPSAddr, ":")[0]` here?

subcommand/server-acl-init/command_test.go
```diff
@@ -1668,8 +1213,8 @@ func completeSetup(t *testing.T, prefix string) (*fake.Clientset, *testutil.Test
 // a Consul API client initialized for the secondary DC, the replication token
```
lkysow, line 1213: docs need updating for extra return value

subcommand/server-acl-init/command_test.go
```diff
@@ -1680,15 +1225,15 @@ func completeReplicatedSetup(t *testing.T, prefix string) (*fake.Clientset, *api
 // a Consul API client initialized for the secondary DC and a
 // cleanup function that should be called at the end of the test that cleans
```
lkysow, line 1226: docs need updating for extra return value

subcommand/server-acl-init/command_test.go
lkysow, line 1213: `completeSetup` doesn't need the `prefix` arg anymore so it should be removed

subcommand/server-acl-init/command.go
```diff
@@ -46,6 +46,8 @@ type Command struct {
 	flagConsulCACert              string
```
lkysow, line 46: can remove variables:
* `flagReleaseName`
* `flagServerLabelSelector`
* `flagReplicas`

subcommand/server-acl-init/command_test.go
```diff
@@ -37,18 +37,14 @@ func TestRun_FlagValidation(t *testing.T) {
 	}{
```
lkysow, line 37:
```
var releaseName = "release-name"
var resourcePrefix = "release-name-consul"
```

can be just resourcePrefix now. Everywhere releaseName is used can be replaced with resourcePrefix

lkysow:
I've tested this now and it works like a charm. I tried to mess it up by deleting pods and starting upgrades but it always worked!

> Am I missing any test cases?

No, I think you're good.

> Flag names

I don't know the best word for IP or hostname but not port. I think `server-address` is fine if it's documented that you shouldn't add the port.

ishustava:
@lkysow it could also be called `-server-host`. As far as I can tell, that's how this part of the [URL](https://en.wikipedia.org/wiki/URL) is called.

subcommand/server-acl-init/command.go
```diff
@@ -73,14 +75,11 @@ type Command struct {
...
-		"Name of Consul Helm release. Deprecated: Use -server-label-selector=component=server,app=consul,release=<release-name> instead")
-	c.flags.StringVar(&c.flagServerLabelSelector, "server-label-selector", "",
-		"Selector (label query) to select Consul server statefulset pods, supports '=', '==', and '!='. (e.g. -l key1=value1,key2=value2)")
 	c.flags.StringVar(&c.flagResourcePrefix, "resource-prefix", "",
 		"Prefix to use for Kubernetes resources. If not set, the \"<release-name>-consul\" prefix is used, where <release-name> is the value set by the -release-name flag.")
-	c.flags.IntVar(&c.flagReplicas, "expected-replicas", 1,
-		"Number of expected Consul server replicas")
+	c.flags.Var((*flags.AppendSliceValue)(&c.flagServerHosts), "server-address",
+		"The address of the Consul server(s), may be provided multiple times. At least one value is required.")
+	c.flags.UintVar(&c.flagServerPort, "server-port", 8500, "The HTTP or HTTPS port of the Consul server.")
```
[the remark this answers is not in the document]
ishustava: Yes!

subcommand/server-acl-init/command_test.go
ishustava: Nice! Good catch!

subcommand/server-acl-init/command_test.go
ishustava: 🤣 I don't know! Probably because ☕ hasn't kicked in ⛔ 🧠

subcommand/server-acl-init/command.go
ishustava: 👍 Ah that was probably me failing at git rebasing 😃

lkysow:
But then you're using it as a possible join url in the next pr. Also join and join-wan use the term "address":
```
  -join=<value>
     Address of an agent to join at start time. Can be specified
     multiple times.

  -join-wan=<value>
     Address of an agent to join -wan at start time. Can be specified
     multiple times.
```

ishustava:
Agree. I left it as `server-address`. 

I've made updates based on your suggestions (thank you so much for this careful review!).

subcommand/server-acl-init/command_test.go
lkysow: still need this change

subcommand/server-acl-init/command_test.go
```diff
@@ -1233,6 +1211,11 @@ func mockReplicatedSetup(t *testing.T, bootToken string) (*fake.Clientset, *api.
 // replicatedSetup is a helper function for completeReplicatedSetup and
 // mockReplicatedSetup. If bootToken is empty, it will run the server-acl-init
 // command to set up replication. Otherwise it will do it through config.
+// Returns the Kubernetes client for the secondary DC,
+// a Consul API client initialized for the secondary DC,
+// ACL replication token, the address of the secondary Consul server, and a
```
lkysow, line 1216: ordering is incorrect. It's address of secondary server, replication token

subcommand/server-acl-init/command_test.go
ishustava: 🤦‍♀️ updated, and also found a few more no longer relevant references to Pods. Ty!

subcommand/server-acl-init/command_test.go
ishustava: Updated.

PR #239, opened 2020-03-31
Update policies for namespaces

lkysow:
* node:write needs to be in the default namespace since nodes aren't
namespaced
* connect inject and catalog sync need global acl tokens and policies when namespaces are enabled because a global token is required to create namespaces

### DIFF
--- a/subcommand/server-acl-init/command.go
+++ b/subcommand/server-acl-init/command.go
@@ -397,7 +397,13 @@ func (c *Command) Run(args []string) int {
 			return 1
 		}
 
-		err = c.createLocalACL("catalog-sync", syncRules, consulDC, consulClient)
+		// If namespaces are enabled, the policy and token needs to be global
+		// to be allowed to create namespaces.
+		if c.flagEnableNamespaces {
+			err = c.createGlobalACL("catalog-sync", syncRules, consulDC, consulClient)
+		} else {
+			err = c.createLocalACL("catalog-sync", syncRules, consulDC, consulClient)
+		}
 		if err != nil {
 			c.Log.Error(err.Error())
 			return 1
@@ -411,7 +417,14 @@ func (c *Command) Run(args []string) int {
 			return 1
 		}
 
-		err = c.createLocalACL("connect-inject", injectRules, consulDC, consulClient)
+		// If namespaces are enabled, the policy and token needs to be global
+		// to be allowed to create namespaces.
+		if c.flagEnableNamespaces {
+			err = c.createGlobalACL("connect-inject", injectRules, consulDC, consulClient)
+		} else {
+			err = c.createLocalACL("connect-inject", injectRules, consulDC, consulClient)
+		}
+
 		if err != nil {
 			c.Log.Error(err.Error())
 			return 1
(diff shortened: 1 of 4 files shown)


subcommand/server-acl-init/command.go
```diff
@@ -391,27 +391,27 @@ func (c *Command) Run(args []string) int {
...
 	if c.flagCreateInjectToken {
-		injectRules, err := c.injectRules()
+		injectRules, localToken, err := c.injectRules()
 		if err != nil {
 			c.Log.Error("Error templating inject rules", "err", err)
 			return 1
 		}
 
-		err = c.createLocalACL("connect-inject", injectRules, consulDC, consulClient)
+		err = c.createACL("connect-inject", injectRules, localToken, consulDC, consulClient)
```
ishustava, line 414: Why do we need to return the boolean signaling whether the policy/token should be local above? Couldn't we just pass `!c.flagEnableNamespaces`?

subcommand/server-acl-init/command.go
```diff
@@ -464,6 +464,8 @@ func (c *Command) Run(args []string) int {
 			c.Log.Error("Error templating acl replication token rules", "err", err)
 			return 1
 		}
+		// Policy must be global because it replicates from the primary DC
+		// and so the primary DC needs to accept the token.
```
ishustava, line 468:
```suggestion
		// and so the primary DC needs to be able to accept the token.
```

subcommand/server-acl-init/rules.go
```diff
@@ -174,12 +183,12 @@ operator = "write"
 agent_prefix "" {
   policy = "read"
 }
+node_prefix "" {
+  policy = "write"
+}
```
ishustava, line 188: What is the difference between these two placements? As far as I can tell the below placement means the node:write policy will apply to all namespaces with `namespace_prefix` being `""`.

subcommand/server-acl-init/command.go
lkysow: Yes but I did it this way because it's not the fact that we're using namespaces that means we need a global token (see the client policy). It's the fact that this specific component needs to be able to create namespaces. I thought it best to put the "knows that this component needs to create namespaces" knowledge into the rules function because it's already giving the rules operator:write because it knows this component needs to create namespaces so the knowledge is contained within that function.

subcommand/server-acl-init/rules.go
lkysow: Yeah I agree but it's not allowed in there, it errors out.

lkysow: `rpc error making call: rpc error making call: Invalid namespace acl policy: namespace policies may not grant write privileges to nodes`

subcommand/server-acl-init/command.go
ishustava: Interesting. I found it a bit confusing because there's nothing that function does to actually make that decision; it's especially evident in the tests where we're testing for boolean values but they are always the opposite of enableNamespaces. You could separate it into a variable to make it more clear.

If you feel strongly about keeping in the function, I would recommend renaming the variable to something more boolean-like, e.g. `isLocal`, because on an initial read I thought it's an actual token.

subcommand/server-acl-init/rules.go
ishustava: Do you mean specifically the `write` policy? We have examples in [our docs](https://www.consul.io/docs/acl/acl-rules.html#namespace-rules-enterprise) that show node:read policy, so I'm pretty sure we can put `node` polices under `namespace_prefix`. Am I missing something?

ishustava: Oh strange, when I was writing my comment I didn't see your second response. That makes sense.

subcommand/server-acl-init/command.go
lkysow: That's fair. I've moved it out of rules and into the `if` statement.

PR #245, opened 2020-04-09
Consul will support envoy 1.14.1 with Consul 1.8

i0rek:
See https://github.com/hashicorp/consul/pull/7624.


lkysow:
Latest release has envoy 1.16.0!

PR #248, opened 2020-04-20
Change perms from 0400 to 0600 for acl-token

lkysow:
If the acl-init command is re-run, it needs to be able to overwrite the
previously written acl-token file.

In some testing, the `service-init` init container has later commands fail. Then Kubernetes restarts the init container and it continuously fails with the error:
```
Error writing token to file "/consul/service/acl-token": open /consul/service/acl-token: permission denied
```
because that file has already been written but it has 0400 permissions so can't be overwritten.

To reproduce, edit `mesh-gateway-deployment`:
```
                ls -la /consul/service
                {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}
                consul-k8s acl-init \
                  -secret-name="{{ template "consul.fullname" . }}-mesh-gateway-acl-token" \
                  -k8s-namespace={{ .Release.Namespace }} \
                  -token-sink-file=/consul/service/acl-token
                {{ end }}
                exit 1
```

The first run-through will write the token. Subsequent run throughs will always fail.

### DIFF
--- a/subcommand/acl-init/command_test.go
+++ b/subcommand/acl-init/command_test.go
@@ -85,3 +85,48 @@ func TestRun_TokenSinkFileErr(t *testing.T) {
 		`Error writing token to file "/this/filepath/does/not/exist": open /this/filepath/does/not/exist: no such file or directory`,
 	)
 }
+
+// Test that if the command is run twice it succeeds. This test is the result
+// of a bug that we discovered where the command failed on second runs because
+// the token file only had read permissions (0400).
+func TestRun_TokenSinkFileTwice(t *testing.T) {
+	t.Parallel()
+	require := require.New(t)
+	tmpDir, err := ioutil.TempDir("", "")
+	require.NoError(err)
+	defer os.RemoveAll(tmpDir)
+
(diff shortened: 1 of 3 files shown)


ishustava:
@lkysow This change looks good. It'd be good to add a test or an assertion to check file permissions since this was a bug. I think your comment pretty much covers the intention behind it, but just to make sure it won't be overlooked in the future.

lkysow:
Absolutely, i've added a test for this!

subcommand/acl-init/command_test.go
```diff
@@ -85,3 +85,48 @@ func TestRun_TokenSinkFileErr(t *testing.T) {
 		`Error writing token to file "/this/filepath/does/not/exist": open /this/filepath/does/not/exist: no such file or directory`,
 	)
 }
+
+// Test that if the command is run twice it succeeds. This test is the result
+// of a bug that we discovered where the command failed on second runs because
```
ishustava, line 90:
```suggestion
// of a bug that we discovered where the command failed on subsequent runs because
```

Issue #246, opened 2020-04-15
Consul connect environment variable injection fails when using different datacenter upstream


lkysow:
If setting an upstream annotation like:
```
    "consul.hashicorp.com/connect-service-upstreams": "static-server:1234:dc2"
```
The environment variables `STATIC_SERVER_CONNECT_SERVICE_HOST=127.0.0.1`, `STATIC_SERVER_CONNECT_SERVICE_PORT=1234` aren't injected.

This is due to the logic here: https://github.com/hashicorp/consul-k8s/blob/master/connect-inject/container_env.go#L18-L20 which doesn't account for the colon-separated datacenter.

ishustava:
Fixed by #250

PR #250, opened 2020-04-23
handle upstream annotations that contain a datacenter segment

DerekStrickland:


DerekStrickland:
Should handle [this](https://github.com/hashicorp/consul-k8s/issues/246) issue

shellfu:
Any status on getting this merged in at all?

connect-inject/container_env_test.go
```diff
@@ -0,0 +1,78 @@
...
+			},
+		}
+	}
+
+	cases := []struct {
+		Name        string
+		Pod         func(*corev1.Pod) *corev1.Pod
+		ServiceHost string // Service host name used to craft env var names
+		ServicePort string // Service port that should be set to env var
+	}{
```
ishustava, line 33: I'd recommend just hard-coding the upstream values and also adding a testcase for an upstream without datacenter. I suggest changing this to:
```go
cases := []struct {
  name string
  upstream string
}{
  "Upstream with datacenter",
  "static-server:7890:dc1",
},
{
  "Upstream without datacenter",
  "static-server:7890",
}
```
Then you could move the minimal pod inside the `for` loop and remove the `serviceHost`, `servicePort`, and `upstreamAnnotation` variables.

```go

	for _, tt := range cases {
		t.Run(tt.Name, func(t *testing.T) {
			require := require.New(t)

			var h Handler
			envVars := h.containerEnvVars(&corev1.Pod{
				Annotations: map[string]string{
					annotationService: "foo",
                                        annotationUpsteams: tt.upstream,
				},
			      },
		           }
	                })

			....
}
```

connect-inject/container_env_test.go
```diff
@@ -0,0 +1,78 @@
...
+					portEntryFound = true
+					if envVars[i].Value == tt.ServicePort {
+						portCorrectlySet = true
+					}
+				}
+			}
+
+			require.True(hostEntryFound)
+			require.True(portEntryFound)
+			require.True(portCorrectlySet)
```
ishustava, line 75: I think we always add two env variables per upstream. In that case, would changing this to one assertion work?

```go
require.ElementsMatch(envVars, []corev1.EnvVar{
	Name:  "STATIC_SERVER_CONNECT_SERVICE_HOST",
	Value: "127.0.0.1",
}, corev1.EnvVar{
	Name:  "STATIC_SERVER_CONNECT_SERVICE_PORT",
	Value: "7809",
})
```

connect-inject/container_env_test.go
DerekStrickland: Thanks for the feedback @ishustava .  I'll take a stab in my off-hours and ping when I've updated.

DerekStrickland: @ishustava  I made the modifications.  Thanks for the input.  Simplified things a lot.

connect-inject/container_env_test.go
DerekStrickland: done

PR #259, opened 2020-05-07
Extract http flags from consul/command pkg

lkysow:
Extract flags from `consul/command` into our own package: `subcommand/flags`. Remove the flags we weren't using, that leaves:
* `-http-addr`
  * Not used anywhere but we use `CONSUL_HTTP_ADDR` everywhere so this flag matches. We do use it in our tests.
* `-token`
  * Not used anywhere but we use `CONSUL_HTTP_TOKEN` in a lot of locations so this flag matches.
* `-token-file`
  * Used in mesh gateways and lifecycle sidecars
* `-ca-file`
  * Used by get-auto-encrypt-ca command. `CONSUL_CACERT` is used by almost all of our components.
* `-ca-path`
   * This isn't used by anything currently but I can see us needing this in the future to support intermediate certs.
* `-tls-server-name`
  * Used by get-auto-encrypt-ca

Testing: Use `lkysow/consul-k8s-dev:may27-flags`. Test with all of our components, they'll crash  if a flag isn't supported. Note, I've tested this myself.
```yaml
global:
  imageK8S: lkysow/consul-k8s-dev:may27-flags
  image: consul:beta
  name: consul
  tls:
    enabled: true
    autoEncryptEnabled: true
  federation:
    enabled: true
  acls:
    manageSystemACLs: true
connectInject:
  enabled: true
syncCatalog:
  enabled: true
meshGateway:
  enabled: true
```


subcommand/inject-connect/command.go
```diff
@@ -16,22 +16,15 @@ import (
...
 	"github.com/hashicorp/consul/api"
-	"github.com/hashicorp/consul/command/flags"
 	"github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/cli"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
-type arrayFlags []string
```
lkysow, line 28: This wasn't used.

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -232,10 +232,6 @@ func TestRun_ConsulCommandFlags(t *testing.T) {
 		"-token=abc",
 		"-token-file=/token/file",
 		"-ca-file=/ca/file",
-		"-ca-path=/ca/path",
```
lkysow, line 235: These weren't actually used by any lifecycle-sidecar callers.

lkysow:
@hashicorp/consul-eco-platform this is a draft of what it looks like to remove our dependency on `consul/command`.

subcommand/flags/flag_duration_value.go
```diff
@@ -0,0 +1,64 @@
...
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+// Taken from https://github.com/hashicorp/consul/blob/master/command/flags/flag_duration_value.go
+// This was done so we don't depend on internal Consul implementation.
+
+// DurationValue provides a flag value that's aware if it has been set.
+type DurationValue struct {
```
ishustava, line 14: This type is currently only used by the [sync-catalog command](https://github.com/hashicorp/consul-k8s/blob/177d59b5d92f7cab73c67087072172d33f21dd3a/subcommand/sync-catalog/command.go#L46). I've also found that you could just use `time.Duration` and the flags would parse correctly (see [this](https://github.com/hashicorp/consul-k8s/blob/177d59b5d92f7cab73c67087072172d33f21dd3a/subcommand/lifecycle-sidecar/command.go#L26) example in the lifecycle-sidecar command). I'm wondering if we can replace the type of the `flagConsulWritePeriod` with `time.Duration`, then maybe we don't need this at all?

I have tried it out quickly, and it seems to be working ([branch](https://github.com/hashicorp/consul-k8s/tree/remove-flags-durationvalue)). This I think would also be fine to refactor afterward as it seems like it's expanding the scope of this task.

subcommand/flags/flag_map_value.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Taken from https://github.com/hashicorp/consul/blob/master/command/flags/flag_map_value.go
+// This was done so we don't depend on internal Consul implementation.
+
+// Ensure implements
+var _ flag.Value = (*FlagMapValue)(nil)
+
+// FlagMapValue is a flag implementation used to provide key=value semantics
+// multiple times.
+type FlagMapValue map[string]string
```
ishustava, line 17: I couldn't find how this type is used?

subcommand/flags/flag_slice_value.go
```diff
@@ -0,0 +1,23 @@
+package flags
+
+import "strings"
+
+// Taken from https://github.com/hashicorp/consul/blob/master/command/flags/flag_slice_value.go
```
ishustava, line 5: Could you make all the links tied to a specific git sha? If this code is moved or deleted, then we'd lose this reference.

subcommand/flags/http.go
```diff
@@ -0,0 +1,144 @@
...
+// Consul implementation.
+
+// HTTPFlags are flags used to configure communication with a Consul agent.
+type HTTPFlags struct {
+	address       StringValue
+	token         StringValue
+	tokenFile     StringValue
+	caFile        StringValue
+	caPath        StringValue
+	tlsServerName StringValue
```
ishustava, line 22: Should we keep client cert and client key? If someone is using this binary standalone they may be passing these to the commands because they've set up the client to accept mTLS connections.

subcommand/flags/usage.go
```diff
@@ -0,0 +1,117 @@
+package flags
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+	"io"
+	"strings"
+
+	text "github.com/kr/text"
```
ishustava, line 10: Do we need to alias it?

subcommand/flags/usage.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return u.String()
+}
+
+type Usager struct {
+	Usage string
+	Flags *flag.FlagSet
+}
+
+func (u *Usager) String() string {
```
ishustava, line 30: Do we need to test this 😬 ?

go.mod
```diff
@@ -18,9 +18,11 @@ require (
...
 	github.com/imdario/mergo v0.3.8 // indirect
 	github.com/json-iterator/go v1.1.8 // indirect
+	github.com/kr/text v0.1.0
 	github.com/mattbaird/jsonpatch v0.0.0-20171005235357-81af80346b1a
 	github.com/mitchellh/cli v1.0.0
 	github.com/mitchellh/go-homedir v1.1.0
+	github.com/mitchellh/mapstructure v1.1.2
 	github.com/onsi/ginkgo v1.10.3 // indirect
 	github.com/onsi/gomega v1.7.1 // indirect
 	github.com/prometheus/common v0.0.0-20181126121408-4724e9255275
```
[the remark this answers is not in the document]
lkysow: yep, ran go mod tidy

subcommand/flags/flag_duration_value.go
lkysow: I think it's fine to include, thanks for doing the leg work. Done.

subcommand/flags/flag_map_value.go
lkysow: removed

subcommand/flags/flag_slice_value.go
lkysow: good call, done!

subcommand/flags/http.go
lkysow: right, added back in

subcommand/flags/usage.go
lkysow: no, was just copied. removed.

subcommand/flags/usage.go
lkysow: I've added a test!

go.mod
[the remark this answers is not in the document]
ishustava: Danke!

subcommand/flags/usage.go
ishustava: 🙏

PR #260, opened 2020-05-12
Resolve hostnames in service-address command

lkysow:
Consul does not support hostnames for mesh gateway wan addresses. For
now to support running on EKS we will resolve the EKS load balancer
hostnames to IPs.

In the future when Consul supports hostnames we will want to remove this
change because the IPs may change and the hostname won't.


lkysow:
Note: I've tested this in EKS and it solves the issue.

PR #264, opened 2020-05-19
Update server acl init to create tokens for gateways

adilyse:
This allows the server acl init job to create base tokens for any
provided ingress and terminating gateway names.

For terminating gateways, users will need to update these created
tokens to include permissions to the services they represent. To
do this, they should add separate policies or attach service identities
to avoid having this job overwrite their configuration.

When Consul namespaces are enabled, users can provide a namespace with
each gateway name to specify which Consul namespace to scope the gateway
token to. In this case, the gateway flag values can be provided as
`<gateway-name>.<gateway-namespace>`. If a namespace is not provided,
the token will be created for the `default` namespace.

### DIFF
--- a/subcommand/server-acl-init/command.go
+++ b/subcommand/server-acl-init/command.go
@@ -49,7 +49,9 @@ type Command struct {
 
 	flagCreateSnapshotAgentToken bool
 
-	flagCreateMeshGatewayToken bool
+	flagCreateMeshGatewayToken  bool
+	flagIngressGatewayNames     []string
+	flagTerminatingGatewayNames []string
 
 	// Flags to configure Consul connection
 	flagServerAddresses     []string
@@ -124,6 +126,14 @@ func (c *Command) init() {
 		"[Enterprise Only] Toggle for creating a token for the Consul snapshot agent deployment.")
 	c.flags.BoolVar(&c.flagCreateMeshGatewayToken, "create-mesh-gateway-token", false,
 		"Toggle for creating a token for a Connect mesh gateway.")
+	c.flags.Var((*flags.AppendSliceValue)(&c.flagIngressGatewayNames), "ingress-gateway-name",
+		"Name of an ingress gateway that needs an acl token. May be specified multiple times. "+
+			"[Enterprise Only] If using Consul namespaces and registering the gateway outside of the "+
+			"default namespace, specify the value in the form <GatewayName>.<ConsulNamespace>.")
+	c.flags.Var((*flags.AppendSliceValue)(&c.flagTerminatingGatewayNames), "terminating-gateway-name",
+		"Name of a terminating gateway that needs an acl token. May be specified multiple times. "+
+			"[Enterprise Only] If using Consul namespaces and registering the gateway outside of the "+
+			"default namespace, specify the value in the form <GatewayName>.<ConsulNamespace>.")
 
 	c.flags.Var((*flags.AppendSliceValue)(&c.flagServerAddresses), "server-address",
 		"The IP, DNS name or the cloud auto-join string of the Consul server(s). If providing IPs or DNS names, may be specified multiple times."+
@@ -138,7 +148,7 @@ func (c *Command) init() {
 
 	c.flags.BoolVar(&c.flagEnableNamespaces, "enable-namespaces", false,
 		"[Enterprise Only] Enables namespaces, in either a single Consul namespace or mirrored [Enterprise only feature]")
-	c.flags.StringVar(&c.flagConsulSyncDestinationNamespace, "consul-sync-destination-namespace", "default",
+	c.flags.StringVar(&c.flagConsulSyncDestinationNamespace, "consul-sync-destination-namespace", consulDefaultNamespace,
 		"[Enterprise Only] Indicates which Consul namespace that catalog sync will register services into. If "+
 			"'-enable-sync-k8s-namespace-mirroring' is true, this is not used.")
 	c.flags.BoolVar(&c.flagEnableSyncK8SNSMirroring, "enable-sync-k8s-namespace-mirroring", false, "[Enterprise Only] "+
@@ -146,7 +156,7 @@ func (c *Command) init() {
 	c.flags.StringVar(&c.flagSyncK8SNSMirroringPrefix, "sync-k8s-namespace-mirroring-prefix", "",
 		"[Enterprise Only] Prefix that will be added to all k8s namespaces mirrored into Consul by catalog sync "+
 			"if mirroring is enabled.")
-	c.flags.StringVar(&c.flagConsulInjectDestinationNamespace, "consul-inject-destination-namespace", "default",
+	c.flags.StringVar(&c.flagConsulInjectDestinationNamespace, "consul-inject-destination-namespace", consulDefaultNamespace,
 		"[Enterprise Only] Indicates which Consul namespace that the Connect injector will register services into. If "+
 			"'-enable-inject-k8s-namespace-mirroring' is true, this is not used.")
 	c.flags.BoolVar(&c.flagEnableInjectK8SNSMirroring, "enable-inject-k8s-namespace-mirroring", false, "[Enterprise Only] "+
@@ -386,7 +396,7 @@ func (c *Command) Run(args []string) int {
 			},
 		}
 		consulNamespace := api.Namespace{
-			Name: "default",
+			Name: consulDefaultNamespace,
 			ACLs: &aclConfig,
 		}
 		_, _, err = consulClient.Namespaces().Update(&consulNamespace, &api.WriteOptions{})
@@ -491,6 +501,118 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
+	if len(c.flagIngressGatewayNames) > 0 {
+		// Create a token for each ingress gateway name. Each gateway needs a
+		// separate token because users may need to attach different policies
+		// to each gateway token depending on what the services it represents
+		for _, name := range c.flagIngressGatewayNames {
+			if name == "" {
+				c.log.Error("Ingress gateway names cannot be empty")
+				return 1
+			}
+
+			// Parse optional namespace, erroring if a user
+			// provides a namespace when not enabling namespaces.
+			var namespace string
+			if c.flagEnableNamespaces {
+				parts := strings.SplitN(strings.TrimSpace(name), ".", 2)
+				if len(parts) > 1 {
+					// Name and namespace were provided
+					name = parts[0]
+
+					// Use default namespace if provided flag is of the
+					// form "name."
+					if parts[1] != "" {
+						namespace = parts[1]
+					} else {
+						namespace = consulDefaultNamespace
+					}
+				} else {
+					// Use the default Consul namespace
+					namespace = consulDefaultNamespace
+				}
+			} else if strings.ContainsAny(name, ".") {
+				c.log.Error("Gateway names shouldn't include a namespace if Consul namespaces aren't enabled",
+					"gateway-name", name)
+				return 1
+			}
+
+			// Define the gateway rules
+			ingressGatewayRules, err := c.ingressGatewayRules(name, namespace)
+			if err != nil {
+				c.log.Error("Error templating ingress gateway rules", "gateway-name", name,
+					"namespace", namespace, "err", err)
+				return 1
+			}
+
+			// The names in the Helm chart are specified by users and so may not contain
+			// the words "ingress-gateway". We need to create unique names for tokens
+			// across all gateway types and so must suffix with `-ingress-gateway`.
+			tokenName := fmt.Sprintf("%s-ingress-gateway", name)
+			err = c.createLocalACL(tokenName, ingressGatewayRules, consulDC, consulClient)
+			if err != nil {
+				c.log.Error(err.Error())
+				return 1
+			}
+		}
+	}
+
+	if len(c.flagTerminatingGatewayNames) > 0 {
+		// Create a token for each terminating gateway name. Each gateway needs a
+		// separate token because users may need to attach different policies
+		// to each gateway token depending on what the services it represents
+		for _, name := range c.flagTerminatingGatewayNames {
+			if name == "" {
+				c.log.Error("Terminating gateway names cannot be empty")
+				return 1
+			}
+
+			// Parse optional namespace. This does not protect against a user
+			// that provides a namespace with namespaces not enabled.
+			var namespace string
+			if c.flagEnableNamespaces {
+				parts := strings.SplitN(strings.TrimSpace(name), ".", 2)
+				if len(parts) > 1 {
+					// Name and namespace were provided
+					name = parts[0]
+
+					// Use default namespace if provided flag is of the
+					// form "name."
+					if parts[1] != "" {
+						namespace = parts[1]
+					} else {
+						namespace = consulDefaultNamespace
+					}
+				} else {
+					// Use the default Consul namespace
+					namespace = consulDefaultNamespace
+				}
+			} else if strings.ContainsAny(name, ".") {
+				c.log.Error("Gateway names shouldn't include a namespace if Consul namespaces aren't enabled",
+					"gateway-name", name)
+				return 1
+			}
+
+			// Define the gateway rules
+			terminatingGatewayRules, err := c.terminatingGatewayRules(name, namespace)
+			if err != nil {
+				c.log.Error("Error templating terminating gateway rules", "gateway-name", name,
+					"namespace", namespace, "err", err)
+				return 1
+			}
+
+			// The names in the Helm chart are specified by users and so may not contain
+			// the words "ingress-gateway". We need to create unique names for tokens
+			// across all gateway types and so must suffix with `-terminating-gateway`.
+			tokenName := fmt.Sprintf("%s-terminating-gateway", name)
+			err = c.createLocalACL(tokenName, terminatingGatewayRules, consulDC, consulClient)
+			if err != nil {
+				c.log.Error(err.Error())
+				return 1
+			}
+		}
+	}
+
 	if c.flagCreateInjectAuthMethod {
 		err := c.configureConnectInject(consulClient)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


subcommand/server-acl-init/command.go
```diff
@@ -491,6 +497,60 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
+	if len(c.flagIngressGateways) > 0 {
+		// Create a token for each ingress gateway name
```
lkysow, line 501: Can you comment as to *why* we need to create a token for *each* gateway even though they all have the same policy (because after the fact users may need to attach different policies to each gateway token depending on what the gateway's config is)

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +497,60 @@ func (c *Command) Run(args []string) int {
...
+			// Check that the gateway name is not an empty string
+			if gateway == "" {
+				c.log.Error("Ingress gateway names cannot be empty")
+				return 1
+			}
+
+			// Define the gateway rules
+			ingressGatewayRules, err := c.ingressGatewayRules(gateway)
+			if err != nil {
+				c.log.Error("Error templating ingress gateway rules", "gateway name", gateway, "err", err)
```
lkysow, line 512:
```suggestion
				c.log.Error("Error templating ingress gateway rules", "gateway-name", gateway, "err", err)
```

I think we don't use spaces in these keys (could be wrong).

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +497,60 @@ func (c *Command) Run(args []string) int {
...
+			}
+
+			// Names can be user specified in the Helm chart, and so don't
+			// append `-ingress-gateway`. For the ACL components, however, it
+			// is helpful to know the type of component.
```
lkysow, line 517: This comment is confusing to me. Maybe

```
The names in the Helm chart are specified by users and so may not contain the words "ingress-gateway".
We need to create unique names for policies across all gateway types and so must suffix with -ingress-gateway.
```

subcommand/server-acl-init/rules.go
```diff
@@ -11,6 +11,7 @@ type rulesData struct {
 	ConsulSyncDestinationNamespace string
 	EnableSyncK8SNSMirroring       bool
 	SyncK8SNSMirroringPrefix       string
+	GatewayName                    string
```
lkysow, line 14: Can you add a comment that this is only set with ingress and terminating gateways.

adilyse:
This currently creates gateway acl tokens in the Consul `default` namespace. This will work, but we should decide if there are cases where the gateway should be defined in another namespace (for example if `-consul-inject-destination-namespace` is defined).

subcommand/server-acl-init/rules.go
```diff
@@ -186,10 +223,10 @@ namespace_prefix "" {
 }
 {{- end }}
 `
-	return c.renderRules(aclReplicationRulesTpl)
+	return c.renderRules(aclReplicationRulesTpl, "")
 }
 
-func (c *Command) renderRules(tmpl string) (string, error) {
+func (c *Command) renderRules(tmpl, gatewayName string) (string, error) {
```
lkysow, line 229: Adding this here and having to set it to `""` in the non-gateway callers seems like a smell to me. What about splitting this and `rulesData` out into a new method/struct?

```go
type rulesData struct {
	EnableNamespaces               bool
	ConsulSyncDestinationNamespace string
	EnableSyncK8SNSMirroring       bool
	SyncK8SNSMirroringPrefix       string
}

type gatewayRulesData struct {
	rulesData
	GatewayName string
}

func (c *Command) renderRules(tmpl string) (string, error) {
	return c.renderRulesGeneric(tmpl, c.rulesData())
}

func (c *Command) renderGatewayRules(tmpl, gatewayName string) (string, error) {
	// Populate the data that will be used in the template.
	// Not all templates will need all of the fields.
	data := gatewayRulesData{
		rulesData:   c.rulesData(),
		GatewayName: gatewayName,
	}

	return c.renderRulesGeneric(tmpl, data)
}

func (c *Command) rulesData() rulesData {
	return rulesData{
		EnableNamespaces:               c.flagEnableNamespaces,
		ConsulSyncDestinationNamespace: c.flagConsulSyncDestinationNamespace,
		EnableSyncK8SNSMirroring:       c.flagEnableSyncK8SNSMirroring,
		SyncK8SNSMirroringPrefix:       c.flagSyncK8SNSMirroringPrefix,
	}
}

func (c *Command) renderRulesGeneric(tmpl string, data interface{}) (string, error) {
	// Check that it's a valid template
	compiled, err := template.New("root").Parse(strings.TrimSpace(tmpl))
	if err != nil {
		return "", err
	}

	// Render the template
	var buf bytes.Buffer
	err = compiled.Execute(&buf, data)
	if err != nil {
		// Discard possible partial results on error return
		return "", err
	}

	return buf.String(), nil
}

```

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -662,21 +700,24 @@ func TestRun_TokensWithNamespacesEnabled(t *testing.T) {
...
-			policy := policyExists(t, c.PolicyName, consul)
-			require.Equal(c.PolicyDCs, policy.Datacenters)
 
-			// Test that the token was created as a Kubernetes Secret.
-			tokenSecret, err := k8s.CoreV1().Secrets(ns).Get(c.SecretName, metav1.GetOptions{})
-			require.NoError(err)
-			require.NotNil(tokenSecret)
-			token, ok := tokenSecret.Data["token"]
-			require.True(ok)
+			for i, _ := range c.PolicyNames {
```
lkysow, line 704:
```suggestion
			for i := range c.PolicyNames {
```

subcommand/server-acl-init/command_test.go
```diff
@@ -212,24 +248,27 @@ func TestRun_TokensPrimaryDC(t *testing.T) {
...
-			policy := policyExists(t, c.PolicyName, consul)
-			require.Equal(c.PolicyDCs, policy.Datacenters)
 
-			// Test that the token was created as a Kubernetes Secret.
-			tokenSecret, err := k8s.CoreV1().Secrets(ns).Get(c.SecretName, metav1.GetOptions{})
-			require.NoError(err)
-			require.NotNil(tokenSecret)
-			token, ok := tokenSecret.Data["token"]
-			require.True(ok)
+			for i, _ := range c.PolicyNames {
```
lkysow, line 252:
```suggestion
			for i := range c.PolicyNames {
```

subcommand/server-acl-init/command_test.go
```diff
@@ -313,36 +387,36 @@ func TestRun_TokensReplicatedDC(t *testing.T) {
...
-				require.NotNil(r, tokenSecret)
-				token, ok := tokenSecret.Data["token"]
-				require.True(r, ok)
-
-				// Test that the token has the expected policies in Consul.
-				tokenData, _, err := consul.ACL().TokenReadSelf(&api.QueryOptions{Token: string(token)})
-				require.NoError(r, err)
-				require.Equal(r, c.PolicyName, tokenData.Policies[0].Name)
-				require.Equal(r, c.LocalToken, tokenData.Local)
+				for i, _ := range c.PolicyNames {
```
lkysow, line 403:
```suggestion
				for i := range c.PolicyNames {
```

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +497,60 @@ func (c *Command) Run(args []string) int {
...
+			name := fmt.Sprintf("%s-ingress-gateway", gateway)
```
lkysow, line 518: So for more folks this is going to end up with `ingress-gateway-ingress-gateway` and `terminating-gateway-terminating-gateway`. What do you think about suffixing with `-igw` and `-tgw` instead?

lkysow:
> This currently creates gateway acl tokens in the Consul default namespace. This will work, but we should decide if there are cases where the gateway should be defined in another namespace (for example if -consul-inject-destination-namespace is defined).

From the RFC I thought it had to be in the default ns for now?

subcommand/server-acl-init/rules.go
adilyse: I've made the change because it seems like a reasonable suggestion, but nobody likes to hear that their code smells. You should try "what about `<suggestion>`?" or "it might be more modular/less repetitive/`<specific reason>` to `<suggestion>`" instead.

subcommand/server-acl-init/rules.go
adilyse: After making the proposed refactor, this no longer applies.

subcommand/server-acl-init/command.go
```diff
@@ -50,6 +50,8 @@ type Command struct {
 	flagCreateSnapshotAgentToken bool
 
 	flagCreateMeshGatewayToken bool
+	flagIngressGateways        []string
```
lkysow, line 53: What about flagIngressGatewayNames? Looks like we follow the convention of matching these variables with the flags.

adilyse: Updated.

subcommand/server-acl-init/rules.go
```diff
@@ -121,6 +127,42 @@ namespace_prefix "" {
...
+	ingressGatewayRulesTpl := `
+{{- if .EnableNamespaces }}
+namespace "{{ .GatewayNamespace }}" {
+{{- end }}
+  service "{{ .GatewayName }}" {
+     policy = "write"
+  }
+{{- if .EnableNamespaces }}
+}
+{{- end }}
```
lkysow, line 140: I think we need to add
```
node_prefix "" {
  policy = "read"
}
service_prefix "" {
  policy = "read"
}
```

So the ingress can route to other services. It's not crazy permissions since we already have that in the anon token and every envoy sidecar and intentions are still required to actually allow requests to those services.

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +501,108 @@ func (c *Command) Run(args []string) int {
...
+				if len(parts) > 1 {
+					// Name and namespace were provided
+					name = parts[0]
+
+					// Use default namespace if provided flag is of the
+					// form "name."
+					if parts[1] != "" {
+						namespace = parts[1]
+					} else {
+						namespace = "default"
```
lkysow, line 528: Can we pull this out into a constant `DefaultConsulNS`? We're hardcoding this string in a couple places now.

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +501,108 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
+	if len(c.flagIngressGatewayNames) > 0 {
```
lkysow, line 504: What do you think about reducing the duplication here by pulling this code into a generic function for both gateway types:
```go
code := c.createGatewayPolices(
  "ingress-gateway",
  c.flagIngressGatewayNames,
  c.ingressGatewayRules,
  consulDC,
  consulClient)

if code != 0 {
  return code
}

code = c.createGatewayPolices(
  "terminating-gateway",
  c.flagTerminatingGatewayNames,
  c.terminatingGatewayRules,
  consulDC,
  consulClient)

if code != 0 {
  return code
}

// generic func signature
func (c *Command) createGatewayPolices(
	gatewayType string,
	names []string,
	rulesF func(string, string) (string, error),
	consulDC string,
	consulClient *api.Client) int {...}
```

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +501,108 @@ func (c *Command) Run(args []string) int {
...
+		// Create a token for each ingress gateway name. Each gateway needs a
+		// separate token because users may need to attach different policies
+		// to each gateway token depending on what the services it represents
+		for _, name := range c.flagIngressGatewayNames {
```
lkysow, line 508: Since all the code here is within the iteration over `c.flagIngressGatewayNames` we don't need the
```go
if len(c.flagIngressGatewayNames) > 0
```

subcommand/server-acl-init/command.go
```diff
@@ -124,6 +126,14 @@ func (c *Command) init() {
 		"[Enterprise Only] Toggle for creating a token for the Consul snapshot agent deployment.")
 	c.flags.BoolVar(&c.flagCreateMeshGatewayToken, "create-mesh-gateway-token", false,
 		"Toggle for creating a token for a Connect mesh gateway.")
+	c.flags.Var((*flags.AppendSliceValue)(&c.flagIngressGatewayNames), "ingress-gateway-name",
```
lkysow, line 129: Great docs.

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +501,108 @@ func (c *Command) Run(args []string) int {
...
+		// Create a token for each ingress gateway name. Each gateway needs a
+		// separate token because users may need to attach different policies
+		// to each gateway token depending on what the services it represents
+		for _, name := range c.flagIngressGatewayNames {
+			if name == "" {
+				c.log.Error("Ingress gateway names cannot be empty")
+				return 1
+			}
+
+			// Parse optional namespace. This does not protect against a user
```
lkysow, line 514: Did you want to give an error if they have a `.` but namespaces aren't enabled? I'm not sure where it would break in this case later down the line, if it would at all so might be nice to save someone.

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -693,6 +734,172 @@ func TestRun_TokensWithNamespacesEnabled(t *testing.T) {
...
+}`, `
+namespace "namespace2" {
+  service "another-gateway" {
+     policy = "write"
+  }
+}`},
+		},
+	}
+	for _, c := range cases {
+		t.Run(c.TestName, func(t *testing.T) {
```
lkysow, line 849: You might get in trouble here by shadowing `t` from the outer scope.

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +501,108 @@ func (c *Command) Run(args []string) int {
...
+			ingressGatewayRules, err := c.ingressGatewayRules(name, namespace)
+			if err != nil {
+				c.log.Error("Error templating ingress gateway rules", "gateway-name", name,
+					"namespace", namespace, "err", err)
+			}
+
+			// The names in the Helm chart are specified by users and so may not contain
+			// the words "ingress-gateway". We need to create unique names for tokens
+			// across all gateway types and so must suffix with `-ingress-gateway`.
+			tokenName := fmt.Sprintf("%s-ingress-gateway", name)
```
lkysow, line 546: If you make the change to add the prefix in the helm chart this should also have the prefix.
```suggestion
			tokenName := c.withPrefix(fmt.Sprintf("%s-ingress-gateway", name))
```

lkysow: Oops, this is wrong. I see now that withPrefix is called within `createACL`

subcommand/server-acl-init/command.go
adilyse: Added.

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +501,108 @@ func (c *Command) Run(args []string) int {
...
+					namespace = "default"
+				}
+			}
+
+			// Define the gateway rules
+			ingressGatewayRules, err := c.ingressGatewayRules(name, namespace)
+			if err != nil {
+				c.log.Error("Error templating ingress gateway rules", "gateway-name", name,
+					"namespace", namespace, "err", err)
+			}
```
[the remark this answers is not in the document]
adilyse: Updated.

subcommand/server-acl-init/rules.go
```diff
@@ -121,6 +127,48 @@ namespace_prefix "" {
 	return c.renderRules(meshGatewayRulesTpl)
 }
 
+func (c *Command) ingressGatewayRules(name, namespace string) (string, error) {
+	ingressGatewayRulesTpl := `
+{{- if .EnableNamespaces }}
+namespace "{{ .GatewayNamespace }}" {
```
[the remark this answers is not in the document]
adilyse: Given that it's easy to run multiple ingress gateways, I don't think we should support cross-namespace permissions at this point. They're pretty messy anyway and it's better to keep permissions as restricted as possible.

subcommand/server-acl-init/command.go
```diff
@@ -751,3 +754,15 @@ func (c *Command) createAnonymousPolicy() bool {
 			// cross-dc.
 			(c.flagCreateInjectAuthMethod && c.flagCreateACLReplicationToken))
 }
+
+const ConsulDefaultNamespace = "default"
```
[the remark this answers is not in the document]
adilyse: Updated.

subcommand/server-acl-init/command.go
```diff
@@ -491,6 +501,122 @@ func (c *Command) Run(args []string) int {
...
+					if parts[1] != "" {
+						namespace = parts[1]
+					} else {
+						namespace = ConsulDefaultNamespace
+					}
+				} else {
+					// Use the default Consul namespace
+					namespace = ConsulDefaultNamespace
+				}
+			} else {
```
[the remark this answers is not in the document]
adilyse: For me, this format makes it faster to understand the logic behind the code, but I've updated it because that must not be the case for everyone.

subcommand/server-acl-init/command_test.go
```diff
@@ -1516,6 +1516,58 @@ func TestRun_CloudAutoJoin(t *testing.T) {
...
+				"-k8s-namespace=" + ns,
+				"-server-address", strings.Split(testSvr.HTTPAddr, ":")[0],
+				"-server-port", strings.Split(testSvr.HTTPAddr, ":")[1],
+			}
+			cmdArgs = append(cmdArgs, c.flags...)
+			responseCode := cmd.Run(cmdArgs)
+			require.Equal(1, responseCode, ui.ErrorWriter.String())
+		})
+	}
+
```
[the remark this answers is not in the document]
adilyse: Removed.

subcommand/server-acl-init/rules.go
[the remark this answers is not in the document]
lkysow: I think we should support the use case of ingress gateway in default namespace being able to route to services in other consul namespaces but I think this may already work because of the cross-namespace policy we create in the default namespace.

Can you try it out and let me know if it does?

PR #267, opened 2020-05-26
Support resource settings

lkysow:
* Set resource settings for init and lifecycle containers. These are set by default and not overridable. They're set by default to avoid needing a separate flag to turn them on. This is safe because they're set to more than double observed resource requirements so users that don't need resource settings won't be affected. They're not exposed for overriding because they won't change based on use-case, i.e. a service with a lot of throughput may need to customize its envoy proxy cpu/memory however the init container and lifecycle sidecar will need the same cpu/memory as all the other connect pods since they're doing the same job.
* Allow setting resource settings for envoy sidecar via default settings passed in through Helm chart values that will apply to all pods, overridable by annotations on injected pods.
* Flags:
  * `-default-sidecar-proxy-cpu-limit`
  * `-default-sidecar-proxy-cpu-request`
  * `-default-sidecar-proxy-memory-limit`
  * `-default-sidecar-proxy-memory-request`
* Annotations:
  * `consul.hashicorp.com/sidecar-proxy-cpu-limit`
  * `consul.hashicorp.com/sidecar-proxy-cpu-request`
  * `consul.hashicorp.com/sidecar-proxy-memory-limit`
  * `consul.hashicorp.com/sidecar-proxy-memory-request`

See https://github.com/hashicorp/consul-helm/pull/470 for testing.

Refactors @wuvs work from https://github.com/hashicorp/consul-k8s/pull/109.

### DIFF
--- a/connect-inject/envoy_sidecar.go
+++ b/connect-inject/envoy_sidecar.go
@@ -6,17 +6,32 @@ import (
 	"text/template"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 type sidecarContainerCommandData struct {
 	AuthMethod      string
 	ConsulNamespace string
+	Resources corev1.ResourceRequirements
 }
 
 func (h *Handler) envoySidecar(pod *corev1.Pod, k8sNamespace string) (corev1.Container, error) {
 	templateData := sidecarContainerCommandData{
 		AuthMethod:      h.AuthMethod,
 		ConsulNamespace: h.consulNamespace(k8sNamespace),
+	}
+
+	if h.Resources {
+		templateData.Resources = corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse(h.CPULimit),
+				corev1.ResourceMemory: resource.MustParse(h.MemoryLimit),
+			},
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse(h.CPULimit),
+				corev1.ResourceMemory: resource.MustParse(h.MemoryLimit),
+			},
+		}
 	}
 
 	// Render the command
@@ -39,6 +54,7 @@ func (h *Handler) envoySidecar(pod *corev1.Pod, k8sNamespace string) (corev1.Con
 				},
 			},
 		},
+		Resources: templateData.Resources,
 		VolumeMounts: []corev1.VolumeMount{
 			{
 				Name:      volumeName,
--- a/connect-inject/envoy_sidecar.go
+++ b/connect-inject/envoy_sidecar.go
@@ -2,6 +2,7 @@ package connectinject
 
 import (
 	"bytes"
+	"fmt"
 	"strings"
 	"text/template"
 
@@ -12,7 +13,6 @@ import (
 type sidecarContainerCommandData struct {
 	AuthMethod      string
 	ConsulNamespace string
-	Resources corev1.ResourceRequirements
 }
 
 func (h *Handler) envoySidecar(pod *corev1.Pod, k8sNamespace string) (corev1.Container, error) {
@@ -21,24 +21,16 @@ func (h *Handler) envoySidecar(pod *corev1.Pod, k8sNamespace string) (corev1.Con
 		ConsulNamespace: h.consulNamespace(k8sNamespace),
 	}
 
-	if h.Resources {
-		templateData.Resources = corev1.ResourceRequirements{
-			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse(h.CPULimit),
-				corev1.ResourceMemory: resource.MustParse(h.MemoryLimit),
-			},
-			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse(h.CPULimit),
-				corev1.ResourceMemory: resource.MustParse(h.MemoryLimit),
-			},
-		}
-	}
-
 	// Render the command
 	var buf bytes.Buffer
 	tpl := template.Must(template.New("root").Parse(strings.TrimSpace(
 		sidecarPreStopCommandTpl)))
 	err := tpl.Execute(&buf, &templateData)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+
+	resources, err := h.envoySidecarResources(pod)
 	if err != nil {
 		return corev1.Container{}, err
 	}
@@ -54,7 +46,7 @@ func (h *Handler) envoySidecar(pod *corev1.Pod, k8sNamespace string) (corev1.Con
 				},
 			},
 		},
-		Resources: templateData.Resources,
+		Resources: resources,
 		VolumeMounts: []corev1.VolumeMount{
 			{
 				Name:      volumeName,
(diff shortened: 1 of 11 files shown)


subcommand/inject-connect/command.go
```diff
@@ -20,18 +20,12 @@ import (
 	"github.com/hashicorp/consul/command/flags"
 	"github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/cli"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
-type arrayFlags []string
```
lkysow, line 28: Unused

subcommand/inject-connect/command.go
```diff
@@ -131,6 +136,37 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 
+	var cpuLimit, cpuRequest, memoryLimit, memoryRequest resource.Quantity
```
lkysow, line 139: Could move this out into its own function.

connect-inject/envoy_sidecar.go
```diff
@@ -81,6 +89,72 @@ func (h *Handler) envoySidecar(pod *corev1.Pod, k8sNamespace string) (corev1.Con
...
+	// set so that users aren't wondering why it's set to 0 when they didn't specify
+	// a request/limit. If they have explicitly set it to 0 then it will be set
+	// to 0 in the pod spec because we're doing a comparison to the zero-valued
+	// struct.
+
+	// CPU Limit.
+	if anno, ok := pod.Annotations[annotationSidecarProxyCPULimit]; ok {
+		cpuLimit, err := resource.ParseQuantity(anno)
+		if err != nil {
+			return corev1.ResourceRequirements{}, fmt.Errorf("parsing %s annotation: %s", annotationSidecarProxyCPULimit, err)
```
ishustava, line 115: What do you think about also including the value of the annotation in the error messages?

connect-inject/envoy_sidecar_test.go
```diff
@@ -203,3 +204,187 @@ func TestHandlerEnvoySidecar_NamespacesAndAuthMethod(t *testing.T) {
 && /consul/connect-inject/consul logout \
   -token-file="/consul/connect-inject/acl-token"`)
 }
+
+func TestHandlerEnvoySidecar_Resources(t *testing.T) {
+	mem1, err := resource.ParseQuantity("100Mi")
```
ishustava, line 209: This could be `resource.MustParse` that you used in the init container logic. That way we don't need to check for errors.

connect-inject/envoy_sidecar_test.go
```diff
@@ -203,3 +204,187 @@ func TestHandlerEnvoySidecar_NamespacesAndAuthMethod(t *testing.T) {
...
+
+	cases := map[string]struct {
+		handler      Handler
+		annotations  map[string]string
+		expResources corev1.ResourceRequirements
+		expErr       string
+	}{
+		"no defaults, no annotations": {
+			handler:     Handler{},
+			annotations: nil,
```
ishustava, line 228: I think you don't need to set these if they are zero values, unless you wanted to be explicit.

connect-inject/envoy_sidecar_test.go
```diff
@@ -203,3 +204,187 @@ func TestHandlerEnvoySidecar_NamespacesAndAuthMethod(t *testing.T) {
...
+					corev1.ResourceCPU:    cpu2,
+					corev1.ResourceMemory: mem2,
+				},
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:    cpu1,
+					corev1.ResourceMemory: mem1,
+				},
+			},
+		},
+		"defaults set to zero, no annotations": {
```
ishustava, line 296: Do you think this is a valid value we should accept?

subcommand/inject-connect/command.go
```diff
@@ -112,10 +112,15 @@ func (c *Command) init() {
 		"[Enterprise Only] Name of the ACL policy to attach to all created Consul namespaces to allow service "+
 			"discovery across Consul namespaces. Only necessary if ACLs are enabled.")
 
+	// Resource setting flags.
+	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPULimit, "default-sidecar-proxy-cpu-limit", "", "Default sidecar proxy CPU limit")
```
ishustava, line 116:
```suggestion
	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPULimit, "default-sidecar-proxy-cpu-limit", "", "Default sidecar proxy CPU limit.")
```

subcommand/inject-connect/command.go
```diff
@@ -112,10 +112,15 @@ func (c *Command) init() {
...
+	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPURequest, "default-sidecar-proxy-cpu-request", "", "Default sidecar proxy CPU request")
```
ishustava, line 117:
```suggestion
	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPURequest, "default-sidecar-proxy-cpu-request", "", "Default sidecar proxy CPU request.")
```

subcommand/inject-connect/command.go
```diff
@@ -112,10 +112,15 @@ func (c *Command) init() {
...
+	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryLimit, "default-sidecar-proxy-memory-limit", "", "Default sidecar proxy memory limit")
```
ishustava, line 118:
```suggestion
	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryLimit, "default-sidecar-proxy-memory-limit", "", "Default sidecar proxy memory limit.")
```

subcommand/inject-connect/command.go
```diff
@@ -112,10 +112,15 @@ func (c *Command) init() {
...
+	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryRequest, "default-sidecar-proxy-memory-request", "", "Default sidecar proxy memory request")
```
ishustava, line 119:
```suggestion
	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryRequest, "default-sidecar-proxy-memory-request", "", "Default sidecar proxy memory request.")
```

subcommand/inject-connect/command.go
ishustava: I think it's fine.

connect-inject/envoy_sidecar_test.go
lkysow: 🙏

connect-inject/envoy_sidecar_test.go
lkysow: Yeah was trying to be explicit

connect-inject/envoy_sidecar_test.go
lkysow: Yes it's a valid valid that means no limit. I think we should accept it because someone may want to set it explicitly to 0 in an environment that requires it be set (think maybe a admission webhook).

connect-inject/envoy_sidecar.go
lkysow: Good call, changed to `fmt.Errorf("parsing annotation %s:%q: %s", annotationSidecarProxyMemoryRequest, anno, err)`

lkysow:
> One question I had is whether you think we need to validate that `limit >= request` for annotations and defaults. I think this is an invalid configuration and kubernetes will return an error. I'm thinking that at least in the case of defaults, it might be useful to get early feedback before you get to deploying services.

Good call, added.

PR #271, opened 2020-06-03
Add -resolve-hostnames flag

lkysow:
Add a -resolve-hostnames flag to the service-address command that
defaults to false. If true, any service with a hostname as its address
will have that hostname resolved and the first ip written.

If false, (default) the hostname will be written. Consul now supports hostnames for mesh gateway addresses so that's why we're adding this now.

This is useful in the case of EKS where load balancers have hostnames.
For WAN federation we want to use the hostname because the underlying
IPs can change. For ingress gateways this command is used to set the
address field in the service definition which only accepts IPs. This
flag allows the service-address command to be used so the service
registration is valid and then we document that in the EKS case users
shouldn't use the ingress consul DNS and should instead set up their own
DNS that resolves to the load balancer.

### DIFF
--- a/subcommand/service-address/command.go
+++ b/subcommand/service-address/command.go
@@ -27,9 +27,10 @@ type Command struct {
 	flags    *flag.FlagSet
 	k8sFlags *k8sflags.K8SFlags
 
-	flagNamespace   string
-	flagServiceName string
-	flagOutputFile  string
+	flagNamespace        string
+	flagServiceName      string
+	flagOutputFile       string
+	flagResolveHostnames bool
 
 	retryDuration time.Duration
 	k8sClient     kubernetes.Interface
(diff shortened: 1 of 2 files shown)


lkysow:
I've tested this in EKS with Freddie's latest version of Consul that supports hostnames and it works.

lkysow:
So to be more clear, for wan federation this value is used in
```hcl
service {
  tagged_addresses {
    wan {
       address = "<here>"
     }
   }
}
```

But for ingress gateways this value is also used for the actual address of the service so that `<svc>.ingress.consul` resolves to the WAN IP.

```hcl
service {
  address = "<here>"
}
```

In this case, we have to provide a way to resolve to ips since a hostname isn't supported in the address field.

lkysow:
I need to re-test that this is necessary. On Friday I learned we *can* use a hostname for the address field.

lkysow:
After further investigation I've seen that while Consul *will* allow you to set a hostname for the address, DNS resolution will fail so this flag *is* necessary.

PR #272, opened 2020-06-04
Fix bug with defining connect healthcheck

lkysow:
Previously we were setting the alias_service field of alias checks to
the service name instead of the service id. This check would then pass
because Consul would consider an alias for a non-existent service id to
be okay. With https://github.com/hashicorp/consul/pull/7384, now the
check will fail if the service id doesn't exist and so those connect
services will be considered unhealthy and be unroutable.

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -323,7 +323,7 @@ services {
 
   checks {
     name = "Destination Alias"
-    alias_service = "{{ .ServiceName }}"
+    alias_service = "${SERVICE_ID}"
   }
 }
 
--- a/connect-inject/container_init_test.go
+++ b/connect-inject/container_init_test.go
@@ -77,7 +77,7 @@ services {
 
   checks {
     name = "Destination Alias"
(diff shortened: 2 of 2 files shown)


lkysow:
I've tested with consul from master that has https://github.com/hashicorp/consul/pull/7384 merged and this check is now passing correctly. Without this patch all connect services will be failing and unroutable.

lkysow:
`lkysow/consul-k8s-dev:jun04-svc-alias` can be used with `lkysow/consul-jun4` if you want to test

lkysow:
To reproduce:
1. Install consul with `lkysow/consul-jun4` consul image
1. Deploy a connect service
1. Notice that it's failing its alias check
1. If you deploy another service and try to make a call between the two they'll fail.
1. If you dig into the envoy logs at debug level you'll see it's not healthy

Then update consul-k8s image to `lkysow/consul-k8s-dev:jun04-svc-alias` and delete the pods so they redeploy. The check now passes.

PR #273, opened 2020-06-09
Update envoy and consul versions

i0rek:
Consul 1.8 will ship with support for envoy 1.14.2: https://github.com/hashicorp/consul/pull/8057

### DIFF
--- a/connect-inject/handler.go
+++ b/connect-inject/handler.go
@@ -21,8 +21,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


ishustava:
Thanks so much for the PR, Hans! I'm going to close this because we've decided to not default to the latest supported Envoy version to make sure that our consul-k8s docker image is compatible with older versions of consul.

PR #274, opened 2020-06-11
top level build was failing due to missing bin extension for windows

kschoche:
Top level build `make all` was failing due to file extensions for windows.
This change checks if we need to modify binary extension at build time and appends to the binary if necessary.
```
 consul-k8s % make
==> Building Consul - OSes: solaris darwin freebsd linux windows, Architectures: 386 amd64 arm arm64
Building sequentially with go install
--->   solaris/amd64
<snip>
--->   windows/386
cp: /Users/kyle/go/bin/windows_386//consul-k8s: No such file or directory
ERROR: Failed to build Consul for windows/386
make: *** [bin] Error 1
```

### DIFF
--- a/build-support/functions/20-build.sh
+++ b/build-support/functions/20-build.sh
@@ -280,7 +280,13 @@ function build_consul_local {
             then
                GOBIN_EXTRA="${os}_${arch}/"
             fi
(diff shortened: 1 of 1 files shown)


lkysow:
@alvin-huang you may want to pull this into your other scripts

PR #281, opened 2020-06-19
fail if service name != service account name

### LINKED ISSUE #237: First container name might not match service account
When using connect inject with ACLs, if the first container's name doesn't match the service account name then the consul login will fail and the service won't be registered. We should investigate some better logging or error messaging in this case. Or we could use the service account name instead of the first container's name.

ashwin-venkatesh:
This check only occurs when ACLs are enable with connectInject.

Tested that it only performs this check when ACLs are enabled. Screenshot of error when incorrect pod config is applied

Closes #237 

Signed-off-by: Ashwin Venkatesh <ashwin@hashicorp.com>

Screen Shot 2020-06-19 at 10 54 48 AM

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -81,6 +81,12 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
 		panic("No service found. This should be impossible since we default it.")
 	}
 
+	// When ACLs are enabled, the ACL token returned from `consul login` is only
+	// valid for a service with the same name as the ServiceAccountName.
+	if data.AuthMethod != "" && data.ServiceName != pod.Spec.ServiceAccountName {
+		return corev1.Container{}, fmt.Errorf("serviceAccountName %q does not match service name %q", pod.Spec.ServiceAccountName, data.ServiceName)
+	}
+
 	// If a port is specified, then we determine the value of that port
 	// and register that port for the host service.
 	if raw, ok := pod.Annotations[annotationPort]; ok && raw != "" {
--- a/connect-inject/container_init_test.go
+++ b/connect-inject/container_init_test.go
@@ -627,6 +627,7 @@ func TestHandlerContainerInit_namespacesEnabled(t *testing.T) {
 						},
 					},
 				},
+				ServiceAccountName: "web",
 			},
 		}
 	}
@@ -1323,6 +1324,7 @@ func TestHandlerContainerInit_authMethod(t *testing.T) {
 					},
 				},
 			},
+			ServiceAccountName: "foo",
 		},
 	}
 	container, err := h.containerInit(pod, k8sNamespace)
@@ -1373,6 +1375,7 @@ func TestHandlerContainerInit_authMethodAndCentralConfig(t *testing.T) {
 					},
 				},
 			},
+			ServiceAccountName: "foo",
 		},
 	}
 	container, err := h.containerInit(pod, k8sNamespace)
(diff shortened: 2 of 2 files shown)


connect-inject/container_init.go
```diff
@@ -81,6 +82,13 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
 		panic("No service found. This should be impossible since we default it.")
 	}
 
+	// If ACLs are enabled, the ACL token is assumed to be in the ServiceAccount that
+	// shares the name with the Service. If the name of the service does not match that
+	// of the ServiceAccount, the token will not be found leading to an opaque 403 error.
+	if data.AuthMethod != "" && data.ServiceName != pod.Spec.ServiceAccountName {
+		return corev1.Container{}, errors.New(fmt.Sprintf(`serviceAccountName "%s" does not match service name "%s"`, pod.Spec.ServiceAccountName, data.ServiceName))
```
lkysow, line 89:
```suggestion
		return corev1.Container{}, fmt.Errorf("serviceAccountName %q does not match service name %q", pod.Spec.ServiceAccountName, data.ServiceName)
```

I think this can be simplified. `%q` quotes the string.

connect-inject/container_init.go
```diff
@@ -81,6 +82,13 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
 		panic("No service found. This should be impossible since we default it.")
 	}
 
+	// If ACLs are enabled, the ACL token is assumed to be in the ServiceAccount that
+	// shares the name with the Service. If the name of the service does not match that
+	// of the ServiceAccount, the token will not be found leading to an opaque 403 error.
```
lkysow, line 87: Can you reword this to say that the ACL token returned from `consul login` is only valid for a service with the same name as the serviceAccountName. I think the "assumed to be in the ServiceAccount" doesn't fully capture what's going on because the ACL token isn't in the service account. We exchange the service account token for a consul ACL token via `consul login`.

connect-inject/container_init_test.go
```diff
@@ -1514,3 +1517,28 @@ func TestHandlerContainerInit_Resources(t *testing.T) {
...
+					Name: "serviceName",
+				},
+			},
+			ServiceAccountName: "notServiceName",
+		},
+	}
+
+	_, err := h.containerInit(pod, k8sNamespace)
+	require.EqualError(err, `serviceAccountName "notServiceName" does not match service name "foo"`)
+}
```
lkysow, line 1544: Can you add another test for when there is no annotation and we're using the first pod name and also for when ACLs are not enabled and so the names can be mismatched with no error.

ashwin-venkatesh: I thought of writing that test but between https://github.com/hashicorp/consul-k8s/blob/7e9d6501f8431f9b1b7997b4a23341a4435758f7/connect-inject/container_init.go#L79:L83 and https://github.com/hashicorp/consul-k8s/blob/7e9d6501f8431f9b1b7997b4a23341a4435758f7/connect-inject/handler.go#L425:L434, the annotation for `serviceName` is always set and it will panic if the annotation is not.

lkysow: Gotcha, makes sense!

lkysow: Oh, what about a test for when ACLs are not enabled? i.e. `Handler.AuthMethod == ""`.

ashwin-venkatesh: there are other tests up top that would fail and hence i didn’t write one explicitly. i can add one though.

lkysow: Yeah I think an explicit one makes sense here. I looked at the other tests and they don't seem targeted enough to this use-case imo.

ashwin-venkatesh: Done 💥

PR #282, opened 2020-06-23
Always update auth method

ishustava:
Previously, we would only update auth method in Consul
if Consul namespaces are enabled. This was necessary to make sure
that if namespace configuration changes, it is reflected in the auth method.

However, running Consul servers externally is another use case why it's important
to update the auth method. That is because if you run a Helm install, then uninstall
and install again, the auth method saved in Consul is out-of-date:
the service account JWT token saved in the auth method refers to the service account token
that has been deleted from Kubernetes when you ran helm uninstall.

This change proposes to always update the auth method to solve for the
external servers use case.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,8 @@ executors:
       - image: circleci/golang:1.13
     environment:
       - TEST_RESULTS: /tmp/test-results # path to where test results are saved
-      - CONSUL_VERSION: 1.7.0 # Consul's OSS version to use in tests
-      - CONSUL_ENT_VERSION: 1.7.0+ent # Consul's enterprise version to use in tests
+      - CONSUL_VERSION: 1.8.0 # Consul's OSS version to use in tests
+      - CONSUL_ENT_VERSION: 1.8.0+ent # Consul's enterprise version to use in tests
 
 jobs:
   go-fmt-and-vet:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,8 @@ executors:
       - image: circleci/golang:1.13
     environment:
       - TEST_RESULTS: /tmp/test-results # path to where test results are saved
-      - CONSUL_VERSION: 1.8.0 # Consul's OSS version to use in tests
-      - CONSUL_ENT_VERSION: 1.8.0+ent # Consul's enterprise version to use in tests
+      - CONSUL_VERSION: 1.7.0 # Consul's OSS version to use in tests
+      - CONSUL_ENT_VERSION: 1.7.0+ent # Consul's enterprise version to use in tests
 
 jobs:
   go-fmt-and-vet:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -6,8 +6,8 @@ executors:
       - image: circleci/golang:1.13
     environment:
       - TEST_RESULTS: /tmp/test-results # path to where test results are saved
-      - CONSUL_VERSION: 1.7.0 # Consul's OSS version to use in tests
-      - CONSUL_ENT_VERSION: 1.7.0+ent # Consul's enterprise version to use in tests
+      - CONSUL_VERSION: 1.8.0 # Consul's OSS version to use in tests
+      - CONSUL_ENT_VERSION: 1.8.0+ent # Consul's enterprise version to use in tests
```
ishustava, line 10: This is unrelated, but I thought it'd be good to run the latest.

subcommand/server-acl-init/command_test.go
```diff
@@ -680,9 +680,8 @@ func TestRun_ConnectInjectAuthMethod(t *testing.T) {
 	t.Parallel()
 
 	cases := map[string]struct {
-		flags          []string
-		expectedHost   string
-		expectedCACert string
```
ishustava, line 685: this was unused

lkysow:
This one took me a bit to grok. For other reviewers, what's happening is when running with external servers, e.g. Consul on Azure, when you helm install for the first time, we create a Consul auth method that Consul uses to exchange service account tokens for consul acl tokens. The auth method works by Consul servers calling the Kubernetes API and checking if the service account is valid. In order for consul to call the kube api, it itself needs a service account token. On first install, this service account token is stored in the Consul servers.

When the consul installation is deleted, we don't delete the auth method from the consul servers since we're just uninstalling the consul clients in kube. However we *do* delete the service account resource from Kubernetes. This means the consul servers now have an auth method that's trying to use a token that was deleted from Kubernetes.

On re-install, prior to this PR, we would recognize that the auth method exists and not update it. It would then still have the now-deleted service account token and so any `consul login` commands (used by all connect services) would fail because when consul tried to make the API call to kube to check if the service account was valid, kube would reject that call.

subcommand/server-acl-init/connect_inject.go
```diff
@@ -9,83 +9,58 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const defaultKubernetesHost = "https://kubernetes.default.svc"
```
lkysow, line 12: Can you document this const 🙏

subcommand/server-acl-init/command_test.go
lkysow: 🍺

subcommand/server-acl-init/connect_inject.go
ishustava: Done!

ishustava:
> lmk if you'd like me to.

thanks! I think it's OK for just myself to do the testing. I've tested this with HCS, and it works nicely!

PR #284, opened 2020-06-26
Fix tests

ishustava:
Depends on hashicorp/consul#8192

This PR makes a couple of changes:

1. Uses `master` branch for the `hashicorp/consul/sdk` dependency.
1. Fix `server-acl-init` federation tests. Now that consul always enforces ACLs, we have to make sure that we have a valid token when calling the API. Prior to that we called secondary's `agent/join` API with a token that it had no way of verifying (more details in the comment [here](https://github.com/hashicorp/consul-k8s/blob/50be8cc1a974ac0f404c3b72e4831757ecbd2b6f/subcommand/server-acl-init/command_test.go#L1772-L1777)).
1. Standardize on `retry.Run` timeouts in lifecycle-sidecar tests. Sometimes tests were using too short of a timeout compared to other tests. I switched it to just use the default timer.
1. Fix lifecycle-sidecar tests that were leaving running consul agents behind.
1. Bumps Consul version in CI


subcommand/server-acl-init/command_test.go
```diff
@@ -1768,14 +1769,28 @@ func replicatedSetup(t *testing.T, bootToken string) (*fake.Clientset, *api.Clie
 	if bootToken == "" {
 		clientToken = aclReplicationToken
 	}
+
+	// Create a consul client pointing to the primary server.
+	// Note: We need to use the primary server to make the WAN join API call
+	// because the secondary  will not be able to verify this token
```
lkysow, line 1775:
```suggestion
	// because the secondary will not be able to verify this token
```

PR #288, opened 2020-07-06
increase cpu request/limits for lifecycle sidecar

kschoche:
With the current CPU resource limits set to 10m we're noticing high cpu usage in the lifecycle sidecar.
Setting this value to 20m instead shows more stable usage during testing.

### DIFF
--- a/connect-inject/lifecycle_sidecar.go
+++ b/connect-inject/lifecycle_sidecar.go
@@ -8,8 +8,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


ashwin-venkatesh:
Will these constants go away in favor of the flags that will be provided to the container?

kschoche:
This will be brought in via a new PR which integrates the init copy container OOM fixes with the lifecycle sidecar changes.

PR #292, opened 2020-07-09
Update changelog

lkysow:


CHANGELOG.md
```diff
@@ -3,6 +3,18 @@
...
+IMPROVEMENTS:
+
+* Extracted Consul's HTTP flags into our own package so we no longer depend on the internal Consul golang module. [[GH-259](https://github.com/hashicorp/consul-k8s/pull/259)]
+
+BUG FIXES:
+
+* Connect: Update resource settings to fix out of memory errors and CPU usage at 100% of limit. [Fixes [GH-283](https://github.com/hashicorp/consul-k8s/issues/283), [consul-helm GH-515](https://github.com/hashicorp/consul-helm/issues/515)]
+* Connect: Creating a pod with a different service account name than its Consul service name will now result in an error when ACLs are enabled.
+  Previously this would not result in an error, but the pod would not be able to send or receive traffic because its ACL token would be for a
+  different service name. [Fixes [GH-237](https://github.com/hashicorp/consul-k8s/issues/237)]
```
ishustava, line 17: Could we refer to the #281 PR instead? It also references the issue. I don't think we've used "Fixes" keywords in the changelog in the past, probably best to stay consistent.

CHANGELOG.md
```diff
@@ -3,6 +3,18 @@
...
 * ACLs: Always update Kubernetes auth method created by the `server-acl-init` job. Previously, we would only update the auth method if Consul namespaces are enabled. With this change, we always update it to make sure that any configuration changes or updates to the `connect-injector-authmethod-svc-account` are propagated [[GH-282](https://github.com/hashicorp/consul-k8s/pull/282)].
+* Connect: Connect pods have had the following resource settings changed: `consul-connect-inject-init` now has its memory limit set to `150M` up from `25M` and `consul-connect-lifecycle-sidecar` has its CPU request and limit set to `20m` up from `10m`. [[GH-291](https://github.com/hashicorp/consul-k8s/pull/291)]
+
+IMPROVEMENTS:
+
+* Extracted Consul's HTTP flags into our own package so we no longer depend on the internal Consul golang module. [[GH-259](https://github.com/hashicorp/consul-k8s/pull/259)]
+
+BUG FIXES:
+
+* Connect: Update resource settings to fix out of memory errors and CPU usage at 100% of limit. [Fixes [GH-283](https://github.com/hashicorp/consul-k8s/issues/283), [consul-helm GH-515](https://github.com/hashicorp/consul-helm/issues/515)]
```
ishustava, line 14: Is this change different from a breaking change mentioned above? Do you think we should mention it twice or just say in the above change that it fixes the CPU and OOM issues?

CHANGELOG.md
lkysow: Oops this is how I do it with Atlantis 🤦 , forgot we don't.

For linking to the issue, it looks like for bugfixes we usually link to the issue being fixed unless there was no issue creatd in which case we link to the pull request.

CHANGELOG.md
lkysow: It's the same change.

I think listing it twice is helpful because we can give more information in each context. If I'm experiencing a bug with CPU usage and I look at our changelog to see if it's been fixed I won't be looking in the breaking changes section, I'll be looking for the bugfix section.

Once I've found the release with the bugfix I need, I'll read the breaking changes while upgrading.

CHANGELOG.md
ishustava: Ah got it! I thought the PR usually will link to the issue anyway, but what you said makes sense to me!

CHANGELOG.md
ishustava: Sounds good, thanks for clarifying!

PR #298, opened 2020-07-21
Allow init and lifecycle containers resource limits to be configured

adilyse:
Updates the work done in #289.

Changes proposed in this PR:
- Adds resource request and limit flags for both the injected init and lifecycle sidecar containers

How I've tested this PR:
Added and updated tests for the functionality. Changed default values to unique values and deployed to a cluster, verifying that they were applied and in the correct place.

Verify the tests function as expected and include all the relevant cases. Deploy the connect injector using the new flags and verify that any injected pods include the provided values.

Checklist:
- [ x ] Tests added
- [ x ] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+
+* Connect: Add resource request and limit flags for the injected init and lifecycle sidecar containers [[GH-298](https://github.com/hashicorp/consul-k8s/pull/298)].
+
 BUG FIXES:
 
 * Connect: Respect allow/deny list flags when namespaces are disabled. [[GH-296](https://github.com/hashicorp/consul-k8s/issues/296)]
--- a/connect-inject/handler.go
+++ b/connect-inject/handler.go
@@ -78,6 +78,7 @@ const (
 	// service is synced (i.e. re-registered) with the local agent.
 	annotationSyncPeriod = "consul.hashicorp.com/connect-sync-period"
 
+	// annotations for sidecar proxy resource limits
 	annotationSidecarProxyCPULimit      = "consul.hashicorp.com/sidecar-proxy-cpu-limit"
 	annotationSidecarProxyCPURequest    = "consul.hashicorp.com/sidecar-proxy-cpu-request"
 	annotationSidecarProxyMemoryLimit   = "consul.hashicorp.com/sidecar-proxy-memory-limit"
@@ -169,11 +170,20 @@ type Handler struct {
 	// Only necessary if ACLs are enabled.
 	CrossNamespaceACLPolicy string
 
-	// Default resource settings for sidecar proxies.
+	// Default resource settings for sidecar proxies. Some of these
+	// fields may be empty.
 	DefaultProxyCPURequest    resource.Quantity
 	DefaultProxyCPULimit      resource.Quantity
 	DefaultProxyMemoryRequest resource.Quantity
 	DefaultProxyMemoryLimit   resource.Quantity
+
+	// Resource settings for init container. All of these fields
+	// will be populated by the defaults provided in the initial flags.
+	InitContainerResources corev1.ResourceRequirements
+
+	// Resource settings for lifecycle sidecar. All of these fields
+	// will be populated by the defaults provided in the initial flags.
+	LifecycleSidecarResources corev1.ResourceRequirements
 
 	// Log
 	Log hclog.Logger
--- a/subcommand/inject-connect/command.go
+++ b/subcommand/inject-connect/command.go
@@ -20,6 +20,7 @@ import (
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/cli"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -57,6 +58,18 @@ type Command struct {
 	flagDefaultSidecarProxyCPURequest    string
 	flagDefaultSidecarProxyMemoryLimit   string
 	flagDefaultSidecarProxyMemoryRequest string
+
+	// Lifecycle sidecar resource settings.
+	flagLifecycleSidecarCPULimit      string
+	flagLifecycleSidecarCPURequest    string
+	flagLifecycleSidecarMemoryLimit   string
+	flagLifecycleSidecarMemoryRequest string
+
+	// Init container resource settings.
+	flagInitContainerCPULimit      string
+	flagInitContainerCPURequest    string
+	flagInitContainerMemoryLimit   string
+	flagInitContainerMemoryRequest string
 
 	flagSet *flag.FlagSet
 	http    *flags.HTTPFlags
@@ -112,11 +125,23 @@ func (c *Command) init() {
 		"[Enterprise Only] Name of the ACL policy to attach to all created Consul namespaces to allow service "+
 			"discovery across Consul namespaces. Only necessary if ACLs are enabled.")
 
-	// Resource setting flags.
-	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPULimit, "default-sidecar-proxy-cpu-limit", "", "Default sidecar proxy CPU limit.")
+	// Proxy sidecar resource setting flags.
 	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPURequest, "default-sidecar-proxy-cpu-request", "", "Default sidecar proxy CPU request.")
-	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryLimit, "default-sidecar-proxy-memory-limit", "", "Default sidecar proxy memory limit.")
+	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPULimit, "default-sidecar-proxy-cpu-limit", "", "Default sidecar proxy CPU limit.")
 	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryRequest, "default-sidecar-proxy-memory-request", "", "Default sidecar proxy memory request.")
+	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryLimit, "default-sidecar-proxy-memory-limit", "", "Default sidecar proxy memory limit.")
+
+	// Init container resource setting flags.
+	c.flagSet.StringVar(&c.flagInitContainerCPURequest, "init-container-cpu-request", "50m", "Init container CPU request.")
+	c.flagSet.StringVar(&c.flagInitContainerCPULimit, "init-container-cpu-limit", "50m", "Init container CPU limit.")
+	c.flagSet.StringVar(&c.flagInitContainerMemoryRequest, "init-container-memory-request", "25Mi", "Init container memory request.")
+	c.flagSet.StringVar(&c.flagInitContainerMemoryLimit, "init-container-memory-limit", "150Mi", "Init container memory limit.")
+
+	// Lifecycle sidecar resource setting flags.
+	c.flagSet.StringVar(&c.flagLifecycleSidecarCPURequest, "lifecycle-sidecar-cpu-request", "20m", "Lifecycle sidecar CPU request.")
+	c.flagSet.StringVar(&c.flagLifecycleSidecarCPULimit, "lifecycle-sidecar-cpu-limit", "20m", "Lifecycle sidecar CPU limit.")
+	c.flagSet.StringVar(&c.flagLifecycleSidecarMemoryRequest, "lifecycle-sidecar-memory-request", "25Mi", "Lifecycle sidecar memory request.")
+	c.flagSet.StringVar(&c.flagLifecycleSidecarMemoryLimit, "lifecycle-sidecar-memory-limit", "25Mi", "Lifecycle sidecar memory limit.")
 
 	c.http = &flags.HTTPFlags{}
 	flags.Merge(c.flagSet, c.http.Flags())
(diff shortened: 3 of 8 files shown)


connect-inject/handler.go
```diff
@@ -169,12 +170,21 @@ type Handler struct {
...
 
-	// Default resource settings for sidecar proxies.
+	// Default resource settings for sidecar proxies. Some of these
+	// fields may be empty.
 	DefaultProxyCPURequest    resource.Quantity
 	DefaultProxyCPULimit      resource.Quantity
 	DefaultProxyMemoryRequest resource.Quantity
 	DefaultProxyMemoryLimit   resource.Quantity
 
+	// Resource settings for Init Container. All of these fields
```
ishustava, line 180:
```suggestion
	// Resource settings for init container. All of these fields
```

subcommand/inject-connect/command.go
```diff
@@ -58,6 +59,18 @@ type Command struct {
 	flagDefaultSidecarProxyMemoryLimit   string
 	flagDefaultSidecarProxyMemoryRequest string
 
+	// Lifecycle sidecar resource settings.
+	flagLifecycleSidecarCPULimit      string
+	flagLifecycleSidecarCPURequest    string
+	flagLifecycleSidecarMemoryLimit   string
+	flagLifecycleSidecarMemoryRequest string
+
+	// Init copy container resource settings.
```
ishustava, line 68:
```suggestion
	// Init container resource settings.
```

subcommand/inject-connect/command.go
```diff
@@ -113,10 +126,22 @@ func (c *Command) init() {
 			"discovery across Consul namespaces. Only necessary if ACLs are enabled.")
 
 	// Resource setting flags.
```
ishustava, line 128:
```suggestion
	// Sidecar resource setting flags.
```

subcommand/inject-connect/command.go
```diff
@@ -113,10 +126,22 @@ func (c *Command) init() {
...
-	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPULimit, "default-sidecar-proxy-cpu-limit", "", "Default sidecar proxy CPU limit.")
 	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPURequest, "default-sidecar-proxy-cpu-request", "", "Default sidecar proxy CPU request.")
-	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryLimit, "default-sidecar-proxy-memory-limit", "", "Default sidecar proxy memory limit.")
+	c.flagSet.StringVar(&c.flagDefaultSidecarProxyCPULimit, "default-sidecar-proxy-cpu-limit", "", "Default sidecar proxy CPU limit.")
 	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryRequest, "default-sidecar-proxy-memory-request", "", "Default sidecar proxy memory request.")
+	c.flagSet.StringVar(&c.flagDefaultSidecarProxyMemoryLimit, "default-sidecar-proxy-memory-limit", "", "Default sidecar proxy memory limit.")
+
+	// Init container resource setting flags.
+	c.flagSet.StringVar(&c.flagInitContainerCPURequest, "init-container-cpu-request", "50m", "Init copy container CPU request.")
```
ishustava, line 135:
```suggestion
	c.flagSet.StringVar(&c.flagInitContainerCPURequest, "init-container-cpu-request", "50m", "Init container CPU request.")
```

subcommand/inject-connect/command.go
```diff
@@ -113,10 +126,22 @@ func (c *Command) init() {
...
+	c.flagSet.StringVar(&c.flagInitContainerCPULimit, "init-container-cpu-limit", "50m", "Init copy container CPU limit.")
```
ishustava, line 136:
```suggestion
	c.flagSet.StringVar(&c.flagInitContainerCPULimit, "init-container-cpu-limit", "50m", "Init container CPU limit.")
```

subcommand/inject-connect/command.go
```diff
@@ -113,10 +126,22 @@ func (c *Command) init() {
...
+	c.flagSet.StringVar(&c.flagInitContainerMemoryRequest, "init-container-memory-request", "25Mi", "Init copy container memory request.")
```
ishustava, line 137:
```suggestion
	c.flagSet.StringVar(&c.flagInitContainerMemoryRequest, "init-container-memory-request", "25Mi", "Init container memory request.")
```

subcommand/inject-connect/command.go
```diff
@@ -113,10 +126,22 @@ func (c *Command) init() {
...
+	c.flagSet.StringVar(&c.flagInitContainerMemoryLimit, "init-container-memory-limit", "150Mi", "Init copy container memory limit.")
```
ishustava, line 138:
```suggestion
	c.flagSet.StringVar(&c.flagInitContainerMemoryLimit, "init-container-memory-limit", "150Mi", "Init container memory limit.")
```

subcommand/inject-connect/command.go
```diff
@@ -135,50 +160,58 @@ func (c *Command) Run(args []string) int {
...
-	if memoryLimit.Value() != 0 && memoryRequest.Cmp(memoryLimit) > 0 {
+	if sidecarProxyMemoryLimit.Value() != 0 && sidecarProxyMemoryRequest.Cmp(sidecarProxyMemoryLimit) > 0 {
 		c.UI.Error(fmt.Sprintf(
 			"request must be <= limit: -default-sidecar-proxy-memory-request value of %q is greater than the -default-sidecar-proxy-memory-limit value of %q",
 			c.flagDefaultSidecarProxyMemoryRequest, c.flagDefaultSidecarProxyMemoryLimit))
 		return 1
 	}
 
+	// Validate resource request/limit flags and parse into corev1.ResourceRequirements
+	initResources, lifecycleResources, err := c.parseAndValidateResourceFlags()
```
ishustava, line 209: I think sidecar resources should also be validated in this function.

connect-inject/handler.go
```diff
@@ -169,12 +170,21 @@ type Handler struct {
 	// Only necessary if ACLs are enabled.
 	CrossNamespaceACLPolicy string
 
-	// Default resource settings for sidecar proxies.
+	// Default resource settings for sidecar proxies. Some of these
+	// fields may be empty.
 	DefaultProxyCPURequest    resource.Quantity
 	DefaultProxyCPULimit      resource.Quantity
 	DefaultProxyMemoryRequest resource.Quantity
 	DefaultProxyMemoryLimit   resource.Quantity
```
ishustava, line 178: I think it'd be nice to make these into `corev1.ResourceRequirements` for consistency.

subcommand/inject-connect/command.go
```diff
@@ -363,6 +398,98 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 	}
 }
 
+func (c *Command) parseAndValidateResourceFlags() (corev1.ResourceRequirements, corev1.ResourceRequirements, error) {
+	// Init container
+	var initContainerCPULimit, initContainerCPURequest, initContainerMemoryLimit, initContainerMemoryRequest resource.Quantity
```
ishustava, line 403: There is a lot of code repetition between this function and the code in `Run` that validates default sidecar resources. Another refactoring idea is to make another function that will parse and validate resources for a single "entity" and then call it for each of the init container, lifecycle sidecar, and potentially for the sidecar proxy as well.

It could look something like this:
```go
func validateAndParseResource(flagPrefix, cpuRequest, cpuLimit, memoryRequest, memoryLimit string) (corev1.ResourceRequirements, error) {
	var containerCPULimit, containerCPURequest, containerMemoryLimit, containerMemoryRequest resource.Quantity

	// Parse and validate the initContainer resources.
	containerCPURequest, err := resource.ParseQuantity(cpuRequest)
	if err != nil {
		return corev1.ResourceRequirements{}, fmt.Errorf("%s-cpu-request is invalid: %s", flagPrefix, err)
	}
	containerCPULimit, err = resource.ParseQuantity(cpuLimit)
	if err != nil {
		return corev1.ResourceRequirements{}, fmt.Errorf("%s-cpu-limit is invalid: %s", flagPrefix, err)
	}
	if containerCPULimit.Value() != 0 && containerCPURequest.Cmp(containerCPULimit) > 0 {
		return corev1.ResourceRequirements{}, fmt.Errorf(
			"request must be <= limit: %s-cpu-request value of %q is greater than the %s-cpu-limit value of %q",
			flagPrefix, cpuRequest, flagPrefix, cpuLimit)
	}

	containerMemoryRequest, err = resource.ParseQuantity(memoryRequest)
	if err != nil {
		return corev1.ResourceRequirements{}, fmt.Errorf("%s-memory-request is invalid: %s", flagPrefix, err)
	}
	containerMemoryLimit, err = resource.ParseQuantity(memoryLimit)
	if err != nil {
		return corev1.ResourceRequirements{}, fmt.Errorf("%s-memory-limit is invalid: %s", flagPrefix, err)
	}
	if containerMemoryLimit.Value() != 0 && containerMemoryRequest.Cmp(containerMemoryLimit) > 0 {
		return corev1.ResourceRequirements{}, fmt.Errorf(
			"request must be <= limit: %s-memory-request value of %q is greater than the %s-memory-limit value of %q",
			flagPrefix, memoryRequest, flagPrefix, memoryLimit)
	}

	// Put into corev1.ResourceRequirements form
	return corev1.ResourceRequirements{
		Requests: corev1.ResourceList{
			corev1.ResourceCPU:    containerCPURequest,
			corev1.ResourceMemory: containerMemoryRequest,
		},
		Limits: corev1.ResourceList{
			corev1.ResourceCPU:    containerCPULimit,
			corev1.ResourceMemory: containerMemoryLimit,
		},
	}, nil
}
```
Then in this function, you'd just call:
```go
initResources, err := validateAndParseResource("-init-container", c.flagInitContainerCPURequest, c.flagInitContainerCPULimit, c.flagInitContainerMemoryRequest, c.flagInitContainerMemoryLimit)
```

ishustava: This could potentially be improved even more because there's also duplication of validations between CPU and memory resources that could probably be moved into another function as well.

subcommand/inject-connect/command_test.go
```diff
@@ -52,6 +52,66 @@ func TestRun_FlagValidation(t *testing.T) {
...
+			expErr: "request must be <= limit: -lifecycle-sidecar-memory-request value of \"50Mi\" is greater than the -lifecycle-sidecar-memory-limit value of \"25Mi\"",
+		},
+		{
+			flags: []string{"-consul-k8s-image", "foo",
+				"-lifecycle-sidecar-cpu-request=50m",
+				"-lifecycle-sidecar-cpu-limit=25m",
+			},
+			expErr: "request must be <= limit: -lifecycle-sidecar-cpu-request value of \"50m\" is greater than the -lifecycle-sidecar-cpu-limit value of \"25m\"",
+		},
 	}
```
ishustava, line 115: should we also add a test to check the defaults? Normally, we don't test default flags values, but I feel like in this case these values are very specific, so it warrants a test. What do you think?

ishustava: Should we also add a test that checks that the init container memory limit is greater than the size of the consul binary installed on your path? This might be better to do in the enterprise test since the enterprise binary is bigger than the OSS.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+
+* Connect: Add resource request and limit flags for the injected init container and lifecycle sidecar.
```
ishustava, line 5:
```suggestion
* Connect: Add resource request and limit flags for the injected init and lifecycle sidecar containers [[GH-298](https://github.com/hashicorp/consul-k8s/pull/298)].
```

ishustava: Should we also mention what their defaults are and why?

subcommand/inject-connect/command.go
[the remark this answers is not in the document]
ishustava: FWIW I suggested renaming it from `init-copy-container` in #289 because this container does a lot more stuff than just copy the consul binary. I don't mind we name it something more specific though.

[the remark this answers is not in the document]
lkysow: Yeah I guess a more specific name is useful to future-proof ourselves against adding a new init container but we could also just call this init-container and if we do add a new init container in the future use a more specific name for that one's resource settings.

connect-inject/handler.go
adilyse: I looked at this, but it gets complicated because there are annotations associated with these settings and they're allowed to be empty. This is significantly different behavior to the init and lifecycle resource flags so it is reasonable to treat them differently. Also, given that it is not a trivial change, it is scope creep because it's not critical to the addition of init and lifecycle resource flags.

subcommand/inject-connect/command.go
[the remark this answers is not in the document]
adilyse: See this [comment](https://github.com/hashicorp/consul-k8s/pull/298/files#r458969645). It's okay that we treat them differently, they have different behavior.

adilyse: See this [comment](https://github.com/hashicorp/consul-k8s/pull/298/files#r458969645). If you feel strongly that the validation should happen in a function, I can put the proxy validation of flags in its own function.

subcommand/inject-connect/command.go
adilyse: There are two repeats in this function. The "don't repeat yourself" rule of thumb generally is suggested to start at 3 since it's over-optimization before then. This suggestion also decrease the readability of the code and makes it more fragile.

subcommand/inject-connect/command_test.go
adilyse: I took a look at adding a default test and didn't come across a straightforward approach to the problem. If we want to fully exercise the flags to get at the defaults, it needs to run at the command level. The `Run` command, however, only outputs a return value and/or error which isn't the information we need to verify these defaults. I'm open to suggestions if I've missed something.

adilyse: A test checking the size of the consul binary would be nondeterministic, which is an undesirable trait in a unit test. The tests are set up to run against the consul binary that's currently in the path, which will vary from developer to developer and over time. Adding this sort of thing to the pipeline could make sense, but doesn't belong as a unit test.

subcommand/inject-connect/command.go
ishustava: I don't think that comment addresses why sidecar resources flag validation should not be refactored into that function.

subcommand/inject-connect/command.go
ishustava: Per my comment above, there is also sidecar resource validation which is almost exactly the same and it lives in the `Run` function, which makes it 3 repeats.

Why are you saying it'll make the code less readable and fragile? I found it pretty hard to read as it's written right now which why I suggested a refactor.

adilyse: I understand that there is some logic sharing. There are also subtle differences in behavior between the new flags and the previous proxy flags, specifically in the defaulting behavior. I think it's a reasonable suggestion to refactor this but I don't think it should be part of this PR. If you would like me to log an issue for this, let me know and I will do so.

subcommand/inject-connect/command.go
adilyse: See this [comment](https://github.com/hashicorp/consul-k8s/pull/298/files#r459010878).

ishustava:
As I was testing this out, I had a question about how would one disable/remove resources for init and lifecycle containers. I think in a Helm/yaml world you would just set it to an empty hash, but how would we translate that to consul-k8s flags? I couldn't find a way to set these flags in a way that would essentially remove them. I could be missing something.

I'll address the rest of the comments and finish my testing tomorrow.

connect-inject/handler.go
kschoche: I agree!
It gets a little messy dealing with ones that are not specified though :-/

subcommand/inject-connect/command_test.go
kschoche: ^ Agreed, adding to this the consul binary varies in size by oss/ent and also by runtime architecture quite dramatically, some are <100M some nearly 130M, so the test may run and pass in our CI but fail in an e.g. darwin64 environment.

subcommand/inject-connect/command.go
lkysow: After looking at this some more, I'd like to see this refactor happen.

To preserve the ability of not setting the sidecar proxy resources we could add an additional `allowEmpty bool` argument to the suggested `validateAndParseResource` function. This could be used like:

```go
	if !allowEmpty || cpuRequest != "" {
		// Parse and validate the initContainer resources.
		containerCPURequest, err = resource.ParseQuantity(cpuRequest)
		if err != nil {
			return corev1.ResourceRequirements{}, fmt.Errorf("%s-cpu-request is invalid: %s", flagPrefix, err)
		}
	}
```

Alternatively we could default those flags to "0". This isn't a breaking change because there's no difference between 0 and not being set. There's a long comment somewhere why I did it like this but in the interest of less complexity and conformity I don't think it's necessary to keep that if it makes the implementation too complex.

subcommand/inject-connect/command_test.go
```diff
@@ -52,6 +52,66 @@ func TestRun_FlagValidation(t *testing.T) {
 			},
 			expErr: "request must be <= limit: -default-sidecar-proxy-cpu-request value of \"50m\" is greater than the -default-sidecar-proxy-cpu-limit value of \"25m\"",
 		},
+		{
+			flags:  []string{"-consul-k8s-image", "foo", "-init-container-cpu-limit=unparseable"},
+			expErr: "-init-container-cpu-limit is invalid",
```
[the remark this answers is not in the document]
lkysow: Looks like the current error is
```
-init-container-cpu-limit is invalid: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
```

I think we should add the value passed in as part of the error message so it's easy for users to find where they've set the invalid value, e.g.
```
-init-container-cpu-limit "unparseable" is invalid: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
```

subcommand/inject-connect/command_test.go
lkysow: Yeah don't see how we can test the defaults here.

I think we should (in a separate PR) add a circle ci job that checks the size based on the current image. Not sure if that makes more sense in consul-k8s or consul-helm.

ishustava: About testing defaults, we can do something like this:

```go
func TestFlagDefaults(t *testing.T) {
	cmd := Command{}
	cmd.init()

	require.Equal(t, cmd.flagInitContainerCPULimit, "50m")
}
```

I've run it locally, and it works. 

> A test checking the size of the consul binary would be nondeterministic, which is an undesirable trait in a unit test. 

Yeah, I agree. Running it in CI seems more appropriate.

ishustava:
I've finished my testing; everything looked great! I've resolved the issue from my comment above with @lkysow 's suggestion to set flags to `0`, and that worked. 

I don't have any other feedback, other than 👍 to Luke's refactoring comments that if we think this is a good refactor, we should address it here.

CHANGELOG.md
adilyse: Updated. No, I don't think listing the new flags is necessary here. The flags are clearly laid out in the PR and descriptions available through the command's help text.

The pull request didn't exist so I couldn't link it when I wrote this up. It's pretty circular to expect it as part of the PR submission.

connect-inject/handler.go
adilyse: Updated.

subcommand/inject-connect/command.go
adilyse: Updated.

subcommand/inject-connect/command.go
adilyse: We have two sidecars so this isn't deterministic. Updated to `Proxy sidecar resource setting flags.`

subcommand/inject-connect/command.go
adilyse: Updated.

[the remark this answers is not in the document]
adilyse: Removed `copy` from flag descriptions.

subcommand/inject-connect/command.go
adilyse: Updated.

subcommand/inject-connect/command.go
adilyse: Updated.

subcommand/inject-connect/command.go
adilyse: Updated.

subcommand/inject-connect/command.go
```diff
@@ -363,6 +398,98 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 	}
 }
 
+func (c *Command) parseAndValidateResourceFlags() (corev1.ResourceRequirements, corev1.ResourceRequirements, error) {
```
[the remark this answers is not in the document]
adilyse: "A `return` statement without arguments returns the named return values. This is known as a "naked" return. **Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions.** " from [A Tour of Go](https://tour.golang.org/basics/7)

Leaving it as-is until it becomes a problem.

subcommand/inject-connect/command_test.go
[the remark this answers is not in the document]
adilyse: Updated.

subcommand/inject-connect/command_test.go
adilyse: Added default test.

subcommand/inject-connect/command.go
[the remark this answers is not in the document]
lkysow: Oh interesting, good to know. I guess the other pattern of a lambda return can work too.

PR #311, opened 2020-08-19
Remove unnecessary dep on prometheus logging

kschoche:
Changes proposed in this PR:
- i noticed an unnecessary dependency on prometheus logging in lifecycle sidecar so removed it
- updated go.mod as well and did a tidy 
This has an added effect of reducing the size of the consul-k8s binary by 1MB :-)

How I've tested this PR:
`make clean && make` 

That it compiles.

Checklist:
- [ ] Tests added
^ Both unnecessary, just a cleanup


subcommand/lifecycle-sidecar/command.go
```diff
@@ -112,7 +111,7 @@ func (c *Command) Run(args []string) int {
 		case <-time.After(c.flagSyncPeriod):
 			continue
 		case <-c.sigCh:
-			log.Info("SIGINT received, shutting down")
+			logger.Info("SIGINT received, shutting down")
```
lkysow, line 114: omg good find!

PR #316, opened 2020-08-31
Provision webhooks with self-generated certs

ashwin-venkatesh:
Changes proposed in this PR:
- Use cert helpers to generate certs for the webhooks
- Remove dependency on cert-manager

How I've tested this PR:
This was tested in conjunction with this helm PR: https://github.com/hashicorp/consul-helm/pull/590
I updated the cert expiry to be 3 minutes from the default 24 hours to observe cert rotation. I then deployed the image with the controller enabled and with 3 replicas.
I observed the webhooks working successfully and no X509 errors were observed. The controller logs were able to successfully update their certs in unison and the were always able to successfully communicate with the webhooks.

Use `ashwinvenkatesh/consul-k8s:webhook-certs` for the imageK8s in the helm chart. enable controllers and allow multiple replicas. 
Additionally, if `cert-manager` is installed in your cluster, `kubectl delete` the installation.
You should be able to successfully helm install it and create service defaults.

Checklist:
- [x] Tests added


subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,229 @@
...
+	c.flagSet.StringVar(&c.flagKeyFile, "tls-key-file", "",
+		"PEM-encoded TLS private key to serve. If blank, will generate random cert.")
+
+	c.httpFlags = &flags.HTTPFlags{}
+	flags.Merge(c.flagSet, c.httpFlags.Flags())
+	c.help = flags.Usage(help, c.flagSet)
+}
+
+func (c *Command) Run(args []string) int {
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
```
ashwin-venkatesh, line 71: ignore this for now. This will use the UI logger.

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,229 @@
...
+		if err != nil {
+			certGenLog.Error(err, "Error creating K8S client")
+			return 1
+		}
+	}
+
+	var certSource cert.Source = &cert.GenSource{
+		Name:   "Consul Webhook Certificates",
+		Hosts:  strings.Split(c.flagAutoHosts, ","),
+		Expiry: 3 * time.Minute,
```
ashwin-venkatesh, line 98: This will default to 24 hours (which is the default). I had this set to 3 minutes to test how the deployment behaves

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,229 @@
...
+			err = c.updateWebhookConfig(bundle, clientset, ctx, log)
+			if err != nil {
+				continue
+			}
+		} else if err != nil {
+			log.Error(err, "Error getting secret for API server")
+			continue
+		}
+
+		if bytes.Equal(certSecret.Data[corev1.TLSCertKey], bundle.Cert) {
```
ashwin-venkatesh, line 165: Ensure we only update the certs when they have been updated to prevent constantly rewriting the secret

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+		return 1
+	}
+
+	if (c.flagCertFile != "" && c.flagKeyFile == "") || (c.flagCertFile == "" && c.flagKeyFile != "") {
+		c.UI.Error(fmt.Sprintf("invalid arguments: %s", errors.New("both tls-cert-file and tls-key-file must be provided")))
+		return 1
+	}
+
+	if c.clientset == nil {
+		config, err := rest.InClusterConfig()
```
lkysow: Could we run this externally? Other commands use

```
		k8sCfg, err := subcommand.K8SConfig(c.k8s.KubeConfig())
		if err != nil {
			c.UI.Error(fmt.Sprintf("Error retrieving Kubernetes auth: %s", err))
			return 1
		}

```

Except for connect-inject which says it must have in-cluster config (I'm guessing because it needs to receive webhooks)

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+		c.UI.Error(fmt.Sprintf("parsing flagSet: %s", err))
+		return 1
+	}
+	if len(c.flagSet.Args()) > 0 {
+		c.UI.Error(fmt.Sprintf("invalid arguments: %s", errors.New("should have no non-flag arguments")))
+		return 1
+	}
+
+	if c.flagSecretName == "" {
+		c.UI.Error(fmt.Sprintf("invalid arguments: %s", errors.New("secret-name must be set")))
```
lkysow:
```suggestion
		c.UI.Error(fmt.Sprintf("invalid arguments: %s", errors.New("-secret-name must be set")))
```

Curious why you want the `fmt` and `errors.New` instead of
```
		c.UI.Error("invalid arguments: -secret-name must be set")
```

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+		return 1
+	}
+
+	if c.flagCertFile == "" && c.flagKeyFile == "" && c.flagWebhookName == "" {
+		c.UI.Error(fmt.Sprintf("invalid arguments: %s", errors.New("either webhook-name or tls-cert-file and tls-key-file must be provided")))
```
lkysow:
```suggestion
		c.UI.Error(fmt.Sprintf("invalid arguments: %s", errors.New("either -webhook-name or -tls-cert-file and -tls-key-file must be provided")))
```

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+		return 1
+	}
+
+	if (c.flagCertFile != "" && c.flagKeyFile == "") || (c.flagCertFile == "" && c.flagKeyFile != "") {
+		c.UI.Error(fmt.Sprintf("invalid arguments: %s", errors.New("both tls-cert-file and tls-key-file must be provided")))
```
lkysow:
```suggestion
		c.UI.Error(fmt.Sprintf("invalid arguments: %s", errors.New("both -tls-cert-file and -tls-key-file must be provided")))
```

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("error creating K8S client: %s", err))
+			return 1
+		}
+	}
+
+	var certSource cert.Source = &cert.GenSource{
+		Name:  "Consul Webhook Certificates",
+		Hosts: strings.Split(c.flagAutoHosts, ","),
+	}
```
lkysow:
```suggestion
	certSource := &cert.GenSource{
		Name:  "Consul Webhook Certificates",
		Hosts: strings.Split(c.flagAutoHosts, ","),
	}
```

If immediately instantiating I believe the idiom is to use `:=` rather than `var`

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+	}
+
+	// Create the certificate notifier so we can update for certificates,
+	// then start all the background routines for updating certificates.
+	certCh := make(chan cert.Bundle)
+	certNotify := &cert.Notify{Ch: certCh, Source: certSource}
+	defer certNotify.Stop()
+	go certNotify.Start(context.Background())
+	ctx, cancelFunc := context.WithCancel(context.Background())
```
lkysow: Can we create a single context here and use it for both?

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+
+func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, clientset kubernetes.Interface, log cli.Ui) {
+	var bundle cert.Bundle
+	for {
+		select {
+		case bundle = <-ch:
+			log.Info("updated certificate bundle received. updating webhook certs.")
+			// Bundle is updated, set it up
+
+		case <-time.After(30 * time.Minute):
```
lkysow: Can you add a note as to why this is a different time than connect inject

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+			if err != nil {
+				log.Error(fmt.Sprintf("error updating webhook configuration: %s", err))
+				continue
+			}
+		} else if err != nil {
+			log.Error(fmt.Sprintf("error getting secret for API server: %s", err))
+			continue
+		}
+
+		//Dont update secret if the certificate is unchanged
```
lkysow:
```suggestion
		// Don't update secret if the certificate is unchanged.
```

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+				log.Error(fmt.Sprintf("error writing secret to API server: %s", err))
+				continue
+			}
+
+			log.Info("updating webhook configuration with new CA")
+			err = c.updateWebhookConfig(bundle, clientset, ctx)
+			if err != nil {
+				log.Error(fmt.Sprintf("error updating webhook configuration: %s", err))
+				continue
+			}
```
lkysow: Don't we need to `continue` out of here? If we fall through we just update the secret and webhook config again.

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+		log.Info("updating webhook configuration with new CA")
+		err = c.updateWebhookConfig(bundle, clientset, ctx)
+		if err != nil {
+			log.Error(fmt.Sprintf("error updating webhook configuration: %s", err))
+			continue
+		}
+	}
+}
+
+func (c *Command) updateWebhookConfig(bundle cert.Bundle, clientset kubernetes.Interface, ctx context.Context) error {
```
lkysow: Let's put `ctx` as the first arg. I think that's more standard.

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+	if c.flagWebhookName != "" && len(bundle.CACert) > 0 {
+		value := base64.StdEncoding.EncodeToString(bundle.CACert)
+
+		// If there is a MWC name set, then update the CA bundle on all the webhooks on that MWC.
```
lkysow: Should this comment go above line 202?

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+			return err
+		}
+		var patches []string
+		for i := range webhookCfg.Webhooks {
+			patches = append(patches, fmt.Sprintf(
+				`{
+						"op": "add",
+						"path": "/webhooks/%d/clientConfig/caBundle",
+						"value": %q
+					}`, i, value))
```
lkysow: Can we use a struct and json encode instead?

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+
+	clientset kubernetes.Interface
+
+	once sync.Once
+	help string
+}
+
+func (c *Command) init() {
+	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
+	c.flagSet.StringVar(&c.flagWebhookName, "webhook-name", "",
```
lkysow:
```suggestion
	c.flagSet.StringVar(&c.flagWebhookConfigName, "webhook-config-name", "",
```

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+		"MutatingWebhookConfiguration name. If specified, will auto generate cert bundle.")
```
lkysow: I thought it was the existence of `-tls-cert-file` and `-tls-key-file` that controls whether we auto generate the cert bundle?

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+	c.flagSet.StringVar(&c.flagAutoHosts, "tls-hosts", "",
+		"Comma-separated hosts for auto-generated TLS cert. If specified, will auto generate cert bundle.")
+	c.flagSet.StringVar(&c.flagSecretName, "secret-name", "",
+		"Name of the secret to update TLS certificates")
+	c.flagSet.StringVar(&c.flagSecretNamespace, "secret-namespace", "default",
+		"Namespace of the secret to update TLS certificates. Defaults to default")
+	c.flagSet.StringVar(&c.flagCertFile, "tls-cert-file", "",
+		"PEM-encoded TLS certificate to serve. If blank, will generate random cert.")
```
lkysow: Should we mention that they must set both -tls-cert-file and -tls-key-file?

subcommand/cert-manager/command.go
```diff
@@ -0,0 +1,243 @@
...
+	flagSet   *flag.FlagSet
+	httpFlags *flags.HTTPFlags
+	UI        cli.Ui
+
+	flagWebhookName     string // MutatingWebhookConfiguration for updating
+	flagAutoHosts       string // SANs for the auto-generated TLS cert.
+	flagSecretName      string // Name of secret where certificates will be written to.
+	flagSecretNamespace string // Namespace of the secret where certificates will be written to.
+	flagCertFile        string // TLS cert for listening (PEM)
+	flagKeyFile         string // TLS cert private key (PEM)
```
lkysow: I think let's drop this support for the moment.

subcommand/cert-manager/command_test.go
```diff
@@ -0,0 +1,46 @@
...
+			ui := cli.NewMockUi()
+			cmd := Command{
+				UI: ui,
+			}
+			exitCode := cmd.Run(c.flags)
+			require.Equal(tt, 1, exitCode, ui.ErrorWriter.String())
+			require.Contains(tt, ui.ErrorWriter.String(), c.expErr)
+		})
+	}
+}
```
lkysow: I think we need some more tests here. Can we test:
* secret is created and webhook config is updated
* secret/webhook isn't updated if it hasn't changed

subcommand/cert-manager/command.go
ashwin-venkatesh: Done whilst in a hurry. Cleaning up :)

subcommand/cert-manager/command.go
ashwin-venkatesh: The cert file and key file are files on disk and if they are present, we use those and do not generate the bundle. But we should get rid of this behavior for now

subcommand/cert-manager/command_test.go
ashwin-venkatesh: > secret is created and webhook config is updated
done

> secret/webhook isn't updated if it hasn't changed
writing this test is going to be tricky. what will we assert on here?

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+					Namespace: bundle.SecretNamespace,
+				},
+				Data: map[string][]byte{
+					corev1.TLSCertKey:       bundle.Cert,
+					corev1.TLSPrivateKeyKey: bundle.Key,
+				},
+				Type: corev1.SecretTypeTLS,
+			}
+
+			log.Info("creating kubernetes secret")
```
ishustava, line 175:
```suggestion
			log.Infof("creating Kubernetes secret with certificate for the %s webhook", bundle.WebhookName)
```

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+		return 0
+	}
+}
+
+func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.MetaBundle, clientset kubernetes.Interface, log cli.Ui) {
+	var bundle cert.MetaBundle
+	for {
+		select {
+		case bundle = <-ch:
+			log.Info("updated certificate bundle received. updating webhook certs.")
```
ishustava, line 145:
```suggestion
			log.Infof("updated certificate bundle received for %s. updating webhook certs.", bundle.WebhookName)
```

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+				Type: corev1.SecretTypeTLS,
+			}
+
+			log.Info("creating kubernetes secret")
+			if _, err := clientset.CoreV1().Secrets(bundle.SecretNamespace).Create(ctx, secret, metav1.CreateOptions{}); err != nil {
+				log.Error(fmt.Sprintf("error writing secret to API server: %s", err))
+				continue
+			}
+
+			log.Info("updating webhook configuration with new CA")
```
ishustava, line 181: I think this CA will only be new when cert manager starts, and after that it'll keep updating it with the same CA.
```suggestion
			log.Info("updating webhook configuration")
```

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+		if err != nil {
+			log.Error(fmt.Sprintf("error updating secret with certificate: %s", err))
+			continue
+		}
+
+		log.Info("updating webhook configuration with new CA")
+		if err := c.updateWebhookConfig(ctx, bundle, clientset); err != nil {
+			log.Error(fmt.Sprintf("error updating webhook configuration: %s", err))
+			continue
+		}
```
ishustava, line 211: I think we're already upding it above. Do we need to do it twice?

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+
+		if _, err = clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Patch(ctx, metaBundle.WebhookConfigName, types.JSONPatchType, patchesJson, metav1.PatchOptions{}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type webhookConfig struct {
+	Name            string   `json:"name,omitempty"`
```
ishustava, line 244: should we have both the name of the webhook configuration as well as the name of the webhook?

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+	if len(metaBundle.CACert) > 0 {
+		value := base64.StdEncoding.EncodeToString(metaBundle.CACert)
+
+		webhookCfg, err := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(ctx, metaBundle.WebhookConfigName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		var patches []patch
+		for i := range webhookCfg.Webhooks {
+			patches = append(patches, patch{
```
ishustava, line 225: should we check here the name of the `i`-th webhook and only update it if it is the webhook for which we're updating the CA?

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+// so it can exit gracefully. This function is needed for tests
+func (c *Command) interrupt() {
+	c.sigCh <- os.Interrupt
+}
+
+const synopsis = "Starts the Consul Kubernetes webhook-cert-manager"
+const help = `
+Usage: consul-k8s webhook-cert-manager [options]
+
+  Starts the Consul Kubernetes webhook-cert-manager that manages the lifecycle for webhook TLS certificates
```
ishustava, line 275:
```suggestion
  Starts the Consul Kubernetes webhook-cert-manager that manages the lifecycle for webhook TLS certificates.
```

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,275 @@
...
+			},
+		},
+	}
+	webhookTwo := &admissionv1beta1.MutatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: webhookConfigTwo,
+		},
+		Webhooks: []admissionv1beta1.MutatingWebhook{
+			{
+				Name: "webhookOne-under-test",
```
ishustava, line 75:
```suggestion
				Name: "webhookTwo-under-test",
```

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,275 @@
...
+		require.NoError(r, err)
+		require.Equal(r, secret2.Type, v1.SecretTypeTLS)
+
+		webhookConfig1, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), webhookConfigOne, metav1.GetOptions{})
+		require.NoError(t, err)
+		require.NotEqual(r, webhookConfig1.Webhooks[0].ClientConfig.CABundle, caBundleOne)
+
+		webhookConfig2, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), webhookConfigTwo, metav1.GetOptions{})
+		require.NoError(t, err)
+		require.NotEqual(r, webhookConfig2.Webhooks[0].ClientConfig.CABundle, caBundleTwo)
```
ishustava, line 119:
```suggestion
		require.NotEqual(r, webhookConfig2.Webhooks[1].ClientConfig.CABundle, caBundleTwo)
```

ishustava: is there any way we can inject a fake cert generator? this way we can write stronger assertions on the value of the caBundle. This would pass even if caBundle was updated to an empty string.

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,275 @@
...
+
```
ishustava, line 120: extra line

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,275 @@
...
+		require.NotEqual(r, secret2.Data[v1.TLSCertKey], []byte("cert-2"))
+		require.NotEqual(r, secret2.Data[v1.TLSPrivateKeyKey], []byte("private-key-2"))
+
+		webhookConfig1, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), webhookConfigOne, metav1.GetOptions{})
+		require.NoError(t, err)
+		require.NotEqual(r, webhookConfig1.Webhooks[0].ClientConfig.CABundle, caBundleOne)
+
+		webhookConfig2, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), webhookConfigTwo, metav1.GetOptions{})
+		require.NoError(t, err)
+		require.NotEqual(r, webhookConfig2.Webhooks[0].ClientConfig.CABundle, caBundleTwo)
```
ishustava, line 221:
```suggestion
		require.NotEqual(r, webhookConfig2.Webhooks[1].ClientConfig.CABundle, caBundleTwo)
```

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,275 @@
...
+    "name": "webhookTwo",
+    "tlsAutoHosts": [
+      "foo-2",
+      "bar-3",
+      "baz-4"
+    ],
+    "secretName": "secret-deploy-2",
+    "secretNamespace": "default"
+  }
+]`
```
ishustava, line 275: do you think we can add a test to check that the cert bundle is updated when certs are about to expire? I think you'd probably need to figure out a way to make these configurable from tests since we're not going to wait for 24h in tests 😄

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/hashicorp/go-sockaddr v1.0.2 // indirect
 	github.com/hashicorp/go-uuid v1.0.2 // indirect
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
+	github.com/hashicorp/serf v0.9.3
```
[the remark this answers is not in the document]
ishustava: Oh it's because it imported the retry package from serf. We should use this one `github.com/hashicorp/consul/sdk/testutil/retry`

ishustava (commented):
Nice work, @ashwin-venkatesh ! I added some comments and suggestions in-line.

go.mod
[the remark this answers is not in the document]
ashwin-venkatesh: ohh..good catch!!

subcommand/controller/command.go
```diff
@@ -46,14 +47,18 @@ func (c *Command) init() {
 	c.flagSet.BoolVar(&c.flagEnableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	c.flagSet.StringVar(&c.flagWebhookTLSCertDir, "webhook-tls-cert-dir", "",
+		"Directory that contains the tls cert and key required to communicate with the MutatingWebhookConfiguration")
```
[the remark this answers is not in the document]
ashwin-venkatesh: > this is the directory with certs for the webhook itself, not for the controller to communicate with the webhook

The certs here are used by the Handler for the webhook, which does run inside the controller. The MutatingWebhook has the CA certificate, and the handler has the cert & key that are signed by that CA so that they can communicate with the Mutating Webhook.

> do the cert and key have to have special file names so that they'll be picked up by the webhook server?

Yup. When the kind of the secret is TLS secret (which is what the webhook-cert-manager creates), the cert is under the `tls.crt` and the key `tls.key`. The webhook manager defaults to looking for files with those names.

subcommand/controller/command.go
```diff
@@ -93,10 +96,12 @@ func (c *Command) Run(_ []string) int {
 	}
 
 	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
```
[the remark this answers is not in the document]
ashwin-venkatesh: I agree. I think we should use flags for everything!! 😄 

Additionally, I think we should always deploy our webhooks, so maybe it doesn't make sense to have this check at all, and always deploy the webhook. Do you think that makes sense.

subcommand/controller/command.go
```diff
@@ -93,10 +96,12 @@ func (c *Command) Run(_ []string) int {
...
+		mgr.GetWebhookServer().CertDir = c.flagWebhookTLSCertDir
```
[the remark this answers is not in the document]
ashwin-venkatesh: Yes! good catch. Error out if not set is 👍

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+type Command struct {
+	UI cli.Ui
+
+	flagSet   *flag.FlagSet
+	httpFlags *flags.HTTPFlags
```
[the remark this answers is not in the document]
ashwin-venkatesh: good catch.

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,277 @@
...
+
+	configFile, err := ioutil.ReadFile(c.flagConfigFile)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("error reading config file from %s", c.flagConfigFile))
+	}
+	var configs []webhookConfig
+	err = json.Unmarshal(configFile, &configs)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("error unmarshalling config file: %s", err.Error()))
+	}
```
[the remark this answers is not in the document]
ashwin-venkatesh: 🤔 I do like the idea. Do you have suggestions for what validations should be?

subcommand/webhook-cert-manager/command.go
ashwin-venkatesh: In case of helm upgrades where the cert-manager deployment restarts, it will lead to a situation where the secret already exists so it will fall into this flow where a new CA might be generated + secret exists, in which case it is safer to update the WebhookConfig as well. But yeah, this will not be required for every new leaf certificate. thoughts?

subcommand/controller/command.go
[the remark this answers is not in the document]
lkysow: This comes from the generated code because when running locally webhooks don't work (since kube can't call into your laptop. I think the makefile actually used to set this env var for you when running locally but maybe we've lost that target.

Down to switch to a flag though, separate PR makes sense I think.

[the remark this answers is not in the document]
ashwin-venkatesh: In that case, i'll add a task to address this for now and resolve this here.

subcommand/webhook-cert-manager/command_test.go
ashwin-venkatesh: I think this one should be `[0]`, but I think it makes sense to have a webhook-config with multiple webhooks in it to test that every webhook has its CA updated.

subcommand/webhook-cert-manager/command.go
ashwin-venkatesh: We will be updating the CA on every single webhook defined in the webhook configuration as they will share the same CA and cert.

subcommand/webhook-cert-manager/command.go
ashwin-venkatesh: given we will be updating every webhook defined in the webhook config, we shouldn't need of the name of the individual webhooks :)

subcommand/webhook-cert-manager/command_test.go
ashwin-venkatesh: I added a field called `certExpiry` to the command, which can override the default cert expiry if set, but it is not exposed via flags. Created a test which then sets the expiry for the test to 1 second. Let me know what you think

subcommand/controller/command.go
```diff
@@ -46,23 +47,29 @@ func (c *Command) init() {
...
 	if len(c.flagSet.Args()) > 0 {
 		setupLog.Error(errors.New("should have no non-flag arguments"), "invalid arguments")
 		return 1
 	}
-
-	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	if c.flagWebhookTLSCertDir == "" {
+		setupLog.Error(errors.New("-webhook-tls-cert-dir must be set"), "invalid arguments")
+		return 1
+	}
```
ishustava, line 72: We should add tests for it.

subcommand/controller/command.go
```diff
@@ -93,10 +100,12 @@ func (c *Command) Run(_ []string) int {
 	}
 
 	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
+		mgr.GetWebhookServer().CertDir = c.flagWebhookTLSCertDir
```
ishustava, line 103: Could we add a comment saying that setting cert dir will also watch for file changes and rotate certs?

ishustava (approved):
This looks good! I've tested it out and saw the certs being updated. 

One thing I noticed when testing is that all logs are printed to the UI, and I think it would be nice to use a logger so that you can see timestamps to see times when certs were updated, etc. I don't this is blocking, however.

ashwin-venkatesh:
>  it would be nice to use a logger so that you can see timestamps to see times when certs were updated

agreed and done

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+			// are in sync. Because the certificate and key are being read from a secret,
+			// this does not have to be processed as aggressively as the 1 sec time in
+			// the connect inject cert watcher.
+
+		case <-ctx.Done():
+			// Quit
+			return
+		}
+
+		log.Info("getting secret from kubernetes")
```
lkysow, line 198: What do you think about creating a logger with the k/v pairs already set?
```
		iterLog := log.With("mutatingwebhookconfig", bundle.WebhookConfigName, "secret", bundle.SecretName, "secretNS", bundle.SecretNamespace)

```

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+					Namespace: bundle.SecretNamespace,
+				},
+				Data: map[string][]byte{
+					corev1.TLSCertKey:       bundle.Cert,
+					corev1.TLSPrivateKeyKey: bundle.Key,
+				},
+				Type: corev1.SecretTypeTLS,
+			}
+
+			log.Info(fmt.Sprintf("creating Kubernetes secret with certificate for the Mutating Webhook Config %s", bundle.WebhookConfigName))
```
lkysow, line 213:
```suggestion
			iterLog.Info("creating Kubernetes secret with certificate")
```
Then these can just be

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+			return
+		}
+
+		log.Info("getting secret from kubernetes")
+		certSecret, err := clientset.CoreV1().Secrets(bundle.SecretNamespace).Get(ctx, bundle.SecretName, metav1.GetOptions{})
+		if err != nil && k8serrors.IsNotFound(err) {
+			secret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      bundle.SecretName,
+					Namespace: bundle.SecretNamespace,
```
lkysow, line 204: Is this needed if we're setting it on the client call anyway?

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+					corev1.TLSCertKey:       bundle.Cert,
+					corev1.TLSPrivateKeyKey: bundle.Key,
+				},
+				Type: corev1.SecretTypeTLS,
+			}
+
+			log.Info(fmt.Sprintf("creating Kubernetes secret with certificate for the Mutating Webhook Config %s", bundle.WebhookConfigName))
+			if _, err := clientset.CoreV1().Secrets(bundle.SecretNamespace).Create(ctx, secret, metav1.CreateOptions{}); err != nil {
+				log.Error(fmt.Sprintf("Error writing secret to API server: %s", err))
+				continue
```
lkysow, line 216: Ahh, this is why the other loop looped so quickly. If anything errors then we'd want it to retry quickly. Given that should we change the iteration time? Or special case a retry time if there was an error?

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+			continue
+		} else if err != nil {
+			log.Error(fmt.Sprintf("Error getting secret from Kubernetes: %s", err))
+			continue
+		}
+
+		// Don't update secret if the certificate is unchanged.
+		if bytes.Equal(certSecret.Data[corev1.TLSCertKey], bundle.Cert) {
+			continue
+		}
```
lkysow, line 233: Seems like we should check both key and cert to be certain

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+		log.Info("updating secret with new certificate")
+		_, err = clientset.CoreV1().Secrets(bundle.SecretNamespace).Update(ctx, certSecret, metav1.UpdateOptions{})
+		if err != nil {
+			log.Error(fmt.Sprintf("error updating secret with certificate: %s", err))
+			continue
+		}
+
+		log.Info("updating webhook configuration with new CA")
+		if err := c.updateWebhookConfig(ctx, bundle, clientset); err != nil {
+			log.Error(fmt.Sprintf("error updating webhook configuration: %s", err))
```
lkysow, line 247: Best to k/v these
```suggestion
			log.Error("error updating webhook configuration", "err", err))
```

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+			continue
+		}
+	}
+}
+
+func (c *Command) updateWebhookConfig(ctx context.Context, metaBundle cert.MetaBundle, clientset kubernetes.Interface) error {
+	if len(metaBundle.CACert) > 0 {
```
lkysow, line 254: If it's not > 0 shouldn't that be an error?

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,463 @@
...
+		secret1, err := k8s.CoreV1().Secrets("default").Get(context.TODO(), secretOne, metav1.GetOptions{})
+		require.NoError(r, err)
+		require.Equal(r, secret1.Type, v1.SecretTypeTLS)
+
+		secret2, err := k8s.CoreV1().Secrets("default").Get(context.TODO(), secretTwo, metav1.GetOptions{})
+		require.NoError(r, err)
+		require.Equal(r, secret2.Type, v1.SecretTypeTLS)
+
+		webhookConfig1, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), webhookConfigOne, metav1.GetOptions{})
+		require.NoError(t, err)
```
lkysow, line 121:
```suggestion
		require.NoError(r, err)
```

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,463 @@
...
+		require.NotEqual(r, webhookConfig1.Webhooks[0].ClientConfig.CABundle, caBundleOne)
+
+		webhookConfig2, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), webhookConfigTwo, metav1.GetOptions{})
+		require.NoError(t, err)
```
lkysow, line 125:
```suggestion
		require.NoError(r, err)
```

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,463 @@
...
+	retry.RunWith(timer, t, func(r *retry.R) {
+		secret1, err := k8s.CoreV1().Secrets("default").Get(context.TODO(), secretOne, metav1.GetOptions{})
+		require.NoError(r, err)
+		require.Equal(r, secret1.Type, v1.SecretTypeTLS)
+
+		secret2, err := k8s.CoreV1().Secrets("default").Get(context.TODO(), secretTwo, metav1.GetOptions{})
+		require.NoError(r, err)
+		require.Equal(r, secret2.Type, v1.SecretTypeTLS)
+
+		webhookConfig1, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), webhookConfigOne, metav1.GetOptions{})
```
lkysow, line 120: Can we create a single ctx?

subcommand/webhook-cert-manager/command_test.go
lkysow: check your other retry loops too, I see some `t`'s

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,463 @@
...
+		require.NoError(t, err)
+		require.NotEqual(r, webhookConfig2.Webhooks[0].ClientConfig.CABundle, caBundleTwo)
+		require.NotEqual(r, webhookConfig2.Webhooks[1].ClientConfig.CABundle, caBundleTwo)
+		require.Equal(r, webhookConfig2.Webhooks[0].ClientConfig.CABundle, webhookConfig2.Webhooks[1].ClientConfig.CABundle)
+	})
+}
+
+func TestRun_SecretExists(t *testing.T) {
+	t.Parallel()
+	secretOne := "secret-deploy-1"
```
lkysow, line 134: `secretOneName`? Then you don't need `secretOne` and `secret1`

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,463 @@
...
+		webhookConfig1, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), webhookConfigOne, metav1.GetOptions{})
+		require.NoError(t, err)
+		require.NotEqual(r, webhookConfig1.Webhooks[0].ClientConfig.CABundle, caBundleOne)
+	})
+
+	// Wait for certs to be rotated.
+	time.Sleep(2 * time.Second)
+
+	// Check that the certificate is rotated and the secret is updated.
+	retry.RunWith(timer, t, func(r *retry.R) {
```
lkysow, line 319: Can you re-use `timer`?

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,463 @@
...
+		secret1, err := k8s.CoreV1().Secrets("default").Get(context.TODO(), secretOne, metav1.GetOptions{})
+		require.NoError(r, err)
+		require.NotEqual(r, secret1.Data[v1.TLSCertKey], certificate)
+		require.NotEqual(r, secret1.Data[v1.TLSPrivateKeyKey], key)
+	})
+}
+
+func TestValidate(t *testing.T) {
```
lkysow, line 327: nice tests

lkysow (commented):
The code looks great, just have some little things except for the re-looping on errors. I'm moving on to actually running it now.

lkysow:
I'm not seeing the webhook config get caCert set:
```
k get mutatingwebhookconfiguration consul-controller-mutating-webhook-configuration -o yaml
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  annotations:
    meta.helm.sh/release-name: consul
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2020-09-12T00:06:03Z"
  generation: 3
  labels:
    app: consul
    app.kubernetes.io/managed-by: Helm
    chart: consul-helm
    component: controller
    heritage: Helm
    release: consul
  name: consul-controller-mutating-webhook-configuration
  resourceVersion: "1649"
  selfLink: /apis/admissionregistration.k8s.io/v1/mutatingwebhookconfigurations/consul-controller-mutating-webhook-configuration
  uid: 36a0e556-0d51-4987-a030-0e992a921a15
webhooks:
- admissionReviewVersions:
  - v1beta1
  clientConfig:
    service:
      name: consul-controller-webhook
      namespace: default
      path: /mutate-v1alpha1-servicedefaults
      port: 443
  failurePolicy: Fail
  matchPolicy: Exact
  name: mutate-servicedefaults.consul.hashicorp.com
  namespaceSelector: {}
  objectSelector: {}
  reinvocationPolicy: Never
  rules:
  - apiGroups:
    - consul.hashicorp.com
    apiVersions:
    - v1alpha1
    operations:
    - CREATE
    - UPDATE
    resources:
    - servicedefaults
    scope: '*'
  sideEffects: Unknown
  timeoutSeconds: 30
```

ashwin-venkatesh:
> I'm not seeing the webhook config get caCert set

err. that is odd. i have seen the caBundle setup running the same command..does the webhook work if you try creating a service default?

lkysow:
Yes... So now I'm super confused.

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+		go certNotify.Start(ctx)
+	}
+
+	go c.certWatcher(ctx, certCh, c.clientset, c.logger)
+
+	closeNotifiers := func() {
+		for _, notifier := range notifiers {
+			notifier.Stop()
+		}
+	}
```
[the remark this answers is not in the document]
ashwin-venkatesh: good catch. There used to be a defer that would invoke this, but given it is an a for loop that doesnt exit, i removed the defer but forgot to inline this into the call site :)

subcommand/webhook-cert-manager/command.go
```diff
@@ -0,0 +1,343 @@
...
+	if c.SecretName == "" {
+		err = multierror.Append(err, errors.New(`config.SecretName cannot be ""`))
+	}
+	if c.SecretNamespace == "" {
+		err = multierror.Append(err, errors.New(`config.SecretNameSpace cannot be ""`))
+	}
+
+	if err != nil {
+		err.ErrorFormat = func(errs []error) string {
```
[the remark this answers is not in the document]
ashwin-venkatesh: The formatting of the error message would index each error on a new line instead of comma separating them which I felt was more readable.

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,532 @@
...
+
+	_, err = file.Write([]byte(configFileUpdates))
+	require.NoError(t, err)
+
+	exitCh := runCommandAsynchronously(&cmd, []string{
+		"-config-file", file.Name(),
+	})
+	defer stopCommand(t, &cmd, exitCh)
+
+	ctx := context.TODO()
```
lkysow, line 371:
```suggestion
	ctx := context.Background()
```

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,532 @@
...
+	// This first sleep is required to ensure the command had completed validating that the MWC does exist in the cluster
+	// It ensures the MWC is absent when the server actually attempts to update it with a bundle.
+	time.Sleep(1 * time.Millisecond)
```
lkysow, line 374: Is this long enough?

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -0,0 +1,532 @@
...
+	err = k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Delete(ctx, webhookName, metav1.DeleteOptions{})
+	require.NoError(t, err)
+
+	// This sleep ensures the cluster attempts to try updating the MWC every second and fails before we re-create it in the cluster.
+	time.Sleep(2 * time.Second)
```
lkysow, line 379: Can you play with the timeouts so we don't need to wait this long? Can we set the error retry to 100ms?

lkysow (approved):
🎉

subcommand/webhook-cert-manager/command_test.go
ashwin-venkatesh: It ran successfully across multiple runs consistently. Ill bump this and the other timeout too a 100ms and check it out

ashwin-venkatesh: When I bumped it too 100ms, it successfully update the MWC the first time without an error. This sets the retry timer to 30 mins and causes the test to fail. I tried 10ms and that causes it to succeed as well. Dropped it back down to 1ms for consistent results for now.

PR #320, opened 2020-09-03
Add defaults and validation for ServiceDefaults

ashwin-venkatesh:
Changes proposed in this PR:

Add defaulting and validation for ServiceDefaults

Checklist:
- [x] Tests added

How I tested the PR:
• Used the image `ashwinvenkatesh/consul-k8s:validation` to deploy consul via helm with the controller enabled.
• Tried creating service default resource with
  • Invalid mesh gateway mode
  • Invalid expose path protocol (not in "http" and "http2") and path value (not prefixed by "/")
This led to webhook errors which indicated the invalid fields
  • Combinations of various invalid values for the above fields
This led to webhook errors that listed every single invalid field

How I expect people to test the PR:

• You can retry the above scenarios with multiple invalid values to verify every invalid value is mentioned in the returned error from the webhook.

### DIFF
--- a/api/v1alpha1/servicedefaults_types.go
+++ b/api/v1alpha1/servicedefaults_types.go
@@ -180,9 +180,7 @@ func (e ExposeConfig) validate() []*field.Error {
 		if path.Path != "" && !strings.HasPrefix(path.Path, "/") {
 			errs = append(errs, field.Invalid(field.NewPath("spec").Child("expose").Child(fmt.Sprintf("paths[%d]", i)).Child("path"), path.Path, `must begin with a '/'`))
 		}
-		if sliceContains(protocols, path.Protocol) {
-			continue
-		} else {
+		if !sliceContains(protocols, path.Protocol) {
 			errs = append(errs, field.Invalid(field.NewPath("spec").Child("expose").Child(fmt.Sprintf("paths[%d]", i)).Child("protocol"), path.Protocol, notInSliceMessage(protocols)))
 		}
 	}
--- a/api/v1alpha1/servicedefaults_types.go
+++ b/api/v1alpha1/servicedefaults_types.go
@@ -11,6 +11,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
+const (
+	ConsulHashicorpGroup string = "consul.hashicorp.com"
+	ServiceDefaultsKind  string = "servicedefaults"
+)
+
 // ServiceDefaultsSpec defines the desired state of ServiceDefaults
 type ServiceDefaultsSpec struct {
 	// Protocol sets the protocol of the service. This is used by Connect proxies for
@@ -85,16 +90,15 @@ func (in *ServiceDefaults) Validate() error {
 	if err := in.Spec.MeshGateway.validate(); err != nil {
 		allErrs = append(allErrs, err)
 	}
-	if err := in.Spec.Expose.validate(); err != nil {
-		allErrs = append(allErrs, err...)
-	}
-	if len(allErrs) == 0 {
-		return nil
+	allErrs = append(allErrs, in.Spec.Expose.validate()...)
+
+	if len(allErrs) > 0 {
+		return apierrors.NewInvalid(
+			schema.GroupKind{Group: ConsulHashicorpGroup, Kind: ServiceDefaultsKind},
+			in.Name, allErrs)
 	}
 
-	return apierrors.NewInvalid(
-		schema.GroupKind{Group: "consul.hashicorp.com", Kind: "ServiceDefaults"},
-		in.Name, allErrs)
+	return nil
 }
 
 // ExposeConfig describes HTTP paths to expose through Envoy outside of Connect.
@@ -182,9 +186,5 @@ func (e ExposeConfig) validate() []*field.Error {
 			errs = append(errs, field.Invalid(field.NewPath("spec").Child("expose").Child(fmt.Sprintf("paths[%d]", i)).Child("protocol"), path.Protocol, `must be one of "http" or "http2"`))
 		}
 	}
-	if len(errs) == 0 {
-		return nil
-	} else {
-		return errs
-	}
+	return errs
 }
--- a/api/v1alpha1/servicedefaults_types.go
+++ b/api/v1alpha1/servicedefaults_types.go
@@ -1,6 +1,9 @@
 package v1alpha1
 
 import (
+	"fmt"
+	"strings"
+
 	capi "github.com/hashicorp/consul/api"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -88,7 +91,7 @@ func (in *ServiceDefaults) Validate() error {
 		allErrs = append(allErrs, err)
 	}
 	if err := in.Spec.Expose.validate(); err != nil {
-		allErrs = append(allErrs, err)
+		allErrs = append(allErrs, err...)
 	}
 	if len(allErrs) == 0 {
 		return nil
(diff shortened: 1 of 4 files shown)


api/v1alpha1/servicedefaults_types.go
```diff
@@ -52,25 +58,49 @@ func init() {
...
-		s.Spec.ExternalSNI == entry.ExternalSNI
+func (in *ServiceDefaults) MatchesConsul(entry *capi.ServiceConfigEntry) bool {
+	return in.Name == entry.GetName() &&
+		in.Spec.Protocol == entry.Protocol &&
+		in.Spec.MeshGateway.Mode == string(entry.MeshGateway.Mode) &&
+		in.Spec.Expose.matches(entry.Expose) &&
+		in.Spec.ExternalSNI == entry.ExternalSNI
+}
+
+func (in *ServiceDefaults) Default() {
```
ishustava, line 82: could we add docs to these?

api/v1alpha1/servicedefaults_types.go
```diff
@@ -144,3 +174,31 @@ func (e ExposeConfig) toConsul() capi.ExposeConfig {
...
+		if path.Protocol == "" {
+			e.Paths[i].Protocol = "http"
+		}
+	}
+}
+
+func (e ExposeConfig) validate() []*field.Error {
+	var errs field.ErrorList
+	for i, path := range e.Paths {
+		if !strings.HasPrefix(path.Path, "/") {
```
ishustava, line 189: If `path.Path` is empty, this will error, but it probably shouldn't since `path` is not required.

ishustava (commented):
The validating logic looks awesome!

I wasn't super sure if we need defaulting though. It seems like a duplication of logic that already exists in Consul, and so I'm not sure what is the value of having it in the controller as well. It also couples us to consul versions more tightly - if consul changes defaults, you'd need to update to the version of the controller/consul-k8s that has that default.

api/v1alpha1/servicedefaults_types.go
ashwin-venkatesh: Is it valid for a ServiceDefault to have ExposePaths with the path being `""`?

ishustava: Yeah it's valid, because you could set only ports without path.

ishustava (approved):
Looks good!

api/v1alpha1/servicedefaults_types.go
```diff
@@ -52,25 +58,44 @@ func init() {
...
+	}
+	if err := in.Spec.Expose.validate(); err != nil {
+		allErrs = append(allErrs, err...)
+	}
+	if len(allErrs) == 0 {
+		return nil
+	}
+
+	return apierrors.NewInvalid(
+		schema.GroupKind{Group: "consul.hashicorp.com", Kind: "ServiceDefaults"},
```
lkysow, line 97: Can we make constants for these?

api/v1alpha1/servicedefaults_types.go
```diff
@@ -144,3 +169,23 @@ func (e ExposeConfig) toConsul() capi.ExposeConfig {
...
 }
+
+func (e ExposeConfig) validate() []*field.Error {
+	var errs field.ErrorList
+	for i, path := range e.Paths {
+		if path.Path != "" && !strings.HasPrefix(path.Path, "/") {
+			errs = append(errs, field.Invalid(field.NewPath("spec").Child("expose").Child(fmt.Sprintf("paths[%d]", i)).Child("path"), path.Path, `must begin with a '/'`))
+		}
+		switch path.Protocol {
+		case "http", "http2":
```
lkysow, line 180: Can we use something like `var allowPathProtocols = []string{"http", "http2"}` and use it for this case and constructing the error message?

api/v1alpha1/servicedefaults_types_test.go
```diff
@@ -654,3 +654,88 @@ func TestMatchesConsul(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "my-service",
+				},
+				Spec: ServiceDefaultsSpec{
+					MeshGateway: MeshGatewayConfig{
+						Mode: "foobar",
+					},
+				},
+			},
+			`ServiceDefaults.consul.hashicorp.com "my-service" is invalid: spec.meshGateway.mode: Invalid value: "foobar": must be one of "remote", "local", "none" or ""`,
```
lkysow, line 674: Is it normal for this to be camelcased? (ServiceDefauts.consul.hashicorp.com)

api/v1alpha1/servicedefaults_types_test.go
```diff
@@ -654,3 +654,88 @@ func TestMatchesConsul(t *testing.T) {
...
+						Paths: []ExposePath{
+							{
+								Protocol: "invalid-protocol",
+								Path:     "invalid-path",
+							},
+						},
+					},
+				},
+			},
+			`ServiceDefaults.consul.hashicorp.com "my-service" is invalid: [spec.meshGateway.mode: Invalid value: "invalid-mode": must be one of "remote", "local", "none" or "", spec.expose.paths[0].path: Invalid value: "invalid-path": must begin with a '/', spec.expose.paths[0].protocol: Invalid value: "invalid-protocol": must be one of "http" or "http2"]`,
```
lkysow, line 731: nice!!

api/v1alpha1/servicedefaults_types.go
```diff
@@ -52,25 +58,44 @@ func init() {
...
+// Validate validates the fields provided in the spec of the ServiceDefaults and
+// returns an error which lists all invalid fields in the resource spec.
+func (in *ServiceDefaults) Validate() error {
+	var allErrs field.ErrorList
+	if err := in.Spec.MeshGateway.validate(); err != nil {
+		allErrs = append(allErrs, err)
+	}
+	if err := in.Spec.Expose.validate(); err != nil {
+		allErrs = append(allErrs, err...)
+	}
```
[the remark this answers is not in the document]
lkysow: If we do my suggestion about the empty slice

api/v1alpha1/types.go
```diff
@@ -47,3 +48,12 @@ func (m MeshGatewayConfig) toConsul() capi.MeshGatewayConfig {
 		}
 	}
 }
+
+func (m MeshGatewayConfig) validate() *field.Error {
+	switch m.Mode {
+	case "", "local", "remote", "none":
```
lkysow, line 54: Can we make this a `var meshGatewayModes = []string{"", "local", "remote", "none"}` and then use it for this case and constructing the error message?

api/v1alpha1/servicedefaults_types.go
```diff
@@ -52,25 +58,44 @@ func init() {
...
+		in.Spec.Protocol == entry.Protocol &&
+		in.Spec.MeshGateway.Mode == string(entry.MeshGateway.Mode) &&
+		in.Spec.Expose.matches(entry.Expose) &&
+		in.Spec.ExternalSNI == entry.ExternalSNI
+}
+
+// Validate validates the fields provided in the spec of the ServiceDefaults and
+// returns an error which lists all invalid fields in the resource spec.
+func (in *ServiceDefaults) Validate() error {
+	var allErrs field.ErrorList
```
lkysow, line 85: What about validating `protocol`? I thought only tcp, http, http2 were supported?

ashwin-venkatesh: I used [this page](https://www.consul.io/docs/agent/config-entries/service-defaults) to determine the valid values which did not specify that subset. Where did you find your list? Maybe there are some others I might have missed.

api/v1alpha1/servicedefaults_types_test.go
ashwin-venkatesh: I used the reference from the kubebuilder book. It does look kinda ugly though. Ill switch it to being kebab case.

api/v1alpha1/servicedefaults_types.go
lkysow: Gotcha, looking at the consul code there is no further validation (afaict) so nvm.

lkysow (approved):

api/v1alpha1/types.go
```diff
@@ -47,3 +50,30 @@ func (m MeshGatewayConfig) toConsul() capi.MeshGatewayConfig {
...
+	}
+}
+
+func notInSliceMessage(slice []string) string {
+	message := ""
+	for _, s := range slice {
+		message = fmt.Sprintf(`%s, "%s"`, message, s)
+	}
+	runes := []rune(message)
+	return fmt.Sprintf(`must be one of %s`, string(runes[2:]))
```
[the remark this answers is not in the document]
ashwin-venkatesh: I think i tried doing that, but it looked different and hence i went down the rune rabbit hole. Here was the output:
`must be one of \"remote\\\", \\\"local\\\", \\\"none\\\", \\\"\"`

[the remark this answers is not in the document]
lkysow: Ahh, yeah. I think I got it working here with slight modification: https://play.golang.org/p/z5iEJ-vvk8k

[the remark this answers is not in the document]
ashwin-venkatesh: ok that worked but i could have sweared i tried that first 😢
thanks Luke

PR #321, opened 2020-09-08
Connect: only look for service account secrets when creating/updating auth method

ishustava:
Kubernetes service account could have multiple secrets associated with it.
Previously, we were only looking at the first secret in the list and using
that secret to create or update the auth method in Consul.
However, on some platforms the service account may contain other types of secrets.
Specifically, in the case of Openshift, there are two secrets: one for the service account token
and the other for docker config credentials. This second secret gets injected automatically by Openshift.

This PR changes our implementation to use the first secret of type kubernetes.io/service-account-token.

### How I've tested this PR:
Created a docker image (`ishustava/consul-k8s-dev:09-04-2020-5e7491d`) and ran connect acceptance tests with it on openshift.

No infrastructure tests required from the reviewers at this point. We can rely on helm acceptance tests to make sure we didn't break anything.

### Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/server-acl-init/command_test.go
+++ b/subcommand/server-acl-init/command_test.go
@@ -1890,7 +1890,7 @@ func setUpK8sServiceAccount(t *testing.T, k8s *fake.Clientset) (string, string) 
 	sa, _ := k8s.CoreV1().ServiceAccounts(ns).Get(context.Background(), serviceAccountName, metav1.GetOptions{})
 	if sa == nil {
 		// Create a service account that references two secrets.
-		// The second secret is mimicing the behavior on Openshift,
+		// The second secret is mimicking the behavior on Openshift,
 		// where two secrets are injected: one with SA token and one with docker config.
 		_, err := k8s.CoreV1().ServiceAccounts(ns).Create(
 			context.Background(),
--- a/subcommand/server-acl-init/command_test.go
+++ b/subcommand/server-acl-init/command_test.go
@@ -1889,6 +1889,9 @@ func setUpK8sServiceAccount(t *testing.T, k8s *fake.Clientset) (string, string) 
 	serviceAccountName := resourcePrefix + "-connect-injector-authmethod-svc-account"
 	sa, _ := k8s.CoreV1().ServiceAccounts(ns).Get(context.Background(), serviceAccountName, metav1.GetOptions{})
 	if sa == nil {
+		// Create a service account that references two secrets.
+		// The second secret is mimicing the behavior on Openshift,
+		// where two secrets are injected: one with SA token and one with docker config.
 		_, err := k8s.CoreV1().ServiceAccounts(ns).Create(
 			context.Background(),
 			&v1.ServiceAccount{
--- a/subcommand/server-acl-init/command_test.go
+++ b/subcommand/server-acl-init/command_test.go
@@ -88,6 +88,7 @@ func TestRun_Defaults(t *testing.T) {
 		clientset: k8s,
 	}
 	args := []string{
+		"-timeout=1m",
 		"-k8s-namespace=" + ns,
 		"-server-address", strings.Split(testSvr.HTTPAddr, ":")[0],
 		"-server-port", strings.Split(testSvr.HTTPAddr, ":")[1],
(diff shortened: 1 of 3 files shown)


subcommand/server-acl-init/command_test.go
```diff
@@ -88,6 +88,7 @@ func TestRun_Defaults(t *testing.T) {
 		clientset: k8s,
 	}
 	args := []string{
+		"-timeout=1m",
```
ishustava, line 91: adding this timeout to all tests, so that if the test fails, it doesn't run too long.

subcommand/server-acl-init/command_test.go
```diff
@@ -1878,6 +1896,9 @@ func setUpK8sServiceAccount(t *testing.T, k8s *fake.Clientset) (string, string)
 					Name: serviceAccountName,
 				},
 				Secrets: []v1.ObjectReference{
+					{
```
lkysow, line 1899: Can you add a note that this is here to mimic openshift?

subcommand/server-acl-init/connect_inject.go
```diff
@@ -152,16 +152,20 @@ func (c *Command) createAuthMethodTmpl(authMethodName string) (api.ACLAuthMethod
...
+		err = c.untilSucceeds(fmt.Sprintf("getting %s Secret", secret.Name),
+			func() error {
+				var err error
+				saSecret, err = c.clientset.CoreV1().Secrets(c.flagK8sNamespace).Get(context.TODO(), secret.Name, metav1.GetOptions{})
+				return err
+			})
+		if saSecret.Type == apiv1.SecretTypeServiceAccountToken {
+			break
+		}
+	}
```
lkysow, line 168: I know this issue existed with the previous implementation but while we're in here what do you think about doing something like this. As written if we don't find a secret of type ServiceAccountToken then we're going to use the last secret in the list.

```suggestion
	for _, secretRef := range authMethodServiceAccount.Secrets {
	        var secret *apiv1.Secret
		err = c.untilSucceeds(fmt.Sprintf("getting %s Secret", secretRef.Name),
			func() error {
				var err error
				secret, err = c.clientset.CoreV1().Secrets(c.flagK8sNamespace).Get(context.TODO(), secretRef.Name, metav1.GetOptions{})
				return err
			})
		if secret.Type == apiv1.SecretTypeServiceAccountToken {
			saSecret = secret
			break
		}
	}
	if err != nil {
	   return api.ACLAuthMethod{}, err
	}
	if saSecret == nil {
	  return api.ACLAuthMethod{}, fmt.Errorf("Found no secret of type ....")
	}
```

ashwin-venkatesh (approved):
Looks great!

subcommand/server-acl-init/connect_inject.go
ashwin-venkatesh: Will that be a valid secret? Wouldn't it be preferable to error out and say no valid secret found?

lkysow: >Will that be a valid secret? Wouldn't it be preferable to error out and say no valid secret found?

Is this a question for me or did we just have overlapping reviews? :D

ashwin-venkatesh: lol. Definitely an overlapping review :P Sorry for the confusion.

ishustava: Yeah definitely, good catch!

subcommand/server-acl-init/connect_inject.go
```diff
@@ -152,19 +152,29 @@ func (c *Command) createAuthMethodTmpl(authMethodName string) (api.ACLAuthMethod
...
+			saSecret = secret
+			break
+		}
+	}
 	if err != nil {
 		return api.ACLAuthMethod{}, err
 	}
+	if saSecret == nil {
+		return api.ACLAuthMethod{},
+			fmt.Errorf("found no secret of type 'kubernetes.io/service-account-token' associated with the %s service account", saName)
```
lkysow, line 176: Can you write a test for this?

ishustava: I tried to add a test to the command test but realized that I can't assert on the log output; then I saw we don't have tests for other errors in this command, so I thought maybe it's ok. Let me see though if I can add a test just for this function, so I can have more direct access to the error.

lkysow: Oh okay, nvm then.

lkysow (approved):

subcommand/server-acl-init/connect_inject.go
ishustava: I added a test, and it actually helped me find a bug in the code 😄 Thank you for suggesting it! #alwaysaddtests

PR #322, opened 2020-09-09
Update ACLs for consul ent and controller

lkysow:
Changes proposed in this PR:
- Companion to https://github.com/hashicorp/consul-k8s/pull/323. This updates the ACL token we create to make it work for consul enterprise namespaces.

How I've tested this PR: Ran the acceptance tests from https://github.com/hashicorp/consul-helm/pull/601 which tests with secure installation and a Docker image made with a merge of this branch and https://github.com/hashicorp/consul-k8s/pull/323.

How I expect reviewers to test this PR: You could run the acceptance tests yourselves or review the ones I wrote and see that CI/CD passes for them.

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/server-acl-init/rules.go
+++ b/subcommand/server-acl-init/rules.go
@@ -7,10 +7,13 @@ import (
 )
 
 type rulesData struct {
-	EnableNamespaces               bool
-	ConsulSyncDestinationNamespace string
-	EnableSyncK8SNSMirroring       bool
-	SyncK8SNSMirroringPrefix       string
+	EnableNamespaces        bool
+	SyncConsulDestNS        string
+	SyncEnableNSMirroring   bool
+	SyncNSMirroringPrefix   string
+	InjectConsulDestNS      string
+	InjectEnableNSMirroring bool
+	InjectNSMirroringPrefix string
 }
 
 type gatewayRulesData struct {
@@ -179,10 +182,10 @@ func (c *Command) syncRules() (string, error) {
   }
 {{- if .EnableNamespaces }}
 operator = "write"
-{{- if .EnableSyncK8SNSMirroring }}
-namespace_prefix "{{ .SyncK8SNSMirroringPrefix }}" {
+{{- if .SyncEnableNSMirroring }}
+namespace_prefix "{{ .SyncNSMirroringPrefix }}" {
 {{- else }}
-namespace "{{ .ConsulSyncDestinationNamespace }}" {
+namespace "{{ .SyncConsulDestNS }}" {
 {{- end }}
 {{- end }}
   node_prefix "" {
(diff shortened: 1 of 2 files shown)


subcommand/server-acl-init/rules.go
```diff
@@ -7,10 +7,13 @@ import (
 )
 
 type rulesData struct {
-	EnableNamespaces               bool
```
lkysow, line 10: I renamed these a bit:
* Put Sync or Inject at front of variable name
* Renamed "Namespace" to NS for brevity for the longer names
* Changed `K8SNSMirroring` to `NSMirroring` since it's clear we're mirroring k8s namespaces (what other namespaces would we mirror)

subcommand/server-acl-init/rules.go
```diff
@@ -240,26 +243,37 @@ namespace_prefix "" {
 	return c.renderRules(aclReplicationRulesTpl)
 }
 
-// todo: implement namespaces support.
 func (c *Command) controllerRules() (string, error) {
 	controllerRules := `
 operator = "write"
-node_prefix "" {
```
lkysow, line 247: I realized we don't need `node_prefix ""`

subcommand/server-acl-init/rules.go
```diff
@@ -240,26 +243,37 @@ namespace_prefix "" {
...
+{{- end }}
+  service_prefix "" {
+    policy = "write"
+  }
 }
+{{- else }}
 service_prefix "" {
   policy = "write"
 }
+{{- end }}
```
ashwin-venkatesh, line 263: Cant we just call the `{{- end }}` block before we write the `service_prefix` and we wouldnt have to duplicate that snippet of the config?

subcommand/server-acl-init/rules_test.go
```diff
@@ -504,28 +504,51 @@ func TestControllerRules(t *testing.T) {
...
+namespace_prefix "" {
+  service_prefix "" {
+    policy = "write"
+  }
+}`,
+		},
+		{
+			Name:             "namespaces=enabled, mirroring=true, mirroringPrefix=prefix-",
+			EnableNamespaces: true,
+			Mirroring:        true,
```
ashwin-venkatesh, line 545: This is more of a question than a comment on the PR. Do we have a validation somewhere that fails if `mirroring` is `true` but `EnableNamespaces` is `false?

ashwin-venkatesh (approved):
This looks great. The drying up of the `service_prefix` would be nice.

subcommand/server-acl-init/rules.go
lkysow: Yeah so the other ones do that but it results in the formatting having an extra two spaces in OSS:
```
operator = "write"
  service_prefix "" {
    policy = "write"
  }
```

vs

```
operator = "write"
service_prefix "" {
  policy = "write"
}
```

This shows up in the Consul UI and always kinda bothered me. What do you think though?

subcommand/server-acl-init/rules_test.go
lkysow: No, that's a good idea though. I'm guessing that's best done in the helm chart? It could fail out?

subcommand/server-acl-init/rules.go
ashwin-venkatesh: 👀 The formatting on the first one hurts my soul. I'll resolve this because that formatting bothers me as well.

subcommand/server-acl-init/rules_test.go
lkysow: It is inconsistent to only do it here though... I'll change to be like the rest and add a note on our backlog.

subcommand/server-acl-init/rules.go
lkysow: It is inconsistent to only do it here though... I'll change to be like the rest and add a note on our backlog.

kschoche (approved):
Looks great!!

PR #326, opened 2020-09-17
Ensure system recovers quickly from failures or drift in state

ashwin-venkatesh:
Changes proposed in this PR:
- helm upgrades will cause the caBundle to get reset on the mutating webhooks. By "reconciling" the state of the system every second, we ensure the drift in this state has a minimal impact on the uptime of the system. it will now verify that the certificates as well as the CA bundle are "correct" every second and update them if they arent.

- this will also ensure any edits made to the secret or the webhook configuration, which could lead to downtime in the system, are recovered from within a second.

How I've tested this PR:
- helm installed consul with the image.
- performed a helm upgrade once the system had reached a stable state.
- attempted the creation of a service default and it did not throw an x509 error.

- `kubectl edit` the secret with the certificate and/or the MWC for the webhook certs. The system should recover within a few seconds and the webhook and the webhook configuration should be able to communicate with each other successfully.

Image for testing: 
`ashwinvenkatesh/consul-k8s:webhook-certs@sha256:25b85fd6ccbe7e141cd1541264c40cb15f28c1ba815a4df58c6a3949b583e91f`

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/webhook-cert-manager/command.go
+++ b/subcommand/webhook-cert-manager/command.go
@@ -50,6 +50,7 @@ type Command struct {
 	logger hclog.Logger
 
 	certExpiry *time.Duration // override default cert expiry of 24 hours if set (only set in tests)
+	source     cert.Source    // override default cert source of cert.GenSource if set (only in tests)
 }
 
 func (c *Command) init() {
@@ -148,12 +149,16 @@ func (c *Command) Run(args []string) int {
 	} else {
 		expiry = defaultCertExpiry
 	}
-
+	var certSource cert.Source
 	for _, config := range configs {
-		certSource := &cert.GenSource{
-			Name:   "Consul Webhook Certificates",
-			Hosts:  config.TLSAutoHosts,
-			Expiry: expiry,
+		if c.source != nil {
+			certSource = c.source
+		} else {
+			certSource = &cert.GenSource{
+				Name:   "Consul Webhook Certificates",
+				Hosts:  config.TLSAutoHosts,
+				Expiry: expiry,
+			}
 		}
 		certNotify := &cert.Notify{Source: certSource, Ch: certCh, WebhookConfigName: config.Name, SecretName: config.SecretName, SecretNamespace: config.SecretNamespace}
 		notifiers = append(notifiers, certNotify)
@@ -203,6 +208,9 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.MetaBundle, cl
 	}
 }
 
+// reconcileCertificates ensures the secret in the MetaBundle has the latest certificate from the MetaBundle and the caBundles on the
+// MutatingWebhookConfiguration have the latest CA certificate from the MetaBundle. It updates them if they are outdated and exits early
+// if they are up-to date.
 func (c *Command) reconcileCertificates(ctx context.Context, clientset kubernetes.Interface, bundle cert.MetaBundle, log hclog.Logger) error {
 	iterLog := log.With("mutatingwebhookconfig", bundle.WebhookConfigName, "secret", bundle.SecretName, "secretNS", bundle.SecretNamespace)
 
@@ -260,6 +268,8 @@ func (c *Command) reconcileCertificates(ctx context.Context, clientset kubernete
 	return nil
 }
 
+// updateWebhookConfig iterates over every webhook on the specified webhook configuration and updates
+// their caBundle with the CA from the MetaBundle.
 func (c *Command) updateWebhookConfig(ctx context.Context, metaBundle cert.MetaBundle, clientset kubernetes.Interface) error {
 	if len(metaBundle.CACert) == 0 {
 		return errors.New("no CA certificate in the bundle")
@@ -289,6 +299,8 @@ func (c *Command) updateWebhookConfig(ctx context.Context, metaBundle cert.MetaB
 	return nil
 }
 
+// webhookUpdated verifies if every caBundle on the specified webhook configuration matches the desired CA certificate.
+// It returns true if the CA is up-to date and false if it needs to be updated.
 func (c *Command) webhookUpdated(ctx context.Context, bundle cert.MetaBundle, clientset kubernetes.Interface) bool {
 	webhookCfg, err := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(ctx, bundle.WebhookConfigName, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


subcommand/webhook-cert-manager/command.go
```diff
@@ -180,89 +179,78 @@ func (c *Command) Run(args []string) int {
...
+			// CA bundle which will not be in sync with the certs in the system.
+			// This fast reconcile ensure the system recovers fairly quickly in case
+			// the secret or MWC gets deleted or reset.
 
 		case <-ctx.Done():
 			// Quit
 			return
 		}
 
-		if err := c.reconcileCertificates(ctx, clientset, bundle, log); err != nil {
```
lkysow, line 202: I think I like still keeping this code in `reconcileCertificates` so we don't have those `continue`'s. They seem error prone to me

subcommand/webhook-cert-manager/command.go
```diff
@@ -180,89 +179,78 @@ func (c *Command) Run(args []string) int {
...
-			iterLog.Error(fmt.Sprintf("Error writing secret to API server: %s", err))
-			return err
+			iterLog.Info("Updating webhook configuration")
+			if err = c.updateWebhookConfig(ctx, bundle, clientset); err != nil {
+				iterLog.Error("Error updating webhook configuration")
+				continue
+			}
+			continue
+		} else if err != nil {
+			iterLog.Error(fmt.Sprintf("Error getting secret from Kubernetes: %s", err))
```
lkysow, line 229:
```suggestion
			iterLog.Error("getting secret from Kubernetes", "err", err)
```

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -369,19 +369,16 @@ func TestCertWatcher(t *testing.T) {
 	defer stopCommand(t, &cmd, exitCh)
 
 	ctx := context.Background()
-	// This first sleep is required to ensure the command had completed validating that the MWC does exist in the cluster
-	// It ensures the MWC is absent when the server actually attempts to update it with a bundle.
-	time.Sleep(1 * time.Millisecond)
+	// This sleep ensures the webhook has been updated and the reconcile has been successful.
```
lkysow, line 372: Can you instead loop and check that the mwc has its caCert len > 0? That will remove race conditions.

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -369,19 +369,16 @@ func TestCertWatcher(t *testing.T) {
...
+	time.Sleep(100 * time.Millisecond)
 	err = k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Delete(ctx, webhookName, metav1.DeleteOptions{})
 	require.NoError(t, err)
 
-	// This sleep ensures the cluster attempts to try updating the MWC every second and fails before we re-create it in the cluster.
-	time.Sleep(100 * time.Millisecond)
 	_, err = k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Create(ctx, webhook, metav1.CreateOptions{})
 	require.NoError(t, err)
```
lkysow, line 379: You could patch the webhook and set caCert to "" rather than creating/deleting it.

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -369,19 +369,16 @@ func TestCertWatcher(t *testing.T) {
...
 
-	// If this test passes, it implies that the system has recovered before a new certificate is available
-	// or its default retry timer of 30 minutes has forced the reconcile.
+	// If this test passes, it implies that the system has recovered from the MWC
+	// getting deleted and restored withing a reasonable time window
 	timer := &retry.Timer{Timeout: 10 * time.Second, Wait: 500 * time.Millisecond}
 	retry.RunWith(timer, t, func(r *retry.R) {
 		webhookConfig1, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(ctx, webhookName, metav1.GetOptions{})
```
lkysow, line 384:
```
		require.NotEqual(r, webhookConfig1.Webhooks[0].ClientConfig.CABundle, "")
		// This verifies that the MWC is populated with a valid certificate which has a
		// a len of 800 chars. But the length is not deterministic because it is encoded.
		require.True(r, len(webhookConfig1.Webhooks[0].ClientConfig.CABundle) > 800)
```

Can be
```
		// This verifies that the MWC is populated with a valid certificate which has a
		// a len of 800 chars. But the length is not deterministic because it is encoded.
		require.Len(r, 800, webhookConfig1.Webhooks[0].ClientConfig.CABundle)
```

ashwin-venkatesh: The length of the certificate is non deterministic. I can drop the `""` check but the base64 encoding of the ca bundle varies a little bit.

lkysow: Ahh gothca, but yeah we ca just assert len > 800

subcommand/webhook-cert-manager/command_test.go
lkysow: You need to loop on a `Get` to ensure the first iteration has run through and updated the MWC I think.

ashwin-venkatesh: done 😄

lkysow (approved):
Note: I haven't run this code.

ishustava (commented):
This looks pretty good! I've left some comments about improving the tests, but otherwise, this looks good to me. Let me know what you think!

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -369,27 +369,27 @@ func TestCertWatcher(t *testing.T) {
...
 	retry.RunWith(timer, t, func(r *retry.R) {
-		webhookConfig1, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(ctx, webhookName, metav1.GetOptions{})
+		webhookConfig, err := k8s.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(ctx, webhookName, metav1.GetOptions{})
 		require.NoError(r, err)
-		require.NotEqual(r, webhookConfig1.Webhooks[0].ClientConfig.CABundle, "")
 		// This verifies that the MWC is populated with a valid certificate which has a
-		// a len of 800 chars. But the length is not deterministic because it is encoded.
-		require.True(r, len(webhookConfig1.Webhooks[0].ClientConfig.CABundle) > 800)
+		// a len of at least 800 chars. But the length is not deterministic because it is encoded.
+		require.True(r, len(webhookConfig.Webhooks[0].ClientConfig.CABundle) > 800)
```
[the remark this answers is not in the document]
ashwin-venkatesh: I gave it a whirl. let me know what you think.

ishustava (approved):
Looks great, @ashwin-venkatesh !!

PR #329, opened 2020-09-23
Implement webhook support for consul ent

lkysow:
- if running ent and namespace mirroring is enabled then allow multiple
resources with the same name across namespaces.
- make Validator structs public and remove constructors

How I've tested this PR:
- I haven't

Checklist:
- [x] Tests added

Not sure if we should make this a struct like we did for `ConfigEntryController` and set the EnableNamespaces and EnableNSMirroring as fields.


ashwin-venkatesh (approved):
This was simpler than I was worried it could be. Additionally, we don't need this check for ProxyDefaults given they are global so 🎉

kschoche (approved):
Looks great!

PR #340, opened 2020-09-29
Replace reflect.DeepEqual with gocmp.Equal

ashwin-venkatesh:
- go-cmp has a more robust library for compares as it allows ignoring
  unexported fields.
- Refactored the MatchesConsul tests for serviceDefaults to resemble other MatchesConsul tests

How I've tested this PR:
- existing tests for MatchesConsul continue to pass.

- this is a refactor so existing tests passing should imply it was successful.

### DIFF
--- a/api/v1alpha1/proxydefaults_types.go
+++ b/api/v1alpha1/proxydefaults_types.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/hashicorp/consul/api"
 	capi "github.com/hashicorp/consul/api"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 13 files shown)


api/v1alpha1/proxydefaults_types_test.go
```diff
@@ -86,11 +91,25 @@ func TestProxyDefaults_MatchesConsul(t *testing.T) {
...
+				Name: common.Global,
+				Kind: capi.ProxyDefaults,
+			},
+			Matches: false,
 		},
 	}
 	for name, c := range cases {
 		t.Run(name, func(t *testing.T) {
-			require.True(t, c.Ours.MatchesConsul(c.Theirs))
+			require.Equal(t, c.Ours.MatchesConsul(c.Theirs), c.Matches)
```
lkysow, line 112:
```suggestion
			require.Equal(t,  c.Matches, c.Ours.MatchesConsul(c.Theirs))
```
exp should be before act

lkysow (approved):
👍 after rebase

api/v1alpha1/proxydefaults_types.go
```diff
@@ -123,15 +124,16 @@ func (in *ProxyDefaults) ToConsul() api.ConfigEntry {
...
 	if !ok {
 		return false
 	}
-	proxyDefCand.Namespace = ""
-	proxyDefCand.CreateIndex = 0
-	proxyDefCand.ModifyIndex = 0
+	// Zero out fields from consul that we don't want to compare on.
+	configEntry.Namespace = ""
+	configEntry.CreateIndex = 0
+	configEntry.ModifyIndex = 0
```
ishustava, line 134: Do you think we could use `cmpopts.IgnoreFields` instead? Otherwise, we're modifying the `candidate` param since it's a pointer, and I believe we're later trying to read `ModifyIndex` in the controller.

ishustava (commented):
Looks good! I left one question, but otherwise, this looks good to me.

api/v1alpha1/proxydefaults_types.go
ashwin-venkatesh: TIL that was an option. That was an excellent suggestion!!! Done!

ishustava (approved):
🎉

PR #343, opened 2020-09-30
Use metadata field from configEntry

ashwin-venkatesh:
This will determine if the resource is managed from outside the current cluster

Changes proposed in this PR:
- Add flag `datacenter` that is a required flag for the controller for startup. This string will be used in the meta() field of the config entries managed by Consul
- Add datacenter specific metadata on the Config Entry Meta field before creating the resource in Consul.
This field will be ignore when checking for equality between resources with `cmp.Equal()`.
- A config entry will only be update/deleted from Consul if the metadata on it indicates that it was created in the datacenter the controller is running in.

How I've tested this PR:
- Adding additional acceptance tests to validate this behavior.

- Ideally, one should be able to use the consul CLI to create config entries an installed Consul. These will not have the meta information set on them. Attempting to create a Custom Resource that shares that same name should lead the the resource getting accepted by the cluster, but will fail reconcile and its status should mention that the sync has failed with a reason indicating the Config is managed externally.
- Deleting this resource will not have the side-effect of deleting the Config Entry from Consul.

Checklist:
- [x] Tests added

### DIFF
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -78,7 +78,7 @@ func invalidPathPrefix(path string) bool {
 	return path != "" && !strings.HasPrefix(path, "/")
 }
 
-func getMeta(datacenter string) map[string]string {
+func meta(datacenter string) map[string]string {
 	return map[string]string{
 		common.SourceKey:     common.SourceValue,
 		common.DatacenterKey: datacenter,
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hashicorp/consul-k8s/api/common"
 	capi "github.com/hashicorp/consul/api"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
@@ -75,4 +76,11 @@ func sliceContains(slice []string, entry string) bool {
 
 func invalidPathPrefix(path string) bool {
 	return path != "" && !strings.HasPrefix(path, "/")
+}
+
+func getMeta(datacenter string) map[string]string {
+	return map[string]string{
+		common.SourceKey:     common.SourceValue,
+		common.DatacenterKey: datacenter,
+	}
 }
--- a/controller/configentry_controller.go
+++ b/controller/configentry_controller.go
@@ -115,25 +115,24 @@ func (r *ConfigEntryController) ReconcileEntry(
 				Namespace: r.consulNamespace(req.Namespace, configEntry.ConsulNamespaced()),
 			})
 
-			if err != nil {
+			// Ignore the error where the config entry isn't found in Consul.
+			// It is indicative of desired state.
+			if err != nil && !isNotFoundErr(err) {
 				return ctrl.Result{}, fmt.Errorf("getting config entry from consul: %w", err)
-			}
-
-			// Only delete the resource from Consul if it is owned by our datacenter.
-			if entry.GetMeta()[common.DatacenterKey] == r.DatacenterName {
-				// Our finalizer is present, so we need to delete the config entry
-				// from consul.
-				_, err := r.ConsulClient.ConfigEntries().Delete(configEntry.ConsulKind(), configEntry.Name(), &capi.WriteOptions{
-					Namespace: r.consulNamespace(req.Namespace, configEntry.ConsulNamespaced()),
-				})
-				if err != nil {
-					return ctrl.Result{}, fmt.Errorf("deleting config entry from consul: %w", err)
+			} else if err == nil {
+				// Only delete the resource from Consul if it is owned by our datacenter.
+				if entry.GetMeta()[common.DatacenterKey] == r.DatacenterName {
+					_, err := r.ConsulClient.ConfigEntries().Delete(configEntry.ConsulKind(), configEntry.Name(), &capi.WriteOptions{
+						Namespace: r.consulNamespace(req.Namespace, configEntry.ConsulNamespaced()),
+					})
+					if err != nil {
+						return ctrl.Result{}, fmt.Errorf("deleting config entry from consul: %w", err)
+					}
+					logger.Info("deletion from Consul successful")
+				} else {
+					logger.Info("config entry in Consul was created in another datacenter - skipping delete from Consul", "external-datacenter", entry.GetMeta()[common.DatacenterKey])
 				}
-				logger.Info("deletion from Consul successful")
-			} else {
-				logger.Info("config entry in Consul was created in another datacenter - skipping delete from Consul", "external-datacenter", entry.GetMeta()[common.DatacenterKey])
 			}
-
 			// remove our finalizer from the list and update it.
 			configEntry.RemoveFinalizer(FinalizerName)
 			if err := crdCtrl.Update(ctx, configEntry); err != nil {
--- a/controller/configentry_controller.go
+++ b/controller/configentry_controller.go
@@ -131,6 +131,8 @@ func (r *ConfigEntryController) ReconcileEntry(
 					return ctrl.Result{}, fmt.Errorf("deleting config entry from consul: %w", err)
 				}
 				logger.Info("deletion from Consul successful")
+			} else {
+				logger.Info("config entry in Consul was created in another datacenter - skipping delete from Consul", "external-datacenter", entry.GetMeta()[common.DatacenterKey])
 			}
 
 			// remove our finalizer from the list and update it.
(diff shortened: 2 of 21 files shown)


controller/configentry_controller.go
```diff
@@ -105,15 +111,27 @@ func (r *ConfigEntryController) ReconcileEntry(
...
+				// Our finalizer is present, so we need to delete the config entry
+				// from consul.
+				_, err := r.ConsulClient.ConfigEntries().Delete(configEntry.ConsulKind(), configEntry.Name(), &capi.WriteOptions{
+					Namespace: r.consulNamespace(req.Namespace, configEntry.ConsulNamespaced()),
+				})
+				if err != nil {
+					return ctrl.Result{}, fmt.Errorf("deleting config entry from consul: %w", err)
+				}
+				logger.Info("deletion from Consul successful")
 			}
```
lkysow, line 134:
```suggestion
			} else {
			    logger.Info("config entry in Consul was created by another datacenter - skipping delete from Consul", "other-datacenter", entry.GetMeta()[common.DatacenterKey])
			}
```

controller/configentry_controller.go
```diff
@@ -167,9 +185,16 @@ func (r *ConfigEntryController) ReconcileEntry(
 		return r.syncFailed(ctx, logger, crdCtrl, configEntry, ConsulAgentError, err)
 	}
 
+	// Check if the config entry is managed by our datacenter.
+	// Do not process resource if the entry was not created within our datacenter
+	// as it was created in a different cluster within the federation.
+	if entry.GetMeta()[common.DatacenterKey] != r.DatacenterName {
+		return r.syncFailed(ctx, logger, crdCtrl, configEntry, ExternallyManagedConfigError, errors.New("config entry managed in different datacenter"))
```
lkysow, line 192: Should this error contain the other dc's name?

controller/configentry_controller_test.go
```diff
@@ -1104,3 +1129,228 @@ func TestConfigEntryControllers_setsSyncedToTrue(t *testing.T) {
...
+			for _, configEntry := range c.consulPrereq {
+				written, _, err := consulClient.ConfigEntries().Set(configEntry, nil)
+				req.NoError(err)
+				req.True(written)
+			}
+
+			// We haven't run reconcile yet so we must create the config entry
+			// in Consul ourselves.
+			{
+				// Create the resource with differnt datacenter on metadata
```
lkysow, line 1330:
```suggestion
				// Create the resource with different datacenter on metadata
```

controller/configentry_controller_test.go
```diff
@@ -1104,3 +1129,228 @@ func TestConfigEntryControllers_setsSyncedToTrue(t *testing.T) {
...
+				r := c.reconciler(client, consulClient, logrtest.TestLogger{T: t})
+				resp, err := r.Reconcile(ctrl.Request{
+					NamespacedName: namespacedName,
+				})
+				req.NoError(err)
+				req.False(resp.Requeue)
+
+				entry, _, err := consulClient.ConfigEntries().Get(c.consulKind, c.configEntryResourceWithDeletion.Name(), nil)
+				req.NoError(err)
+				req.Equal(entry.GetMeta()[common.DatacenterKey], "different-datacenter")
```
lkysow, line 1352: Needs to test that it's deleted from kube

controller/configentry_controller_test.go
```diff
@@ -1104,3 +1129,228 @@ func TestConfigEntryControllers_setsSyncedToTrue(t *testing.T) {
...
+			}
+		})
+	}
+}
```
lkysow, line 1356: Also need a test that if it exists in another DC that it doesn't modify the resource in consul and it sets its status accordingly.

subcommand/controller/command.go
```diff
@@ -56,6 +57,8 @@ func (c *Command) init() {
 	c.flagSet.BoolVar(&c.flagEnableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	c.flagSet.StringVar(&c.flagDatacenterName, "datacenter-name", "",
```
lkysow, line 60: Should we use `-datacenter`? That's what Consul uses I think?

controller/configentry_controller_test.go
ashwin-venkatesh: writing that test as we speak :)

controller/configentry_controller.go
ashwin-venkatesh: I on the fence with this one. I guess it is ok to share this information? 

Also, there is a good chance it might not have that info either ie it could just be blank (it would require every "system") that creates a config entry to set the meta fields.

ashwin-venkatesh: Well, i added it, but we should test this a little bit

api/v1alpha1/types.go
```diff
@@ -76,3 +77,10 @@ func sliceContains(slice []string, entry string) bool {
 func invalidPathPrefix(path string) bool {
 	return path != "" && !strings.HasPrefix(path, "/")
 }
+
+func getMeta(datacenter string) map[string]string {
```
lkysow, line 81: Just `meta` or `configEntryMeta` could work too

controller/configentry_controller.go
```diff
@@ -167,9 +187,16 @@ func (r *ConfigEntryController) ReconcileEntry(
 		return r.syncFailed(ctx, logger, crdCtrl, configEntry, ConsulAgentError, err)
 	}
 
+	// Check if the config entry is managed by our datacenter.
+	// Do not process resource if the entry was not created within our datacenter
+	// as it was created in a different cluster within the federation.
```
lkysow, line 192:
```suggestion
	// as it was created in a different cluster which will be managing that config entry.
```

controller/configentry_controller.go
```diff
@@ -167,9 +187,16 @@ func (r *ConfigEntryController) ReconcileEntry(
...
+	if entry.GetMeta()[common.DatacenterKey] != r.DatacenterName {
+		return r.syncFailed(ctx, logger, crdCtrl, configEntry, ExternallyManagedConfigError, errors.New(fmt.Sprintf("config entry managed in different datacenter: %s", entry.GetMeta()[common.DatacenterKey])))
```
lkysow, line 194:
```suggestion
		return r.syncFailed(ctx, logger, crdCtrl, configEntry, ExternallyManagedConfigError, fmt.Errorf("config entry managed in different datacenter: %q", entry.GetMeta()[common.DatacenterKey]))
```

controller/configentry_controller_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
+const DatacenterName = "datacenter"
```
lkysow, line 25:
```suggestion
const datacenterName = "datacenter
```

controller/configentry_controller_test.go
```diff
@@ -1104,3 +1129,489 @@ func TestConfigEntryControllers_setsSyncedToTrue(t *testing.T) {
...
+
+			// We haven't run reconcile yet. We must create the config entry
+			// in Consul ourselves in a different datacenter.
+			{
+				written, _, err := consulClient.ConfigEntries().Set(c.configEntryResource.ToConsul("different-datacenter"), nil)
+				req.NoError(err)
+				req.True(written)
+			}
+
+			// Now run reconcile which should update the entry in Consul.
```
lkysow, line 1330:
```suggestion
			// Now run reconcile which should **not** update the entry in Consul.
```

lkysow (approved):

lkysow:
You'll need to rebase once #344 is merged in. I tried it out and looks like there's a bug in the enterprise logic: https://github.com/hashicorp/consul-k8s/pull/345

controller/configentry_controller.go
```diff
@@ -105,15 +110,29 @@ func (r *ConfigEntryController) ReconcileEntry(
...
-			_, err := r.ConsulClient.ConfigEntries().Delete(configEntry.ConsulKind(), configEntry.Name(), &capi.WriteOptions{
+			// Check to see if consul has config entry with the same name
+			entry, _, err := r.ConsulClient.ConfigEntries().Get(configEntry.ConsulKind(), configEntry.Name(), &capi.QueryOptions{
 				Namespace: r.consulNamespace(req.Namespace, configEntry.ConsulNamespaced()),
 			})
+
 			if err != nil {
-				return ctrl.Result{}, fmt.Errorf("deleting config entry from consul: %w", err)
+				return ctrl.Result{}, fmt.Errorf("getting config entry from consul: %w", err)
+			}
```
lkysow, line 120: So this shouldn't actually error out since if it is deleted then it's actually good.

ashwin-venkatesh: Nice catch. Updated this check to ignore notFound errors (I'm time traveling. Ignore timestamps.)

ashwin-venkatesh: Thanks for helping with finding the bug BTW. I really appreciate it.

controller/configentry_controller.go
```diff
@@ -105,16 +110,31 @@ func (r *ConfigEntryController) ReconcileEntry(
...
 
+			// Ignore the error where the config entry isn't found in Consul.
+			// It is indicative of desired state.
+			if err != nil && !isNotFoundErr(err) {
+				return ctrl.Result{}, fmt.Errorf("getting config entry from consul: %w", err)
+			} else if err == nil {
+				// Only delete the resource from Consul if it is owned by our datacenter.
+				if entry.GetMeta()[common.DatacenterKey] == r.DatacenterName {
+					// Our finalizer is present, so we need to delete the config entry
+					// from consul.
```
lkysow, line 126: Can probably drop this comment. Doesn't make much sense here since we're so deep into the deletion process here.

lkysow (approved):

kschoche (approved):
🦞 Looks great!

PR #348, opened 2020-10-02
Add -log-level flag to controller

lkysow:


ashwin-venkatesh (approved):
🦊

subcommand/controller/command.go
```diff
@@ -231,6 +255,24 @@ func (c *Command) Run(args []string) int {
 	return 0
 }
 
+// toLevel returns the zapcore level  and an error if lvl is not supported.
+func toLevel(lvl string) (zapcore.Level, error) {
```
ishustava, line 259: You could also use the [`UnmarshalText`](https://godoc.org/go.uber.org/zap/zapcore#Level.UnmarshalText) function from the `zapcore` library. Not sure if you saw it and decided not to use it for some reason. The only difference between theirs and your implementation is that theirs also accepts an all-caps string for log level, but I don't think it would be bad if we do too. WDYT?

lkysow: 💯

ishustava (approved):
🎉

subcommand/controller/command.go
lkysow: I also switched to using their constants. The one maybe issue is that they actually support more levels than we've documented but... probably okay?

ishustava: Awesome! Yeah, I think that's fine, given that we might switch it out to hclog.

PR #350, opened 2020-10-06
Don't set envoy flag `--max-obj-name-len`

lkysow:
This flag was deprecated and ignored in Envoy 1.11 and will result
in an error for Envoy 1.15.

How I've tested this PR:
* Run with `ghcr.io/lkysow/consul-k8s-dev:oct06-envoy` against envoy versions 1.14.4, 1.13.4, 1.12.6, 1.11.2

* You can see here (https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.11.0#deprecated) that the flag is deprecated so unless you want to, there's no need to test it yourself that connect still works.

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 ## UNRELEASED
 
+BREAKING CHANGES:
+* Connect: No longer set `--max-obj-name-len` flag when executing `envoy`. This flag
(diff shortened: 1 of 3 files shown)


kschoche (approved):
Looks 🏆

ishustava (approved):

PR #352, opened 2020-10-07
L7 intentions

ashwin-venkatesh:
Changes proposed in this PR:
Add support for L7 intentions

How I've tested this PR:
Local tests. 
Pending: Testing via helm

This PR is mostly just an update to the existing struct. At this point, there is not enough documentation for the fields but that will be updated when consul updates docs for service intentions as config entries.


api/v1alpha1/serviceintentions_types.go
```diff
@@ -175,8 +251,18 @@ func (in *ServiceIntentions) Validate() error {
...
+		}
+		if len(source.Permissions) == 0 {
+			if err := source.Action.validate(path.Child("sources").Index(i)); err != nil {
+				errs = append(errs, err)
+			}
+		} else {
+			if err := source.Permissions.validate(path.Child("sources").Index(i)); err != nil {
+				errs = append(errs, err...)
+			}
 		}
```
lkysow, line 266:
```suggestion
		if len(source.Permissions) == 0 {
			if err := source.Action.validate(path.Child("sources").Index(i)); err != nil {
				errs = append(errs, err)
			}
		} else if err := source.Permissions.validate(path.Child("sources").Index(i)); err != nil {
		        errs = append(errs, err...)
		}
```

lkysow (approved):

api/v1alpha1/serviceintentions_types.go
```diff
@@ -148,13 +178,59 @@ func (in *SourceIntention) toConsul() *capi.SourceIntention {
...
 }
 
 func (in IntentionAction) toConsul() capi.IntentionAction {
 	return capi.IntentionAction(in)
 }
+
+func (in IntentionPermissions) toConsul() []*capi.IntentionPermission {
+	if in == nil {
+		return nil
+	}
```
ishustava, line 193: we probably don't need this since a nil array is an empty array.

api/v1alpha1/serviceintentions_types.go
```diff
@@ -187,6 +272,32 @@ func (in *ServiceIntentions) Validate() error {
...
+
+func (in *IntentionHTTPPermission) validate(path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	if invalidPathPrefix(in.PathPrefix) {
+		errs = append(errs, field.Invalid(path.Child("pathPrefix"), in.PathPrefix, `must begin with a '/'`))
+	}
+	if invalidPathPrefix(in.PathExact) {
+		errs = append(errs, field.Invalid(path.Child("pathExact"), in.PathExact, `must begin with a '/'`))
+	}
+	return errs
```
ishustava, line 298: do you think we need to validate that a valid HTTP method was provided? This is not a blocker by any means, could be added at a later time once we have more docs and stuff from consul.

ishustava (approved):
Looks good! A couple of comments, but nothing blocking.

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: good catch!!

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: I was wondering the same thing. I wasn't sure if casing was a problem or if wild cards were supported which I why I punted on that for now. If the exhaustive list of methods is GET, PUT, POST, PATCH & DELETE (do you know if we support CONNECT, OPTIONS, TRACE and HEAD)??

ashwin-venkatesh: PS: i had to look up MDN to get the exhaustive list of http methods 😛

ishustava: Ah, fair enough! Definitely makes sense to punt! Haha, I don't even know what is MDN 🤯

ashwin-venkatesh: Thanks for the review @ishustava 💥

PR #356, opened 2020-10-09
Add test for command shutting down cleanly

lkysow:
Also update logs and errors


lkysow:
Kyle 🦖 🇹🇭 'd this on zoom

PR #358, opened 2020-10-09
Add missing additional columns

lkysow:
Missing from service-intentions and splitter but on the rest

These show up on
```
$ k get servicesplitter
NAME   AGE
foo    7s
$ k get servicedefaults
NAME                     SYNCED
foo                      True
```


ashwin-venkatesh (approved):
great catch Luke!! 🦊

kschoche (approved):

PR #362, opened 2020-10-15
Support Consul namespaces for service-intentions

lkysow:
Changes proposed in this PR:
- Fix support for service-intentions and Consul enterprise. Previously the defaults we were setting for the namespace fields weren't getting applied to the object in the webhook because we weren't returning any patches. When the entry got into the controller, its destination.namespace field was `""`.  We would then attempt to create a namespace with name `""` which would fail. In addition, if the namespace creation didn't error out, we would create the config entry in the namespace `"default"` instead of what we wanted to be the default namespace (e.g. if mirroring we wanted the namespace to equal the kube namespace).

- In this PR, we reject requests that have source.namespace or destination.namespace specified and consul namespaces are disabled.

- Additionally, if a destination namespace is explicitly specified, and namespaces are enabled, we create the service intention in the specified destination namespace.

- In case the destination namespace is not specified and namespaces are enabled, we use our existing logic to determine the consul namespace, depending on mirroring, mirror-prefix and destination consul namespace values.

- This PR allows validates that atleast one source value is set on an intention.
- It also ensures that exposePaths.protocol does not fail is `""` is set as it's value as that defaulting is managed by Consul.
How I've tested this PR:
- Acceptance tests are passing https://github.com/hashicorp/consul-helm/pull/645 using this image. Previously they were failing for Consul enterprise.

Closes https://github.com/hashicorp/consul-k8s/issues/367

- code, review acceptance tests

Checklist:
- [x] Tests added

### DIFF
--- a/api/v1alpha1/serviceintentions_types_test.go
+++ b/api/v1alpha1/serviceintentions_types_test.go
@@ -1,7 +1,6 @@
 package v1alpha1
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -199,7 +198,8 @@ func TestServiceIntentions_ToConsul(t *testing.T) {
 				},
 				Spec: ServiceIntentionsSpec{
 					Destination: Destination{
-						Name: "svc-name",
+						Name:      "svc-name",
+						Namespace: "dest-ns",
 					},
 					Sources: []*SourceIntention{
 						{
@@ -248,8 +248,9 @@ func TestServiceIntentions_ToConsul(t *testing.T) {
 				},
 			},
 			Exp: &capi.ServiceIntentionsConfigEntry{
-				Kind: capi.ServiceIntentions,
-				Name: "svc-name",
+				Kind:      capi.ServiceIntentions,
+				Name:      "svc-name",
+				Namespace: "dest-ns",
 				Sources: []*capi.SourceIntention{
 					{
 						Name:        "svc1",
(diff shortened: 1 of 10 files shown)


api/v1alpha1/serviceintentions_webhook.go
```diff
@@ -85,10 +90,33 @@ func (v *ServiceIntentionsWebhook) Handle(ctx context.Context, req admission.Req
...
 }
 
 func (v *ServiceIntentionsWebhook) InjectDecoder(d *admission.Decoder) error {
 	v.decoder = d
 	return nil
 }
+
+// defaultingPatches returns the patches needed to set fields to their
+// defaults.
+func (v *ServiceIntentionsWebhook) defaultingPatches(err error, svcIntentions *ServiceIntentions) ([]jsonpatch.Operation, error) {
```
lkysow, line 106: This only returns patches in consul ent but I didn't short-circuit in case we add defaulting logic for oss later.

api/v1alpha1/serviceintentions_webhook_test.go
```diff
@@ -3,18 +3,20 @@ package v1alpha1
...
+	"gomodules.xyz/jsonpatch/v2"
 	"k8s.io/api/admission/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-func TestValidateServiceIntentions_Create(t *testing.T) {
+func TestHandle_ServiceIntentions_Create(t *testing.T) {
```
lkysow, line 19: These tests actually call `Handle()` unlike the other webhook tests that do call `Validate()`

api/v1alpha1/serviceintentions_webhook_test.go
```diff
@@ -398,3 +400,167 @@ func TestValidateServiceIntentions_Update(t *testing.T) {
 		})
 	}
 }
+
+// Test that we return patches to set Consul namespace fields to their defaults.
+// This test also tests OSS where we expect no patches since OSS has no
+// Consul namespaces.
+func TestHandle_ServiceIntentions_Patches(t *testing.T) {
```
lkysow, line 407: I added a new test rather than adding permutations to the existing tests because I thought that would make them too complicated.

api/v1alpha1/serviceintentions_webhook.go
```diff
@@ -41,7 +43,10 @@ func (v *ServiceIntentionsWebhook) Handle(ctx context.Context, req admission.Req
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
-	svcIntentions.Default()
+	defaultingPatches, err := v.defaultingPatches(err, &svcIntentions)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
```
lkysow, line 48: Not sure if this should be 500 or 400. If there's an error, it will happen every time but the error could be due to the input or our internal defaulting logic.

api/v1alpha1/serviceintentions_types_test.go
```diff
@@ -641,10 +643,20 @@ func TestServiceIntentions_Default(t *testing.T) {
...
+		// the input shouldn't change since we don't set any namespace defaults
+		// in OSS.
+		for _, namespacesEnabled := range []bool{false, true} {
+			testName := fmt.Sprintf("%s namespaces=%t", name, namespacesEnabled)
+			t.Run(testName, func(t *testing.T) {
+				testCase.input.Default(namespacesEnabled)
+				if namespacesEnabled {
+					require.True(t, cmp.Equal(testCase.input, testCase.output))
+				} else {
+					require.True(t, cmp.Equal(testCase.input, testCase.input))
```
ashwin-venkatesh, line 656: I think this assertion will always be true.

api/v1alpha1/serviceintentions_webhook.go
ashwin-venkatesh: This comment just stopped me from mentioning we should short circuit it 😛

api/v1alpha1/serviceintentions_webhook.go
ashwin-venkatesh: You're right. If malformed JSON makes it here, it's probably because the API server messed up in some weird way. I'm OK with a 500 here, given i'd expect the API server to return a 400

api/v1alpha1/serviceintentions_webhook_test.go
```diff
@@ -398,3 +400,167 @@ func TestValidateServiceIntentions_Update(t *testing.T) {
...
+			},
+			expPatches: []jsonpatch.Operation{},
+		},
+		"destination.namespace empty": {
+			newResource: &ServiceIntentions{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "foo-intention",
+					Namespace: "bar",
+				},
+				Spec: ServiceIntentionsSpec{
```
ashwin-venkatesh, line 442: So this is kinda unrelated to the changes proposed here, but this test reminded me of it. Consul barfs if the len of []source is 0. We can add that validation as a separate PR and i dont want to block this, but if it isn't too much work, could we handle it in this PR? This is not a blocker on this PR.

ashwin-venkatesh (commented):
LOVE IT!

api/v1alpha1/serviceintentions_types_test.go
lkysow: Isn't that the best kind of assertion?

ashwin-venkatesh: i mean if you phrase it that way 😂

ishustava (approved):
Looks great! I tested it out on a cluster too with both single namespace and namespace mirroring settings.

lkysow (commented):
@ashwin-venkatesh this looks good. I can't approve my own PRs :D

ashwin-venkatesh:
> @ashwin-venkatesh this looks good. I can't approve my own PRs :D

LOL..ill approve it for you ;)

ashwin-venkatesh (approved):
🦊

PR #363, opened 2020-10-15
Health Checks support for consul service mesh

kschoche:
Changes proposed in this PR:
- Adds a new controller to connect-inject which is responsible for synchronizing kubernetes pod health checks with consul service instance health checks. This has the effect of letting Consul know about the pod's readiness states thereby allowing Consul to direct service mesh traffic only to "healthy" pods.
- Adds a new Consul TTL health check for every connect-injected pod. 
- Introduces a new pod label which is applied to connect-injected pods via the mutating webhook: "consul.hashicorp.com/connect-inject-status".
- Includes Upsert/Delete functionality as well as periodic reconcile which does a full reconcile of all pods on a configurable frequency defaulted to "1m".
- Can be disabled/enabled through helm, default is disabled.
- TLS & autoencrypt are only supported.

How I've tested this PR:
Unit tests have been added which cover permutations of Upsert/Reconcile states, acceptance tests are pending in the associated consul-helm PR which covers end-to-end.
Manual tests are also possible when using consul-helm/hc-base to deploy with `connectInject.healthChecks.enabled=true`

Check out the code and run the unit tests

Checklist:
- [x] Tests added


lkysow (approved):
🦊

PR #365, opened 2020-10-23
Fail sync if consul delete returns an error

ashwin-venkatesh:
Changes proposed in this PR:
- In case consul fails while attempting to delete a config entry, update the sync value to being failed. This can happen when there are other config entries that depend on the config entry being deleted. For example, a service-defaults entry being deleted before service-splitter/service-router/service-resolver entries that depend on the service-defaults.
- Add Age column to CRD printer columns

How I've tested this PR: Deployed it in local cluster and viewed the changes. Wasn't sure what a failing test case would look like for this particular error so I would like some feedback on that. Maybe just a standalone test for service-defaults with service splitters? Most of the existing tests test every config entry but that assertion will probably not be the case here.

Checklist:
- [ ] Tests added

### DIFF
--- a/api/v1alpha1/serviceresolver_types.go
+++ b/api/v1alpha1/serviceresolver_types.go
@@ -248,7 +248,7 @@ type ServiceResolverFailover struct {
 	ServiceSubset string `json:"serviceSubset,omitempty"`
 	// Namespace is the namespace to resolve the requested service from to form
 	// the failover group of instances. If empty the current namespace is used.
-	Namespace string `json:"namespaces,omitempty"`
+	Namespace string `json:"namespace,omitempty"`
 	// Datacenters is a fixed list of datacenters to try during failover.
 	Datacenters []string `json:"datacenters,omitempty"`
 }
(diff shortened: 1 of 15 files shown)


api/v1alpha1/serviceresolver_types.go
```diff
@@ -247,7 +248,7 @@ type ServiceResolverFailover struct {
 	ServiceSubset string `json:"serviceSubset,omitempty"`
 	// Namespace is the namespace to resolve the requested service from to form
 	// the failover group of instances. If empty the current namespace is used.
-	Namespace string `json:"namespaces,omitempty"`
+	Namespace string `json:"namespace,omitempty"`
```
kschoche, line 251: 👍🏽

kschoche (approved):
🔥

lkysow (approved):

controller/configentry_controller_test.go
```diff
@@ -2073,3 +2075,112 @@ func TestConfigEntryControllers_doesNotDeleteUnownedConfig(t *testing.T) {
...
+	require.False(t, resp.Requeue)
+
+	resp, err = svcSplitterReconciler.Reconcile(ctrl.Request{NamespacedName: splitterNamespacedName})
+	require.NoError(t, err)
+	require.False(t, resp.Requeue)
+
+	err = client.Get(ctx, defaultsNamespacedName, defaults)
+	require.NoError(t, err)
+
+	// Update service-default with deletion timestamp so that it attempts deletion on reconcile.
```
ishustava, line 2164:
```suggestion
	// Update service-defaults with deletion timestamp so that it attempts deletion on reconcile.
```

ishustava (approved):
Looks great!

Issue #364, opened 2020-10-22
Replication token cannot replicate namespaced tokens


lkysow:
Our ACL replication rules (https://github.com/hashicorp/consul-k8s/blob/master/subcommand/server-acl-init/rules.go#L225-L242)

Contain
```hcl
acl = "write"
```

But not
```hcl
namespace prefix "" {
  acl = "write"
}
```

This means that replication of tokens from non-default namespaces won't work. We should add this to our rules.

PR #370, opened 2020-10-27
Replication token needs acl write on all ns's

lkysow:
If namespaces are enabled, the replication token will need to be able to
replicate all tokens and policies from all namespaces. Without this
functionality, users won't be able to use namespaced
tokens/policies/roles in secondary datacenters (since they won't be
replicated).

Fixes https://github.com/hashicorp/consul-k8s/issues/364

How I've tested this PR:
* Federated two clusters with ACLs enabled
* Created a namespace `luke`
* Created a policy and token in namespace `luke`:
    ```hcl
    acl = "read"
    ```
* On secondary DC, exec'd into server and ran:
    ```bash
    curl -H "X-Consul-Token: 5207f7dd-cfb0-7d7f-b0a5-964295bcb3a1" -k https://localhost:8501/v1/acl/token/self
    ACL not found
    ```
* Modified replication ACL policy to have `namespace_prefix ""`
* Confirmed that (after a 5m delay) that the API call worked

* code review

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+BUG FIXES:
+* Federation: ensure replication ACL token can replicate policies and tokens in Consul namespaces other than `default` (Consul-enterprise only). [[GH-364](https://github.com/hashicorp/consul-k8s/issues/364)]
+
 ## 0.19.0 (October 12, 2020)
 
 FEATURES:
--- a/subcommand/server-acl-init/rules.go
+++ b/subcommand/server-acl-init/rules.go
@@ -222,7 +222,6 @@ func (c *Command) aclReplicationRules() (string, error) {
 	// datacenters during federation since in order to start ACL replication,
 	// we need a token with both replication and agent permissions.
 	aclReplicationRulesTpl := `
-acl = "write"
 operator = "write"
 agent_prefix "" {
   policy = "read"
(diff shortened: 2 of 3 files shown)


subcommand/server-acl-init/rules.go
```diff
@@ -233,6 +232,7 @@ node_prefix "" {
 {{- if .EnableNamespaces }}
 namespace_prefix "" {
 {{- end }}
+  acl = "write"
```
lkysow, line 235: The key is here. Moving this inside the `namespace_prefix ""` block.

ashwin-venkatesh (approved):
🦊

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ IMPROVEMENTS:
 
 BUG FIXES:
 * Connect: use the first secret of type `kubernetes.io/service-account-token` when creating/updating auth method. [[GH-350](https://github.com/hashicorp/consul-k8s/pull/321)]
+* Federation: ensure replication ACL token can replicate policies and tokens in Consul namespaces other than `default` (Consul-enterprise only). [[GH-364](https://github.com/hashicorp/consul-k8s/issues/364)]
```
ndhanushkodi, line 43: Should this be in unreleased?

subcommand/server-acl-init/rules.go
lkysow: yes exactly. If it was `namespace "blah"` then it would be only for `blah` and if it was `namespace_prefix "blah-"` it would be for `blah-.*`

CHANGELOG.md
lkysow: fixed!

ndhanushkodi (approved):

PR #371, opened 2020-10-29
Update tests for health checks acls (and some refactoring)

lkysow:
Also add back enterprise test build tag.

### DIFF
--- a/subcommand/server-acl-init/command_ent_test.go
+++ b/subcommand/server-acl-init/command_ent_test.go
@@ -714,6 +714,27 @@ func TestRun_TokensWithNamespacesEnabled(tt *testing.T) {
 			SecretNames: []string{resourcePrefix + "-acl-replication-acl-token"},
 			LocalToken:  false,
 		},
+		"inject token (deprecated)": {
+			TokenFlags:  []string{"-create-inject-namespace-token", "-enable-namespaces"},
+			PolicyNames: []string{"connect-inject-token"},
+			PolicyDCs:   nil,
+			SecretNames: []string{resourcePrefix + "-connect-inject-acl-token"},
+			LocalToken:  false,
+		},
+		"inject token with namespaces": {
+			TokenFlags:  []string{"-create-inject-token", "-enable-namespaces"},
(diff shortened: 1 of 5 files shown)


subcommand/server-acl-init/command_ent_test.go
```diff
@@ -1,3 +1,5 @@
+// +build enterprise
```
lkysow, line 1: @kschoche without this the enterprise tests run with the oss binary (and they'll all fail)

kschoche: Ahhh completely forgot to switch this back, that explains so much

kschoche (approved):

subcommand/server-acl-init/command_ent_test.go
```diff
@@ -16,105 +17,100 @@ import (
```
lkysow, line 17: You did a great job updating these tests to pass both `-create-inject-auth-method` and `-create-inject-token` however in my opinion looking at it now, it just makes those tests too complicated when it requires an additional `t.Run()` outside of the existing `t.Run()`.

I think we can argue that we can get away with just switching to the new `-create-inject-token` for all the ent tests because we've already tested in the oss tests that the deprecated flag works to create the auth method and because the tradeoff in test complexity isn't worth it.

subcommand/server-acl-init/command_ent_test.go
lkysow: image

I set this in Goland to get everything to build (add the enterprise tag).

subcommand/server-acl-init/command_ent_test.go
kschoche: I was thinking about this when I was doing it, it got ugly fast. I think we can just get one back compatible test to validate namespaces for ent and we'd be OK setting the rest to current valid flags, thoughts?

kschoche (approved):

subcommand/server-acl-init/command_ent_test.go
lkysow: Yeah that would work too.

PR #380, opened 2020-11-05
Set failing check when pod is not yet running

lkysow:
Once the inject init container completes, the service instance is
registered with Consul. We want to immediately register a failing health
check at this time, otherwise the pod will receive traffic even though
its readiness check hasn't passed and its non-init containers might not
even be running yet.

* To test in a cluster, use Helm:
  ```yaml
  global:
    name: consul
    imageK8S: ghcr.io/lkysow/consul-k8s-dev:nov04-early-reg3
  connectInject:
    enabled: true
    healthChecks:
      enabled: true
  server:
    replicas: 1
    bootstrapExpect: 1
  ```
* Then create a Pod that will never start up. The below pod uses an invald image which can never be pulled.

  ```yaml
  apiVersion: v1
  kind: ServiceAccount
  metadata:
    name: static-client
  ---
  apiVersion: v1
  kind: Pod
  metadata:
    name: static-client
    annotations:
      "consul.hashicorp.com/connect-service-upstreams": "static-server:1234"
  spec:
    containers:
      - name: static-client
        image: tutum/l:latest
        command: [ "/bin/sh", "-c", "--" ]
        args: [ "while true; do sleep 30; done;" ]
        ports:
          - containerPort: 8080
            name: http
    serviceAccountName: static-client
  ```
* Notice that the health check is created and its output set to "Pod is pending"

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -10,6 +10,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const InjectInitContainerName = "consul-connect-inject-init"
+
 type initContainerCommandData struct {
 	ServiceName      string
 	ProxyServiceName string
@@ -193,7 +195,7 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
 	}
 
 	return corev1.Container{
-		Name:  "consul-connect-inject-init",
+		Name:  InjectInitContainerName,
 		Image: h.ImageConsul,
 		Env: []corev1.EnvVar{
 			{
--- a/connect-inject/health_check_resource.go
+++ b/connect-inject/health_check_resource.go
@@ -24,6 +24,8 @@ const (
 
 	// kubernetesSuccessReasonMsg will be passed for passing health check's Reason to Consul.
 	kubernetesSuccessReasonMsg = "Kubernetes Health Checks Passing"
+
+	podPendingReasonMsg = "Pod is pending"
 )
 
 type HealthCheckResource struct {
@@ -230,6 +232,13 @@ func (h *HealthCheckResource) getServiceCheck(client *api.Client, healthCheckID 
 // ready state of the pod along with the reason message which will be passed into the Notes
 // field of the Consul health check.
 func (h *HealthCheckResource) getReadyStatusAndReason(pod *corev1.Pod) (string, string, error) {
+	// A pod might be pending if the init containers have run but the non-init
+	// containers haven't reached running state. In this case we set a failing health
+	// check so the pod doesn't receive traffic before it's ready.
+	if pod.Status.Phase == corev1.PodPending {
+		return api.HealthCritical, podPendingReasonMsg, nil
+	}
+
 	for _, cond := range pod.Status.Conditions {
 		var consulStatus, reason string
 		if cond.Type == corev1.PodReady {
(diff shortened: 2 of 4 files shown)


lkysow:
image

connect-inject/container_init.go
```diff
@@ -10,6 +10,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const InjectInitContainerName = "consul-connect-inject-init"
```
lkysow, line 13: Made this a constant so we can use it in the health checks code.

connect-inject/health_check_resource_test.go
```diff
@@ -270,33 +321,6 @@ func TestReconcilePod(t *testing.T) {
 			},
 			"",
 		},
-		{
-			"PodNotRunning will be ignored for processing",
```
lkysow, line 274: This case was removed because we no longer ignore non-running pods.

connect-inject/health_check_resource_test.go
```diff
@@ -366,7 +392,10 @@ func TestReconcilePod(t *testing.T) {
 			require.NoError(err)
 			// Get the agent checks if they were registered.
 			actual := getConsulAgentChecks(t, client)
-			require.True(cmp.Equal(actual, tt.Expected, cmpopts.IgnoreFields(api.AgentCheck{}, ignoredFields...)))
+
+			cmpOpts := cmpopts.IgnoreFields(api.AgentCheck{}, ignoredFields...)
+			require.True(cmp.Equal(actual, tt.Expected, cmpOpts),
+				cmp.Diff(actual, tt.Expected, cmpOpts))
```
lkysow, line 398: This change just ensures that if the `cmp.Equal` fails, we print out the diff.

connect-inject/health_check_resource_test.go
```diff
@@ -432,6 +461,115 @@ func TestReconcile_IgnorePodsWithoutInjectLabel(t *testing.T) {
 	require.Nil(actual)
 }
 
+// Test pod statuses that the reconciler should ignore.
```
lkysow, line 464: Open to opinion on this test case. I logged all the state transitions while developing this feature so thought we may as well encode that in a test case.

connect-inject/health_check_resource.go
```diff
@@ -232,6 +234,13 @@ func (h *HealthCheckResource) getServiceCheck(client *api.Client, healthCheckID
 // ready state of the pod along with the reason message which will be passed into the Notes
 // field of the Consul health check.
 func (h *HealthCheckResource) getReadyStatusAndReason(pod *corev1.Pod) (string, string, error) {
+	// A pod might be pending if the init containers have run but the non-init
+	// containers haven't been started. In this case we set a failing health
```
kschoche, line 238:
```suggestion
	// containers haven't reached running state. In this case we set a failing health
```
I'm not entirely sure how to re-word this but I think we need to instead state that we're going after pods that have not reached PodRunning, the containers have likely "started" but just **haven't reached the point where readiness probes have run.**

lkysow: I think your re-word is good.

ndhanushkodi (commented):
Looks great, I really liked the comments and tests, they really helped give context. I just had one question about whether one of the if conditions captures a unique case we want to handle separately or if we could just let that fall through to being handled later in that method.

lkysow:
https://github.com/hashicorp/consul-helm/pull/674 will ensure the check output shows up in the UI immediatey

connect-inject/health_check_resource.go
```diff
@@ -232,6 +234,13 @@ func (h *HealthCheckResource) getServiceCheck(client *api.Client, healthCheckID
...
+	// check so the pod doesn't receive traffic before it's ready.
+	if pod.Status.Phase == corev1.PodPending {
+		return api.HealthCritical, podPendingReasonMsg, nil
+	}
+
```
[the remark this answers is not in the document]
lkysow: Yeah good question, my idea was that it would be better to say `Pod is pending` than `containers with unready status: [static-client consul-connect-envoy-sidecar consul-connect-lifecycle-sidecar]` because I thought that would be more understandable. For me when I see the latter I would wonder if the readiness checks were failing.

Here's what the pod status looks like when the init container has terminated but it's still pending:

``json

```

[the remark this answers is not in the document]
ndhanushkodi: Thanks for the explanation! Yeah that makes sense, `Pod is pending` sounds like something you would just have to wait for, while `containers with unready status: [static-client consul-connect-envoy-sidecar consul-connect-lifecycle-sidecar]` sounds like an error you'd want to fix.

kschoche (approved):
Fantastic work, this will have a nice UX impact.

ndhanushkodi (approved):

PR #376, opened 2020-11-02
add support of consul namespaces to health checks

kschoche:
[Consul-ENT only]

Changes proposed in this PR:
- connect-inject now adds a new annotation when Mutate is called "consul.hashicorp.com/consul-destination-namespace" when namespaces and mirroring are enabled.
- health checks are created in namespaces if they are supported+used
- existing unit tests are modified a bit to be able to re-use code between oss and consul-ent

How I've tested this PR:
unit tests have been added for the patch functionality as well as health checks resource.

How I expect reviewers to test this PR:
Unit tests.
Alternatively running and end to end test manually by:
1. deploying with consul-helm consul-ent, along with 
```
global:
  imageK8S: kschoche/consul-k8s-dev
  enableConsulNamespaces: true
  name: consul
  image: hashicorp/consul-enterprise:1.9.0-ent-beta1
  tls:
    enabled: true
server:
  enterpriseLicense:
    secretName: 'ent-license'
    secretKey: 'key'
connectInject:
  healthChecks:
    enabled: true
  enabled: true
  consulNamespaces:
    mirroringK8S: true
```
2. Deploy a connect-injected application with k8s readiness probes into a k8s namespace

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

PR #381, opened 2020-11-05
add namespace_prefix to acls rule for connect inject

kschoche:
Changes proposed in this PR:
- update the ACL rules for connect-inject when health checks are enabled alongside namespaces and update the unit tests to reflect this.
NOTE: dependency on the following PR to be merged first #376 

How I've tested this PR:
unit-testing

run the unit tests
it should be sufficient to test with ACLs+namespaces+mirroring enabled against consul-ent.
acceptance test coverage will be added to consul-helm

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/server-acl-init/rules.go
+++ b/subcommand/server-acl-init/rules.go
@@ -217,17 +217,14 @@ node_prefix "" {
 }
 {{- if .EnableNamespaces }}
 namespace_prefix "" {
+{{- end }}
   service_prefix "" {
     policy = "write"
   }
-}
-{{- else }}
-service_prefix "" {
-  policy = "write"
+{{- if .EnableNamespaces }}
 }
 {{- end }}
-{{- end }}
-`
+{{- end }}`
 	return c.renderRules(injectRulesTpl)
 }
 
--- a/subcommand/server-acl-init/rules.go
+++ b/subcommand/server-acl-init/rules.go
@@ -212,12 +212,18 @@ func (c *Command) injectRules() (string, error) {
 operator = "write"
 {{- end }}
 {{- if .EnableHealthChecks }}
+{{- if .EnableNamespaces }}
+namespace_prefix "" {
+{{- end }}
   node_prefix "" {
      policy = "write"
   }
   service_prefix "" {
      policy = "write"
   }
+{{- if .EnableNamespaces }}
+}
+{{- end }}
 {{- end }}
 `
 	return c.renderRules(injectRulesTpl)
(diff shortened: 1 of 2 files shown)


ashwin-venkatesh (approved):
looks great!

subcommand/server-acl-init/rules.go
```diff
@@ -212,16 +212,18 @@ func (c *Command) injectRules() (string, error) {
...
-     policy = "write"
+    policy = "write"
   }
-{{- if .EnableNamespaces }}
+}
+{{- else }}
+service_prefix "" {
+  policy = "write"
 }
 {{- end }}
```
ashwin-venkatesh: Something like:
```
{{- if .EnableHealthChecks }}
node_prefix "" {
  policy = "write"
}
{{- if .EnableNamespaces }}
namespace_prefix "" {
{{- end }}
  service_prefix "" {
    policy = "write"
  }
{{- if .EnableNamespaces }}
}
{{- end }}
```
It seems to have been our approach in other places in this file but I could be talked out of it

subcommand/server-acl-init/rules_test.go
```diff
@@ -518,25 +518,25 @@ operator = "write"`,
...
+  policy = "write"
+}`,
 		},
 		{
 			"Namespaces are enabled, health checks controller enabled",
 			true,
 			true,
 			`
 operator = "write"
+node_prefix "" {
```
ashwin-venkatesh, line 534: The indentation of these tests might need to have their indentations updated if you're ok with the above changes

ashwin-venkatesh (commented):
This looks great. Have a cosmetic suggestion.

subcommand/server-acl-init/rules.go
kschoche: I had initially thought this was a good route, but then we end up with spacing problems on the service_prefix portion, which wouldn't be indented if namespaces were disabled.
I guess the tradeoff is ugly resultant hcl vs slightly more complex generation code.
Thoughts?

lkysow: Yeah this bothers me too 😂 . But let's just follow the existing pattern for now and then maybe we can change it for all of them at some point.

ashwin-venkatesh: Yeah. Im with Luke. Let's be consistent now and we can re-jigger the entire file down the road.

kschoche: alrighty, can do!

ashwin-venkatesh (approved):

lkysow (approved):

PR #382, opened 2020-11-05
Dont attempt to create '*' namespace

ashwin-venkatesh:
Changes proposed in this PR:
- Do not attempt to create a wildcard namespace when verifying if it exists
Interestingly, the consul API does not return `*` namespace in a get request. This leads to us attempting to create the namespace when `*` is provided as a destination namespace for service intentions. This PR ensures we do not attempt to create the namespace when provided.

How I've tested this PR:
- Unit tests
- Created an image which i deployed with consul-helm and tried to create a service intention where the destination intention was "*"

- Code Review

Checklist:
- [x] Tests added

### DIFF
--- a/namespaces/namespaces_test.go
+++ b/namespaces/namespaces_test.go
@@ -71,44 +71,9 @@ func TestEnsureExists_AlreadyExists(tt *testing.T) {
 
 // Test that if the namespace already exists the function succeeds.
 func TestEnsureExists_WildcardNamespace_AlreadyExists(tt *testing.T) {
-	for _, c := range []struct {
-		ACLsEnabled bool
-	}{
-		{
-			ACLsEnabled: true,
-		},
-		{
-			ACLsEnabled: false,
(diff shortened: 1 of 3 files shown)


kschoche (approved):
Awesome work!

namespaces/namespaces_test.go
```diff
@@ -69,6 +69,48 @@ func TestEnsureExists_AlreadyExists(tt *testing.T) {
 	}
 }
 
+// Test that if the namespace already exists the function succeeds.
```
lkysow, line 72: Should we simplify it to

```
// Test that if the namespace already exists the function succeeds.
func TestEnsureExists_WildcardNamespace(tt *testing.T) {
	created, err := EnsureExists(nil, "*", "")
	require.NoError(tt, err)
	require.False(tt, created)
}
```

lkysow (approved):

PR #385, opened 2020-11-06
Add descriptions and missing validation for service-intentions

ashwin-venkatesh:
Changes proposed in this PR:
- Add inline descriptions for the various fields of service-intentions
- Add missing validations for service intentions

How I've tested this PR:
- Unit tests

This was work that was previously skipped as we did not have enough documentation regarding service-intentions.

- Code Review

Checklist:
- [x] Tests added

### DIFF
--- a/api/v1alpha1/serviceintentions_types.go
+++ b/api/v1alpha1/serviceintentions_types.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"encoding/json"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -18,12 +19,21 @@ import (
 
 // ServiceIntentionsSpec defines the desired state of ServiceIntentions
 type ServiceIntentionsSpec struct {
-	Destination Destination      `json:"destination,omitempty"`
-	Sources     SourceIntentions `json:"sources,omitempty"`
+	Destination Destination `json:"destination,omitempty"`
+	// Sources is the list of all intention sources and the authorization granted to those sources.
+	// The order of this list does not matter, but out of convenience Consul will always store this
+	// reverse sorted by intention precedence, as that is the order that they will be evaluated at enforcement time.
+	Sources SourceIntentions `json:"sources,omitempty"`
 }
 
 type Destination struct {
-	Name      string `json:"name,omitempty"`
+	// Name is the destination of all intentions defined in this config entry.
+	// This may be set to the wildcard character (*) to match
+	// all services that don't otherwise have intentions defined.
+	Name string `json:"name,omitempty"`
+	// Namespace specifies the namespace the config entry will apply to.
+	// This may be set to the wildcard character (*) to match all services
+	// in all namespaces that don't otherwise have intentions defined.
 	Namespace string `json:"namespace,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


api/v1alpha1/serviceintentions_types.go
```diff
@@ -18,12 +19,21 @@ import (
 
 // ServiceIntentionsSpec defines the desired state of ServiceIntentions
 type ServiceIntentionsSpec struct {
-	Destination Destination      `json:"destination,omitempty"`
-	Sources     SourceIntentions `json:"sources,omitempty"`
+	Destination Destination `json:"destination,omitempty"`
```
lkysow, line 22: What about Destination?

api/v1alpha1/serviceintentions_types.go
```diff
@@ -296,12 +333,34 @@ func (in IntentionPermissions) validate(path *field.Path) field.ErrorList {
 
 func (in *IntentionHTTPPermission) validate(path *field.Path) field.ErrorList {
 	var errs field.ErrorList
+	if (in.PathExact != "" && in.PathPrefix != "") || (in.PathExact != "" && in.PathRegex != "") || (in.PathPrefix != "" && in.PathRegex != "") {
+		errs = append(errs, field.Invalid(path, in, `At most only one of pathExact, pathPrefix, or pathRegex may be configured.`))
```
lkysow, line 337: The other places we json encode `in`.

api/v1alpha1/serviceintentions_types.go
```diff
@@ -296,12 +333,34 @@ func (in IntentionPermissions) validate(path *field.Path) field.ErrorList {
...
+	}
 	if invalidPathPrefix(in.PathPrefix) {
 		errs = append(errs, field.Invalid(path.Child("pathPrefix"), in.PathPrefix, `must begin with a '/'`))
 	}
 	if invalidPathPrefix(in.PathExact) {
 		errs = append(errs, field.Invalid(path.Child("pathExact"), in.PathExact, `must begin with a '/'`))
 	}
+	for i, method := range in.Methods {
+		methods := []string{"GET", "HEAD", "POST", "PUT", "DELETE", "CONNECT", "OPTIONS", "TRACE", "PATCH"}
```
lkysow, line 346: did you get this from consul?

api/v1alpha1/serviceintentions_types.go
```diff
@@ -296,12 +333,34 @@ func (in IntentionPermissions) validate(path *field.Path) field.ErrorList {
...
+		if !sliceContains(methods, strings.ToUpper(method)) {
```
lkysow, line 347: does consul also not care about casing?

api/v1alpha1/serviceintentions_types.go
```diff
@@ -296,12 +333,34 @@ func (in IntentionPermissions) validate(path *field.Path) field.ErrorList {
...
+	return errs
+}
+func (in IntentionHTTPHeaderPermissions) validate(path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, permission := range in {
+		if (permission.Regex != "" && permission.Suffix != "") || (permission.Regex != "" && permission.Prefix != "") || (permission.Regex != "" && permission.Exact != "") ||
+			(permission.Regex != "" && permission.Present) || (permission.Suffix != "" && permission.Prefix != "") || (permission.Suffix != "" && permission.Exact != "") ||
+			(permission.Suffix != "" && permission.Present) || (permission.Prefix != "" && permission.Exact != "") ||
+			(permission.Prefix != "" && permission.Present) || (permission.Exact != "" && permission.Present) {
+			errs = append(errs, field.Invalid(path.Index(i), in[i], `At most only one of exact, prefix, suffix, regex, or present may be configured.`))
```
lkysow, line 361: Could we count how many strings are non-empty and error if count != 1?

ishustava: I like this idea, but I think we'd need to error if count > 1 since 0 is also a valid count. It seems that consul validates it in this way too.

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: Checked consul. It does want everything to be uppercase. Fixing the validation.

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: Good catch. Updating this check.

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: I did from the docs. But when I looked at the Consul code, they were using headers from the HTTP library so I switched to using those and added a validation to ensure one doesnt repeat a methods as well.

lkysow: Sorry, I mean instead of passing `in` to the field error that ends up getting encoded in error messages as the golang struct in other locations we're json encoding `in` first

ashwin-venkatesh: whoops..this was the response to the question about the accepted header fields. sorry.

yes yes, i updated it to json encode `in` and pass that into the error field 😄

api/v1alpha1/serviceintentions_types.go
```diff
@@ -296,11 +334,73 @@ func (in IntentionPermissions) validate(path *field.Path) field.ErrorList {
...
+		}
+		if permission.Prefix != "" {
+			hdrParts++
+		}
+		if permission.Suffix != "" {
+			hdrParts++
+		}
+		if hdrParts != 1 {
+			errs = append(errs, field.Invalid(path.Index(i), in[i], `At most only one of exact, prefix, suffix, regex, or present may be configured.`))
+		}
```
lkysow, line 403: `in[i]` should be json encoded here.

Unrelated and not a blocker, I think a helper:
```go
func numNotEmpty(ss ...string) int {
	count := 0
	for _, s := range ss {
		if s != "" {
			count++
		}
	}
	return count
}
```

Would clean this up even further.

lkysow (approved):

api/v1alpha1/serviceintentions_types.go
```diff
@@ -296,11 +334,64 @@ func (in IntentionPermissions) validate(path *field.Path) field.ErrorList {
...
+
+func (in IntentionHTTPHeaderPermissions) validate(path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, permission := range in {
+		hdrParts := 0
+		if permission.Present {
+			hdrParts++
+		}
+		hdrParts += numNotEmpty(permission.Exact, permission.Regex, permission.Prefix, permission.Suffix)
+		if hdrParts != 1 {
```
ishustava, line 391: `0` is valid here too.
```suggestion
		if hdrParts > 1 {
```

api/v1alpha1/serviceintentions_types.go
```diff
@@ -296,11 +334,64 @@ func (in IntentionPermissions) validate(path *field.Path) field.ErrorList {
...
+}
+
+func (in IntentionHTTPHeaderPermissions) validate(path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, permission := range in {
+		hdrParts := 0
+		if permission.Present {
+			hdrParts++
+		}
+		hdrParts += numNotEmpty(permission.Exact, permission.Regex, permission.Prefix, permission.Suffix)
```
ishustava, line 390: why couldn't we call `numNotEmpty` with `permission.Present` too like so:
```suggestion
		hdrParts := numNotEmpty(permission.Exact, permission.Regex, permission.Prefix, permission.Suffix, permission.Present)
```

ishustava (commented):
Looking good! I've left a couple of comments about, the main one being allowing `hdrParts` to be `0` as well as `1`. Once that's fixed, it's good to go!

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: `Present` is a bool value and not a string value 😅

ishustava: Oh haha! ignore me then!

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: good catch

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: lol..will do. I think everything else should be resolved now?

api/v1alpha1/serviceintentions_types.go
ashwin-venkatesh: i kinda copied this from consul but i think they check this only if it exists so i miscopied it

ishustava (approved):

PR #388, opened 2020-11-10
Tweak check id and output

lkysow:
Purely aesthetic changes. None of these are required, more of just a conversation starter based on some discrepancies I've noticed with other built-in checks.

* Changed success output from title case `Kubernetes Health Checks
Passing` to sentence case `Kubernetes health checks passing` to match
other built-in check outputs.
* Removed `-ttl` from end of check id because the type of check doesn't
matter for users and is listed as a separate field in the API responses.
* Changed check id to use `namespace/service-id` instead of
`namespace_service-id` to match Consul and Kube's regular formats
* Changed `service-id_kubernetes-health-check` to
`service_id/kubernetes-health-check` for purely asthetic reasons: clashing
`_` and `-`'s looks jarring.

What it would look like (with `/` now instead of `:`):
image

How some of the other built-in checks look:
image

### DIFF
--- a/connect-inject/health_check_resource.go
+++ b/connect-inject/health_check_resource.go
@@ -23,7 +23,7 @@ const (
 	labelInject = "consul.hashicorp.com/connect-inject-status"
 
 	// kubernetesSuccessReasonMsg will be passed for passing health check's Reason to Consul.
-	kubernetesSuccessReasonMsg = "Kubernetes Health Checks Passing"
+	kubernetesSuccessReasonMsg = "Kubernetes health checks passing"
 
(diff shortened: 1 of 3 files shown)


kschoche (approved):
This looks great!
I love the changes wrt the text formatting of the messages.
I also like the `namespace/sid` formatting, it looks much better.
I'm not totally sold on the `service:kubernetes-health-check` part, I'd argue the same that the `:` is also maybe a bit jarring but I could be convinced otherwise.

lkysow:
@kschoche swapped to `/`

ndhanushkodi (approved):
Looks great!! 🎉

PR #389, opened 2020-11-11
Shutdown Subcommands with SIGTERM and SIGINT

lawliet89:
Changes proposed in this PR:

- Kubernetes sends SIGTERM to shut down containers.
- Other tools like Argo also adopts the same behaviour to shut containers down
- Connect Lifecycle Sidecar should also listen to SIGTERM for shutting down

How I've tested this PR:

- *Pending review*

- Start a connect injected pod and check logs for SIGTERM when pod is deleted

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/inject-connect/command_test.go
+++ b/subcommand/inject-connect/command_test.go
@@ -211,7 +211,7 @@ func TestRun_CommandFailsWithInvalidListener(t *testing.T) {
 // command cleanly.
 func TestRun_CommandExitsCleanlyAfterSignal(t *testing.T) {
 
-	t.Run("SIGNIT", testSignalHandling(syscall.SIGINT))
+	t.Run("SIGINT", testSignalHandling(syscall.SIGINT))
 	t.Run("SIGTERM", testSignalHandling(syscall.SIGTERM))
 }
 
--- a/subcommand/inject-connect/command_test.go
+++ b/subcommand/inject-connect/command_test.go
@@ -3,6 +3,7 @@ package connectinject
 import (
 	"fmt"
 	"os"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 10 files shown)


subcommand/lifecycle-sidecar/command.go
```diff
@@ -47,12 +48,12 @@ func (c *Command) init() {
...
 	c.help = flags.Usage(help, c.flagSet)
 
-	// Wait on an interrupt to exit. This channel must be initialized before
+	// Wait on an interrupt or terminate to exit. This channel must be initialized before
 	// Run() is called so that there are no race conditions where the channel
 	// is not defined.
 	if c.sigCh == nil {
 		c.sigCh = make(chan os.Signal, 1)
-		signal.Notify(c.sigCh, os.Interrupt)
+		signal.Notify(c.sigCh, os.Interrupt, syscall.SIGTERM)
```
ashwin-venkatesh, line 56: Update `os.Interrupt` to `syscall.SIGINT` for consistency.
```suggestion
		signal.Notify(c.sigCh, syscall.SIGINT, syscall.SIGTERM)
```

ashwin-venkatesh (commented):
Thanks so much for the PR!! This looks good. I just have one suggestion.

ashwin-venkatesh (commented):
I think we perform similar signal handling in the following files as well:
• webhook-cert-manager/command.go
• inject-connect/command.go
• sync-catalog/command.go

Could you update them as well as a part of this PR so that we are consistent in our approach to signal handling across the repo. Thanks @lawliet89

ashwin-venkatesh (commented):
One last thing, could you include unit tests similar to what we have here: https://github.com/hashicorp/consul-k8s/blob/2b86fdd2d2b7fb601f42fcca70912936e5994dae/subcommand/inject-connect/command_test.go#L211 just to verify this behavior

lawliet89:
@ashwin-venkatesh I have added `SIGTERM` handling to the subcommands you listed and have added tests for all of them.

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -25,6 +26,28 @@ func TestRun_Defaults(t *testing.T) {
...
+
+		ui := cli.NewMockUi()
+		cmd := Command{
+			UI: ui,
+		}
+		// Run async because we need to kill it when the test is over.
+		exitChan := runCommandAsynchronously(&cmd, []string{
+			"-service-config", configFile,
+		})
+		defer stopCommand(t, &cmd, exitChan, sig)
```
ashwin-venkatesh, line 47: This test will need an assertion similar to `inject_connect_test.go`
```
	// Send the interrupt.
	cmd.interrupt()

	// Assert that it exits cleanly or timeout.
	select {
	case exitCode := <-exitChan:
		require.Equal(t, 0, exitCode, ui.ErrorWriter.String())
	case <-time.After(time.Second * 1):
		// Fail if the stopCh was not caught.
		require.Fail(t, "timeout waiting for command to exit")
	}
```

subcommand/sync-catalog/command_test.go
```diff
@@ -81,6 +83,35 @@ func TestRun_Defaults_SyncsConsulServiceToK8s(t *testing.T) {
...
+			logger: hclog.New(&hclog.LoggerOptions{
+				Name:  t.Name(),
+				Level: hclog.Debug,
+			}),
+		}
+
+		exitChan := runCommandAsynchronously(&cmd, []string{
+			"-http-addr", testServer.HTTPAddr,
+		})
+		defer stopCommand(t, &cmd, exitChan, sig)
```
ashwin-venkatesh, line 111: This test will need as assert. Something like
```
	// Send the interrupt.
	cmd.interrupt()

	// Assert that it exits cleanly or timeout.
	select {
	case exitCode := <-exitChan:
		require.Equal(t, 0, exitCode, ui.ErrorWriter.String())
	case <-time.After(time.Second * 1):
		// Fail if the stopCh was not caught.
		require.Fail(t, "timeout waiting for command to exit")
	}
```

subcommand/webhook-cert-manager/command_test.go
```diff
@@ -18,6 +19,74 @@ import (
...
+		require.NoError(t, err)
+		defer os.Remove(file.Name())
+
+		_, err = file.Write([]byte(configFile))
+		require.NoError(t, err)
+
+		exitCh := runCommandAsynchronously(&cmd, []string{
+			"-config-file", file.Name(),
+		})
+		defer stopCommand(t, &cmd, exitCh, sig)
```
ashwin-venkatesh, line 86: same comment as above
```
	// Send the interrupt.
	cmd.interrupt()

	// Assert that it exits cleanly or timeout.
	select {
	case exitCode := <-exitChan:
		require.Equal(t, 0, exitCode, ui.ErrorWriter.String())
	case <-time.After(time.Second * 1):
		// Fail if the stopCh was not caught.
		require.Fail(t, "timeout waiting for command to exit")
	}
```

ashwin-venkatesh (commented):
This looks good. Thanks for adding the tests but they will need some assertions on them to ensure the signal handling works as expected.

lawliet89:
@ashwin-venkatesh I have added the assertions you suggested. I gave sync catalog more time because it doesn't seem to be able to exit within 1s.

subcommand/inject-connect/command_test.go
```diff
@@ -206,38 +207,46 @@ func TestRun_CommandFailsWithInvalidListener(t *testing.T) {
...
-		UI:        ui,
-		clientset: k8sClient,
-	}
-	ports := freeport.MustTake(1)
+func TestRun_CommandExitsCleanlyAfterSignal(t *testing.T) {
 
-	// NOTE: This url doesn't matter because Consul is never called.
-	os.Setenv(api.HTTPAddrEnvName, "http://0.0.0.0:9999")
-	defer os.Unsetenv(api.HTTPAddrEnvName)
+	t.Run("SIGNIT", testSignalHandling(syscall.SIGINT))
```
ashwin-venkatesh, line 214:
```suggestion
	t.Run("SIGINT", testSignalHandling(syscall.SIGINT))
```

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -25,6 +26,37 @@ func TestRun_Defaults(t *testing.T) {
...
+
+		ui := cli.NewMockUi()
+		cmd := Command{
+			UI: ui,
+		}
+		// Run async because we need to kill it when the test is over.
+		exitChan := runCommandAsynchronously(&cmd, []string{
+			"-service-config", configFile,
+		})
+		cmd.sendSignal(sig)
```
ashwin-venkatesh, line 47: If we are using `cmd.sendSignal(sig)` in the tests, I dont think we need to change the signature of `stopCommand()` to add a signal right? `stopCommand()` can continue to invoke `cmd.interrupt()`

subcommand/sync-catalog/command_test.go
```diff
@@ -81,6 +83,44 @@ func TestRun_Defaults_SyncsConsulServiceToK8s(t *testing.T) {
...
+		exitChan := runCommandAsynchronously(&cmd, []string{
+			"-http-addr", testServer.HTTPAddr,
+		})
+		cmd.sendSignal(sig)
+
+		// Assert that it exits cleanly or timeout.
+		select {
+		case exitCode := <-exitChan:
+			require.Equal(t, 0, exitCode, ui.ErrorWriter.String())
+		case <-time.After(time.Second * 10):
```
ashwin-venkatesh, line 117: Does it really take up to 10 seconds sometimes? Can you add a comment here describing why it is `10s` here as compared to other tests

subcommand/sync-catalog/command.go
```diff
@@ -9,9 +9,10 @@ import (
 	"os/signal"
 	"regexp"
 	"sync"
+	"syscall"
 	"time"
 
-	"github.com/deckarep/golang-set"
+	mapset "github.com/deckarep/golang-set"
```
ashwin-venkatesh, line 15: Can this change be unset please

ashwin-venkatesh (commented):
Ok I have the last few suggestions but I think we will be in a position to merge this once those changes are made! Thanks @lawliet89

subcommand/sync-catalog/command.go
lawliet89: It's automatically added by `goimports`. I've removed it.

subcommand/lifecycle-sidecar/command_test.go
lawliet89: Done. And also undid the changes to the other `stopCommand`s I've made previously.

ashwin-venkatesh (approved):
🥳 Excellent!!

ashwin-venkatesh:
Thanks @lawliet89 !!

PR #400, opened 2020-11-20
Add -log-level flag to inject-connect, healthchecks debug logs

lkysow:
Changes proposed in this PR:
- Add `-log-level` flag to `inject-connect` so we can configure the health check log level
- Add debug logs to health checks so we can see when it's setting a check status
- Move logger construction to `common` so it can be reused

How I've tested this PR:
- I haven't yet!

- code

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/delete-completed-job/command.go
+++ b/subcommand/delete-completed-job/command.go
@@ -4,13 +4,12 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"os"
 	"sync"
 	"time"
 
 	"github.com/hashicorp/consul-k8s/subcommand"
+	"github.com/hashicorp/consul-k8s/subcommand/common"
 	"github.com/hashicorp/consul-k8s/subcommand/flags"
-	"github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/cli"
 	v1 "k8s.io/api/batch/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -18,6 +17,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
+
+const logLevel = "info"
 
 // Command is the command for deleting completed jobs.
 type Command struct {
@@ -95,10 +96,11 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
-	logger := hclog.New(&hclog.LoggerOptions{
-		Level:  hclog.Info,
-		Output: os.Stderr,
-	})
+	logger, err := common.Logger(logLevel)
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
 
 	// Wait for job to complete.
 	logger.Info(fmt.Sprintf("waiting for job %q to complete successfully", jobName))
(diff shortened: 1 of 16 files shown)


subcommand/delete-completed-job/command.go
```diff
@@ -95,10 +94,7 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
-	logger := hclog.New(&hclog.LoggerOptions{
-		Level:  hclog.Info,
-		Output: os.Stderr,
-	})
+	logger := hclog.Default()
```
lkysow, line 97: This is equivalent.

subcommand/lifecycle-sidecar/command.go
```diff
@@ -142,10 +143,6 @@ func (c *Command) validateFlags() error {
 	if err != nil {
 		return fmt.Errorf("-consul-binary %q not found: %s", c.flagConsulBinary, err)
 	}
-	logLevel := hclog.LevelFromString(c.flagLogLevel)
```
lkysow, line 145: No need to validate this here since it's done now via `common.Logger()`

kschoche (approved):
looks great!

subcommand/lifecycle-sidecar/command.go
ishustava: We left some of these tests the other commands, e.g. `create-federation-secret`. I think it's useful to keep these tests because they provide additional validation that commands actually check and return the error in the UI.

subcommand/delete-completed-job/command.go
ishustava: Should we use the constructor here?

ishustava (approved):
Looks good! Left a couple of comments, but they are not blocking.

subcommand/delete-completed-job/command.go
lkysow: You mean swap it back to hclog.New() or use `common.Logger("info")?

subcommand/lifecycle-sidecar/command.go
lkysow: This isn't a test though, this was called in `validateFlags()`:
```go
	err := c.validateFlags()
	if err != nil {
		c.UI.Error("Error: " + err.Error())
		return 1
	}

	logger, err := common.Logger(c.flagLogLevel)
	if err != nil {
		c.UI.Error(err.Error())
		return 1
	}

```

But now that `common.Logger` validates `c.flagLogLevel` it doesn't need to also be validate here.

subcommand/delete-completed-job/command.go
ishustava: use `common.Logger(hclog.Info)` for consistency?

subcommand/lifecycle-sidecar/command.go
ishustava: Oh sorry! please ignore my comment then 😄 🤦‍♀️

subcommand/delete-completed-job/command.go
ishustava: Not a huge deal for me one way or the other btw.

lkysow: Good idea because we might add more to that function and expect it to apply to all loggers.

PR #403, opened 2020-11-25
Set -envoy-version for Consul connect

lkysow:
- Make flags -envoy-image and -consul-image required because the Helm
chart has set them for a number of releases now (-consul-image since 2 years ago and -envoy-image since 9 months ago) and because their
defaults if not set were out of date.
- Try and parse Envoy version from -envoy-image tag, e.g.
  `envoy:1.16.0` will result in `1.16.0`.
- Add new flag -envoy-version that can be used to set the Envoy version
  if it can't be parsed from -envoy-image tag.

How I've tested this PR:
Ran with
```yaml
global:
  imageK8S: "ghcr.io/lkysow/consul-k8s-dev:nov24"
  name: consul
server:
  replicas: 1
  bootstrapExpect: 1
connectInject:
  enabled: true
```
and acceptance tests: https://github.com/hashicorp/consul-helm/pull/706

In addition to the integration tests passing, if you run with this image you'll see the envoy version set in the `bootstrap.yaml` envoy file:
```
k exec static-client -c consul-connect-envoy-sidecar -- cat /consul/connect-inject/envoy-bootstrap.yaml | jq .node.metadata
{
  "namespace": "default",
  "envoy_version": "1.16.0"
}
```

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 IMPROVEMENTS:
 * Connect: Add `-log-level` flag to `inject-connect` command. [[GH-400](https://github.com/hashicorp/consul-k8s/pull/400)]
 
+BREAKING CHANGES:
+* Connect: the flags `-envoy-image` and `-consul-image` for command `inject-connect` are now required. [[GH-403](https://github.com/hashicorp/consul-k8s/pull/403)]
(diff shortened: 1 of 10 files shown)


lkysow:
Closing because we've determined we don't need to set `-envoy-version` because Consul will default to the latest supported envoy version.

PR #406, opened 2020-11-26
Document how to CRD. Add helm copier.

lkysow:
- Add how to create CRDs to contributing.md.
- Add a script that will copy new CRDs to consul-helm in the format
expected. This will reduce manual copying errors.

I've tested this by running through the readme to create ingress and terminating gateway CRDs and the script was used for https://github.com/hashicorp/consul-helm/pull/714

Reviewers should review the script and suggest improvements to the instructions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,3 +57,327 @@ If the changes in your PR do not conflict with any of the existing code in the p
 automatic rebasing when the PR is accepted into the code. However, if there are conflicts (there will be
 a warning on the PR that reads "This branch cannot be rebased due to conflicts"), you will need to manually
 rebase the branch on master, fixing any conflicts along the way before the code can be merged.
+
+## Creating a new CRD
+
+### The Structs
+1. Run the generate command:
+    ```bash
+    operator-sdk create api --group consul --version v1alpha1 --kind IngressGateway --controller --namespaced=true --make=false --resource=true
+    ```
+1. Re-order the file so it looks like:
+    ```go
+    func init() {
+    	SchemeBuilder.Register(&IngressGateway{}, &IngressGatewayList{})
+    }
+    
+    // +kubebuilder:object:root=true
+    // +kubebuilder:subresource:status
+    
+    // IngressGateway is the Schema for the ingressgateways API
+    type IngressGateway struct {
+    	metav1.TypeMeta   `json:",inline"`
+    	metav1.ObjectMeta `json:"metadata,omitempty"`
+    
+    	Spec   IngressGatewaySpec   `json:"spec,omitempty"`
+    	Status IngressGatewayStatus `json:"status,omitempty"`
+    }
+    
+    // +kubebuilder:object:root=true
+    
+    // IngressGatewayList contains a list of IngressGateway
+    type IngressGatewayList struct {
+    	metav1.TypeMeta `json:",inline"`
+    	metav1.ListMeta `json:"metadata,omitempty"`
+    	Items           []IngressGateway `json:"items"`
+    }
+    
+    // IngressGatewaySpec defines the desired state of IngressGateway
+    type IngressGatewaySpec struct {
+    	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+    	// Important: Run "make" to regenerate code after modifying this file
+    
+    	// Foo is an example field of IngressGateway. Edit IngressGateway_types.go to remove/update
+    	Foo string `json:"foo,omitempty"`
+    }
+    
+    // IngressGatewayStatus defines the observed state of IngressGateway
+    type IngressGatewayStatus struct {
+    	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+    	// Important: Run "make" to regenerate code after modifying this file
+    }
+    ```
+1. Add kubebuilder status metadata to the `IngressGateway struct`:
+    ```go
+    // ServiceRouter is the Schema for the servicerouters API
+    // +kubebuilder:printcolumn:name="Synced",type="string",JSONPath=".status.conditions[?(@.type==\"Synced\")].status",description="The sync status of the resource with Consul"
+    // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="The age of the resource"
+    type ServiceRouter struct {
+    ```
+1. Delete `IngressGatewayStatus` struct. We use a common status struct.
+1. Use the `Status` struct instead and embed it:
+    ```diff
+    // IngressGateway is the Schema for the ingressgateways API
+    type IngressGateway struct {
+    	metav1.TypeMeta   `json:",inline"`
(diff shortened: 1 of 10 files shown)


config/webhook/manifests.v1beta1.yaml
```diff
@@ -7,7 +7,6 @@ metadata:
   name: mutating-webhook-configuration
 webhooks:
 - clientConfig:
-    caBundle: Cg==
```
lkysow, line 10: Removed by `0.4.1`: https://github.com/kubernetes-sigs/controller-tools/pull/495

config/crd/bases/consul.hashicorp.com_proxydefaults.yaml
```diff
@@ -43,7 +43,8 @@ spec:
           properties:
             config:
               description: Config is an arbitrary map of configuration values used by Connect proxies. Any values that your proxy allows can be configured globally here. Supports JSON config values. See https://www.consul.io/docs/connect/proxies/envoy#configuration-formatting
-              type: Any
+              format: byte
```
lkysow, line 46: The new version of the controller-tools changes this. `Any` was a bug actually: https://github.com/kubernetes-sigs/controller-tools/pull/505. It was never supported. Our CRD in helm use `object`. I've modified the hack script to swap this when copying over to `consul-helm`.

CONTRIBUTING.md
```diff
@@ -57,3 +57,325 @@ If the changes in your PR do not conflict with any of the existing code in the p
...
+   what validation needs to be done.
+   
+   Things to keep in mind:
+   1. Re-use the `sliceContains` and `notInSliceMessage` helper methods where applicable.
+   1. If the invalid field is an entire struct, encode as json (look for `asJSON` for an example).
+   1. `validateNamespaces` should be a separate method.
+   1. If the field can have a `nil` pointer, check for that, e.g.
+        ```go
+        func (in *ServiceRouteHTTPMatchHeader) validate(path *field.Path) *field.Error {
+            if in == nil {
```
kschoche, line 180: should this be `in` or `path` ?

CONTRIBUTING.md
```diff
@@ -57,3 +57,325 @@ If the changes in your PR do not conflict with any of the existing code in the p
...
+            t.Run(name, func(t *testing.T) {
+                require.Equal(t, c.Matches, c.Ours.MatchesConsul(c.Theirs))
+            })
+        }
+    }
+    ```
+1. The test for `ToConsul` will re-use the same struct as for the previous test cases
+   `empty field matches` and `all fields set matches` but renamed to `empty fields` and `every field set`
+   and with any `Namespace` set on the Consul struct removed and `CreateIndex`/`ModifyIndex` removed
+   since the `ToConsul` method won't set those.
```
kschoche, line 214: This is a mouthful, wondering if it can be rephrased with bullets for keep/remove?

CONTRIBUTING.md
```diff
@@ -57,3 +57,325 @@ If the changes in your PR do not conflict with any of the existing code in the p
...
+            - UPDATE
+          resources:
+            - ingressgateways
+      sideEffects: None
+    ```
+1. Update `templates/controller-clusterrole.yaml` to allow the controller to
+   manage your resource type.
+
+### Test
+1. Build a Docker image for consul-k8s
```
kschoche, line 327:
```suggestion
1. Build a Docker image for consul-k8s via `make dev-docker` and tagging your image appropriately.
```

hack/crds-to-consul-helm/main.go
```diff
@@ -0,0 +1,87 @@
+// Script to move generated CRD yaml into consul-helm and modify it to match
+// the expected consul-helm format.
```
kschoche, line 2: just a thought but it might be work mentioning somewhere in here that the binary (appears to be) expects to be run from CWD and not from top level consul-helm? Maybe I'm not reading the code correctly

kschoche (approved):
Looks awesome, great work, I cant wait to follow it the next time I'm building a new resource!

CONTRIBUTING.md
lkysow: It needs to be `in` because if the CRD is:
```
spec:
  routes:
    - match:
        pathPrefix: ....
```

Then `ServiceRouter.Routes[0].match.header` will be `nil` but we blindly call
```go
	errs = append(errs, in.Match.validate(path.Child("match"))...)
```

So we need to check inside `validate` if our pointer caller is `nil`. We could also check outside of the call to `validate()` but this is the pattern we've followed.

`path` won't be `nil` because we're always passing it properly.

hack/crds-to-consul-helm/main.go
lkysow: Yeah thanks for calling this out. I think the better pattern is to call this through the `Makefile`. I've pushed a new commit with that change. Then you can run `make ctrl-crd-copy helm=../consul-helm` and it will work.

kschoche: Ah fantastic idea, love it!

CONTRIBUTING.md
lkysow: 👍 split it up

ndhanushkodi (approved):
This is such a helpful change, thanks for taking the time to document all this!

CONTRIBUTING.md
```diff
@@ -57,3 +57,327 @@ If the changes in your PR do not conflict with any of the existing code in the p
...
+1. Use the `Status` struct instead and embed it:
+    ```diff
+    // IngressGateway is the Schema for the ingressgateways API
+    type IngressGateway struct {
+    	metav1.TypeMeta   `json:",inline"`
+    	metav1.ObjectMeta `json:"metadata,omitempty"`
+    
+    	Spec   IngressGatewaySpec   `json:"spec,omitempty"`
+    -	Status IngressGatewayStatus `json:"status,omitempty"`
+    +	Status `json:"status,omitempty"`
```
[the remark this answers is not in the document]
lkysow: So this is actually our own Status object defined here: https://github.com/hashicorp/consul-k8s/blob/master/api/v1alpha1/status.go#L71

CONTRIBUTING.md
```diff
@@ -57,3 +57,325 @@ If the changes in your PR do not conflict with any of the existing code in the p
...
+### Spec Methods
+1. Run `make ctrl-generate` to implement the deep copy methods.
+1. Implement all the methods for `ConfigEntryResource` in the `_types.go` file. If using goland you can
+   automatically stub out all the methods by using Code -> Generate -> IngressGateway -> ConfigEntryResource.
+1. Use existing implementations of other types to implement the methods. We have to
+   copy their code because we can't use a common struct that implements the methods
+   because that messes up the CRD code generation. 
+   
+   You should be able to follow the other "normal" types. The non-normal types
+   are `ServiceIntention` and `ProxyDefault` because they have special behaviour
```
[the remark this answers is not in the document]
lkysow: 😂 I should probably use the American spelling for consistency with ya'll 😄

PR #409, opened 2020-12-16
fix signal handling for lifecycle sidecar

kschoche:
This fixes issues related to stale service registrations created when pods are terminated, causing services to be registered with no pods backing them. This was caused by a race condition in shutdown logic of the lifecycle-sidecar, so the fix address shutdown path for it.

Changes proposed in this PR:
- move signal handling to a go func in lifecycle sidecar so that it can exit cleanly and avoid race conditions if resync were to take a long time.

How I've tested this PR:
existing unit tests + manual tests

Spin up latest release of consul-k8s and deploy a sample app that is connect injected, it is helpful to see the issue in the UI so enable that as well. Once the app is online delete the application, you can watch the logs of the lifecycle-sidecar for the pod side by side with the agent logs and see that lifecycle sidecar runs a register after the pod de-registers its service, or see that in the UI the service never goes away.

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/lifecycle-sidecar/command.go
+++ b/subcommand/lifecycle-sidecar/command.go
@@ -1,6 +1,7 @@
 package subcommand
 
 import (
+	"context"
 	"errors"
 	"flag"
 	"fmt"
@@ -90,6 +91,18 @@ func (c *Command) Run(args []string) int {
 	c.consulCommand = append(c.consulCommand, c.parseConsulFlags()...)
 	c.consulCommand = append(c.consulCommand, c.flagServiceConfig)
 
+	// ctx that we pass in to the main work loop, signal handling is handled in another thread
+	// due to the length of time it can take for the cmd to complete causing synchronization issues
+	// on shutdown. Also passing a context in so that it can interrupt the cmd and exit cleanly.
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	go func() {
+		select {
+		case sig := <-c.sigCh:
+			logger.Info(fmt.Sprintf("%s received, shutting down", sig))
+			cancelFunc()
+			return
+		}
+	}()
 	// The main work loop. We continually re-register our service every
 	// syncPeriod. Consul is smart enough to know when the service hasn't changed
 	// and so won't update any indices. This means we won't be causing a lot
@@ -98,22 +111,24 @@ func (c *Command) Run(args []string) int {
 	//
 	// The loop will only exit when the Pod is shut down and we receive a SIGINT.
 	for {
-		cmd := exec.Command(c.flagConsulBinary, c.consulCommand...)
+		start := time.Now()
+		cmd := exec.CommandContext(ctx, c.flagConsulBinary, c.consulCommand...)
 
 		// Run the command and record the stdout and stderr output
 		output, err := cmd.CombinedOutput()
+		// Currently this command is taking >7s, need to investigate
+		logger.Debug("time to run resync:", "time", time.Since(start))
+
 		if err != nil {
 			logger.Error("failed to sync service", "output", strings.TrimSpace(string(output)), "err", err)
 		} else {
 			logger.Info("successfully synced service", "output", strings.TrimSpace(string(output)))
 		}
-
-		// Re-loop after syncPeriod or exit if we receive interrupt or terminate signals.
 		select {
+		// Re-loop after syncPeriod or exit if we receive interrupt or terminate signals.
 		case <-time.After(c.flagSyncPeriod):
 			continue
-		case sig := <-c.sigCh:
-			logger.Info(fmt.Sprintf("%s received, shutting down", sig))
+		case <-ctx.Done():
 			return 0
 		}
 	}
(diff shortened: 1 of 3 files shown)


subcommand/lifecycle-sidecar/command.go
```diff
@@ -98,22 +111,24 @@ func (c *Command) Run(args []string) int {
 	//
 	// The loop will only exit when the Pod is shut down and we receive a SIGINT.
 	for {
-		cmd := exec.Command(c.flagConsulBinary, c.consulCommand...)
+		start := time.Now()
```
ashwin-venkatesh, line 114: Clarifying question: Are we creating a `start` time for solely for debugging or do we intend on utilizing it in for something else as well?

subcommand/lifecycle-sidecar/command.go
```diff
@@ -98,22 +111,24 @@ func (c *Command) Run(args []string) int {
...
+		cmd := exec.CommandContext(ctx, c.flagConsulBinary, c.consulCommand...)
```
ashwin-venkatesh, line 115: I like this!!

ashwin-venkatesh (commented):
I really like the updated pattern for signal handling and passing a context into `exex.CommandContext()`.

I might be out of the loop here, but does this address a specific problem we are running into? Also, given we don't really follow a pattern of logging time taken by a command, I'm unsure if this is something we would want to merge vs use to identify root cause and fix independently.

subcommand/lifecycle-sidecar/command.go
kschoche: This was purely for debugging, I can back it out, but right now it's on the order of 7-10s to run the command (possibly a bug in consul?) with a single registered service, so I was just printing it out.

kschoche:
> I might be out of the loop here, but does this address a specific problem we are running into? Also, given we don't really follow a pattern of logging time taken by a command, I'm unsure if this is something we would want to merge vs use to identify root cause and fix independently.

Sorry about that, I feel silly for not expanding in the PR.

This addresses the issues that we've been seeing with services in the UI showing up after the application has been deleted. What was happening was that an application would get deleted, the pods would terminate and their preStop hooks would deregister the service, (normal), however due to the time between when we're running the `cmd.CombinedOutput()` and when we actually catch the term signal for lifecycle sidecar we would actually re-register the pod on the way out nearly every time!

This fixes the race condition between lifecycle-sidecar getting a term signal and pod de/re-registration.

There is also a past known issue where we suspected pods were not executing their `preStop` hooks due  
abnormal termination scenarios from k8s, which occasionally left the services registered with Consul even though the pod was gone. This looks+reads+smells like the actual cause of that problem as opposed to "abnormal termination."
The underlying issue that caused all of this to surface in the latest release is the abnormally long time it is currently taking for our register command to complete, which means we are not able to catch our term signal about 50% of the time without re-registering the pod.

Let me know if that doesn't make sense and I can demo it to you, and I'll also add steps to repro to the PR.

subcommand/lifecycle-sidecar/command.go
lkysow: I think it's good to keep this but we could put it in with the existing `failed to sync service` and `successfully synced service` log entries:

```go
		if err != nil {
			logger.Error("failed to sync service", "output", strings.TrimSpace(string(output)), "err", err, "duration", time.Since(start))
		} else {
			logger.Info("successfully synced service", "output", strings.TrimSpace(string(output)), "duration", time.Since(start))
		}
```

subcommand/lifecycle-sidecar/command.go
```diff
@@ -98,22 +111,24 @@ func (c *Command) Run(args []string) int {
...
 
 		// Run the command and record the stdout and stderr output
 		output, err := cmd.CombinedOutput()
+		// Currently this command is taking >7s, need to investigate
+		logger.Debug("time to run resync:", "time", time.Since(start))
```
lkysow, line 120: I know this was just for debugging but just making a note that we should clean it up.

lkysow (commented):
This is looking good. Like you mentioned, we also need a test case. We should be able to start the command, send the signal and see that the service isn't re-registered in consul.

ashwin-venkatesh (approved):
Nice work!

subcommand/lifecycle-sidecar/command_test.go
```diff
@@ -36,10 +36,19 @@ func testRunSignalHandling(sig os.Signal) func(*testing.T) {
...
+
 		ui := cli.NewMockUi()
 		cmd := Command{
 			UI: ui,
 		}
+
+		client, err := api.NewClient(&api.Config{
+			Address: a.HTTPAddr,
+		})
+		require.NoError(t, err)
```
lkysow, line 51: This test still passes if I revert the changes made because the command isn't given the http-addr so it can't register the services regardless of the signal handling.

We just need to add `-http-addr` and `-sync-period` flags to the line below:
```go
		exitChan := runCommandAsynchronously(&cmd, []string{
			"-service-config", configFile,
			"-http-addr", a.HTTPAddr,
			"-sync-period", "1s",
		})
		cmd.sendSignal(sig)
```

And then the test fails without the changes and passes with them.

lkysow (approved):
🦊 ! See comment about test, also needs changelog entry.

PR #412, opened 2020-12-17
Give better error when policy was created manually

lkysow:
In cases where a user creates a policy out-of-band with the same name as
one consul-k8s tries to update, give a better error.

This is definitely an edge case but has occurred when a user created the
cross-namespace-acl-policy manually when trying to work around another
issue. This will at least give a better error message in this case.

How I've tested this PR:
- unit tests

- code

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/server-acl-init/command_test.go
+++ b/subcommand/server-acl-init/command_test.go
@@ -412,8 +412,7 @@ func TestRun_TokensReplicatedDC(t *testing.T) {
 	for _, c := range cases {
 		t.Run(c.TestName, func(t *testing.T) {
 			bootToken := "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
-			tokenFile, fileCleanup := writeTempFile(t, bootToken)
-			defer fileCleanup()
+			tokenFile := writeTempFile(t, bootToken)
 
 			k8s, consul, secondaryAddr, cleanup := mockReplicatedSetup(t, bootToken)
 			setUpK8sServiceAccount(t, k8s, ns)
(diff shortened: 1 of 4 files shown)


kschoche (approved):
Looks great!

subcommand/server-acl-init/create_or_update_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	// Start Consul.
+	bootToken := "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+	svr, err := testutil.NewTestServerConfigT(t, func(c *testutil.TestServerConfig) {
+		c.ACL.Enabled = true
+		c.ACL.Tokens.Master = bootToken
+	})
+	require.NoError(err)
+	svr.WaitForServiceIntentions(t)
```
ishustava, line 36: I'm wondering if we need this since this command or test aren't creating service intentions.

ishustava (approved):
Looks great! A couple of minor comments, but they are not blocking.

subcommand/server-acl-init/create_or_update_test.go
lkysow: Changed to `WaitForLeader()`

subcommand/server-acl-init/command_test.go
```diff
@@ -1097,6 +1097,69 @@ func TestRun_SyncPolicyUpdates(t *testing.T) {
...
+// that we try to update will work for testing.
+func TestRun_ErrorsOnDuplicateACLPolicy(t *testing.T) {
+	t.Parallel()
+	require := require.New(t)
+
+	// Create Consul with ACLs already bootstrapped so that we can
+	// then seed it with our manually created policy.
+	bootToken := "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+	tokenFile, fileCleanup := writeTempFile(t, bootToken)
+	defer fileCleanup()
```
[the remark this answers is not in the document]
lkysow: Done!

PR #413, opened 2020-12-18
Default namespaces for CRDs that support namespace when enabled

thisisnotashwin:
Changes proposed in this PR:
- Default the namespace fields of resources when namespaces are enabled so that the controller doesnt continuously attempt to reconcile them as the value presented by consul does not match the spec (because Consul defaults the namespace field where we don't).

How to test this PR:
- Code Review
- Deploy `consul-enterprise` with the consul-k8s image `ashwinvenkatesh/consul-k8s:default-ns`. Allow namespaces on the `consul` installation.

Verify that creating an `ingress-gateway`, `terminating-gateway` and `service-router` without namespace fields reconciles just once and does not lead to a reconcile loop.

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@
 BUG FIXES:
 * CRDs: Fix issue where a `ServiceIntentions` resource could be continually resynced with Consul
   because Consul's internal representation had a different order for an array than the Kubernetes resource. [[GH-416](https://github.com/hashicorp/consul-k8s/pull/416)] 
-* CRDs: default the `namespace` fields on resources where Consul performs namespace defaulting to prevent constant re-syncing.
+* CRDs: **(Consul Enterprise only)** default the `namespace` fields on resources where Consul performs namespace defaulting to prevent constant re-syncing.
   [[GH-413](https://github.com/hashicorp/consul-k8s/pull/413)]
+
 ## 0.22.0 (December 21, 2020)
 
 BUG FIXES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 BUG FIXES:
 * CRDs: Fix issue where a `ServiceIntentions` resource could be continually resynced with Consul
   because Consul's internal representation had a different order for an array than the Kubernetes resource. [[GH-416](https://github.com/hashicorp/consul-k8s/pull/416)] 
-
+* CRDs: default the `namespace` fields on resources where Consul performs namespace defaulting to prevent constant re-syncing.
+  [[GH-413](https://github.com/hashicorp/consul-k8s/pull/413)]
 ## 0.22.0 (December 21, 2020)
 
 BUG FIXES:
--- a/api/common/configentry.go
+++ b/api/common/configentry.go
@@ -57,6 +57,7 @@ type ConfigEntryResource interface {
 	DeepCopyObject() runtime.Object
 	// Validate returns an error if the resource is invalid.
 	Validate(namespacesEnabled bool) error
-	// Default sets the namespace field on the config entry spec to their default values if namespaces are enabled.
+	// DefaultNamespaceFields sets Consul namespace fields on the config entry
+	// spec to their default values if namespaces are enabled.
 	DefaultNamespaceFields(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string)
 }
--- a/api/common/configentry.go
+++ b/api/common/configentry.go
@@ -58,5 +58,5 @@ type ConfigEntryResource interface {
 	// Validate returns an error if the resource is invalid.
 	Validate(namespacesEnabled bool) error
 	// Default sets the namespace field on the config entry spec to their default values if namespaces are enabled.
-	Default(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string)
+	DefaultNamespaceFields(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string)
 }
--- a/api/common/configentry.go
+++ b/api/common/configentry.go
@@ -57,4 +57,6 @@ type ConfigEntryResource interface {
 	DeepCopyObject() runtime.Object
 	// Validate returns an error if the resource is invalid.
 	Validate(namespacesEnabled bool) error
+	// Default sets the namespace field on the config entry spec to their default values if namespaces are enabled.
+	Default(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string)
 }
--- a/api/common/configentry_webhook.go
+++ b/api/common/configentry_webhook.go
@@ -34,7 +34,7 @@ func ValidateConfigEntry(
 	consulDestinationNamespace string,
 	nsMirroringPrefix string) admission.Response {
 
-	defaultingPatches, err := defaultingPatches(cfgEntry, enableConsulNamespaces, nsMirroring, consulDestinationNamespace, nsMirroringPrefix)
+	defaultingPatches, err := DefaultingPatches(cfgEntry, enableConsulNamespaces, nsMirroring, consulDestinationNamespace, nsMirroringPrefix)
 	if err != nil {
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
@@ -67,9 +67,9 @@ func ValidateConfigEntry(
 	return admission.Patched(fmt.Sprintf("valid %s request", cfgEntry.KubeKind()), defaultingPatches...)
 }
 
-// defaultingPatches returns the patches needed to set fields to their
+// DefaultingPatches returns the patches needed to set fields to their
 // defaults.
-func defaultingPatches(cfgEntry ConfigEntryResource, enableConsulNamespaces bool, nsMirroring bool, consulDestinationNamespace string, nsMirroringPrefix string) ([]jsonpatch.Operation, error) {
+func DefaultingPatches(cfgEntry ConfigEntryResource, enableConsulNamespaces bool, nsMirroring bool, consulDestinationNamespace string, nsMirroringPrefix string) ([]jsonpatch.Operation, error) {
 	beforeDefaulting, err := json.Marshal(cfgEntry)
 	if err != nil {
 		return nil, fmt.Errorf("marshalling input: %s", err)
--- a/api/common/configentry_webhook.go
+++ b/api/common/configentry_webhook.go
@@ -69,13 +69,13 @@ func ValidateConfigEntry(
 
 // defaultingPatches returns the patches needed to set fields to their
 // defaults.
-func defaultingPatches(svcIntentions ConfigEntryResource, enableConsulNamespaces bool, nsMirroring bool, consulDestinationNamespace string, nsMirroringPrefix string) ([]jsonpatch.Operation, error) {
-	beforeDefaulting, err := json.Marshal(svcIntentions)
+func defaultingPatches(cfgEntry ConfigEntryResource, enableConsulNamespaces bool, nsMirroring bool, consulDestinationNamespace string, nsMirroringPrefix string) ([]jsonpatch.Operation, error) {
+	beforeDefaulting, err := json.Marshal(cfgEntry)
 	if err != nil {
 		return nil, fmt.Errorf("marshalling input: %s", err)
 	}
-	svcIntentions.Default(enableConsulNamespaces, consulDestinationNamespace, nsMirroring, nsMirroringPrefix)
-	afterDefaulting, err := json.Marshal(svcIntentions)
+	cfgEntry.DefaultNamespaceFields(enableConsulNamespaces, consulDestinationNamespace, nsMirroring, nsMirroringPrefix)
+	afterDefaulting, err := json.Marshal(cfgEntry)
 	if err != nil {
 		return nil, fmt.Errorf("marshalling after defaulting: %s", err)
 	}
--- a/api/common/configentry_webhook.go
+++ b/api/common/configentry_webhook.go
@@ -2,10 +2,12 @@ package common
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 
 	"github.com/go-logr/logr"
+	"gomodules.xyz/jsonpatch/v2"
 	"k8s.io/api/admission/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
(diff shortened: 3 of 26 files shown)


api/v1alpha1/serviceresolver_types.go
```diff
@@ -306,6 +307,28 @@ func (in *ServiceResolver) Validate(namespacesEnabled bool) error {
...
+func (in *ServiceResolver) Default(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string) {
+	// If namespaces are enabled we want to set the namespaces field to it's
+	// default. If namespaces are not enabled (i.e. OSS) we don't set the
+	// namespace fields because this would cause errors
+	// making API calls (because namespace fields can't be set in OSS).
+	if consulNamespacesEnabled {
+		namespace := namespaces.ConsulNamespace(in.Namespace, consulNamespacesEnabled, destinationNamespace, mirroring, prefix)
+		if in.Spec.Redirect != nil {
+			if in.Spec.Redirect.Namespace == "" {
+				in.Spec.Redirect.Namespace = namespace
```
lkysow, line 319: This doesn't get defaulted for some reason.

api/v1alpha1/serviceresolver_types.go
```diff
@@ -306,6 +307,28 @@ func (in *ServiceResolver) Validate(namespacesEnabled bool) error {
...
+			}
+		}
+		for k, v := range in.Spec.Failover {
+			if v.Namespace == "" {
+				failover := in.Spec.Failover[k]
+				failover.Namespace = namespace
```
lkysow, line 325: This doesn't default either.

api/v1alpha1/servicesplitter_types.go
```diff
@@ -170,6 +171,21 @@ func (in *ServiceSplitter) Validate(namespacesEnabled bool) error {
...
+func (in *ServiceSplitter) Default(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string) {
+	// If namespaces are enabled we want to set the namespace fields to it's
+	// default. If namespaces are not enabled (i.e. OSS) we don't set the
+	// namespace fields because this would cause errors
+	// making API calls (because namespace fields can't be set in OSS).
+	if consulNamespacesEnabled {
+		namespace := namespaces.ConsulNamespace(in.Namespace, consulNamespacesEnabled, destinationNamespace, mirroring, prefix)
+		for i, s := range in.Spec.Splits {
+			if s.Namespace == "" {
+				in.Spec.Splits[i].Namespace = namespace
```
lkysow, line 183: doesn't default

api/common/configentry.go
```diff
@@ -57,4 +57,6 @@ type ConfigEntryResource interface {
 	DeepCopyObject() runtime.Object
 	// Validate returns an error if the resource is invalid.
 	Validate(namespacesEnabled bool) error
+	// Default sets the namespace field on the config entry spec to their default values if namespaces are enabled.
+	Default(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string)
```
lkysow, line 61: Should we call this `DefaultNamespaceFields`?

lkysow (commented):
Looks good, I like the approach.

Can we remove the special case for serviceintentions now?

I checked out which places default, it's inconsistent. I've made comments where the defaulting doesn't happen.

api/common/configentry.go
ashwin-venkatesh: Sounds good

ashwin-venkatesh:
> Can we remove the special case for serviceintentions now?

We might have to keep that around because the validation behavior for Service Intentions are different from the other CRDs (we compare the name and namespace of the destination from the spec which is unique to the CRD)

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 BUG FIXES:
 * CRDs: Fix issue where a `ServiceIntentions` resource could be continually resynced with Consul
   because Consul's internal representation had a different order for an array than the Kubernetes resource. [[GH-416](https://github.com/hashicorp/consul-k8s/pull/416)] 
-
+* CRDs: default the `namespace` fields on resources where Consul performs namespace defaulting to prevent constant re-syncing.
+  [[GH-413](https://github.com/hashicorp/consul-k8s/pull/413)]
```
lkysow, line 7:
```suggestion
* CRDs: **(Consul Enterprise only)** default the `namespace` fields on resources where Consul performs namespace defaulting to prevent constant re-syncing.
  [[GH-413](https://github.com/hashicorp/consul-k8s/pull/413)]
  
```

api/common/configentry.go
```diff
@@ -57,4 +57,6 @@ type ConfigEntryResource interface {
 	DeepCopyObject() runtime.Object
 	// Validate returns an error if the resource is invalid.
 	Validate(namespacesEnabled bool) error
+	// Default sets the namespace field on the config entry spec to their default values if namespaces are enabled.
```
lkysow, line 60:
```suggestion
	// DefaultNamespaceFields sets Consul namespace fields on the config entry spec to their default values if namespaces are enabled.
```

api/v1alpha1/ingressgateway_types.go
```diff
@@ -225,6 +226,23 @@ func (in *IngressGateway) Validate(namespacesEnabled bool) error {
 	return nil
 }
 
+func (in *IngressGateway) DefaultNamespaceFields(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string) {
+	// If namespaces are enabled we want to set the namespace fields to it's
```
lkysow, line 230:
```suggestion
	// If namespaces are enabled we want to set the namespace fields to their
```

api/v1alpha1/ingressgateway_types.go
```diff
@@ -225,6 +226,23 @@ func (in *IngressGateway) Validate(namespacesEnabled bool) error {
...
+	// default. If namespaces are not enabled (i.e. OSS) we don't set the
```
lkysow, line 231:
```suggestion
	// defaults. If namespaces are not enabled (i.e. OSS) we don't set the
```

api/v1alpha1/ingressgateway_types_test.go
```diff
@@ -459,6 +460,91 @@ func TestIngressGateway_Validate(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		}
+
+		t.Run(name, func(t *testing.T) {
+			input.DefaultNamespaceFields(s.enabled, s.destinationNamespace, s.mirroring, s.prefix)
+			require.True(t, cmp.Equal(input, output))
+		})
```
lkysow, line 544: nit: I think the rest of our tests execute the `t.Run` on the next line after the `for range case` loop.

api/v1alpha1/proxydefaults_types.go
```diff
@@ -176,6 +176,11 @@ func (in *ProxyDefaults) Validate(namespacesEnabled bool) error {
 	return nil
 }
 
+// Default has no behaviour here as proxy-defaults have no namespace specific fields.
```
lkysow, line 179:
```suggestion
// DefaultNamespaceFields has no behaviour here as proxy-defaults have no namespace specific fields.
```

Also the other locations.

api/v1alpha1/servicerouter_types.go
```diff
@@ -255,6 +256,23 @@ func (in *ServiceRouter) Validate(namespacesEnabled bool) error {
 	return nil
 }
 
+func (in *ServiceRouter) DefaultNamespaceFields(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string) {
+	// If namespaces are enabled we want to set the namespace fields to it's
+	// default. If namespaces are not enabled (i.e. OSS) we don't set the
+	// namespace fields because this would cause errors
+	// making API calls (because namespace fields can't be set in OSS).
```
lkysow, line 263:
```suggestion
	// If namespaces are enabled we want to set the namespace fields to their
	// defaults. If namespaces are not enabled (i.e. OSS) we don't set the
	// namespace fields because this would cause errors
	// making API calls (because namespace fields can't be set in OSS).
```

api/v1alpha1/serviceresolver_types.go
```diff
@@ -306,6 +306,9 @@ func (in *ServiceResolver) Validate(namespacesEnabled bool) error {
 	return nil
 }
 
+func (in *ServiceResolver) DefaultNamespaceFields(_ bool, _ string, _ bool, _ string) {
```
lkysow, line 309: Missing the same comment about this not having defaults.

api/common/configentry_webhook.go
```diff
@@ -56,5 +64,25 @@ func ValidateConfigEntry(
 	if err := cfgEntry.Validate(enableConsulNamespaces); err != nil {
 		return admission.Errored(http.StatusBadRequest, err)
 	}
-	return admission.Allowed(fmt.Sprintf("valid %s request", cfgEntry.KubeKind()))
+	return admission.Patched(fmt.Sprintf("valid %s request", cfgEntry.KubeKind()), defaultingPatches...)
+}
+
+// defaultingPatches returns the patches needed to set fields to their
+// defaults.
+func defaultingPatches(cfgEntry ConfigEntryResource, enableConsulNamespaces bool, nsMirroring bool, consulDestinationNamespace string, nsMirroringPrefix string) ([]jsonpatch.Operation, error) {
```
lkysow, line 72: If we export this I think we can use it in serviceintentions and get rid of the duplicate code.

```go
	defaultingPatches, err := common.DefaultingPatches(&svcIntentions, v.EnableConsulNamespaces, v.EnableNSMirroring, v.ConsulDestinationNamespace, v.NSMirroringPrefix)
	if err != nil {
		return admission.Errored(http.StatusInternalServerError, err)
	}

```

lkysow (commented):
This is looking good! I haven't tested it out yet. I think we might be able to reduce the code duplication between `serviceintentions_webhook`.

lkysow (approved):
I tested this all out and it works great.

CHANGELOG.md
thisisnotashwin: oops..fixing it..somehow i noticed this and i blanked

api/v1alpha1/ingressgateway_types_test.go
```diff
@@ -459,6 +460,90 @@ func TestIngressGateway_Validate(t *testing.T) {
 	}
 }
 
+// Test defaulting behavior when namespaces are enabled as well as disabled.
+func TestIngressGateway_Default(t *testing.T) {
```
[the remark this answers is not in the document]
thisisnotashwin: good catch

api/common/configentry_webhook_test.go
```diff
@@ -112,7 +114,9 @@ func TestValidateConfigEntry(t *testing.T) {
 				lister,
 				c.newResource,
 				c.enableNamespaces,
-				c.nsMirroring)
+				c.nsMirroring,
+				c.consulDestinationNS,
+				c.nsMirroringPrefix)
 			require.Equal(t, c.expAllow, response.Allowed)
```
ishustava, line 120: Do we need to add an assertion here that the defaulting has worked (maybe check that the `DefaultNamespaceFields`  method was called on the mock config entry)?

api/v1alpha1/ingressgateway_types.go
```diff
@@ -225,6 +226,24 @@ func (in *IngressGateway) Validate(namespacesEnabled bool) error {
...
 }
 
+// DefaultNamespaceFields sets the namespace field on spec.listeners[].services to their default values if namespaces are enabled.
+func (in *IngressGateway) DefaultNamespaceFields(consulNamespacesEnabled bool, destinationNamespace string, mirroring bool, prefix string) {
+	// If namespaces are enabled we want to set the namespace fields to their
+	// defaults. If namespaces are not enabled (i.e. OSS) we don't set the
+	// namespace fields because this would cause errors
+	// making API calls (because namespace fields can't be set in OSS).
+	if consulNamespacesEnabled {
+		namespace := namespaces.ConsulNamespace(in.Namespace, consulNamespacesEnabled, destinationNamespace, mirroring, prefix)
```
ishustava, line 236:
```suggestion
                // Default to the current namespace (i.e. the namespace of this config entry).
		namespace := namespaces.ConsulNamespace(in.Namespace, consulNamespacesEnabled, destinationNamespace, mirroring, prefix)
```

api/v1alpha1/serviceresolver_types.go
```diff
@@ -306,6 +306,11 @@ func (in *ServiceResolver) Validate(namespacesEnabled bool) error {
 	return nil
 }
 
+// DefaultNamespaceFields has no behaviour here as service-resolver have namespace fields
+// that do not default.
+func (in *ServiceResolver) DefaultNamespaceFields(_ bool, _ string, _ bool, _ string) {
+}
```
ishustava, line 312: I'm curious why we didn't need to say `return` here as we did in others.

api/v1alpha1/terminatinggateway_types_test.go
```diff
@@ -279,6 +280,80 @@ func TestTerminatingGateway_Validate(t *testing.T) {
...
+			prefix:               "",
+			expectedDestination:  "bar",
+		},
+		"mirroringWithPrefix": {
+			enabled:              true,
+			destinationNamespace: "",
+			mirroring:            true,
+			prefix:               "ns-",
+			expectedDestination:  "ns-bar",
+		},
```
ishustava, line 319: I think it'd be good to add a test case to all of these that this method doesn't overwrite if the namespace field is already set

ishustava (commented):
Looking pretty good, Ashwin! I only had one suggestion that I think is holding the approval (to add a test case checking that the `DefaultNamespaceFields` implementations won't overwrite a field that's already set); the rest of my comments are pretty minor. Great work!

api/v1alpha1/serviceresolver_types.go
thisisnotashwin: it was unintentional. Fixed.

api/common/configentry_webhook_test.go
thisisnotashwin: Added a test for the same. It is fairly primitive but let me know if that works of if you'd like something a little more fleshed out.

ishustava: that works, thanks!

ishustava (approved):
Looks great! Thanks for addressing the comments.

PR #416, opened 2021-01-06
Ensure ServiceIntentions aren't resynced

lkysow:
Consul will internally sort the array of ServiceIntention sources by
precedence. For example, if it receives:

```
sources:
  - name: "*"
    action: allow
  - name: bar
    action: allow
```

It will return it in precedence sorted order:

```
sources:
  - name: bar
    action: allow
  - name: "*"
    action: allow
```

This sorting breaks our MatchesConsul function because we think the
resource has changed and so we will update it continually since it will
never match.

This change uses gocmp's SortSlices to re-sort the sources array so that
both the Kube resource and the Consul resource can be compared despite
Consul's sorting. We use a JSON encoded representation of the source
element with the fields that Consul set automatically zero'd out.

How I've tested this PR:
- unit tests
- tested with resource:
    ```yaml
    kind: ServiceIntentions
    metadata:
      name: foo
    spec:
      destination:
        name: foo
      sources:
        - name: "*"
          action: allow
        - name: bar
          action: allow
    ```

  Before:
   ```
    2021-01-06T22:04:06.335Z  INFO  webhooks.serviceintentions  validate create {"name": "foo"}
    2021-01-06T22:04:06.347Z  INFO  webhooks.serviceintentions  validate update {"name": "foo"}
    2021-01-06T22:04:06.365Z  INFO  controller.serviceintentions  config entry not found in consul  {"request": "default/foo"}
    2021-01-06T22:04:06.377Z  INFO  controller.serviceintentions  config entry created  {"request": "default/foo", "request-time": "11.594086ms"}
    2021-01-06T22:04:06.390Z  INFO  controller.serviceintentions  config entry does not match consul  {"request": "default/foo", "modify-index": 511}
    2021-01-06T22:04:06.394Z  INFO  controller.serviceintentions  config entry updated  {"request": "default/foo", "request-time": "3.454279ms"}
   ```

   After:
    ```
    2021-01-06T22:07:19.745Z  INFO  webhooks.serviceintentions  validate create {"name": "foo"}
    2021-01-06T22:07:19.751Z  INFO  webhooks.serviceintentions  validate update {"name": "foo"}
    2021-01-06T22:07:19.759Z  INFO  controller.serviceintentions  config entry not found in consul  {"request": "default/foo"}
    2021-01-06T22:07:19.762Z  INFO  controller.serviceintentions  config entry created  {"request": "default/foo", "request-time": "3.048659ms"}
    ```

How I expect reviewers to test this PR: Code

Checklist:
- [x] Tests added

### DIFF
--- a/api/v1alpha1/serviceintentions_types.go
+++ b/api/v1alpha1/serviceintentions_types.go
@@ -3,6 +3,7 @@ package v1alpha1
 import (
 	"encoding/json"
 	"net/http"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -222,6 +223,7 @@ func (in *ServiceIntentions) MatchesConsul(candidate api.ConfigEntry) bool {
 	if !ok {
 		return false
 	}
+
 	// No datacenter is passed to ToConsul as we ignore the Meta field when checking for equality.
 	return cmp.Equal(
 		in.ToConsul(""),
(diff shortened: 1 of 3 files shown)


api/v1alpha1/serviceintentions_types.go
```diff
@@ -222,6 +223,28 @@ func (in *ServiceIntentions) MatchesConsul(candidate api.ConfigEntry) bool {
 	if !ok {
 		return false
 	}
+
+	// sourceSortKey returns a string that can be used to sort intention
+	// sources.
+	sourceSortKey := func(ixn *capi.SourceIntention) string {
```
ashwin-venkatesh, line 229: can we pull this out into a private func for readability? not really a blocker.

ashwin-venkatesh (approved):
This looks great!! Have a minor comment but it isn't blocking.

ndhanushkodi (approved):
Looks great 🎉 !!

PR #418, opened 2021-01-12
Remove -default-protocol and annotation support

lkysow:
Changes proposed in this PR:
- Remove support for `-default-protocol` and `-enable-central-config` flags passed to `consul-k8s inject-connect` command. These flags would cause Consul to write a `service-defaults` config entry that would set the protocol of the service.
- Remove support for the `consul.hashicorp.com/connect-service-protocol` annotation. Connect pods that have this annotation will not be injected and the injector will return an error.

How I've tested this PR:
* Using `ghcr.io/lkysow/consul-k8s-dev:jan11` I created a service that sets the annotation and confirmed the replicaset won't start with error: 
  ```
  Error creating: admission webhook "consul-connect-injector.consul.hashicorp.com" denied the request: Error validating pod: 
  the consul.hashicorp.com/connect-service-protocol annotation is no longer supported. Instead, create a ServiceDefaults 
  resource (see www.consul.io/docs/k8s/crds/upgrade-to-crds)
  ```

* code

Checklist:
- [x] Tests added

Companion to: https://github.com/hashicorp/consul-helm/pull/763


thisisnotashwin (approved):
Looks great!! Good work @lkysow 🦊

ishustava (approved):
Looks perfect! Looking at all the code removals gives you such a satisfying feeling 😃

PR #420, opened 2021-01-12
Give better error if config entry already created

lkysow:
Previously if the config entry already existed in Consul we'd give an
error like:

    config entry managed in different datacenter: "other-dc"

This worked well if the config entry had been created by a controller
running in another Kube cluster, but if the user had created the config
entry directly in Consul then it wouldn't have the datacenter annotation
and so the error would actually read:

    config entry managed in different datacenter: ""

Which was confusing.

This change changes the error in that case to read:

    config entry already exists in Consul

How I've tested this PR:

Checklist:
- [x] Tests added

### DIFF
--- a/controller/configentry_controller_test.go
+++ b/controller/configentry_controller_test.go
@@ -1843,16 +1843,26 @@ func TestConfigEntryControllers_doesNotCreateUnownedConfigEntry(t *testing.T) {
 	kubeNS := "default"
 
 	cases := []struct {
-		kubeKind            string
-		consulKind          string
-		consulPrereqs       []capi.ConfigEntry
-		configEntryResource common.ConfigEntryResource
-		reconciler          func(client.Client, *capi.Client, logr.Logger) testReconciler
+		datacenterAnnotation string
+		expErr               string
 	}{
 		{
-			kubeKind:   "ServiceDefaults",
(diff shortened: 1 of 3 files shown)


controller/configentry_controller_test.go
```diff
@@ -1843,292 +1843,36 @@ func TestConfigEntryControllers_doesNotCreateUnownedConfigEntry(t *testing.T) {
 	kubeNS := "default"
 
 	cases := []struct {
-		kubeKind            string
```
lkysow, line 1846: I removed the cases for each config entry and instead added cases based on the value of the `source-datacenter` metadata since now we expect a different error message if it's empty vs set to something else vs set to our datacenter.

I think it's okay to remove the per-config-entry cases because
1. they were testing the same underlying code
2. I would need to add config entry x cases permutations to test this new functionality which I think is going to make the tests too complicated
3. each config entry has unit tests to ensure it implements the ConfigEntry interface properly (which is all we require to be working here)

thisisnotashwin (approved):
🎉

controller/configentry_controller_test.go
lkysow: https://github.com/hashicorp/consul-k8s/pull/424

kschoche (approved):
🔥

PR #421, opened 2021-01-21
Error when mesh gateway mode not set

lkysow:
To prevent users from being confused as to why their traffic isn't being
routed, give an error on connect injection if they've configured an
upstream to use a datacenter but they don't have a proxy-defaults config
with mesh gateway mode set to local or remote.

NOTE: No ACL changes need to be made because `proxy-defaults` does not require an ACL for reads (https://www.consul.io/api-docs/config#get-configuration)

How I tested:
* Install with:
  ```yaml
  global:
    name: consul
    imageK8S: ghcr.io/lkysow/consul-k8s-dev:jan20
    acls:
      manageSystemACLs: true
  connectInject:
    enabled: true
  controller:
    enabled: true
  server:
    replicas: 1
  ```
* Apply
  ```yaml
  apiVersion: v1
  kind: ServiceAccount
  metadata:
    name: static-client
  ---
  apiVersion: v1
  kind: Pod
  metadata:
    name: luke
    annotations:
      "consul.hashicorp.com/connect-service-upstreams": "static-server:1234:dc2"
  spec:
    containers:
      - name: static-client
        image: tutum/curl:latest
        command: [ "/bin/sh", "-c", "--" ]
        args: [ "while true; do sleep 30; done;" ]
        ports:
          - containerPort: 8080
            name: http
    serviceAccountName: static-client
  ```

* Errors with
  ```
  Error from server: error when creating "scratch/static-client.yaml": admission webhook "consul-connect-injector.consul.hashicorp.com" denied the request: Error configuring injection init container: upstream "static-server:1234:dc2" is invalid: there is no ProxyDefaults config to set mesh gateway mode
  ```
* Apply
  ```yaml
  kind: ProxyDefaults
  metadata:
    name: global
  spec:
    meshGateway:
      mode: local
  ```
* Re-apply static-client, it will schedule

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/container_init_test.go
+++ b/connect-inject/container_init_test.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"testing"
 
+	capi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/sdk/testutil"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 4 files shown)


subcommand/server-acl-init/command_test.go
```diff
@@ -1906,6 +1906,7 @@ func completeBootstrappedSetup(t *testing.T, masterToken string) (*fake.Clientse
 		c.ACL.Tokens.Master = masterToken
 	})
 	require.NoError(t, err)
+	svr.WaitForActiveCARoot(t)
```
lkysow, line 1909: Test flake: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/1542/workflows/3b1ad5c1-6bce-43a6-8b67-a9742dbf8557/jobs/7574

connect-inject/container_init_test.go
```diff
@@ -589,7 +591,29 @@ services {
 		t.Run(tt.Name, func(t *testing.T) {
 			require := require.New(t)
 
-			var h Handler
+			// Create a Consul server/client and proxy-defaults config because
```
lkysow, line 594: These tests didn't change but now require consulClient to be instantiated because they test datacenter upstreams.

thisisnotashwin (approved):
EXCELLENT!

kschoche (approved):
This is a super nice fix to have, great work!

PR #424, opened 2021-01-25
Simplify controller test cases

lkysow:
Refactor the CRD controller test cases to make them smaller and easier to maintain.

Previously each test had a case for each config entry type. This PR leaves the cases for the CRUD tests but for the edge case tests it removes the cases and just uses `ServiceDefaults` because it's a simple config entry with no prereqs. This change was made to reduce the complexity of the edge case tests and because all the cases were testing the exact same code paths of the generic `configentry_controller`.

Adds one permutation to `TestConfigEntryControllers_doesNotDeleteUnownedConfig` to test what happens when the metadata key is empty.

### DIFF
--- a/controller/configentry_controller_test.go
+++ b/controller/configentry_controller_test.go
@@ -1253,239 +1253,58 @@ func TestConfigEntryControllers_errorUpdatesSyncStatus(t *testing.T) {
 	t.Parallel()
 	kubeNS := "default"
 
-	cases := []struct {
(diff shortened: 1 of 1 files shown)


thisisnotashwin (commented):
❤️  everything about this PR

thisisnotashwin (approved):
NICE! 🦊

kschoche (approved):
This is awesome, love the cleanup!

Issue #373, opened 2020-10-30
[docs] Document how to connect to specific pod using connect


LaumiH:
**Edit from @lkysow:** See https://github.com/hashicorp/consul-k8s/issues/373#issuecomment-769351345 for how to accomplish this. This issue now tracks documenting the process properly.

---

If I understand correctly, Consul for now can only resolve a service of type StatefulSet via `<service>.service.<datacenter>.consul`.

What I am missing is the possibility to address the individual pods of the set via their Headless Service, as it is possible with K8s DNS <pod-0>.<service>.<namespace>.svc.cluster.local.

My use case is a cross-cluster MongoDB deployment, where the individual members can be addressed via their Consul resolvable names across cluster boundaries.

Of course this is useful with any StatefulSet and not limited to databases.

Explaining further what I want to achieve:
- setup Consul Federation with Mesh Gateways
- setup StatefulSet (e.g. MongoDB) in both clusters, annotate so Consul manages service
- annotate accordingly, so that upstream dependencies of form mongo-0.mongo.service.dc1.consul can be found connecting to various ports in each pod, e.g. `'consul.hashicorp.com/connect-service-upstreams': 'mongo-0:mongo:1234:dc1,mongo-1:mongo:5678:dc2'`
- use Consul to connect the mongo pods across clusters

Additionally, a sidecar would be helpful to configure all this automatically and set the replicaset inside mongo, but this is a different topic :)

I am also referring to [this forum post](https://discuss.hashicorp.com/t/kubernetes-service-mesh-access-to-service-instances/15218/6), where a user asked for the same feature.

lkysow:
Hi Laura,
This might be possible if we can use the filter expression of the service resolver (https://www.consul.io/docs/agent/config-entries/service-resolver#filter) to select a service by its ID. The ID of the service in consul is `<service-name>-<pod id>`, e.g. `consul-consul-server-0`.

Perhaps you can test this out?

LaumiH:
Thanks for the quick response! I will look into it now.

LaumiH:
Ok, so I'm stuck and need help. I am new to Consul and cannot transfer what I am reading about the service resolver into k8s usage.

My understanding is as follows: 
I have created a Headless Service for my mongo StatefulSet, which gets [synced](https://www.consul.io/docs/k8s/service-sync#clusterip) into Consul by default or if I annotate it accordingly. 
[...]

PR #427, opened 2021-01-28
Use an underscore instead of a dash for the pod name meta key

ishustava:
Consul [filtering]((https://www.consul.io/api/features/filtering#selectors)) API only supports ASCII character, numbers, and underscores
as selectors.
So if someone tries to use that meta key for e.g. a service-resolver,
it won't work.

Changes proposed in this PR:
- Switch the meta key for the pod name from `pod-name` to `pod_name`

How I've tested this PR:
- Steps outlined [here](https://github.com/hashicorp/consul-k8s/issues/373#issuecomment-769351345).

- code review

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ BREAKING CHANGES
   In practice, this would have already been causing issues since without that
   config setting, traffic wouldn't have been routed through mesh gateways and
   so would not be actually making it to the other service.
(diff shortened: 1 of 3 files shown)


thisisnotashwin (approved):
_bellissima_

kschoche (approved):
🌲

PR #428, opened 2021-01-30
Rename lifecycle-sidecar command to consul-sidecar

ndhanushkodi:
Changes proposed in this PR:
- Originally, the lifecycle-sidecar's functionality was to register the
Connect service with Consul. This sidecar will in the near future
support merged metrics functionality, so this renaming supports that.

**NOTE:** will wait to merge until the corresponding consul-helm change to change any calls to flags with "lifecycle-sidecar" in the name.

How I've tested this PR:
- Consul-k8s tests

- Code review

Checklist:
- [x] Tests added (no new tests, rename)


commands.go
```diff
@@ -35,8 +35,8 @@ func init() {
 			return &cmdInjectConnect.Command{UI: ui}, nil
 		},
 
-		"lifecycle-sidecar": func() (cli.Command, error) {
-			return &cmdLifecycleSidecar.Command{UI: ui}, nil
+		"consul-sidecar": func() (cli.Command, error) {
```
lkysow, line 38: Can we make `lifecycle-sidecar` an alias for `consul-sidecar`? If we're not adding any mandatory flags then it would be backwards compat I think?

ndhanushkodi: Oh interesting, yeah! Would that mean just having lifecycle-sidecar return the consul-sidecar command, i.e:
```
"lifecycle-sidecar": func() (cli.Command, error) {
			return &cmdConsulSidecar.Command{UI: ui}, nil
```

Hmm, this command is only called by the pod-mutating-webhook, right? So wouldn't nobody but our own code be calling consul-sidecar? If that's the case, I feel like we wouldn't need to create the alias since within the webhook handler we just change which command gets run for the sidecar. Its not exposed in any way through the helm chart.

There are also some inject-connect flags, like lifecycle-sidecar-cpu-request that have been renamed, so I was thinking the helm values would accept `global.lifecycleSidecarContainer.resources.requests.cpu` and `global.consulSidecarContainer.resources.requests.cpu` but both would point to the consul-sidecar-cpu-request flag in consul-k8s.

lkysow:
Oh yeah right because after upgrade new pods will just use the new command
name. Nevermind!

connect-inject/consul_sidecar_test.go
```diff
@@ -27,14 +27,14 @@ var (
...
-func TestLifecycleSidecar_Default(t *testing.T) {
+// Test that the consul sidecar is as expected.
+func TestConsulSidecar_Default(t *testing.T) {
 	handler := Handler{
-		Log:                       hclog.Default().Named("handler"),
-		ImageConsulK8S:            "hashicorp/consul-k8s:9.9.9",
-		LifecycleSidecarResources: lifecycleResources,
+		Log:                    hclog.Default().Named("handler"),
+		ImageConsulK8S:         "hashicorp/consul-k8s:9.9.9",
+		ConsulSidecarResources: lifecycleResources,
```
lkysow, line 35: need to rename `lifecycleResources` too

connect-inject/consul_sidecar_test.go
```diff
@@ -174,7 +174,7 @@ func TestLifecycleSidecar_TLS(t *testing.T) {
 			},
 		},
 		Command: []string{
-			"consul-k8s", "lifecycle-sidecar",
+			"consul-k8s", "consul-sidecar",
 			"-service-config", "/consul/connect-inject/service.hcl",
 			"-consul-binary", "/consul/connect-inject/consul",
 		},
```
lkysow, line 180: need to rename the `lifecycleResources` below

connect-inject/health_check_resource_test.go
```diff
@@ -566,7 +566,7 @@ func TestReconcile_IgnoreStatuses(t *testing.T) {
 					Ready: false,
 				},
 				{
-					Name: "consul-connect-lifecycle-sidecar",
+					Name: "consul-connect-consul-sidecar",
```
lkysow, line 569: Should we call this `consul-sidecar` or `consul-connect-sidecar`?

connect-inject/health_check_resource_test.go
```diff
@@ -792,7 +792,7 @@ var unreadyAppContainers = []corev1.ContainerStatus{
 		Ready: false,
 	},
 	{
-		Name: "consul-connect-lifecycle-sidecar",
+		Name: "consul-connect-lifconsulecycle-sidecar",
```
lkysow, line 795: typo

subcommand/consul-sidecar/command.go
```diff
@@ -187,9 +187,9 @@ func (c *Command) Help() string {
 	return c.help
 }
 
-const synopsis = "Connect lifecycle sidecar."
+const synopsis = "Connect consul sidecar."
```
lkysow, line 190:
```suggestion
const synopsis = "Consul connect sidecar."
```

subcommand/consul-sidecar/command.go
```diff
@@ -187,9 +187,9 @@ func (c *Command) Help() string {
...
 const help = `
-Usage: consul-k8s lifecycle-sidecar [options]
+Usage: consul-k8s consul-sidecar [options]
```
lkysow, line 192: There's `lifecycleResources` on line `258` and a couple other instances.

lkysow (commented):
Looking good. A couple places for lifecycle were missed that I could see. Do a quick grep in case I missed some others.

connect-inject/health_check_resource_test.go
ndhanushkodi: I think I still like `consul-connect-consul-sidecar` because the component and command is consul-sidecar and the consul-connect is a prefix for other containers like the envoy sidecar name to specify its only added with Connect. `consul-sidecar` alone doesn't let you know this is a Connect-related feature, and loses consistency with the envoy container name and `consul-connect-sidecar` also loses consistency with the envoy container name. Wdyt?

connect-inject/consul_sidecar_test.go
ndhanushkodi: Nice thanks for catching this for some reason I overlooked that as a k8s specific thing but that's just the word lifecycle on its own. Renamed!

subcommand/consul-sidecar/command.go
ndhanushkodi: Maybe it can be "Consul sidecar for Connect"?

connect-inject/health_check_resource_test.go
ndhanushkodi: thanks!

connect-inject/consul_sidecar_test.go
```diff
@@ -27,14 +27,14 @@ var (
 // would require a lot of boilerplate to get at the underlying patches that would
 // complicate understanding the tests (which are simple).
 
-// Test that the lifecycle sidecar is as expected.
-func TestLifecycleSidecar_Default(t *testing.T) {
+// Test that the consul sidecar is as expected.
```
thisisnotashwin, line 30:
```suggestion
// Test that the Consul sidecar is as expected.
```

connect-inject/handler.go
```diff
@@ -187,9 +187,9 @@ type Handler struct {
 	// will be populated by the defaults provided in the initial flags.
 	InitContainerResources corev1.ResourceRequirements
 
-	// Resource settings for lifecycle sidecar. All of these fields
+	// Resource settings for consul sidecar. All of these fields
```
thisisnotashwin, line 190:
```suggestion
	// Resource settings for Consul sidecar. All of these fields
```

connect-inject/handler.go
```diff
@@ -342,7 +342,7 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
 		[]corev1.Container{container},
 		"/spec/initContainers")...)
 
-	// Add the Envoy and lifecycle sidecars.
+	// Add the Envoy and consul sidecars.
```
thisisnotashwin, line 345:
```suggestion
	// Add the Envoy and Consul sidecars.
```

connect-inject/health_check_resource_test.go
thisisnotashwin: I think I'm with Luke on this one. I made a similar comment in another place in the code before I noticed it was being discussed here. Personally, i wouldn't mind renaming `consul-connect-envoy-sider` to `envoy-sidecar` and the calling the sidecar in question the `consul-sidecar`. 
I like the argument for consistency but in this case, IMO, it feels more confusing to parse.

subcommand/inject-connect/command.go
```diff
@@ -565,55 +565,55 @@ func (c *Command) parseAndValidateResourceFlags() (corev1.ResourceRequirements,
...
 	}
 
-	// Lifecycle sidecar
-	var lifecycleSidecarCPULimit, lifecycleSidecarCPURequest, lifecycleSidecarMemoryLimit, lifecycleSidecarMemoryRequest resource.Quantity
+	// Consul sidecar
+	var consulSidecarCPULimit, consulSidecarCPURequest, consulSidecarMemoryLimit, consulSidecarMemoryRequest resource.Quantity
 
-	// Parse and validate the lifecycle sidecar resources
-	lifecycleSidecarCPURequest, err = resource.ParseQuantity(c.flagLifecycleSidecarCPURequest)
+	// Parse and validate the consul sidecar resources
```
thisisnotashwin, line 571:
```suggestion
	// Parse and validate the Consul sidecar resources
```

thisisnotashwin (commented):
Excellent work here @ndhanushkodi !! Have a few `consul` -> `Consul` suggestions. Also, threw in my opinion on the name of the sidecar.

connect-inject/health_check_resource_test.go
ndhanushkodi: I've renamed the containers to `consul-sidecar` and `envoy sidecar` :)

thisisnotashwin (approved):
Excellent work!!

lkysow (approved):
🌀

PR #431, opened 2021-02-02
Include object with delete events

lkysow:
Refactor our controller implementation to include the resource as part
of the deletion event. During a delete event, we get the full object
however previously we would discard this object and record only the key
of the object that was deleted. Then later when we would dequeue this
event, the object could not be retrieved from our cache because it was
deleted from Kubernetes and our cache is usually up to date with the
state of Kubernetes.

This will be useful for cleaning up force deleted pods because we can
use the pod object from the event to determine which Consul service
instance that pod got registered as (and potentially delete it if it's
still registered).

Checklist:
- [x] Tests added

### DIFF
--- a/helper/controller/controller.go
+++ b/helper/controller/controller.go
@@ -25,6 +25,20 @@ type Controller struct {
 	informer cache.SharedIndexInformer
 }
 
+// Event is something that occurred to the resources we're watching.
+type Event struct {
+	// Key is in the form of <namespace>/<name>, e.g. default/pod-abc123,
+	// and corresponds to the resource modified.
+	Key string
+	// Obj holds the resource that was modified at the time of the event
+	// occurring. If possible, the resource should be retrieved from the informer
+	// cache, instead of using this field because the cache will be more up to
+	// date at the time the event is processed.
+	// In some cases, such as a delete event, the resource will no longer exist
+	// in the cache and then it is useful to have the resource here.
+	Obj interface{}
+}
+
 // Run starts the Controller and blocks until stopCh is closed.
 //
 // Important: Callers must ensure that Run is only called once at a time.
@@ -52,21 +66,21 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 			key, err := cache.MetaNamespaceKeyFunc(obj)
 			c.Log.Debug("queue", "op", "add", "key", key)
 			if err == nil {
-				queue.Add(key)
+				queue.Add(Event{Key: key, Obj: obj})
 			}
 		},
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			key, err := cache.MetaNamespaceKeyFunc(newObj)
 			c.Log.Debug("queue", "op", "update", "key", key)
 			if err == nil {
-				queue.Add(key)
+				queue.Add(Event{Key: key, Obj: newObj})
 			}
 		},
 		DeleteFunc: func(obj interface{}) {
 			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 			c.Log.Debug("queue", "op", "delete", "key", key)
 			if err == nil {
-				queue.Add(key)
+				queue.Add(Event{Key: key, Obj: obj})
 			}
 		},
 	})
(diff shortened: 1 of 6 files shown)


helper/controller/controller.go
```diff
@@ -25,6 +25,20 @@ type Controller struct {
 	informer cache.SharedIndexInformer
 }
 
+// Event is something that occurred to the resources we're watching.
```
lkysow, line 28: Instead of passing string keys into the queue, pass this struct now.

helper/controller/controller.go
```diff
@@ -149,44 +163,47 @@ func (c *Controller) processSingle(
...
 	// If we got the item successfully, call the proper method
 	if err == nil {
-		c.Log.Debug("processing object", "key", keyRaw, "exists", exists)
+		c.Log.Debug("processing object", "key", key, "exists", exists)
 		c.Log.Trace("processing object", "object", item)
 		if !exists {
-			err = c.Resource.Delete(keyRaw)
+			// In the case of deletes, the item is no longer in the cache so
+			// we use the copy we got at the time of the event (event.Obj).
+			err = c.Resource.Delete(key, event.Obj)
```
lkysow, line 190: This is really the major change here. Previously we would just pass `key` into `Delete()`. Now we also pass `event.Obj` which contains the resource that was included with the delete event from kube. This is our only copy of the resource since `informer` has already had it cleared from its cache.

helper/controller/controller_test.go
```diff
@@ -26,7 +26,7 @@ func TestController_initialData(t *testing.T) {
 	require := require.New(t)
```
lkysow, line 26: These tests worked by exposing a `data` map that contained all the updates via the callbacks. To test the new `Delete()` functionality, I have exposed a new map called `deleted` that lists the keys that were deleted and their values so we can test that `Delete(key, obj)` was called as expected.

helper/controller/controller.go
```diff
@@ -52,21 +66,21 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
...
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			key, err := cache.MetaNamespaceKeyFunc(newObj)
 			c.Log.Debug("queue", "op", "update", "key", key)
 			if err == nil {
-				queue.Add(key)
+				queue.Add(Event{Key: key, Obj: newObj})
 			}
 		},
 		DeleteFunc: func(obj interface{}) {
 			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
```
kschoche, line 80: I'm curious if this should be `MetaNamespaceKeyFunc` instead of `DeletionHandlingMetaNamespaceKeyFunc` because this was intended to skip when the object has been deleted in certain circumstances?

// DeletionHandlingMetaNamespaceKeyFunc checks for DeletedFinalStateUnknown objects before calling MetaNamespaceKeyFunc.

kschoche (approved):
This looks great @lkysow !
I only had one comment which is just a clarifying question.
Ran the tests and they look solid.

helper/controller/controller.go
lkysow: I read a bit into this via https://lairdnelson.wordpress.com/2018/01/07/understanding-kubernetes-tools-cache-package-part-0/ and I still don't fully understand it (so could definitely be wrong). What I primarily don't understand is how if the type of obj is `DeletedFinalStateUnknown` that it would have a **different** key than if it wasn't.

My guess is that this is an artifact of the way the queue works where it does its own "compression" and modifies the events in the queue. Maybe in this case it modifies the event such that the key can't be read from the normal struct fields.

Another thing I'm confused by is that in my testing the watch events for DELETE always have the full pod object.

tl;dr: definitely open to understanding this further but because I don't right now I think we should leave it as `DeletionHandlingMetaNamespaceKeyFunc` because I'm wont to change something I don't understand and because this seems so work with all my testing.

kschoche: Sounds good. I was equally confused. But if it works for our needs and we can prove that with automated testing we're good!

lkysow: To follow-up on this, I did figure it out. The controller reads from a queue called a DeltaFIFO. That queue tries to give a consistent sequence of updates to match what's happening in Kubernetes, e.g. [(delete, foo), (update, bar), etc.]

It watches the pod resource in Kubernetes but it will also do a periodic resync where it gets the full list of pods. In that situation, imagine it had the following list of events: `[(add, foo)]`, and the it calls Kube for the list of pods and kube says there are no pods.

In this case, it needs to "make up" a `(delete, foo)` event so its queue is consistent. Usually, when it has a `(delete, foo)` event that it got from Kube via its watch, `foo` will be the most up-to-date version of the foo pod. However in this case where it's making up the delete event, it can only use the version of `foo` it received in its `(add, foo)` event.

In order to make this clear to callers that the `(delete, foo)` event they're getting contains a possibly out-of-date `foo` object, the `DeltaFIFO` queue wraps `foo` in another struct: `DeletedFinalStateUnknown{key: key, obj: foo}`.  This is why we have to use the `DeletionHandlingMetaNamespaceKeyFunc` function here, because its possible that instead of getting a regular kube object, we're actually getting the `DeletedFinalStateUnknown` struct as `obj` which requires a different method to get the `key` out.

ishustava: That's a very useful investigation! 

Looking at the implementation of the `DeletionHandlingMetaNamespaceKeyFunc`, I'm curious if `obj` can be a `DeletedFinalStateUnknown` type, then we'd be adding that to our queue below on line 83. In that case, do we need to do the same check and extract the actual object from the `DeletedFinalStateUnknown` object?

Or if the included object information is stale, then should our controller know about that? I mean if the information in the object can be stale in some edge case, then it seems like we can't rely on the included object 100% of the time and need some sort of "back-up" method to still process the deletion correctly.

lkysow: Great catch. We do need to handle unwrapping `DeletedFinalStateUnknown`. I've modified the code to do so and also added tests. I could find no better way to test than pulling out the `DeleteFunc()` into a private function and calling it in tests.

I've updated the docs on `Delete()` to say that `obj` may be out of date. In our specific use-case the field we're accessing is `Status.HostIP` which afaik can't change once the pod is scheduled so we're okay to use the potentially stale object. Even if it's possibly wrong, we still have the `Reconcile()` loop to save us.

ishustava (approved):
Looks great, Luke! I really like the idea of adding the `obj` to delete events! Thanks for adding the fix with unwrapping `DeletedFinalStateUnknown` objects.

PR #432, opened 2021-02-03
Refactor inject-connect command

lkysow:
This refactor paves the way for adding the cleanup controller. It
removes the if/else statement around starting the mutating webhook
server when health checks are enabled vs when they're not and instead
starts it the same way in both cases.

Previously the code looked like:

```
if healthChecks {
  // start goroutine for health checks
  // start goroutine for webhook server
  // use select{} to wait
} else {
  // start webhook server
}
```

Now it looks like:

```
// start goroutine for webhook server
if healthChecks {
  // start goroutine for health checks
}
// use select{} to wait
```

This makes it easier to later do the below because we're already starting everything in goroutines and using select to wait:

```
// start goroutine for webhook server
if cleanup {
  // start goroutine for cleanup controller
}
if healthChecks {
  // start goroutine for health checks
}
// use select{} to wait
```

Otherwise we'd need to do:
```
if healthChecks || cleanup {
  if healthChecks {
    // start goroutine for health checks
  }
  if cleanup {
    // start goroutine for webhook server
  }
  // use select{} to wait
} else {
  // start webhook server
}
```

Somewhat related, it also removes special validation around the CONSUL_HTTP_ADDR. This is because I discovered that earlier in the file we were already doing:
```go
	// create Consul API config object
	cfg := api.DefaultConfig()
	c.http.MergeOntoConfig(cfg)
```

This code already picks up `CONSUL_HTTP_ADDR` and so I don't think it needs to be done separately. This also matches the pattern in two other locations where we don't explicitly validate for that env var (https://github.com/hashicorp/consul-k8s/blob/master/subcommand/create-federation-secret/command.go#L191-L209, https://github.com/hashicorp/consul-k8s/blob/master/subcommand/get-consul-client-ca/command.go#L127-L154).

How I've tested this PR:
* I ran the connect inject acceptance tests locally

* You can run the acceptance tests with `go test ./... -v -p 1 -run TestConnectInject -failfast -no-cleanup-on-failure -consul-k8s-image=ghcr.io/lkysow/consul-k8s-dev:feb03-inject-refactor3`

Checklist:
- [x] Tests added

Fixes https://github.com/hashicorp/consul-k8s/issues/436

### DIFF
--- a/subcommand/inject-connect/command_test.go
+++ b/subcommand/inject-connect/command_test.go
@@ -153,11 +153,6 @@ func TestRun_FlagValidation(t *testing.T) {
 			},
 			expErr: "request must be <= limit: -consul-sidecar-cpu-request value of \"50m\" is greater than the -consul-sidecar-cpu-limit value of \"25m\"",
 		},
-		{
-			flags: []string{"-consul-k8s-image", "hashicorpdev/consul-k8s:latest", "-consul-image", "foo", "-envoy-image", "envoy:1.16.0",
-				"-enable-health-checks-controller=true"},
-			expErr: "CONSUL_HTTP_ADDR is not specified",
-		},
 	}
 
 	for _, c := range cases {
@@ -192,7 +187,7 @@ func TestRun_ResourceLimitDefaults(t *testing.T) {
 	require.Equal(t, cmd.flagConsulSidecarMemoryLimit, "50Mi")
 }
 
-func TestRun_ValidationHealthCheckEnv(t *testing.T) {
+func TestRun_ValidationConsulHTTPAddr(t *testing.T) {
 	cases := []struct {
 		name    string
 		envVars []string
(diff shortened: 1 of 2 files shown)


subcommand/inject-connect/command_test.go
```diff
@@ -153,11 +153,6 @@ func TestRun_FlagValidation(t *testing.T) {
 			},
 			expErr: "request must be <= limit: -lifecycle-sidecar-cpu-request value of \"50m\" is greater than the -lifecycle-sidecar-cpu-limit value of \"25m\"",
 		},
-		{
```
lkysow, line 156: I removed this test because we're using `cfg := api.DefaultConfig()` now defaults the address if it's not set in the env var (and therefore doesn't error out). This no longer will error if that env var isn't set, instead it will set the address to 127.0.0.1 (which will likely break later) so this is a reduction in safety but does match the other locations mentioned in the PR description.

subcommand/inject-connect/command_test.go
```diff
@@ -192,18 +187,18 @@ func TestRun_ResourceLimitDefaults(t *testing.T) {
...
+func TestRun_ValidationConsulHTTPAddr(t *testing.T) {
 	cases := []struct {
 		name    string
 		envVars []string
 		flags   []string
 		expErr  string
 	}{
 		{
-			envVars: []string{api.HTTPAddrEnvName, "0.0.0.0:999999"},
+			envVars: []string{api.HTTPAddrEnvName, "%"},
```
lkysow, line 198: had to switch to `%` because now that we're prepending `http://`, `http://0.0.00:999999` isn't  actually an invalid url.

kschoche (approved):
💯 well done!

ishustava (approved):
👍 Looks good!

ishustava:
I have tested it out too and it all works!

PR #434, opened 2021-02-04
Set User-Agent header on Consul API calls

lkysow:
* Upgrade to latest consul api package
* Create common consul package that implements NewClient(). This
  function returns a Consul client that will set the header
  `User-Agent: consul-k8s/<version>` on all calls to Consul.
* Swap to using this new constructor everywhere.
* Add a linter that ensures we use our NewClient() function everywhere
  in the future. Cribbed from https://github.com/hashicorp/lint-consul-retry.

How I've tested this PR:
* unit test
* ran a request printer: `docker run -p 8501:8443 --rm -t mendhak/http-https-echo:17`

  then ran `CONSUL_HTTP_SSL_VERIFY=false consul-k8s get-consul-client-ca -output-file /tmp/out -server-addr localhost -server-port 8501`
  
  ```
  "GET /v1/agent/connect/ca/roots HTTP/1.1" 200 458 "-" "consul-k8s/0.23.0-dev"
  -----------------
  {
      "path": "/v1/agent/connect/ca/roots",
      "headers": {
          "host": "localhost:8501",
          "user-agent": "consul-k8s/0.23.0-dev",
          "accept-encoding": "gzip"
      },
      "method": "GET",
      "body": "",
      "fresh": false,
      "hostname": "localhost",
  ```

Checklist:
- [x] Tests added


namespaces/namespaces_test.go
```diff
@@ -107,6 +107,7 @@ func TestEnsureExists_CreatesNS(tt *testing.T) {
 			})
 			req.NoError(err)
 			defer consul.Stop()
+			consul.WaitForLeader(t)
```
lkysow, line 110: flaky test

consul/consul.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"fmt"
+
+	"github.com/hashicorp/consul-k8s/version"
+	capi "github.com/hashicorp/consul/api"
+)
+
+// NewClient returns a Consul API client. It adds a required User-Agent
+// header that describes the version of consul-k8s making the call.
+func NewClient(config *capi.Config) (*capi.Client, error) {
```
lkysow, line 12: @ishustava what do you think about this new package location?

thisisnotashwin: lol. A joke that haunts @ndhanushkodi, @ishustava and @thisisnotashwin!

ishustava: Makes sense!

connect-inject/health_check_resource.go
```diff
@@ -278,7 +279,7 @@ func (h *HealthCheckResource) getConsulClient(pod *corev1.Pod) (*api.Client, err
 	if pod.Annotations[annotationConsulNamespace] != "" {
 		localConfig.Namespace = pod.Annotations[annotationConsulNamespace]
 	}
-	localClient, err := api.NewClient(localConfig)
+	localClient, err := consul.NewClient(localConfig)
```
kschoche, line 282: is it a bit weird to be using `localConfig := api.DefaultConfig()` and then passing that to `consul.NewClient(localConfig)` ? 
Without duplicating too much, is it possible to make these calls to the same library?

kschoche (approved):
This is great stuff, and I like the linter too!
Tested it out and looks good.
I left one comment which isn't a blocker but I'd like to get other opinions on whether or not we should keep using `DefaultConfig()` from API or duplicate it inside the new lib from a code readability standpoint.
So cool 
```
consul-k8s % go run hack/lint-api-new-client/main.go
Found code using github.com/hashicorp/consul/api.NewClient()
instead of github.com/hashicorp/consul-k8s/consul.NewClient()
in the following files:
-  github.com/hashicorp/consul-k8s/connect-inject/health_check_resource.go
exit status 1
```

connect-inject/health_check_resource.go
lkysow: Yeah it's a good idea. Here's what I found:

Out of the 9 places we use `consul.NewClient()`, in 4 of them we call `api.DefaultConfig` so they'd benefit from implementing `DefaultConfig()` in this new package. Note that in all those locations we still end up importing `consul/api` because another part of the code uses it.

In the other 5 locations, we're using the `api.Config` struct directly. I don't think it makes sense to duplicate this struct in our `consul` package because it's got a lot of fields (https://github.com/hashicorp/consul/blob/master/api/api.go#L281) and we'd need to implement a transform function to get it back into the `api.Config` struct.

Here's the commit for what it looks like: https://github.com/hashicorp/consul-k8s/commit/e56248073cd36d4d13cd3b4f09c9bebbbf6be9e8

Do you think we should do it despite not all cases being covered?

ndhanushkodi: I'm thinking of the consul package as a place where common use cases for the api are covered there. That makes sense for things like NewClient which is a couple lines of functionality, and very commonly used, but the commit you linked seems like just a redirection for a call to the api package and if we're ok with using the api package in other more custom cases, I don't think we need to redirect one line calls to the consul package.

ndhanushkodi (approved):
Looks great!!

connect-inject/health_check_resource.go
kschoche: I think we’re good to go on this one, I wanted it more for looks than function and if it doesn’t achieve the looks factor because it cannot be consistently applied then there is no reason to bother with it IMHO 💯

lkysow: okay sounds good thanks for the feedback!

PR #435, opened 2021-02-04
Cleanup controller support for Consul NS's

lkysow:
Checklist:
- [x] Tests added

Implementation Notes:
* There's no Consul API to list all services in all namespaces so instead we need to first get all namespaces and then iterate through them and list all services.
* To keep the OSS and enterprise code paths as close as possible, we use a fake namespace in oss with value empty string `""` which can then be safely passed to `QueryOptions{Namespace: ""}` without error.

Acceptance tests: https://github.com/hashicorp/consul-helm/pull/826.
Docker image: `ghcr.io/lkysow/consul-k8s-dev:feb11`


ishustava (approved):
Looks great!

ndhanushkodi (approved):
Looks great!!

PR #433, opened 2021-02-04
Implement cleanup controller

lkysow:
* Runs by default
* Watches for pod delete events and ensures there are no stale Consul
service instances
* Runs a Reconcile loop on a timer that checks all Consul services
* Adds new connect-inject flags: `-enable-cleanup-controller`,
`-cleanup-controller-reconcile-period`.
* Adds new `k8s-namespace` meta key to connect service instances
* Adds new server-acl-init flag: `-enable-cleanup-controller` (default
true) and modifies the ACL templates.

How I've tested this PR:
* manually and with acceptance tests: https://github.com/hashicorp/consul-helm/pull/806
* reconcile can be tested by registering a service directly with a consul client:
  * port-forward to consul client on 8500
  * use `svc.json`:
    ```
    {
      "ID": "redis1",
      "Name": "redis",
      "Tags": ["primary", "v1"],
      "Address": "127.0.0.1",
      "Port": 8000,
      "Meta": {
        "pod-name": "dead",
        "k8s-namespace": "default"
      }
    }
    ```
   * `curl -X PUT -v --data @svc.json localhost:8500/v1/agent/service/register`
   * wait for reconcile to run (can bump default time with the helm value)
    ```
    2021-02-10T00:50:32.329Z [DEBUG] cleanupResource: starting reconcile
    2021-02-10T00:50:32.342Z [INFO]  cleanupResource: found service instance from terminated pod still registered: pod=dead id=redis1
    2021-02-10T00:50:32.346Z [INFO]  cleanupResource: service instance deregistered: id=redis1
    2021-02-10T00:50:32.346Z [DEBUG] cleanupResource: finished reconcile
    ```

How I expect reviewers to test this PR:
* try it out with `ghcr.io/lkysow/consul-k8s-dev:feb09` and see if you can break it

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

PR #437, opened 2021-02-11
Refactor health resource to match cleanup resource

lkysow:
- refactor Run loop
- use ConsulPort and ConsulScheme instead of ConsulURL
- reconcile() does not return an error since it is logged during that
function

Matches https://github.com/hashicorp/consul-k8s/pull/433#discussion_r573365394


ishustava (approved):

kschoche (approved):
Beautiful work.

PR #438, opened 2021-02-12
Add metrics merging and Prometheus annotations to connect-injected Pods

ndhanushkodi:
Adds metrics merging feature to consul-k8s.

Changes proposed in this PR:
- inject-connect takes default metrics flags
- supports annotations for overriding the default flags
- inject-connect passes metrics flags to connect-inject handler
- handler now has functions to grab configuration from the default flags or the overridden values from the annotations
- from the handler:
  - init container adds envoy_prometheus_bind_addr and flags to consul connect envoy
  - consul-sidecar is given the appropriate flags to run the merged metrics server
  - added prometheus annotations to pod to scrape from

How I've tested this PR:
- consul-helm (`test-metrics` branch)
- consul-k8s (`merged-metrics` branch)
- consul (`metrics-merging` branch)
- Build the consul and consul-k8s images or use:
  - `gcr.io/nitya-293720/consul-k8s-dev:metrics28`
  - `gcr.io/nitya-293720/consul-dev:metrics10`
  - Set those in values.yaml, make sure connectInject is enabled
- Deploy prometheus: `helm install prom prometheus-community/prometheus`
- Helm install consul
- `k apply -f example-app.yaml`
example-app.yaml: 
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: prometheus-example-app
  name: prometheus-example-app
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: prometheus-example-app
  template:
    metadata:
      annotations:
        'consul.hashicorp.com/connect-inject': 'true'
      labels:
        app.kubernetes.io/name: prometheus-example-app
    spec:
      containers:
      - name: prometheus-example-app
        image: quay.io/brancz/prometheus-example-app:v0.3.0
        ports:
        - name: web
          containerPort: 8080
```
- Port forward to prometheus
  - `export POD_NAME=$(kubectl get pods --namespace default -l "app=prometheus,component=server" -o jsonpath="{.items[0].metadata.name}")`
  - `kubectl --namespace default port-forward $POD_NAME 9090`
- go to localhost:9090 and look at targets, you should see prometheus-example-app under kubernetes pods
- try graphing the "version" metric (an app metric) and an envoy metric

- Give the steps above a shot
- Code review

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/consul_sidecar.go
+++ b/connect-inject/consul_sidecar.go
@@ -1,12 +1,13 @@
 package connectinject
 
 import (
+	"fmt"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 )
 
-func (h *Handler) consulSidecar(pod *corev1.Pod) corev1.Container {
+func (h *Handler) consulSidecar(pod *corev1.Pod) (corev1.Container, error) {
 	command := []string{
 		"consul-k8s",
 		"consul-sidecar",
@@ -19,6 +20,32 @@ func (h *Handler) consulSidecar(pod *corev1.Pod) corev1.Container {
 
 	if period, ok := pod.Annotations[annotationSyncPeriod]; ok {
 		command = append(command, "-sync-period="+strings.TrimSpace(period))
+	}
+
+	run, err := h.shouldRunMergedMetricsServer(pod)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+
+	// If we need to run the merged metrics server, configure consul
+	// sidecar with the appropriate metrics flags.
+	if run {
+		mergedMetricsPort, err := h.mergedMetricsPort(pod)
+		if err != nil {
+			return corev1.Container{}, err
+		}
+		serviceMetricsPath := h.serviceMetricsPath(pod)
+
+		// Don't need to check the error since it's checked in the call to
+		// h.shouldRunMergedMetricsServer() above.
+		serviceMetricsPort, _ := h.serviceMetricsPort(pod)
+
+		command = append(command, []string{
+			"-enable-metrics-merging=true",
+			fmt.Sprintf("-merged-metrics-port=%s", mergedMetricsPort),
+			fmt.Sprintf("-service-metrics-port=%s", serviceMetricsPort),
+			fmt.Sprintf("-service-metrics-path=%s", serviceMetricsPath),
+		}...)
 	}
 
 	envVariables := []corev1.EnvVar{
@@ -65,5 +92,5 @@ func (h *Handler) consulSidecar(pod *corev1.Pod) corev1.Container {
 		},
 		Command:   command,
 		Resources: h.ConsulSidecarResources,
-	}
+	}, nil
 }
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -42,6 +42,17 @@ type initContainerCommandData struct {
 	// The PEM-encoded CA certificate to use when
 	// communicating with Consul clients
 	ConsulCACert string
+	// EnableMetrics adds a listener to Envoy where Prometheus will scrape
+	// metrics from.
+	EnableMetrics bool
+	// PrometheusScrapeListener configures the listener on Envoy where
+	// Prometheus will scrape metrics from.
+	PrometheusScrapeListener string
+	// PrometheusScrapePath configures the path on the listener on Envoy where
+	// Prometheus will scrape metrics from.
+	PrometheusScrapePath string
+	// PrometheusBackendPort configures where the listener on Envoy will point to.
+	PrometheusBackendPort string
 }
 
 type initContainerCommandUpstreamData struct {
@@ -83,6 +94,41 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
 		if port, _ := portValue(pod, raw); port > 0 {
 			data.ServicePort = port
 		}
+	}
+
+	// If metrics are enabled, the init container should set up
+	// envoy_prometheus_bind_addr so there's a listener on 0.0.0.0 that points
+	// to a metrics backend. The metrics backend is determined by the call to
+	// h.shouldRunMergedMetricsServer(). If there is a merged metrics server,
+	// the backend would be that server. If we are not running the merged
+	// metrics server, the backend should just be the Envoy metrics endpoint.
+	enableMetrics, err := h.enableMetrics(pod)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+	data.EnableMetrics = enableMetrics
+
+	prometheusScrapePort, err := h.prometheusScrapePort(pod)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+	data.PrometheusScrapeListener = fmt.Sprintf("0.0.0.0:%s", prometheusScrapePort)
+
+	// This determines how to configure the consul connect envoy command: what
+	// metrics backend to use and what path to expose on the
+	// envoy_prometheus_bind_addr listener for scraping.
+	run, err := h.shouldRunMergedMetricsServer(pod)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+	if run {
+		prometheusScrapePath := h.prometheusScrapePath(pod)
+		mergedMetricsPort, err := h.mergedMetricsPort(pod)
+		if err != nil {
+			return corev1.Container{}, err
+		}
+		data.PrometheusScrapePath = prometheusScrapePath
+		data.PrometheusBackendPort = mergedMetricsPort
 	}
 
 	var tags []string
@@ -203,7 +249,7 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
 	var buf bytes.Buffer
 	tpl := template.Must(template.New("root").Parse(strings.TrimSpace(
 		initContainerCommandTpl)))
-	err := tpl.Execute(&buf, &data)
+	err = tpl.Execute(&buf, &data)
 	if err != nil {
 		return corev1.Container{}, err
 	}
@@ -317,6 +363,11 @@ services {
   }
 
   proxy {
+	{{- if .EnableMetrics }}
+	config {
+	  envoy_prometheus_bind_addr = "{{ .PrometheusScrapeListener }}"
+	}
+	{{- end }}
     destination_service_name = "{{ .ServiceName }}"
     destination_service_id = "${SERVICE_ID}"
     {{- if (gt .ServicePort 0) }}
@@ -389,6 +440,12 @@ chmod 444 /consul/connect-inject/acl-token
 # Generate the envoy bootstrap code
 /bin/consul connect envoy \
   -proxy-id="${PROXY_SERVICE_ID}" \
+  {{- if .PrometheusScrapePath }}
+  -prometheus-scrape-path="{{ .PrometheusScrapePath }}" \
+  {{- end }}
+  {{- if .PrometheusBackendPort }}
+  -prometheus-backend-port="{{ .PrometheusBackendPort }}" \
+  {{- end }}
   {{- if .AuthMethod }}
   -token-file="/consul/connect-inject/acl-token" \
   {{- end }}
--- a/connect-inject/container_init_test.go
+++ b/connect-inject/container_init_test.go
@@ -165,6 +165,51 @@ services {
 		},
 
 		{
+			"Metrics enabled with custom Prometheus scrape port",
+			func(pod *corev1.Pod) *corev1.Pod {
+				pod.Annotations[annotationService] = "web"
+				pod.Annotations[annotationEnableMetrics] = "true"
+				pod.Annotations[annotationPrometheusScrapePort] = "22222"
+				return pod
+			},
+			`proxy {
+	config {
+	  envoy_prometheus_bind_addr = "0.0.0.0:22222"
+	}
+    destination_service_name = "web"
+    destination_service_id = "${SERVICE_ID}"
+  }`,
+			"",
+		},
+
+		{
+			"When running the merged metrics server, configures consul connect envoy command",
+			func(pod *corev1.Pod) *corev1.Pod {
+				// The annotations to enable metrics, enable merging, and
+				// service metrics port make the condition to run the merged
+				// metrics server true. When that is the case,
+				// prometheusScrapePath and mergedMetricsPort should get
+				// rendered as -prometheus-scrape-path and
+				// -prometheus-backend-port to the consul connect envoy command.
+				pod.Annotations[annotationService] = "web"
+				pod.Annotations[annotationEnableMetrics] = "true"
+				pod.Annotations[annotationEnableMetricsMerging] = "true"
+				pod.Annotations[annotationMergedMetricsPort] = "20100"
+				pod.Annotations[annotationServiceMetricsPort] = "1234"
+				pod.Annotations[annotationPrometheusScrapePort] = "22222"
+				pod.Annotations[annotationPrometheusScrapePath] = "/scrape-path"
+				return pod
+			},
+			`# Generate the envoy bootstrap code
+/bin/consul connect envoy \
+  -proxy-id="${PROXY_SERVICE_ID}" \
+  -prometheus-scrape-path="/scrape-path" \
+  -prometheus-backend-port="20100" \
+  -bootstrap > /consul/connect-inject/envoy-bootstrap.yaml`,
+			"",
+		},
+
+		{
 			"Upstream",
 			func(pod *corev1.Pod) *corev1.Pod {
 				pod.Annotations[annotationService] = "web"
--- a/connect-inject/handler.go
+++ b/connect-inject/handler.go
@@ -82,6 +82,17 @@ const (
 	annotationSidecarProxyMemoryLimit   = "consul.hashicorp.com/sidecar-proxy-memory-limit"
 	annotationSidecarProxyMemoryRequest = "consul.hashicorp.com/sidecar-proxy-memory-request"
 
+	// annotations for metrics to configure where Prometheus scrapes
+	// metrics from, whether to run a merged metrics endpoint on the consul
+	// sidecar, and configure the connect service metrics.
+	annotationEnableMetrics        = "consul.hashicorp.com/enable-metrics"
+	annotationEnableMetricsMerging = "consul.hashicorp.com/enable-metrics-merging"
+	annotationMergedMetricsPort    = "consul.hashicorp.com/merged-metrics-port"
+	annotationPrometheusScrapePort = "consul.hashicorp.com/prometheus-scrape-port"
+	annotationPrometheusScrapePath = "consul.hashicorp.com/prometheus-scrape-path"
+	annotationServiceMetricsPort   = "consul.hashicorp.com/service-metrics-port"
+	annotationServiceMetricsPath   = "consul.hashicorp.com/service-metrics-path"
+
 	// annotationEnvoyExtraArgs is a space-separated list of arguments to be passed to the
 	// envoy binary. See list of args here: https://www.envoyproxy.io/docs/envoy/latest/operations/cli
 	// e.g. consul.hashicorp.com/envoy-extra-args: "--log-level debug --disable-hot-restart"
@@ -94,6 +105,8 @@ const (
 
 	// annotationConsulNamespace is the Consul namespace the service is registered into.
 	annotationConsulNamespace = "consul.hashicorp.com/consul-namespace"
+
+	defaultServiceMetricsPath = "/metrics"
 )
 
 var (
@@ -182,6 +195,15 @@ type Handler struct {
 	DefaultProxyCPULimit      resource.Quantity
 	DefaultProxyMemoryRequest resource.Quantity
 	DefaultProxyMemoryLimit   resource.Quantity
+
+	// Default metrics settings. These will configure where Prometheus scrapes
+	// metrics from, and whether to run a merged metrics endpoint on the consul
+	// sidecar. These can be overridden via pod annotations.
+	DefaultEnableMetrics        bool
+	DefaultEnableMetricsMerging bool
+	DefaultMergedMetricsPort    string
+	DefaultPrometheusScrapePort string
+	DefaultPrometheusScrapePath string
 
 	// Resource settings for init container. All of these fields
 	// will be populated by the defaults provided in the initial flags.
@@ -352,7 +374,15 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
 			},
 		}
 	}
-	connectContainer := h.consulSidecar(&pod)
+	connectContainer, err := h.consulSidecar(&pod)
+	if err != nil {
+		h.Log.Error("Error configuring consul sidecar container", "err", err, "Request Name", req.Name)
+		return &v1beta1.AdmissionResponse{
+			Result: &metav1.Status{
+				Message: fmt.Sprintf("Error configuring consul sidecar container: %s", err),
+			},
+		}
+	}
 	patches = append(patches, addContainer(
 		pod.Spec.Containers,
 		[]corev1.Container{esContainer, connectContainer},
@@ -364,6 +394,22 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
 		map[string]string{
 			annotationStatus: injected,
 		})...)
+
+	// Add annotations for metrics
+	promAnnotations, err := h.prometheusAnnotations(&pod)
+	if err != nil {
+		h.Log.Error("Error configuring prometheus annotations", "err", err, "Request Name", req.Name)
+		return &v1beta1.AdmissionResponse{
+			Result: &metav1.Status{
+				Message: fmt.Sprintf("Error configuring prometheus annotations: %s", err),
+			},
+		}
+	}
+	if promAnnotations != nil {
+		patches = append(patches, updateAnnotation(
+			pod.Annotations,
+			promAnnotations)...)
+	}
 
 	// Add Pod label for health checks
 	patches = append(patches, updateLabels(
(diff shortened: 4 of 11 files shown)


connect-inject/container_init_test.go
```diff
@@ -158,6 +162,122 @@ services {
 }
 `,
 			"",
+			Handler{},
+			"",
+		},
+
+		// TODO do we need all of these tests, if the handler tests the enableMetrics and prometheusScrapePort methods?
```
ndhanushkodi, line 169: There are some common validation methods in subcommand/common, and connect-inject/handler. I think the edge case testing should be handled there, and here it should just be "happy path" cases of enabling metrics and configuring the prometheus scrape port and seeing they are configured.

connect-inject/handler.go
```diff
@@ -504,6 +532,91 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
 	return nil
 }
 
+// enableMetrics returns the default value in the handler, or overrides that
```
ndhanushkodi, line 535: All of these methods should be tested in handler_test

subcommand/consul-sidecar/command.go
```diff
@@ -44,6 +54,11 @@ func (c *Command) init() {
 		"Log verbosity level. Supported values (in order of detail) are \"trace\", "+
 			"\"debug\", \"info\", \"warn\", and \"error\". Defaults to info.")
 
+	// TODO: do these need to be validated if they're validated before being passed in?
```
ndhanushkodi, line 57: QUESTION: Do we need to validate these flags again? They are being validated by inject-connect, and then values that are passed to consul-sidecar are done via code.

subcommand/consul-sidecar/command.go
```diff
@@ -80,29 +95,66 @@ func (c *Command) Run(args []string) int {
...
 	go func() {
 		select {
 		case sig := <-c.sigCh:
-			logger.Info(fmt.Sprintf("%s received, shutting down", sig))
+			c.logger.Info(fmt.Sprintf("%s received, shutting down", sig))
 			cancelFunc()
 			return
 		}
 	}()
+
```
ndhanushkodi, line 127: TODO: needs more logging, it's hard to tell if this is starting up without errors and if it's doing ok scraping metrics

subcommand/consul-sidecar/command.go
```diff
@@ -125,12 +177,60 @@ func (c *Command) Run(args []string) int {
 		// Re-loop after syncPeriod or exit if we receive interrupt or terminate signals.
 		case <-time.After(c.flagSyncPeriod):
 			continue
-		case <-ctx.Done():
+		case <-signalCtx.Done():
 			return 0
 		}
 	}
 }
```
ndhanushkodi, line 185: How to test the merged metrics server? Is it worth running it? It has nowhere realistic to scrape from.

Some interesting behaviors maybe worth testing: if it can't fetch app metrics, it should log and just return envoy metrics in the response. If it can't fetch envoy metrics, it should log and not fetch anything.

Should we make a mock server that either returns "fake metrics" or returns a 404, and 404 simulates no metrics endpoint and "fake metrics" simulates metrics.

thisisnotashwin: We could potentially have a mock server that we start when we want to test "fake metrics" but dont start at all when we want to test it returning a `404`

ndhanushkodi: Thanks ashwin! I like this idea.

connect-inject/container_init.go
```diff
@@ -85,6 +96,35 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
 		}
 	}
 
+	enableMetrics, err := h.enableMetrics(pod)
```
kschoche, line 99: Are there common cases where `enableMetrics` is true but `run` is not?
If so, maybe we could document that somewhere in the code?
If not, you could potentially just wrap everything inside `if run`. 🤷🏽

connect-inject/handler.go
```diff
@@ -8,7 +8,7 @@ import (
 	"net/http"
 	"strconv"
 
-	"github.com/deckarep/golang-set"
+	mapset "github.com/deckarep/golang-set"
```
kschoche, line 11: I'm assuming your IDE brought this in, happens to everyone, could you remove it?

connect-inject/handler.go
```diff
@@ -504,6 +548,173 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+	enableMetricsMerging, err := h.enableMetricsMerging(pod)
+	if err != nil {
+		return false, err
+	}
+	serviceMetricsPort, err := h.serviceMetricsPort(pod)
+	if err != nil {
+		return false, err
+	}
+
+	// Don't need to check error here since serviceMetricsPort has been
```
kschoche, line 680: 💯

connect-inject/handler.go
```diff
@@ -504,6 +548,173 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+	// validated by calling h.serviceMetricsPort above
+	smp, _ := strconv.Atoi(serviceMetricsPort)
+	// h.Log.Info("shouldRun", "enableMetrics", enableMetrics, "enableMetricsMerging", enableMetricsMerging, "serviceMetricsPort", serviceMetricsPort)
```
kschoche, line 683:
```suggestion
	
```
I assume you meant to take this out!

subcommand/consul-sidecar/command.go
```diff
@@ -44,6 +56,10 @@ func (c *Command) init() {
 		"Log verbosity level. Supported values (in order of detail) are \"trace\", "+
 			"\"debug\", \"info\", \"warn\", and \"error\". Defaults to info.")
 
+	c.flagSet.BoolVar(&c.flagEnableMetricsMerging, "enable-metrics-merging", false, "Enables consul sidecar to run a merged metrics endpoint. Defaults to false.")
+	c.flagSet.StringVar(&c.flagMergedMetricsPort, "merged-metrics-port", "20100", "Port to serve merged Envoy and application metrics. Defaults to 20100.")
+	c.flagSet.StringVar(&c.flagServiceMetricsPort, "service-metrics-port", "0", "Port where application metrics are being served. Defaults to 20100.")
```
kschoche, line 61: You've got the default value set to "0" and the text says "defaults to 20100", assuming this is a typo?
```suggestion
	c.flagSet.StringVar(&c.flagServiceMetricsPort, "service-metrics-port", "20100", "Port where application metrics are being served. Defaults to 20100.")
```

subcommand/common/common.go
```diff
@@ -30,3 +32,16 @@ func Logger(level string) (hclog.Logger, error) {
...
 }
+
+// ValidatePort converts flags representing ports into integer and validates
+// that it's in the port range.
+func ValidatePort(flagName, flagValue string) error {
+	port, err := strconv.Atoi(flagValue)
+	if err != nil {
+		return errors.New(fmt.Sprintf("%s value of %s is not a valid integer.", flagName, flagValue))
+	}
+	if port <= 1023 || port > 65353 {
```
kschoche, line 43: Bonus points if you can find a library that does this validation, I looked and couldn't find one.
In lieu of reading `/proc/sys/net/ipv4/ip_local_port_range`, I think this is safe for now.
In theory this may be an invalid range based on how the kernel/network is configured.

kschoche: Also a really dumb question on my part but did you mean `65535` instead of `65353`?

subcommand/consul-sidecar/command.go
```diff
@@ -80,57 +96,165 @@ func (c *Command) Run(args []string) int {
...
+func (c *Command) shutdownMetricsServer(server *http.Server) {
+	shutdownCtx, shutdownCancelFunc := context.WithTimeout(context.Background(), metricsServerShutdownTimeout)
+	defer shutdownCancelFunc()
+
+	// If the server hasn't started up yet before this function is called, it
+	// could be nil.
+	if server != nil {
+		c.logger.Info("Merged metrics server exists, attempting to gracefully shut down server.")
+		if err := server.Shutdown(shutdownCtx); err != nil {
+			c.logger.Error(fmt.Sprintf("Server shutdown failed: %s", err))
```
kschoche, line 195: Should we return/exit from here so we don't print the next line in failure cases?

subcommand/consul-sidecar/command.go
```diff
@@ -80,57 +96,165 @@ func (c *Command) Run(args []string) int {
...
+// mergedMetricsHandler has the logic to append both Envoy and service metrics
+// together, logging if it's unsuccessful at either.
+func (c *Command) mergedMetricsHandler(rw http.ResponseWriter, r *http.Request) {
+	// The default http.Client timeout is indefinite, so adding a timeout makes
+	// sure that requests don't hang.
+	netClient := &http.Client{
+		Timeout: time.Second * 10,
+	}
+
+	c.logger.Info("Handling /stats/prometheus")
```
kschoche, line 221: Curious if you think this might get a bit spammy over time in the logs?

subcommand/consul-sidecar/command.go
```diff
@@ -80,57 +96,165 @@ func (c *Command) Run(args []string) int {
...
+		return
+	}
+
+	appMetricsAddr := fmt.Sprintf("http://127.0.0.1:%s%s", c.flagServiceMetricsPort, c.flagServiceMetricsPath)
+	appMetrics, err := netClient.Get(appMetricsAddr)
+	if err != nil {
+		// If there are no service metrics, we want to log the warning and continue
+		// to write envoy metrics into the response, so we won't return if
+		// there's an error.
+		c.logger.Error(fmt.Sprintf("Error scraping service metrics: %s", err.Error()))
```
kschoche, line 238: super minor nit: you mentioned this is a warning in the comments and logging as an error, is it really an error?

kschoche (approved):
@ndhanushkodi  This is awesome work!! I struggled a bit to find mistakes (🥇 ) so my apologies if there are some "nit" comments.
I've left a few comments along with a few suggestions.
Cheers!

connect-inject/container_init.go
```diff
@@ -317,6 +357,11 @@ services {
   }
 
   proxy {
+	{{- if .EnableMetrics }}
+	config {
+		envoy_prometheus_bind_addr = "{{ .PrometheusScrapeListener }}"
```
thisisnotashwin, line 362:
```suggestion
	      envoy_prometheus_bind_addr = "{{ .PrometheusScrapeListener }}"
```

connect-inject/handler.go
```diff
@@ -504,6 +548,173 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+// serviceMetricsPath returns a default of /metrics, or overrides
+// that with the annotation if provided.
+func (h *Handler) serviceMetricsPath(pod *corev1.Pod) (string, error) {
+	defaultServiceMetricsPath := "/metrics"
+
+	if raw, ok := pod.Annotations[annotationServiceMetricsPath]; ok && raw != "" {
+		return raw, nil
+	}
+
+	return defaultServiceMetricsPath, nil
```
thisisnotashwin, line 633: You could potentially have a file level const `const string defaultServiceMetricsPath = "/metrics"`
```suggestion
	if raw, ok := pod.Annotations[annotationServiceMetricsPath]; ok && raw != "" {
		return raw, nil
	}
	return "/metrics", nil
```

connect-inject/handler.go
```diff
@@ -504,6 +548,173 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+// prometheusScrapePath returns the default value in the handler, or overrides
+// that with the annotation if provided.
+func (h *Handler) prometheusScrapePath(pod *corev1.Pod) (string, error) {
+	defaultPrometheusScrapePath := h.DefaultPrometheusScrapePath
+
+	if raw, ok := pod.Annotations[annotationPrometheusScrapePath]; ok && raw != "" {
+		return raw, nil
+	}
+
+	return defaultPrometheusScrapePath, nil
```
thisisnotashwin, line 604:
```suggestion
	if raw, ok := pod.Annotations[annotationPrometheusScrapePath]; ok && raw != "" {
		return raw, nil
	}
	return h.DefaultPrometheusScrapePath, nil
```

thisisnotashwin: Also, do we need to return an error here? It doesnt look like it ever gets set.

connect-inject/handler.go
thisisnotashwin: Also, do we need to return an error here? It doesnt look like it ever gets set.

connect-inject/handler.go
```diff
@@ -504,6 +548,173 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+
+	return determineAndValidatePort(pod, annotationMergedMetricsPort, defaultMergedMetricsPort)
+}
+
+// prometheusScrapePort returns the default value in the handler, or overrides
+// that with the annotation if provided.
+func (h *Handler) prometheusScrapePort(pod *corev1.Pod) (string, error) {
+	defaultPrometheusScrapePort := h.DefaultPrometheusScrapePort
+
+	return determineAndValidatePort(pod, annotationPrometheusScrapePort, defaultPrometheusScrapePort)
```
thisisnotashwin, line 592:
```suggestion
	return determineAndValidatePort(pod, annotationPrometheusScrapePort, h.DefaultPrometheusScrapePort)
```

connect-inject/handler.go
```diff
@@ -504,6 +548,173 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+	}
+	return enabled, nil
+}
+
+// mergedMetricsPort returns the default value in the handler, or overrides
+// that with the annotation if provided.
+func (h *Handler) mergedMetricsPort(pod *corev1.Pod) (string, error) {
+	defaultMergedMetricsPort := h.DefaultMergedMetricsPort
+
+	return determineAndValidatePort(pod, annotationMergedMetricsPort, defaultMergedMetricsPort)
```
thisisnotashwin, line 584:
```suggestion
	return determineAndValidatePort(pod, annotationMergedMetricsPort, h.DefaultMergedMetricsPort)
```

subcommand/common/common.go
thisisnotashwin, line 43:
```suggestion
	if port < 1024 || port > 65353 {
```

thisisnotashwin: As Kyle mentioned, is it `65535` and not `65353`? Could you add a one-liner here and on `handler.go:700` as to why this range.

subcommand/consul-sidecar/command.go
```diff
@@ -80,57 +96,165 @@ func (c *Command) Run(args []string) int {
...
+	// The default http.Client timeout is indefinite, so adding a timeout makes
+	// sure that requests don't hang.
+	netClient := &http.Client{
+		Timeout: time.Second * 10,
+	}
+
+	c.logger.Info("Handling /stats/prometheus")
+
+	envoyMetricsAddr := fmt.Sprintf("http://127.0.0.1:19000/stats/prometheus")
+	envoyMetrics, err := netClient.Get(envoyMetricsAddr)
```
thisisnotashwin, line 224: You can constantize `envoyMetricsAddr` and pass the constant directly to `netCliet.Get()`

subcommand/consul-sidecar/command.go
```diff
@@ -80,57 +96,165 @@ func (c *Command) Run(args []string) int {
...
+
+	mergedMetricsServerAddr := fmt.Sprintf("127.0.0.1:%s", c.flagMergedMetricsPort)
+	server := &http.Server{Addr: mergedMetricsServerAddr, Handler: mux}
+
+	return server
+}
+
+// mergedMetricsHandler has the logic to append both Envoy and service metrics
+// together, logging if it's unsuccessful at either.
+func (c *Command) mergedMetricsHandler(rw http.ResponseWriter, r *http.Request) {
```
thisisnotashwin, line 214:
```suggestion
func (c *Command) mergedMetricsHandler(rw http.ResponseWriter, _ *http.Request) {
```

thisisnotashwin, line 214: I was giving this workflow a read and had a general question/suggestion:
Does it make sense to read and write the `envoy` metrics and then attempt to read and then write the `app` metrics? It might simplify some of the error handling and plus you wouldnt have to check if the response from the application metrics end point was `nil` as you can determine that from if the read attempt returned an error.
Just a thought.

thisisnotashwin (commented):
Hey Nitya! love this approach. it looks really great. have a few comments and suggestions but it looks excellent!

connect-inject/consul_sidecar.go
```diff
@@ -21,6 +22,32 @@ func (h *Handler) consulSidecar(pod *corev1.Pod) corev1.Container {
...
+			return corev1.Container{}, err
+		}
+		serviceMetricsPort, err := h.serviceMetricsPort(pod)
+		if err != nil {
+			return corev1.Container{}, err
+		}
+		serviceMetricsPath, err := h.serviceMetricsPath(pod)
+		if err != nil {
+			return corev1.Container{}, err
+		}
```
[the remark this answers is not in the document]
ndhanushkodi: `h.shouldRunMergedMetricsServer(pod)` calls `h.enableMetrics(pod)`, `h.enableMetricsMerging(pod)`, and `h.serviceMetricsPort(pod)`, so we could avoid checking the error for `serviceMetricsPort`. I can do that and add a comment describing why we're not checking the error.

connect-inject/consul_sidecar.go
```diff
@@ -21,6 +22,32 @@ func (h *Handler) consulSidecar(pod *corev1.Pod) corev1.Container {
...
+		command = append(command, "-enable-metrics-merging=true")
+		command = append(command, fmt.Sprintf("-merged-metrics-port=%s", mergedMetricsPort))
+		command = append(command, fmt.Sprintf("-service-metrics-port=%s", serviceMetricsPort))
+		command = append(command, fmt.Sprintf("-service-metrics-path=%s", serviceMetricsPath))
```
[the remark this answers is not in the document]
ndhanushkodi: Nice, I like this change!

connect-inject/container_init.go
ndhanushkodi: Yeah the case where you'd want `data.EnableMetrics` and `data.PrometheusScrapeListener` to be set but not the values in `if run` is if your app doesn't have any metrics and you still want Envoy sidecar metrics. So you'd enable metrics, but not metrics merging, which means prometheus annotations are still added to your app, and envoy_prometheus_bind_addr gets configured, but the merged metrics server wouldn't be running. I'll document that in a comment, that's a good thing to call out, thanks for catching it!

connect-inject/handler.go
ndhanushkodi: Yup!! Thanks!!

subcommand/consul-sidecar/command.go
ndhanushkodi: So this one shouldn't default to 20100. I had made a typo in the Defaults sentence. It is only set in connect-inject/consul_sidecar.go if: metrics and metrics merging are enabled, and the service metrics port is greater than 0. So I thought I could default `serviceMetricsPort` to 0, since if the merged metrics server is not running, it will not get used, and if it is running, it will definitely get set to something that is not 0. I changed the sentence to "Defaults to 0." I added a comment to explain this as well, let me know if its clear.

subcommand/common/common.go
ndhanushkodi: Oh you're absolutely right I don't know where my number came from, changed that! Yeah it would be nice to use a library but yeah I also couldn't find one.

subcommand/consul-sidecar/command.go
ndhanushkodi: Nice catch!

subcommand/consul-sidecar/command.go
ndhanushkodi: Yeah it'll be good to delete that especially since Prometheus is scraping from this endpoint periodically.

subcommand/consul-sidecar/command.go
ndhanushkodi: Changed it to logger.Warn, since that's really what it should be.

connect-inject/handler.go
ndhanushkodi: Nope we don't need an error for that method, I'll remove it

subcommand/common/common.go
ndhanushkodi: Yup!!

connect-inject/handler.go
ndhanushkodi: Changed it to use the const, and got rid of the error, thanks!!

subcommand/consul-sidecar/command.go
ndhanushkodi: Ah I like that flow. I changed it to match that, and added a return if there was an error trying to netClient.Get(appMetricsAddr), since we will have already written envoy metrics at that point.

connect-inject/handler.go
```diff
@@ -504,6 +550,163 @@ func (h *Handler) defaultAnnotations(pod *corev1.Pod, patches *[]jsonpatch.JsonP
...
+// If the annotation does not exist, return the default port.
+func determineAndValidatePort(pod *corev1.Pod, annotation string, defaultPort string) (string, error) {
+	if raw, ok := pod.Annotations[annotation]; ok && raw != "" {
+		port, err := portValue(pod, raw)
+		if err != nil {
+			return "", fmt.Errorf("%s annotation value of %s is not a valid integer.", annotation, raw)
+		}
+		// This checks if the port is in the valid port range.
+		if port < 1024 || port > 65535 {
+			return "", fmt.Errorf("%s annotation value of %d is not in the port range 1024-65353.", annotation, port)
```
thisisnotashwin, line 693:
```suggestion
			return "", fmt.Errorf("%s annotation value of %d is not in the port range 1024-65535.", annotation, port)
```

connect-inject/handler_test.go
```diff
@@ -586,6 +586,485 @@ func TestHandlerDefaultAnnotations(t *testing.T) {
...
+		},
+		{
+			Name: "Invalid annotation (integer not in port range)",
+			Pod: func(pod *corev1.Pod) *corev1.Pod {
+				pod.Annotations["consul.hashicorp.com/test-annotation-port"] = "100000"
+				return pod
+			},
+			Annotation: "consul.hashicorp.com/test-annotation-port",
+			Expected:   "",
+			Err:        "consul.hashicorp.com/test-annotation-port annotation value of 100000 is not in the port range 1024-65353.",
```
thisisnotashwin, line 1048:
```suggestion
			Err:        "consul.hashicorp.com/test-annotation-port annotation value of 100000 is not in the port range 1024-65535.",
```

subcommand/common/common.go
```diff
@@ -30,3 +32,17 @@ func Logger(level string) (hclog.Logger, error) {
...
+// ValidatePort converts flags representing ports into integer and validates
+// that it's in the port range.
+func ValidatePort(flagName, flagValue string) error {
+	port, err := strconv.Atoi(flagValue)
+	if err != nil {
+		return errors.New(fmt.Sprintf("%s value of %s is not a valid integer.", flagName, flagValue))
+	}
+	// This checks if the port is in the valid port range.
+	if port < 1024 || port > 65535 {
+		return errors.New(fmt.Sprintf("%s value of %d is not in the port range 1024-65353.", flagName, port))
```
thisisnotashwin, line 45:
```suggestion
		return errors.New(fmt.Sprintf("%s value of %d is not in the port range 1024-65535.", flagName, port))
```

subcommand/common/common_test.go
```diff
@@ -17,3 +17,12 @@ func TestLogger(t *testing.T) {
...
 	require.True(t, lgr.IsDebug())
 }
+
+func TestValidatePort(t *testing.T) {
+	err := ValidatePort("-test-flag-name", "1234")
+	require.NoError(t, err)
+	err = ValidatePort("-test-flag-name", "invalid-port")
+	require.EqualError(t, err, "-test-flag-name value of invalid-port is not a valid integer.")
+	err = ValidatePort("-test-flag-name", "22")
+	require.EqualError(t, err, "-test-flag-name value of 22 is not in the port range 1024-65353.")
```
thisisnotashwin, line 27:
```suggestion
	require.EqualError(t, err, "-test-flag-name value of 22 is not in the port range 1024-65535.")
```

thisisnotashwin (approved):
🥳 Excellent work Nitya!! have some suggestions that should update error message associated with 65353 -> 65535

subcommand/common/common.go
```diff
@@ -30,3 +32,17 @@ func Logger(level string) (hclog.Logger, error) {
...
+
+// ValidatePort converts flags representing ports into integer and validates
+// that it's in the port range.
+func ValidatePort(flagName, flagValue string) error {
+	port, err := strconv.Atoi(flagValue)
+	if err != nil {
+		return errors.New(fmt.Sprintf("%s value of %s is not a valid integer.", flagName, flagValue))
+	}
+	// This checks if the port is in the valid port range.
+	if port < 1024 || port > 65535 {
```
lkysow, line 44: Ports < 1024 are valid I think.

lkysow: Is this checking for unprivileged ports?

subcommand/consul-sidecar/command.go
```diff
@@ -80,57 +103,166 @@ func (c *Command) Run(args []string) int {
...
+	// for loop using shutdownMetricsServer().
+	var server *http.Server
+	if c.flagEnableMetricsMerging {
+		c.logger.Info("Metrics is enabled, creating merged metrics server.")
+		server = c.createMergedMetricsServer()
+
+		// Run the merged metrics server
+		c.logger.Info("Running merged metrics server.")
+		go func() {
+			if err = server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
```
lkysow, line 148: If the metrics merging server fails to start (e.g. it can't bind to its port) should the sidecar container continue to run? I would have thought we'd error out so the user is aware.

subcommand/common/common.go
ndhanushkodi: Yes that's accurate, sorry for the imprecise wording. It's checking for unprivileged ports and will return an error if you configure something in the unprivileged port range. I can make a change to the comment to clarify.

subcommand/consul-sidecar/command.go
ndhanushkodi: That's a good point. I suppose if a user misconfigures metrics with something like a conflicting port and there's only one log at the top of the consul-sidecar logs it will be fairly hidden to them that it's broken, so I think it makes sense to error out. I was reading [here](https://stackoverflow.com/questions/25920963/in-what-situation-will-http-listenandserve-return) that it seems common to `log.Fatal` here. `hclog` doesn't have a log.Fatal function but we can just `os.Exit(1)` after logging the error.

subcommand/common/common.go
lkysow: Is there a possible case where a user exposes their prom metrics on a privileged port because they're running their container as root? In that case we only want to ensure that the metrics merged port is unprivileged and valid and then we want to check that the scrape port is simply valid but it can be privileged or unprivileged.

subcommand/consul-sidecar/command.go
lkysow: This is exactly what happened to me. I mixed up the scrape port and merged port annotations and I set the merged port to be equal to the scrape port. Then the merged server failed to start because it tried to bind to a port that was already bound to (since that was the port my app was exposing its metrics on).

PR #441, opened 2021-02-19
Tproxy split init containers

kschoche:
Changes proposed in this PR:
- split the existing connect init container into two init containers:  1 does the consul binary copy and the other does the existing leg-work for registration. This is part of the tproxy refactor which sees the existing init container functionality replaced by a new consul-k8s command (incoming PR).

How I've tested this PR:
manually deployed and confirmed connect inject still works with the correct init containers.
new and existing unit tests work.

deploy kschoche/consul-k8s-dev and a sample connect injected app to confirm it still mutates and now has 2 containers instead of 1.

```
Events:
  Type     Reason     Age   From               Message
  ----     ------     ----  ----               -------
  Normal   Scheduled  22s   default-scheduler  Successfully assigned default/counting to [redacted]
  Normal   Pulled     21s   kubelet            Container image "hashicorp/consul:1.9.3" already present on machine
  Normal   Created    21s   kubelet            Created container consul-connect-inject-init-copy-container
  Normal   Started    21s   kubelet            Started container consul-connect-inject-init-copy-container
  Normal   Pulled     17s   kubelet            Container image "hashicorp/consul:1.9.3" already present on machine
  Normal   Created    17s   kubelet            Created container consul-connect-inject-init
  Normal   Started    16s   kubelet            Started container consul-connect-inject-init
  Normal   Pulled     12s   kubelet            Container image "hashicorp/counting-service:0.0.2" already present on machine
  Normal   Started    11s   kubelet            Started container counting
```

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -12,9 +12,10 @@ import (
 )
 
 const (
-	InjectInitContainerName = "consul-connect-inject-init"
-	MetaKeyPodName          = "pod-name"
-	MetaKeyKubeNS           = "k8s-namespace"
+	InjectInitCopyContainerName = "copy-consul-bin"
+	InjectInitContainerName     = "consul-connect-inject-init"
+	MetaKeyPodName              = "pod-name"
+	MetaKeyKubeNS               = "k8s-namespace"
 )
 
 type initContainerCommandData struct {
@@ -61,6 +62,25 @@ type initContainerCommandUpstreamData struct {
 	ConsulUpstreamNamespace string
 	Datacenter              string
 	Query                   string
+}
+
+// containerInitCopyContainer returns the init container spec for the copy container which places
+// the consul binary into the shared volume.
+func (h *Handler) containerInitCopyContainer() corev1.Container {
+	// Copy the Consul binary from the image to the shared volume.
+	cmd := "cp /bin/consul /consul/connect-inject/consul"
+	return corev1.Container{
+		Name:      InjectInitCopyContainerName,
+		Image:     h.ImageConsul,
+		Resources: h.InitContainerResources,
+		VolumeMounts: []corev1.VolumeMount{
+			corev1.VolumeMount{
+				Name:      volumeName,
+				MountPath: "/consul/connect-inject",
+			},
+		},
+		Command: []string{"/bin/sh", "-ec", cmd},
+	}
 }
 
 // containerInit returns the init container spec for registering the Consul
@@ -452,8 +472,4 @@ chmod 444 /consul/connect-inject/acl-token
   {{- if .ConsulNamespace }}
   -namespace="{{ .ConsulNamespace }}" \
   {{- end }}
-  -bootstrap > /consul/connect-inject/envoy-bootstrap.yaml
-
-# Copy the Consul binary
-cp /bin/consul /consul/connect-inject/consul
-`
+  -bootstrap > /consul/connect-inject/envoy-bootstrap.yaml`
--- a/connect-inject/handler_test.go
+++ b/connect-inject/handler_test.go
@@ -99,6 +99,10 @@ func TestHandlerHandle(t *testing.T) {
 				},
 				{
 					Operation: "add",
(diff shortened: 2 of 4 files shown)


subcommand/consul-sidecar/command.go
```diff
@@ -1,4 +1,4 @@
-package subcommand
+package consulsidecar
```
kschoche, line 1: Everything past this in the PR is a merge from feature-metrics which is the base of feature-tproxy.
I've rebased feature-tproxy off feature-metrics and pushed it but for some reason it's not showing up here.
Will try to sort it out before merging.

connect-inject/container_init.go
```diff
@@ -12,9 +12,10 @@ import (
 )
 
 const (
-	InjectInitContainerName = "consul-connect-inject-init"
-	MetaKeyPodName          = "pod-name"
-	MetaKeyKubeNS           = "k8s-namespace"
+	InjectInitCopyContainerName = "consul-connect-inject-init-copy-container"
```
kschoche, line 15: Thoughts on naming the new init container?

ndhanushkodi: My vote for container names would be for something like:
`consul-connect-copy-init`
`consul-connect-inject-init` and maybe even change the name of the consul-k8s command to `inject-init` rather than `consul-init`.

connect-inject/handler_test.go
```diff
@@ -97,6 +97,10 @@ func TestHandlerHandle(t *testing.T) {
 					Operation: "add",
 					Path:      "/spec/initContainers",
 				},
+				{
+					Operation: "add",
+					Path:      "/spec/initContainers/-",
```
ndhanushkodi, line 102: What's the difference between L102 and L98? Do you know what determines if you need a - or not? Can we use the - for both?

ndhanushkodi (approved):
Looks good! Had a thought on the name and was curious about the JSON patches.

connect-inject/handler_test.go
kschoche: The slash is for when there is more than one entry from the patch operation. So if there was a single entry it would just be `/spec/initContainer` and if there were more than one, the additional ones would have the `/-`.
Does that make sense?

ndhanushkodi: I see, thanks!!!

ishustava (commented):
Looking pretty good! I've left a couple of comments and chimed in on the naming conversation 😄 Once we decide on those, it's good to go. I've also left a couple of minor edits just to follow the godoc format.

I've run connect acceptance tests with an image from this branch, and they pass!

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+ENHANCEMENTS:
+* Connect: the init container for connect injected pods has been split into two containers.
+
```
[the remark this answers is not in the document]
kschoche:
```suggestion
```

connect-inject/container_init.go
kschoche, line 15:
```suggestion
	InjectInitCopyContainerName = "copy-consul-bin"
```

connect-inject/container_init.go
```diff
@@ -63,6 +64,26 @@ type initContainerCommandUpstreamData struct {
 	Query                   string
 }
 
+// containerInitCopyContainer returns the init container spec for the copy container which places
+// the consul binary into the shared volume.
+func (h *Handler) containerInitCopyContainer() (corev1.Container, error) {
```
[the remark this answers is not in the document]
kschoche: ah good catch!

ishustava (approved):
Looks good! (assuming we solve the diff before merging)

kschoche:
> Looks good! (assuming we solve the diff before merging)

Oddly enough if I pull both from origin and diff the two the metrics bits dont show up in `git diff`, I wonder if this is a strange quirk of github? both branches have commit `a756490c0bb3cd835a06ec580daa1adfb9cf7653` :-/

PR #443, opened 2021-02-23
Consul sidecar will exit if merged metrics server is unable to start.

ndhanushkodi:
Changes proposed in this PR:
- Exits if the merged metrics server can't come up
- Clarify comments/error messages about unprivileged ports

Addresses additional comments from https://github.com/hashicorp/consul-k8s/pull/438

How I've tested this PR:
Unit tests

Code review

### DIFF
--- a/connect-inject/handler.go
+++ b/connect-inject/handler.go
@@ -581,13 +581,13 @@ func (h *Handler) enableMetricsMerging(pod *corev1.Pod) (bool, error) {
 // mergedMetricsPort returns the default value in the handler, or overrides
 // that with the annotation if provided.
 func (h *Handler) mergedMetricsPort(pod *corev1.Pod) (string, error) {
-	return determineAndValidatePort(pod, annotationMergedMetricsPort, h.DefaultMergedMetricsPort)
+	return determineAndValidatePort(pod, annotationMergedMetricsPort, h.DefaultMergedMetricsPort, false)
 }
 
 // prometheusScrapePort returns the default value in the handler, or overrides
 // that with the annotation if provided.
 func (h *Handler) prometheusScrapePort(pod *corev1.Pod) (string, error) {
-	return determineAndValidatePort(pod, annotationPrometheusScrapePort, h.DefaultPrometheusScrapePort)
+	return determineAndValidatePort(pod, annotationPrometheusScrapePort, h.DefaultPrometheusScrapePort, false)
 }
 
 // prometheusScrapePath returns the default value in the handler, or overrides
@@ -600,21 +600,26 @@ func (h *Handler) prometheusScrapePath(pod *corev1.Pod) string {
 	return h.DefaultPrometheusScrapePath
 }
 
-// serviceMetricsPort returns the port used to register the service with Consul,
-// or overrides that with the annotation if provided.
+// serviceMetricsPort returns the port the service exposes metrics on. This will
+// default to the port used to register the service with Consul, and can be
+// overridden with the annotation if provided.
 func (h *Handler) serviceMetricsPort(pod *corev1.Pod) (string, error) {
 	// The annotationPort is the port used to register the service with Consul.
 	// If that has been set, it'll be used as the port for getting service
 	// metrics as well, unless overridden by the service-metrics-port annotation.
 	if raw, ok := pod.Annotations[annotationPort]; ok && raw != "" {
-		return determineAndValidatePort(pod, annotationServiceMetricsPort, raw)
+		// The service metrics port can be privileged if the service author has
+		// written their service in such a way that it expects to be able to use
+		// privileged ports. So, the port metrics are exposed on the service can
+		// be privileged.
+		return determineAndValidatePort(pod, annotationServiceMetricsPort, raw, true)
 	}
 
 	// If the annotationPort is not set, the serviceMetrics port will be 0
 	// unless overridden by the service-metrics-port annotation. If the service
 	// metrics port is 0, the consul sidecar will not run a merged metrics
 	// server.
-	return determineAndValidatePort(pod, annotationServiceMetricsPort, "0")
+	return determineAndValidatePort(pod, annotationServiceMetricsPort, "0", true)
 }
 
 // serviceMetricsPath returns a default of /metrics, or overrides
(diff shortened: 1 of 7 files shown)


subcommand/consul-sidecar/command.go
```diff
@@ -156,6 +156,7 @@ func (c *Command) Run(args []string) int {
 		go func() {
 			if err = server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
 				c.logger.Error(err.Error())
+				os.Exit(1)
```
ndhanushkodi, line 159: should this just be `return 1` instead?

ndhanushkodi: Actually, os.Exit(1) won't work in a goroutine. I need to somehow signal to Run to exit.

lkysow: See the command/inject-connect for how we do this. It uses a channel that the error gets sent to.

connect-inject/handler.go
```diff
@@ -688,9 +688,9 @@ func determineAndValidatePort(pod *corev1.Pod, annotation string, defaultPort st
 		if err != nil {
 			return "", fmt.Errorf("%s annotation value of %s is not a valid integer.", annotation, raw)
 		}
-		// This checks if the port is in the valid port range.
```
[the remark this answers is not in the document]
ndhanushkodi: It now takes a flag for privileged, and the service metrics port is allowed to be privileged, and merged metrics and prometheus scrape ports are unprivileged

connect-inject/handler.go
```diff
@@ -607,14 +607,14 @@ func (h *Handler) serviceMetricsPort(pod *corev1.Pod) (string, error) {
 	// If that has been set, it'll be used as the port for getting service
 	// metrics as well, unless overridden by the service-metrics-port annotation.
 	if raw, ok := pod.Annotations[annotationPort]; ok && raw != "" {
-		return determineAndValidatePort(pod, annotationServiceMetricsPort, raw)
+		return determineAndValidatePort(pod, annotationServiceMetricsPort, raw, true)
```
thisisnotashwin, line 610: Does it make sense to add a text snippet here describing why the serviceMetricsPort can use a privileged port but the scrape port and merged metrics port cannot? It might be tricky to understand the rationale behind doing so in the absence of the context of Luke's PR comment.

subcommand/consul-sidecar/command.go
```diff
@@ -165,34 +166,44 @@ func (c *Command) Run(args []string) int {
...
-		case <-time.After(c.flagSyncPeriod):
-			continue
-		case <-signalCtx.Done():
-			// After the signal is received, wait for the merged metrics server
-			// to gracefully shutdown as well if it has been enabled. This can
-			// take up to metricsServerShutdownTimeout seconds.
-			if c.flagEnableMetricsMerging {
-				c.logger.Info("Attempting to shut down metrics server.")
-				c.shutdownMetricsServer(server)
+	go func() {
```
thisisnotashwin, line 169: I like this cleanup a lot!

thisisnotashwin (commented):
Nicely done! It would be nice to elaborate why the service metrics port can be privileged unlike the other ports.

subcommand/consul-sidecar/command.go
```diff
@@ -165,34 +166,44 @@ func (c *Command) Run(args []string) int {
...
+			c.logger.Info("Attempting to shut down metrics server.")
+			c.shutdownMetricsServer(server)
+		}
+		return 0
+	case err := <-srvExitCh:
+		c.logger.Error(fmt.Sprintf("Metrics server error: %v", err))
+		return 1
 	}
+
 }
```
lkysow, line 207: below we discussed how `defer shutdownCancelFunc()` is not needed and how `server` cannot be `nil`.

subcommand/consul-sidecar/command_test.go
```diff
@@ -31,13 +31,13 @@ func TestRun_Defaults(t *testing.T) {
 }
 
 func TestRun_ExitsCleanlyonSignals(t *testing.T) {
```
lkysow, line 33: Is there a reason we have these test cases using this pattern of a top-level function that calls a private method rather than using the typical pattern of cases declared and then looping over those cases:

```go
func TestRun_SignalHandlingRegistration(t *testing.T) {
	cases := map[string]os.Signal{
		"SIGINT":  syscall.SIGINT,
		"SIGTERM": syscall.SIGTERM,
	}

	for name, signal := range cases {

```

lkysow (approved):
Looks great!

One little thing I noticed is that errors and log lines end with periods. For errors we typically avoid punctation because it looks weird when you wrap errors (`error 1.: error 2.`) and for the logs elsewhere we don't end with periods so we should follow that pattern.

connect-inject/handler.go
```diff
@@ -682,16 +682,28 @@ func (h *Handler) shouldRunMergedMetricsServer(pod *corev1.Pod) (bool, error) {
 // determineAndValidatePort behaves as follows:
```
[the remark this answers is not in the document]
ndhanushkodi: I tried to clarify that as:
```go
// serviceMetricsPort returns the port the service exposes metrics on. This will
// default to the port used to register the service with Consul, and can be
// overridden with the annotation if provided.
```

subcommand/consul-sidecar/command.go
ndhanushkodi: Ah sorry I had forgotten that. Added with a comment.

connect-inject/handler.go
ndhanushkodi: Totally! Added this.

connect-inject/handler.go
```diff
@@ -682,16 +682,28 @@ func (h *Handler) shouldRunMergedMetricsServer(pod *corev1.Pod) (bool, error) {
...
 // If the annotation does not exist, return the default port.
-func determineAndValidatePort(pod *corev1.Pod, annotation string, defaultPort string) (string, error) {
+// If the privileged flag is true, it will allow the port to be in the
+// privileged port range of 1-1023. Otherwise, it will only allow ports in the
+// unprivileged range of 1024-65535.
+func determineAndValidatePort(pod *corev1.Pod, annotation string, defaultPort string, privileged bool) (string, error) {
 	if raw, ok := pod.Annotations[annotation]; ok && raw != "" {
 		port, err := portValue(pod, raw)
 		if err != nil {
 			return "", fmt.Errorf("%s annotation value of %s is not a valid integer.", annotation, raw)
```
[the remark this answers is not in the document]
ndhanushkodi: Took a look through for errors and logs with periods and removed them. Thanks for pointing this out!

subcommand/consul-sidecar/command_test.go
ndhanushkodi: I think we should use the typical pattern, I went ahead and switched that. Thanks!

connect-inject/handler.go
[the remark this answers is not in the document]
lkysow: 👍

subcommand/consul-sidecar/command.go
ndhanushkodi: Actually, the go vet check seems to be failing due to not calling the cancel func. I think we should still `defer shutdownCancelFunc()` and leave the comment about why its not necessary but required for go vet.

connect-inject/handler.go
```diff
@@ -682,16 +682,28 @@ func (h *Handler) shouldRunMergedMetricsServer(pod *corev1.Pod) (bool, error) {
...
+			if port < 1 || port > 65535 {
+				return "", fmt.Errorf("%s annotation value of %d is not in the valid port range 1-65535.", annotation, port)
+			}
+		} else {
+			// This checks if the port is in the unprivileged port range.
+			if port < 1024 || port > 65535 {
+				return "", fmt.Errorf("%s annotation value of %d is not in the unprivileged port range 1024-65535.", annotation, port)
+			}
 		}
+
```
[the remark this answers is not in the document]
ndhanushkodi: if `privileged` is true, and you input a valid privileged port it falls into the "else if" condition and returns the unprivileged error message. the "else if" condition needs a `!privileged`. Only noticed this when the test failed :)

thisisnotashwin (approved):
Brilliant!

PR #446, opened 2021-02-24
Add consul login api

kschoche:
Changes proposed in this PR:
- Create a new API for the `consul login` command which uses only Consul APIs and removes a dependency on the Consul binary.
- Currently just putting this out there to see how people feel about the testing, its a bit complicated because we would need to bootstrap auth methods to fully duplicate the tests that `hashicorp/consul` already does on the same code.
- I've made one tweak to the code that the consul command runs which is to use ioutil to write the file instead of their own Sink library which seems to only additionaly do a fsync() on the file after writing. In our case this is not necessary as fsync is for disk persistence in the event of a crash and we're writing to an emptyDir volume which is destroyed on pod restart.

How I've tested this PR:
Unit tests so far.
Further integration is a little bit blocked on an incoming PR which adds the `consul-k8s consul-init` command.

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/common/common.go
+++ b/subcommand/common/common.go
@@ -4,9 +4,12 @@ package common
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strconv"
+	"strings"
 
+	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
 )
 
@@ -43,6 +46,35 @@ func ValidatePort(flagName, flagValue string) error {
 	// This checks if the port is in the valid port range.
 	if port < 1024 || port > 65535 {
 		return errors.New(fmt.Sprintf("%s value of %d is not in the port range 1024-65535.", flagName, port))
+	}
+	return nil
+}
+
+// ConsulAclLogin issues an ACL().Login to Consul and writes out the token to tokenSinkFile.
+func ConsulAclLogin(client *api.Client, bearerTokenFile, authMethodName, bearerToken, tokenSinkFile string, meta map[string]string) error {
+	data, err := ioutil.ReadFile(bearerTokenFile)
+	if err != nil {
+		return fmt.Errorf("unable to read bearerTokenFile: %v, err: %v", bearerTokenFile, err)
+	}
+	bearerToken = strings.TrimSpace(string(data))
+	if bearerToken == "" {
+		return fmt.Errorf("no bearer token found in %s", bearerTokenFile)
+	}
+	// Do the login.
+	req := &api.ACLLoginParams{
+		AuthMethod:  authMethodName,
+		BearerToken: bearerToken,
+		Meta:        meta,
+	}
+	tok, _, err := client.ACL().Login(req, nil)
+	if err != nil {
+		return fmt.Errorf("error logging in: %s", err)
+	}
+
+	// Write the token out to file.
+	payload := []byte(tok.SecretID)
+	if err := ioutil.WriteFile(tokenSinkFile, payload, 444); err != nil {
+		return fmt.Errorf("error writing token to file sink: %v", err)
 	}
 	return nil
 }
--- a/subcommand/common/common_test.go
+++ b/subcommand/common/common_test.go
@@ -1,8 +1,13 @@
 package common
 
 import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"testing"
 
+	"github.com/hashicorp/consul/api"
 	"github.com/stretchr/testify/require"
 )
 
@@ -25,4 +30,56 @@ func TestValidatePort(t *testing.T) {
 	require.EqualError(t, err, "-test-flag-name value of invalid-port is not a valid integer.")
 	err = ValidatePort("-test-flag-name", "22")
 	require.EqualError(t, err, "-test-flag-name value of 22 is not in the port range 1024-65535.")
+}
+
+// TestConsulAclLogin ensures that our implementation of consul login hits `/v1/acl/login`.
+func TestConsulAclLogin(t *testing.T) {
+	t.Parallel()
+	require := require.New(t)
+
+	type APICall struct {
+		Method string
+		Path   string
+	}
+	var consulAPICalls []APICall
+
+	// Stub the read of the bearerTokenFile inside the login function.
+	bearerTokenFile, err := ioutil.TempFile("", "bearerTokenFile")
+	require.NoError(err)
+	_, err = bearerTokenFile.WriteString("foo")
+	require.NoError(err)
+
+	// Start the Consul server.
+	consulServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Record all the API calls made.
+		consulAPICalls = append(consulAPICalls, APICall{
+			Method: r.Method,
+			Path:   r.URL.Path,
+		})
+	}))
+	defer consulServer.Close()
+	serverURL, err := url.Parse(consulServer.URL)
+	require.NoError(err)
+	clientConfig := &api.Config{Address: serverURL.String()}
+	client, err := api.NewClient(clientConfig)
+	require.NoError(err)
+
+	// We expect this to fail in the internal client.ACL().Login() path because we've not
+	// bootstrapped any auth methods.
+	err = ConsulAclLogin(
+		client,
+		bearerTokenFile.Name(),
+		"foo",
+		"/foo",
+		"/foo",
+		nil,
+	)
+	require.Error(err, "error logging in: EOF")
+	// Ensure that the /v1/acl/login url was correctly hit.
+	require.Equal([]APICall{
+		{
+			"POST",
+			"/v1/acl/login",
+		},
+	}, consulAPICalls)
 }


connect-inject/container_init.go
```diff
@@ -323,6 +323,21 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
...
+         defined in the default namespace */}}
+  -namespace="default" \
+  {{- else }}
+  -namespace="{{ .ConsulNamespace }}" \
+  {{- end }}
+  {{- end }}
+  -meta="pod=${POD_NAMESPACE}/${POD_NAME}"
+{{- /* The acl token file needs to be read by the consul-sidecar which runs
+       as non-root user consul-k8s. TODO: will this be necessary anymore? */}}
+{{- end }}
```
kschoche, line 340: When the endpoints controller is ready we will remove everything past this in the command tmpl

connect-inject/container_init.go
```diff
@@ -458,7 +453,7 @@ chmod 444 /consul/connect-inject/acl-token
   /consul/connect-inject/service.hcl
 
 # Generate the envoy bootstrap code
-/bin/consul connect envoy \
+/consul/connect-inject/consul connect envoy \
```
kschoche, line 456: Since we are now using consul-k8s image instead of consul the path will be from the shared vol instead of `/bin/`.. update all the tests accordingly.

subcommand/common/common.go
```diff
@@ -46,3 +51,48 @@ func ValidatePort(flagName, flagValue string) error {
...
+	payload := []byte(tok.SecretID)
+	if err := ioutil.WriteFile(tokenSinkFile, payload, 444); err != nil {
+		return fmt.Errorf("error writing token to file sink: %v", err)
+	}
+	return nil
+}
+
+// WriteTempFile writes contents to a temporary file and returns the file
+// name. It will remove the file once the test completes.
+func WriteTempFile(t *testing.T, contents string) string {
```
kschoche, line 86: This was moved from server-acl-init so we can re-use it.

subcommand/common/common_test.go
```diff
@@ -26,3 +31,59 @@ func TestValidatePort(t *testing.T) {
 	err = ValidatePort("-test-flag-name", "22")
 	require.EqualError(t, err, "-test-flag-name value of 22 is not in the port range 1024-65535.")
 }
+
+// TestConsulAclLogin ensures that our implementation of consul login hits `/v1/acl/login`.
+func TestConsulAclLogin(t *testing.T) {
```
kschoche, line 36:
```suggestion
// TestConsulLogin ensures that our implementation of consul login hits `/v1/acl/login`.
func TestConsulLogin(t *testing.T) {
```

subcommand/common/common_test.go
```diff
@@ -26,3 +31,59 @@ func TestValidatePort(t *testing.T) {
...
+	require.NoError(err)
+
+	// Start the Consul server.
+	consulServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Record all the API calls made.
+		consulAPICalls = append(consulAPICalls, APICall{
+			Method: r.Method,
+			Path:   r.URL.Path,
+		})
+		b := "{\n  \"AccessorID\": \"926e2bd2-b344-d91b-0c83-ae89f372cd9b\",\n  \"SecretID\": \"b78d37c7-0ca7-5f4d-99ee-6d9975ce4586\",\n  \"Description\": \"token created via login\",\n  \"Roles\": [\n    {\n      \"ID\": \"3356c67c-5535-403a-ad79-c1d5f9df8fc7\",\n      \"Name\": \"demo\"\n    }\n  ],\n  \"ServiceIdentities\": [\n    {\n      \"ServiceName\": \"example\"\n    }\n  ],\n  \"Local\": true,\n  \"AuthMethod\": \"minikube\",\n  \"CreateTime\": \"2019-04-29T10:08:08.404370762-05:00\",\n  \"Hash\": \"nLimyD+7l6miiHEBmN/tvCelAmE/SbIXxcnTzG3pbGY=\",\n  \"CreateIndex\": 36,\n  \"ModifyIndex\": 36\n}"
```
kschoche, line 61: This is a sample API output from consul.io that we are mocking.

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,185 @@
...
+		"Metadata to be passed to consul login api.")
+
+	c.flagSet.StringVar(&c.flagBearerTokenFile, "bearer-token-file", bearerTokenFile,
+		"Token file to be used for consul login api.")
+	c.flagSet.StringVar(&c.flagTokenSinkFile, "token-sink-file", tokenSinkFile,
+		"Token sink file to be used for consul login api.")
+	//c.flagSet.StringVar(&c.flagBearerTokenFile, "bearer-token-file", "",
+	//	"Token file to be used for consul login api.")
+	//c.flagSet.StringVar(&c.flagTokenSinkFile, "token-sink-file", "",
+	//	"Token sink file to be used for consul login api.")
```
kschoche, line 93:
```suggestion
```

subcommand/common/common.go
kschoche: Note to reviewers: 
I'm moving this back and just having it be duplicated in `server-acl-init/command_test.go` and `consul-init/command_test.go` because putting it here brings in `testing` package deps to the common package.
This has the added

connect-inject/cleanup_resource_ent_test.go
```diff
@@ -124,7 +125,8 @@ func TestReconcile_ConsulNamespaces(t *testing.T) {
 
 			// Run Reconcile.
 			cleanupResource.reconcile()
-
+			// FIXME: timing issue here popping up also in main branch
+			time.Sleep(time.Second * 1)
```
kschoche, line 129: Temp fix until this one gets merged.
https://github.com/hashicorp/consul-k8s/pull/449/files

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+		c.consulClient, err = consul.NewClient(cfg)
+		if err != nil {
+			c.UI.Error("unable to get client connection")
+			return 1
+		}
+	}
+	ErrCh := make(chan error)
+	ExitCh := make(chan bool)
+	meta := map[string]string{"pod": strings.Split(c.flagMeta, "=")[1]}
+	go func() {
```
lkysow, line 109: This loop ignore any sigterms right now. I'm also not sure it needs to run in a go routine? Would something like this work:

```go
	retries := 0
	for {
		err = common.ConsulLogin(c.consulClient, c.flagBearerTokenFile, c.flagACLAuthMethod, c.flagTokenSinkFile, meta)
		if err == nil {
			break
		}
		retries++
		if retries == c.numACLLoginRetries {
			c.UI.Error("hit maximum retries for consul login")
			return 1
		}
		c.UI.Error(fmt.Sprintf("consul login failed; retrying: %s", err))

		select {
		case <-time.After(1 * time.Second):
			// retry loop
		case sig := <-c.sigCh:
			c.UI.Info(fmt.Sprintf("%s received, shutting down", sig))
			return 0
		}
	}
```

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+		retries := 0
+		// This is hardcoded because we hardcode it in the init container.
+		for {
+			err = common.ConsulLogin(c.consulClient, c.flagBearerTokenFile, c.flagACLAuthMethod, c.flagTokenSinkFile, meta)
+			if err != nil {
+				retries++
+				time.Sleep(time.Second * 1)
```
lkysow, line 116: can we make this a const

subcommand/consul-init/command_test.go
```diff
@@ -0,0 +1,201 @@
...
+func min(x, y int) int {
+	if x > y {
+		return y
+	}
+	return x
+}
+
+// writeTempFile writes contents to a temporary file and returns the file
+// name. It will remove the file once the test completes.
+func writeTempFile(t *testing.T, contents string) string {
```
lkysow, line 145: rather than copying this can we put it in `common/test_util`?

subcommand/consul-init/command_test.go
```diff
@@ -0,0 +1,201 @@
...
+			bearerTokenFile := writeTempFile(t, "bearerTokenFile")
+			tokenFile := writeTempFile(t, "")
+
+			// Start the mock Consul server.
+			counter := 0
+			consulServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r != nil && r.URL.Path == "/v1/acl/login" && r.Method == "POST" {
+					counter++
+					// sample response from https://consul.io/api-docs/acl#sample-response
+					b := "{\n  \"AccessorID\": \"926e2bd2-b344-d91b-0c83-ae89f372cd9b\",\n  \"SecretID\": \"b78d37c7-0ca7-5f4d-99ee-6d9975ce4586\",\n  \"Description\": \"token created via login\",\n  \"Roles\": [\n    {\n      \"ID\": \"3356c67c-5535-403a-ad79-c1d5f9df8fc7\",\n      \"Name\": \"demo\"\n    }\n  ],\n  \"ServiceIdentities\": [\n    {\n      \"ServiceName\": \"example\"\n    }\n  ],\n  \"Local\": true,\n  \"AuthMethod\": \"minikube\",\n  \"CreateTime\": \"2019-04-29T10:08:08.404370762-05:00\",\n  \"Hash\": \"nLimyD+7l6miiHEBmN/tvCelAmE/SbIXxcnTzG3pbGY=\",\n  \"CreateIndex\": 36,\n  \"ModifyIndex\": 36\n}"
```
lkysow, line 99: see my comment on other test

subcommand/consul-init/command_test.go
```diff
@@ -0,0 +1,201 @@
...
+					if !c.TestRetry || (c.TestRetry && c.LoginAttemptsCount == counter) {
```
lkysow, line 100: nice! Love this method

lkysow (commented):
This is looking good! I think the biggest thing I'm wondering is that right now this command is only run when ACLs are enabled but the plan is for this command to eventually run when ACLs are disabled so does it make sense to write it right now so it essentially no-ops if ACLs are disabled?

Because otherwise we're writing tests that `-auth-method` is a required flag but then later it will be run without that flag (if ACLs are disabled).

subcommand/consul-init/command_test.go
```diff
@@ -0,0 +1,201 @@
...
+	_, err = file.WriteString(contents)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		os.Remove(file.Name())
+	})
+	return file.Name()
+}
+
+func TestSignalHandling(t *testing.T) {
```
lkysow, line 158: can we move this up so it's grouped with the other test methods and above any private test helpers

connect-inject/container_init.go
```diff
@@ -323,6 +323,21 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
...
+  {{- /* If namespace mirroring is enabled, the auth method is
+         defined in the default namespace */}}
+  -namespace="default" \
+  {{- else }}
+  -namespace="{{ .ConsulNamespace }}" \
+  {{- end }}
+  {{- end }}
+  -meta="pod=${POD_NAMESPACE}/${POD_NAME}"
+{{- /* The acl token file needs to be read by the consul-sidecar which runs
+       as non-root user consul-k8s. TODO: will this be necessary anymore? */}}
```
ndhanushkodi, line 339: Reminder to remove this comment, I saw Luke say in his review that you would have the consul-init command itself write the file with 444 permissions.

commands.go
```diff
@@ -31,6 +32,10 @@ func init() {
 			return &cmdACLInit.Command{UI: ui}, nil
 		},
 
+		"consul-init": func() (cli.Command, error) {
```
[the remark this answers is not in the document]
kschoche: Yeah I think connect-init sounds better!  %s/ time.

subcommand/common/common.go
```diff
@@ -46,3 +49,33 @@ func ValidatePort(flagName, flagValue string) error {
 	}
 	return nil
 }
+
+// ConsulLogin issues an ACL().Login to Consul and writes out the token to tokenSinkFile.
+// The logic of this is taken from the `consul login` command.
+func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkFile string, meta map[string]string) error {
```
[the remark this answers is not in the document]
kschoche: Yeah, I think the plan was to potentially use it elsewhere anytime we are fetching ACLs.

subcommand/common/common_test.go
```diff
@@ -26,3 +31,59 @@ func TestValidatePort(t *testing.T) {
...
```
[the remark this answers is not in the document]
kschoche: Great idea!

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/hashicorp/consul-k8s/consul"
+	"github.com/hashicorp/consul-k8s/subcommand/common"
+	"github.com/hashicorp/consul-k8s/subcommand/flags"
+	"github.com/hashicorp/consul/api"
+	"github.com/mitchellh/cli"
+)
+
+const bearerTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+const tokenSinkFile = "/consul/connect-inject/acl-token"
```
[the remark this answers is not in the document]
kschoche: Let me think about this, I think it's a good idea, it is a const in the init_container.go, but we could do that here as well.

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+
+type Command struct {
+	UI cli.Ui
+
+	flagACLAuthMethod   string // Auth Method to use for ACLs, if enabled
+	flagMeta            string // Flag for metadata to consul login.
```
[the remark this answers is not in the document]
kschoche: So this field is used exclusively by `consul login` and is hard-coded by the init_container.
I think I can probably rename it to "-consul-login-meta" or honestly remove it altogether because at runtime we will actually know how to build up the string "pod=ns/podname".
Do we expect to be able to pass in additional metadata to `consul login` or do you envision us having other metadata to be passed into this command? Right now it does not support it, but I do plan to expand the command with additional flags so that it can do the rest of the init code!
Should I remove / rename this field?

subcommand/common/common_test.go
ndhanushkodi: This is a nifty way to stub out the server and assert on the api call! I love what you've done here.

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+	flagBearerTokenFile string // Location of the bearer token.
+	flagTokenSinkFile   string // Location to write the output token.
+	numACLLoginRetries  int    // Number of times to attempt to retry ACL().Login(), default: 3.
```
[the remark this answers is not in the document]
kschoche: Great question. I thought about controlling it via a flag, but just made it a var for simplicity sake.
I could add that, but I'm not sure there is a ton of value in making it configurable?? Definitely open to ideas.

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+	sigCh chan os.Signal
+	once  sync.Once
+	help  string
+	cert  atomic.Value
+}
+
+func (c *Command) init() {
+	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
+	c.flagSet.StringVar(&c.flagACLAuthMethod, "method", "",
+		"The name of the Kubernetes Auth Method to use for connectInjection if ACLs are enabled.")
```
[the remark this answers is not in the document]
kschoche: ah great catch, I completely forgot to look at this whole section when I was cleaning up!

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+	c.flagSet.StringVar(&c.flagMeta, "meta", "",
+		"Metadata to be passed to consul login api.")
+	c.flagSet.StringVar(&c.flagBearerTokenFile, "bearer-token-file", bearerTokenFile,
+		"Token file to be used for consul login api.")
```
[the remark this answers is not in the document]
kschoche: Yes!

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+	c.flagSet.StringVar(&c.flagTokenSinkFile, "token-sink-file", tokenSinkFile,
+		"Token sink file to be used for consul login api.")
+
+	c.http = &flags.HTTPFlags{}
+
+	flags.Merge(c.flagSet, c.http.Flags())
+	c.help = flags.Usage(help, c.flagSet)
+
+	// Default number of times to attempt ACL().Login()
+	c.numACLLoginRetries = 3
```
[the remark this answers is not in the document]
kschoche: This is intentional, and in the test it was just to help visualize what was going on.
Per the comment above about it being a flag or not we can make a decision on this!

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := c.flagSet.Parse(args); err != nil {
+		return 1
+	}
+
+	// Validate flags.
+	if c.flagACLAuthMethod == "" {
+		c.UI.Error("-method must be set")
+		return 1
+	}
+	if c.flagMeta == "" {
```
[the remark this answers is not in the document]
kschoche: So this command is not generally consumable by users, it won't do anything meaningful standalone, meta will always be set outside of tests in the current incarnation of the code. However..  I could wrap these 4 validation checks inside of some sort if `if c.flagACLMethod {       }` ?
I don't think its a massive overhaul of the code or tests to leave these as is for current PR since it ONLY functions as a means to fetch ACLs.
Thoughts ?

subcommand/consul-init/command.go
kschoche: Love it!

subcommand/consul-init/command_test.go
kschoche: Yes absolutely this is where it should go! I initially put it in common and then backed that out bc it brought in testing deps to common... thanks!

connect-inject/container_init.go
```diff
@@ -323,6 +323,21 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
 // and the connect-proxy service should come after the "main" service
 // because its alias health check depends on the main service to exist.
 const initContainerCommandTpl = `
+{{- if .AuthMethod }}
```
ndhanushkodi, line 326: Is this just the command without the loop that waits for the consul client to have a service instance registered with the local agent? (i.e is this only don't the consul login for now)?

ndhanushkodi (approved):
Just a clarification on the scope of this PR, the consul login logic and tests and command look great!!

connect-inject/cleanup_resource_ent_test.go
```diff
@@ -199,7 +199,7 @@ func TestDelete_ConsulNamespaces(t *testing.T) {
 			server, err := testutil.NewTestServerConfigT(t, nil)
 			defer server.Stop()
 			require.NoError(err)
-			server.WaitForLeader(t)
+			server.WaitForSerfCheck(t)
```
kschoche, line 202: will rebase this at the end.

connect-inject/container_init.go
ndhanushkodi: oh I think you can ignore this, I think that is the case, and for right now the command is mostly doing the consul login.

kschoche: Yeah it is only doing consul login right now. trying to strike a line between this PR and the next PR which will bring in non-ACL support, etc.

connect-inject/container_init.go
kschoche: Just did! I'm going through comments now and will push an update in ~30m

subcommand/consul-init/command.go
ishustava: What do you think about using `backoff.Retry` (example [here](https://github.com/hashicorp/consul-k8s/blob/125f7e37bc2ab16a7ed432849fbbfe13ff58e802/subcommand/get-consul-client-ca/command.go#L101-L115)). I think this package will already handle a lot the signals and make the logic much cleaner. You'd need to pass in `backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), 3)` as the second argument so that it stops after 3 tries.

subcommand/consul-init/command.go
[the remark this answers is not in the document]
ndhanushkodi: If we do ever need to add additional metadata to consul login in the future it'll probably be nice to have the []string. I think it'd be ok for us to assume the order of the flags and things like that since it'll always be passed programmatically and consistently for all connect-injected pods.

subcommand/consul-init/command.go
```diff
@@ -0,0 +1,162 @@
...
+		c.http.MergeOntoConfig(cfg)
+		c.consulClient, err = consul.NewClient(cfg)
+		if err != nil {
+			c.UI.Error("unable to get client connection")
+			return 1
+		}
+	}
+	ErrCh := make(chan error)
+	ExitCh := make(chan bool)
+	meta := map[string]string{"pod": strings.Split(c.flagMeta, "=")[1]}
```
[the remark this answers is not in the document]
kschoche: Ok you've convinced me on this, and actually `consul login` also supports multiple `meta` params, so I'll add support for it!

subcommand/consul-init/command.go
kschoche: @ishustava great idea, I like it and its much cleaner looking I'm not sure I understand how it handles signals. In `get-consul-client-ca` I don't see it either and we're not testing that command for any signal handling. Could you can show me? 🙇🏽

ishustava: What is the reason to have signal handling in this command?

In get-consul-client-ca, we don't handle interrupt signals because it doesn't start any goroutines, and so if an interrupt or a term signal is sent, it doesn't need to drain or stop any other processes that it's running. We also don't do it in other similar commands, e.g server-acl-init, probably for the same reason.

lkysow: We do loop for the login so wouldn't it be nice if you could delete the pod and it would exit quickly if it was in its `init`?

lkysow: I'm fine with not having it too if it's deemed too complicated.

ishustava: >We do loop for the login so wouldn't it be nice if you could delete the pod and it would exit quickly if it was in its init

If you delete a pod, then it should still exit quickly. If you have a program that doesn't handle interrupts and you send it an interrupt, it should still exit as soon as you send that signal because it's handled by golang and OS. In this command, we don't need any graceful shutdown behavior so I think it's ok to skip OS signal handling. Maybe I'm misunderstanding something?

lkysow: Ahh I didn't know that there was default signal handling that gets turned off by `signal.Notify` (https://golang.org/pkg/os/signal/).

connect-inject/container_init.go
```diff
@@ -323,6 +323,19 @@ func (h *Handler) containerInit(pod *corev1.Pod, k8sNamespace string) (corev1.Co
...
+         defined in the default namespace */}}
+  -namespace="default" \
+  {{- else }}
+  -namespace="{{ .ConsulNamespace }}" \
+  {{- end }}
+  {{- end }}
+  -meta="pod=${POD_NAMESPACE}/${POD_NAME}"
+{{- end }}
 {{- if .ConsulCACert}}
 export CONSUL_HTTP_ADDR="https://${HOST_IP}:8501"
```
lkysow, line 340: exports need to be first

lkysow (commented):
I tested everything out and it's looking great. Just some small suggested changes remaining.

connect-inject/container_init.go
kschoche: great catch thx! got mixed up during a refactor from the PR

subcommand/connect-init/command_test.go
```diff
@@ -0,0 +1,181 @@
...
+		return y
+	}
+	return x
+}
+
+// This function starts the command asynchronously and returns a non-blocking chan.
+// When finished, the command will send its exit code to the channel.
+// Note that it's the responsibility of the caller to terminate the command by calling stopCommand,
+// otherwise it can run forever.
+func runCommandAsynchronously(cmd *Command, args []string) chan int {
```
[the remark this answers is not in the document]
kschoche: yup! forgot to push this back up :D

subcommand/common/common.go
```diff
@@ -46,3 +49,36 @@ func ValidatePort(flagName, flagValue string) error {
...
+	if meta == nil {
+		return fmt.Errorf("invalid meta")
```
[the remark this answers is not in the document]
kschoche: Its an invalid input, we require it to be passed to consul login.

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,147 @@
...
+	// Default number of times to attempt ACL().Login()
+	c.numLoginRetries = 3
+
+	// Wait on an interrupt or terminate for exit, be sure to init it before running
+	// the controller so that we don't receive an interrupt before it's ready.
+	if c.sigCh == nil {
+		c.sigCh = make(chan os.Signal, 1)
+		signal.Notify(c.sigCh, syscall.SIGINT, syscall.SIGTERM)
+	}
+}
```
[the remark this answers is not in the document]
kschoche: rip all this out then..

subcommand/common/common_test.go
```diff
@@ -26,3 +59,97 @@ func TestValidatePort(t *testing.T) {
...
+	consulServer, client, bearerTokenFile, tokenFile := startMockServer(t, "", &counter)
+	defer consulServer.Close()
+	err := ConsulLogin(
+		client,
+		bearerTokenFile,
+		testAuthMethod,
+		tokenFile,
+		testPodMeta,
+	)
+	require.Error(err, "unable to read bearerTokenFile")
```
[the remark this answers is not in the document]
kschoche: great catch!

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,147 @@
...
+	c.flagSet.StringVar(&c.flagTokenSinkFile, "token-sink-file", tokenSinkFile,
+		"The most recent token's SecretID is kept up to date in this file.")
+
+	c.http = &flags.HTTPFlags{}
+
+	flags.Merge(c.flagSet, c.http.Flags())
+	c.help = flags.Usage(help, c.flagSet)
+
+	// Default number of times to attempt ACL().Login()
+	c.numLoginRetries = 3
```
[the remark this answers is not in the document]
kschoche: Yup makes sense, have made it a constant.

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,147 @@
...
+	return 0
+}
+
+func (c *Command) interrupt() {
+	c.sendSignal(syscall.SIGINT)
+}
+
+func (c *Command) sendSignal(sig os.Signal) {
+	c.sigCh <- sig
+}
```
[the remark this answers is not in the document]
kschoche: great catch thx

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Validate flags.
+	if c.flagACLAuthMethod == "" {
+		c.UI.Error("-method must be set")
+		return 1
+	}
+	if c.flagMeta == nil {
+		c.UI.Error("-meta must be set")
+		return 1
+	}
```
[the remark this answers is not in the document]
lkysow: clarified that we want to make it required because we used it before

subcommand/connect-init/command_test.go
```diff
@@ -0,0 +1,181 @@
...
+	}{
+		{
+			flags:  []string{"-method", testAuthMethod},
+			expErr: "-meta must be set",
+		},
+		{
+			flags:  []string{"-meta", "pod=abcdefg"},
+			expErr: "-method must be set",
+		},
+	}
```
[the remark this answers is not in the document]
kschoche: I have them taking in defaults right now in the flag parsing, wasn't sure how to handle it, what do you think?

subcommand/common/common_test.go
```diff
@@ -26,3 +33,128 @@ func TestValidatePort(t *testing.T) {
...
+	data, err := ioutil.ReadFile(tokenFile)
+	require.NoError(err)
+	require.Equal(string(data), "b78d37c7-0ca7-5f4d-99ee-6d9975ce4586")
+}
+
+func TestConsulLogin_EmptyBearerTokenFile(t *testing.T) {
+	t.Parallel()
+	require := require.New(t)
+	counter := 0
+	client, bearerTokenFile, tokenFile := startMockServer(t, "", &counter)
```
ishustava, line 63: why do we need the mock server in this test? I think this error happens before any calls to consul are made.

subcommand/connect-init/command_test.go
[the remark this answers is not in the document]
lkysow: I'd say follow the other commands we have and how they deal with flag defaults. If there is no prior art then I'd say if a flag has a default then we don't need to check if it's `nil` and then we also wouldn't write a test for it.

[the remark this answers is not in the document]
kschoche: Yeah that is what I ended up doing here. but let me look at some other tests and update the PR accordingly.

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,125 @@
...
+		if err == nil {
+			return nil
+		}
+		retries++
+		if retries == numLoginRetries {
+			return err
+		}
+		c.UI.Error(fmt.Sprintf("consul login failed; retrying: %s", err))
+		return err
+	}, backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), numLoginRetries))
```
ishustava, line 104: I think this can be just:
```suggestion
	err = backoff.Retry(func() error {
		err := common.ConsulLogin(c.consulClient, c.flagBearerTokenFile, c.flagACLAuthMethod, c.flagTokenSinkFile, c.flagMeta)
		if err != nil {
		  c.UI.Error(fmt.Sprintf("Consul login failed; retrying: %s", err))
		}
		return err
	}, backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), numLoginRetries))
```

subcommand/connect-init/command_test.go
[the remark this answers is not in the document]
lkysow: Oh interesting. I would have said that if:
```go
	if c.flagBearerTokenFile == "" {
		c.UI.Error("-bearer-token-file must be set")
		return 1
	}
	if c.flagTokenSinkFile == "" {
		c.UI.Error("-token-sink-file must be set")
		return 1
	}
```

Will never run because those have defaults then we can just delete those lines and then no need to test.

ishustava (commented):
Looking pretty good! I really like the work you did here! I left some comments; they are mostly minor suggestions or questions. Otherwise, it looks good to me!

subcommand/common/common_test.go
kschoche: Yup, you're right, I was just using the same mock setup for each of the tests, I can be a bit more explicit with the code.

subcommand/common/common_test.go
```diff
@@ -26,3 +59,97 @@ func TestValidatePort(t *testing.T) {
...
+	)
+	require.Error(err, "unable to read bearerTokenFile")
+}
+
+func TestConsulLogin_NoBearerTokenFile(t *testing.T) {
+	t.Parallel()
+	require := require.New(t)
+	counter := 0
+	consulServer, client, _, tokenFile := startMockServer(t, "foo", &counter)
+	randFileName := fmt.Sprintf("/foo/%d/%d", rand.Int(), rand.Int())
```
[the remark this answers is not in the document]
kschoche: I picked random file name because I wanted to prevent any chance of accidentally finding "foo" as a valid file on a test. I guess its not likely though.

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,147 @@
...
+	consulClient *api.Client
+
+	sigCh chan os.Signal
+	once  sync.Once
+	help  string
+}
+
+func (c *Command) init() {
+	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
+	c.flagSet.StringVar(&c.flagACLAuthMethod, "method", "",
```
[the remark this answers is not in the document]
kschoche: absolutely, great catch!

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Default number of times to attempt ACL().Login()
+	c.numLoginRetries = 3
+
+	// Wait on an interrupt or terminate for exit, be sure to init it before running
+	// the controller so that we don't receive an interrupt before it's ready.
+	if c.sigCh == nil {
+		c.sigCh = make(chan os.Signal, 1)
+		signal.Notify(c.sigCh, syscall.SIGINT, syscall.SIGTERM)
+	}
```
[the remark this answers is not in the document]
ishustava: sorry! I removed this comment, but github still remembered it and posted it!

subcommand/common/common_test.go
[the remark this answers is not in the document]
ishustava: sounds good!

subcommand/common/common_test.go
```diff
@@ -26,3 +59,97 @@ func TestValidatePort(t *testing.T) {
...
+		testAuthMethod,
+		randFileName,
+		testPodMeta,
+	)
+	require.Error(err, "error writing token to file sink")
+}
+
+func startMockServer(t *testing.T, bearerTokenContents string, apiCallCounter *int) (*httptest.Server, *api.Client, string, string) {
+	bearerTokenFile := WriteTempFile(t, bearerTokenContents)
+	tokenFile := WriteTempFile(t, "")
```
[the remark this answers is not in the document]
kschoche: yeah great point, cleaned that up!

subcommand/common/common_test.go
```diff
@@ -26,3 +33,127 @@ func TestValidatePort(t *testing.T) {
...
+	err := ConsulLogin(
+		client,
+		bearerTokenFile,
+		testAuthMethod,
+		tokenFile,
+		testPodMeta,
+	)
+	require.NoError(err)
+	require.Equal(counter, 1)
+	// validate that the token file was written to disk
```
ishustava, line 56:
```suggestion
	// Validate that the token file was written to disk.
```

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,112 @@
...
+		cfg := api.DefaultConfig()
+		c.http.MergeOntoConfig(cfg)
+		c.consulClient, err = consul.NewClient(cfg)
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Unable to get client connection: %s", err))
+			return 1
+		}
+	}
+
+	backoff.Retry(func() error {
```
ishustava, line 85: we need to save err that's returned by `backoff.Retry`
```suggestion
	err := backoff.Retry(func() error {
```

subcommand/connect-init/command.go
```diff
@@ -0,0 +1,112 @@
...
+
+	c.http = &flags.HTTPFlags{}
+
+	flags.Merge(c.flagSet, c.http.Flags())
+	c.help = flags.Usage(help, c.flagSet)
+}
+
+func (c *Command) Run(args []string) int {
+	c.once.Do(c.init)
+	var err error
```
ishustava, line 59:
```suggestion
```

ishustava (approved):
🎉 

Left a couple of minor edits, but otherwise looks great!

subcommand/connect-init/command_test.go
```diff
@@ -0,0 +1,150 @@
...
+	require.Contains(t, ui.ErrorWriter.String(), "Hit maximum retries for consul login")
+}
+
+func TestRun_withRetries(t *testing.T) {
+	cases := []struct {
+		Description        string
+		TestRetry          bool
+		LoginAttemptsCount int
+		ExpCode            int
+		ExpErr             string
```
lkysow, line 60: `ExpErr` isn't used anymore

lkysow (approved):
🦊 great work. Note I didn't dockerize it myself on this last review.

PR #449, opened 2021-03-04
wait for serf check before starting the test

kschoche:
Changes proposed in this PR:
- wait for serf checks to pass before proceeding with the testing because the creation of the `consul` service is driven by completion of the serf check. Without this wait or a manual validation of the consul service existing *in the catalog* the test will run reconcile and pass before the consul service exists in the catalog resulting in the ENT tests failing because they use `Catalog()` API calls to validate final states.

How I've tested this PR:
Unit test fix with ent consul binary

pass CI

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/cleanup_resource_ent_test.go
+++ b/connect-inject/cleanup_resource_ent_test.go
@@ -89,6 +89,7 @@ func TestReconcile_ConsulNamespaces(t *testing.T) {
 			server.WaitForLeader(t)
 			consulClient, err := capi.NewClient(&capi.Config{Address: server.HTTPAddr})
 			require.NoError(err)
+			server.WaitForSerfCheck(t)
 
 			// Register Consul services.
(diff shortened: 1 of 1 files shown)


connect-inject/cleanup_resource_ent_test.go
```diff
@@ -89,6 +89,7 @@ func TestReconcile_ConsulNamespaces(t *testing.T) {
 			server.WaitForLeader(t)
```
lkysow, line 89: let's replace all the `WaitForLeader` with the `WaitForSerfCheck`? Unless you tried that and it didn't work.

kschoche: Yup good idea, I'll make the change!

lkysow (approved):

thisisnotashwin (approved):

PR #453, opened 2021-03-12
Add flag for service registration in consul sidecar

ndhanushkodi:
Changes proposed in this PR:

- Add flag for service registration in consul sidecar.
- This flag allows configuring the consul sidecar to run service
registration, or disable it in favor of a centralized service
regitstration controller.

How I've tested this PR:
Unit tests

Code review
Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/consul-sidecar/command.go
+++ b/subcommand/consul-sidecar/command.go
@@ -27,12 +27,13 @@ const envoyMetricsAddr = "http://127.0.0.1:19000/stats/prometheus"
 type Command struct {
 	UI cli.Ui
 
-	http              *flags.HTTPFlags
-	flagServiceConfig string
-	flagConsulBinary  string
-	flagSyncPeriod    time.Duration
-	flagSet           *flag.FlagSet
-	flagLogLevel      string
+	http                          *flags.HTTPFlags
+	flagEnableServiceRegistration bool
+	flagServiceConfig             string
+	flagConsulBinary              string
+	flagSyncPeriod                time.Duration
+	flagSet                       *flag.FlagSet
+	flagLogLevel                  string
 
 	// Flags to configure metrics merging
 	flagEnableMetricsMerging bool
@@ -59,6 +60,7 @@ type metricsGetter interface {
 
 func (c *Command) init() {
 	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
+	c.flagSet.BoolVar(&c.flagEnableServiceRegistration, "enable-service-registration", true, "Enables consul sidecar to register the service with consul every sync period. Defaults to true.")
 	c.flagSet.StringVar(&c.flagServiceConfig, "service-config", "", "Path to the service config file")
 	c.flagSet.StringVar(&c.flagConsulBinary, "consul-binary", "consul", "Path to a consul binary")
 	c.flagSet.DurationVar(&c.flagSyncPeriod, "sync-period", 10*time.Second, "Time between syncing the service registration. Defaults to 10s.")
@@ -115,7 +117,8 @@ func (c *Command) Run(args []string) int {
 	c.logger = logger
 
 	// Log initial configuration
-	c.logger.Info("Command configuration", "service-config", c.flagServiceConfig,
+	c.logger.Info("Command configuration", "enable-service-registration", c.flagEnableServiceRegistration,
+		"service-config", c.flagServiceConfig,
 		"consul-binary", c.flagConsulBinary,
 		"sync-period", c.flagSyncPeriod,
 		"log-level", c.flagLogLevel,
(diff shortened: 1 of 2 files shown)


subcommand/consul-sidecar/command.go
```diff
@@ -297,20 +302,22 @@ func (c *Command) validateFlags() error {
...
+	if c.flagEnableServiceRegistration {
+		if c.flagServiceConfig == "" {
+			return errors.New("-service-config must be set")
+		}
+		if c.flagConsulBinary == "" {
+			return errors.New("-consul-binary must be set")
+		}
+		_, err := os.Stat(c.flagServiceConfig)
+		if os.IsNotExist(err) {
+			err = fmt.Errorf("-service-config file %q not found", c.flagServiceConfig)
```
kschoche, line 314:
```suggestion
```
I think this is unnecessary?

subcommand/consul-sidecar/command.go
```diff
@@ -166,27 +165,33 @@ func (c *Command) Run(args []string) int {
...
+		c.consulCommand = []string{"services", "register"}
+		c.consulCommand = append(c.consulCommand, c.parseConsulFlags()...)
+		c.consulCommand = append(c.consulCommand, c.flagServiceConfig)
+
+		go func() {
+			for {
+				start := time.Now()
+				cmd := exec.CommandContext(signalCtx, c.flagConsulBinary, c.consulCommand...)
+
+				// Run the command and record the stdout and stderr output
```
kschoche, line 178:
```suggestion
				// Run the command and record the stdout and stderr output.
```

subcommand/consul-sidecar/command.go
ndhanushkodi: thanks for catching that!

subcommand/consul-sidecar/command_test.go
```diff
@@ -79,6 +79,67 @@ func TestRunSignalHandlingRegistration(t *testing.T) {
...
+				for {
+					_, err := conn.Write([]byte("hello"))
+					if err != nil {
+						break
+					}
+				}
+			}()
+
+			// Send a signal to consul-sidecar. The merged metrics server can take
+			// up to metricsServerShutdownTimeout to finish cleaning up.
```
kschoche, line 124: I love the comments here, great work!

kschoche (approved):
Awesome work! I just had 2 small comments which are minor fixes.

subcommand/consul-sidecar/command_test.go
```diff
@@ -79,6 +79,67 @@ func TestRunSignalHandlingRegistration(t *testing.T) {
...
+				// Fail if the signal was not caught.
+				require.Fail(t, "timeout waiting for command to exit")
+			}
+		})
+	}
+}
+
 func TestRunSignalHandlingAllProcessesEnabled(t *testing.T) {
 	cases := map[string]os.Signal{
 		"SIGINT":  syscall.SIGINT,
```
ishustava, line 145: Commenting here because I can't comment on the exact line: I think we'd need to add a test to flag validation to check that the service registration flags are validated only if `-enable-service-registration` is set.

ishustava (commented):
Looks great! Awesome work especially with the testing!!

I left a couple of comments with some improvements. Nothing super major but will wait for those to be resolved before approving.

subcommand/consul-sidecar/command.go
```diff
@@ -297,20 +302,22 @@ func (c *Command) validateFlags() error {
...
+		}
+		_, err := os.Stat(c.flagServiceConfig)
+		if os.IsNotExist(err) {
+			err = fmt.Errorf("-service-config file %q not found", c.flagServiceConfig)
+			return fmt.Errorf("-service-config file %q not found", c.flagServiceConfig)
+		}
+		_, err = exec.LookPath(c.flagConsulBinary)
+		if err != nil {
+			return fmt.Errorf("-consul-binary %q not found: %s", c.flagConsulBinary, err)
+		}
```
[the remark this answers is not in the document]
ndhanushkodi: yes this is a great point, added that check to validate flags. Later on when we add endpoints controller and set enable-service-registration to false, we will want to write some tests around not running consul-sidecar altogether.

subcommand/consul-sidecar/command_test.go
```diff
@@ -79,6 +79,67 @@ func TestRunSignalHandlingRegistration(t *testing.T) {
...
+				}
+			})
+			go func() {
+				for {
+					_, err := conn.Write([]byte("hello"))
+					if err != nil {
+						break
+					}
+				}
+			}()
```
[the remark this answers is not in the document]
ndhanushkodi: yes it will, since it will not be able to write to the connection, so it will break out of the for loop, exiting the goroutine so it can be cleaned up. It's a great idea to add a comment there, thanks!

subcommand/consul-sidecar/command_test.go
ndhanushkodi: The flag validation tests are structured to test only error cases (if they don't error the test would run forever). However the signal handling tests implicitly test the end to end behavior of the non-error cases, including when -enable-service-registration is set to false. WDYT?

ishustava: Oh, you're right! Nevermind!

ishustava (approved):
🎉

PR #454, opened 2021-03-12
Webhook refactor

thisisnotashwin:
Refactor the connect-inject webhook to use the `admission.Webhook` library for it's behavior. The behavior of the webhook should remain unchanged. The difference now is that the webhook does not manage its own certificates. Rather it relies on the webhook cert manager to provision certificates for it as a mounted secret.

This allows us to remove our dependency from DiskCerts and also remove the code involved with cert generation on the connect webhook. This is hopefully allow for the a degree of HA with the deployment as the webhook certificates will be consistent across the pods and webhook configuration.

I have also updated some of the methods within the connect package that previously has method signatures with (*Pod) -> (Pod) as they were not performing modifications on the Pod object.

How to test this PR: Code Review

There will be a companion Helm PR where these changes shall be acceptance tested and will try and ensure behavior is still the same.


connect-inject/handler.go
```diff
@@ -399,85 +315,53 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
+
+	updatedPodJson, err := json.Marshal(pod)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
 
-		resp.Patch = patch
-		patchType := v1beta1.PatchTypeJSONPatch
-		resp.PatchType = &patchType
+	patches, err := jsonpatch.CreatePatch(origPodJson, updatedPodJson)
```
ndhanushkodi, line 345: Nice, I like this over compiling the list of patches with each step!

connect-inject/handler.go
```diff
@@ -399,85 +315,53 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
-		[]corev1.Container{esContainer},
-		"/spec/containers")...)
-
-	// Add annotations so that we know we're injected
-	patches = append(patches, updateAnnotation(
-		pod.Annotations,
-		map[string]string{
-			annotationStatus: injected,
-		})...)
+	pod.Annotations[annotationStatus] = injected
```
ndhanushkodi, line 318: Do we need a nil check here for `pod.Annotations`?

subcommand/inject-connect/command.go
```diff
@@ -454,6 +375,7 @@ func (c *Command) Run(args []string) int {
 		mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 			Scheme:             scheme,
 			LeaderElection:     false,
+			Port:               8080,
```
ndhanushkodi, line 378: This should be `c.flagListen`

ndhanushkodi (commented):
Had a few comments during an initial lookover, I like how it's coming together!

subcommand/inject-connect/command.go
thisisnotashwin: 🙏 thank you!!

ishustava (commented):
The approach looks good to me! 

I had one minor question, but overall I think this is the right direction. Thanks for doing the refactor!

connect-inject/handler.go
thisisnotashwin: Added a comment here!!

thisisnotashwin:
The failing tests seem related to the refactor with consul-init. It was failing on the branch and i didnt spend too much time digging into it.

connect-inject/handler.go
```diff
@@ -215,238 +216,132 @@ type Handler struct {
...
-		}
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("error configuring prometheus annotations: %s", err))
 	}
-	if promAnnotations != nil {
-		patches = append(patches, updateAnnotation(
-			pod.Annotations,
-			promAnnotations)...)
+	for key, value := range promAnnotations {
+		pod.Annotations[key] = value
 	}
```
ishustava, line 321: Not a blocker, but this could potentially move to the `prometheusAnnotations` function. We can just let it modify the annotations on the pod similar to how we do it with `defaultAnnotations`.

connect-inject/handler.go
ishustava: yeah I agree! looks much cleaner!

helper/cert/source_disk.go
```diff
@@ -1,103 +0,0 @@
-package cert
```
ishustava, line 1: ❤️

subcommand/inject-connect/command.go
```diff
@@ -129,16 +123,10 @@ func init() {
 
 func (c *Command) init() {
 	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
-	c.flagSet.StringVar(&c.flagListen, "listen", ":8080", "Address to bind listener to.")
+	c.flagSet.IntVar(&c.flagListen, "listen", 8080, "Address to bind listener to.")
```
ishustava, line 126: I wonder if we should add it to changelog now as a breaking change. I think we might forget to mention it when we merge the feature branch.

Also, should it be port only?
```suggestion
	c.flagSet.IntVar(&c.flagListen, "listen", 8080, "Port to bind listener to.")
```

subcommand/inject-connect/command.go
```diff
@@ -368,83 +356,16 @@ func (c *Command) Run(args []string) int {
...
-	}
-	if c.flagCertFile != "" {
-		certSource = &cert.DiskSource{
-			CertPath: c.flagCertFile,
-			KeyPath:  c.flagKeyFile,
-		}
-	}
-
 	// Create the certificate notifier so we can update for certificates,
 	// then start all the background routines for updating certificates.
```
ishustava, line 360:
```suggestion
	// Create a context to be used by the processes started in this command.
```

subcommand/inject-connect/command.go
```diff
@@ -368,83 +356,16 @@ func (c *Command) Run(args []string) int {
...
-	go func() {
-		c.UI.Info(fmt.Sprintf("Listening on %q...", c.flagListen))
-		if err := server.ListenAndServeTLS("", ""); err != nil {
-			c.UI.Error(fmt.Sprintf("Error listening: %s", err))
-			serverErrors <- err
-		}
-	}()
-
-	// Create a channel for all controllers' exits.
+	// Create a channel for all controllers' exits
```
ishustava, line 368:
```suggestion
	// Create a channel for all controllers' exits.
```

subcommand/inject-connect/command.go
ishustava: Alternatively, to keep backward compatibility, we could keep our current set up and parse the string into host and port. Then set both when create new manager.

subcommand/inject-connect/command_test.go
```diff
@@ -272,7 +253,7 @@ func testSignalHandling(sig os.Signal) func(*testing.T) {
 		select {
 		case exitCode := <-exitChan:
 			require.Equal(t, 0, exitCode, ui.ErrorWriter.String())
-		case <-time.After(time.Second * 1):
+		case <-time.After(time.Second * 2):
```
ishustava, line 256: curious why we needed to bump this time?

ishustava (commented):
Looks great!!! I ❤️  how you cleaned up the handler and that we don't need cert watcher anymore and to run our own webhook server!

I left some minor comments. The only blocking question on my side is whether we should keep the `-listen` flag backward compatible.

subcommand/inject-connect/command_test.go
thisisnotashwin: Nice catch! The manager did not exit within 1 second but once i bumped the wait time to 2 seconds, we would always exit 0. My suspicion here is that the manager just takes a little longer to exit once it receives the cancel signal.

subcommand/inject-connect/command.go
thisisnotashwin: I liked your second suggestion of maintaining backwards compatibility so i updated it to do the same and added some tests around the same!

subcommand/inject-connect/command_test.go
ishustava: oh got it! thanks for clarifying!

ishustava (approved):
Looks great!

connect-inject/handler.go
```diff
@@ -215,238 +216,127 @@ type Handler struct {
...
-		"/spec/containers")...)
+	pod.Spec.Containers = append(pod.Spec.Containers, connectContainer)
 
-	// Add annotations so that we know we're injected
-	patches = append(patches, updateAnnotation(
-		pod.Annotations,
-		map[string]string{
-			annotationStatus: injected,
-		})...)
+	// The annotations here have already been initialized be h.defaultAnnotations()
```
kschoche, line 307:
```suggestion
	// The annotations here have already been initialized by h.defaultAnnotations()
```

connect-inject/handler.go
```diff
@@ -456,18 +346,16 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
-				Result: &metav1.Status{
-					Message: fmt.Sprintf("Error checking or creating namespace: %s", err),
-				},
-			}
+			return admission.Errored(http.StatusInternalServerError, fmt.Errorf("error checking or creating namespace: %s", err))
 		}
 	}
 
-	return resp
+	// return a Patched response along with the patches we intend on applying to the
```
kschoche, line 353:
```suggestion
	// Return a Patched response along with the patches we intend on applying to the
```

connect-inject/handler.go
```diff
@@ -779,3 +639,8 @@ func findServiceAccountVolumeMount(pod *corev1.Pod) (corev1.VolumeMount, error)
 
 	return volumeMount, nil
 }
+
+func (h *Handler) InjectDecoder(d *admission.Decoder) error {
+	h.decoder = d
+	return nil
+}
```
kschoche, line 646:
```suggestion
```

Feel free to let me know why this is needed, but I pulled the branch and (after fixing a bug in the endpoints_controller_test()) it looks like everything compiles and the tests run fine without it?

Maybe there's a discrepancy in libraries that I have vs yours?

btw I had to apply this patch to get it to build, I'm not sure what the state of the base branch is or what the plans are with merging it, but for the PR to pass tests I needed this:
```
kyle@kyles-mbp consul-k8s % git diff
diff --git a/connect-inject/endpoints_controller_test.go b/connect-inject/endpoints_controller_test.go
index 643ea0c..592e3ae 100644
--- a/connect-inject/endpoints_controller_test.go
+++ b/connect-inject/endpoints_controller_test.go
@@ -223,7 +223,7 @@ func TestProcessUpstreams(t *testing.T) {
                                DenyK8sNamespacesSet:  mapset.NewSetWith(),
                        }

-                       upstreams, err := ep.processUpstreams(tt.pod())
+                       upstreams, err := ep.processUpstreams(*tt.pod())
                        if tt.expErr != "" {
                                require.EqualError(t, err, tt.expErr)
                        } else {
```

connect-inject/handler.go
```diff
@@ -506,22 +394,16 @@ func (h *Handler) shouldInject(pod *corev1.Pod, namespace string) (bool, error)
...
 		if cs := pod.Spec.Containers; len(cs) > 0 {
-			// Create the patch for this first, so that the Annotation
-			// object will be created if necessary
-			*patches = append(*patches, updateAnnotation(
-				pod.Annotations,
-				map[string]string{annotationService: cs[0].Name})...)
-
 			// Set the annotation for checking in shouldInject
-			pod.ObjectMeta.Annotations[annotationService] = cs[0].Name
+			pod.Annotations[annotationService] = cs[0].Name
```
kschoche, line 406: I was just curious about why we no longer need to use the `pod.ObjectMeta.Annotation` fields, is it because that was used for the old patching and isn't necessary now? I do see that we had sometimes been using `pod.Annotations[]` elsewhere instead of ObjectMeta. 🤷🏽  Non-blocker, just curious.

kschoche (approved):
This looks great.
I left a couple "nit" suggestions, and two questions but otherwise 🆗 !
(the compilation error of course being a blocker but I'm sure it'll be sorted)

connect-inject/handler.go
thisisnotashwin: So this is a runtime dependency and is used by the Manager to provide a decoder to the handler to enable it to decode Kubernetes objects in the fly. While in the tests we can provision a decoder directly, this method is how Dependency Injection is done during runtime. Does this answer your question?

WRT to patch, yeah, the stacked PRs have caused this at the moment and a rebase should potentially fix it. Im holding off on the rebase till the endpoints controller gets merged, but it will be address in the code that does get merged.

connect-inject/handler.go
thisisnotashwin: pod.Annotations[] == pod.ObjectMeta.Annotations[] and is supported natively by the kubernetes library and is shorter and im lazy 😄

connect-inject/handler.go
kschoche: Cool, thanks for the clarification!

subcommand/inject-connect/command_test.go
```diff
@@ -188,53 +202,21 @@ func TestRun_ResourceLimitDefaults(t *testing.T) {
...
-		"-consul-k8s-image", "hashicorp/consul-k8s", "-consul-image", "foo", "-envoy-image", "envoy:1.16.0",
-		"-enable-health-checks-controller=true",
-		"-http-addr=http://0.0.0.0:9999",
-		"-listen", "999999",
-	})
+	flags := []string{"-consul-k8s-image", "hashicorp/consul-k8s", "-consul-image", "foo", "-envoy-image", "envoy:1.16.0",
+		"-enable-health-checks-controller=true"}
+
+	os.Setenv(api.HTTPAddrEnvName, "%")
+	code := cmd.Run(flags)
```
ndhanushkodi, line 215: Did this work in CI? I think this would run into the same issue I was running into, where there is no kubeconfig for the manager.

I learned from @lkysow  that the original purpose of the `TestRun_CommandFailsWithInvalidListener` test was to make sure that when there was an error that the whole command would actually exit rather than keep on running (apparently there was a bug/issue where the healthchecks controller would keep running).

Making sure that the command exits on validation errors still seems to satisfy that purpose and makes sense to me. So, if we can do that while skipping the manager itself running (if the manager causes this test to fail) then  that would be good. Currently, the only way I can think of doing that is adding a conditional around running the manager, and simply not running it during tests that do `cmd.Run`

ndhanushkodi (commented):
I love the cleanup and simplification of the code you've done!! Amazing work!! Had a few suggestions around comments/logging and a question/thought about the tests in CI with the manager running.

subcommand/inject-connect/command_test.go
ndhanushkodi: Oh right I realize you can't test this in CI until after you can rebase on endpoints controller. Can we take a closer look at this test after the rebase and make sure it's testing the original purpose well, and that it passes?

connect-inject/handler.go
```diff
@@ -638,32 +508,30 @@ func (h *Handler) serviceMetricsPath(pod *corev1.Pod) string {
 // prometheusAnnotations returns the Prometheus scraping configuration
 // annotations. It returns a nil map if metrics are not enabled and annotations
 // should not be set.
-func (h *Handler) prometheusAnnotations(pod *corev1.Pod) (map[string]string, error) {
-	enableMetrics, err := h.enableMetrics(pod)
+func (h *Handler) prometheusAnnotations(pod *corev1.Pod) error {
```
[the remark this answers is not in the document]
thisisnotashwin: I had updated this based on a suggestion from Iryna. We are updated the annotations on the pod within this method on lines 523-525 and are hence passing in a pointer reference.

connect-inject/handler.go
```diff
@@ -638,32 +508,30 @@ func (h *Handler) serviceMetricsPath(pod *corev1.Pod) string {
...
+	enableMetrics, err := h.enableMetrics(*pod)
```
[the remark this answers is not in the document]
thisisnotashwin: This now goes back to being required 😄

connect-inject/handler.go
```diff
@@ -638,32 +508,30 @@ func (h *Handler) serviceMetricsPath(pod *corev1.Pod) string {
...
 	if err != nil {
-		return map[string]string{}, err
+		return err
 	}
-	prometheusScrapePort, err := h.prometheusScrapePort(pod)
+	prometheusScrapePort, err := h.prometheusScrapePort(*pod)
```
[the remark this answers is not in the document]
thisisnotashwin: This now goes back to being required 😄

connect-inject/handler.go
```diff
@@ -638,32 +508,30 @@ func (h *Handler) serviceMetricsPath(pod *corev1.Pod) string {
...
 	if err != nil {
-		return map[string]string{}, err
+		return err
 	}
-	prometheusScrapePath := h.prometheusScrapePath(pod)
+	prometheusScrapePath := h.prometheusScrapePath(*pod)
```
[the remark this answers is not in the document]
thisisnotashwin: This now goes back to being required 😄

subcommand/inject-connect/command_test.go
thisisnotashwin: That sounds good! This diff definitely makes it harder to look at. The validation test for the invalid listener now fails before starting the manager because the webhook runs in the same process as the manager does but the validation error occurs before the manager starts and hence does run successfully in CI. 

`inject-connect/command.go` L386-395 capture this validation error and they will return before the manager starts.

connect-inject/handler.go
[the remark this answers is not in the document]
ndhanushkodi: Ahh nice I missed that! will resolve this and the related!

subcommand/inject-connect/command_test.go
ndhanushkodi: Ah I see, that makes sense!

ndhanushkodi (approved):
Looks great!!

PR #456, opened 2021-03-15
Fix incorrect validation for ServiceResolver

lkysow:
Changes proposed in this PR:
- `ServiceResolver` validation: if hashPolicy.field is set fieldValue must also be set 
- `ServiceResolver` validation: hashPolicy.field can be empty if hashPolicy.sourceIP is true

How I've tested this PR:
- unit tests

- review tests & changes

Checklist:
- [x] Tests added

### DIFF
--- a/api/v1alpha1/serviceresolver_types.go
+++ b/api/v1alpha1/serviceresolver_types.go
@@ -478,16 +478,21 @@ func (in *LoadBalancer) validate(path *field.Path) field.ErrorList {
 
 func (in HashPolicy) validate(path *field.Path) field.ErrorList {
 	var errs field.ErrorList
-	validFields := []string{"header", "cookie", "query_parameter"}
-	if !sliceContains(validFields, in.Field) {
-		errs = append(errs, field.Invalid(path.Child("field"), in.Field,
-			notInSliceMessage(validFields)))
+	if in.Field != "" {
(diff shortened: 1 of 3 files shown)


api/v1alpha1/serviceresolver_types.go
```diff
@@ -478,16 +478,21 @@ func (in *LoadBalancer) validate(path *field.Path) field.ErrorList {
...
 
-	if in.Field != "" && in.SourceIP {
-		asJSON, _ := json.Marshal(in)
-		errs = append(errs, field.Invalid(path, string(asJSON),
-			"cannot set both field and sourceIP"))
+		if in.SourceIP {
+			asJSON, _ := json.Marshal(in)
+			errs = append(errs, field.Invalid(path, string(asJSON),
+				"cannot set both field and sourceIP"))
+		} else if in.FieldValue == "" {
```
lkysow, line 492: using else if here because if they're setting:
```
field: cookie
sourceIP: true
```

then thought only one error (can't set both field and sourceIP) is clearer than two: (can't set both field and sourceIP, if field is set fieldValue must also be set)

thisisnotashwin (approved):
Nicely done!!

ndhanushkodi (approved):
Looks good to me!

PR #457, opened 2021-03-17
Add watcher for agent pods to endpoints controller

thisisnotashwin:
Changes proposed in this PR:
- This watcher watches for Consul Agent pods to be in a running phase and the condition ready to be true and then reconcile all endpoints that have a ready/not-ready address that share a node name with that of the consul agent pod.

How I've tested this PR:
- Unit tests
- Iryna tested it

- Code review

Will link an acceptance test run to show that behavior hasnt been impacted.

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -11,11 +11,17 @@ import (
 	"github.com/hashicorp/consul/api"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 type EndpointsController struct {
@@ -387,7 +393,11 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
 func (r *EndpointsController) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Endpoints{}).
-		Complete(r)
+		Watches(
+			&source.Kind{Type: &corev1.Pod{}},
+			&handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(r.requestsForRunningAgentPods)},
+			builder.WithPredicates(predicate.NewPredicateFuncs(r.filterAgentPods)),
+		).Complete(r)
 }
 
 // getConsulClient returns an *api.Client that points at the consul agent local to the pod.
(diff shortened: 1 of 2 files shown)


connect-inject/endpoints_controller.go
```diff
@@ -442,7 +448,86 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
...
+	labels := meta.GetLabels()
+	app, ok := labels["app"]
+	if !ok {
+		return false
+	}
+	component, ok := labels["component"]
+	if !ok {
+		return false
+	}
+	// todo: do we want to filter on other labels?
```
kschoche, line 472: I believe we had talked about filtering on `release` as well?
I don't think it is totally required, but the more strict we are the better?

thisisnotashwin: That does make sense. We would need the value of the release to be something that is provided to the command struct which we currently do not perform. I know we want to make this change within the endpoints reconciler as well so we should be able to make the change in both places when we have resolved some of those specifics.

connect-inject/endpoints_controller_test.go
```diff
@@ -849,6 +851,525 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app":       "consul",
+						"component": "not-client",
+					},
+				},
+			},
+			expected: false,
+		},
+		"label[app]=consul and label[component]!=client": {
```
kschoche, line 916:
```suggestion
		"label[app]!=consul and label[component]!=client": {
```

connect-inject/endpoints_controller.go
```diff
@@ -442,7 +448,86 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
...
+	}
+	return false
+}
+
+// requestsForRunningAgentPods creates a slice of requests for the endpoints controller. It enqueues a request for each endpoint that needs to be reconciled.
+// It iterates through the list of endpoints and creates a request for those endpoints that have an address that are on the same node as the new Consul Agent pod.
+// It receives a Pod Object which is a Consul Agent that has been filtered by filterAgentPods and only processes endpoints for pods where the Ready condition is true.
+func (r EndpointsController) requestsForRunningAgentPods(object handler.MapObject) []ctrl.Request {
+	var consulClientPod corev1.Pod
+	r.Log.Info("received update for consulClientPod", "podName", object.Meta.GetName(), "consulClientPod", object.Object)
```
ishustava, line 484: Will this print the whole pod object? if so it may be too verbose for the info log and hard to read. wdyt?

connect-inject/endpoints_controller.go
```diff
@@ -442,7 +448,86 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
...
+	err := r.Client.Get(r.Context, types.NamespacedName{Name: object.Meta.GetName(), Namespace: object.Meta.GetNamespace()}, &consulClientPod)
+	if k8serrors.IsNotFound(err) {
+		// Ignore if consulClientPod is not found
```
ishustava, line 487:
```suggestion
		// Ignore if consulClientPod is not found.
```

connect-inject/endpoints_controller.go
```diff
@@ -442,7 +448,86 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
...
+	}
+	// We only care about agents that are already running.
+	if consulClientPod.Status.Phase != corev1.PodRunning {
+		r.Log.Info("ignoring consulClientPod because it's not running", "consulClientPod", consulClientPod.Name)
+		return []ctrl.Request{}
+	}
+	// We only care about pods that are ready because we can't talk to the consul client otherwise.
+	for _, cond := range consulClientPod.Status.Conditions {
+		if cond.Type == corev1.PodReady && cond.Status != corev1.ConditionTrue {
+			// Ignore if consulClientPod is not ready
```
ishustava, line 502:
```suggestion
			// Ignore if consulClientPod is not ready.
```

connect-inject/endpoints_controller.go
```diff
@@ -442,7 +448,86 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
...
+
+	// Get the list of all endpoints.
+	var endpointsList corev1.EndpointsList
+	err = r.Client.List(r.Context, &endpointsList)
+	if err != nil {
+		r.Log.Error(err, "failed to list endpoints")
+		return []ctrl.Request{}
+	}
+
+	// Re-enqueue requests for endpoints from this agent
```
ishustava, line 516:
```suggestion
	// Re-enqueue requests for endpoints from this agent.
```

connect-inject/endpoints_controller_test.go
```diff
@@ -849,6 +851,525 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
...
+
+	for name, test := range cases {
+		t.Run(name, func(t *testing.T) {
+			controller := EndpointsController{}
+
+			result := controller.filterAgentPods(test.meta, nil)
+			require.Equal(t, test.expected, result)
+		})
+	}
+}
```
ishustava, line 937: nice! this was super easy to read!

connect-inject/endpoints_controller_test.go
```diff
@@ -849,6 +851,525 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
 	}
 }
 
+func TestFilterAgentPods(t *testing.T) {
```
ishustava, line 854: we should add `t.Parallel()` to all tests I think.

connect-inject/endpoints_controller.go
```diff
@@ -442,7 +448,86 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
...
+	// todo: do we want to filter on other labels?
+	if app == "consul" && component == "client" {
+		return true
+	}
+	return false
+}
+
+// requestsForRunningAgentPods creates a slice of requests for the endpoints controller. It enqueues a request for each endpoint that needs to be reconciled.
+// It iterates through the list of endpoints and creates a request for those endpoints that have an address that are on the same node as the new Consul Agent pod.
+// It receives a Pod Object which is a Consul Agent that has been filtered by filterAgentPods and only processes endpoints for pods where the Ready condition is true.
```
ishustava, line 481: could we break this up so the lines aren't too long?

connect-inject/endpoints_controller_test.go
```diff
@@ -849,6 +851,525 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
...
+				Context: ctx,
+				Scheme:  s,
+				Log:     logger,
+			}
+
+			requests := controller.requestsForRunningAgentPods(handler.MapObject{Meta: test.agentPod})
+			require.ElementsMatch(t, requests, test.expectedRequests)
+		})
+	}
+}
```
ishustava, line 1371: do you think we need a test case for when pod is not found?

ishustava (approved):
Looks great!! I've left some minor comments, but nothing blocking.

Re acceptance test, I think we'd need to write a new one for this case (currently we're not testing it). I think it's ok for it to come after the PR is merged, but it's up to you. I think we just need to make sure we have coverage before we merge the feature branch.

connect-inject/endpoints_controller.go
thisisnotashwin: good catch. im going to only print the podName. It seems like enough information!

connect-inject/endpoints_controller_test.go
thisisnotashwin: done :)

connect-inject/endpoints_controller.go
thisisnotashwin: DONE 😄

connect-inject/endpoints_controller.go
```diff
@@ -442,7 +448,86 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
...
-		Complete(r)
+		Watches(
+			&source.Kind{Type: &corev1.Pod{}},
+			&handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(r.requestsForRunningAgentPods)},
+			builder.WithPredicates(predicate.NewPredicateFuncs(r.filterAgentPods)),
+		).Complete(r)
+}
+
+// filterAgentPods receives meta and object information for Kubernetes resources that are being watched,
+// which in this case are Pods. It only returns true if the Pod is a Consul Agent Pod. It reads the labels
```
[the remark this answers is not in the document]
thisisnotashwin: Always take care of your future self, that's what I say 😉

connect-inject/endpoints_controller_test.go
thisisnotashwin: DONE!

connect-inject/endpoints_controller.go
thisisnotashwin: done done done done

connect-inject/endpoints_controller.go
```diff
@@ -439,6 +449,89 @@ func shouldIgnore(namespace string, denySet, allowSet mapset.Set) bool {
...
+	}
+	return false
+}
+
+// requestsForRunningAgentPods creates a slice of requests for the endpoints controller.
+// It enqueues a request for each endpoint that needs to be reconciled. It iterates through
+// the list of endpoints and creates a request for those endpoints that have an address that
+// are on the same node as the new Consul Agent pod. It receives a Pod Object which is a
+// Consul Agent that has been filtered by filterAgentPods and only processes endpoints
+// for pods where the Ready condition is true.
```
ndhanushkodi, line 483:
```suggestion
// for client agent pods where the Ready condition is true.
```
For this one I initially thought it was talking about the pods referenced by the endpoints, so added the specification just to be clear :)

ndhanushkodi (approved):
Great work Ashwin, it was really well structured, and the unit tests have great coverage and are easy to understand! I had a few nitpicky comment-related suggestions, and please to feel free to pick and choose which ones you like and ignore ones you don't!

connect-inject/endpoints_controller.go
```diff
@@ -421,6 +431,92 @@ func shouldIgnore(namespace string, denySet, allowSet mapset.Set) bool {
 	return false
 }
 
+// filterAgentPods receives meta and object information for Kubernetes resources that are being watched,
+// which in this case are Pods. It only returns true if the Pod is a Consul Client Agent Pod. It reads the labels
+// from the meta of the resource and uses the values of the "app" and "component" label to validate that
+// the Pod is a Consul Client Agent.
+func (r EndpointsController) filterAgentPods(meta metav1.Object, object runtime.Object) bool {
```
lkysow, line 438: @thisisnotashwin These should be function pointers I think: `r *EndpointsController`

connect-inject/endpoints_controller.go
```diff
@@ -387,7 +393,11 @@ func (r *EndpointsController) Logger(name types.NamespacedName) logr.Logger {
 func (r *EndpointsController) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Endpoints{}).
-		Complete(r)
+		Watches(
+			&source.Kind{Type: &corev1.Pod{}},
+			&handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(r.requestsForRunningAgentPods)},
+			builder.WithPredicates(predicate.NewPredicateFuncs(r.filterAgentPods)),
+		).Complete(r)
```
lkysow, line 400: woah this is fancy! Cool!

connect-inject/endpoints_controller.go
thisisnotashwin: sorry for the late response. They don't seem to use any values from the `r` struct. this might be something I dont quite understand here but why would it need to be a function pointer?

lkysow: Just because the other functions in this struct do use the struct pointer so these should be consistent. Or if they're not accessing anything on the struct they could be functions instead.

thisisnotashwin: that is a good point. i looked at the other method and i think it does make sense to use pointers to keep them consistent. plus i could add some logging here just for some added information in which case i can use the logger from the struct! good catch!

PR #458, opened 2021-03-18
Add readiness check to controller-manager

thisisnotashwin:
Changes proposed in this PR:
- Add readiness check to manager

This leads to a readiness check configured as follows in the deployment
```
livenessProbe:
  httpGet:
    path: /readyz/ready
    port: 9445
    scheme: HTTP
  failureThreshold: 2
  initialDelaySeconds: 1
  periodSeconds: 2
  successThreshold: 1
  timeoutSeconds: 5
readinessProbe:
  httpGet:
    path: /readyz/ready
    port: 9445
    scheme: HTTP
  failureThreshold: 2
  initialDelaySeconds: 2
  periodSeconds: 2
  successThreshold: 1
  timeoutSeconds: 5
```

I have re-used the readiness check but can easily create a separate liveness check which can be a different path.

How I've tested this PR: created an image from this build and tinkered with the helm config on the `webhook-refactor` consul-helm branch and saw it fail until i figured out the config to make it go green

How I expect reviewers to test this PR: this is a draft but i wanted feedback on:
1) what would be a good liveness check and do we need one?
2) what would be a good readiness check and do we need one?

Currently i check if the webhook certs exist on disk but that was for trial purposes.

Checklist:
- [ ] Tests added


kschoche:
This is pretty neat!
If I had to pick "something" to use as a health check for the controller I would imagine it should probably be  GET on `/v1/status/leader` because a huge chunk of the controller is issuing client calls against the agent? Thoughts?

thisisnotashwin:
I do like that idea..my only question is wouldnt it be taken care of by the reconciler that would reconcile the endpoints as the Client pods became ready? but i do like it more than what is currently in place

ndhanushkodi:
I don't think we need a liveness check since if the manager exits with any error, the command should exit, triggering a Pod restart.

For readiness, we only want reconcile requests going to the controller if Consul is in a state to handle service registrations, and so @kschoche's idea makes sense to me. 

> wouldnt it be taken care of by the reconciler that would reconcile the endpoints as the Client pods became ready?
I am not sure I am following-- I'm not sure that the reconciler would take care of making sure Consul is up. I think it would make service registration requests against an un-ready Consul, so it would make sense to me to check that Consul has a leader and can perform writes before sending traffic to the controller. Let me know if I'm missing something here though!!

thisisnotashwin:
@ndhanushkodi Sorry, I phrased that poorly. Let me try that again.

Given the reconcile loop of the endpoints controller will "fail" to register an endpoint when the Consul servers/agents arent ready, it will re-enqueue those endpoints which will eventually get reconciled when the Consul agents do become alive. So does it matter that the controller starts receiving requests before Consul is ready?

This gets trickier with a webhook, as when a webhook isn't ready, it can actively reject requests from incoming Pods. I was thinking of gearing readiness to when the webhook is ready to serve requests and those dont seem as coupled to the availability of the Consul server IMO. 

wrt the liveness check, I agree. We used to previously have one so I was just wondering out loud.

ishustava:
>If I had to pick "something" to use as a health check for the controller I would imagine it should probably be GET on /v1/status/leader because a huge chunk of the controller is issuing client calls against the agent? Thoughts?

I like this idea as well. I think this is something that [sync catalog does too](https://github.com/hashicorp/consul-k8s/blob/33b2fdaa57eade0d7cd351cf2354595ce13799c1/subcommand/sync-catalog/command.go#L361-L370) and it performs very similar operation of syncing services from Kubernetes to Consul. Maybe the ideal scenario would be to do some sort of `and` check to make sure that both the controller and the webhook are healthy:
1. for controller, check that the consul clients are available
2. for webhook, check for certs.

ndhanushkodi:
>Given the reconcile loop of the endpoints controller will "fail" to register an endpoint when the Consul servers/agents arent ready, it will re-enqueue those endpoints which will eventually get reconciled when the Consul agents do become alive. So does it matter that the controller starts receiving requests before Consul is ready?

Thanks for this clarification, I totally see what you're saying. I am not sure what's better, letting the requests error or stopping them from reaching the controller. I suppose it could be a little nicer to use the readiness check to stop it from reaching the controller so there isn't a bunch of churn upfront.

If we do go ahead with the readiness check I do like @ishustava's suggestion of waiting for both of these conditions.

thisisnotashwin:
Closing as this has now been done correctly!

PR #459, opened 2021-03-19
add tls support and tests to connect-init command

kschoche:
Changes proposed in this PR:
- Add TLS tests to `connect-init/command_test.go`
- Modify the `common.GenerateServerCerts` function to do it's own cleanup and propogate this to existing tests.
The reason for this is that it helped to clean up my test so that I didn't have to generate server certs when TLS is disabled in the table test due to scoping of the cleanup func, as well as it generally cleans up existing code.
- Remove old copy of generateServerCerts from `server-acl-init` test and convert it to the one in common.
- In case it isn't obvious `connect-init` already works with TLS so no actual code is changed. Inside of `connect-inject/container_init.go` we are passing in the required ENV variables in the command template to enable it to be picked up by the consul client:
```
{{- if .ConsulCACert}}
export CONSUL_HTTP_ADDR="https://${HOST_IP}:8501"
export CONSUL_GRPC_ADDR="https://${HOST_IP}:8502"
export CONSUL_CACERT=/consul/connect-inject/consul-ca.pem
cat <<EOF >/consul/connect-inject/consul-ca.pem
{{ .ConsulCACert }}
EOF
{{- else}}
{{- end}}
consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
  {{- if .AuthMethod }}
```
How I've tested this PR:
Unit tests have been updated to enable TLS and pass.
Manually tested TLS enabled with a deployment.

Unit tests have been updated to enable TLS and pass.
Manually tested TLS enabled with a deployment.

Checklist:
- [x] Tests added


subcommand/connect-init/command_test.go
```diff
@@ -45,140 +45,235 @@ func TestRun_FlagValidation(t *testing.T) {
...
 
-	// Start Consul server.
-	server, err := testutil.NewTestServerConfigT(t, nil)
-	defer server.Stop()
-	require.NoError(t, err)
-	server.WaitForLeader(t)
-	consulClient, err := api.NewClient(&api.Config{Address: server.HTTPAddr})
-	require.NoError(t, err)
+			var caFile, certFile, keyFile string
+			//defer cleanup()
```
kschoche, line 213:
```suggestion
```

subcommand/connect-init/command_test.go
```diff
@@ -45,140 +45,230 @@ func TestRun_FlagValidation(t *testing.T) {
...
-		Name:        testAuthMethod,
-		Description: "Kubernetes Auth Method",
-		Type:        "kubernetes",
-		Config: map[string]interface{}{
-			"Host":              k8sMockServer.URL,
-			"CACert":            serviceAccountCACert,
-			"ServiceAccountJWT": serviceAccountJWTToken,
+	cases := []struct {
+		name   string
+		secure bool
```
lkysow, line 55: we use `secure` in the acceptance tests to mean acls & tls so maybe we can call this `tls` just to avoid confusion.

subcommand/connect-init/command_test.go
```diff
@@ -45,140 +45,230 @@ func TestRun_FlagValidation(t *testing.T) {
...
-		AuthMethod:  testAuthMethod,
-		BindType:    api.BindingRuleBindTypeService,
-		BindName:    "${serviceaccount.name}",
-		Selector:    "serviceaccount.name!=default",
-	}
-	_, _, err = consulClient.ACL().BindingRuleCreate(&aclBindingRule, nil)
-	require.NoError(t, err)
+			var caFile, certFile, keyFile string
+			// Start Consul server with ACLs enabled and default deny policy.
+			var masterToken = "b78d37c7-0ca7-5f4d-99ee-6d9975ce4586"
```
lkysow, line 74:
```suggestion
			masterToken := "b78d37c7-0ca7-5f4d-99ee-6d9975ce4586"
```

subcommand/connect-init/command_test.go
```diff
@@ -45,140 +45,230 @@ func TestRun_FlagValidation(t *testing.T) {
...
+				c.ACL.Tokens.Master = masterToken
+				if test.secure {
+					caFile, certFile, keyFile = common.GenerateServerCerts(t)
+					c.CAFile = caFile
+					c.CertFile = certFile
+					c.KeyFile = keyFile
+				}
+			})
+			defer server.Stop()
+			require.NoError(t, err)
```
lkysow, line 87:
```suggestion
			require.NoError(t, err)
			defer server.Stop()
```

`server` could be nil if `err != nil` which would result in panic

lkysow: I saw that below as well

lkysow (approved):
Nice work! Good catch on the generate certs duplication.

subcommand/connect-init/command_test.go
lkysow: applies below too

kschoche: good point, will change, thanks!

subcommand/connect-init/command_test.go
kschoche: Great catch, found this 3 other places! thx!

subcommand/common/test_util.go
```diff
@@ -46,12 +44,12 @@ func GenerateServerCerts(t *testing.T) (string, string, string, func()) {
 	_, err = certKeyFile.WriteString(keyPem)
 	require.NoError(err)
 
-	cleanupFunc := func() {
+	t.Cleanup(func() {
 		os.Remove(caFile.Name())
 		os.Remove(certFile.Name())
 		os.Remove(certKeyFile.Name())
```
ishustava, line 50: Great! I really like this pattern!

subcommand/server-acl-init/command_test.go
```diff
@@ -2075,50 +2073,6 @@ func getBootToken(t *testing.T, k8s *fake.Clientset, prefix string, k8sNamespace
 	return string(bootToken)
 }
 
-// generateServerCerts generates Consul CA
-// and a server certificate and saves them to temp files.
-// It returns file names in this order:
-// CA certificate, server certificate, and server key.
-// Note that it's the responsibility of the caller to
-// remove the temporary files created by this function.
-func generateServerCerts(t *testing.T) (string, string, string, func()) {
```
ishustava, line 2084: nice catch!

ishustava (approved):
Looks great 🎉 !! Thanks for adding these tests 🙏

PR #461, opened 2021-03-23
tproxy use quotes in ServicesWithFilter's filter

kschoche:
Changes proposed in this PR:
- when filter items have `-`'s we need to wrap the query in quotes when passing a filter to `consulClient.Agent().ServicesWithFilter()`.
- Update unit tests to use hyphenated names for pods and namespaces.

How I've tested this PR:
Changing `testPodName` from `counting` to `counting-foo` resulted in the unit tests failing, after the change they pass.

code review

Checklist:
- [x] Tests added


ndhanushkodi (approved):
Looks good to me! I like the change to use the tests to drive it out.

thisisnotashwin (approved):
sweet

PR #465, opened 2021-03-24
Tproxy writefile changes

kschoche:
Changes proposed in this PR:
- `ioutil.WriteFile()` will create the new file with the requested permissions prior to actually writing the file. This causes us to accidentally create it with "0444" which will not allow us to complete the write.
- Use `ioutil.WriteFile(.., .., 0666)` in combination with  `os.Chmod(0444)`  to create the file with the correct privs, and remove it first if it already exists in the event that we failed the command and the container is restarted.
- We missed this in the previous tests because we were using `WriteTempFile()` to create the files prior to passing them into the functions to be tested so that we could have easy cleanup, and that had default 0666 privs on it. In end to end testing this produced permission denied errors.

How I've tested this PR:
unit tests
run this on live system with ACL Login.

unit tests passes and live test with ACLs enabled.

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/common/common.go
+++ b/subcommand/common/common.go
@@ -75,10 +75,25 @@ func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkF
 		return fmt.Errorf("error logging in: %s", err)
 	}
 
-	// Write the token out to file with permissions so consul-k8s user can read.
-	payload := []byte(tok.SecretID)
-	if err := ioutil.WriteFile(tokenSinkFile, payload, 0444); err != nil {
+	if err := WriteFileWithPerms(tokenSinkFile, tok.SecretID, 0444); err != nil {
 		return fmt.Errorf("error writing token to file sink: %v", err)
 	}
 	return nil
+}
+
+// WriteFileWithPerms will write payload as the contents of the outputFile and set permissions.
+func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
+	// ioutil.WriteFile truncates existing files or if they do not exist writes them as FileMode(0666), but only if
+	// the file is currently writable. If the file exists it will already likely be read-only. Remove it.
+	if _, err := os.Stat(outputFile); err == nil {
+		err = os.Remove(outputFile)
+		if err != nil {
+			return fmt.Errorf("unable to delete existing token file: %s", err)
+		}
+	}
+	err := ioutil.WriteFile(outputFile, []byte(payload), os.FileMode(0666))
+	if err != nil {
+		return fmt.Errorf("unable to write file: %s", err)
+	}
+	return os.Chmod(outputFile, mode)
 }
--- a/subcommand/common/common.go
+++ b/subcommand/common/common.go
@@ -86,8 +86,7 @@ func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
 	// os.WriteFile truncates existing files and overwrites them, but only if they are writable.
 	// If the file exists it will already likely be read-only. Remove it first.
 	if _, err := os.Stat(outputFile); err == nil {
-		err = os.Remove(outputFile)
-		if err != nil {
+		if err = os.Remove(outputFile); err != nil {
 			return fmt.Errorf("unable to delete existing file: %s", err)
 		}
 	}
--- a/subcommand/common/common.go
+++ b/subcommand/common/common.go
@@ -83,16 +83,15 @@ func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkF
 
 // WriteFileWithPerms will write payload as the contents of the outputFile and set permissions.
 func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
-	// ioutil.WriteFile truncates existing files or if they do not exist writes them as FileMode(0666), but only if
-	// the file is currently writable. If the file exists it will already likely be read-only. Remove it.
+	// os.WriteFile truncates existing files and overwrites them, but only if they are writable.
+	// If the file exists it will already likely be read-only. Remove it first.
 	if _, err := os.Stat(outputFile); err == nil {
 		err = os.Remove(outputFile)
 		if err != nil {
-			return fmt.Errorf("unable to delete existing token file: %s", err)
+			return fmt.Errorf("unable to delete existing file: %s", err)
 		}
 	}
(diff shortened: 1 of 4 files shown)


subcommand/common/common.go
```diff
@@ -75,10 +75,25 @@ func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkF
...
 }
+
+// WriteFileWithPerms will write payload as the contents of the outputFile and set permissions.
+func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
+	// ioutil.WriteFile truncates existing files or if they do not exist writes them as FileMode(0666), but only if
+	// the file is currently writable. If the file exists it will already likely be read-only. Remove it.
+	if _, err := os.Stat(outputFile); err == nil {
+		err = os.Remove(outputFile)
+		if err != nil {
+			return fmt.Errorf("unable to delete existing token file: %s", err)
```
thisisnotashwin, line 91:
```suggestion
			return fmt.Errorf("unable to delete existing file: %s", err)
```

subcommand/common/common.go
```diff
@@ -75,10 +75,25 @@ func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkF
...
+		}
```
thisisnotashwin, line 92:
```suggestion
		if err = os.Remove(outputFile); err != nil {
			return fmt.Errorf("unable to delete existing token file: %s", err)
		}
```

subcommand/common/common.go
```diff
@@ -75,10 +75,25 @@ func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkF
...
+	}
+	err := ioutil.WriteFile(outputFile, []byte(payload), os.FileMode(0666))
+	if err != nil {
+		return fmt.Errorf("unable to write file: %s", err)
+	}
```
thisisnotashwin, line 97: Can we use `0777` instead of `0666`? Shouldn't matter given we `os.Chmod()` it anyway right?
```suggestion
	if err := ioutil.WriteFile(outputFile, []byte(payload), os.ModePerm); err != nil {
		return fmt.Errorf("unable to write file: %s", err)
	}
```

subcommand/common/common.go
```diff
@@ -75,10 +75,25 @@ func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkF
...
+func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
+	// ioutil.WriteFile truncates existing files or if they do not exist writes them as FileMode(0666), but only if
+	// the file is currently writable. If the file exists it will already likely be read-only. Remove it.
+	if _, err := os.Stat(outputFile); err == nil {
+		err = os.Remove(outputFile)
+		if err != nil {
+			return fmt.Errorf("unable to delete existing token file: %s", err)
+		}
+	}
+	err := ioutil.WriteFile(outputFile, []byte(payload), os.FileMode(0666))
```
thisisnotashwin, line 94:
```suggestion
	err := os.WriteFile(outputFile, []byte(payload), os.FileMode(0666))
```

thisisnotashwin: Looks like `ioutil.WriteFile()` just invokes `os.WriteFile()` under the hood

thisisnotashwin (commented):
This looks good. Left a few suggestions!!

subcommand/common/common.go
kschoche: Great suggestions here and good catch on the os/ioutil changes.
I think going 0777 is fine as well.
Funny story: You are unable to apply suggested changes to the same line and all three of these touch (but dont change) the same lines so I have to do this all manually :D

subcommand/common/common.go
```diff
@@ -75,10 +75,24 @@ func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkF
...
-	// Write the token out to file with permissions so consul-k8s user can read.
-	payload := []byte(tok.SecretID)
-	if err := ioutil.WriteFile(tokenSinkFile, payload, 0444); err != nil {
+	if err := WriteFileWithPerms(tokenSinkFile, tok.SecretID, 0444); err != nil {
 		return fmt.Errorf("error writing token to file sink: %v", err)
 	}
 	return nil
 }
+
+// WriteFileWithPerms will write payload as the contents of the outputFile and set permissions.
```
ndhanushkodi, line 84:
```suggestion
// WriteFileWithPerms will write payload as the contents of the outputFile and set permissions after writing the contents. This function is necessary since using ioutil.WriteFile() alone will create the new file with the requested permissions prior to actually writing the file, so you can't set read-only permissions.
```
Wanted to capture the context for why this function was written here as well!

ndhanushkodi (approved):
Looks great! 🎉 If you think the context is good, I added a suggestion to the function comment why we can't use ioutil.WriteFile alone

subcommand/common/common.go
kschoche: After a little bit of research this is in go 1.16 but not 1.14 so skipping for now! :)

subcommand/common/common.go
kschoche: Love it, thanks @ndhanushkodi !!

subcommand/common/common.go
```diff
@@ -75,10 +75,24 @@ func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkF
...
+
+// WriteFileWithPerms will write payload as the contents of the outputFile and set permissions after writing the contents. This function is necessary since using ioutil.WriteFile() alone will create the new file with the requested permissions prior to actually writing the file, so you can't set read-only permissions.
+func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
+	// os.WriteFile truncates existing files and overwrites them, but only if they are writable.
+	// If the file exists it will already likely be read-only. Remove it first.
+	if _, err := os.Stat(outputFile); err == nil {
+		err = os.Remove(outputFile)
+		if err != nil {
+			return fmt.Errorf("unable to delete existing file: %s", err)
+		}
```
thisisnotashwin, line 92:
```suggestion
		if err = os.Remove(outputFile); err != nil {
			return fmt.Errorf("unable to delete existing file: %s", err)
		}
```

thisisnotashwin (approved):
💥 nicely done!!

PR #466, opened 2021-03-25
Fix last updated time

thisisnotashwin:
Changes proposed in this PR:
- Update the type of lastSynced time to allows a nil value to be set. Previously, the reconciler would fail to set it to `null` causing an error if the resource could not sync successfully. This now sets it to nil.

How I've tested this PR: 
Helm install with imageK8S: `ashwinvenkatesh/consul-k8s@sha256:eae7a1d7e3aab3cb2060ee3a882803f225d1cb676c9c8283f3ef7a01d2cde056`

apply crd:
```yaml
kind: ServiceSplitter
metadata:
  name: service-a
spec:
  splits:
    - weight: 90
      serviceSubset: v1
    - weight: 10
      serviceSubset: v2
``` 

Check controller logs. Previously an error message would show up:
`invalid: status.lastSyncedTime: Invalid value: \"null\": status.lastSyncedTime in body must be of type string: \"null\""`
It no longer does

How I expect reviewers to test this PR: code-review

Checklist:
- [ ] Tests added


lkysow (approved):
remember to rebase since this PR has the squashed commits

ndhanushkodi (approved):
🎉 Nice!! Thanks for fixing this!!

PR #467, opened 2021-03-25
Enable endpoints controller

ishustava:
This PR enables the endpoints controller and configures the connect-init container to now search for services. It removes the old functionality where the services were registered in the init container. The full list of changes is below

#### container_init
Remove service registration from the init container template and the corresponding tests. 

Note that I have removed some of the tests prematurely, even though that logic has not yet been implemented in the endpoints controller. Namely, tests for metrics configuration and namespaces. However, I think that is OK as these tests will be added back when we do implement those. 

For other removed tests, if there wasn't already coverage in the endpoints controller, I've moved them there.

#### consul-sidecar
Consul sidecar is now only injected if we need to run the metrics merging server. The `consulSidecar` function now fails if you try to run it when metrics merging isn't enabled, and if it is, we  configure it to not register services.

#### envoy-sidecar
Remove the preStop hook from the envoy sidecar container. The hook will no longer work as there is no service.hcl file, plus services should now be deregistered only be the endpoints controller.

#### connect-init
Remove the temporary `-skip-service-registration-polling` flag since we now always need to poll for services. 

#### endpoints-controller
Enable endpoints controller in the inject-connect command.

---

How I've tested this PR:
Ran connect, controller, mesh-, ingress-, and terminating gateway acceptance tests only in insecure configuration as endpoints controller does not yet support TLS and ACLs [link](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/2683/workflows/9cf6736b-4409-4621-9a2a-c01eff82cd3f/jobs/9837).

Note that I've added an acceptance test for when consul clients are restarted [here](https://github.com/hashicorp/consul-helm/pull/886/files#diff-891cc7979b4831efe0da57ce148d04232bfc98129bf291559d41e24c9ee0be6aR169-R197). 

Acceptance tests. However, if you would like or have time, you could also test it manually.

Checklist:
- [x] Tests added


connect-inject/patch.go
```diff
@@ -1,131 +0,0 @@
-package connectinject
```
ishustava, line 1: everything here was unused, except for `escapeJSONPointer`, which was only used in tests. I've moved it to the test that was using it instead.

subcommand/connect-init/command_test.go
```diff
@@ -859,8 +865,8 @@ var (
 		Name: "counting-sidecar-proxy",
 		Kind: "connect-proxy",
 		Proxy: &api.AgentServiceConnectProxyConfig{
-			DestinationServiceName: "foo",
-			DestinationServiceID:   "foo",
+			DestinationServiceName: "counting",
+			DestinationServiceID:   "counting-counting",
```
ishustava, line 869: this is to fix errors in consul agent logs

connect-inject/handler.go
```diff
@@ -291,26 +207,39 @@ func (h *Handler) Handle(_ context.Context, req admission.Request) admission.Res
...
 	// pod.Annotations has already been initialized by h.defaultAnnotations()
 	// and does not need to be checked for being a nil value.
 	pod.Annotations[annotationStatus] = injected
 
 	// Add annotations for metrics.
-	err = h.prometheusAnnotations(&pod)
+	err = h.prometheusAnnotations(pod)
 	if err != nil {
 		h.Log.Error("Error configuring prometheus annotations", "err", err, "Request Name", req.Name)
 		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("error configuring prometheus annotations: %s", err))
```
thisisnotashwin, line 245:
```suggestion
	if err = h.prometheusAnnotations(pod); err != nil {
		h.Log.Error("Error configuring prometheus annotations", "err", err, "Request Name", req.Name)
		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("error configuring prometheus annotations: %s", err))
```

connect-inject/handler.go
```diff
@@ -291,26 +207,39 @@ func (h *Handler) Handle(_ context.Context, req admission.Request) admission.Res
...
 	}
-	pod.Spec.Containers = append(pod.Spec.Containers, connectContainer)
 
 	// pod.Annotations has already been initialized by h.defaultAnnotations()
 	// and does not need to be checked for being a nil value.
 	pod.Annotations[annotationStatus] = injected
 
 	// Add annotations for metrics.
-	err = h.prometheusAnnotations(&pod)
+	err = h.prometheusAnnotations(pod)
```
thisisnotashwin, line 242: Also, given we are actually updating the pod's annotations for metrics in `prometheusAnnotations` I *think*, we should send a pointer reference and not a value one.

connect-inject/handler_test.go
```diff
@@ -348,7 +384,23 @@ func TestHandlerHandle(t *testing.T) {
...
+					Operation: "add",
+					Path:      "/metadata/annotations/" + escapeJSONPointer("prometheus.io/scrape"),
+				},
+				{
+					Operation: "add",
+					Path:      "/metadata/annotations/" + escapeJSONPointer("prometheus.io/port"),
+				},
+				{
+					Operation: "add",
+					Path:      "/metadata/annotations/" + escapeJSONPointer("prometheus.io/path"),
```
thisisnotashwin, line 403: If one were to make a request to add `"prometheus.io/*"` to the constants in the `annotations.go` file, how much would you hate them? Asking for a friend 😛
The friend has observed that they have been used twice in tests and once in handler.go

thisisnotashwin (commented):
This looks so excellent!! Have a few comments and suggestions. The one regarding passing in a pod pointer reference to metrics annotations is the one that is potentially "blocking"

connect-inject/annotations.go
```diff
@@ -0,0 +1,89 @@
+package connectinject
```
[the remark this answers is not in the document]
kschoche: seconding this comment ^

connect-inject/endpoints_controller.go
```diff
@@ -313,6 +336,14 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(k8sSvcName, k8sSvcNam
 	return nil
 }
 
+// serviceInstancesForK8SServiceNameAndNamespace calls Consul's ServicesWithFilter to get the list
+// of services instances that have the provided k8sServiceName and k8sServiceNamespace in their metadata.
+func serviceInstancesForK8SServiceNameAndNamespace(k8sServiceName, k8sServiceNamespace string, client *api.Client) (map[string]*api.AgentService, error) {
```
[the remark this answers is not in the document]
ishustava: 😄 That was your idea!

connect-inject/annotations.go
```diff
@@ -0,0 +1,89 @@
...
+	annotationMeta = "consul.hashicorp.com/service-meta-"
+
+	// annotationSyncPeriod controls the -sync-period flag passed to the
+	// consul-k8s consul-sidecar command. This flag controls how often the
+	// service is synced (i.e. re-registered) with the local agent.
+	//
+	// Deprecated: This annotation is no longer supported.
+	annotationSyncPeriod = "consul.hashicorp.com/connect-sync-period"
+
+	// annotations for sidecar proxy resource limits
```
kschoche, line 60:
```suggestion
	// annotations for sidecar proxy resource limits.
```

connect-inject/annotations.go
```diff
@@ -0,0 +1,89 @@
...
+	// envoy binary. See list of args here: https://www.envoyproxy.io/docs/envoy/latest/operations/cli
+	// e.g. consul.hashicorp.com/envoy-extra-args: "--log-level debug --disable-hot-restart"
+	// The arguments passed in via this annotation will take precendence over arguments
+	// passed via the -envoy-extra-args flag.
+	annotationEnvoyExtraArgs = "consul.hashicorp.com/envoy-extra-args"
+
+	// annotationConsulNamespace is the Consul namespace the service is registered into.
+	annotationConsulNamespace = "consul.hashicorp.com/consul-namespace"
+
+	// injected is used as the annotation value for annotationInjected
```
kschoche, line 87:
```suggestion
	// injected is used as the annotation value for annotationInjected.
```

connect-inject/container_init.go
```diff
@@ -349,111 +203,9 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
   {{- end }}
   {{- end }}
 
-# Register the service. The HCL is stored in the volume so that
```
kschoche, line 352: Removing all of this 🥳

connect-inject/endpoints_controller_test.go
```diff
@@ -1920,6 +2000,110 @@ func TestRequestsForRunningAgentPods(t *testing.T) {
...
+					},
+					Meta: map[string]string{"k8s-service-name": k8sSvc, "k8s-namespace": k8sNS},
+				},
+			},
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			servicesInConsul := []*api.AgentServiceRegistration{
```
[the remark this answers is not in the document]
ishustava: I was thinking about that case too, but then decided that it's essentially covered by the first three cases. Now that I'm thinking about it though, I think you're right! let me add those.

connect-inject/handler.go
ishustava: Oh you're right! I totally missed that, good catch!

connect-inject/endpoints_controller_test.go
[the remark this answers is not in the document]
kschoche: Kind of a test of whether or not consul client filtering works, but if we were going to add additional services along this line I would vote that one of them should run in a different k8s namespace.

connect-inject/health_check_resource_test.go
```diff
@@ -349,6 +349,7 @@ func TestReconcilePod(t *testing.T) {
 			// Get a server, client, and handler.
 			server, client, resource := testServerAgentResourceAndController(t, tt.Pod)
 			defer server.Stop()
+			server.WaitForLeader(t)
```
kschoche, line 352: nice catch! have been finding one of these every week..

connect-inject/endpoints_controller_test.go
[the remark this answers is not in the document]
ishustava: Done!

connect-inject/handler.go
ishustava: Even though the tests didn't fail because `pod.Annotations` is a pointer/map, I've changed it to a pointer to make it more obvious that this function mutates an object.

connect-inject/handler_test.go
ishustava: Done!

connect-inject/endpoints_controller_test.go
```diff
@@ -115,6 +115,7 @@ func TestProcessUpstreams(t *testing.T) {
 		expected    []api.Upstream
 		expErr      string
 		configEntry func() api.ConfigEntry
+		consulDown  bool
```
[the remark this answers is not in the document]
ishustava: updated!

connect-inject/endpoints_controller_test.go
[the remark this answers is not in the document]
ishustava: @kschoche that makes sense, I'll update the test with your suggestion

thisisnotashwin (approved):
🍏 📗 🟢 💚 🥗 🟩

kschoche (approved):
Excellent work!

PR #468, opened 2021-03-26
Rebase master on feature-tproxy

thisisnotashwin:
Changes proposed in this PR:
- Rebase master to get the latest updates to controller-runtime and associated libraries
- Update code and test fixtures to use the updated libraries.

How I've tested this PR:

Checklist:
- [ ] Tests added

### DIFF
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -49,24 +49,24 @@ type EndpointsController struct {
 	ReleaseNamespace string
 	Log              logr.Logger
 	Scheme           *runtime.Scheme
-	Ctx              context.Context
+	context.Context
 }
 
-func (r *EndpointsController) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var serviceEndpoints corev1.Endpoints
 
 	if shouldIgnore(req.Namespace, r.DenyK8sNamespacesSet, r.AllowK8sNamespacesSet) {
 		return ctrl.Result{}, nil
 	}
 
-	err := r.Client.Get(r.Ctx, req.NamespacedName, &serviceEndpoints)
+	err := r.Client.Get(ctx, req.NamespacedName, &serviceEndpoints)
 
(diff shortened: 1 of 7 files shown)


connect-inject/endpoints_controller.go
```diff
@@ -473,10 +472,11 @@ func (r EndpointsController) filterAgentPods(meta metav1.Object, object runtime.
...
 // Consul Agent that has been filtered by filterAgentPods and only enqueues endpoints
 // for client agent pods where the Ready condition is true.
-func (r EndpointsController) requestsForRunningAgentPods(object handler.MapObject) []ctrl.Request {
+func (r EndpointsController) requestsForRunningAgentPods(object client.Object) []ctrl.Request {
 	var consulClientPod corev1.Pod
-	r.Log.Info("received update for consulClientPod", "podName", object.Meta.GetName())
-	err := r.Client.Get(r.Ctx, types.NamespacedName{Name: object.Meta.GetName(), Namespace: object.Meta.GetNamespace()}, &consulClientPod)
+	ctx := context.Background()
+	r.Log.Info("received update for consulClientPod", "podName", object.GetName())
+	err := r.Client.Get(ctx, types.NamespacedName{Name: object.GetName(), Namespace: object.GetNamespace()}, &consulClientPod)
```
ishustava, line 479: Previously, we were using the same context from the controller, but now we can't I guess because we don't have it yet? Could we pass it from the manager so that it's still one context for the controller?

thisisnotashwin: yeah.. i was trying to think of a way of attempting to re-use it from the manager, but also wanted to avoid having the Reconciler pass around a context and also have one defined in the struct. let me revisit this real quick

thisisnotashwin: added back context but I only used it in that one function. I would like Reconcile to use the context that is passed to it from the machinery tho

thisisnotashwin: I do like not having to create a new context here though 😄

ndhanushkodi (approved):
This looks great to me! Thanks for doing this.

kschoche (approved):
🚒 great work! love the new controller runtime interfaces too, much more readable.

PR #469, opened 2021-03-31
Refactor metrics merging to work with endpoints-controller

ndhanushkodi:
Changes proposed in this PR:
- Refactor metrics merging to work with endpoints-controller

    Previously, all metrics configuration was dealt with in the mutating
    webhook handler. Now, since service/proxy registration happens in
    endpoints-controller, some of the metrics configuration needs to be used
    in endpoints-controller as well. There is still some functionality in
    the handler that also needs metrics configuration, such as adding
    prometheus annotations and running the consul sidecar.

    This refactor pulls out common configuration for metrics into a
    MetricsConfiguration struct and adds the methods for getting values from
    flags and annotations to that struct, so it can be commonly used between
    endpoints-controller and the webhook handler.

~I wasn't able to E2E test this due to consul-k8s and consul-helm being incompatible on the feature-tproxy branch. Once we make the Helm PR for feature-tproxy, we should have confidence that the acceptance tests cover the metrics merging functionality.~

How I've tested this PR:
- Unit tests, Acceptance test for merged metrics with image "gcr.io/nitya-293720/consul-k8s-dev:tproxymetrics2" (with small changes to https://github.com/hashicorp/consul-helm/pull/886 to add a service and kustomize.yaml)

- Code Review

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/consul_sidecar.go
+++ b/connect-inject/consul_sidecar.go
@@ -1,7 +1,6 @@
 package connectinject
 
 import (
-	"errors"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
@@ -12,36 +11,19 @@ import (
 // It always disables service registration because for connect we no longer
 // need to keep services registered as this is handled in the endpoints-controller.
 func (h *Handler) consulSidecar(pod corev1.Pod) (corev1.Container, error) {
-	run, err := h.shouldRunMergedMetricsServer(pod)
+	metricsPorts, err := h.MetricsConfig.mergedMetricsServerConfiguration(pod)
 	if err != nil {
 		return corev1.Container{}, err
 	}
-
-	// This should never happen because we only call this function in the handler if
-	// we need to run the metrics merging server. This check is here just in case.
-	if !run {
-		return corev1.Container{}, errors.New("metrics merging should be enabled in order to inject the consul-sidecar")
-	}
-
-	// Configure consul sidecar with the appropriate metrics flags.
-	mergedMetricsPort, err := h.mergedMetricsPort(pod)
-	if err != nil {
-		return corev1.Container{}, err
-	}
-	serviceMetricsPath := h.serviceMetricsPath(pod)
-
-	// Don't need to check the error since it's checked in the call to
-	// h.shouldRunMergedMetricsServer() above.
-	serviceMetricsPort, _ := h.serviceMetricsPort(pod)
 
 	command := []string{
 		"consul-k8s",
 		"consul-sidecar",
 		"-enable-service-registration=false",
 		"-enable-metrics-merging=true",
-		fmt.Sprintf("-merged-metrics-port=%s", mergedMetricsPort),
-		fmt.Sprintf("-service-metrics-port=%s", serviceMetricsPort),
-		fmt.Sprintf("-service-metrics-path=%s", serviceMetricsPath),
+		fmt.Sprintf("-merged-metrics-port=%s", metricsPorts.mergedPort),
+		fmt.Sprintf("-service-metrics-port=%s", metricsPorts.servicePort),
+		fmt.Sprintf("-service-metrics-path=%s", metricsPorts.servicePath),
 	}
 
 	return corev1.Container{
--- a/connect-inject/consul_sidecar_test.go
+++ b/connect-inject/consul_sidecar_test.go
@@ -6,35 +6,19 @@ import (
 	"github.com/hashicorp/go-hclog"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
-
-var (
-	consulSidecarResources = corev1.ResourceRequirements{
-		Requests: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("10m"),
-			corev1.ResourceMemory: resource.MustParse("25Mi"),
-		},
-		Limits: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("20m"),
-			corev1.ResourceMemory: resource.MustParse("50Mi"),
-		},
-	}
-)
-
-// NOTE: This is tested here rather than in handler_test because doing it there
-// would require a lot of boilerplate to get at the underlying patches that would
-// complicate understanding the tests (which are simple).
 
 // Test that if the conditions for running a merged metrics server are true,
 // that we pass the metrics flags to consul sidecar.
 func TestConsulSidecar_MetricsFlags(t *testing.T) {
 	handler := Handler{
-		Log:                         hclog.Default().Named("handler"),
-		ImageConsulK8S:              "hashicorp/consul-k8s:9.9.9",
-		DefaultEnableMetrics:        true,
-		DefaultEnableMetricsMerging: true,
+		Log:            hclog.Default().Named("handler"),
+		ImageConsulK8S: "hashicorp/consul-k8s:9.9.9",
+		MetricsConfig: MetricsConfig{
+			DefaultEnableMetrics:        true,
+			DefaultEnableMetricsMerging: true,
+		},
 	}
 	container, err := handler.consulSidecar(corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
@@ -58,23 +42,4 @@ func TestConsulSidecar_MetricsFlags(t *testing.T) {
 	require.Contains(t, container.Command, "-merged-metrics-port=20100")
 	require.Contains(t, container.Command, "-service-metrics-port=8080")
 	require.Contains(t, container.Command, "-service-metrics-path=/metrics")
-}
-
-// Test that the Consul sidecar errors when metrics merging is disabled.
-func TestConsulSidecar_ErrorsWhenMetricsMergingIsDisabled(t *testing.T) {
-	handler := Handler{
-		Log:                    hclog.Default().Named("handler"),
-		ImageConsulK8S:         "hashicorp/consul-k8s:9.9.9",
-		ConsulSidecarResources: consulSidecarResources,
-	}
-	_, err := handler.consulSidecar(corev1.Pod{
-		Spec: corev1.PodSpec{
-			Containers: []corev1.Container{
-				{
-					Name: "web",
-				},
-			},
-		},
-	})
-	require.EqualError(t, err, "metrics merging should be enabled in order to inject the consul-sidecar")
 }
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -77,34 +77,16 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 		return corev1.Container{}, fmt.Errorf("serviceAccountName %q does not match service name %q", pod.Spec.ServiceAccountName, data.ServiceName)
 	}
 
-	// If metrics are enabled, the init container should set up
-	// envoy_prometheus_bind_addr so there's a listener on 0.0.0.0 that points
-	// to a metrics backend. The metrics backend is determined by the call to
-	// h.shouldRunMergedMetricsServer(). If there is a merged metrics server,
-	// the backend would be that server. If we are not running the merged
-	// metrics server, the backend should just be the Envoy metrics endpoint.
-	enableMetrics, err := h.enableMetrics(pod)
-	if err != nil {
-		return corev1.Container{}, err
-	}
-	data.EnableMetrics = enableMetrics
-
-	prometheusScrapePort, err := h.prometheusScrapePort(pod)
-	if err != nil {
-		return corev1.Container{}, err
-	}
-	data.PrometheusScrapeListener = fmt.Sprintf("0.0.0.0:%s", prometheusScrapePort)
-
 	// This determines how to configure the consul connect envoy command: what
 	// metrics backend to use and what path to expose on the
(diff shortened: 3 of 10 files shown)


connect-inject/common_inject_configuration.go
```diff
@@ -0,0 +1,160 @@
+package connectinject
+
+import (
+	"fmt"
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type InjectConfiguration struct {
```
thisisnotashwin, line 10: We could call this metrics configuration given it's specificity?

connect-inject/common_inject_configuration.go
```diff
@@ -0,0 +1,160 @@
+package connectinject
+
```
thisisnotashwin, line 2: this file could potentially read `metrics_configuration.go`?

connect-inject/consul_sidecar.go
```diff
@@ -12,7 +12,7 @@ import (
 // It always disables service registration because for connect we no longer
 // need to keep services registered as this is handled in the endpoints-controller.
 func (h *Handler) consulSidecar(pod corev1.Pod) (corev1.Container, error) {
-	run, err := h.shouldRunMergedMetricsServer(pod)
+	run, err := h.InjectConfig.shouldRunMergedMetricsServer(pod)
```
thisisnotashwin, line 15: here is an idea: can we create a method on `InjectConfig` that returns all the ports you need if `run=true` but an `error` otherwise? We seem to make 4 distinct calls to `h.InjectConfig` which made me wonder if we could just do it in 1?

thisisnotashwin (commented):
My thoughts: I really like the direction of this refactor/cleanup!! In terms of naming, i'd migrate `InjectConfiguration` -> `MetricsConfiguration` given the specificity of the config.

I also potentially like grouping all of our requests and limits into a similar struct (not a part of this change here, but more as a future adoption of this pattern of grouping related entities into their own struct and file for readability.

connect-inject/common_inject_configuration.go
kschoche: agree on the naming bits here!

connect-inject/consul_sidecar.go
kschoche: Yeah good point. I had this same thought in the original code, would be nice to just have a function to return all the ports!

kschoche:
I really like this, great work so far. Echoing a few of the naming related things Ashwin mentioned otherwise I think it's a good approach.

ndhanushkodi:
Thanks @thisisnotashwin and @kschoche for your feedback!! I've made edits based on this and cleaned up the PR, so re-requesting your reviews for this more final version.

connect-inject/handler.go
```diff
@@ -116,11 +112,7 @@ type Handler struct {
 	// Default metrics settings. These will configure where Prometheus scrapes
 	// metrics from, and whether to run a merged metrics endpoint on the consul
 	// sidecar. These can be overridden via pod annotations.
```
thisisnotashwin, line 114: This might need to be re-written just a little bit.

thisisnotashwin (commented):
This looks really good. Have a few suggestions but it should be close to ready after those changes.

connect-inject/handler.go
lkysow: +1

connect-inject/metrics_configuration.go
```diff
@@ -0,0 +1,194 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type MetricsConfiguration struct {
```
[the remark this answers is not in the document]
lkysow: needs godoc too

connect-inject/consul_sidecar_test.go
```diff
@@ -31,10 +31,12 @@ var (
 // that we pass the metrics flags to consul sidecar.
 func TestConsulSidecar_MetricsFlags(t *testing.T) {
```
lkysow, line 32: the var `consulSidecarResources` above here can be deleted since it's not used

connect-inject/container_init.go
```diff
@@ -77,34 +77,16 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
...
-	if err != nil {
-		return corev1.Container{}, err
-	}
-	data.PrometheusScrapeListener = fmt.Sprintf("0.0.0.0:%s", prometheusScrapePort)
-
 	// This determines how to configure the consul connect envoy command: what
 	// metrics backend to use and what path to expose on the
 	// envoy_prometheus_bind_addr listener for scraping.
-	run, err := h.shouldRunMergedMetricsServer(pod)
+	run, err := h.MetricsConfiguration.shouldRunMergedMetricsServer(pod)
```
lkysow, line 83: I think `run` here is too generic. What about `metricsServer`?

connect-inject/endpoints_controller.go
```diff
@@ -225,7 +226,25 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
 	proxyConfig := &api.AgentServiceConnectProxyConfig{
 		DestinationServiceName: serviceName,
 		DestinationServiceID:   serviceID,
-		Config:                 nil, // TODO: add config for metrics (upcoming PR)
+		Config:                 map[string]interface{}{},
```
lkysow, line 229: nit: I think we're to use `make(map[string]interface{})` for instantiations of empty maps.

connect-inject/endpoints_controller.go
```diff
@@ -225,7 +226,25 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
...
+	if err != nil {
+		return nil, nil, err
+	}
+	prometheusScrapePort, err := r.MetricsConfiguration.prometheusScrapePort(pod)
+	if err != nil {
+		return nil, nil, err
+	}
+	prometheusScrapeListener := fmt.Sprintf("0.0.0.0:%s", prometheusScrapePort)
+	if enableMetrics {
+		proxyConfig.Config["envoy_prometheus_bind_addr"] = prometheusScrapeListener
```
lkysow, line 247: `envoy_prometheus_bind_addr` should be a const

connect-inject/endpoints_controller.go
```diff
@@ -225,7 +226,25 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
...
+	// If metrics are enabled, the proxyConfig should set envoy_prometheus_bind_addr to a listener on 0.0.0.0 on
+	// the prometheusScrapePort that points to a metrics backend. The backend for this listener will be determined by
+	// the envoy bootstrapping command (consul connect envoy) configuration in the init container. If there is a merged
+	// metrics server, the backend would be that server. If we are not running the merged metrics server, the backend
+	// should just be the Envoy metrics endpoint.
+	enableMetrics, err := r.MetricsConfiguration.enableMetrics(pod)
+	if err != nil {
+		return nil, nil, err
+	}
+	prometheusScrapePort, err := r.MetricsConfiguration.prometheusScrapePort(pod)
```
lkysow, line 241: these can be nested under a check for `enableMetrics`:

```go
	if enableMetrics {
		prometheusScrapePort, err := r.MetricsConfiguration.prometheusScrapePort(pod)
		if err != nil {
			return nil, nil, err
		}
		prometheusScrapeListener := fmt.Sprintf("0.0.0.0:%s", prometheusScrapePort)
		proxyConfig.Config["envoy_prometheus_bind_addr"] = prometheusScrapeListener
	}

```

lkysow (approved):
LGTM

connect-inject/metrics_configuration.go
```diff
@@ -0,0 +1,194 @@
...
+	DefaultPrometheusScrapePath string
+}
+
+const (
+	defaultServiceMetricsPath = "/metrics"
+)
+
+// mergedMetricsServerConfiguration is called when running a merged metrics server and used to return ports necessary to
+// configure the merged metrics server.
+func (mc MetricsConfiguration) mergedMetricsServerConfiguration(pod corev1.Pod) (string, string, string, error) {
```
lkysow, line 25: oh one thing I forgot to say is that at the very least, each return value needs to be documented here and it might make sense to actually return a struct here, e.g. `MetricsPorts` with fields `MergedPort`, `ServicePort`, `MetricsPath`.

connect-inject/metrics_configuration.go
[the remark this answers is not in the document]
kschoche: +1 on `MetricsConfig`

connect-inject/metrics_configuration.go
```diff
@@ -0,0 +1,194 @@
...
+	// Don't need to check the error since it's checked in the call to
+	// mc.shouldRunMergedMetricsServer() above.
+	serviceMetricsPort, _ := mc.serviceMetricsPort(pod)
+
+	serviceMetricsPath := mc.serviceMetricsPath(pod)
+
+	return mergedMetricsPort, serviceMetricsPort, serviceMetricsPath, nil
+}
+
+// enableMetrics returns the default value in the handler, or overrides that
```
kschoche, line 52: super minor nit: but this function and the next have the same comment, maybe their comments could be changed so they explain how they're not the same!

connect-inject/consul_sidecar_test.go
ndhanushkodi: I also deleted the comment above since it was probably for different tests (when consul sidecar had other functionality)

connect-inject/endpoints_controller.go
ndhanushkodi: nice cleanup!

connect-inject/metrics_configuration.go
ndhanushkodi: yeah that's a good point, I updated this!

connect-inject/metrics_configuration.go
ndhanushkodi: good idea, I changed this to use the struct. It could be unexported since it's only used within the package.

connect-inject/consul_sidecar.go
```diff
@@ -12,36 +11,19 @@ import (
...
 		"consul-k8s",
 		"consul-sidecar",
 		"-enable-service-registration=false",
 		"-enable-metrics-merging=true",
-		fmt.Sprintf("-merged-metrics-port=%s", mergedMetricsPort),
-		fmt.Sprintf("-service-metrics-port=%s", serviceMetricsPort),
-		fmt.Sprintf("-service-metrics-path=%s", serviceMetricsPath),
+		fmt.Sprintf("-merged-metrics-port=%s", metricsPorts.mergedPort),
+		fmt.Sprintf("-service-metrics-port=%s", metricsPorts.servicePort),
+		fmt.Sprintf("-service-metrics-path=%s", metricsPorts.servicePath),
```
kschoche, line 26: nice, I ❤️ this cleanup!

connect-inject/metrics_configuration.go
```diff
@@ -0,0 +1,206 @@
...
+			return "", fmt.Errorf("%s annotation value of %d is not in the valid port range 1-65535", annotation, port)
+		} else if !privileged && (port < 1024 || port > 65535) {
+			return "", fmt.Errorf("%s annotation value of %d is not in the unprivileged port range 1024-65535", annotation, port)
+		}
+
+		// if the annotation exists, return the validated port
+		return fmt.Sprint(port), nil
+	}
+
+	// if the annotation does not exist, return the default
```
kschoche, line 197:
```suggestion
	// If the annotation does not exist, return the default.
```

connect-inject/metrics_configuration.go
```diff
@@ -0,0 +1,206 @@
...
+			return "", fmt.Errorf("%s annotation value of %s is not a valid integer", annotation, raw)
+		}
+
+		if privileged && (port < 1 || port > 65535) {
+			return "", fmt.Errorf("%s annotation value of %d is not in the valid port range 1-65535", annotation, port)
+		} else if !privileged && (port < 1024 || port > 65535) {
+			return "", fmt.Errorf("%s annotation value of %d is not in the unprivileged port range 1024-65535", annotation, port)
+		}
+
+		// if the annotation exists, return the validated port
```
kschoche, line 193:
```suggestion
		// If the annotation exists, return the validated port.
```

connect-inject/metrics_configuration.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return false, err
+	}
+	serviceMetricsPort, err := mc.serviceMetricsPort(pod)
+	if err != nil {
+		return false, err
+	}
+
+	// Don't need to check error here since serviceMetricsPort has been
+	// validated by calling mc.serviceMetricsPort above
```
kschoche, line 165:
```suggestion
	// validated by calling mc.serviceMetricsPort above.
```

connect-inject/metrics_configuration_test.go
```diff
@@ -0,0 +1,528 @@
+package connectinject
+
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
```
kschoche, line 9:
```suggestion
	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"github.com/stretchr/testify/require"

```

kschoche (approved):
Awesome stuff here! I love the cleanup that you did.
I left a few nonblocking comments but it looks great!

PR #470, opened 2021-03-31
Add acls and tls to endpoints controller

kschoche:
Changes proposed in this PR:
- Adds support for ACLs and TLS to endpoints controller
- Minor change to how we deal with consul clients in the endpoints controller+tests.

How I've tested this PR:
Unit tests for ACLS+TLS have been added which cover normal paths as well as the three places that where we request a new consulClient:
1. processUpstreams()
2. 2x in Reconcile()  (deregisterServiceOnAllAgents() and regular Reconcile)

Manually run the connect-inject acceptance tests using https://github.com/hashicorp/consul-helm/pull/886/, with `imageK8S: "kschoche/consul-k8s-dev"` and uncommenting the TLS table tests.

Run unit tests / manually deploy and run acceptance tests.

Note: I'm willing to table-fy and add ACL/TLS support to all of the endpoints-controller tests if we think this is really necessary.
I think the tests which use the consul client (`TestReconcileDeleteEndpoint()`, `TestReconcileUpdateEndpoint()`) where I have NOT added ACL/TLS tests are already covered by the the tests I added. I'm willing to make changes but for the sake of readability I didn't want to unnecessarily add complexity to the already long+awesome tests. Thoughts?

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -102,7 +102,7 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 
 				if hasBeenInjected(pod) {
 					// Create client for Consul agent local to the pod.
-					client, err := r.getConsulClient(r.ConsulScheme, r.ConsulPort, pod.Status.HostIP)
+					client, err := r.getConsulClient(pod.Status.HostIP)
 					if err != nil {
 						r.Log.Error(err, "failed to create a new Consul client", "address", pod.Status.HostIP)
 						return ctrl.Result{}, err
@@ -302,7 +302,7 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 	// On each agent, we need to get services matching "k8s-service-name" and "k8s-namespace" metadata.
 	for _, pod := range list.Items {
 		// Create client for this agent.
-		client, err := r.getConsulClient(r.ConsulScheme, r.ConsulPort, pod.Status.PodIP)
+		client, err := r.getConsulClient(pod.Status.PodIP)
 		if err != nil {
 			r.Log.Error(err, "failed to create a new Consul client", "address", pod.Status.PodIP)
 			return err
@@ -413,8 +413,8 @@ func (r *EndpointsController) processUpstreams(pod corev1.Pod) ([]api.Upstream, 
 }
 
 // getConsulClient returns an *api.Client that points at the consul agent local to the pod.
-func (r *EndpointsController) getConsulClient(scheme, port, ip string) (*api.Client, error) {
-	newAddr := fmt.Sprintf("%s://%s:%s", scheme, ip, port)
+func (r *EndpointsController) getConsulClient(ip string) (*api.Client, error) {
+	newAddr := fmt.Sprintf("%s://%s:%s", r.ConsulScheme, ip, r.ConsulPort)
 	localConfig := r.ConsulClientCfg
 	localConfig.Address = newAddr
 
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -102,7 +102,7 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 
 				if hasBeenInjected(pod) {
 					// Create client for Consul agent local to the pod.
-					client, err := r.getConsulClient(pod.Status.HostIP)
+					client, err := r.remoteConsulClient(pod.Status.HostIP)
 					if err != nil {
 						r.Log.Error(err, "failed to create a new Consul client", "address", pod.Status.HostIP)
 						return ctrl.Result{}, err
@@ -302,7 +302,7 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 	// On each agent, we need to get services matching "k8s-service-name" and "k8s-namespace" metadata.
 	for _, pod := range list.Items {
 		// Create client for this agent.
-		client, err := r.getConsulClient(pod.Status.PodIP)
+		client, err := r.remoteConsulClient(pod.Status.PodIP)
 		if err != nil {
 			r.Log.Error(err, "failed to create a new Consul client", "address", pod.Status.PodIP)
 			return err
@@ -412,8 +412,8 @@ func (r *EndpointsController) processUpstreams(pod corev1.Pod) ([]api.Upstream, 
 	return upstreams, nil
 }
 
-// getConsulClient returns an *api.Client that points at the consul agent local to the pod.
-func (r *EndpointsController) getConsulClient(ip string) (*api.Client, error) {
+// remoteConsulClient returns an *api.Client that points at the consul agent local to the pod.
+func (r *EndpointsController) remoteConsulClient(ip string) (*api.Client, error) {
 	newAddr := fmt.Sprintf("%s://%s:%s", r.ConsulScheme, ip, r.ConsulPort)
 	localConfig := r.ConsulClientCfg
 	localConfig.Address = newAddr
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -50,6 +50,11 @@ type EndpointsController struct {
 	Log              logr.Logger
 	Scheme           *runtime.Scheme
 	context.Context
+
+	// GetClientFunc allows us to specify how to get a consul client handle.
+	// This is used so that we can provide our own function for testing that is
+	// not dependent on having then ENV set up to pick up tokens and ca certs.
+	GetClientFunc func(string, string, string) (*api.Client, error)
 }
 
 func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
@@ -100,7 +105,7 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 
 				if hasBeenInjected(pod) {
 					// Create client for Consul agent local to the pod.
-					client, err := r.getConsulClient(pod.Status.HostIP)
+					client, err := r.GetClientFunc(r.ConsulScheme, r.ConsulPort, pod.Status.HostIP)
 					if err != nil {
 						r.Log.Error(err, "failed to create a new Consul client", "address", pod.Status.HostIP)
 						return ctrl.Result{}, err
@@ -300,7 +305,7 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 	// On each agent, we need to get services matching "k8s-service-name" and "k8s-namespace" metadata.
 	for _, pod := range list.Items {
 		// Create client for this agent.
-		client, err := r.getConsulClient(pod.Status.PodIP)
+		client, err := r.GetClientFunc(r.ConsulScheme, r.ConsulPort, pod.Status.PodIP)
 		if err != nil {
 			r.Log.Error(err, "failed to create a new Consul client", "address", pod.Status.PodIP)
 			return err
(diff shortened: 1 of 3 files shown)


connect-inject/endpoints_controller_test.go
```diff
@@ -1201,79 +1275,113 @@ func TestReconcileUpdateEndpoint(t *testing.T) {
...
 	}
-	for _, tt := range cases {
-		t.Run(tt.name, func(t *testing.T) {
-			// The agent pod needs to have the address 127.0.0.1 so when the
-			// code gets the agent pods via the label component=client, and
-			// makes requests against the agent API, it will actually hit the
-			// test server we have on localhost.
-			fakeClientPod := createPod("fake-consul-client", "127.0.0.1", false)
-			fakeClientPod.Labels = map[string]string{"component": "client", "app": "consul", "release": "consul"}
+	for _, secure := range []bool{true, false} {
```
kschoche, line 1278: This is just adding a second for loop + indents.

connect-inject/endpoints_controller.go
```diff
@@ -50,6 +50,11 @@ type EndpointsController struct {
 	Log              logr.Logger
 	Scheme           *runtime.Scheme
 	context.Context
+
+	// GetClientFunc allows us to specify how to get a consul client handle.
+	// This is used so that we can provide our own function for testing that is
+	// not dependent on having then ENV set up to pick up tokens and ca certs.
+	GetClientFunc func(string, string, string) (*api.Client, error)
```
ndhanushkodi, line 57:
```suggestion
	GetClient func(string, string, string) (*api.Client, error)
```
I think the "get" verb is enough to imply it's a function, and usually I don't see go function names have `Func` in them.

connect-inject/endpoints_controller_test.go
```diff
@@ -632,6 +701,10 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 				require.NoError(t, err)
 			}
 
+			getclientfunction := func(string, string, string) (*api.Client, error) {
```
ndhanushkodi, line 704:
```suggestion
			getTestClient := func(string, string, string) (*api.Client, error) {
```

connect-inject/endpoints_controller_test.go
```diff
@@ -643,6 +716,7 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 				DenyK8sNamespacesSet:  mapset.NewSetWith(),
 				ReleaseName:           "consul",
 				ReleaseNamespace:      "default",
+				GetClientFunc:         getclientfunction,
```
ndhanushkodi, line 719:
```suggestion
				GetClient:         getTestClient,
```

connect-inject/endpoints_controller_test.go
```diff
@@ -1201,79 +1275,113 @@ func TestReconcileUpdateEndpoint(t *testing.T) {
...
+						c.ACL.Tokens.Master = masterToken
+						c.CAFile = caFile
+						c.CertFile = certFile
+						c.KeyFile = keyFile
+					}
+					c.NodeName = nodeName
+				})
+				require.NoError(t, err)
+				defer consul.Stop()
+				consul.WaitForSerfCheck(t)
```
ndhanushkodi, line 1308: Is WaitForSerfCheck longer than WaitForLeader? If that makes more sense can we update Create/Delete tests to also use WaitForSerfCheck?

connect-inject/endpoints_controller_test.go
```diff
@@ -1201,79 +1275,113 @@ func TestReconcileUpdateEndpoint(t *testing.T) {
...
-			})
-			require.NoError(t, err)
-			require.False(t, resp.Requeue)
+				// Override the internal implementation of getConsulClient() because
+				// there is not a straight-forward way of providing the ACL token and ca info
+				// in tests through the env.
+				// Normally this is provided through the Command via `-ca-file`, etc, httpFlags
+				// which are later re-read for each new consul client in api.DefaultConfig(), but this test does
+				// not run as part of a Command and so api.DefaultConfig() would set our TLS and ACL config to empty.
+				// We can re-use this client as the test defines the fakeClientPod to be 127.0.0.1.
```
ndhanushkodi, line 1339:
```suggestion
				// We can re-use consulClient as the local agent client since the test defines the fakeClientPod to be 127.0.0.1.
```

connect-inject/endpoints_controller_test.go
```diff
@@ -1371,6 +1479,10 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
 				require.NoError(t, err)
 			}
 
+			getclientfunction := func(string, string, string) (*api.Client, error) {
```
ndhanushkodi, line 1482:
```suggestion
			getTestClient := func(string, string, string) (*api.Client, error) {
```

connect-inject/endpoints_controller_test.go
```diff
@@ -1201,79 +1275,113 @@ func TestReconcileUpdateEndpoint(t *testing.T) {
...
-			require.NoError(t, err)
-			require.False(t, resp.Requeue)
+				// Override the internal implementation of getConsulClient() because
+				// there is not a straight-forward way of providing the ACL token and ca info
+				// in tests through the env.
+				// Normally this is provided through the Command via `-ca-file`, etc, httpFlags
+				// which are later re-read for each new consul client in api.DefaultConfig(), but this test does
+				// not run as part of a Command and so api.DefaultConfig() would set our TLS and ACL config to empty.
+				// We can re-use this client as the test defines the fakeClientPod to be 127.0.0.1.
+				getclientfunction := func(string, string, string) (*api.Client, error) {
```
ndhanushkodi, line 1340:
```suggestion
				getTestClient := func(string, string, string) (*api.Client, error) {
```

connect-inject/endpoints_controller_test.go
```diff
@@ -704,7 +778,7 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 // For the register and deregister codepath, this also tests that they work when the Consul service name is different
 // from the K8s service name.
 // This test covers EndpointsController.deregisterServiceOnAllAgents when services should be selectively deregistered
-// since the map will not be nil.
+// since the map will not be nil. This test also runs each test with ACLs+TLS enabled and disabled.
```
ndhanushkodi, line 781:
```suggestion
// since the map will not be nil. This test also runs each test with ACLs+TLS enabled and disabled, since it covers all the cases where a Consul client is created.
```

ndhanushkodi (commented):
Nice work Kyle!! I just had a few naming comments and small things. I think it makes sense that just TestReconcileUpdateEndpoint has the ACL/TLS tests since they cover the cases a client is created, I just added more to the comment on TestReconcileUpdateEndpoint to reflect that. I wish there was a way to have consistency with Create/Delete so you don't need to have context about why those tests are only for Update but I can't think of another way, so it makes sense to keep it the way it is. Maybe the next reviewer will think of something :)

connect-inject/endpoints_controller_test.go
kschoche: Yeah WaitForSerfCheck waits the longest, ensures that we're able to replicate ACLs and read them back, its not totally necessary unless we're doing ACLs tho.

ndhanushkodi: Ahh I see, that makes sense why it's here but not for Create/Delete!

connect-inject/endpoints_controller_test.go
```diff
@@ -106,6 +107,74 @@ func TestHasBeenInjected(t *testing.T) {
...
+			CAFile: caFile,
+		},
+		Token: masterToken,
+	}
+	consulClient, err := api.NewClient(cfg)
+	require.NoError(t, err)
+	addr := strings.Split(consul.HTTPSAddr, ":")
+	consulPort := addr[1]
+
+	ce, _ := api.MakeConfigEntry(api.ProxyDefaults, "pd")
```
lkysow, line 146:
```suggestion
	ce, _ := api.MakeConfigEntry(api.ProxyDefaults, "global")
```

I don't know how this worked because inside processUpstreams we look for a proxy defaults with name `global`. I think somewhere when you call `ConfigEntries().Set` it's ignoring the `pd` and using `global` but may as well set it to `global` here to avoid confusion.

lkysow (commented):
I like the idea of trying to get in here and test this. I thought it was a bit confusing as implemented because there's a number of things involved in getting clients:
1. `EndpointsController.ConsulClient`
2. `EndpointsController.GetClient`

We're only mocking out one of these.

I messed around with making `ConsulClientCfg` a field:

```go
EndpointsController {
  ConsulClient *api.Client
  ConsulClientCfg *api.Config
}
```

And then having `GetConsulClient` (which should probably be renamed to something like `RemoteConsulClient` or `ExternalConsulClient`) use `r.ConsulClientCfg` when generating its client config:

```go
// GetConsulClient returns an *api.Client that points at the consul agent local to the pod.
func (r *EndpointsController) GetConsulClient(scheme, port, ip string) (*api.Client, error) {
	newAddr := fmt.Sprintf("%s://%s:%s", scheme, ip, port)
	localConfig := r.ConsulClientCfg
	localConfig.Address = newAddr

	return consul.NewClient(localConfig)
}
```

Then in the tests you can pass in `ConsulClientCfg: cfg` which in most places is already generated. Thanks to you setting the IP of the pod to `127.0.0.1`, this will then work.

I think this approach might be better because:
1. less confusion about clients and getting clients and what the difference between the two is
2. requires less work in tests since you don't need to generate a mock closure, just pass in the cfg you already needed to create the client
3. It's "less powerful" because it doesn't involve a closure and less powerful mocking is preferred due to simplicity


Not dead set on it though so interested in hearing why the current approach was taken.

connect-inject/endpoints_controller_test.go
kschoche: Good catch on this. Not sure how it worked either..

kschoche:
> Not dead set on it though so interested in hearing why the current approach was taken.
Current approach was really just taken because I couldnt think of what you suggested, I stared at the code too long after running into issues with GetConsulClient() overwriting the token!
I like this idea, and it's much cleaner, thanks for the suggestion!!

subcommand/inject-connect/command.go
```diff
@@ -418,6 +418,7 @@ func (c *Command) Run(args []string) int {
 			ReleaseName:           c.flagReleaseName,
 			ReleaseNamespace:      c.flagReleaseNamespace,
 			Context:               ctx,
+			ConsulClientCfg:       api.DefaultConfig(),
```
lkysow, line 421: I think we may need to use `cfg` here because there's a bunch of code above that is doing stuff to it after it gets it out of `api.DefaultConfig()`:

```go
	// create Consul API config object
	cfg := api.DefaultConfig()
	c.http.MergeOntoConfig(cfg)
	if cfg.TLSConfig.CAFile == "" && c.flagConsulCACert != "" {
		cfg.TLSConfig.CAFile = c.flagConsulCACert
	}
	consulURLRaw := cfg.Address
	// cfg.Address may or may not be prefixed with scheme.
	if !strings.Contains(cfg.Address, "://") {
		consulURLRaw = fmt.Sprintf("%s://%s", cfg.Scheme, cfg.Address)
	}
	consulURL, err := url.Parse(consulURLRaw)
	if err != nil {
		c.UI.Error(fmt.Sprintf("error parsing consul address %q: %s", consulURLRaw, err))
		return 1
	}

	// load CA file contents
	var consulCACert []byte
	if cfg.TLSConfig.CAFile != "" {
		var err error
		consulCACert, err = ioutil.ReadFile(cfg.TLSConfig.CAFile)
		if err != nil {
			c.UI.Error(fmt.Sprintf("error reading Consul's CA cert file %q: %s", cfg.TLSConfig.CAFile, err))
			return 1
		}
	}
```

I haven't tested this out in an acceptance test but I'm guessing we need to use that config?

lkysow (approved):
looks good, just have important question about which `cfg` we use in `command.go`.

lkysow:
Oops, I see you requested we run in acceptance test. I haven't done that so will do so as well.

subcommand/inject-connect/command.go
kschoche: Ah good catch. At runtime it worked for me deployed on a cluster as-is, but we definitely can use cfg instead here.
I believe  `api.DefaultConfig()` wouldn't have picked up flag overrides on command line, thx!

connect-inject/endpoints_controller.go
```diff
@@ -413,7 +415,7 @@ func (r *EndpointsController) processUpstreams(pod corev1.Pod) ([]api.Upstream,
 // getConsulClient returns an *api.Client that points at the consul agent local to the pod.
 func (r *EndpointsController) getConsulClient(ip string) (*api.Client, error) {
```
ndhanushkodi, line 416: I liked Luke's suggestion from [here](https://github.com/hashicorp/consul-k8s/pull/470#pullrequestreview-628110014) to name this `remoteConsulClient` or `externalConsulClient`

ndhanushkodi (approved):
I love the new approach, and this looks great to me! I ran the TestConnectInject acceptance test as well with secure:true and they pass 🎉 awesome work!!

connect-inject/endpoints_controller.go
kschoche: `remoteConsulClient` looks great, switching to that, thanks!

PR #478, opened 2021-04-08
Remove default service name annotation

ndhanushkodi:
Changes proposed in this PR:
- Remove the default service name annotation
- ~There's a check in container_init.go to make sure the annotation and the service account name match. That check has been modified so it only compares the annotation and the service account name if the annotation is set.~
- ~If the annotation is *not* set, then the error handling moves to the connect-init command in the init container. When it polls for the matching service, if the k8s-service-name metadata does not match the service account name, it will fail to set up the init container so the user will continue to see the error before the pod comes up, rather than a mismatched token being created.~
- The connect-init command uses the values of the new `service-name` and `service-account-name` flags, and if `service-name` is set (via the annotation) compares `service-account-name` to that. If `service-name` is not set, it compares `service-account-name` to the kubernetes service name.

How I've tested this PR:
- Acceptance tests, unit tests. (Will wait till [this](https://github.com/hashicorp/consul-helm/pull/897) goes green to merge)

- Code review

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -2,7 +2,6 @@ package connectinject
 
 import (
 	"bytes"
-	"fmt"
 	"strings"
 	"text/template"
 
@@ -64,25 +63,15 @@ func (h *Handler) containerInitCopyContainer() corev1.Container {
 // service, setting up the Envoy bootstrap, etc.
 func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Container, error) {
 	data := initContainerCommandData{
-		ServiceName:               pod.Annotations[annotationService],
 		AuthMethod:                h.AuthMethod,
 		ConsulNamespace:           h.consulNamespace(k8sNamespace),
 		NamespaceMirroringEnabled: h.EnableK8SNSMirroring,
 		ConsulCACert:              h.ConsulCACert,
 	}
 
-	// When ACLs are enabled, the ACL token returned from `consul login` is only
-	// valid for a service with the same name as the ServiceAccountName. If the
-	// annotation for service name is set, check against that.
-	if data.AuthMethod != "" && data.ServiceName != "" && data.ServiceName != pod.Spec.ServiceAccountName {
-		return corev1.Container{}, fmt.Errorf("serviceAccountName %q does not match service name %q", pod.Spec.ServiceAccountName, data.ServiceName)
-	}
-
-	// If the service name annotation does not exist, set the service account name flag contents for the init container
-	// command. If the init container sees the service account name flag as non-empty, it will check it against the
-	// Kubernetes service name to ensure they are the same.
-	if data.ServiceName == "" {
+	if data.AuthMethod != "" {
 		data.ServiceAccountName = pod.Spec.ServiceAccountName
+		data.ServiceName = pod.Annotations[annotationService]
 	}
 
 	// This determines how to configure the consul connect envoy command: what
@@ -183,6 +172,7 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
   {{- if .AuthMethod }}
   -acl-auth-method="{{ .AuthMethod }}" \
   -service-account-name="{{ .ServiceAccountName }}" \
+  -service-name="{{ .ServiceName }}" \
   {{- if .ConsulNamespace }}
   {{- if .NamespaceMirroringEnabled }}
   {{- /* If namespace mirroring is enabled, the auth method is
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -15,9 +15,9 @@ const (
 )
 
 type initContainerCommandData struct {
-	// todo: remove once service name annotation is removed
-	ServiceName string
-	AuthMethod  string
+	ServiceName        string
+	ServiceAccountName string
+	AuthMethod         string
 	// ConsulNamespace is the Consul namespace to register the service
 	// and proxy in. An empty string indicates namespaces are not
 	// enabled in Consul (necessary for OSS).
@@ -72,9 +72,17 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 	}
 
 	// When ACLs are enabled, the ACL token returned from `consul login` is only
-	// valid for a service with the same name as the ServiceAccountName.
-	if data.AuthMethod != "" && data.ServiceName != pod.Spec.ServiceAccountName {
+	// valid for a service with the same name as the ServiceAccountName. If the
+	// annotation for service name is set, check against that.
+	if data.AuthMethod != "" && data.ServiceName != "" && data.ServiceName != pod.Spec.ServiceAccountName {
 		return corev1.Container{}, fmt.Errorf("serviceAccountName %q does not match service name %q", pod.Spec.ServiceAccountName, data.ServiceName)
+	}
+
+	// If the service name annotation does not exist, set the service account name flag contents for the init container
+	// command. If the init container sees the service account name flag as non-empty, it will check it against the
+	// Kubernetes service name to ensure they are the same.
+	if data.ServiceName == "" {
+		data.ServiceAccountName = pod.Spec.ServiceAccountName
 	}
 
 	// This determines how to configure the consul connect envoy command: what
@@ -174,6 +182,7 @@ export CONSUL_GRPC_ADDR="${HOST_IP}:8502"
 consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
   {{- if .AuthMethod }}
   -acl-auth-method="{{ .AuthMethod }}" \
+  -service-account-name="{{ .ServiceAccountName }}" \
   {{- if .ConsulNamespace }}
   {{- if .NamespaceMirroringEnabled }}
   {{- /* If namespace mirroring is enabled, the auth method is
--- a/connect-inject/container_init_test.go
+++ b/connect-inject/container_init_test.go
@@ -43,10 +43,12 @@ func TestHandlerContainerInit(t *testing.T) {
 	}
 
 	cases := []struct {
-		Name   string
-		Pod    func(*corev1.Pod) *corev1.Pod
-		Cmd    string // Strings.Contains test
-		CmdNot string // Not contains
+		Name    string
+		Pod     func(*corev1.Pod) *corev1.Pod
+		Handler Handler
+		ExpErr  string
+		Cmd     string // Strings.Contains test
+		CmdNot  string // Not contains
 	}{
 		// The first test checks the whole template. Subsequent tests check
 		// the parts that change.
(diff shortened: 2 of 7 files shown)


connect-inject/container_init.go
```diff
@@ -72,11 +72,19 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
...
-	if data.AuthMethod != "" && data.ServiceName != pod.Spec.ServiceAccountName {
+	// valid for a service with the same name as the ServiceAccountName. If the
+	// annotation for service name is set, check against that.
+	if data.AuthMethod != "" && data.ServiceName != "" && data.ServiceName != pod.Spec.ServiceAccountName {
 		return corev1.Container{}, fmt.Errorf("serviceAccountName %q does not match service name %q", pod.Spec.ServiceAccountName, data.ServiceName)
 	}
 
+	// If the service name annotation does not exist, set the service account name flag contents for the init container
+	// command. If the init container sees the service account name flag as non-empty, it will check it against the
+	// Kubernetes service name to ensure they are the same.
```
ndhanushkodi, line 83: Note here and in the tests that if the service name annotation has *not* been set, the flag will be set as empty, which is meaningful in the connect-init command.

connect-inject/container_init.go
```diff
@@ -72,11 +72,19 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
...
+	if data.ServiceName == "" {
```
kschoche, line 84: Minor nit : I think everything will "just work" but if the user has a serviceaccount and ACLs are disabled in consul, we're filling this in for no reason. It wont end up templated into the command so it should work.
But should we guard it with `if data.AuthMethod != "" && ` for readability sake?

ndhanushkodi: That's a great point, I'll add that!

connect-inject/container_init_test.go
```diff
@@ -247,6 +325,7 @@ export CONSUL_HTTP_ADDR="${HOST_IP}:8500"
 export CONSUL_GRPC_ADDR="${HOST_IP}:8502"
 consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
   -acl-auth-method="auth-method" \
+  -service-account-name="" \
```
kschoche, line 328: scratching my head on this one a little bit.
Isn't it invalid to have auth-method and no service-account-name? Shouldnt we always pass it because we'll always know it at this point?

connect-inject/handler_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/deckarep/golang-set"
+	mapset "github.com/deckarep/golang-set"
```
kschoche, line 9:
```suggestion
	"github.com/deckarep/golang-set"
```
I'm struggling to find how to fix this, but its a setting in the IDE to disable unnecessary import aliasing

ndhanushkodi: Ooh I'll look for the setting, thanks!

connect-inject/container_init_test.go
ndhanushkodi: I needed a way to know whether to compare the service account name to the kubernetes service name or not, since we only want to compare them if the annotation is not set. I had originally done that with a separate flag like `-service-name-annotation-set` but had the idea to consolidate to one flag, just `-service-account-name` and do the check based on whether the flag had a value or not. This was essentially the way to signal from container_init.go to the connect-init command to do the check or not.

ndhanushkodi: If we always set the service account name, it will always check it against the kube service name which is undesirable if the service name annotation has been set to be different from the kube service name.

kschoche: Ah right I'm remembering now, okay that makes sense, thanks!

connect-inject/container_init.go
```diff
@@ -72,11 +72,19 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 	}
 
 	// When ACLs are enabled, the ACL token returned from `consul login` is only
-	// valid for a service with the same name as the ServiceAccountName.
-	if data.AuthMethod != "" && data.ServiceName != pod.Spec.ServiceAccountName {
+	// valid for a service with the same name as the ServiceAccountName. If the
+	// annotation for service name is set, check against that.
+	if data.AuthMethod != "" && data.ServiceName != "" && data.ServiceName != pod.Spec.ServiceAccountName {
 		return corev1.Container{}, fmt.Errorf("serviceAccountName %q does not match service name %q", pod.Spec.ServiceAccountName, data.ServiceName)
 	}
```
ishustava, line 79: I'm curious if you think it'd make sense to consolidate it and fail in both cases in the init container. That way we'd have a consistent UX for this use-case

connect-inject/container_init_test.go
```diff
@@ -68,6 +72,73 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
...
+			Handler{
+				AuthMethod: "an-auth-method",
+			},
+			"",
+			`/bin/sh -ec 
+export CONSUL_HTTP_ADDR="${HOST_IP}:8500"
+export CONSUL_GRPC_ADDR="${HOST_IP}:8502"
+consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
+  -acl-auth-method="an-auth-method" \
+`,
```
ishustava, line 97: Hmm as I'm reading this test name, I'm curious why is the `-service-account-name` not passed here?

connect-inject/handler_test.go
ishustava: These are the settings I have for imports:

Screen Shot 2021-04-09 at 12 35 35 PM

subcommand/connect-init/command.go
```diff
@@ -130,7 +133,7 @@ func (c *Command) Run(args []string) int {
 			c.flagPodNamespace)
 		serviceList, err := consulClient.Agent().ServicesWithFilter(filter)
 		if err != nil {
-			c.UI.Error(fmt.Sprintf("Unable to get Agent services: %s", err))
+			c.UI.Error(fmt.Sprintf("unable to get Agent services: %s", err))
```
ishustava, line 136: We typically capitalize output to the UI, so I think it should stay as is.

subcommand/connect-init/command.go
```diff
@@ -140,6 +143,15 @@ func (c *Command) Run(args []string) int {
 		}
 		for _, svc := range serviceList {
 			c.UI.Info(fmt.Sprintf("Registered service has been detected: %s", svc.Service))
+			// When ACLs are enabled: If the flagServiceAccountName is empty, it means the service name pod annotation
+			// was set, so the check for service account name == consul service name has already occurred in
+			// container_init.go. If the flagServiceAccountName is not empty, we need to check whether it matches the
+			// Kubernetes service name.
+			if c.flagACLAuthMethod != "" && c.flagServiceAccountName != "" && svc.Meta[connectinject.MetaKeyKubeServiceName] != c.flagServiceAccountName {
```
ishustava, line 150: In general, we should fail if the service account doesn't match the consul service account name rather than Kubernetes. But I guess in this case, consul service name == kubernetes svc name. If it didn't then we would have failed in the webhook. If you don't think we should consolidate these two cases as I've mentioned above, then I think adding a comment here would be nice.

ishustava (commented):
Looking really good, Nitya!! I left some comments and I think I'd prefer if we consolidated the two cases rather than have two ways to fail for the same reason. Let me know what you think!

connect-inject/container_init_test.go
ndhanushkodi: You are right it should have been! The tests have been changed now with the new approach you suggested

connect-inject/container_init.go
ndhanushkodi: So I've just been trying it out, and I like the idea, since it requires a lot less comments to understand and one additional flag for the connect-init command. Will modify this PR shortly with the changes

connect-inject/handler_test.go
ndhanushkodi: I'm using the same settings and it won't remove that right now :(

go.mod
```diff
@@ -29,7 +29,6 @@ require (
 	github.com/mitchellh/mapstructure v1.4.1 // indirect
 	github.com/stretchr/testify v1.6.1
 	go.uber.org/zap v1.15.0
-	golang.org/x/net v0.0.0-20201110031124-69a78807bb2b
 	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208 // indirect
```
ndhanushkodi, line 32: Ran a go mod tidy

connect-inject/container_init.go
ndhanushkodi: I went ahead and had both cases fail in the init container

connect-inject/handler_test.go
kschoche: Doh! No worries :D

kschoche (approved):
Love the new changes, this is great @ndhanushkodi!

subcommand/connect-init/command.go
```diff
@@ -140,6 +150,19 @@ func (c *Command) Run(args []string) int {
...
 		for _, svc := range serviceList {
 			c.UI.Info(fmt.Sprintf("Registered service has been detected: %s", svc.Service))
+			if c.flagACLAuthMethod != "" {
+				if c.flagServiceName != "" && c.flagServiceAccountName != c.flagServiceName {
+					// Set the error but return nil so we don't retry.
+					errServiceNameMismatch = fmt.Errorf("service account name %s doesn't match annotation service name %s", c.flagServiceAccountName, c.flagServiceName)
+					return nil
+				}
+
+				if c.flagServiceName == "" && c.flagServiceAccountName != svc.Meta[connectinject.MetaKeyKubeServiceName] {
```
ishustava, line 160: I think it'd be better to compare it with the consul service name here to make it a bit more obvious even though the two names should be equal. This error is specifically about consul service name != service account name:
```suggestion
				if c.flagServiceName == "" && c.flagServiceAccountName != svc.Name {
```

subcommand/connect-init/command.go
```diff
@@ -140,6 +150,19 @@ func (c *Command) Run(args []string) int {
...
+					// Set the error but return nil so we don't retry.
+					errServiceNameMismatch = fmt.Errorf("service account name %s doesn't match Kubernetes service name %s", c.flagServiceAccountName, svc.Meta[connectinject.MetaKeyKubeServiceName])
```
ishustava, line 162:
```suggestion
					errServiceNameMismatch = fmt.Errorf("service account name %s doesn't match Consul service name %s", c.flagServiceAccountName, svc.Name)
```

ishustava (approved):
Looks great!! Thanks so much for making the changes. A couple of minor suggestions, but they won't block the approval!

PR #479, opened 2021-04-08
Add HA support to endpoints controller

thisisnotashwin:
Changes proposed in this PR:
- Allow for leader election which will allow us to run multiple instances of the endpoints controller. When one of the instances gets deleted, another running instance grabs the lease and becomes the new leader.
- All instances will serve the webhook server but the requests will be round robin-ed between them.

How I've tested this PR:
- Created an image from it and ran the acceptance tests against it. I had to update some of the RBAC to ensure the controller has CRUD access on configmaps and also can read from leases (this was a change introduced by the update to the latest controller runtime that was made to the config-entry controller and was now made to connect-inject controller. The acceptance tests passed with the number of replicas of the connect-inject deployment set to 3.
- Manually deployed the controller with 3 instances and deleted the leader to observe another pod grab the lease and become the new leader while reconciling the endpoints.
- `ashwinvenkatesh/consul-k8s@sha256:2dc7eb5523645ecc7198212af8fa5056286ac01d9a015a5afcc5b4e95d50d695`

- Hit approve 😛


lkysow (approved):
If only all tickets were this easy!

thisisnotashwin:
> we make an acceptance test of sorts to validate it end to end?

That is a great idea. I did run the acceptance test with multiple instances running to ensure that there isnt a "breaking" change and manually checked that leaders got swapped when a pod got deleted. I guess i am on the fence about writing an acceptance test for this specifically only stems from the fact that we did not build in the leader election logic, and if someone makes a mistakes the existing tests should fail 😄

lkysow:
> Okay so silly question: I don't think we need to include unit testing for the manager's leader election itself but should we make an acceptance test of sorts to validate it end to end?

We should bump our replicas to 2 at some point. Then I think our acceptance tests as-is would be a good test.

lkysow:
If we added upgrade testing this would also be a good test since during the deploy they'll need to hand-off elections.

thisisnotashwin:
> If we added upgrade testing this would also be a good test since during the deploy they'll need to hand-off elections.

That sounds like a good idea.

PR #480, opened 2021-04-08
remove processing of pods in PodPending phase in endpoints controller

kschoche:
Changes proposed in this PR:
- remove check for PodPending in `getReadyStatusAndReason()` as pods which are in PodPending phase are not part of the endpoints object and as such we would never be reconciling a pod from an endpoint object that is in this state.

How I've tested this PR:
Per @ishustava 's excellent investigation pods which are in PodPending phase are not included in the endpoints object list so we would never need to have this check.
https://github.com/hashicorp/consul-k8s/pull/476#discussion_r610080955
Unit tests still pass

Read background in link above, code review

Checklist:
- [ ] Tests added


ndhanushkodi (approved):
🎉

thisisnotashwin (approved):

PR #481, opened 2021-04-09
Enable tproxy for the Consul service mesh

ishustava:
Changes proposed in this PR:
* add new -enable-transparent-proxy flag to inject-connect command,
  which will enable all transparent proxy mode for all applications running
  on the Consul service mesh.
* add a new pod annotation consul.hashicorp.com/transparent-proxy to allow
  enabling or disabling transparent proxy mode.
* endpoints-controller will configure proxy service registrations to enable
  transparent proxy when the above flag is passed or when tproxy annotation is set.
  It will also add service's cluster IP to both the service and proxy registrations
  as a tagged address so that Consul can configure Envoy filter traffic based on that IP.
* consul-connect-inject-init container will apply traffic redirection rules
  via consul connect redirect-traffic command which runs iptables under the hood.
  To run the command, the init container needs to both be a root user and have
  NET_ADMIN capability.

How I've tested this PR:
Ran acceptance tests in hashicorp/consul-helm#905. Note that we can't yet run enterprise acceptance tests as we're still waiting for some fixes to make it to the enterprise image

- Code review
- If you'd like you can try it out with consul-k8s image `ishustava/consul-k8s-dev:04-13-2021-65883b7` and consul image: `ishustava/consul-dev:tproxy-test`

Checklist:
- [x] Tests added

### DIFF
--- a/build-support/docker/Dev.dockerfile
+++ b/build-support/docker/Dev.dockerfile
@@ -1,2 +1,8 @@
 FROM hashicorp/consul-k8s:latest
+
+# change the user to root so we can install stuff
+USER root
+RUN apk update && apk add iptables
+USER consul-k8s
+
 COPY pkg/bin/linux_amd64/consul-k8s /bin
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -2,6 +2,7 @@ package connectinject
 
 import (
 	"bytes"
+	"strconv"
 	"strings"
 	"text/template"
 
@@ -11,6 +12,8 @@ import (
 const (
 	InjectInitCopyContainerName = "copy-consul-bin"
 	InjectInitContainerName     = "consul-connect-inject-init"
+	rootUserAndGroupID          = 0
+	netAdminCapability          = "NET_ADMIN"
 )
 
 type initContainerCommandData struct {
@@ -29,15 +32,16 @@ type initContainerCommandData struct {
 	// EnableMetrics adds a listener to Envoy where Prometheus will scrape
 	// metrics from.
 	EnableMetrics bool
-	// todo: remove once endpoints controller supports metrics
-	// PrometheusScrapeListener configures the listener on Envoy where
-	// Prometheus will scrape metrics from.
-	PrometheusScrapeListener string
 	// PrometheusScrapePath configures the path on the listener on Envoy where
 	// Prometheus will scrape metrics from.
 	PrometheusScrapePath string
 	// PrometheusBackendPort configures where the listener on Envoy will point to.
 	PrometheusBackendPort string
+
+	// EnableTransparentProxy configures this init container to run in transparent proxy mode,
+	// i.e. run consul connect redirect-traffic command and add the required privileges to the
+	// container to do that.
+	EnableTransparentProxy bool
 }
 
 // containerInitCopyContainer returns the init container spec for the copy container which places
@@ -62,11 +66,18 @@ func (h *Handler) containerInitCopyContainer() corev1.Container {
 // containerInit returns the init container spec for registering the Consul
 // service, setting up the Envoy bootstrap, etc.
 func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Container, error) {
+	// Check if tproxy is enabled on this pod.
+	tproxyEnabled, err := transparentProxyEnabled(pod, h.EnableTransparentProxy)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+
 	data := initContainerCommandData{
 		AuthMethod:                h.AuthMethod,
 		ConsulNamespace:           h.consulNamespace(k8sNamespace),
 		NamespaceMirroringEnabled: h.EnableK8SNSMirroring,
 		ConsulCACert:              h.ConsulCACert,
+		EnableTransparentProxy:    tproxyEnabled,
 	}
 
 	if data.AuthMethod != "" {
@@ -119,7 +130,7 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 		return corev1.Container{}, err
 	}
 
-	return corev1.Container{
+	container := corev1.Container{
 		Name:  InjectInitContainerName,
 		Image: h.ImageConsulK8S,
 		Env: []corev1.EnvVar{
@@ -151,7 +162,36 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 		Resources:    h.InitContainerResources,
 		VolumeMounts: volMounts,
 		Command:      []string{"/bin/sh", "-ec", buf.String()},
-	}, nil
+	}
+
+	if tproxyEnabled {
+		// Running consul connect redirect-traffic with iptables
+		// requires both being a root user and having NET_ADMIN capability.
+		container.SecurityContext = &corev1.SecurityContext{
+			RunAsUser:  pointerToInt64(rootUserAndGroupID),
+			RunAsGroup: pointerToInt64(rootUserAndGroupID),
+			Capabilities: &corev1.Capabilities{
+				Add: []corev1.Capability{netAdminCapability},
+			},
+		}
+	}
+
+	return container, nil
+}
+
+// transparentProxyEnabled returns true if transparent proxy should is enabled for this pod.
+// It returns an error when the annotation value cannot be parsed by strconv.ParseBool.
+func transparentProxyEnabled(pod corev1.Pod, globalEnabled bool) (bool, error) {
+	if raw, ok := pod.Annotations[annotationTransparentProxy]; ok {
+		return strconv.ParseBool(raw)
+	}
+
+	return globalEnabled, nil
+}
+
+// pointerToInt64 takes an int64 and returns a pointer to it.
+func pointerToInt64(i int64) *int64 {
+	return &i
 }
 
 // initContainerCommandTpl is the template for the command executed by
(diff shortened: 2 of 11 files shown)


ishustava:
Hey @whiskeysierra 

We're going to release the docs together with the next Helm, consul-k8s and consul beta release (probably this week).

build-support/docker/Dev.dockerfile
```diff
@@ -1,2 +1,8 @@
 FROM hashicorp/consul-k8s:latest
+
+# change the user to root so we can install stuff
+USER root
+RUN apk update && apk add iptables
+USER consul-k8s
```
ishustava, line 6: we should be able to remove this after we release, but for now it's useful to have for the dev images.

connect-inject/container_init_test.go
```diff
@@ -192,7 +261,6 @@ func TestHandlerContainerInit_namespacesEnabled(t *testing.T) {
 				EnableNamespaces:           true,
 				ConsulDestinationNamespace: "default",
 			},
-			k8sNamespace,
```
ishustava, line 195: this was unused

connect-inject/container_init.go
```diff
@@ -29,15 +32,16 @@ type initContainerCommandData struct {
 	// EnableMetrics adds a listener to Envoy where Prometheus will scrape
 	// metrics from.
 	EnableMetrics bool
-	// todo: remove once endpoints controller supports metrics
-	// PrometheusScrapeListener configures the listener on Envoy where
-	// Prometheus will scrape metrics from.
-	PrometheusScrapeListener string
```
ishustava, line 35: this was also unused once we added support for metrics merging to the endpoints controller

connect-inject/container_init.go
```diff
@@ -151,7 +162,36 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
...
+			Capabilities: &corev1.Capabilities{
+				Add: []corev1.Capability{netAdminCapability},
+			},
+		}
+	}
+
+	return container, nil
+}
+
+// transparentProxyEnabled returns true if transparent proxy should is enabled for this pod.
```
kschoche, line 182:
```suggestion
// transparentProxyEnabled returns true if transparent proxy should be enabled for this pod.
```

connect-inject/container_init_test.go
```diff
@@ -142,6 +141,77 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
...
+			pod := minimal()
+			if c.annotationEnabled != nil {
+				pod.Annotations[annotationTransparentProxy] = strconv.FormatBool(*c.annotationEnabled)
+			}
+
+			expectedSecurityContext := &corev1.SecurityContext{
+				RunAsUser:  pointerToInt64(0),
+				RunAsGroup: pointerToInt64(0),
+				Capabilities: &corev1.Capabilities{
+					Add: []corev1.Capability{"NET_ADMIN"},
```
kschoche, line 193: minor nit:
```suggestion
					Add: []corev1.Capability{netAdminCapability},
```
It is bugging me that there isn't a const for NET_ADMIN in a k8s library for us somewhere :-(  I get that it's a posix thing not a k8s thing but its a common field for a security context!

connect-inject/container_init_test.go
```diff
@@ -294,6 +358,105 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
...
+			Handler{
+				EnableNamespaces:           true,
+				ConsulDestinationNamespace: "default",
+				EnableTransparentProxy:     true,
+			},
+			`/bin/sh -ec 
+export CONSUL_HTTP_ADDR="${HOST_IP}:8500"
+export CONSUL_GRPC_ADDR="${HOST_IP}:8502"
+consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
+  -consul-service-namespace="default" \
```
kschoche, line 376: no changes needed here, just a general comment that its kinda reading weird now that we have `consul-service-namespace` and `namespace` used across the various connect commands now, I *think* they all mean the same thing but obv we cannot change them on the consul side.

connect-inject/endpoints_controller_test.go
```diff
@@ -804,7 +805,7 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 			})
 			require.NoError(t, err)
 			defer consul.Stop()
-			consul.WaitForSerfCheck(t)
+			consul.WaitForServiceIntentions(t)
```
kschoche, line 808: This is brilliant, I totally messed this up when I put SerfCheck instead of ServiceIntentions, this is probably also the cause of the flakey tests I've seen today! Awesome work!

subcommand/inject-connect/command.go
```diff
@@ -150,6 +153,8 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagCrossNamespaceACLPolicy, "consul-cross-namespace-acl-policy", "",
 		"[Enterprise Only] Name of the ACL policy to attach to all created Consul namespaces to allow service "+
 			"discovery across Consul namespaces. Only necessary if ACLs are enabled.")
+	c.flagSet.BoolVar(&c.flagEnableTransparentProxy, "enable-transparent-proxy", false,
```
kschoche, line 156: Is the plan to keep this disabled by default in master?

kschoche (approved):
This looks fantastic! I had a couple non-blocking questions but overall I really like the approach and the testing! great work.

connect-inject/container_init.go
```diff
@@ -202,4 +242,16 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
   {{- if .ConsulNamespace }}
   -namespace="{{ .ConsulNamespace }}" \
   {{- end }}
-  -bootstrap > /consul/connect-inject/envoy-bootstrap.yaml`
+  -bootstrap > /consul/connect-inject/envoy-bootstrap.yaml
+
+{{- if .EnableTransparentProxy }}
+
```
lkysow, line 248:
```suggestion
```

connect-inject/container_init.go
```diff
@@ -202,4 +242,16 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
...
+# Apply traffic redirection rules.
+/consul/connect-inject/consul connect redirect-traffic \
+  {{- if .ConsulNamespace }}
+  -namespace="{{ .ConsulNamespace }}" \
+  {{- end }}
+  -proxy-id="$(cat /consul/connect-inject/proxyid)" \
+  -proxy-uid=0
```
lkysow, line 255: If someone changes `rootUserAndGroupID` will this also need to be updated? If so, maybe this should be added as a templated field.

connect-inject/container_init_test.go
```diff
@@ -142,6 +141,77 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
...
+
+			if c.expectEnabled {
+				require.Equal(t, expectedSecurityContext, container.SecurityContext)
+				cmd := strings.Join(container.Command, " ")
+				require.Contains(t, cmd, expectedCmd)
+			} else {
+				require.Nil(t, container.SecurityContext)
+				cmd := strings.Join(container.Command, " ")
+				require.NotContains(t, cmd, expectedCmd)
+			}
```
lkysow, line 210: minor refactor option

```suggestion
			require.NoError(t, err)
			actualCmd := strings.Join(container.Command, " ")

			if c.expectEnabled {
				require.Equal(t, expectedSecurityContext, container.SecurityContext)
				require.Contains(t, actualCmd, expectedCmd)
			} else {
				require.Nil(t, container.SecurityContext)
				require.NotContains(t, actualCmd, expectedCmd)
			}
```

connect-inject/endpoints_controller.go
```diff
@@ -32,6 +33,7 @@ const (
 	kubernetesSuccessReasonMsg = "Kubernetes health checks passing"
 	podPendingReasonMsg        = "Pod is pending"
```
lkysow, line 34: my ide says this isn't used anymore

connect-inject/endpoints_controller.go
```diff
@@ -334,12 +340,49 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
...
+					Port:    int(servicePort.Port),
+				}
+			}
+
+			service.TaggedAddresses = taggedAddresses
+			proxyService.TaggedAddresses = taggedAddresses
+
+			proxyService.Proxy.Mode = api.ProxyModeTransparent
+		} else {
+			r.Log.Info("skipping syncing service cluster IP to Consul", "name", k8sService.Name, "ns", k8sService.Namespace, "ip", k8sService.Spec.ClusterIP)
```
lkysow, line 382: should this be debug level? Wondering if there's tons of services the you don't want this. Although I guess this only applies to headless services?

subcommand/inject-connect/command.go
```diff
@@ -88,6 +88,9 @@ type Command struct {
 	flagInitContainerMemoryLimit   string
 	flagInitContainerMemoryRequest string
 
+	// Transparent proxy flag(s).
```
lkysow, line 91: but I only see one... 😆

lkysow (approved):
Looks great. I haven't had a chance to try it out yet but won't block.

subcommand/inject-connect/command.go
ishustava: It'll default to true in the helm chart. I suppose we can default it true here as well for consistency.

connect-inject/container_init.go
ishustava: That one is only used by the init container, where is `-proxy-uid` is the user id of the Envoy proxy container. Once we don't run as root, this will need to be changed. I think we can add it to the template once we support running envoy as non-root.

Note that to support tproxy and be able to run iptables we need that init container to be root, unfortunately.

connect-inject/container_init_test.go
ishustava: great catch! thank you!

connect-inject/container_init_test.go
ishustava: Yeah, the problem is that some of these are consul cli flags and some are flags for consul-k8s commands. For consul, there is only one namespace (the consul one), whereas in consul-k8s, it could be either consul or k8s namespace and so we need to disambiguate. Not sure if there is a way around it.

connect-inject/endpoints_controller.go
ishustava: your IDE is 100% correct! great catch!

connect-inject/endpoints_controller.go
ishustava: Hmm, yeah this applies only to headless services. But maybe still this should be `debug` because it's not really useful to you since everything will still work just using pod IPs. I'll change it to debug.

ishustava: or I won't 😄 It looks like zap logger doesn't have `debug` log level?

connect-inject/container_init.go
ishustava: I'll add a comment saying that.

PR #485, opened 2021-04-14
Add support for field TransparentProxy to SvcDefaults and ProxyDefaults

thisisnotashwin:
Changes proposed in this PR:
- Add support for the new Transparent Proxy field on ServiceDefaults and ProxyDefaults. Something to note here is that currently, the state/config_entry.go always sets a zero value for the field if it is set to a `nil` value when it gets returned from Consul. Hence a lot of text fixtures needed to be updated to ensure that they explicitly expect the TransparentProxy to be set. Additionally, if the incoming CRD does not have the value set, we explicitly set the value to its zero value so that `MatchesConsul` returns true. 
- This will be fixed in Consul before the next beta release potentially and parts of the fixture can be reverted.

- Update the webhook versions from v1beta1 to v1 and v1beta1 support is deprecated in Kubernetes 1.16+

How I've tested this PR: Added some test. Successful acceptance test when the CRDs were updated in helm.

How I expect reviewers to test this PR: Code Review.

Checklist:
- [x] Tests added


api/v1alpha1/proxydefaults_types_test.go
```diff
@@ -138,6 +147,9 @@ func TestProxyDefaults_ToConsul(t *testing.T) {
 					common.SourceKey:     common.SourceValue,
 					common.DatacenterKey: "datacenter",
 				},
+				TransparentProxy: &capi.TransparentProxyConfig{
+					OutboundListenerPort: 0,
+				},
```
lkysow, line 152: This is the zero field issue?

api/v1alpha1/servicedefaults_types.go
```diff
@@ -55,32 +55,8 @@ type ServiceDefaultsSpec struct {
...
-	Path string `json:"path,omitempty"`
-
-	// LocalPathPort is the port that the service is listening on for the given path.
-	LocalPathPort int `json:"localPathPort,omitempty"`
-
-	// Protocol describes the upstream's service protocol.
-	// Valid values are "http" and "http2", defaults to "http".
-	Protocol string `json:"protocol,omitempty"`
+	// TransparentProxy controls configuration specific to proxies in transparent mode.
+	TransparentProxy *TransparentProxyConfig `json:"transparentProxy,omitempty"`
```
lkysow, line 59: nit: looks like our other config entries don't suffix `Config` for the helper structs so maybe we can just use `TransparentProxy` and swap `MeshGatewayConfig` to `MeshGateway`

api/v1alpha1/types.go
```diff
@@ -30,6 +30,37 @@ const (
 	MeshGatewayModeRemote MeshGatewayMode = "remote"
 )
 
+// ExposeConfig describes HTTP paths to expose through Envoy outside of Connect.
```
lkysow, line 33: nit: looks like the MeshGateway stuff above here isn't used at all and can be deleted

api/v1alpha1/types.go
```diff
@@ -39,8 +70,8 @@ type MeshGatewayConfig struct {
 }
 
 // toConsul returns the MeshGatewayConfig for the entry
-func (m MeshGatewayConfig) toConsul() capi.MeshGatewayConfig {
-	mode := capi.MeshGatewayMode(m.Mode)
+func (in MeshGatewayConfig) toConsul() capi.MeshGatewayConfig {
```
lkysow, line 73: 🍺

api/v1alpha1/types.go
```diff
@@ -53,14 +84,59 @@ func (m MeshGatewayConfig) toConsul() capi.MeshGatewayConfig {
...
 	modes := []string{"remote", "local", "none", ""}
-	if !sliceContains(modes, m.Mode) {
-		return field.Invalid(path.Child("mode"), m.Mode, notInSliceMessage(modes))
+	if !sliceContains(modes, in.Mode) {
+		return field.Invalid(path.Child("mode"), in.Mode, notInSliceMessage(modes))
 	}
 	return nil
 }
 
+// toConsul returns the ExposeConfig for the entry
```
lkysow, line 95: can kill these doc blocks, none of the other entries have them for `toConsul`

lkysow (approved):

api/v1alpha1/proxydefaults_types_test.go
thisisnotashwin: yup!

api/v1alpha1/servicedefaults_types.go
thisisnotashwin: Updated ExposeConfig to Expose as well!!

config/crd/bases/consul.hashicorp.com_proxydefaults.yaml
```diff
@@ -80,6 +80,13 @@ spec:
                     description: Mode is the mode that should be used for the upstream connection. One of none, local, or remote.
                     type: string
                 type: object
+              transparentProxy:
+                description: TransparentProxy controls configuration specific to proxies in transparent mode.
+                properties:
+                  outboundListenerPort:
+                    description: The port of the listener where outbound application traffic is being redirected to.
```
ndhanushkodi, line 87: Do you think we should add something like "this would not typically be set via the CRD, but this field exists to match the config-entry", or is it not worth it with the validation error?

config/crd/bases/consul.hashicorp.com_servicedefaults.yaml
```diff
@@ -83,6 +83,13 @@ spec:
               protocol:
                 description: Protocol sets the protocol of the service. This is used by Connect proxies for things like observability features and to unlock usage of the service-splitter and service-router config entries for a service.
                 type: string
+              transparentProxy:
+                description: TransparentProxy controls configuration specific to proxies in transparent mode.
+                properties:
+                  outboundListenerPort:
+                    description: The port of the listener where outbound application traffic is being redirected to.
```
ndhanushkodi, line 90: Same q as above :)

connect-inject/endpoints_controller_test.go
```diff
@@ -560,6 +560,7 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 						DestinationServiceID:   "pod1-service-created",
 						LocalServiceAddress:    "",
 						LocalServicePort:       0,
+						TransparentProxy:       &api.TransparentProxyConfig{},
```
ndhanushkodi, line 563: Are these necessary if empty?

ndhanushkodi (approved):
Looks great Ashwin!! Just had some non-blocking questions!

config/crd/bases/consul.hashicorp.com_proxydefaults.yaml
thisisnotashwin: When a user does try to explicitly set this field, a validation error is returned from the webhook 😄 
The error message returned by the webhook is: Invalid value: v1alpha1.TransparentProxy{OutboundListenerPort:<value>}: use the annotation `consul.hashicorp.com/transparent-proxy-outbound-listener-port` to configure the Outbound Listener Port

config/crd/bases/consul.hashicorp.com_servicedefaults.yaml
thisisnotashwin: same answer as above ;)

connect-inject/endpoints_controller_test.go
thisisnotashwin: Unfortunately, not in this case as it expects a pointer. An empty value would equate to a `nil` and not an empty field. this should be fixed in the future and we should be able to revert this change when Consul returns a nil value when provided a nil value.

ndhanushkodi: Ah sweet, I wasn't sure if this fell under that case, thanks for the explanation!!

config/crd/bases/consul.hashicorp.com_proxydefaults.yaml
ndhanushkodi: Yeah I had noticed that, and that should be enough :)

PR #489, opened 2021-04-15
Update the handler to use zap logger

thisisnotashwin:
Changes proposed in this PR:
- Previously the handler was using hclog, which logs the webhook logs to a different output from the controller logs. In order for them to be written to where the controller logs are, we update the handler to now use zaplogger over hclog.

How I've tested this PR: Run tests

How I expect reviewers to test this PR: Code review

Checklist:
- [ ] Tests added

### DIFF
--- a/connect-inject/handler.go
+++ b/connect-inject/handler.go
@@ -174,6 +174,8 @@ func (h *Handler) Handle(_ context.Context, req admission.Request) admission.Res
 		return admission.Allowed(fmt.Sprintf("%s %s does not require injection", pod.Kind, pod.Name))
 	}
 
+	h.Log.Info("received pod", "name", pod.Name, "ns", pod.Namespace)
+
 	// Add our volume that will be shared by the init container and
 	// the sidecar for passing data in the pod.
 	pod.Spec.Volumes = append(pod.Spec.Volumes, h.containerVolume())
--- a/connect-inject/handler.go
+++ b/connect-inject/handler.go
@@ -9,9 +9,9 @@ import (
 	"strconv"
 
 	"github.com/deckarep/golang-set"
+	"github.com/go-logr/logr"
 	"github.com/hashicorp/consul-k8s/namespaces"
 	"github.com/hashicorp/consul/api"
-	"github.com/hashicorp/go-hclog"
 	"gomodules.xyz/jsonpatch/v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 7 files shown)


subcommand/connect-init/command.go
```diff
@@ -24,7 +24,7 @@ const (
 	// The number of times to attempt ACL Login.
 	numLoginRetries = 3
 	// The number of times to attempt to read this service (60s).
-	defaultServicePollingRetries = 60
+	defaultServicePollingRetries = 120
```
ishustava, line 27: this change seems unrelated?

subcommand/inject-connect/command.go
```diff
@@ -351,8 +346,14 @@ func (c *Command) Run(args []string) int {
 	allowK8sNamespaces := flags.ToSet(c.flagAllowK8sNamespacesList)
 	denyK8sNamespaces := flags.ToSet(c.flagDenyK8sNamespacesList)
 
-	// Start the endpoints controller
-	zapLogger := zap.New(zap.UseDevMode(true), zap.Level(zapcore.InfoLevel))
+	var zapLevel zapcore.Level
+	if err := zapLevel.UnmarshalText([]byte(c.flagLogLevel)); err != nil {
+		c.UI.Error(fmt.Sprintf("Error parsing -log-level %q: %s", c.flagLogLevel, err.Error()))
+		return 1
+	}
```
ishustava, line 353: does zaplogger support the same levels as hclog?

thisisnotashwin: good catch..trace is not supported!! updating the PR!

subcommand/connect-init/command.go
thisisnotashwin: yeah..i just crammed it into this PR but I can remove it from this one and add another PR for the same

subcommand/inject-connect/command.go
thisisnotashwin: Updated the flag description to only include the levels supported by zaplogger!

connect-inject/handler.go
```diff
@@ -198,15 +198,15 @@ func (h *Handler) Handle(_ context.Context, req admission.Request) admission.Res
 	// the Envoy configuration.
 	initContainer, err := h.containerInit(pod, req.Namespace)
 	if err != nil {
```
ishustava, line 200: since we're changing it, I wonder if we should add an `Info` level log line saying the name/namespace of the pod it's mutating, maybe on line 176?

ishustava (approved):

connect-inject/handler.go
thisisnotashwin: Done!! 😺

kschoche (approved):
Excellent!

PR #493, opened 2021-04-20
use rootless containers where possible

kschoche:
Changes proposed in this PR:
- Add a security context to the init copy container and the envoy sidecar and ensure they run as non-root.
- Add `RunAsNonRoot: false` to the connt-init container so that it can run as root even if the Pod has this disabled. This should address https://github.com/hashicorp/consul-helm/issues/918. 
- Remove an unused param in the envoySidecar() function.

How I've tested this PR:
unit tests.
I've deployed this into a test cluster with a sample app that has:
  (1) no pod security contexts  
 and
 (2) pod security context : 
```
  securityContext:
    runAsUser: 1000
    runAsGroup: 3000
    fsGroup: 2000
    runAsNonRoot: true
``` 
and confirmed that the pod deploys succesfully with the changes to the init container and fails without them. 

unit tests / run it.

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## UNRELEASED
 
-FEATURES:
-* Connect: Add a security context to the init copy container and the envoy sidecar and ensure they do not run as root. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
-
 BUG FIXES:
-* Connect: Use `RunAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
+* Connect: Use `runAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
+
+BREAKING CHANGES:
+* Connect: Add a security context to the init copy container and the envoy sidecar and ensure they do not run as root. If a pod container shares the same `runAsUser` (5995) as Envoy an error is returned on scheduling. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
 
 ## 0.26.0-beta1 (April 16, 2021)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## UNRELEASED
 
 FEATURES:
-* Connect: Add a security context to the init copy container and the envoy sidecar and ensure they do not run as root. Also use `RunAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
+* Connect: Add a security context to the init copy container and the envoy sidecar and ensure they do not run as root. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
 
+BUG FIXES:
+* Connect: Use `RunAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
 
 ## 0.26.0-beta1 (April 16, 2021)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+FEATURES:
+* Connect: Add a security context to the init copy container and the envoy sidecar and ensure they do not run as root. Also use `RunAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
+
+
 ## 0.26.0-beta1 (April 16, 2021)
 
 BREAKING CHANGES:
--- a/connect-inject/container_init.go
+++ b/connect-inject/container_init.go
@@ -2,6 +2,7 @@ package connectinject
 
 import (
 	"bytes"
+	"fmt"
 	"strconv"
 	"strings"
 	"text/template"
@@ -13,6 +14,7 @@ const (
 	InjectInitCopyContainerName = "copy-consul-bin"
 	InjectInitContainerName     = "consul-connect-inject-init"
 	rootUserAndGroupID          = 0
+	envoyUserAndGroupID         = 5995
 	netAdminCapability          = "NET_ADMIN"
 )
 
@@ -37,6 +39,8 @@ type initContainerCommandData struct {
 	PrometheusScrapePath string
 	// PrometheusBackendPort configures where the listener on Envoy will point to.
 	PrometheusBackendPort string
+	// EnvoyUID is the UID that `consul connect redirect-traffic` will use when tproxy is enabled.
+	EnvoyUID string
 
 	// EnableTransparentProxy configures this init container to run in transparent proxy mode,
 	// i.e. run consul connect redirect-traffic command and add the required privileges to the
@@ -60,6 +64,12 @@ func (h *Handler) containerInitCopyContainer() corev1.Container {
 			},
 		},
 		Command: []string{"/bin/sh", "-ec", cmd},
+		SecurityContext: &corev1.SecurityContext{
+			RunAsUser:              pointerToInt64(envoyUserAndGroupID),
+			RunAsGroup:             pointerToInt64(envoyUserAndGroupID),
+			RunAsNonRoot:           pointerToBool(true),
+			ReadOnlyRootFilesystem: pointerToBool(true),
+		},
 	}
 }
 
@@ -78,6 +88,7 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 		NamespaceMirroringEnabled: h.EnableK8SNSMirroring,
 		ConsulCACert:              h.ConsulCACert,
 		EnableTransparentProxy:    tproxyEnabled,
+		EnvoyUID:                  fmt.Sprintf("%d", envoyUserAndGroupID),
 	}
 
 	if data.AuthMethod != "" {
@@ -170,6 +181,8 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 		container.SecurityContext = &corev1.SecurityContext{
 			RunAsUser:  pointerToInt64(rootUserAndGroupID),
 			RunAsGroup: pointerToInt64(rootUserAndGroupID),
+			// RunAsNonRoot overrides any setting in the Pod so that we can still run as root here as required.
+			RunAsNonRoot: pointerToBool(false),
 			Capabilities: &corev1.Capabilities{
 				Add: []corev1.Capability{netAdminCapability},
 			},
@@ -191,6 +204,11 @@ func transparentProxyEnabled(pod corev1.Pod, globalEnabled bool) (bool, error) {
 
 // pointerToInt64 takes an int64 and returns a pointer to it.
 func pointerToInt64(i int64) *int64 {
+	return &i
+}
+
+// pointerToBool takes a bool and returns a pointer to it.
+func pointerToBool(i bool) *bool {
 	return &i
 }
 
@@ -254,6 +272,6 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
   -namespace="{{ .ConsulNamespace }}" \
   {{- end }}
   -proxy-id="$(cat /consul/connect-inject/proxyid)" \
-  -proxy-uid=0
+  -proxy-uid={{ .EnvoyUID }}
 {{- end }}
 `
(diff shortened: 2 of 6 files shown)


connect-inject/envoy_sidecar.go
```diff
@@ -10,7 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
-func (h *Handler) envoySidecar(pod corev1.Pod, k8sNamespace string) (corev1.Container, error) {
```
kschoche, line 13: k8sNamespace was unused.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+FEATURES:
+* Connect: Add a security context to the init copy container and the envoy sidecar and ensure they do not run as root. Also use `RunAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
+
```
ishustava, line 5: >Also use `RunAsNonRoot: false` for connect-init's container when tproxy is enabled.

should this be recorded as a bug?

connect-inject/container_init.go
```diff
@@ -170,6 +181,8 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 		container.SecurityContext = &corev1.SecurityContext{
 			RunAsUser:  pointerToInt64(rootUserAndGroupID),
 			RunAsGroup: pointerToInt64(rootUserAndGroupID),
+			// RunAsNonRoot overrides any setting in the Pod so that we can still run as root here as required.
+			RunAsNonRoot: pointerToBool(false),
```
ishustava, line 185: 🙏

connect-inject/container_init.go
```diff
@@ -194,6 +207,11 @@ func pointerToInt64(i int64) *int64 {
 	return &i
 }
 
+// pointerToBool takes a bool and returns a pointer to it.
+func pointerToBool(i bool) *bool {
+	return &i
+}
```
ishustava, line 213:
```suggestion
func pointerToBool(b bool) *bool {
	return &b
}
```

connect-inject/container_init.go
```diff
@@ -78,6 +88,7 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 		NamespaceMirroringEnabled: h.EnableK8SNSMirroring,
 		ConsulCACert:              h.ConsulCACert,
 		EnableTransparentProxy:    tproxyEnabled,
+		EnvoyUID:                  fmt.Sprintf("%d", envoyUserAndGroupID),
```
ishustava, line 91: you could also use `strconv.Itoa(envoyUserAndGroupID)`

connect-inject/envoy_sidecar.go
ishustava: 🙏

connect-inject/envoy_sidecar.go
```diff
@@ -21,6 +21,13 @@ func (h *Handler) envoySidecar(pod corev1.Pod, k8sNamespace string) (corev1.Cont
 		return corev1.Container{}, err
 	}
 
+	if pod.Spec.SecurityContext != nil {
+		// User container and Envoy container cannot have the same UID.
+		if pod.Spec.SecurityContext.RunAsUser != nil && *pod.Spec.SecurityContext.RunAsUser == envoyUserAndGroupID {
```
ishustava, line 26: it looks we're only checking the security context of the pod spec, but we probably should also check the security context of the user container.

connect-inject/handler.go
```diff
@@ -205,8 +204,8 @@ func (h *Handler) Handle(_ context.Context, req admission.Request) admission.Res
 	}
 	pod.Spec.InitContainers = append(pod.Spec.InitContainers, initContainer)
 
-	// Add the Envoy and Consul sidecars.
-	envoySidecar, err := h.envoySidecar(pod, req.Namespace)
+	// Add the Envoy sidecar.
```
ishustava, line 207: 🙏

ishustava (commented):
Looks great!! The only thing that's blocking I think is failing when security context of the application container (not only pod) uses the same uid as Envoy. Everything else looks great!

CHANGELOG.md
kschoche: Sure! good idea!

connect-inject/envoy_sidecar.go
kschoche: Ah right, I completely forgot about this, thanks for double checking. Will add this!

connect-inject/container_init.go
```diff
@@ -37,6 +38,8 @@ type initContainerCommandData struct {
 	PrometheusScrapePath string
 	// PrometheusBackendPort configures where the listener on Envoy will point to.
 	PrometheusBackendPort string
+	// EnvoyUID is the UID that `consul connect redirect-traffic` will use when tproxy is enabled.
```
lkysow, line 41: I found this comment confusing because it says it's envoy's UID but then it says it's for the command `consul connect redirect-traffic`.

Upon further reading I see that this is what gets passed into that command's `-proxy-uid` flag so that makes sense now but I'm thinking this comment should either just say "this is envoy's linux user id" without talking about the `redirect-traffic` command *or* it should say "this is envoy's linux user id that will be passed to the -proxy-uid flag of consul connect redirect-traffic".

lkysow, line 41:
```suggestion
	// EnvoyUID is the Linux user ID that `consul connect redirect-traffic` will use when tproxy is enabled.
```

For folks that don't know what UID stands for.

connect-inject/container_init.go
```diff
@@ -78,6 +87,7 @@ func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Con
 		NamespaceMirroringEnabled: h.EnableK8SNSMirroring,
 		ConsulCACert:              h.ConsulCACert,
 		EnableTransparentProxy:    tproxyEnabled,
+		EnvoyUID:                  strconv.Itoa(envoyUserAndGroupID),
```
lkysow, line 90: I think we can pass it in as an `int` since the template will render it corrctly.

connect-inject/container_init_test.go
```diff
@@ -467,9 +468,9 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
 			container, err := h.containerInit(*tt.Pod(minimal()), k8sNamespace)
 			require.NoError(err)
 			actual := strings.Join(container.Command, " ")
-			require.Equal(actual, tt.Cmd)
+			require.Equal(tt.Cmd, actual)
 			if tt.CmdNot != "" {
-				require.NotContains(actual, tt.CmdNot)
+				require.NotContains(tt.CmdNot, actual)
```
lkysow, line 473: Looks like none of the test cases set `CmdNot` so you can remove this field altogether.

connect-inject/container_init_test.go
```diff
@@ -467,9 +468,9 @@ consul-k8s connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
 			container, err := h.containerInit(*tt.Pod(minimal()), k8sNamespace)
 			require.NoError(err)
 			actual := strings.Join(container.Command, " ")
-			require.Equal(actual, tt.Cmd)
+			require.Equal(tt.Cmd, actual)
```
lkysow, line 471: 🍺

connect-inject/envoy_sidecar.go
```diff
@@ -21,6 +21,21 @@ func (h *Handler) envoySidecar(pod corev1.Pod, k8sNamespace string) (corev1.Cont
...
+			return corev1.Container{}, fmt.Errorf("pod security context cannot have the same uid as envoy: %v", envoyUserAndGroupID)
+		}
+	}
+	// Ensure that none of the user's containers have the same UID as Envoy. At this point in injection the handler
+	// has only injected init containers so all containers defined in pod.Spec.Containers are from the user.
+	for _, c := range pod.Spec.Containers {
+		// User container and Envoy container cannot have the same UID.
+		if c.SecurityContext != nil && c.SecurityContext.RunAsUser != nil && *c.SecurityContext.RunAsUser == envoyUserAndGroupID {
+			return corev1.Container{}, fmt.Errorf("user containers cannot have the same uid as envoy: %v", envoyUserAndGroupID)
```
lkysow, line 35:
```suggestion
			return corev1.Container{}, fmt.Errorf("container %q has runAsUser set to the same uid %q as envoy which is not allowed", c.Name, envoyUserAndGroupID)
```

This error will be shown during `kubectl describe` so I think it should make sense in that context and be as helpful as possible. A user won't think of their containers as "user containers", they'll just think of them as their pod's containers. Nice to include the specific container's name too. This is just a suggestion for a possible error that meets that criteria.

connect-inject/envoy_sidecar_test.go
```diff
@@ -42,6 +43,51 @@ func TestHandlerEnvoySidecar(t *testing.T) {
...
+						RunAsUser: pointerToInt64(envoyUserAndGroupID),
+					},
+				},
+			},
+		},
+	}
+	_, err := h.envoySidecar(pod)
+	require.Error(err, fmt.Sprintf("user containers cannot have the same uid as envoy: %v", envoyUserAndGroupID))
+}
+
```
lkysow, line 90: 👍  good test coverage.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## UNRELEASED
 
+FEATURES:
+* Connect: Add a security context to the init copy container and the envoy sidecar and ensure they do not run as root. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
+
+BUG FIXES:
+* Connect: Use `RunAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
+
```
ishustava, line 8: I think we should also call out that we will error out if you have the same user as envoy as a breaking change. Unless we want to error out only when tproxy is enabled.

lkysow (commented):
Looks good. Mainly wondering about the uid for the copy container.

connect-inject/container_init.go
```diff
@@ -60,6 +63,12 @@ func (h *Handler) containerInitCopyContainer() corev1.Container {
 			},
 		},
 		Command: []string{"/bin/sh", "-ec", cmd},
+		SecurityContext: &corev1.SecurityContext{
+			RunAsUser:              pointerToInt64(envoyUserAndGroupID),
+			RunAsGroup:             pointerToInt64(envoyUserAndGroupID),
```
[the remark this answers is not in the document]
kschoche: Without setting a UID it will run as root because that is the default of the consul container.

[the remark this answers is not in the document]
lkysow: Ahh interesting. So we can pick any random uid here as long at it's not 0 and we're good?

Can you add a comment explaining that here 🙏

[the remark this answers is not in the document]
kschoche: Yup, exactly! And I just picked the same UID as envoy so I didn't have to make another constant. I'll definitely add a comment to clarify thanks for pointing this out!!
```
$ kubectl exec -it counting -c envoy-sidecar -- sh
/ $ ls -lah /consul/connect-inject/consul
-rwxr-xr-x    1 5995     5995      111.7M Apr 20 20:44 /consul/connect-inject/consul
```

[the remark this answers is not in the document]
ishustava: Maybe to avoid confusion, we should use a different user? In the RFC, we proposed `5996`.

ishustava (approved):

CHANGELOG.md
kschoche: Yeah good point. But now I have a FEATURE + BREAKING CHANGE + BUG FIX entry in the changelog against the same PR, can we consolidate this somehow?

ishustava: Haha, yes that's true. Maybe change the non-bug fix to a breaking change and add a note about what's breaking.

connect-inject/container_init_test.go
kschoche: nice catch!

CHANGELOG.md
kschoche: @ishustava  how does this look?

connect-inject/container_init_test.go
```diff
@@ -248,7 +249,6 @@ func TestHandlerContainerInit_namespacesEnabled(t *testing.T) {
 		Pod     func(*corev1.Pod) *corev1.Pod
 		Handler Handler
 		Cmd     string // Strings.Contains test
-		CmdNot  string // Not contains
```
kschoche, line 251: Unused in this test.

CHANGELOG.md
ishustava: Perfect!

lkysow (approved):

PR #494, opened 2021-04-20
Add mesh resource

thisisnotashwin:
Changes proposed in this PR:
- Add support for new Cluster resource. This resource will potentially be renamed to `Mesh` and this PR will be marked ready when that change has been made.

How I've tested this PR: Run unit tests and acceptance tests locally

How I expect reviewers to test this PR: The CI will fail as the latest released consul image does not currently have mesh support but the beta2 release will. Once an image for beta2 will be available, this CI config will be updated with it to ensure all tests pass. Review this PR assuming the CI is green!! Thank you!!

Checklist:
- [x] Tests added


api/v1alpha1/mesh_types.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// MeshSpec defines the desired state of Mesh
+type MeshSpec struct {
+	TransparentProxy TransparentProxyMeshConfig `json:"transparentProxy,omitempty"`
+}
+
+type TransparentProxyMeshConfig struct {
+	CatalogDestinationsOnly bool `json:"catalogDestinationsOnly,omitempty"`
+}
```
ishustava, line 51: Do these need docs?

api/v1alpha1/mesh_webhook_test.go
```diff
@@ -0,0 +1,101 @@
...
+				Spec: MeshSpec{
+					TransparentProxy: TransparentProxyMeshConfig{
+						CatalogDestinationsOnly: true,
+					},
+				},
+			},
+			expAllow:      false,
+			expErrMessage: "mesh resource already defined - only one mesh entry is supported",
+		},
+		"name not global": {
```
ishustava, line 56:
```suggestion
		"name not mesh": {
```

ishustava (approved):
Looks good! I think it's just missing docs and descriptions on the CRD but I'm not going to block on that.

config/crd/kustomization.yaml
```diff
@@ -9,6 +9,7 @@ resources:
 - bases/consul.hashicorp.com_serviceintentions.yaml
 - bases/consul.hashicorp.com_ingressgateways.yaml
 - bases/consul.hashicorp.com_terminatinggateways.yaml
+- bases/consul.hashicorp.com_clusters.yaml
```
ndhanushkodi, line 12: Should this be _meshes.yaml? (And the rest in this file as well?)

ndhanushkodi (approved):
Looks great, just wanted to check if the kustomization file needs updating with the new name.

api/v1alpha1/mesh_types.go
thisisnotashwin: Added :)

config/crd/kustomization.yaml
thisisnotashwin: good catch!!

PR #495, opened 2021-04-20
Allow unstructured schema for the ProxyDefaults CRD

ishustava:
Prior to `apiextensions.k8s.io/v1`, structured schema was [optional](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema) for CRD spec, but now it is required. Since `config` field can have arbitrary keys, we need to tell Kubernetes to allow that explicitly now that we've upgraded from `v1beta1` to `v1`.

Changes proposed in this PR:
- Set `x-kubernetes-preserve-unknown-fields` to `true` for the `config` field in `ProxyDefaults`

How I've tested this PR:
by running acceptance tests

code review (acceptance tests in hashicorp/consul-helm#921)

Checklist:
- [ ] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+BUG FIXES:
+* CRDs: Fix a bug where the `config` field in `ProxyDefaults` CR was not synced to Consul because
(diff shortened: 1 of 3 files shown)


thisisnotashwin (approved):
🙏 thanks so much!!

lkysow (approved):
🚀

PR #501, opened 2021-04-22
use hclog instead of UI logging for connect-init

kschoche:
Changes proposed in this PR:
- Only use UI.Error/Info/etc for startup or validation logging, then switch to use hclog to log everything else.
- Update the command tests accordingly.

How I've tested this PR:
Manual testing + updated unit tests.
```
2021-04-22T13:18:07.053-0500 [INFO]  Consul login complete
2021-04-22T13:18:07.055-0500 [INFO]  Registered service has been detected: service=counting
2021-04-22T13:18:07.055-0500 [INFO]  Registered service has been detected: service=counting-sidecar-proxy
2021-04-22T13:18:07.055-0500 [INFO]  Connect initialization completed
```

code review, unit tests

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/connect-init/command_test.go
+++ b/subcommand/connect-init/command_test.go
@@ -36,6 +36,10 @@ func TestRun_FlagValidation(t *testing.T) {
 			flags:  []string{"-pod-name", testPodName, "-pod-namespace", testPodNamespace, "-acl-auth-method", testAuthMethod},
 			expErr: "-service-account-name must be set when ACLs are enabled",
 		},
+		{
+			flags:  []string{"-pod-name", testPodName, "-pod-namespace", testPodNamespace, "-acl-auth-method", testAuthMethod, "-service-account-name", "foo", "-log-level", "invalid"},
+			expErr: "unknown log level: invalid",
+		},
 	}
 	for _, c := range cases {
 		t.Run(c.expErr, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


subcommand/connect-init/command_test.go
```diff
@@ -209,29 +217,28 @@ func TestRun_ServicePollingWithACLsAndTLS(t *testing.T) {
...
-				require.NoError(t, err)
-				require.Equal(t, "token created via login: {\"pod\":\"default-ns/counting-pod\"}", token.Description)
-
-				// Validate contents of proxyFile.
-				data, err := ioutil.ReadFile(proxyFile)
-				require.NoError(t, err)
-				require.Contains(t, string(data), "counting-counting-sidecar-proxy")
+			if test.expCode != 0 {
+				require.Equal(t, test.expCode, code)
+				return
```
thisisnotashwin, line 222: I really like this fixture. The only suggestion I have is potentially using a `bool` for whether an error is expected or not. If it is expected, assert that the return code is 1, else assert that it is 0. That can make the if check on line 220 a little cleaner maybe? 
`if test.expError {}`

thisisnotashwin (approved):
Great work Kyle!! I really like the cleanup in the tests 😄 
Have a suggestion for one of the tests which should hopefully be straightforward to implement.

subcommand/connect-init/command_test.go
kschoche: Love it, great suggestion. When I was poking around at it I was thinking that expCode here kinda looked funky but couldnt think of a better way. thanks!

ndhanushkodi (approved):
Looks good to me. I was a little on the fence about removing all testing of these error messages, but I suppose it's not that common to test logging and it may not be worth the effort.

PR #502, opened 2021-04-22
Update service defaults

thisisnotashwin:
Changes proposed in this PR:
- Add support for `Mode` and `UpstreamConfigs` to Service Defaults.

- Error out in validation if `Mode` is set as we expect users to use annotations for the same

How I've tested this PR: 

Checklist:
- [x] Tests added


kschoche (approved):
👍🏽

api/v1alpha1/servicedefaults_types.go
```diff
@@ -56,7 +58,92 @@ type ServiceDefaultsSpec struct {
...
+	// ConnectTimeoutMs is the number of milliseconds to timeout making a new
+	// connection to this upstream. Defaults to 5000 (5 seconds) if not set.
+	ConnectTimeoutMs int `json:"connectTimeoutMs,omitempty"`
+	// Limits are the set of limits that are applied to the proxy for a specific upstream of a
+	// service instance.
+	Limits *UpstreamLimits `json:"limits,omitempty"`
+	// PassiveHealthCheck configuration determines how upstream proxy instances will
+	// be monitored for removal from the load balancing pool.
+	PassiveHealthCheck *PassiveHealthCheck `json:"passiveHealthCheck,omitempty"`
+	// MeshGatewayConfig controls how Mesh Gateways are configured and used
```
ishustava, line 117:
```suggestion
	// MeshGatewayConfig controls how Mesh Gateways are configured and used.
```

api/v1alpha1/servicedefaults_types.go
```diff
@@ -153,12 +240,20 @@ func (in *ServiceDefaults) Validate(namespacesEnabled bool) error {
 	var allErrs field.ErrorList
```
ishustava, line 240: should `ToConsul` function convert UpstreamConfig?

api/v1alpha1/servicedefaults_types.go
```diff
@@ -153,12 +240,20 @@ func (in *ServiceDefaults) Validate(namespacesEnabled bool) error {
...
 	path := field.NewPath("spec")
 
+	validProtocols := []string{"tcp", "http", "http2", "grpc"}
+	if in.Spec.Protocol != "" && !sliceContains(validProtocols, in.Spec.Protocol) {
+		allErrs = append(allErrs, field.Invalid(path.Child("protocol"), in.Spec.Protocol, notInSliceMessage(validProtocols)))
+	}
```
ishustava, line 246: good catch! I didn't realize we don't validate the protocol

config/crd/bases/consul.hashicorp.com_servicedefaults.yaml
```diff
@@ -80,16 +80,133 @@ spec:
                     description: Mode is the mode that should be used for the upstream connection. One of none, local, or remote.
                     type: string
                 type: object
+              mode:
+                description: 'Mode can be one of "direct" or "transparent". "transparent" represents that inbound and outbound application traffic is being captured and redirected through the proxy. This mode does not enable the traffic redirection itself. Instead it signals Consul to configure Envoy as if traffic is already being redirected. "direct" represents that the proxy''s listeners must be dialed directly by the local application and other proxies. Note: This cannot be set using the CRD and should be set using annotations on the services that are part of the mesh.'
```
ishustava, line 84: It's weird that it added another single quote here to the `proxy's` and now it's `proxy''s` Don't know if there's a way to escape that.

api/v1alpha1/servicedefaults_types.go
ishustava: Maybe we should also add a test that sets all possible values set for service defaults and verifies that it gets written to consul.

ishustava (commented):
This looks good, Ashwin! The only thing is that I'm not sure that it works because we don't convert `UpstreamConfig` in the `ToConsul` function? Maybe having a test that sets all possible values of the service default and checks that it was written to Consul would be good as a validation since acceptance tests also don't set these values?

thisisnotashwin:
@ishustava I completely forgot about to Consul 😞 . The last few changes were all to ensure that things dont get set it Consul that it just slipped my mind. Will fix that up right away!!

api/v1alpha1/servicedefaults_types.go
thisisnotashwin: Yeah..this was definitely a massive oversight by me

config/crd/bases/consul.hashicorp.com_servicedefaults.yaml
thisisnotashwin: yeah..i poked around and couldnt really find a way to prevent that

thisisnotashwin:
I updated the acceptance test fixture locally with some upstream configs and verified that they got set

ishustava (approved):
🎉

PR #503, opened 2021-04-23
Update health check status based on the endpoints object instead of pod

ishustava:
Previously, we were updating health check status based on the pod status
of the pod that the endpoints address is pointing to. However, since the
endpoints controller is not watching pod objects, it will not always have
the most updated pod information in the cache. This results in incorrect behavior
where sometimes the pod we get will have stale information and we update the
health check status in Consul incorrectly.

This PR changes this behavior to instead get the health check status from
the endpoints object. The endpoints object already has information about ready
and not ready addresses, and it will always have updated cache since we receive
events based on updates from the Kubernetes API server to the cached Informer.

How I've tested this PR:
Acceptance test [here](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/2887/workflows/a020c8e2-16e3-40bc-bd85-dfd31bb699f4)

code review

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -115,10 +115,17 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	// Register all addresses of this Endpoints object as service instances in Consul.
 	for _, subset := range serviceEndpoints.Subsets {
-		// Do the same thing for all addresses, regardless of whether they're ready.
-		allAddresses := append(subset.Addresses, subset.NotReadyAddresses...)
+		// Combine all addresses to a map, with a value indicating whether an address is ready or not.
+		allAddresses := make(map[corev1.EndpointAddress]string)
+		for _, readyAddress := range subset.Addresses {
+			allAddresses[readyAddress] = api.HealthPassing
+		}
 
-		for _, address := range allAddresses {
+		for _, notReadyAddress := range subset.NotReadyAddresses {
+			allAddresses[notReadyAddress] = api.HealthCritical
+		}
+
+		for address, healthStatus := range allAddresses {
 			if address.TargetRef != nil && address.TargetRef.Kind == "Pod" {
 				// Get pod associated with this address.
 				var pod corev1.Pod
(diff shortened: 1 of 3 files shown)


connect-inject/endpoints_controller.go
```diff
@@ -166,13 +173,9 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
...
 					// This is required because ServiceRegister() does not update the TTL if the service already exists.
-					status, reason, err := getReadyStatusAndReason(pod)
-					if err != nil {
-						r.Log.Error(err, "failed to get status and reason from pod", "name", serviceRegistration.Name)
-						return ctrl.Result{}, err
-					}
-					r.Log.Info("updating TTL health check for service", "name", serviceRegistration.Name, "reason", reason, "status", status)
-					err = client.Agent().UpdateTTL(getConsulHealthCheckID(pod, serviceRegistration.ID), reason, status)
+					reason := getHealthCheckStatusReason(address, healthStatus)
+					r.Log.Info("updating TTL health check for service", "name", serviceRegistration.Name, "reason", reason, "status", healthStatus)
```
lkysow, line 177: nit: I don't think users care (or know) what a TTL health check is. I think it would be fine to say "updating health check status"

ishustava: yeah let me change that.

connect-inject/endpoints_controller.go
```diff
@@ -392,24 +386,23 @@ func getConsulHealthCheckID(pod corev1.Pod, serviceID string) string {
...
+	}
+
+	// Get the pod name from the address if it points to a pod,
+	// otherwise use the IP of the endpoint address in the reason message.
+	if address.TargetRef != nil && address.TargetRef.Kind == "Pod" {
+		return fmt.Sprintf("Pod \"%s/%s\" is not ready", address.TargetRef.Namespace, address.TargetRef.Name)
+	} else {
+		// If TargetRef is not a pod, the endpoints controller will ignore it and not call this function,
+		// and so we don't need to worry about the reason message in that case.
+		return ""
```
lkysow, line 404: If this is a case we shouldn't hit should this say something like "error: targetRef should not be null"?

lkysow (approved):
Nice work!

connect-inject/endpoints_controller.go
ishustava: Yeah, I was on the fence on whether to return an error here, but could certainly do so! It will look cleaner from an API perspective even though it's a private function.

lkysow: Rather than an error I was thinking you could literally return that string. But error would work too.

ndhanushkodi: An error or the string literal suggested above sounds good to me

ndhanushkodi (approved):
Looks great to me, awesome job tracking this down and I like the decision to look at the Endpoints object. Thanks for all the context in the PR as well. Nice work!!!

connect-inject/endpoints_controller.go
ishustava: I've refactored it to instead take pod name and namespace as args to the function. I think it makes more sense since we're using the address only to get the name and the namespace anyways.

PR #509, opened 2021-05-04
 Support upgrades for connect refactor

ndhanushkodi:
Before the connect refactor, service registration in Consul was managed by the lifecycle sidecar, which would re-register the service with Consul every 10s. Now, service registration is managed by Endpoints controller.

In order to support upgrades to the refactored Endpoints controller, we need Endpoints controller to NOT register or deregister any services managed by lifecycle sidecar. To do this, the annotation `consul.hashicorp.com/connect-inject-managed-by` is added to pods by the mutating webhook, so endpoints controller will ignore older services managed by lifecycle sidecar (legacy services) for service registration/deregistration. 

To make sure endpoints controller only deregisters services managed by endpoints controller, a meta key indicating the service is managed by endpoints controller is added to the Consul service registration. 

To support health checks for legacy services, the Endpoints controller will always update the healthcheck for any pod, whether it's managed by Endpoints controller or not.

How I've tested this PR:
* Helm install v0.31.1 with the following values
  * <details>
  </details>
* k apply -f static-client.yaml
  * <details>
  </details>
* k apply -f static-server.yaml
  * <details>
  </details>
* Helm install latest with the following values
  * <details>
  </details>
* Check that the services/checks still exist: 
```
k exec pod/nitya-consul-9hhn7 -- curl -s localhost:8500/v1/agent/services?ns=foo | jq .
k exec pod/nitya-consul-9hhn7 -- curl -s localhost:8500/v1/agent/checks?ns=foo | jq .
```
* Delete the pods of one of the static-client deployment
* Check that the services/checks still exist, and now the services will have metadata that they are managed by endpoints controller.
* Mark the static-server unhealthy and see that the check is updated even though it's a legacy service.

Code review and if you can, the steps above.

Checklist:
- [x] Tests added


connect-inject/endpoints_controller.go
```diff
@@ -156,30 +162,42 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
...
 					// and the connect-proxy service should come after the "main" service
 					// because its alias health check depends on the main service existing.
-					r.Log.Info("registering service with Consul", "name", serviceRegistration.Name)
-					err = client.Agent().ServiceRegister(serviceRegistration)
-					if err != nil {
-						r.Log.Error(err, "failed to register service", "name", serviceRegistration.Name)
-						return ctrl.Result{}, err
-					}
+					if managedByEndpointsController {
+						//fmt.Println("here")
```
kschoche, line 166:
```suggestion
```

connect-inject/endpoints_controller.go
```diff
@@ -156,30 +162,42 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
...
+					err = r.upsertHealthCheck(pod, client, serviceID, proxyServiceID, proxyServiceName, healthCheckID, healthStatus)
 					if err != nil {
-						r.Log.Error(err, "failed to update health check status for service", "name", serviceRegistration.Name)
+						r.Log.Error(err, "failed to update health check status for service", "name", serviceName)
 						return ctrl.Result{}, err
 					}
+					checks, _ := client.Agent().Checks()
+					fmt.Printf("\n\nCHECKS: %+v\n", checks)
+					fmt.Printf("0:%+v\n", checks["service:pod1-service-created-sidecar-proxy:1"])
+					fmt.Printf("1:%+v\n", checks["service:pod1-service-created-sidecar-proxy:2"])
```
kschoche, line 200:
```suggestion
```

connect-inject/endpoints_controller.go
```diff
@@ -156,30 +162,42 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
...
+					proxyServiceName := getProxyServiceName(pod, serviceEndpoints)
+					proxyServiceID := getProxyServiceID(pod, serviceEndpoints)
+					healthCheckID := getConsulHealthCheckID(pod, serviceID)
+					err = r.upsertHealthCheck(pod, client, serviceID, proxyServiceID, proxyServiceName, healthCheckID, healthStatus)
 					if err != nil {
-						r.Log.Error(err, "failed to update health check status for service", "name", serviceRegistration.Name)
+						r.Log.Error(err, "failed to update health check status for service", "name", serviceName)
 						return ctrl.Result{}, err
 					}
+					checks, _ := client.Agent().Checks()
```
kschoche, line 197: Assuming this one was also not meant to make it in?

connect-inject/endpoints_controller.go
```diff
@@ -210,6 +228,111 @@ func (r *EndpointsController) SetupWithManager(mgr ctrl.Manager) error {
...
+	checks, err := client.Agent().ChecksWithFilter(filter)
+	if err != nil {
+		return nil, err
+	}
+	// This will be nil (does not exist) or an actual check.
+	return checks[healthCheckID], nil
+}
+
+// registerConsulHealthCheck registers a TTL health check for the service on this Agent.
+// The Agent is local to the Pod which has a kubernetes health check.
```
kschoche, line 243: nit: the pod may not have a k8s health check, but it will have a Ready status which is what we're actually sync'ing.

connect-inject/endpoints_controller.go
```diff
@@ -210,6 +228,111 @@ func (r *EndpointsController) SetupWithManager(mgr ctrl.Manager) error {
...
+			return err
+		}
+
+		// Also update it, the reason this is separate is there is no way to set the Output field of the health check
+		// at creation time, and this is what is displayed on the UI as opposed to the Notes field.
+		err = r.updateConsulHealthCheckStatus(client, healthCheckID, status, reason)
+		if err != nil {
+			return err
+		}
+	} else if serviceCheck.Status != status {
```
kschoche, line 299: We could in theory always update, regardless of whether or not the status has changed, it is a TTL check after-all. It's a tradeoff between making an extra call and the code looking possibly cleaner?

Definitely not a blocker but I thought I'd raise the question in case anyone else thought it. I'm normally not a fan of issuing unnecessary IO ;)

connect-inject/endpoints_controller.go
ndhanushkodi: Thank you!!! Sorry I missed some cleanups!!

connect-inject/endpoints_controller.go
ndhanushkodi: I'd imagine that a large majority of the time the status doesn't change since people like to keep their apps up for a majority of the time and since it would avoid the extra call a majority of the time I'd lean towards leaving the check

connect-inject/endpoints_controller.go
ndhanushkodi: Updated this to use the phrase "readiness status", WDYT?

connect-inject/endpoints_controller.go
kschoche: Yup, however an update to the endpoints object is almost always driven by a Ready status change, but you're right that for most of the pods referenced by this endpoint object that we're reconciling it wouldnt need to be updated. i'm cool either way!

connect-inject/endpoints_controller.go
kschoche: Love it!

connect-inject/endpoints_controller.go
ndhanushkodi: Oh yes that's true! I didn't think about that. Hmm, I think I still slightly lean towards keeping the check for a reduction in calls even though its not as many as I originally imagined.

kschoche: Cool!

connect-inject/endpoints_controller_test.go
```diff
@@ -925,11 +927,321 @@ func TestReconcileUpdateEndpoint(t *testing.T) {
 		expectedProxySvcInstances  []*api.CatalogService
 		expectedAgentHealthChecks  []*api.AgentCheck
 	}{
+		// Legacy services are not managed by endpoints controller, but endpoints controller
+		// will still add/update the legacy service's health checks.
+		{
```
ndhanushkodi, line 932: To reviewers since EP controller tests are long:
The tests I added were that for a legacy service: 
- the health check is added when the pod is healthy
- the health check is added when the pod is unhealthy
- the health check is updated from healthy --> unhealthy
- the health check is updated from unhealthy --> healthy

Since when an agent is rolled it will lose the health checks, I wanted to test that they are added back for a legacy service, and that they are updated if they already exist.

connect-inject/endpoints_controller.go
```diff
@@ -210,6 +223,111 @@ func (r *EndpointsController) SetupWithManager(mgr ctrl.Manager) error {
 		).Complete(r)
 }
 
+// getServiceCheck will return the health check for this pod and service if it exists.
+func getServiceCheck(client *api.Client, healthCheckID string) (*api.AgentCheck, error) {
+	filter := fmt.Sprintf("CheckID == `%s`", healthCheckID)
```
kschoche, line 228: I'm wondering if we may encounter a bug similar to the one where we weren't using `%q` when defining a filter in `connect-init` ? I'm guessing the `'s around "%s" make it work?

kschoche (approved):
This looks great!

lkysow (commented):
Sorry @ndhanushkodi. I couldn't complete this review. I removed myself from reviewers and added the plat team back and @ishustava was assigned.

I didn't get to finish but overall my concerns are that we're intertwining the code paths for endpoints-controller managed pods and legacy pods and that I'm worried that this is going to set us up for confusion.

I was looking to see if there was a clean way at the top of the `for` loop to deal with the legacy pod and then continue the loop without carrying the `isRegisteredByEndpointsCtrl` boolean throughout the rest of the loop code. I think that someone will add something at the bottom of the loop and be unaware that the code is dealing with legacy pods.

lkysow (approved):
Approving for now to not block.

ndhanushkodi:
> Correct me if my understanding is wrong: we're not creating the check here because we're going to create it after the service register so the check also gets created for non-endpoints-ctrl pods?

Yes

>If so, why don't we create it here and then after serviceRegister if the service isn't managed by endpoints controller we create the check?

Addressed below.

> overall my concerns are that we're intertwining the code paths for endpoints-controller managed pods and legacy pods and that I'm worried that this is going to set us up for confusion.
>I was looking to see if there was a clean way at the top of the for loop to deal with the legacy pod and then continue the loop without carrying the isRegisteredByEndpointsCtrl boolean throughout the rest of the loop code. I think that someone will add something at the bottom of the loop and be unaware that the code is dealing with legacy pods.

@lkysow I'm addressing these comments here:

The reason why I made the health check registration separate from the service registration was so we could have logic along the lines of:

```
// logic to create local client for agent
if managedByEndpointsController{
// create the service registration (without the healthcheck)
// register the service
}
// idempotently upsert the healthcheck (legacy and endpoints controller managed pods)
```

I could see how a future contributor might add logic meant for endpoints-controller managed pods to the end of that, and have that apply to legacy and endpoints controller managed pods.  

One option to reduce confusion is with a comment describing that the code after the if managedByEndpointsController block applies to legacy and endpoints controller pods.  I will go ahead and do this option in this PR so when this is merged with the existing logic it's less confusing.

Then, in a future PR, we could make a refactor to have the logic look like this:

```
// logic to create local client for agent
if legacy{
// upsert healthcheck for legacy pod
} else {
// create service registration with healthcheck
// register the service
}
```

@ishustava and @kschoche I'm looking for feedback on whether you think I should go forward with the future refactor option after merging this PR with some more documentation. I'm also looking for feedback on that refactor idea.

connect-inject/endpoints_controller.go
ndhanushkodi: yeah I think so, and I didn't notice issues testing it end to end, and that codepath would be used for updating the health check which I did test :)

connect-inject/annotations.go
```diff
@@ -102,6 +108,9 @@ const (
 
 	// injected is used as the annotation value for annotationInjected.
 	injected = "injected"
+
+	// endpointsController is the value for keyManagedBy.
+	endpointsController = "consul-k8s-endpoints-controller"
```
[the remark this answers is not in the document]
ndhanushkodi: This is a great suggestion, changed this

ishustava:
@ndhanushkodi 

> Then, in a future PR, we could make a refactor to have the logic look like this:

I like the refactor idea to make it more clear.

ndhanushkodi:
Thanks @ishustava @kschoche !

PR #510, opened 2021-05-04
add namespace toggle to tproxy

kschoche:
Changes proposed in this PR:
- Adds ability to define default tproxy mode settings at the k8s namespace level.
- ~~renames tproxy annotation to "consul.hashicorp.com/transparent-proxy-default"~~

How I've tested this PR:
Unit tests, manual testing via deploying a simple counting service with no tproxy annotations set, then toggling the namespace label true/false and redeploying:
```
% kubectl describe pod counting | grep redirect
      # Apply traffic redirection rules.
      /consul/connect-inject/consul connect redirect-traffic \
% kubectl get namespaces --show-labels| grep transparent
default           Active   261d   consul.hashicorp.com/transparent-proxy=true
% kubectl delete -f svc.yaml
service "counting" deleted
serviceaccount "counting" deleted
pod "counting" deleted
% kubectl label namespace default consul.hashicorp.com/transparent-proxy=false --overwrite
namespace/default labeled
% kubectl apply -f svc.yaml
service/counting created
serviceaccount/counting created
pod/counting created
% kubectl describe pod counting | grep redirect
%
```

NOTE: https://github.com/hashicorp/consul-helm/pull/942 must be merged in order to support this.

Unit tests, manual test

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,13 @@ IMPROVEMENTS:
   * `consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs` - Comma-separated list of IPs or CIDRs to exclude.
   * `consul.hashicorp.com/transparent-proxy-exclude-uids` - Comma-separated list of Linux user IDs to exclude.
 
+* Connect: Add the ability to set default tproxy mode at namespace level via label. [[GH-501](https://github.com/hashicorp/consul-k8s/pull/510)]
+
+  Setting `consul.hashicorp.com/transparent-proxy` to `true/false` will have the following behaviour:
+  * If set as a `pod annotation` - This will define whether tproxy is enabled/disabled for the pod.
+  * If set as a `namespace label` - This will define the default behaviour for pods in this namespace which do not have their respective annotation set.
+  * If not set on either pod or namespace - The default behaviour will be defined by the `connectInject.transparentProxy.defaultEnabled` helm value.
+
 BUG FIXES:
 * Connect: Use `runAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
 * CRDs: Fix a bug where the `config` field in `ProxyDefaults` CR was not synced to Consul because
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,9 @@ IMPROVEMENTS:
 
 * Connect: Add the ability to set default tproxy mode at namespace level via label. [[GH-501](https://github.com/hashicorp/consul-k8s/pull/510)]
 
-  Setting `consul.hashicorp.com/transparent-proxy` to `true/false` will have the following behaviour:
-  * If set as a `pod annotation` - This will define whether tproxy is enabled/disabled for the pod.
-  * If set as a `namespace label` - This will define the default behaviour for pods in this namespace which do not have their respective annotation set.
-  * If not set on either pod or namespace - The default behaviour will be defined by the `connectInject.transparentProxy.defaultEnabled` helm value.
+  Setting the annotation `consul.hashicorp.com/transparent-proxy` to `true/false` will define whether tproxy is enabled/disabled for the pod.
+  * If a `namespace` has the label `consul.hashicorp.com/transparent-proxy` set to `true/false` - this will define the default behaviour for pods in this namespace which do not also have the respective annotation set.
+  * If not set on both pod and namespace - The default tproxy behaviour will be defined by the value of `-enable-transparent-proxy` to the `consul-k8s inject-connect` command.
 
 BUG FIXES:
 * Connect: Use `runAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,9 @@ IMPROVEMENTS:
 
 * Connect: Add the ability to set default tproxy mode at namespace level via label. [[GH-501](https://github.com/hashicorp/consul-k8s/pull/510)]
 
-  Setting the annotation `consul.hashicorp.com/transparent-proxy` to `true/false` will define whether tproxy is enabled/disabled for the pod.
-  * If a `namespace` has the label `consul.hashicorp.com/transparent-proxy` set to `true/false` - this will define the default behaviour for pods in this namespace which do not also have the respective annotation set.
-  * If not set on both pod and namespace - The default tproxy behaviour will be defined by the value of `-enable-transparent-proxy` to the `consul-k8s inject-connect` command.
+  * Setting the annotation `consul.hashicorp.com/transparent-proxy` to `true/false` will define whether tproxy is enabled/disabled for the pod.
+  * Setting the label `consul.hashicorp.com/transparent-proxy` to `true/false` on a namespace will define the default behavior for pods in that namespace, which do not also have the annotation set.
+  * The default tproxy behavior will be defined by the value of `-enable-transparent-proxy` flag to the `consul-k8s inject-connect` command. It can be overridden in a namespace by the the label on the namespace or for a pod using the annotation on the pod.
 
 BUG FIXES:
 * Connect: Use `runAsNonRoot: false` for connect-init's container when tproxy is enabled. [[GH-493](https://github.com/hashicorp/consul-k8s/pull/493)]
--- a/connect-inject/annotations.go
+++ b/connect-inject/annotations.go
@@ -84,9 +84,9 @@ const (
 	// annotationConsulNamespace is the Consul namespace the service is registered into.
 	annotationConsulNamespace = "consul.hashicorp.com/consul-namespace"
 
-	// annotationTransparentProxy enables or disables transparent proxy mode for a given pod.
+	// annotationTransparentProxy enables or disables default transparent proxy behaviour for a given pod or namespace.
 	// This annotation takes a boolean value (true/false).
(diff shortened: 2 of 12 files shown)


connect-inject/annotations.go
```diff
@@ -84,9 +84,9 @@ const (
 	// annotationConsulNamespace is the Consul namespace the service is registered into.
 	annotationConsulNamespace = "consul.hashicorp.com/consul-namespace"
 
-	// annotationTransparentProxy enables or disables transparent proxy mode for a given pod.
+	// annotationTransparentProxy enables or disables default transparent proxy behaviour for a given pod or namespace.
 	// This annotation takes a boolean value (true/false).
-	annotationTransparentProxy = "consul.hashicorp.com/transparent-proxy"
+	annotationTransparentProxy = "consul.hashicorp.com/transparent-proxy-default"
```
lkysow, line 89: Curious about this change. In my mind it makes sense as `transparent-proxy: true`. I'm guessing you're thinking that if I have it on a namespace *and* on the pod then the namespace annotation is only the default. But when it's on a pod it's not the default, it's the actual value that gets used.

We *could* have `transparent-proxy-default: true` for the namespace and `transparent-proxy: true` for the pod but I think this is probably too complicated. I think most users will understand `transparent-proxy: true` at the namespace level can be overridden at the pod level.

kschoche: I'm with you on this point. It was suggested as a means to make it a bit more descriptive of the behaviour at the namespace level, but I think it's creating a bit more confusion than clarity. Totally fine with moving this back to how it was!

ishustava: Yep, @lkysow you're absolutely right. This should only apply to the namespace label since you could still override the namespace setting with a pod annotation. It doesn't make sense to have `-default` for the pod annotation since there isn't anything else that could override it.

ishustava: >We could have transparent-proxy-default: true for the namespace and transparent-proxy: true for the pod but I think this is probably too complicated. I think most users will understand transparent-proxy: true at the namespace level can be overridden at the pod level.

The reason we thought it makes sense for the namespace is to make it similar the helm value (`transparentProxy.defaultEnabled`) in that the behavior is very similar between the two: one applies to the cluster defautl and the other to the namespace.

I think you're right that it might be intuitive for the users to just have `transparent-proxy` label but we wanted to make it explicit and consistent with the helm value.

lkysow: > I think you're right that it might be intuitive for the users to just have transparent-proxy label but we wanted to make it explicit and consistent with the helm value.

I think that it's more likely that users mess this up and put the wrong label on the pod or namespace. In Helm there's only one value `defaultEnabled`, not two: `enabled`, `defaultEnabled`. If we had both then I think users would mess it up there too.

I'd vote to stick with `transparent-proxy` for both namespace and pod.

connect-inject/annotations.go
```diff
@@ -84,7 +84,9 @@ const (
 	// annotationConsulNamespace is the Consul namespace the service is registered into.
 	annotationConsulNamespace = "consul.hashicorp.com/consul-namespace"
 
-	// annotationTransparentProxy enables or disables transparent proxy mode for a given pod.
+	// annotationTransparentProxy enables or disables transparent proxy for a given pod. It can also bet set as a label
```
kschoche, line 87:
```suggestion
	// annotationTransparentProxy enables or disables transparent proxy for a given pod. It can also be set as a label
```

connect-inject/container_init.go
```diff
@@ -92,16 +92,16 @@ func (h *Handler) containerInitCopyContainer() corev1.Container {
 
 // containerInit returns the init container spec for registering the Consul
 // service, setting up the Envoy bootstrap, etc.
-func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Container, error) {
+func (h *Handler) containerInit(namespace *corev1.Namespace, pod corev1.Pod) (corev1.Container, error) {
 	// Check if tproxy is enabled on this pod.
-	tproxyEnabled, err := transparentProxyEnabled(pod, h.EnableTransparentProxy)
+	tproxyEnabled, err := transparentProxyEnabled(namespace, pod, h.EnableTransparentProxy)
```
kschoche, line 97: We now pass the namespace into containerInit from the handler, this allows us to (1) re-use the `transparentProxyEnabled()` function between containerInit/endpointsController and (2) makes testing simpler so we do not need to load up a k8s clientset for every containerInit test.

CHANGELOG.md
```diff
@@ -22,6 +22,13 @@ IMPROVEMENTS:
   * `consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs` - Comma-separated list of IPs or CIDRs to exclude.
   * `consul.hashicorp.com/transparent-proxy-exclude-uids` - Comma-separated list of Linux user IDs to exclude.
 
+* Connect: Add the ability to set default tproxy mode at namespace level via label. [[GH-501](https://github.com/hashicorp/consul-k8s/pull/510)]
+
+  Setting `consul.hashicorp.com/transparent-proxy` to `true/false` will have the following behaviour:
+  * If set as a `pod annotation` - This will define whether tproxy is enabled/disabled for the pod.
+  * If set as a `namespace label` - This will define the default behaviour for pods in this namespace which do not have their respective annotation set.
+  * If not set on either pod or namespace - The default behaviour will be defined by the `connectInject.transparentProxy.defaultEnabled` helm value.
```
thisisnotashwin, line 30: Given this a change in `consul-k8s`, can you rephrase the `connectInject.transparentProxy.defaultEnabled` helm value as the equivalent flag and sub-command it gets set on in `consul-k8s`?
Additionally, I'd phrase it as annotation on the Pod and label on the Namespace. I don't think `pod annotation` and `namespace label` seem accurate. 
Conversely:
Setting the annotations `consul.hashicorp.com/transparent-proxy` to `true/false` on the pod will define whether tproxy is enabled/disabled for the pod. Maybe?

connect-inject/annotations.go
```diff
@@ -84,7 +84,9 @@ const (
 	// annotationConsulNamespace is the Consul namespace the service is registered into.
 	annotationConsulNamespace = "consul.hashicorp.com/consul-namespace"
 
-	// annotationTransparentProxy enables or disables transparent proxy mode for a given pod.
+	// annotationTransparentProxy enables or disables transparent proxy for a given pod. It can also be set as a label
+	// on a namespace to define the default behaviour for connect-injected pods which do not otherwise override this setting
+	// with their own annotation.
 	// This annotation takes a boolean value (true/false).
 	annotationTransparentProxy = "consul.hashicorp.com/transparent-proxy"
```
thisisnotashwin, line 91: can we rename this to `keyTransparentProxy`?

connect-inject/container_init.go
```diff
@@ -92,16 +92,16 @@ func (h *Handler) containerInitCopyContainer() corev1.Container {
 
 // containerInit returns the init container spec for registering the Consul
 // service, setting up the Envoy bootstrap, etc.
-func (h *Handler) containerInit(pod corev1.Pod, k8sNamespace string) (corev1.Container, error) {
+func (h *Handler) containerInit(namespace *corev1.Namespace, pod corev1.Pod) (corev1.Container, error) {
```
thisisnotashwin, line 95: Does the namespace here have to be a pointer given we are only going to read from it?

thisisnotashwin (commented):
This looks really great!! Have a few suggestions for improvements!!

connect-inject/container_init.go
kschoche: great point, no need!

connect-inject/annotations.go
kschoche: I was on the fence on this, but it makes sense, sounds good!

ndhanushkodi (approved):
Looks great Kyle!!

CHANGELOG.md
```diff
@@ -22,6 +22,12 @@ IMPROVEMENTS:
   * `consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs` - Comma-separated list of IPs or CIDRs to exclude.
   * `consul.hashicorp.com/transparent-proxy-exclude-uids` - Comma-separated list of Linux user IDs to exclude.
 
+* Connect: Add the ability to set default tproxy mode at namespace level via label. [[GH-501](https://github.com/hashicorp/consul-k8s/pull/510)]
+
+  Setting the annotation `consul.hashicorp.com/transparent-proxy` to `true/false` will define whether tproxy is enabled/disabled for the pod.
+  * If a `namespace` has the label `consul.hashicorp.com/transparent-proxy` set to `true/false` - this will define the default behaviour for pods in this namespace which do not also have the respective annotation set.
```
thisisnotashwin, line 28:
```suggestion
  * Setting the annotation `consul.hashicorp.com/transparent-proxy` to `true/false` will define whether tproxy is enabled/disabled for the pod.
  * Setting the label `consul.hashicorp.com/transparent-proxy` set to `true/false` will define the default behaviour for pods in this namespace which do not also have the respective annotation set.
```

CHANGELOG.md
```diff
@@ -22,6 +22,12 @@ IMPROVEMENTS:
...
+  * If not set on both pod and namespace - The default tproxy behaviour will be defined by the value of `-enable-transparent-proxy` to the `consul-k8s inject-connect` command.
```
thisisnotashwin, line 29:
```suggestion
  * The default tproxy behavior will be defined by the value of `-enable-transparent-proxy` flag to the `consul-k8s inject-connect` command. It can be overridden in a namespace by the the label on the namespace or for a pod using the annotation on the pod.
```

thisisnotashwin (approved):
great work bud!!

PR #511, opened 2021-05-04
Don't support syncing multiple addresses when k8s service has multiple ports

ishustava:
Previously, we set a tagged address for each Kubernetes service port,
using the "virtual-" prefix for the tagged address key. For example,
if a service port is named "tcp", then we synced it to Consul with
the tagged address key "virtual-tcp".

However, since Consul doesn't really support multiple ports on a service yet,
this is unnecessary. So instead, we will find the service's target port
that is equal to the service port we are registering with Consul
and use that as the Port for the tagged address. Otherwise,
if we can't find the target port, we will set the tagged address's port to 0
(it will still have the cluster IP set).

How I've tested this PR:
- unit tests

- code review

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -378,17 +378,20 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
 					// If the targetPortValue is the consulServicePort, then this is the service port we'll use as the tagged address.
 					if targetPortValue == int32(consulServicePort) {
 						k8sServicePort = sp.Port
+						break
 					}
 				} else if sp.TargetPort.Type == intstr.Int && sp.TargetPort.IntVal != 0 {
 					// If the target port is a non-zero int, we can compare that port directly with the Consul service port.
 					if sp.TargetPort.IntVal == int32(consulServicePort) {
 						k8sServicePort = sp.Port
+						break
 					}
 				} else {
 					// If targetPort is not specified, then the service port is used as the target port,
 					// and we can compare the service port with the Consul service port.
 					if sp.Port == int32(consulServicePort) {
 						k8sServicePort = sp.Port
+						break
 					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


connect-inject/endpoints_controller.go
```diff
@@ -32,6 +33,10 @@ const (
 	MetaKeyKubeNS              = "k8s-namespace"
 	kubernetesSuccessReasonMsg = "Kubernetes health checks passing"
 	envoyPrometheusBindAddr    = "envoy_prometheus_bind_addr"
+
+	// clusterIPTaggedAddressName is the key for the tagged address to store the service's cluster IP and service port
+	// in Consul. Note: This value should not be changed without a corresponding change in Consul.
+	// TODO: change this to a constant shared with Consul to avoid accidentally changing this.
```
ishustava, line 39: We will update it at a future time.

connect-inject/endpoints_controller.go
```diff
@@ -356,18 +362,42 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
...
+					// If the target port is a non-zero int, we can compare that port directly with the Consul service port.
+					if sp.TargetPort.IntVal == int32(consulServicePort) {
+						k8sServicePort = sp.Port
+					}
+				} else {
+					// If targetPort is not specified, then the service port is used as the target port,
+					// and we can compare the service port with the Consul service port.
+					if sp.Port == int32(consulServicePort) {
+						k8sServicePort = sp.Port
+					}
```
thisisnotashwin, line 392: As we expect only a single port to match the `consulServicePort`, can we add a `break` inside the block where `portValue == int32(consulServicePort)`? We potentially dont need to go through the other ports, plus it makes it clear that only a single spec.Port is expected to be a match vs the last one that matches.

kschoche (approved):
Looks great!

thisisnotashwin (approved):
This looks great!! have a suggestion that isnt blocking!! Nice work

connect-inject/endpoints_controller.go
ishustava: Hmm I don't think we need a break since it's it will only go into the other if blocks if one of them is false. I.e. if one of the if in the if-else evaluates to `true` then it won't evaluate others. WDYT?

ishustava: Done! great catch!

thisisnotashwin: I should have phrased this comment better!

PR #517, opened 2021-05-14
Overwrite k8s HTTP probes when transparent proxy is enabled.

ishustava:
When this is enabled, the mutating webhook will also overwrite the HTTP probes
of the application container to point them to Envoy, while saving the original
ports of the probes as annotations. The endpoints controller will set Expose
configuration for the proxy, setting the LocalPathPort to the port saved
in the annotation.

- Add a new flag -transparent-proxy-default-overwrite-probes (default to true)
  to allow overwriting k8s probes.
- Allow setting this on a per-pod basis via the "consul.hashicorp.com/transparent-proxy-overwrite-probes"
  annotation
- Add new annotations to allow for choosing custom ports for Envoy listeners to expose
  readiness and liveness probes
- Also, rename the `-enable-transparent-proxy` to `-default-enable-transparent-proxy`

How I've tested this PR:
Acceptance test and the helm PR here: hashicorp/consul-helm#953

code review

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 
 	"github.com/deckarep/golang-set"
@@ -39,6 +40,14 @@ const (
 	// in Consul. Note: This value should not be changed without a corresponding change in Consul.
 	// TODO: change this to a constant shared with Consul to avoid accidentally changing this.
 	clusterIPTaggedAddressName = "virtual"
+
+	// defaultExposedPathsListenerPortLiveness is the default port that we will use as the ListenerPort
+	// for the Expose configuration of the proxy registration for a liveness probe.
+	defaultExposedPathsListenerPortLiveness = 20300
+
+	// defaultExposedPathsListenerPortReadiness is the default port that we will use as the ListenerPort
+	// for the Expose configuration of the proxy registration for a readiness probe.
+	defaultExposedPathsListenerPortReadiness = 20301
 )
 
 type EndpointsController struct {
@@ -82,11 +91,14 @@ type EndpointsController struct {
 	// EnableTransparentProxy controls whether transparent proxy should be enabled
 	// for all proxy service registrations.
 	EnableTransparentProxy bool
+	// TProxyOverwriteProbes controls whether the endpoints controller should expose pod's HTTP probes
+	// via Envoy proxy.
+	TProxyOverwriteProbes bool
 
 	MetricsConfig MetricsConfig
 	Log           logr.Logger
-	Scheme        *runtime.Scheme
 
+	Scheme *runtime.Scheme
 	context.Context
 }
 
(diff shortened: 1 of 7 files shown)


connect-inject/handler_test.go
```diff
@@ -1189,6 +1277,146 @@ func TestShouldInject(t *testing.T) {
...
+			}
+			err := h.overwriteProbes(corev1.Namespace{}, pod)
+			require.NoError(t, err)
+			if c.readinessProbe != nil {
+				require.Equal(t, c.expReadinessPort, pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Port.IntValue())
+				if c.expOriginalReadinessPort != 0 {
+					require.Equal(t, strconv.Itoa(c.expOriginalReadinessPort), pod.Annotations[annotationOriginalReadinessProbePort])
+				}
+			}
+			if c.livenessProbe != nil && c.expLivenessPort != 0 {
```
thisisnotashwin, line 1410: qq: why does the `if` check on L1404 look different (not symmetrical) to the one on L1410? They seems equivalent right? am i missing something?

thisisnotashwin (commented):
This looks very excellent. 

I had a broad suggestion. I think the `handler.go` method `overwriteProbes()` should be able to set the desired values for the `Port` on both the probes as it should be able to read the annotations of the pod. This was the endpoint controller does not have to check the annotations to determine what the value of the port should be and it can just read it off the pod spec.

connect-inject/handler_test.go
ishustava: Turned out it wasn't need and was probably left over from another iteration of this test. Thanks so much for catching this 😄

thisisnotashwin (approved):
This PR should live in a museum!!

connect-inject/endpoints_controller.go
```diff
@@ -530,6 +542,45 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
...
 			r.Log.Info("skipping syncing service cluster IP to Consul", "name", k8sService.Name, "ns", k8sService.Namespace, "ip", k8sService.Spec.ClusterIP)
 		}
+
+		// Expose k8s probes as Envoy listeners if needed.
+		overwriteProbes, err := shouldOverwriteProbes(pod, r.TProxyOverwriteProbes)
+		if err != nil {
+			return nil, nil, err
+		}
+		if overwriteProbes {
+			if cs := pod.Spec.Containers; len(cs) > 0 {
```
ndhanushkodi, line 552: I know for service name and service port we would look at the first container name/port in the past, but those were also overrideable by configuration. In this case, there's no configuration to pick which container is the app container to rewrite the healthcheck. Is it alright to assume here that it should be the first container?

I think yes because we don't support multi-port services and because it's not a limiting requirement to the app dev that the first container is their main app since there isn't an order the containers will run in. But just wanted to voice it out loud in case you thought of something along these lines as well.

ndhanushkodi (approved):
+1 to what ashwin said :)

connect-inject/endpoints_controller.go
ishustava: Yep, that's a really good point. Thanks for bringing this up!

You're right that with a port you could still choose a port from any container, but now we will always overwrite probes for the first container. Thinking about it, I think we could make it work and use the same container to overwrite probes as the container that has the port from the provided port annotation. That will probably be the most accurate, but I'll do it a separate PR. WDYT?

ndhanushkodi: Yes I think that makes sense-- if the port annotation exists, look for the container with a port that matches the annotation, else use the first container? Separate PR totally makes sense as this PR is valuable on its own!

PR #519, opened 2021-05-18
fully process the endpoints object before returning error

kschoche:
Changes proposed in this PR:
- Allow the endpoints controller to fully process the endpoints object prior to returning error in the case where processing one of the target addresses fails.

How I've tested this PR:
unit test added which creates an address which does not have a backing pod so that when we reconcile it that address fails and then we assert that the others still got processed.

code review + unit tests
Note to reviewers: I thought about adding tests cases which cover other errors in the reconcile path aside from what is covered in the `TestReconcileCreateEndpoints()` but I'm not sure they are totally necessary to validate the code changes. I'd be happy to add additional test cases to other tests of course.

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/endpoints_controller_test.go
+++ b/connect-inject/endpoints_controller_test.go
@@ -485,6 +485,7 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 		expectedConsulSvcInstances []*api.CatalogService
 		expectedProxySvcInstances  []*api.CatalogService
 		expectedAgentHealthChecks  []*api.AgentCheck
+		expErr                     string
 	}{
 		{
 			name:          "Empty endpoints",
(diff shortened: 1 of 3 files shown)


connect-inject/endpoints_controller_test.go
```diff
@@ -689,6 +690,132 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 				},
 			},
 		},
+		{
+			// This test has 3 addresses, but only 2 are backed by pod resources. This will cause Reconcile to error
+			// on the invalid address but continue and process the other addresses. We check for error specific to
+			// pod3 being non-existant at the end, and validate the other 2 addresses have services.
```
kschoche, line 696:
```suggestion
			// pod3 being non-existant at the end, and validate the other 2 addresses have service instances.
```

thisisnotashwin (approved):
great job on this one @kschoche!!

ndhanushkodi (approved):
Yup I think the test case in the "create" test makes sense since it covers the create/update logic paths and this doesn't apply for delete. Nice work!!!

PR #521, opened 2021-05-19
Don't set security context when on OpenShift

ishustava:
* Envoy container should not have security context set when on OpenShift
  and when tproxy is disabled because we want to let OpenShift set a random
  user/group. When tproxy is enabled though, we need to still set it on OpenShift
  because we require Envoy to have a specific user for the traffic redirection to work.
* init-copy-container doesn't need to have security context set whenever running on OpenShift
  (regardless of whether tproxy is enabled or not) so that OpenShift can set its own random user/group.

How I've tested this PR:
[acceptance tests on openshift](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/3020/workflows/7784b7b3-dfea-4980-b52b-263a57f49ea8/jobs/12238)

code review

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/container_init_test.go
+++ b/connect-inject/container_init_test.go
@@ -690,10 +690,7 @@ func TestHandlerInitCopyContainer(t *testing.T) {
 
 	for _, openShiftEnabled := range openShiftEnabledCases {
 		t.Run(fmt.Sprintf("openshift enabled: %t", openShiftEnabled), func(t *testing.T) {
-			h := Handler{}
-			if openShiftEnabled {
-				h.EnableOpenShift = true
-			}
+			h := Handler{EnableOpenShift: openShiftEnabled}
 
 			container := h.initCopyContainer()
 
--- a/connect-inject/container_init_test.go
+++ b/connect-inject/container_init_test.go
@@ -1,6 +1,7 @@
 package connectinject
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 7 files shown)


connect-inject/container_init_test.go
```diff
@@ -684,19 +685,34 @@ func TestHandlerContainerInit_Resources(t *testing.T) {
...
-		ReadOnlyRootFilesystem: pointerToBool(true),
+func TestHandlerInitCopyContainer(t *testing.T) {
+	openShiftEnabledCases := []bool{false, true}
+
+	for _, openShiftEnabled := range openShiftEnabledCases {
+		t.Run(fmt.Sprintf("openshift enabled: %t", openShiftEnabled), func(t *testing.T) {
+			h := Handler{}
+			if openShiftEnabled {
+				h.EnableOpenShift = true
+			}
```
lkysow, line 696:
```suggestion
			h := Handler{
			  EnableOpenShift: h.EnableOpenShift
			}
```

connect-inject/envoy_sidecar.go
```diff
@@ -55,13 +40,40 @@ func (h *Handler) envoySidecar(pod corev1.Pod) (corev1.Container, error) {
 			},
 		},
 		Command: cmd,
-		SecurityContext: &corev1.SecurityContext{
+	}
+
+	// If not running in transparent proxy mode and in an OpenShift environment,
+	// skip setting the security context and let OpenShift set it for us.
+	// When transparent proxy is enabled, then Envoy needs to run as our specific user
+	// so that traffic redirection will work.
```
lkysow, line 48: This comment makes more sense to me to be on line 54 right above `if tproxyEnabled || !h.EnableOpenShift`

lkysow (approved):
Nice work!

connect-inject/envoy_sidecar.go
ishustava: Good point, updated!

ndhanushkodi (approved):
Looks great!!

PR #524, opened 2021-05-25
consul-connect-inject-init needs to run privileged: true when tproxy is enabled

ishustava:
On OpenShift, if we don't set this value, the container will not provisioned with proper
privileges to run iptabels commands

How I've tested this PR:
manually, on openshift

code review

Checklist:
- [x] Tests added


thisisnotashwin (approved):
masterpiece

kschoche (approved):
Brilliant!

PR #527, opened 2021-06-01
Don't register duplicate services from different k8s namespaces

ishustava:
When Consul namespaces are not enabled and we are processing a service
that already exists in Consul but with a different k8s namespace meta,
we skip service registration to avoid service name collisions.

How I've tested this PR: 
unit-tests

code review

Checklist:
- [x] Tests added

### DIFF
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -180,7 +180,7 @@ func (r *EndpointsController) SetupWithManager(mgr ctrl.Manager) error {
 		).Complete(r)
 }
 
-// registerServicesAndHealthCheck create Consul registrations for the service and proxy and register them with Consul.
+// registerServicesAndHealthCheck creates Consul registrations for the service and proxy and register them with Consul.
 // It also upserts a Kubernetes health check for the service based on whether the endpoint address is ready.
 func (r *EndpointsController) registerServicesAndHealthCheck(ctx context.Context, serviceEndpoints corev1.Endpoints, address corev1.EndpointAddress, healthStatus string, endpointAddressMap map[string]bool) error {
 	// Get pod associated with this address.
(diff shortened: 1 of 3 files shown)


connect-inject/endpoints_controller.go
```diff
@@ -247,6 +180,103 @@ func (r *EndpointsController) SetupWithManager(mgr ctrl.Manager) error {
 		).Complete(r)
 }
 
+// registerServicesAndHealthCheck create Consul registrations for the service and proxy and register them with Consul.
```
kschoche, line 183: Minor nit :
```suggestion
// registerServicesAndHealthCheck will create Consul registrations for the service and proxy and register them with Consul.
```

This is completely a nonblocking comment but I was wondering about the name of the function in that we're not "registering" the health check (actually we are in the case it doesn't exist) but rather its an Upsert process, so maybe we could call it `registerServicesAndUpsertHealthCheck()` or `registerServicesAndUpdateHealthCheck` or something?
I think it is fine if we dont make a change but wondered if you had given any minor thought to it?

kschoche (approved):
Most excellent work!
I had a couple non-blocking questions/comments, but it looks great!

connect-inject/endpoints_controller.go
ishustava: I think it's OK to refer to both as `register`. Looking at Consul API for [health checks](https://www.consul.io/api-docs/agent/check#register-check), it seems like upserting a health check is the same as registering a health check, so I think this name is fine.

There is a distinction in that we don't always register health check and just update the health status, but I don't think this needs to be reflected in the name of the function necessarily.

kschoche: Sweet, thanks! Yeah I was on the fence about even mentioning it but thought I'd ask if you'd thought about it. Sounds good!

ishustava: I'm glad you brought it up!

ndhanushkodi (approved):
🎉

PR #529, opened 2021-06-08
Update type for timeouts from time.Duration to metav1.Duration

thisisnotashwin:
- This allows the values of these timeouts to be set as durations via the CRD allowing timeouts to be set as '5s' instead of an integer that time.Duration is resolved as.

How I've tested this PR:

Checklist:
- [ x ] Tests added


lkysow (approved):
What's the upgrade path look like for folks that were setting this using an integer? Should we document that in the changelog?

thisisnotashwin:
> What's the upgrade path look like for folks that were setting this using an integer? Should we document that in the changelog?

An integer is a valid duration. 5 translates to 5ns. So their existing resources will continue to be valid (and _i think_ continue to represent the same duration as well)

thisisnotashwin:
Ill update the changelog with the same for clarity.

PR #530, opened 2021-06-10
Set owner reference to secrets created by webhook cert manager

thisisnotashwin:
Changes proposed in this PR:

When the certificate secret is created or updated, set an OwnerReference on the secret as the `webhook-cert-manager` deployment. This ensures that deletion of the deployment will also delete the secrets. This addresses the race condition bug that we sometimes see when re-installing consul on a cluster that had a consul deleted from it. This was because the helm delete would not delete the existing secrets with certificates. When the controller would get created with a new installation, it would mount the existing secret (which was stale) and the secret on disk would get rotated before the cert watcher started which would lead to the controller using certificates signed by a CA different from the CA bundle on the MWC which would lead to x509 errors.

This change would ensure the secrets get deleted every single time and hence, a new secret would always get created during a helm install. This also ensure an existing secret, when updated is updated with the owner ref ensuring helm upgrades or installs to a cluster with an existing secret give people the desired behavior as well.

How I've tested this PR: https://github.com/hashicorp/consul-helm/pull/987

How I expect reviewers to test this PR: Code review

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/webhook-cert-manager/command.go
+++ b/subcommand/webhook-cert-manager/command.go
@@ -44,6 +44,9 @@ type Command struct {
 	flagConfigFile string
 	flagLogLevel   string
 
+	flagDeploymentName      string
+	flagDeploymentNamespace string
+
 	clientset kubernetes.Interface
 
 	once   sync.Once
@@ -59,6 +62,10 @@ func (c *Command) init() {
 	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
 	c.flagSet.StringVar(&c.flagConfigFile, "config-file", "",
 		"Path to a config file to read webhook configs from. This file must be in JSON format.")
+	c.flagSet.StringVar(&c.flagDeploymentName, "deployment-name", "",
+		"Name of deployment that the cert-manager pod is managed by.")
+	c.flagSet.StringVar(&c.flagDeploymentNamespace, "deployment-namespace", "",
+		"Namespace of deployment that the cert-manager pod is managed by.")
 	c.flagSet.StringVar(&c.flagLogLevel, "log-level", "info",
 		"Log verbosity level. Supported values (in order of detail) are \"trace\", "+
 			"\"debug\", \"info\", \"warn\", and \"error\".")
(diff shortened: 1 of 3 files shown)


subcommand/webhook-cert-manager/command.go
```diff
@@ -59,6 +62,10 @@ func (c *Command) init() {
 	c.flagSet = flag.NewFlagSet("", flag.ContinueOnError)
 	c.flagSet.StringVar(&c.flagConfigFile, "config-file", "",
 		"Path to a config file to read webhook configs from. This file must be in JSON format.")
+	c.flagSet.StringVar(&c.flagDeploymentName, "deployment-name", "",
+		"Name of deployment that is the owner of the secret")
```
lkysow, line 66: Not clear what "the secret" is. I think we can just say name of the deployment this pod is running in. What we do with that is an impl detail

subcommand/webhook-cert-manager/command.go
```diff
@@ -251,6 +281,14 @@ func (c *Command) reconcileCertificates(ctx context.Context, clientset kubernete
 
 	certSecret.Data[corev1.TLSCertKey] = bundle.Cert
 	certSecret.Data[corev1.TLSPrivateKeyKey] = bundle.Key
+	certSecret.OwnerReferences = []metav1.OwnerReference{
```
lkysow, line 284: add a comment that this is here to update existing secrets that were created before we added ownerReference?

subcommand/webhook-cert-manager/command.go
thisisnotashwin: done

subcommand/webhook-cert-manager/command.go
thisisnotashwin: done

lkysow (approved):
LGTM. Did you find out what happens if the cert-manager deployment is deleted?

thisisnotashwin:
> LGTM. Did you find out what happens if the cert-manager deployment is deleted?

oh oh I did!! i forgot to put it in here. you were right. the pods just keep them in memory and continue to work. it only becomes problematic if the cert expires or the pod has to restart.

kschoche (approved):
🧀

PR #531, opened 2021-06-14
Add a flag for federation to server-acl-init

kschoche:
Changes proposed in this PR:
- Add new flag `-federation` which will be set and will drive creation of ACL tokens. This is primarily to help with HCP use cases.
- Adds a new field called `primary` to the command which is read from the server configuration and controls whether or not we create the anonymous token and whether or not we suffix the policy names with a datacenter.
Note: There is a second PR which will be created shortly that adds the flag in consul-helm.

How I've tested this PR:
Federated acceptance tests passing.
Unit tests passing.

Code review + unit tests.

Checklist:
- [x] Tests added


subcommand/server-acl-init/command.go
```diff
@@ -743,45 +752,51 @@ func (c *Command) consulDatacenter(client *api.Client) (string, error) {
...
 	dc, ok := agentCfg["Config"]["Datacenter"].(string)
 	if !ok {
-		return "", fmt.Errorf("could not cast Config.Datacenter as string: %s", agentCfg)
+		return "", "", fmt.Errorf("could not cast Config.Datacenter as string: %s", agentCfg)
 	}
 	if dc == "" {
-		return "", fmt.Errorf("value of Config.Datacenter was empty string: %s", agentCfg)
+		return "", "", fmt.Errorf("value of Config.Datacenter was empty string: %s", agentCfg)
+	}
+	primaryDC, ok := agentCfg["DebugConfig"]["PrimaryDatacenter"].(string)
```
kschoche, line 770: In the future we will have `agentCfg["Config"]["PrimaryDatacenter"]` but since it doesnt exist going to fall back on the `DebugConfig` and address it again in the future when we have an overlapping release with consul.

thisisnotashwin:
@kschoche out of curiosity, has this been tested with the Consul cloud team? Just wanted to validate that this solution solves the problems they were running into

subcommand/server-acl-init/command.go
thisisnotashwin: Can we swap this over with the consul `1.10-rc` release?

kschoche: It's not in yet afaik, not sure when it will be in tbh.
But it definitely will not work with older consul versions at that point. We could do some sort of attempt on both Debug/Config which is doable.

subcommand/server-acl-init/command.go
```diff
@@ -732,9 +741,9 @@ func (c *Command) withPrefix(resource string) string {
 	return fmt.Sprintf("%s-%s", c.flagResourcePrefix, resource)
 }
 
-// consulDatacenter returns the current datacenter name using the
+// consulDatacenter returns the current datacenter name and the primary datacenter using the
 // /agent/self API endpoint.
-func (c *Command) consulDatacenter(client *api.Client) (string, error) {
+func (c *Command) consulDatacenter(client *api.Client) (string, string, error) {
```
thisisnotashwin, line 746: Can we rename this to be a little more detailed about the datacenter being returned? eg `consulDatacenterList()` potentially?

thisisnotashwin (approved):
This looks good!! Have a comment that isnt blocking! Happy to have this merged once we have verified that is fixes the Consul Cloud federation problem.

subcommand/server-acl-init/command.go
lkysow: This should try `Config` and if it's not there, fall back to `DebugConfig`. Then we'll be safe if it's removed from `DebugConfig`.

lkysow (approved):
Good to merge. We can fix my comments in a later PR.

subcommand/server-acl-init/command.go
```diff
@@ -781,6 +797,8 @@ func (c *Command) createAnonymousPolicy() bool {
 			// on cross-dc API calls. The cross-dc API calls thus use the anonymous
 			// token. Cross-dc API calls are needed by the Connect proxies to talk
 			// cross-dc.
+			(c.flagCreateInjectToken && c.flagFederation) ||
+			// OR to maintain back compat
 			(c.flagCreateInjectToken && c.flagCreateACLReplicationToken))
```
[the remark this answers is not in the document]
lkysow: outdated.

PR #533, opened 2021-06-14
Add support for TransparentProxy.DialedDirectly

lkysow:
* renamed `types.go` => `shared_type.go` to better reflect what that
file contains.
* Added `DialedDirectly` field to `TransparentProxy` struct
* Renamed `CatalogDestinationsOnly` to `MeshDestinationsOnly` to match
Consul renaming (https://github.com/hashicorp/consul/pull/10397)
* Added tests to validate for ProxyDefaults since that config entry now
has more fields. The validate methods were technically tested through
ServiceDefaults tests but I added them to ProxyDefaults since they both
rely on that shared validation.
* Fixed validation for OutboundListenerPort so that it returned the port
as the error rather than the entire TransparentProxy struct.
* Updated Consul api package to latest
* Updated controller-runtime to 0.6.0 because I was getting an error on
0.5.0.
* Regenerated YAML CRDs

How I've tested this PR:

Checklist:
- [x] Tests added


lkysow (commented):
This will work once Consul is released that has the `MeshServicesOnly` field name changes.

thisisnotashwin (approved):
NICE!

thisisnotashwin:
I did a `go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.0` in the project folder and it bumped up some of the dependencies in the `go.mod` so i checked them in. I can revert those changes if we don't want to update some of those dependencies.

PR #545, opened 2021-06-24
Enable custom consul-client ports (HTTP, HTTPS & GRPC) for connect-inject

### LINKED ISSUE #544: Make Consul client ports customizable for connect-inject init container
#### Feature Description

This feature request is related to what is described in the [consul-helm issue #997](https://github.com/hashicorp/consul-helm/issues/997). Basically in certain environments `hostPorts` are not available and thus the currently only feasible replacement are `nodePorts`. By using `topologyKeys` (K8s < 1.21) or `Topology Aware Hints` (K8s 1.21+) the `nodePort` loadbalancing can be prevent in order to enforce that the inject sidecars establish connectivity to their node-local Consul client.

In order to avoid having to extend the `service-node-port-range` to include the `85XY` ports, the hardcoded Consul client ports in consul-k8s should be customizable.

#### Use Case(s)

* Environments where `hostPorts` are not available, make it possible to evade to `nodePorts`
* Provide more flexibility by making the hardcoded ports customizable

#### Contributions

We (@flavioaiello, @pree & @karras) are already working on a PR.

karras:
Changes proposed in this PR:
- This PR makes it possible to customize the hardcoded Consul client ports for the connect-inject init container by providing optional parameters. If the parameters are not specified the default Consul client ports are used.
- As the parameters are optional, backwards compatibility should be guaranteed, i.e. we'll open another PR for consul-helm to allow configuring these new port parameters.
- See https://github.com/hashicorp/consul-helm/pull/1004 for the consul-helm counterpart
- Fixes #544 

How I've tested this PR:
- Built a consul-k8s container image based on this PR and uploaded it to Dockerhub. Then adjusted the image flag in consul-helm to deploy it accordingly. Finally tested the inject workflow by deploying a simple NGINX hello world app and scaling the replicas to a higher number. All Pods were correctly registered.
- Do the same as above but this time adjust the parameters and check in the connect-inject deployment if the parameters are there. When deploying an example NGINX app check the logs if they now try to connect to the Consul client with the specified ports (this will of course fail unless the client ports were adjusted as well).

- Deploy connect-inject based on this PR, check if everything is still working
- Deploy connect-inject based on this PR, then configure the new parameters but with default ports and check if everything is still working
 - Deploy connect-inject based on this PR, then configure the new parameters but with custom ports and check if everything is still working

Checklist:
- [ ] Tests added
- [n/a] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

### DIFF
--- a/subcommand/inject-connect/command.go
+++ b/subcommand/inject-connect/command.go
@@ -94,6 +94,11 @@ type Command struct {
 
 	flagEnableOpenShift bool
 
+	// Consul client ports
+	flagConsulClientPortHTTP  string
+	flagConsulClientPortHTTPS string
+	flagConsulClientPortGRPC  string
+
 	flagSet *flag.FlagSet
 	http    *flags.HTTPFlags
 
@@ -162,6 +167,12 @@ func (c *Command) init() {
 		"Overwrite Kubernetes probes to point to Envoy by default when in Transparent Proxy mode.")
 	c.flagSet.BoolVar(&c.flagEnableOpenShift, "enable-openshift", false,
 		"Indicates that the command runs in an OpenShift cluster.")
+	c.flagSet.StringVar(&c.flagConsulClientPortHTTP, "consul-client-port-http", "8500",
+		"Consul client HTTP port, defaults to 8500.")
+	c.flagSet.StringVar(&c.flagConsulClientPortHTTPS, "consul-client-port-https", "8501",
+		"Consul client HTTPS port, defaults to 8501.")
+	c.flagSet.StringVar(&c.flagConsulClientPortGRPC, "consul-client-port-grpc", "8502",
+		"Consul client GRPC port, defaults to 8502.")
 	c.flagSet.StringVar(&c.flagLogLevel, "log-level", zapcore.InfoLevel.String(),
 		fmt.Sprintf("Log verbosity level. Supported values (in order of detail) are "+
 			"%q, %q, %q, and %q.", zapcore.DebugLevel.String(), zapcore.InfoLevel.String(), zapcore.WarnLevel.String(), zapcore.ErrorLevel.String()))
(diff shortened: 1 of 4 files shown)


karras:
@pree is currently working on the unit tests, then the PR will be ready for review.

karras:
@david-yu This PR is ready as well. Meaning all PRs are now complete for review. Thanks!

subcommand/inject-connect/command.go
```diff
@@ -162,6 +167,12 @@ func (c *Command) init() {
 		"Overwrite Kubernetes probes to point to Envoy by default when in Transparent Proxy mode.")
 	c.flagSet.BoolVar(&c.flagEnableOpenShift, "enable-openshift", false,
 		"Indicates that the command runs in an OpenShift cluster.")
+	c.flagSet.StringVar(&c.flagConsulClientPortHTTP, "consul-client-port-http", "8500",
+		"Consul client HTTP port, defaults to 8500.")
+	c.flagSet.StringVar(&c.flagConsulClientPortHTTPS, "consul-client-port-https", "8501",
+		"Consul client HTTPS port, defaults to 8501.")
```
ishustava, line 173: You might want to re-use the fact that we're setting CONSUL_HTTP_ADDR as env variable for the deployment and parsing the URL to get the port here: https://github.com/hashicorp/consul-k8s/blob/d49e91985755fe71b123c933619bc26a4c0e16ec/subcommand/inject-connect/command.go#L313-L327

ishustava (commented):
This mostly looks good. A couple of comments/questions.

I left a comment inline about re-using existing code that parses http/https ports into a URL. I don't think it's a blocker but might make the code easier depending on the outcome of my second question below.

My other comment/question is about making changes to the endpoints controller. I think as written, the endpoints controller will continue to use [host IP](https://github.com/hashicorp/consul-k8s/blob/d49e91985755fe71b123c933619bc26a4c0e16ec/connect-inject/endpoints_controller.go#L204) and the port that is being parsed from the env variable set [here](https://github.com/hashicorp/consul-helm/blob/b1b21ddf9e044be557004f1674a3cd6516ec5ccc/templates/connect-inject-deployment.yaml#L73-L78). ~~I don't think hashicorp/consul-helm#1004 changes that and so I'm pretty sure this will fail since endpoints controller will not be able to register any services.~~ I saw that you do change it in hashicorp/consul-helm#999, see my comment below!

ishustava:
>I think as written, the endpoints controller will continue to use host IP and the port that is being parsed from the env variable set here.

Thinking a bit more about this. I don't think you can change it to be <hostIP>:<nodePort> because that, with node-local topology keys, will route to the consul client that is node-local with the connect-inject deployment rather than the application pod. As a result, all services will be registered with one client (the one that's local to connect-injector), and the init container running when your application pod starts will not be able to find the service when the application pod is on a different node from connect-injector.

Only one option comes to mind to work around this: to use pod IPs for the endpoints controller. I'm not sure if using pod IPs would circumvent the service topology enforcement, but if it does, then that could be an option. In that case, I think you need to supply different http/https ports for the init container and the endpoints controller. 

I haven't tested any of this out though, so could be totally wrong!

As a side note, you could try to run acceptance tests in the Helm chart to test this out. I think the `connect` package should give you enough confidence that you haven't broken anything. We have some docs on how to set that up and run them locally: https://github.com/hashicorp/consul-helm/blob/master/CONTRIBUTING.md#acceptance-tests.

karras:
@ishustava Thanks for the review and additional inputs, much appreciated! As discussed in the issue with David it can't be merged. Thus I'm closing this and the other related PRs and issues to prevent cluttering and prepare for the upcoming repo consolidation into consul-k8s.

PR #547, opened 2021-06-28
Clean up unused release scripts/makefile targets

lkysow:
Clean up unused Make targets and build scripts. Some are still needed by the consul-eco-releases but these ones aren't.


lkysow:
@alvin-huang I'm looking at everything inside https://github.com/hashicorp/consul-k8s/tree/master/build-support and wondering if we can delete a lot of those scripts as well?

alvin-huang:
Yeah that whole directory can be deleted. You might need to delete the relevant parts in the Makefile as well.

Makefile
```diff
@@ -35,65 +22,13 @@ CI_DEV_DOCKER_WORKDIR?=.
...
-	@$(SHELL) $(CURDIR)/build-support/scripts/build-local.sh
-
 dev:
 	@$(SHELL) $(CURDIR)/build-support/scripts/build-local.sh -o $(GOOS) -a $(GOARCH)
 
 dev-docker:
 	@$(SHELL) $(CURDIR)/build-support/scripts/build-local.sh -o linux -a amd64
 	@docker build -t '$(DEV_IMAGE)' --build-arg 'GIT_COMMIT=$(GIT_COMMIT)' --build-arg 'GIT_DIRTY=$(GIT_DIRTY)' --build-arg 'GIT_DESCRIBE=$(GIT_DESCRIBE)' -f $(CURDIR)/build-support/docker/Dev.dockerfile $(CURDIR)
 
-dev-tree:
```
alvin-huang, line 94: consul-eco-releases still runs `make dev-tree` in the respective repos. However, we could move this into the release repo instead. Just noting it here that if this is removed and we release, it will be broken until that change is implemented.

alvin-huang:
Do you think anyone will need these targets to build locally at all? If not this is fine to remove.

Makefile
lkysow: gotcha, I'll put that bac

lkysow:
> Do you think anyone will need these targets to build locally at all? If not this is fine to remove.

I left in `make dev-docker` and `make dev`. I can add back other targets if folks were using them.

kschoche (approved):
very nice!

thisisnotashwin (approved):
🦊 NICE!!

Issue #549, opened 2021-07-05
Upstream environment variables not being set


lkysow:
`<NAME>_CONNECT_SERVICE_HOST` and `<NAME>_CONNECT_SERVICE_PORT` environment variables are no longer being set. (originally reported https://github.com/hashicorp/consul-helm/issues/981#issuecomment-873141197)

PR #550, opened 2021-07-05
Fix bug where upstream env vars not set

lkysow:
Fixes issue where we weren't setting the upstream environment variables
when the upstream annotation was set:
* `<NAME>_CONNECT_SERVICE_HOST`
* `<NAME>_CONNECT_SERVICE_PORT`

Since we're modifying a slice's values during iteration we must access
the element via reference using the index.

Fixes https://github.com/hashicorp/consul-k8s/issues/549

How I've tested this PR:
* unit tests

* code

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+BUG FIXES:
+* Connect: Fix bug where environment variables `<NAME>_CONNECT_SERVICE_HOST` and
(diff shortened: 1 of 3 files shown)


kschoche (approved):

thisisnotashwin (approved):
good catch!

sadjamz (approved):

PR #551, opened 2021-07-05
Change error handling when scraping metrics

### LINKED ISSUE #546: Metrics Merging should ignore non 2xx service metrics
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue

When Consul Connect has metrics merging enabled, the merged metrics should ignore service metrics that return non 2xx responses. For example, the following service does not have any metrics and the merged metrics is invalid:

image

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Deploy services without service metrics and add it to the service mesh
1. Enable metrics merging

### Consul info for both Client and Server

Version 1.10

### Operating system and Environment details

OS, Architecture, and any other information you can provide about the environment.

GKE 1.20

### Log Fragments

Include appropriate Client or Server log fragments. If the log is longer than a few dozen lines, please include the URL to the [gist](https://gist.github.com/) of the log instead of posting it in the issue. Use `-log-level=TRACE` on the client and server to capture the maximum log detail.

N/A

lkysow:
Fixes some issues with metrics merging. Namely that if there were any errors getting metrics from Envoy then we'd return a 200 with empty metrics instead of a 500 _and_ that if the service returned an error, e.g. 400, we'd just stick the body of its response in our metrics output which would break Prometheus.

* If Envoy returns an error then also respond with a 500 in our merged
metrics response so that Prometheus will know that we had an error, not
that there are no metrics.
* If the service metrics return with a non-2xx status code then don't
include the response body in the merged metrics. This will stop issues
where users accidentally turn on metrics merging but they don't have an
exporter and so their metrics endpoint returns 404. I could have
responded with a 500 in this case in order to indicate that there is an
error, however I think it's more likely that users are accidentally
turning on metrics merging and the error indication is accomplished via
a new metric (see below).
* Append a new metric that indicates the success of the service
scraping. This can be used for alerting by users since the response code
of the service metrics response is discarded:
  * success: consul_metrics_merging_service_metrics_success 1
  * fail: consul_metrics_merging_service_metrics_success 0
* modify logging to use key/value pairs
* Fixes #546

How I've tested this PR:
- manually and unit tests

- code

Checklist:
- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 BREAKING CHANGES:
 * Control Plane
   * Update minimum go version for project to 1.17 [[GH-878](https://github.com/hashicorp/consul-k8s/pull/878)]
+  * Add boolean metric to merged metrics response `consul_merged_service_metrics_success` to indicate if service metrics were
+    scraped successfully. [[GH-551](https://github.com/hashicorp/consul-k8s/pull/551)]
 
 IMPROVEMENTS:
 * CLI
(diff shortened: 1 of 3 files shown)


lkysow:
@ndhanushkodi, interested in your thoughts on this. (see also exporter guidelines: https://prometheus.io/docs/instrumenting/writing_exporters/#failed-scrapes)

ndhanushkodi (commented):
The behavior looks quite nice to me. Just thinking out loud, could it make sense to have a metric similar to `consul_metrics_merging_service_metrics_success` for the envoy?  But, I think it makes more sense what you're doing here so that the user sees the Prometheus scrape 500/404 from Envoy rather than a successful metrics scraping when something is definitely configured wrong.

I think all of the changes you've listed makes sense.

lkysow:
> could it make sense to have a metric similar to consul_metrics_merging_service_metrics_success for the envoy

I think that that metric now comes out of us returning a 500/200 depending on what happened. I'm sure there's a way to track that in prometheus?

ishustava (approved):
Looks good!

lkysow:
Will rebase changelog after approval

ndhanushkodi (approved):
Looks great! Thanks for fixing this!!

PR #552, opened 2021-07-06
Handle TransparentProxy field CRD syncing

lkysow:
There is a bug (https://github.com/hashicorp/consul-helm/issues/1023) where if you're on Consul <1.10.0 then proxy-defaults no longer works. This is because we were always setting the `transparentProxy` key, even when it was `nil` on the source CRD.

Fixing this bug then caused another issue where Consul always returns the `transparentProxy` key as an empty struct, even when it was written as a nil pointer (https://github.com/hashicorp/consul/issues/10595). To fix this for now, when we do our equals comparison we need to treat the empty struct and nil pointer as equal.

* This fix applies for `ServiceDefaults` and `ProxyDefaults` CRDs that
both allow setting the `transparentProxy` key.
* First, if `transparentProxy` is not set on the CRD, don't write the
empty struct into Consul, instead set the key to `nil`. This fixes a bug
where the CRDs don't work with Consul <1.10 since earlier versions of
Consul don't support that key.
* Second, there is a bug in Consul
(hashicorp/consul#10595) where even if you
pass `transparentProxy` as `nil`, it is returned as the empty struct. To
handle that, write a custom comparator that treats the empty struct for
that field as equal to a nil pointer. This ensures we don't continually
re-sync the CRD.

Fixes https://github.com/hashicorp/consul-helm/issues/1023

How I've tested this PR:
- unit tests

- code

Checklist:
- [x] Tests added

### DIFF
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/hashicorp/consul-k8s/api/common"
 	capi "github.com/hashicorp/consul/api"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -121,7 +122,7 @@ func (in Expose) validate(path *field.Path) field.ErrorList {
 
 func (in *TransparentProxy) toConsul() *capi.TransparentProxyConfig {
 	if in == nil {
-		return &capi.TransparentProxyConfig{OutboundListenerPort: 0}
+		return nil
 	}
 	return &capi.TransparentProxyConfig{
 		OutboundListenerPort: in.OutboundListenerPort,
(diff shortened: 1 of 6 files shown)


api/v1alpha1/shared_types.go
```diff
@@ -121,7 +122,7 @@ func (in Expose) validate(path *field.Path) field.ErrorList {
 
 func (in *TransparentProxy) toConsul() *capi.TransparentProxyConfig {
 	if in == nil {
-		return &capi.TransparentProxyConfig{OutboundListenerPort: 0}
+		return nil
```
lkysow, line 125: Here is the change to write it to Consul as nil if we receive it as nil (which caused the other issues).

lkysow:
I actually haven't tested this against 1.9 which I should do.

api/v1alpha1/shared_types.go
```diff
@@ -169,3 +170,29 @@ func meta(datacenter string) map[string]string {
...
+// 1.10.0 (https://github.com/hashicorp/consul/issues/10595) where Consul will
+// always return the empty struct for the TransparentProxy key even if it was
+// written as a nil pointer. With the default comparator, a nil pointer and
+// empty struct are treated as different and so we would always treat the
+// CRD as not synced and would continually try and write it to Consul.
+func transparentProxyConfigComparer(a, b *capi.TransparentProxyConfig) bool {
+	empty := capi.TransparentProxyConfig{}
+	if a == nil && b == nil {
+		return true
+	}
```
ishustava, line 186: Should we add a comment saying why we need to compare when both are nil (I'm guessing is so that we can dereference it later)?

ishustava (approved):
Looks good! Thanks for fixing it 🙏

sadjamz (approved):
The code and testing made sense to me! Good catch.

api/v1alpha1/shared_types.go
lkysow: I refactored this to just:

```go
	if a == nil && b != nil && *b == empty {
		return true
	}
	if b == nil && a != nil && *a == empty {
		return true
	}
```

PR #566, opened 2021-07-21
Add golangci-lint to consul-k8s repo as github action

kschoche:
Changes proposed in this PR:
- Add golangci-lint to the pipeline as a github action
- Add .golangci.yaml file to top level dir in repo so that users can run it locally via `golangci-lint run`.
- Update all of consul-k8s (minus the tests files) to be in-line with code standards.

Users should install golangci-lint locally via `brew install golangci-lint` and target version `1.41.1`

How I've tested this PR:
Run it locally + in CI in this branch.

Pull the branch and run it locally with various changes to your code.

Checklist:
- [ ] Tests added

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -17,7 +17,7 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v1.41.1
+          version: v1.23.6
 
           # Optional: working directory, useful for monorepos
           # TODO: we may need to modify this when monorepo comes, it could be helpful for when we test
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,12 +1,10 @@
 name: golangci-lint
-on:
-  push:
-    tags:
-      - v*
-    branches:
-      - master
-      - main
-  pull_request:
+on: [push, pull_request]
+  tags:
+    - v*
+  branches:
+    - master
+    - main
 jobs:
   golangci:
     name: lint
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -16,19 +16,9 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
           version: v1.41.1
-
           # Optional: working directory, useful for monorepos
           # TODO: we may need to modify this when monorepo comes, it could be helpful for when we test
           # only control-plane components in a PR, or some other scenario.
           # working-directory: somedir
-
-          # Optional: golangci-lint command line arguments.
-          # args: --issues-exit-code=0
-          # args: --config .golangci.yml
-
-          # Optional: show only new issues if it's a pull request. The default value is `false`.
-          # TODO: if we fix all the issues.. remove
-          # only-new-issues: true
 
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,34 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
+  pull_request:
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.29
+
+          # Optional: working directory, useful for monorepos
+          # TODO: we may need to modify this when monorepo comes, it could be helpful for when we test
+          # only control-plane components in a PR, or some other scenario.
+          # working-directory: somedir
+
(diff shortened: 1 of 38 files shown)


.golangci.yml
```diff
@@ -0,0 +1,34 @@
...
+issues:
+  # Disable the default exclude list so that all excludes are explicitly
+  # defined in this file.
+  exclude-use-default: false
+
+  # Only report deltas off of main branch
+  new: true
+
+  exclude-rules:
+    # These pop up as a result of command tests and can be safely gnored.
```
kschoche, line 13:
```suggestion
    # These pop up as a result of command tests and can be safely ignored. 
```

.golangci.yml
```diff
@@ -0,0 +1,34 @@
...
+  new: true
+
+  exclude-rules:
+    # These pop up as a result of command tests and can be safely ignored. 
+    - linters: [errcheck]
+      text: 'Error return value of `(.*).Stop` is not checked'
+    - linters: [errcheck]
+      text: 'Error return value of `(.*).Write` is not checked'
+    - linters: [errcheck]
+      text: 'Error return value of `(.*).Remove` is not checked'
```
ishustava, line 19: I think these two could be useful for non-test code because, for example, there are a lot of things that implement  the [`Writer` interface](https://pkg.go.dev/io#Writer) and could be using that function. If this is something we really should ignore for tests, then maybe there's a way to exclude only test files? Btw, which test code is this fixing?

.golangci.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    # Allow unused parameters to start with an underscore. Arguments with a name
+    # of '_' are already ignored.
+    # Ignoring longer names that start with underscore allow for better
+    # self-documentation than a single underscore by itself.  Underscore arguments
+    # should generally only be used when a function is implementing an interface.
+    - linters: [unparam]
+      text: '`_[^`]*` is unused'
```
ishustava, line 27: Why do we need this rule?

ishustava (commented):
Great work, Kyle! I left some questions/comments inline.

One other question I have is that I'm not sure if it works locally. I copied the config file onto a branch I was working off of for another PR `delete-acl-token` and `golangci-lint run` didn't  output anything. Then when I added `new-from-rev: master` it showed a lot of errors. I'm not sure why that is the case because when I run `git diff HEAD~` it shows the right diff.  Also, the errors it returned seemed kind of noisy. For example:

```
connect-inject/endpoints_controller_test.go:2203:15: mnd: Magic number: 20000, in <assign> detected (gomnd)
					Port:    20000,
					         ^
```

I wonder if we can dial those down a bit since they don't seem that important. 

Additionally, the errcheck didn't ignore `os.Remove` calls that didn't check for errors.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,39 @@
+name: golangci-lint
+on:
+  push:
```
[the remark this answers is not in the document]
kschoche: I'm still not totally sure on this, but will have another look into it!

subcommand/consul-sidecar/command.go
```diff
@@ -268,13 +265,21 @@ func (c *Command) mergedMetricsHandler(rw http.ResponseWriter, _ *http.Request)
 	}
 
 	// Write Envoy metrics to the response.
-	defer envoyMetrics.Body.Close()
+	defer func() {
+		err = envoyMetrics.Body.Close()
+		if err != nil {
+			c.logger.Info(fmt.Sprintf("Error closing envoy metrics body: %s", err.Error()))
```
ishustava, line 271: I think this should be `c.logger.Error` rather than `Info`

subcommand/create-federation-secret/command.go
```diff
@@ -295,7 +295,7 @@ func (c *Command) replicationToken(logger hclog.Logger) ([]byte, error) {
 	// requires all servers to be running and a leader elected.
 	// This will run forever but it's running as a Helm hook so Helm will timeout
 	// after a configurable time period.
-	backoff.Retry(func() error {
+	_ = backoff.Retry(func() error {
```
ishustava, line 298: should we not ignore this error?

ishustava (approved):
Looks good! A couple of comments, but nothing blocking.

.golangci.yml
```diff
@@ -0,0 +1,30 @@
+linters:
+  # enables all defaults + the below, `golangci-lint linters` to see the list of active linters.
```
ndhanushkodi, line 2: Nice, thanks for this comment!

connect-inject/handler.go
```diff
@@ -8,26 +8,21 @@ import (
 	"net/http"
 	"strconv"
 
-	"github.com/deckarep/golang-set"
+	mapset "github.com/deckarep/golang-set"
 	"github.com/go-logr/logr"
```
ndhanushkodi, line 12: Was this added due to linter or IDE? (My IDE automatically adds it still so I can't quite tell but if the linter wants this that's great!

subcommand/consul-sidecar/command.go
```diff
@@ -136,12 +136,9 @@ func (c *Command) Run(args []string) int {
...
 	go func() {
-		select {
-		case sig := <-c.sigCh:
-			c.logger.Info(fmt.Sprintf("%s received, shutting down", sig))
-			cancelFunc()
-			return
-		}
+		sig := <-c.sigCh
+		c.logger.Info(fmt.Sprintf("%s received, shutting down", sig))
+		cancelFunc()
```
ndhanushkodi, line 141: This is amazing that the linter tells us this simplification!

ndhanushkodi (approved):
Amazing work Kyle, I got a taste of how nice some of the simplifications are!

PR #567, opened 2021-07-21
Merge consul-helm and consul-k8s into one repository

ndhanushkodi:
Co-authored-by: Saad Jamal <saad@hashicorp.com>

This PR has an almost working circleci config that runs:
- linters/fmt/vet for consul-helm and consul-k8s
- unit tests for control-plane
- unit tests for helm templates, and other modules in helm
- acceptance tests

We are looking for early feedback on the circleci config before doing a code freeze on the helm repository.


.circleci/config.yml
```diff
@@ -1,4 +1,7 @@
+# Originally from consul-k8s
```
ndhanushkodi, line 1: This is the file we would like reviewed.

ndhanushkodi:
Hey @ishustava and @thisisnotashwin, we were just looking for any early feedback you might have on just the `.circleci/config.yml` file in this PR. That way when we do the code freeze, we're not waiting on as much code review changes and can migrate quickly.

.circleci/config.yml
```diff
@@ -143,44 +281,652 @@ jobs:
...
+          fail_only: true
+          failure_message: "Acceptance tests against Kind with Kubernetes v1.21 failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+
+  update-helm-charts-index:
+    docker:
+      - image: docker.mirror.hashicorp.services/circleci/golang:latest
+    steps:
+      - checkout
+      - run:
+          name: verify Chart version matches tag version
```
ishustava, line 783: I think we need to set working dir here so that it can find Chart.yaml
```suggestion
          name: verify Chart version matches tag version
          working_directory: *charts-consul-path
```

ishustava (commented):
circleci config looks good! just one small comment, everything else is :shipit: !

ndhanushkodi:
> This looks good Nitya!! My only question/remark is regarding whether the pipeline in question is the PR pipeline or the nightly pipelines? This looks perfect in terms of the nightly pipeline. We would probably only run acceptance and acceptance t-proxy on the PR pipelines if I understand this correctly right?

Yup Ashwin that's right, we would just delete the nightly's section, just ran it yesterday to confirm they still work. 

> I think we need to set working dir here so that it can find Chart.yaml

Thanks for this catch Iryna!!

.circleci/config.yml
ndhanushkodi: thank you!! nice catch!!

PR #569, opened 2021-07-26
add some additional log msgs for connect-init

kschoche:
Changes proposed in this PR:
- After some feedback it seemed helpful to point users at the controller logs if connect-init is stuck in a wait loop, add this.

How I've tested this PR:
code review

code review

Checklist:
- [ ] Tests added


thisisnotashwin (approved):
This change does make sense. I wonder if there should be an FAQ section or something similar where people can kinda search for this information beforehand. It seems a little weird for it to be in log line only.

kschoche:
> This change does make sense. I wonder if there should be an FAQ section or something similar where people can kinda search for this information beforehand. It seems a little weird for it to be in log line only.

@thisisnotashwin - good point! I thought we had scoped out adding it to here for GA : https://www.consul.io/docs/troubleshoot/common-errors but it looks like that didn't happen.

sadjamz (approved):
This is a better log message, makes sense to me!

PR #573, opened 2021-07-27
Fix logging in controller

lkysow:
Previously this would always log:
```
    handler.connect received pod    {"name": "", "ns": ""}
```

How I've tested this PR:
* Ran it and saw the logs

Checklist:
- Tests added (logging so no tests needed)
- CHANGELOG entry added (not sure this is necessary)


kschoche (approved):
Nice catch!

thisisnotashwin (approved):
🦊

PR #571, opened 2021-07-27
connect: endpoints controller deletes ACL token when service is deregistered

ishustava:
Fixes #540 

Changes proposed in this PR:
- Modify endpoints controller to delete ACL tokens for each service instance that it deregisters
- Remove TLS+ACLs table tests from endpoints controller tests. These tests were testing that endpoints controller works with a client configured to have TLS and ACLs. I thought this test was not necessary because there isn't any code in the controller that behaves differently if the consul client is configured with any of those and as a result there's no way these tests could fail. The tests testing to the new ACL logic are there but they are only testing the logic that was added and configure test agent to accommodate for that.
- Create test package under `helper` and move `GenerateServerCerts` function from `subcommand/common` there because it was used outside of `subcommand`. 
- Create a helper test function to set up auth methods and refactor existing `connect-init` command tests to use that function.
- Minor editing fixes of comments etc.

How I've tested this PR:
* unit tests 
* acceptance tests in hashicorp/consul-helm#1056

How I expect reviewers to test this PR:
- code review

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

PR #580, opened 2021-07-30
connect: don't get pods from the API to determine if ACL should be deleted

ishustava:
This is a follow-up to #571 

Endpoints controller may have stale cache for pod objects since it's only
watching the endpoints objects. As a result when we try to get a pod
from cache, it may be stale. For the case of deleting ACL tokens,
we can't tolerate stale cache reads since we need to know for sure
if the pod exists to determine if an ACL token for that pod should be deleted.
This commit changes it to instead use the pod info stored in the endpoints addresses
as TargetRef.

This can be seen in the failed acceptance tests build [here](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/3477/workflows/85035325-7c4d-48ef-9ea2-fdbb3a9a83f3/jobs/15657).

Also, remove `fmt.Printf` and `fmt.Println` statements in tests and code as I think they were left accidentally.

How I've tested this PR:
[acceptance tests](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/3490/workflows/7ec1fc0b-0f7c-4c05-af84-2b71780fd03a)

- code review

Checklist:
- [x] Tests added
- N/A CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

### DIFF
--- a/connect-inject/endpoints_controller.go
+++ b/connect-inject/endpoints_controller.go
@@ -746,7 +746,8 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 
 // reconcileACLTokensForService finds the ACL tokens that belongs to the service and deletes it from Consul.
 // It will only check for ACL tokens that have been created with the auth method this controller
-// has been configured with.
+// has been configured with and will only delete tokens for pods that aren't in endpointPods
+// (endpointPods is a set of pods that the endpoints object is pointing to).
 func (r *EndpointsController) reconcileACLTokensForService(client *api.Client, serviceName, k8sNS string, endpointPods mapset.Set) error {
 	tokens, _, err := client.ACL().TokenList(nil)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


ndhanushkodi (approved):
Nice find, and this looks great! Was this failing consistently in acceptance tests or intermittently?

ishustava:
> Was this failing consistently in acceptance tests or intermittently?

It failed consistently, although I haven't ran it that many times.

sadjamz (approved):
Great work Iryna! Only thing I was thinking of commenting was perhaps adding a sentence in the comment on top of the function `reconcileACLTokensForService()` describing again what endpointPods is. But you already did that in line 127. So probably not required.

ishustava:
>Only thing I was thinking of commenting was perhaps adding a sentence in the comment on top of the function reconcileACLTokensForService()

yeah, I think it couldn't hurt to add that!

PR #581, opened 2021-07-30
connect: skip service registration only for duplicate services on k8s

ishustava:
Previously, we would skip service registration if we find a duplicate service
even when the 'k8s-namespace' meta key is not present on the service.
However, that is not correct behavior since that would ignore any service
instance that could exist on other platforms (e.g. VMs) and that
doesn't have this meta key. We should instead only check services that
have the 'k8s-namespace' meta key.

How I've tested this PR:
- unit tests

- code review

Checklist:
- [x] Tests added (removed in this case 😄)

### DIFF
--- a/connect-inject/endpoints_controller_test.go
+++ b/connect-inject/endpoints_controller_test.go
@@ -4657,94 +4657,79 @@ func TestCreateServiceRegistrations_withTransparentProxy(t *testing.T) {
 func TestRegisterServicesAndHealthCheck_skipsWhenDuplicateServiceFound(t *testing.T) {
 	t.Parallel()
 
-	cases := map[string]struct {
-		consulServiceMeta map[string]string
-	}{
-		"different k8s namespace meta": {
-			consulServiceMeta: map[string]string{MetaKeyKubeNS: "some-other-ns"},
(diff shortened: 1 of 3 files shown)


connect-inject/endpoints_controller_test.go
```diff
@@ -4656,93 +4656,78 @@ func TestCreateServiceRegistrations_withTransparentProxy(t *testing.T) {
...
 
-	cases := map[string]struct {
-		consulServiceMeta map[string]string
-	}{
-		"different k8s namespace meta": {
-			consulServiceMeta: map[string]string{MetaKeyKubeNS: "some-other-ns"},
-		},
-		"no k8s namespace meta": {
-			consulServiceMeta: nil,
-		},
```
ishustava, line 4667: This diff is showing deleting this case and in-lining everything into the test since we only have one case now.

kschoche: Hooray deleting code :)

kschoche (approved):
Looks great!

ndhanushkodi (approved):
Great catch, looks good!

PR #584, opened 2021-07-30
connect-init: Allow setting -k8s-service-name flag

### LINKED ISSUE #578: connect-init command breaks workaround for supporting multiple ports in sidecar proxy
As of consul-helm chart v0.32.0 which runs consul 1.10 and consul-k8s 0.26.0 we can no longer use our workaround to support multiple ports in sidecar proxy as suggested in open issue https://github.com/hashicorp/consul/issues/5388#issuecomment-854061709. Our workaround was functional as of Consul 1.9.4 (consul-helm 0.31.1).

The consul upgrade to 1.10 is integrated with Kubernetes objects. Kubernetes services are now part of the standard deployment of any pod that is ran on the mesh.

Each pod is now tied to a Kubernetes service and only one Kubernetes service is expected since a pod can only expose one port due to Consul's limitation to supporting one port per pod. In consul-k8s, the new connect-init function makes that assumption and breaks the current workaround we have in place to support multiple ports.

It is made clear in the code that no more than two services are expected (https://github.com/hashicorp/consul-k8s/blob/v0.26.0/subcommand/connect-init/command.go#L166).  Since it doesn't log the services that are returned, I presume that it's returning all the services associated with that pod.

`filter := fmt.Sprintf("Meta[%q] == %q and Meta[%q] == %q", connectinject.MetaKeyPodName, c.flagPodName, connectinject.MetaKeyKubeNS, c.flagPodNamespace)` (https://github.com/hashicorp/consul-k8s/blob/v0.26.0/subcommand/connect-init/command.go#L159)

I think adding the service name to the filter will bring back exactly the Kubernetes services tied to the service name.

Our workaround is tied to manually adding sidecars to the main application container. Here is an example:

POD A --->service 1 (port 9888) with associated Kubernetes Service 1
            --->service 2 (port 9889) with associated Kubernetes Service 2

We have POD A which exposes two ports each representing a different service on POD A. We created the corresponding Kubernetes service for each of the services exposed in POD A.

In this example I think the filter for Pod A for service 1 is returning Kubernetes Services 1 and 2 since the filter is only applied to the pod/namespace and not pod/namespace/service.

ndhanushkodi:
Changes proposed in this PR:
- connect-init: Setting `-k8s-service-name` will narrow search in Consul for service in init container to services with a meta key matching that K8s service name. The reason to use `-k8s-service-name` rather than `-service-name` as originally discussed, is because `-service-name` is an already existing flag which we expect to be the Consul service name from the pod annotation, which in many cases will NOT match the K8s service name.

How I've tested this PR:
- unit tests

- code review

Fixes #578 

Checklist:
- [x] Tests added

### DIFF
--- a/subcommand/connect-init/command.go
+++ b/subcommand/connect-init/command.go
@@ -37,7 +37,8 @@ type Command struct {
 	flagAuthMethodNamespace    string // Consul namespace the auth-method is defined in.
 	flagConsulServiceNamespace string // Consul destination namespace for the service.
 	flagServiceAccountName     string // Service account name.
-	flagServiceName            string // Service name.
+	flagServiceName            string // Consul Service name.
+	flagK8sServiceName         string // Kubernetes Service name.
 	flagLogLevel               string
 	flagLogJSON                bool
 
@@ -62,7 +63,10 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagAuthMethodNamespace, "auth-method-namespace", "", "Consul namespace the auth-method is defined in")
 	c.flagSet.StringVar(&c.flagConsulServiceNamespace, "consul-service-namespace", "", "Consul destination namespace of the service.")
 	c.flagSet.StringVar(&c.flagServiceAccountName, "service-account-name", "", "Service account name on the pod.")
-	c.flagSet.StringVar(&c.flagServiceName, "service-name", "", "Service name as specified via the pod annotation.")
+	c.flagSet.StringVar(&c.flagServiceName, "service-name", "", "Service name in Consul as specified via the pod annotation.")
+	c.flagSet.StringVar(&c.flagK8sServiceName, "k8s-service-name", "",
+		"Kubernetes service name matching this pod. If provided, the connect-init command will wait for a "+
+			"service to be registered in Consul that also matches the value of this flag.")
 	c.flagSet.StringVar(&c.flagLogLevel, "log-level", "info",
 		"Log verbosity level. Supported values (in order of detail) are \"trace\", "+
 			"\"debug\", \"info\", \"warn\", and \"error\".")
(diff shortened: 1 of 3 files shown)


subcommand/connect-init/command.go
```diff
@@ -63,6 +64,9 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagConsulServiceNamespace, "consul-service-namespace", "", "Consul destination namespace of the service.")
 	c.flagSet.StringVar(&c.flagServiceAccountName, "service-account-name", "", "Service account name on the pod.")
 	c.flagSet.StringVar(&c.flagServiceName, "service-name", "", "Service name as specified via the pod annotation.")
```
ishustava, line 66: nit: should we update the description of this one to say that it's the consul service name? No one is really running `--help` on this command, but maybe just for ourselves?

subcommand/connect-init/command_test.go
```diff
@@ -105,6 +106,21 @@ func TestRun_ServicePollingWithACLsAndTLS(t *testing.T) {
...
+			k8sServiceName:     "counting",
+		},
+		{
+			name:               "ACLs enabled, k8s-service-name flag is set and doesn't match the service in Consul",
+			tls:                false,
+			serviceAccountName: "counting",
+			serviceName:        "",
+			k8sServiceName:     "not-in-consul",
+			expFail:            true,
+		},
```
ishustava, line 123: Do we need to add these tests cases to `TestRun_ServicePollingOnly` to test this code without ACLs? I'm thinking that since ACL token affects the kind of services you have access to, it might good to add those too.

ishustava (approved):
Looks great! Thanks for making this fix so quickly 🙏 . I left a couple of minor comments, but they are not blocking.

subcommand/connect-init/command.go
ndhanushkodi: Good idea!

kschoche (approved):
👍🏽 
Looks good, I think it might be worth adding the extra tests that Iryna mentioned unless it no longer is necessary!
Great work!

subcommand/connect-init/command_test.go
ndhanushkodi: Good idea, added this!

ndhanushkodi:
Closing due to this not being able to fully workaround multiple ports

PR #590, opened 2021-08-06
Use docker-entrypoint when running consul

lkysow:
Previously we were executing consul directly via /bin/consul and we were
using the Kubernetes `command` key which overrides the default Docker
ENTRYPOINT.

This had no issues, however I propose that it's best to execute via the
entrypoint so we keep things consistent with how the Docker image was
supposed to be run. This is also consistent with Vault's Helm chart. If
you step through the entrypoint script, nothing actually changes with
how we execute consul because we don't trigger any of the `if`
statements. So there is no real effect right now.

In order to support using the entrypoint script, we need to disable part of the
script that attempts to change the ownership of the /consul directory
because the ownership is already set via Kube. To disable this we set the
`DISABLE_PERM_MGMT` env var to `true`.

Entrypoint script is here: https://github.com/hashicorp/docker-consul/blob/master/0.X/docker-entrypoint.sh

How I've tested this PR:
* acceptance tests

* code

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 IMPROVEMENTS:
 * Add flags `-log-level`, `-log-json` to all subcommands to control log level and json formatting. [[GH-523](https://github.com/hashicorp/consul-k8s/pull/523)]
-* Execute consul clients and servers using the docker entrypoint to keep consistent. [[GH-590](https://github.com/hashicorp/consul-k8s/pull/590)]
+* Execute Consul clients and servers using the Docker entrypoint for consistency. [[GH-590](https://github.com/hashicorp/consul-k8s/pull/590)]
 
 BUG FIXES:
 * Connect: Use `AdmissionregistrationV1` instead of `AdmissionregistrationV1beta1` API as it was deprecated in k8s 1.16. [[GH-558](https://github.com/hashicorp/consul-k8s/pull/558)]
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 IMPROVEMENTS:
 * Add flags `-log-level`, `-log-json` to all subcommands to control log level and json formatting. [[GH-523](https://github.com/hashicorp/consul-k8s/pull/523)]
+* Execute consul clients and servers using the docker entrypoint to keep consistent. [[GH-590](https://github.com/hashicorp/consul-k8s/pull/590)]
```
t-eckert, line 5: Nit: "Consul" instead of "consul" and "Docker" instead of "docker" to be consistent with previous Changelog items.

t-eckert (approved):
Looks good to me, given my pretty limited knowledge of the domain. I only had a nitpick in the Changelog.

CHANGELOG.md
ndhanushkodi, line 5:
```suggestion
* Execute consul clients and servers using the docker entrypoint for consistency. [[GH-590](https://github.com/hashicorp/consul-k8s/pull/590)]
```

ndhanushkodi (approved):
Looks good!!

CHANGELOG.md
lkysow, line 5:
```suggestion
* Execute Consul clients and servers using the Docker entrypoint to keep consistent. [[GH-590](https://github.com/hashicorp/consul-k8s/pull/590)]
```

lkysow: Nice catch!

PR #591, opened 2021-08-06
Docker image for OpenShift

lkysow:
Changes proposed in this PR:
- Add new Dockerfile for building OpenShift images based on redhat's UBI image
- Move Dockerfile for our regular image from the releases repo to here so that both the regular and UBI images are in the same repo

#### UBI Changes
- Base image changes from alpine to registry.access.redhat.com/ubi8/ubi-minimal:8.4
- the license file must live at /licenses/mozilla.txt
- creating the user and group is moved to after the package installs since the groupadd utility is not available before the package installs
- microdnf is used instead of apk to install packages since apk is an alpine utility and microdnf is a smaller yum
- install wget, unzip to allow downloading from releases.hashicorp.com
- install shadow-utils to add the groupadd and adduser utilities
- don't install su-exec since it's not available in ubi
- the only supported arch is amd64 so don't build for other archs
- use gpgpconf --kill to stop dirmngr and gpg-agent (copied from vault)
- the consul-k8s group is hardcoded to 1000 and user is hardcoded to 100

How I've tested this PR:
- ran acceptance tests for openshift manually

- compare the two dockerfiles and ask any questions about the changes to the UBI file

### DIFF
--- a/control-plane/build-support/docker/Release.dockerfile
+++ b/control-plane/build-support/docker/Release.dockerfile
@@ -1,0 +1,70 @@
+# This Dockerfile creates a production release image for the project. This
+# downloads the release from releases.hashicorp.com and therefore requires that
+# the release is published before building the Docker image.
+#
+# We don't rebuild the software because we want the exact checksums and
+# binary signatures to match the software and our builds aren't fully
+# reproducible currently.
+FROM alpine:3.13
+
+# NAME and VERSION are the name of the software in releases.hashicorp.com
+# and the version to download. Example: NAME=consul VERSION=1.2.3.
+ARG NAME
+ARG VERSION
+
+LABEL name=$NAME \
+      maintainer="Consul Team <consul@hashicorp.com>" \
+      vendor="HashiCorp" \
+      version=$VERSION \
+      release=$VERSION \
+      summary="The consul-k8s binary includes first-class integrations between Consul and Kubernetes." \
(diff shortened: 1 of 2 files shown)


lkysow:
Can also compare against https://github.com/hashicorp/vault-k8s/blob/master/build/docker/Release.ubi.dockerfile

lkysow:
Tests are failing for some other issue.

sadjamz:
Questions from Aug 11 Sync:
Are we sure amd64 is the only supported architecture?
Why are we calling `gpgconf --kill all`?

kschoche (approved):
Looks great!

david-yu:
@sadjamz We should support x86 to start since that is currently what Vault supports as the architecture for OpenShift clusters ([vault](https://catalog.redhat.com/software/containers/hashicorp/vault/5fda55bd2937386820429e0c?container-tabs=overview&gti-tabs=registry-tokens))

sadjamz (approved):
Thanks Luke!

lkysow:
> Why are we calling gpgconf --kill all?

Looked into this, this explains it: https://unix.stackexchange.com/questions/188668/how-does-gpg-agent-work. It stops the process from running since it's not needed after the image is built.

> Are we sure amd64 is the only supported architecture?

No, there are other OpenShift architectures however we're starting with just amd64 support and will revisit.

PR #604, opened 2021-08-12
Add ability to specify port for ui service

lkysow:
I followed the same pattern as for `nodePort`. This is generally useful for specifying load balancer ports and we have this option already for mesh gateways. This is also useful for running in Minikube because `minikube tunnel` will open up `localhost:<port>` and so you can have the Consul UI on `localhost:8500` rather than `localhost:80`.

Taken from https://github.com/hashicorp/consul-helm/pull/1062

How I've tested this PR:
* helm tests

How I expect reviewers to test this PR: code

Checklist:
- [x] Bats tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+IMPROVEMENTS:
+* Helm Chart
+  * Add ability to specify port for ui service. [[GH-604](https://github.com/hashicorp/consul-k8s/pull/604)]
+
 ## 0.33.0 (August 12, 2021)
 
 BREAKING CHANGES:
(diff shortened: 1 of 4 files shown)


lkysow:
From @kschoche 

> Looks great! Given that you mentioned it's potentially helpful in minikube, could we update a doc as well somewhere pointing this out? No worries if not.

Our minikube guide uses `minikube service consul-ui` (https://learn.hashicorp.com/tutorials/consul/kubernetes-minikube?utm_source=consul.io&utm_medium=docs&utm_content=k8s&utm_term=mk#access-the-consul-ui) which is another way to do it so I don't think we need to update that guide.

`minikube tunnel` is nice if you hav multiple services but that's a unique use-case.

kschoche (approved):
Nice!

sadjamz (approved):
This'll be really useful (I think for the CLI too?) thanks Luke!

PR #607, opened 2021-08-12
Fix make dev tree changelog path

ndhanushkodi:
The makefile has to be called inside the control-plane directory so this fixes the path for changelog in make dev-tree


sadjamz (approved):
Putting a note here, let's write some more comments later explaining the whole CD'ing path and everything.

PR #608, opened 2021-08-12
Fix make dev-tree git functions

ndhanushkodi:
Changes proposed in this PR:
- The git functions were failing because they should be run from top level


sadjamz (approved):
Good comments.

PR #619, opened 2021-08-17
[sync-catalog] Fix NodePort register service with wrong internal IP when multiple internal IPs are reported on the node

kolorful:
This PR fixed an issue where consul-sync will register NodePort services with wrong internal IPs when multiple internal IPs are reported on the node, and this is a common pattern when using the AWS CNI plugin.

When using the AWS CNI plugin, multiple ENIs could be used and the secondary ENI with multiple IPs are not used by the node itself, but assigned to individual pods. In this case kubelet will end up reporting all used IPs on all ENIs as internal IPs on the node. ([Detailed explanation](https://github.com/kubernetes/kubernetes/issues/61921#issuecomment-384089455))

Kubernetes has since [made sure the first IP would be equal to the IP on eth0](https://github.com/kubernetes/kubernetes/blob/fe099b2abdb023b21a17cd6a127e381b846c1a1f/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go) ([reference 2](https://github.com/kubernetes/kubernetes/issues/61921#issuecomment-586036594)) and it would be the actual IP of the Node.

Prior to my fix the consul-sync program always sync the last matching IP to consul and it ends up being a Pod IP on that node. This PR fixed it by always picking the first matching IP as how Kubernetes is implemented.

Changes proposed in this PR:
- Use the first found Node IP address when syncing NodePort service to consul

How I've tested this PR:

- make test
- Built a docker image with my change then tested in our AWS cluster that has multiple ENIs and verify it now syncs the correct Node IP address for NodePort service.

- Run test suite should be good enough since I've already tested in a real cluster.

Checklist:
- [x] Bats tests added

### DIFF
--- a/control-plane/catalog/to-consul/resource.go
+++ b/control-plane/catalog/to-consul/resource.go
@@ -553,6 +553,12 @@ func (t *ServiceResource) generateRegistrations(key string) {
 						r.Service.Address = address.Address
 
 						t.consulMap[key] = append(t.consulMap[key], &r)
+						// Only consider the first address that matches. In some cases
+						// there will be multiple addresses like when using AWS CNI.
+						// In those cases, Kubernetes will ensure eth0 is always the first
+						// address in the list.
+						// See https://github.com/kubernetes/kubernetes/blob/b559434c02f903dbcd46ee7d6c78b216d3f0aca0/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1462-L1464
+						break
 					}
 				}
 
@@ -568,6 +574,12 @@ func (t *ServiceResource) generateRegistrations(key string) {
 							r.Service.Address = address.Address
 
 							t.consulMap[key] = append(t.consulMap[key], &r)
+							// Only consider the first address that matches. In some cases
+							// there will be multiple addresses like when using AWS CNI.
+							// In those cases, Kubernetes will ensure eth0 is always the first
+							// address in the list.
+							// See https://github.com/kubernetes/kubernetes/blob/b559434c02f903dbcd46ee7d6c78b216d3f0aca0/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1462-L1464
+							break
 						}
 					}
 				}
--- a/control-plane/catalog/to-consul/resource_test.go
+++ b/control-plane/catalog/to-consul/resource_test.go
@@ -1527,6 +1527,7 @@ func createNodes(t *testing.T, client *fake.Clientset) (*apiv1.Node, *apiv1.Node
 			Addresses: []apiv1.NodeAddress{
 				{Type: apiv1.NodeExternalIP, Address: "1.2.3.4"},
 				{Type: apiv1.NodeInternalIP, Address: "4.5.6.7"},
+				{Type: apiv1.NodeInternalIP, Address: "7.8.9.10"},
 			},
 		},
 	}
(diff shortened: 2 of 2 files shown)


control-plane/catalog/to-consul/resource.go
```diff
@@ -553,6 +553,7 @@ func (t *ServiceResource) generateRegistrations(key string) {
 						r.Service.Address = address.Address
 
 						t.consulMap[key] = append(t.consulMap[key], &r)
+						break
```
lkysow, line 556:
```suggestion
						// Only consider the first address that matches. In some cases
						// there will be multiple addresses like when using AWS CNI.
						// In those cases, Kubernetes will ensure eth0 is always the first
						// address in the list.
						// See https://github.com/kubernetes/kubernetes/blob/b559434c02f903dbcd46ee7d6c78b216d3f0aca0/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1462-L1464
						break
```

control-plane/catalog/to-consul/resource.go
```diff
@@ -568,6 +569,7 @@ func (t *ServiceResource) generateRegistrations(key string) {
 							r.Service.Address = address.Address
 
 							t.consulMap[key] = append(t.consulMap[key], &r)
+							break
```
lkysow, line 572:
```suggestion
							// Only consider the first address that matches. In some cases
							// there will be multiple addresses like when using AWS CNI.
							// In those cases, Kubernetes will ensure eth0 is always the first
							// address in the list.
							// See https://github.com/kubernetes/kubernetes/blob/b559434c02f903dbcd46ee7d6c78b216d3f0aca0/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1462-L1464
							break
```

lkysow (commented):
Thanks for the PR! Looks good, just one suggestion for comments.

kolorful:
@lkysow thanks for reviewing the PR! I've merged in your suggestion and squashed the commit.

lkysow (approved):

kolorful:
Hi @lkysow, is it possible to backport this change to 0.25, 0.26, 0.33 because we are using `consul:1.9.5` + `consul-k8s 0.25` and consul 1.10 introduced a [backward incompatible change](https://github.com/hashicorp/consul/pull/9981) which caused catalog-sync to fail on errors like:
```
2021-10-01T18:33:37.897Z [WARN]  to-consul/sink: error registering service: node-name=k8s-sync service-name=tstsvc service="&{ tstsvc-b48e46ea737f tstsvc [awsnrt2] map[external-k8s-ns:tstsvc external-source:kubernetes port-http:80 port-https:443] 30080 10.135.200.155  map[] {0 0} false 0 0  <nil> <nil>   }" err="Unexpected response code: 400 (Request decode failed: json: unknown field "SocketPath")"
```

lkysow:
Hi @kolorful are you not able to upgrade to consul 1.10 and consul-k8s latest? Or is there an issue with those versions together?

kolorful:
Hi @lkysow we plan to, but it would take some time before we can get there. If back-port is easy it would be really appreciated so we don't need to maintain a fork for the time being. Thank you :)

lkysow:
Ahh I see. We actually don't do any backport releases of consul-k8s right now so it's unfortunately not as easy as running a script.

kolorful:
I see, that's fine, thank you!

Issue #602, opened 2021-08-12
Allow container securityContext to be set by Helm chart


tarquin-the-brave:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Is your feature request related to a problem? Please describe.

The output of the Helm chart fails to pass [`kubeaudit all`](https://github.com/Shopify/kubeaudit).

e.g. for the consul server statefulSet:

```
---------------- Results for ---------------

  apiVersion: apps/v1
  kind: StatefulSet
  metadata:
    name: RELEASE-NAME-consul-server
    namespace: default

--------------------------------------------

-- [error] CapabilityOrSecurityContextMissing
   Message: Security Context not set. The Security Context should be specified and all Capabilities should be dropped by setting the Drop list to ALL.
   Metadata:
      Container: consul

-- [error] AllowPrivilegeEscalationNil
   Message: allowPrivilegeEscalation not set which allows privilege escalation. It should be set to 'false'.
   Metadata:
      Container: consul

-- [warning] PrivilegedNil
   Message: privileged is not set in container SecurityContext. Privileged defaults to 'false' but it should be explicitly set to 'false'.                                     Metadata:
      Container: consul

-- [error] ReadOnlyRootFilesystemNil                                                                                                                                           Message: readOnlyRootFilesystem is not set in container SecurityContext. It should be set to 'true'.
   Metadata:                                                                                                                                                                      Container: consul
```

These can all be resolved by setting them in the [`container securityContext`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core) (which is different to the [`Pod securityContext`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podsecuritycontext-v1-core)).

#### Feature Description

In the same manner as the Pod security context: expose fields in `values.yaml`, `containerSecurityContext`, under `client` and `server` sections and template in the given security context like so:

```
securityContext:
[...]

PR #620, opened 2021-08-17
WIP: #602 Allow container securityContext to be set

tarquin-the-brave:
Changes proposed in this PR:
- Allow the container securityContext, for containers in the client and server Pods, to be configured by the Helm chart.

This PR doesn't add in any default values for the containers' securityContext. 

How I've tested this PR: TODO

How I expect reviewers to test this PR: TODO

Checklist:
- [ ] Bats tests added


charts/consul/templates/client-daemonset.yaml
```diff
@@ -353,6 +353,10 @@ spec:
             {{- toYaml .Values.client.resources | nindent 12 }}
             {{- end }}
           {{- end }}
+          {{- if .Values.client.containerSecurityContext.client }}
+          securityContext:
+            {{- toYaml .Values.client.containerSecurityContext.client | nindent 12 }}
+          {{- end }}
```
tarquin-the-brave, line 359: I notice the Pod `securityContext`s are wrapped in a:

```
securityContext:
    ...
{{- end }}
```

I don't know that much about openshift, would these container Security contexts need to have this same conditional wrapper too?

charts/consul/values.yaml
```diff
@@ -471,6 +471,14 @@ server:
     runAsUser: 100
     fsGroup: 1000
 
+  # The container securityContext for the server pods.  In addition to the [Pod
+  # secuirtyContext](#v-server-securitycontext) this can set the capabilities
+  # of processes running in the container and ensure the root file systems in
+  # the container is read-only.
+  # @type: map
+  # @recurse: false
+  containerSecurityContext: null
```
lkysow, line 480: I think we need `server` within here to match `client` and to future-proof ourselves.

lkysow (commented):
Thanks for the PR! Can you also write some bats tests for this change? We should test that you can set `securityContext` for each container and that if openshift is enabled the settings are ignored.

kschoche:
One additional comment is that we do need to run in escalated modes when transparent proxy is enabled, this is done in part of the webhook (https://github.com/hashicorp/consul-k8s/blob/37875730318912169571f996f37ab02055e3565c/control-plane/connect-inject/container_init.go#L203) where the sidecars are attached, I do not think that the tool you're using would catch this since it resides in the consul-k8s binary.
I think in order for this to actually work at runtime we'd need to gate it on transparent proxy being disabled, somehow? Unfortunately that is challenging because you can enable transparent proxy at runtime via a pod annotation. @lkysow thoughts?

lkysow:
@tarquin-the-brave you should be able to rebase off of master to get the builds passing.

@kschoche makes a good point. This only applies to some of the pods we create through the helm chart, not the containers we create via injection.

tarquin-the-brave:
> One additional comment is that we do need to run in escalated modes when transparent proxy is enabled, this is done in part of the webhook (
> 
> https://github.com/hashicorp/consul-k8s/blob/37875730318912169571f996f37ab02055e3565c/control-plane/connect-inject/container_init.go#L203
> 
> ) where the sidecars are attached, I do not think that the tool you're using would catch this since it resides in the consul-k8s binary.
> I think in order for this to actually work at runtime we'd need to gate it on transparent proxy being disabled, somehow? Unfortunately that is challenging because you can enable transparent proxy at runtime via a pod annotation. @lkysow thoughts?

This change is only affecting the Pods deployed directly by the Helm chart.  Init and sidecar containers injected by the webhook server won't be affected.

Whether that's something you'd want to make configurable is another question.  There's a strong case to not want the initContainer's securityContext configurable: You need the capabilities it sets to manipulate the iptables (I assume) for transparent proxying. For the Envoy sidecar proxy you're also doing some container securityContext hardening https://github.com/hashicorp/consul-k8s/blob/37875730318912169571f996f37ab02055e3565c/control-plane/connect-inject/envoy_sidecar.go#L69-L74

These are the fields that overlap with the Pod securityContext.  I imagine they're set here so that even if the the Pod that's being injected hasn't set them we ensure the sidecar proxy being injected has them.

The motivation for this PR was "there's further security hardening you can only do in the container securityContext, the Pods defined in this Helm chart don't set any fields in the container securityContext.  Make it configurable so we can set some more fields and harden our Pods".  This PR doesn't do any security hardening in itself. Where the injected Pods are already setting some fields in the container securityContext it would be messy to take values from the users and correct/overlay them so the application can run.  A better solution in that case would be "harden the injected containers to the maximum that kubernetes allows, only giving containers the capabilities they need".  That is pretty much our security requirements.  This [guidance from the NSA](https://media.defense.gov/2021/Aug/03/2002820425/-1/-1/1/CTR_KUBERNETES%20HARDENING%20GUIDANCE.PDF) has a lot of good stuff in it on that.

In our use case we're actually running "Native Integration Mode" so we're not using the sidecar injection.  I've also not made this change to the sidecar-injector Pod either, but can do so that this change covers "all Pods deployed by Helm chart".  For our purposes, this PR with that addition would allow us to set the hardening fields we need. It's very much "make it configurable, we'll do the hardening".  But the paragraph above got me thinking whether or not this is the right change for this project...

Every container that's deployed is running functionality delivered by "the Consul project".  You know what permissions and capabilities different components need, and thus what "max hardened" would look like (at least in terms of k8s API field values).  Does it make sense to make "Pod hardening fields" configurable at all?  Unless you want the Helm chart to accommodate people  swapping in their own rebuilds of components that might need more permissions.  But that does come at the risk to all the other users of "they make their Pods less secure".

I suppose, like with the Pod securityContexts, this PR could be followed up with one to set defaults for these values that provide max hardening.  As you mentioned in #602, you were still discussing internally what those defaults should be.  When we do our testing, we could feed back to you "we tried max hardening the containers and this is what we found...".

What do you think?  Would this PR with the recommended tests added, and having the webhook server container with securityContext configurable (for completeness), be something you'd want to take?

lkysow:
If you can also determine the correct settings and submit that via PR then that would be welcome. The reason we're not doing that ourselves is simply due to time/priority constraints.

tarquin-the-brave:
Yeah.  As part of integrating this change we're going to do the testing to work out what the maximum container securityContext is that won't break any of the components.  We'll then be configuring this chart with those values, so we're more than happy to follow up with a PR to set them as defaults.

charts/consul/values.yaml
```diff
@@ -471,6 +471,18 @@ server:
     runAsUser: 100
     fsGroup: 1000
 
+  # The container securityContext for each container in the server pods.  In
+  # addition to the [Pod secuirtyContext](#v-server-securitycontext) this can
```
lkysow, line 475:
```suggestion
  # addition to the Pod's SecurityContext this can
```

Linking would be nice but we don't do this elsewhere and might break later.

charts/consul/values.yaml
```diff
@@ -804,6 +816,26 @@ client:
     runAsUser: 100
     fsGroup: 1000
 
+  # The container securityContext for each container in the client pods.  In
+  # addition to the [Pod secuirtyContext](#v-client-securitycontext) this can
```
lkysow, line 820:
```suggestion
  # addition to the Pod's SecurityContext this can
```

lkysow (approved):
This is great work. Just one small comment and we're good to go.

tarquin-the-brave:
@lkysow Thanks for your help!

Looks like I leave this with you to merge.

It'll be a couple of weeks yet, but when we do our Pod security hardening exercise I'll make sure to follow this PR up with one to set some strong defaults for these values.

PR #624, opened 2021-08-18
Combined .gitignore across merged repos.

sadjamz:
Changes proposed in this PR:
- Previously we had two .gitignores, one from the control-plane and one from charts/consul. This PR makes one .gitignore that is basically the set union of these two.
- '.idea/' is also ignored incase anyone uses GoLand or any other JetBrains IDE.

How I've tested this PR:
Manually tried adding files and didn't see them when typing `git status`.

Code review.

Checklist:
- [ ] Bats tests added


.gitignore
```diff
@@ -4,3 +4,6 @@
 terraform.tfstate*
 terraform.tfvars
 values.dev.yaml
+bin/
+pkg/
+.idea/
```
ndhanushkodi, line 9: Should we include control-plane bin and pkg folders too? I think the reason we ignore them is temporary stuff gets created in there in makefile scripts or when we do builds, so might make sense to ignore them inside the module as well
```suggestion
.idea/
control-plane/bin
control-plane/pkg
```

ndhanushkodi (approved):
Had an additional suggestion, but looks good to me!

.gitignore
ndhanushkodi: maybe it automatically includes control-plane/bin and control-plane/pkg and we don't have to do this? I already see those not being added by git

.gitignore
```diff
@@ -4,3 +4,6 @@
 terraform.tfstate*
 terraform.tfvars
 values.dev.yaml
+bin/
+pkg/
```
lkysow, line 8: Do we have these dirs?

t-eckert: Wouldn't they be generated by running `go build` and the like?

.gitignore
lkysow: once cli comes in then it'll probably have those dirs too so probably good to have it at the top-level

lkysow (approved):

PR #626, opened 2021-08-19
Added healthchecks to the connect inject webhook.

NicoletaPopoviciu:
Based on #458 
Changes proposed in this PR:
- added readiness and liveness checks to the connect inject webhook
- @ishustava + I  think that only the webhook should have the readiness check:
  - It doesn't make sense to incorporate it into the readiness status because controller doesn't receive any traffic and doesn't listen on any ports (only the webhook does).
  - If consul cluster is unhealthy, the endpoints controller will keep requeuing the request, and the readiness check will not help in changing that behavior. It will also not help to add a liveness check in that case, since the current behavior of requeuing the request is better than letting kubernetes to restart the container.

How I've tested this PR:
-unit tests
-acceptance tests

-code review

Checklist:
- [x] tests added


.circleci/config.yml
```diff
@@ -68,7 +68,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:latest"
```
NicoletaPopoviciu, line 71: reminder to revert before merging

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -193,6 +193,26 @@ spec:
...
+            periodSeconds: 2
+            successThreshold: 1
+            timeoutSeconds: 5
+          readinessProbe:
+            httpGet:
+              path: /readyz/ready
+              port: 9445
+              scheme: HTTP
+            failureThreshold: 2
+            initialDelaySeconds: 2
```
thisisnotashwin, line 212: This is more out of curiosity than anything else, but is there a reason the `initialDelaySeconds` is 2 here while it's value is 1 in the livenessProbe?

control-plane/connect-inject/health_checks_test.go
```diff
@@ -0,0 +1,53 @@
...
+			rc := ReadinessCheck{tmpDir}
+			err = rc.Ready(nil)
+			if tt.expectError {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+
```
ishustava, line 48:
```suggestion
```

ishustava (approved):
Looks great Nicoleta 🎉 😃 🐶 !!

control-plane/connect-inject/health_checks_test.go
```diff
@@ -0,0 +1,53 @@
...
+		keyFileContents  *string
+		expectError      bool
+	}{
+		{"Both cert and key files not present.", nil, nil, true},
+		{"Cert file not empty and key file missing.", ptrToString("test"), nil, true},
+		{"Key file not empty and cert file missing.", nil, ptrToString("test"), true},
+		{"Both cert and key files are present and not empty.", ptrToString("test"), ptrToString("test"), false},
+		{"Both cert and key files are present but both are empty.", ptrToString(""), ptrToString(""), true},
+		{"Both cert and key files are present but key file is empty.", ptrToString("test"), ptrToString(""), true},
+		{"Both cert and key files are present but cert file is empty.", ptrToString(""), ptrToString("test"), true},
```
thisisnotashwin, line 24: This is excellent!!

thisisnotashwin (approved):
This looks great!! Excellent first PR @NicoletaPopoviciu

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -193,6 +193,26 @@ spec:
...
                 {{- end }}
                 {{- end }}
+          livenessProbe:
+            httpGet:
+              path: /readyz/ready
+              port: 9445
+              scheme: HTTP
+            failureThreshold: 2
+            initialDelaySeconds: 1
+            periodSeconds: 2
```
lkysow, line 203: I'm curious if it needs to be checked every 2 seconds? How likely is it for the certs to be removed from disk?

charts/consul/templates/connect-inject-deployment.yaml
ishustava: We've used the same configuration as before for both probes and didn't re-evaluate whether those make sense since it's essentially the same check as before. We'll take a deeper look on Monday.

charts/consul/templates/connect-inject-deployment.yaml
ishustava: Yep, we'll take another look. We've added the same configuration as before the refactor, but we will re-think whether those still make sense.

NicoletaPopoviciu: We kept periodSeconds=2, as it was the value we had before the t-proxy refactor; we looked more into Kubernetes documentation, and the default value for periodSeconds is 10. Do you have any guidance as to whether we should increase the value, and by how much? And to your point once there the certificates are available, they are unlikely to be removed.

charts/consul/templates/connect-inject-deployment.yaml
NicoletaPopoviciu: We took another look, the two probes are initially triggered 1 second apart to keep the two requests staggered.

thisisnotashwin: AHA!! good find!! thanks for the update Nicoleta!!

PR #628, opened 2021-08-20
Fix issue where VPCs weren't deleted

lkysow:
There's a race condition were even though the NAT gateway was just
deleted, AWS won't delete the VPC because it still sees the NAT gateway
as a dependency. This fix retries the VPC delete up to 10 times (waiting
1s between retries).

How I've tested this PR:
* I haven't been able to

* code


lkysow:
See https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/2428/workflows/7e4649a1-5cbc-4a19-86a7-725e54f99dfc/jobs/14170. When I went into the console this morning the VPC deleted cleanly so I think it's just a race condition.

ishustava (approved):

ndhanushkodi (approved):
Nice fix!

PR #630, opened 2021-08-22
Add namespace to all templates

ndhanushkodi:
Changes proposed in this PR:
- Add namespace to all templates. We saw this error when deploying with the CLI to the `consul` namespace, and a few resources were getting created in `default`. This only occurs using the Helm Go SDK, not the Helm CLI.
- NOTE: The test is not a catch all. It only greps for "namespace: " in every file. This means, if that string exists anywhere in the file, it won't error out. We couldn't use `yq` because this is still templated yaml. It would catch a majority of cases. If you have suggestions for making this better please let me know!

How I've tested this PR:
acceptance and unit tests.

review, make any suggestions to testing

Checklist:
- [x] Bats tests added


lkysow (approved):

t-eckert (approved):
LGTM

PR #691, opened 2021-08-27
Move hack scripts to top-level 

lkysow:
* Move some hack scripts to top level that apply to the whole repo
* Also add Makefile to top-level for the targets that moved
* Add target for aws acceptance tests
* Delete unused circleci config in charts/consul
* Update helm gen script for new directory of chart

How I've tested this PR:
* ran the targets and a test build

* code


lkysow:
> This looks good. Does it make sense to migrate the crds-to-consul-helm in control-plane/hack? It needs to be updated anyway and maybe we could update it to be a top level hack scrip.t

Yeah that makes sense!

thisisnotashwin (approved):
Excellent work @lkysow !! Thanks for doing this

ishustava (approved):
Sorry for the delay! This looks good!

PR #696, opened 2021-08-30
Create Admin Partitions when enabled

thisisnotashwin:
Changes proposed in this PR:
- Create a partition when Admin Partitions are enabled.
- The Kubernetes cluster that has servers (server cluster) will have clients installed in the default partition.
- Workload clusters will have clients deployed in the partition whose name is provided in the values file.

How I've tested this PR:
- Deployed the servers using this values file
```
global:
  tls:
    enabled: true
  image: ashwinvenkatesh/consul@sha256:08b61b39c2c45679f328ff618626540981261d37abb0e96246fa965f4c60e163
  imageK8S: ashwinvenkatesh/consul-k8s@sha256:fd9f407059514cc986369228f65d6ff3e127b9e37af0dd2513a494e12be3b5e3
  adminPartitions:
    enabled: true
server:
  exposeGossipAndRPCPorts: true
  enterpriseLicense:
    secretName: license
    secretKey: key
connectInject:
  enabled: true
controller:
  enabled: true
ui:
  service:
    type: NodePort
    nodePort:
      https: 30100
```
Things of note are this is a Consul Enterprise image and the license was provided as a secret. Additionally, an explicit nodePort has been assigned to the UI service. This is to expose Consul's https port.

Now, copy the TLS secrets and the Ent License to the workload-cluster.
```
global:
  enabled: false
  image: ashwinvenkatesh/consul@sha256:08b61b39c2c45679f328ff618626540981261d37abb0e96246fa965f4c60e163
  imageK8S: ashwinvenkatesh/consul-k8s@sha256:fd9f407059514cc986369228f65d6ff3e127b9e37af0dd2513a494e12be3b5e3
  adminPartitions:
    enabled: true
    name: "alpha"
  tls:
    enabled: true
    caCert:
      secretName: consul-consul-ca-cert
      secretKey: tls.crt
    caKey:
      secretName: consul-consul-ca-key
      secretKey: tls.key
server:
  enterpriseLicense:
    secretName: license
    secretKey: key
externalServers:
  enabled: true
  hosts: ["10.128.0.50"]
  httpsPort: 30100
  tlsServerName: server.dc1.consul
client:
  enabled: true
  exposeGossipPorts: true
  join:
  - "10.128.0.50"
  - "10.128.0.52"
  - "10.128.15.192"
connectInject:
  enabled: true
controller:
  enabled: true
```
The name of the partition here is "alpha".  The ips specified under `hosts` and `join` are internal IPs for the nodes on GKE. the node IPs can be found using the `kubectl get nodes -o wide` command. Additionally, the firewalls rules on GKE needed to be updated to ensure that the node network and pod network from the server cluster and workload cluster could communicate all the nodes.

The clients logs in the server client state that they are in the `default` partition and the logs in the workload cluster state that they are in the `alpha` partition.
Validated the Connect works on the default cluster. Connect needs to make some changes in order for it to be supported on workload clusters and that is being worked on.

Follow the above steps. DM me if you would like to pair on the above. Re-do the above steps and deploy "connect" apps in the default cluster.

Checklist:
- [x] Tests added


control-plane/subcommand/partition-init/command_ent_test.go
```diff
@@ -0,0 +1,74 @@
+//go:build enterprise
```
lkysow, line 1: do we need this one? The others don't have that.

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+	c.flags.UintVar(&c.flagServerPort, "server-port", 8500, "The HTTP or HTTPS port of the Consul server. Defaults to 8500.")
+	c.flags.StringVar(&c.flagConsulCACert, "consul-ca-cert", "",
+		"Path to the PEM-encoded CA certificate of the Consul cluster.")
+	c.flags.StringVar(&c.flagConsulTLSServerName, "consul-tls-server-name", "",
+		"The server name to set as the SNI header when sending HTTPS requests to Consul.")
+	c.flags.BoolVar(&c.flagUseHTTPS, "use-https", false,
+		"Toggle for using HTTPS for all API calls to Consul.")
+
+	c.flags.DurationVar(&c.flagTimeout, "timeout", 10*time.Minute,
+		"How long we'll try to bootstrap ACLs for before timing out, e.g. 1ms, 2s, 3m")
```
lkysow, line 72: says "bootstrapACLs"

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	return nil
+}
+
+const synopsis = "Initialize an Admin Partition on Consul."
+const help = `
+Usage: consul-k8s-control-plane partition-init [options]
+
+  Bootstraps Consul with non-default Admin Partitions.
+  It will run indefinitely the partition has been created. It is idempotent
```
lkysow, line 205: Won't it run until it's timeout?

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+func (c *Command) Help() string {
+	c.once.Do(c.init)
+	return c.help
+}
+
+// Run bootstraps ACLs on Consul servers and writes the bootstrap token to a
+// Kubernetes secret.
+// Given various flags, it will also create policies and associated ACL tokens
+// and store the tokens as Kubernetes Secrets.
+// The function will retry its tasks indefinitely until they are complete.
```
lkysow, line 100: Incorrect docs

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+		Token:   "", // TODO: Figure out if a token is required here and which one that is.
+		TLSConfig: api.TLSConfig{
+			Address: c.flagConsulTLSServerName,
+			CAFile:  c.flagConsulCACert,
+		},
+	})
+	if err != nil {
+		c.log.Error(fmt.Sprintf("Error creating Consul client for addr %q: %s", serverAddr, err))
+		return 1
+	}
```
lkysow, line 158: Did you contemplate factoring out the duplication here between server-acl-init?

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+
+	partition, _, err := consulClient.Partitions().Read(c.cmdTimeout, c.flagPartitionName, &api.QueryOptions{})
```
lkysow, line 160: nit: can pass `nil` for query opts

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+		Address: serverAddr,
+		Scheme:  scheme,
+		Token:   "", // TODO: Figure out if a token is required here and which one that is.
+		TLSConfig: api.TLSConfig{
+			Address: c.flagConsulTLSServerName,
+			CAFile:  c.flagConsulCACert,
+		},
+	})
+	if err != nil {
+		c.log.Error(fmt.Sprintf("Error creating Consul client for addr %q: %s", serverAddr, err))
```
lkysow, line 156: Looks like the pattern you're following is to use `c.UI` for fatal issues and `c.log` for runtime logs. By that pattern this should use `c.UI`

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	partition, _, err := consulClient.Partitions().Read(c.cmdTimeout, c.flagPartitionName, &api.QueryOptions{})
+	// The API does not return an error if the Partition does not exist. It returns a nil Partition.
+	if partition == nil {
+		for {
+			// Retry Admin Partition creation until it succeeds, or we reach the command timeout.
+			_, _, err = consulClient.Partitions().Create(c.cmdTimeout, &api.AdminPartition{Name: c.flagPartitionName}, &api.WriteOptions{})
+			if err == nil {
+				c.log.Info(fmt.Sprintf("Successfully created Admin Partition '%s'", c.flagPartitionName))
```
lkysow, line 167: hclog supports k/v pairs so this should read `c.log.Info("Successfully created admin partition", "name", c.flagPartitionName)

lkysow: same for the other logs (you can k/v them)

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+	flagConsulCACert        string
+	flagConsulTLSServerName string
+	flagUseHTTPS            bool
+
+	flagLogLevel string
+	flagLogJSON  bool
+	flagTimeout  time.Duration
+
+	// cmdTimeout is cancelled when the command timeout is reached.
+	cmdTimeout    context.Context
```
lkysow, line 43: @t-eckert and I actually refactored this to just be `ctx` in server-acl-init

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+				continue
+			case <-c.cmdTimeout.Done():
+				return 1
+			}
+		}
+	} else if err != nil {
+		c.log.Error(fmt.Sprintf("Error reading Partition '%s' from Consul: %s", c.flagPartitionName, err))
+		return 1
+	}
+	c.log.Info(fmt.Sprintf("Admin Partition '%s' exists. Exiting.", c.flagPartitionName))
```
lkysow, line 185: Would be nice if it said "ap already exists" if it already exits or "created successfully" if it gets created just so I know exactly what happened

control-plane/subcommand/partition-init/command_ent_test.go
```diff
@@ -0,0 +1,74 @@
...
+	cmd.init()
+	args := []string{
+		"-server-address=" + strings.Split(a.HTTPAddr, ":")[0],
+		"-server-port=" + strings.Split(a.HTTPAddr, ":")[1],
+		"-partition-name", partitionName,
+	}
+
+	responseCode := cmd.Run(args)
+
+	require.Equal(t, 0, responseCode)
```
lkysow, line 71: You should assert that the partition is actually created.

control-plane/subcommand/partition-init/command_ent_test.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// TODO: Write tests with ACLs enabled
```
lkysow, line 74: Needs tests:
* if the partition already exists it works
* times out after the timeout

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,208 @@
...
+			case <-time.After(c.retryDuration):
+				continue
+			case <-c.cmdTimeout.Done():
+				return 1
+			}
+		}
+	} else if err != nil {
+		c.log.Error(fmt.Sprintf("Error reading Partition '%s' from Consul: %s", c.flagPartitionName, err))
+		return 1
+	}
```
lkysow, line 184: thinking: is there a use-case where we want it to retry even if at first it can't talk to consul for some reason? Like what if the consul server isn't up yet for some reason?

lkysow (commented):
🎉 looking good

charts/consul/test/unit/partition-init-podsecuritypolicy.bats
```diff
@@ -0,0 +1,48 @@
...
+      --set 'global.adminPartitions.enabled=true' \
+      --set 'server.enabled=false' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "partitionInit/PodSecurityPolicy: disabled with global.adminPartitions.enabled=true and servers = true" {
+  cd `chart_dir`
+ assert_empty helm template \
```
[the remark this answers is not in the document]
thisisnotashwin: fixed!! im upset that auto-indent doesnt fix this.

control-plane/go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/google/go-cmp v0.5.6
 	github.com/google/go-querystring v1.0.0 // indirect
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
-	github.com/hashicorp/consul/api v1.9.0
+	github.com/hashicorp/consul/api v1.4.1-0.20210827004034-d2e50fd130ae
```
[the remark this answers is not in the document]
thisisnotashwin: yeah..i wish i knew 😞 It is supposed to use tags but I don't think its doing that.

control-plane/subcommand/partition-init/command_ent_test.go
thisisnotashwin: yeah. that happened accidently.

control-plane/subcommand/partition-init/command.go
thisisnotashwin: I think it does that right now. I'll change the phrasing to add "already" to the statement. When an AP is created, there is a different log line that says that a partition was create and then returns. (line 168)

control-plane/subcommand/partition-init/command.go
thisisnotashwin: makes sense

charts/consul/values.yaml
```diff
@@ -29,6 +29,18 @@ global:
...
+  # It additionally indicates that you are running Consul Enterprise v1.11+ with a valid Consul Enterprise
+  # license and would like to make use of configuration beyond registering everything into
+  # the `default` Consul partition.
+  adminPartitions:
+    # If true, the Helm chart will enable Admin Partitions for the cluster. The clients in the server cluster
+    # must be installed in the default partition.
+    enabled: false
+    # The name of the Admin Partition. Must be "default" in the server cluster ie the Kubernetes cluster that
+    # the Consul server pods are deployed onto.
+    name: "default"
```
lkysow, line 42: Do you do any validation that if servers are enabled that this == default?

control-plane/subcommand/common/common.go
```diff
@@ -115,3 +118,44 @@ func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
 	}
 	return os.Chmod(outputFile, mode)
 }
+
+// GetHTTPSScheme return the string value "https" if the bool https is true and "http" if it is false.
+func GetHTTPSScheme(https bool) string {
```
lkysow, line 123: nit: seems overkill to function this

control-plane/subcommand/common/common.go
```diff
@@ -115,3 +118,44 @@ func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
...
+func GetResolvedServerAddresses(serverAddresses []string, providers map[string]discover.Provider, logger hclog.Logger) ([]string, error) {
+	var err error
+	if len(serverAddresses) == 1 && strings.Contains(serverAddresses[0], "provider=") {
+		serverAddresses, err = godiscover.ConsulServerAddresses(serverAddresses[0], providers, logger)
+		if err != nil {
+			logger.Error(fmt.Sprintf("Unable to discover any Consul addresses from %q: %s", serverAddresses[0], err))
+			return nil, err
+		}
+	}
+	return serverAddresses, nil
```
lkysow, line 141:
```suggestion
	if len(serverAddresses) != 1 || !strings.Contains(serverAddresses[0], "provider=") {
		return serverAddresses, nil
	}
	return godiscover.ConsulServerAddresses(serverAddresses[0], providers, logger)
```

just nits, take em or leave em:
1. Don't see the need to re-use the serverAddresses variable, can redefine a new one if you end up using godiscover
2. logging makes sense to me in the caller. So basically functions return errs and it's up to the callers to log. Otherwise you end up with duplicate logs (both caller and callee) or you end up diving into functions so see if they log the errors
3. exit-early on simple cases pattern
4. hoping we can just return the godiscver error straight up, but maybe it's not very readable?

control-plane/subcommand/common/common.go
```diff
@@ -115,3 +118,44 @@ func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
...
+		Address: serverAddr,
+		Scheme:  scheme,
+		Token:   token,
+		TLSConfig: api.TLSConfig{
+			Address: tlsServerName,
+			CAFile:  consulCACert,
+		},
+	})
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error creating Consul client for addr %q: %s", serverAddr, err))
```
lkysow, line 157: see logging nit above

lkysow: also if you don't log then you can just return from `consul.NewClient` directly

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+	for {
+		partition, _, err := consulClient.Partitions().Read(c.ctx, c.flagPartitionName, nil)
+		// The API does not return an error if the Partition does not exist. It returns a nil Partition.
+		if err != nil {
+			c.log.Error("Error reading Partition from Consul", "name", c.flagPartitionName, "error", err.Error())
+			c.log.Info("Retrying in " + c.retryDuration.String())
```
lkysow, line 137: retry log can be put just before the select

control-plane/subcommand/partition-init/command.go
```diff
@@ -0,0 +1,185 @@
...
+			// Retry Admin Partition creation until it succeeds, or we reach the command timeout.
+			_, _, err = consulClient.Partitions().Create(c.ctx, &api.AdminPartition{
+				Name:        c.flagPartitionName,
+				Description: "Created by Helm installation",
+			}, nil)
+			if err == nil {
+				c.log.Info("Successfully created Admin Partition", "name", c.flagPartitionName)
+				return 0
+			}
+			c.log.Error(fmt.Sprintf("Error creating partition '%s': '%s'", c.flagPartitionName, err))
```
lkysow, line 148: same comment about the k/v pairs for logs here

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -355,19 +342,8 @@ func (c *Command) Run(args []string) int {
...
-	consulClient, err := consul.NewClient(&api.Config{
-		Address: serverAddr,
-		Scheme:  scheme,
-		Token:   bootstrapToken,
-		TLSConfig: api.TLSConfig{
-			Address: c.flagConsulTLSServerName,
-			CAFile:  c.flagConsulCACert,
-		},
-	})
+	consulClient, err := common.GetConsulClientForServer(serverAddresses, c.flagServerPort, scheme, bootstrapToken, c.flagConsulTLSServerName, c.flagConsulCACert, c.log)
```
lkysow, line 345: looking at it now, this function was probably not needed since it basically just constructs this struct but it's fine too

lkysow (approved):

charts/consul/values.yaml
thisisnotashwin: Not right now. I wanted to do this in a follow up task as it is not a must have. Have a task for it: https://app.asana.com/0/1200449541311255/1200910791029388/f

control-plane/subcommand/common/common.go
thisisnotashwin: yeah thats fair. It was a fall out of trying to do a larger refactor and then needed to trim it down.

control-plane/subcommand/server-acl-init/command.go
thisisnotashwin: yeah.. i was middle of the refactor when I kinda realized that it was probably overkill. Ill revert it.

ndhanushkodi (commented):
Looking awesome so far! I've just gotten through reviewing the Helm templates and some of the control-plane code and will continue tomorrow! Barely had comments because it's looking really great!

Other qs that popped up:
Do we explicitly require TLS to enable partitions or just highly recommend?

thisisnotashwin:
Setting up external servers explicitly requires TLS. This is more a safety consideration as compared to Consul not working without it. So in a certain sense, we do not support non-TLS on k8s with externalServers. Are you thinking about adding that as an additional check?

ndhanushkodi (approved):
> we do not support non-TLS on k8s with externalServers. Are you thinking about adding that as an additional check?
It could be worth the additional check, ok with that being a follow on task though!

Awesome work Ashwin!!!

control-plane/subcommand/common/common.go
```diff
@@ -115,3 +117,11 @@ func WriteFileWithPerms(outputFile, payload string, mode os.FileMode) error {
 	}
 	return os.Chmod(outputFile, mode)
 }
+
+// GetResolvedServerAddresses resolves the Consul server address if it has been provided a provider else it returns the server addresses that were input to it.
+func GetResolvedServerAddresses(serverAddresses []string, providers map[string]discover.Provider, logger hclog.Logger) ([]string, error) {
+	if len(serverAddresses) != 1 || !strings.Contains(serverAddresses[0], "provider=") {
```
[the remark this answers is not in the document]
thisisnotashwin: Sure thing :)
There could be a case where we have a single address which is not a provider. For instance, with the Partition service, the single address would be the address of the partition service LB. Ill improve the text here.

PR #704, opened 2021-08-31
Disable openshift tests until we can use UBI images

ishustava:
Changes proposed in this PR:
Currently, all tests fail because the user is not allowed to `mkdir /consul/extra-config` ([link](https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/2503/workflows/2bf83d1d-baf1-4024-907a-7d2385dd08bc/jobs/15161)) in the consul server and client pods. We need to wait for UBI images as that problem is fixed in that docker image.

How I've tested this PR:

Checklist:
- [ ] Tests added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -946,9 +946,10 @@ workflows:
       - cleanup-gcp-resources
       - cleanup-azure-resources
       - cleanup-eks-resources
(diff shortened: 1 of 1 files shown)


thisisnotashwin (approved):
lol..again

PR #706, opened 2021-08-31
Add support for Partition Service

thisisnotashwin:
This is a LoadBalancer service that allows one to expose the server pods. This allows a constant value for client auto-join even if the node IPs of the server pods change.

How I've tested this PR: Bats and by deploying the service.

How I expect reviewers to test this PR: Code Review

When testing, ensure the enterprise license and server cert/key are migrated to the workload cluster. Ensure firewall rules allow communication between node and pod networks of all clusters involved.
Sample Manifests:
servers.yaml
```
global:
  tls:
    enabled: true
  image: ashwinvenkatesh/consul@sha256:08b61b39c2c45679f328ff618626540981261d37abb0e96246fa965f4c60e163
  imageK8S: ashwinvenkatesh/consul-k8s@sha256:fd9f407059514cc986369228f65d6ff3e127b9e37af0dd2513a494e12be3b5e3
  adminPartitions:
    enabled: true
server:
  exposeGossipAndRPCPorts: true
  enterpriseLicense:
    secretName: license
    secretKey: key
connectInject:
  enabled: true
controller:
  enabled: true
```

partition.yaml
```
global:
  enabled: false
  image: ashwinvenkatesh/consul@sha256:08b61b39c2c45679f328ff618626540981261d37abb0e96246fa965f4c60e163
  imageK8S: ashwinvenkatesh/consul-k8s@sha256:fd9f407059514cc986369228f65d6ff3e127b9e37af0dd2513a494e12be3b5e3
  adminPartitions:
    enabled: true
    name: "alpha"
  tls:
    enabled: true
    caCert:
      secretName: consul-consul-ca-cert
      secretKey: tls.crt
    caKey:
      secretName: consul-consul-ca-key
      secretKey: tls.key
server:
  enterpriseLicense:
    secretName: license
    secretKey: key
externalServers:
  enabled: true
  hosts: ["*.*.*.*"] // the loadbalancer external IP
  tlsServerName: server.dc1.consul
client:
  enabled: true
  exposeGossipPorts: true
  join: ["*.*.*.*"] // the loadbalancer external IP
connectInject:
  enabled: true
controller:
  enabled: true
```

Checklist:
- [x] Tests added


charts/consul/values.yaml
```diff
@@ -32,6 +32,17 @@ global:
...
+    # Partition service properties.
+    service:
+      # Annotations to apply to the partition service.
+      #
+      # ```yaml
+      # annotations: |
+      #   "annotation-key": "annotation-value"
+      # ```
+      #
+      # @type: string
```
kschoche, line 44: Is "string" correct as a type? wouldnt this be a map[string]string?

charts/consul/templates/partition-service.yaml
```diff
@@ -0,0 +1,58 @@
...
+spec:
+  type: LoadBalancer
+  # We want the servers to become available even if they're not ready
+  # since this DNS is also used for join operations.
+  publishNotReadyAddresses: true
+  ports:
+    {{- if (or (not .Values.global.tls.enabled) (not .Values.global.tls.httpsOnly)) }}
+    - name: http
+      port: 8500
+      targetPort: 8500
```
ishustava, line 35: Right now externalServers section doesn't support http. Do we plan to support it for partitions? If we do, it should probably come in a later PR.

charts/consul/templates/partition-service.yaml
```diff
@@ -0,0 +1,58 @@
...
+    release: {{ .Release.Name }}
+    component: server
+  annotations:
+    {{- if .Values.global.adminPartitions.service.annotations }}
+    {{ tpl .Values.global.adminPartitions.service.annotations . | nindent 4 | trim }}
+    {{- end }}
+    # This must be set in addition to publishNotReadyAddresses due
+    # to an open issue where it may not work:
+    # https://github.com/kubernetes/kubernetes/issues/58662
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
```
ishustava, line 25: do we need this for the partition service? We'd be joining the servers after they're ready from another partition.

ishustava (commented):
Great work, Ashwin!!! This looks fantastic 😃 

I've left some comments and questions, the main one being about supporting `kind`. 

I did try it with a NodePort service on kind, and it worked beautifully!

charts/consul/values.yaml
thisisnotashwin: We provide the input as a giant `string` as we use the `|` directive to begin. This is how annotations have been defined across the board in our values file TBH.

charts/consul/templates/partition-service.yaml
```diff
@@ -0,0 +1,58 @@
...
+      targetPort: 8301
+    - name: serfwan-tcp
+      port: 8302
+      targetPort: 8302
+    - name: server
+      port: 8300
+      targetPort: 8300
+    - name: dns-tcp
+      port: 8600
+      targetPort: dns-tcp
```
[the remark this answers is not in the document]
thisisnotashwin: you are absolutely right. I copy-pasted this and did not really look into the specific of which ports were retained or removed. This is a good callout. Will remove the serfwan port as well.

charts/consul/templates/partition-service.yaml
thisisnotashwin: Good point!

charts/consul/templates/partition-service.yaml
thisisnotashwin: I had known this and had conveniently forgotten about this. I think we should NOT support HTTP and have removed that section from the service. I will update the RFC with the specific call out.

charts/consul/templates/partition-service.yaml
```diff
@@ -0,0 +1,58 @@
...
+  annotations:
+    {{- if .Values.global.adminPartitions.service.annotations }}
+    {{ tpl .Values.global.adminPartitions.service.annotations . | nindent 4 | trim }}
+    {{- end }}
+    # This must be set in addition to publishNotReadyAddresses due
+    # to an open issue where it may not work:
+    # https://github.com/kubernetes/kubernetes/issues/58662
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
+spec:
+  type: LoadBalancer
```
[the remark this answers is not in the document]
thisisnotashwin: I like the consistency of always relying on the same service to "discover" Consul Servers if that makes sense. I think we should allow users to configure the type and use NodePort so that it is supported on `kind`. I have also enabled the ability to select a nodePort for https, rpc and serf

ishustava (approved):
🎉 Looks great! Thank you making the updates!

charts/consul/templates/partition-service.yaml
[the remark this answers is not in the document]
ishustava: That makes sense to me 😃

kschoche (approved):
Looks great!!

PR #707, opened 2021-08-31
Use command struct ctx in server-acl-init

t-eckert:
## Commits

- Change to single `ctx` in `connect_inject_test`
- Use `ctx` from command struct

- Instead of using `context.TODO()` throughout the `server-acl-init` package, use the `ctx` (previously called `cmdTimeout`) which is created within the `Run` function.

## How I've tested this PR:

I still see test failures when I run this locally. I think I have something configured improperly or am missing a dependency. I'll sync up with @lkysow tomorrow and figure it out.

Check that the rename of `cmdTimeout` to `ctx` doesn't break anything. I think it should be ok. Also just run the tests for `control-plane` 🤷 .

### DIFF
--- a/control-plane/subcommand/server-acl-init/command.go
+++ b/control-plane/subcommand/server-acl-init/command.go
@@ -91,8 +91,8 @@ type Command struct {
 
 	clientset kubernetes.Interface
 
-	// cmdTimeout is cancelled when the command timeout is reached.
-	cmdTimeout    context.Context
+	// ctx is cancelled when the command timeout is reached.
+	ctx           context.Context
 	retryDuration time.Duration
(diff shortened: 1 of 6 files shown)


thisisnotashwin (approved):
Excellent work!!! These clean-ups make me erally happy!!

t-eckert:
> Looks great!! Some of the tests still use `context.Background` multiple times but thinking about it again I don't think the need to necessarily use a shared `ctx` object throughout a single test because the context is never actually being cancelled, golang is just killing the test if it times out.

Cool, if we want to clean this up as the work progresses, we can make that call tomorrow.

PR #710, opened 2021-09-01
Use root context in sync-catalog to-k8s

t-eckert:
Changes proposed in this PR:
- Add a context to the `K8SSink` struct
- Use the context in the struct for processes spawned by `K8SSink`
- Update tests on `K8SSink` to pass in a `context.Background()`
- Update the `sync-catalog` command which instantiates and uses `K8SSink` to pass in the `ctx` created when in its `Run` function

How I've tested this PR:

I ran the unit tests for the `catalog` and `sync-catalog` packages. They passed :).

Unit tests

Checklist:
- [x] Tests added (modified)
- [-] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

### DIFF
--- a/control-plane/catalog/to-k8s/sink.go
+++ b/control-plane/catalog/to-k8s/sink.go
@@ -38,7 +38,7 @@ type Sink interface {
 
 // K8SSink is a Sink implementation that registers services with Kubernetes.
 //
-// K8SSink also implements controller.Resource and is meant to run as a K8S
+// K8SSink also implements controller. Resource and is meant to run as a K8S
 // controller that watches services. This is the primary way that the
 // sink should be run.
 type K8SSink struct {
@@ -50,6 +50,9 @@ type K8SSink struct {
 	// services in Kubernetes. This can be fairly short since no work will be
 	// done if there are no changes.
 	SyncPeriod time.Duration
+
+	// Ctx is used to cancel the Sink
+	Ctx context.Context
 
 	// lock gates concurrent access to all the maps.
 	lock sync.Mutex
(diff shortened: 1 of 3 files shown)


control-plane/catalog/to-k8s/sink.go
```diff
@@ -38,7 +38,7 @@ type Sink interface {
 
 // K8SSink is a Sink implementation that registers services with Kubernetes.
 //
-// K8SSink also implements controller.Resource and is meant to run as a K8S
+// K8SSink also implements controller. Resource and is meant to run as a K8S
```
kschoche, line 41: I think this is saying `controller.Resource` and is not a sentence

control-plane/catalog/to-k8s/sink.go
```diff
@@ -51,6 +51,9 @@ type K8SSink struct {
 	// done if there are no changes.
 	SyncPeriod time.Duration
 
+	// Ctx is used to cancel the Sink
```
kschoche, line 54:
```suggestion
	// Ctx is used to cancel the Sink.
```

control-plane/catalog/to-k8s/sink.go
t-eckert: Oh. Oops.

kschoche (approved):
This looks great, I left just a couple comments.
I wanted to also just leave a general thought: Given that the previous flow of this code basically assumed that the context wouldnt do anything (`context.TODO()`) under any circumstances and we're adding logic where we are expecting that passing a context *should do something* does it make sense to add some testing somewhere?
We are technically changing the expected behaviour?

kschoche:
> > This looks great, I left just a couple comments.
> > I wanted to also just leave a general thought: Given that the previous flow of this code basically assumed that the context wouldnt do anything (`context.TODO()`) under any circumstances and we're adding logic where we are expecting that passing a context _should do something_ does it make sense to add some testing somewhere?
> > We are technically changing the expected behaviour?
> 
> That's a great point that I didn't really stop to reflect on. I'll take a look at where it would be meaningful to add tests.

I can't honestly think of a good testing scenario here that doesn't just test the k8s APIs (unnecessary) handling of ctx so it's probably okay to ignore for now, but just wanted to raise that as a something to brainstorm going forward!

control-plane/catalog/to-k8s/sink.go
```diff
@@ -51,6 +51,9 @@ type K8SSink struct {
 	// done if there are no changes.
 	SyncPeriod time.Duration
 
+	// Ctx is used to cancel the Sink.
+	Ctx context.Context
```
lkysow, line 55: can this be private (lowercase)?

lkysow (approved):
This does look pretty hard to test, but a good call out. I think I'm okay without a test.

control-plane/catalog/to-k8s/sink.go
t-eckert: I tested it and it doesn't work if private. If I'm wrong, we can swing back around and fix it.

Issue #655, opened 2021-01-27
helm:Support ability to configure `consul-webhook-cert-manager` deployment tolerations


HofmannZ:
#### Question

We are spinning up Consul on GKE (Kubernetes 1.19) and noticed that the `consul-connect-injector-webhook-deployment` is not injecting sidecar proxies.

It's giving the following error:

```
[Info]
2021-01-27 04:05:05.870 CETSuccessfully wrote Consul client CA to: /consul/tls/client/ca/tls.crt
[Info]
2021-01-27 04:05:07.673 CETListening on ":8080"...
[Info]
2021-01-27 04:05:08.075 CETUpdated certificate bundle received. Updating certs...
[Error]
2021-01-27 04:05:08.172 CET2021/01/27 03:05:08 http: TLS handshake error from 10.2.1.1:57248: No certificate available.
```

Then we cleaned up the cluster and ran it again with `logLevel: debug`:

```
Error
2021-01-27 04:46:23.415 CET2021-01-27T03:46:23.415Z [ERROR] Error retrieving CA roots from Consul: err="Get "https://consul-server:8501/v1/agent/connect/ca/roots": dial tcp: lookup consul-server on 192.168.0.10:53: no such host"
Error
2021-01-27 04:46:24.462 CET2021-01-27T03:46:24.461Z [ERROR] Error retrieving CA roots from Consul: err="Get "https://consul-server:8501/v1/agent/connect/ca/roots": dial tcp: lookup consul-server on 192.168.0.10:53: no such host"
Info
2021-01-27 04:46:31.109 CETSuccessfully wrote Consul client CA to: /consul/tls/client/ca/tls.crt
Info
2021-01-27 04:46:32.642 CETListening on ":8080"...
Error
2021-01-27 04:46:32.642 CET2021-01-27T03:46:32.642Z [DEBUG] healthCheckResource: starting reconcile
Info
2021-01-27 04:46:32.941 CETUpdated certificate bundle received. Updating certs...
Error
2021-01-27 04:46:33.245 CET2021-01-27T03:46:33.245Z [DEBUG] healthCheckResource: finished reconcile
Error
2021-01-27 04:46:33.246 CET2021-01-27T03:46:33.245Z [DEBUG] healthCheckController: initial cache sync complete
```

This time we did not get the error message nor did we get a confirmation message (but we're not sure if we should get one in the first place.)

Our config:

```yaml
global:
  name: consul
  gossipEncryption:
[...]

PR #712, opened 2021-09-02
Add config for `webhook-cert-manager` tolerations

t-eckert:
Changes proposed in this PR:
- Add configuration for the webhook-cert-manager deployment tolerations

This work closes issue #655 where the user was running Consul on a private cluster in GKE.

How I've tested this PR:
Added bats and manually produced the deployment YAML using helm.

Run bats

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ IMPROVEMENTS:
   * Add readiness, liveness and startup probes to the connect inject deployment. [[GH-626](https://github.com/hashicorp/consul-k8s/pull/626)][[GH-701](https://github.com/hashicorp/consul-k8s/pull/701)]
   * Add support for setting container security contexts on client and server Pods. [[GH-620](https://github.com/hashicorp/consul-k8s/pull/620)]
   * Update Envoy image to 1.18.4 [[GH-699](https://github.com/hashicorp/consul-k8s/pull/699)]
+  * Add configuration for webhook-cert-manager tolerations [[GH-XXX]]()
 * Control Plane
   * Add health endpoint to the connect inject webhook that will be healthy when webhook certs are present and not empty. [[GH-626](https://github.com/hashicorp/consul-k8s/pull/626)]
   * Catalog Sync: Fix issue registering NodePort services with wrong IPs when a node has multiple IP addresses. [[GH-619](https://github.com/hashicorp/consul-k8s/pull/619)]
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ IMPROVEMENTS:
   * Add readiness, liveness and startup probes to the connect inject deployment. [[GH-626](https://github.com/hashicorp/consul-k8s/pull/626)][[GH-701](https://github.com/hashicorp/consul-k8s/pull/701)]
   * Add support for setting container security contexts on client and server Pods. [[GH-620](https://github.com/hashicorp/consul-k8s/pull/620)]
   * Update Envoy image to 1.18.4 [[GH-699](https://github.com/hashicorp/consul-k8s/pull/699)]
+  * Add configuration for webhook-cert-manager tolerations [[GH-XXX]]()
```
lkysow: todo

charts/consul/values.yaml
```diff
@@ -2245,6 +2245,15 @@ terminatingGateways:
   gateways:
     - name: terminating-gateway
 
+# Configuration settings for the webhook-cert-manager
```
lkysow, line 2248: can we expand on this a bit and talk about what the cert manager does? Just so folks aren't confused about what exact certs it's managing.

t-eckert: @kschoche What is your take on what I added?

charts/consul/values.yaml
```diff
@@ -2245,6 +2245,16 @@ terminatingGateways:
...
+# webhook-cert-manager ensures that cert bundles are up to date for mutating webhooks
```
kschoche, line 2249:
```suggestion
# `webhook-cert-manager` ensures that cert bundles are up to date for the mutating webhook.
```

kschoche: Trying to decide if we capitalize it or wrap it in quotes .
Thoughts?

ndhanushkodi (approved):
Looks great Thomas!!

I think the style of tolerations has varied throughout our values.yaml, with some being set to "" and some being set to `null`, but I think since the inconsistency exists already, and this matches one of the patterns, it still works. For example  see https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/values.yaml#L885-L895

t-eckert:
@ndhanushkodi, I noticed that too. I thought it might be a good onboarding task to pick a standard and I can go through and change all of the nullish instances to use that in a later PR.

PR #713, opened 2021-09-03
cli: support install command 

ndhanushkodi:
(This description is lightly edited from the original PR Saad created: https://github.com/hashicorp/consul-k8s-cli/pull/2 We migrated the PR here and migrated some of the review comments as well)

Changes proposed in this PR:
 
* `consul-k8s install` installs the latest Helm chart onto your Kubernetes cluster after performing a few verification steps. Note that once the CLI is moved into the monorepo, the version of the Helm chart deployed will match the version in the repository.
* For details about flags and their intended behavior visit the RFC, linked [here](https://docs.google.com/document/d/13vBeED9OkKUO3isZWIxTRvbH-sxEzvYH9rdJeb9my48/edit#).
* `common/flag` is copied from Waypoint’s internal/pkg subdirectory [here](https://github.com/hashicorp/waypoint/tree/main/internal/pkg/flag). It is a thin layer over the standard Go flag package that provides features such as aliasing and global flags. It was created specifically for mitchellh/cli. See also [this](https://hashicorp.slack.com/archives/C0E7S0EU8/p1628019710015500) Slack thread for the decision to use it.
* `common/terminal` package is adapted from [the Waypoint Plugin SDK](https://github.com/hashicorp/waypoint-plugin-sdk/tree/74d9328929293551499078da388b8d057f3b2341/terminal) and excludes the Status interface and Step interface as described in [doc.go](https://github.com/hashicorp/consul-k8s-cli/blob/install/cmd/common/terminal/doc.go). It allows us to have a consistent UI output with style.
* `common/base.go` describes a base command that can be embedded into all commands, and pass in common functionality like UI and ctx. This pattern was adapted from Waypoint.
* ~`common/utils.go` contains a function `WithInterrupt()` that we take from [signalcontext.go](https://github.com/hashicorp/waypoint/blob/348d2c77fce199952618ccef6433c8844b22583b/internal/pkg/signalcontext/signalcontext.go) in Waypoint's internal/pkg.~
 
How I've tested this PR:
 
* Created rather simple unit tests that validate two different helper functions inside of install.go. The first helper function we test is `validateFlags()` which performs a lot of the basic sanity checks before starting the install. The second helper is `validLabel()` which makes sure that the namespace the user passes is legal, in that if follows [RFC 1123 label convention](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/).
* In addition to the actual unit tests, a lot of manual ad-hoc testing took place. Follow the instructions below to build the CLI and test the various flags, especially the -set ones, and see how they interact with one another.
* Note that more rigorous acceptance tests are soon to come after monorepo is finished. These tests will actually run the installation through and validate it worked properly.
 
* Try it out :
       `git clone https://github.com/hashicorp/consul-k8s-cli.git`
       `cd consul-k8s-cli`
       `go build -o consul-k8s`
       `consul-k8s install`
 
* Try this with different flags and various configurations. Note that if you are running on minikube, it is recommended to install with `-preset=demo`. That way server.replicas is set to 1. I also found it helpful to specify the namespace when manually testing via `-namespace=not-peppertrout`.
 
* Code review :
       Pay careful attention to install.go, specifically the pre-install checks. If you see anything you believe should be pulled out into an individual function and/or tested more thoroughly please comment!

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 
  
Co-authored-by: Saad Jamal [saad@hashicorp.com](mailto:saad@hashicorp.com)
Co-authored-by: Nitya Dhanushkodi [nitya@hashicorp.com](mailto:nitya@hashicorp.com)


cli/README.md
```diff
@@ -0,0 +1,55 @@
...
+ 	Path to a file to customize the installation, such as Consul Helm chart values file. Can be specified multiple times.
+
+  -name=<string>
+ 	Name of the installation. This will be prefixed to resources installed on the cluster.
+
+  -namespace=<string>
+ 	Namespace for the Consul installation. Defaults to “consul”.
+
+  -preset=<string>
+ 	Use an installation preset, one of default, demo, secure. Defaults to "default".
```
ndhanushkodi, line 37: @kschoche's comment: Would it make sense to go into a bit of detail on what these things actually mean?
Also a general comment that if `default` doesn't actually do anything we probably don't need it.

cli/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+  -set=<string>
+ 	Set a value to customize. Can be specified multiple times. Supports Consul Helm chart values.
+
+  -set-file=<string>
+      Set a value to customize via a file. The contents of the file will be set as the value. Can be specified multiple times. Supports Consul Helm chart values.
+
+  -set-string=<string>
+      Set a string value to customize. Can be specified multiple times. Supports Consul Helm chart values.
+
```
ndhanushkodi, line 47: @kschoche's comment: Can we add a bit more information in the readme to define the order of precedence for any of these flags.
Example:
`consul-k8s-cli install -f val.yaml -preset=secure -set="global.tls.enabled=false"` 
I can guess that this will probably break the install somehow due to helm charts themselves, but it is not clear to me what the order of precedence for these flags in general. What if I set them twice, to different values, or set conflicting flags.
I think we can solve most of these questions (for this whole file) by just expanding on how things work in the Readme file.

cli/cmd/install/install_test.go
```diff
@@ -0,0 +1,94 @@
+package install
```
ndhanushkodi, line 1: @kschoche's comment: 

These tests look great!
I know that testing CLI bits is challenging, but I'm wondering if we can add testing around skipConfirm and dryRun having expected behaviour?

Does it make sense to have some additional tests around how we're using step, and also somehow have at least a happy-path test of the actual command?

cli/cmd/install/install.go
```diff
@@ -0,0 +1,456 @@
...
+			Target:  &c.flagKubeContext,
+			Default: "",
+			Usage:   "Kubernetes context to use.",
+		})
+	}
+
+	c.help = c.set.Help()
+}
+
+func (c *Command) Run(args []string) int {
```
ndhanushkodi, line 156: @t-eckert's comment: I'm a little concerned about the length of this function. It was hard for me to keep in mind all of the context I needed as I went through it. I think it would make more sense to break the function down into several functions that are executed by the Run function.

ndhanushkodi (commented):
Migrated the review comments from https://github.com/hashicorp/consul-k8s-cli/pull/2 so we can address them here

cli/cmd/install/install_test.go
ndhanushkodi: Hey @kschoche, so I tried to give this a shot (skipConfirm and dryRun) and in theory it could have been possible by creating a fake actionConfig that fakes out the kubeclient and helm registry client. I got to the point of almost being able to create a fake actionConfig, but the Helm go SDK has the registry package under an "internal" folder, so I can't import the structs without copying that entire package. Since it would require a bit of jumping through hoops to mock out anything related to Helm, I think it may be worth covering dryRun and skipConfirm and potentially any other behaviour involving the Run() function via acceptance tests?

cli/cmd/install/install.go
ndhanushkodi: Broke this up a bit, now only actions pertaining to the main install are in the Run function.

cli/README.md
ndhanushkodi: Changed the wording here and added above this snippet a description of the presets

cli/README.md
ndhanushkodi: Added to the readme a description of the precedence ordering!

cli/cmd/install/install.go
```diff
@@ -0,0 +1,496 @@
...
+		return 1
+	}
+
+	// Handle preset, value files, and set values logic.
+	vals, err := c.mergeValuesFlagsWithPrecedence(settings)
+	if err != nil {
+		c.UI.Output(err.Error(), terminal.WithErrorStyle())
+		return 1
+	}
+	valuesYaml, err := yaml.Marshal(vals)
```
thisisnotashwin, line 244: missing err check

cli/cmd/install/install.go
```diff
@@ -0,0 +1,496 @@
...
+		c.UI.Output("Installation name: %s", DefaultReleaseName, terminal.WithInfoStyle())
+		c.UI.Output("Namespace: %s", c.flagNamespace, terminal.WithInfoStyle())
+
+		if err != nil {
+			c.UI.Output("Overrides:"+"\n"+"%+v", err, terminal.WithInfoStyle())
+		} else if len(vals) == 0 {
+			c.UI.Output("Overrides: "+string(valuesYaml), terminal.WithInfoStyle())
+		} else {
+			c.UI.Output("Overrides:"+"\n"+string(valuesYaml), terminal.WithInfoStyle())
+		}
```
thisisnotashwin, line 258: This could be the error check for the previous error but it is unclear why it is getting checked here IMO. Also, does it not exit if there was an error marshalling the YAML?

cli/cmd/install/install.go
```diff
@@ -0,0 +1,496 @@
...
+		if !(strings.ToLower(confirmation) == "y" || strings.ToLower(confirmation) == "yes") {
+			c.UI.Output("Install aborted. To learn how to customize your installation, run:\nconsul-k8s install --help", terminal.WithInfoStyle())
+			return 1
+		}
+	}
+
+	c.UI.Output("Running Installation", terminal.WithHeaderStyle())
+
+	// Setup the installation action.
+	install := action.NewInstall(actionConfig)
```
thisisnotashwin, line 293: can `actionConfig` be initialized right above this step? It doesnt appear to be used elsewhere.

cli/cmd/install/install_test.go
```diff
@@ -0,0 +1,163 @@
...
+	baseCommand := &common.BaseCommand{
+		Ctx: ctx,
+		Log: log,
+	}
+
+	c := &Command{
+		BaseCommand: baseCommand,
+	}
+	c.init()
+	c.Init()
```
thisisnotashwin, line 36: moving c.Init() inside init() will ensure you only need to invoke it once.

cli/cmd/install/install_test.go
```diff
@@ -0,0 +1,163 @@
...
+	return c
+}
+
+// TestDebugger is used to play with install.go for ad-hoc testing.
+func TestDebugger(t *testing.T) {
+	c := getInitializedCommand(t)
+	c.Run([]string{"-auto-approve", "-f=../../config.yaml"})
+}
+
+func TestCheckForPreviousPVCs(t *testing.T) {
```
thisisnotashwin, line 46: This could be migrated to being a table test.

cli/cmd/install/install_test.go
```diff
@@ -0,0 +1,163 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "irrelevant-pvc",
+		},
+	}
+	c.kubernetes.CoreV1().PersistentVolumeClaims("default").Create(context.TODO(), pvc, metav1.CreateOptions{})
+	err = c.checkForPreviousPVCs()
+	require.NoError(t, err)
+}
+
+func TestCheckForPreviousSecrets(t *testing.T) {
```
thisisnotashwin, line 81: This can be a table test as well.

cli/cmd/install/install_test.go
```diff
@@ -0,0 +1,163 @@
...
+			if err := c.validateFlags(testCase.input); err == nil {
+				t.Errorf("Test case should have failed.")
+			}
+		})
+	}
+}
+
+// TestValidLabel calls validLabel() which checks strings match RFC 1123 label convention.
+func TestValidLabel(t *testing.T) {
+	testCases := []struct {
```
thisisnotashwin, line 136: similar change where `map[string]struct{}` could be used to make the description the key which should make the test slightly easier to read.

cli/cmd/common/base.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"io"
+
+	"github.com/hashicorp/consul-k8s/cli/cmd/common/terminal"
+	"github.com/hashicorp/go-hclog"
+)
+
+// BaseCommand is embedded in all commands to provide common logic and data.
+type BaseCommand struct {
```
thisisnotashwin, line 12: I love this pattern SO much!!

thisisnotashwin (commented):
Heyo!! This is such great work!! The amount of effort that has gone into this is so clear!! Have some comments that should hopefully be straightforward!

cli/cmd/version/version.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// A pre-release marker for the version. If this is "" (empty string)
+	// then it means that it is a final release. Otherwise, this is a pre-release
+	// such as "dev" (in development), "beta", "rc1", etc.
+	VersionPrerelease = ""
+)
+
+// GetHumanVersion composes the parts of the version in a way that's suitable
+// for displaying to humans.
+func GetHumanVersion() string {
```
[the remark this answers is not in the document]
ndhanushkodi: This file is copied from: https://github.com/hashicorp/consul-k8s/blob/main/control-plane/version/version.go, and similar to https://github.com/hashicorp/consul/blob/main/version/version.go which I probably should have referenced in the file itself, and it is only ever modified by release CI. For that reason, I think it's ok if it doesn't have tests, as it's not tech debt in that a human would need to modify this file. WDYT?

cli/README.md
```diff
@@ -0,0 +1,71 @@
+# Consul Kubernetes CLI
+This repository contains a CLI tool for installing and operating [Consul](https://www.consul.io/) on Kubernetes. 
+**Warning** this tool is currently experimental. Do not use it on Consul clusters you care about.
+
+## Installation & Setup
+Currently the tool is not available on any releases page. Instead clone the repository and run `go build -o bin/consul-k8s`
+and proceed to run the binary.
+
+## Commands
+* [consul-k8s install](# consul-k8s install)
```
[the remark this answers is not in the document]
ndhanushkodi: Thanks for catching this!!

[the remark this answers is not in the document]
ndhanushkodi: Thanks for catching this!

cli/cmd/install/install_test.go
```diff
@@ -0,0 +1,163 @@
...
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			if result := validLabel(testCase.input); result != testCase.expected {
+				t.Errorf("Incorrect output, got %v and expected %v", result, testCase.expected)
+			}
+		})
+	}
+}
+
+func assertEqual(t *testing.T, a, b interface{}) {
```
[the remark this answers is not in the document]
ndhanushkodi: Nope it's unused, and yup require.Equal should work, thanks for catching this!!

cli/cmd/install/install_test.go
```diff
@@ -0,0 +1,163 @@
...
+	require.NoError(t, err)
+}
+
+// TestValidateFlags tests the validate flags function.
+func TestValidateFlags(t *testing.T) {
+	// The following cases should all error, if they fail to this test fails.
+	testCases := []struct {
+		input       []string
+		description string
+	}{
```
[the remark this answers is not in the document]
ndhanushkodi: I kept it as a struct but switched the arguments around and added some spacing that I think makes it easier to read :)

cli/cmd/install/install_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/hashicorp/consul-k8s/cli/cmd/common"
+	"github.com/hashicorp/go-hclog"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+// Helper function which sets up a Command struct for you.
```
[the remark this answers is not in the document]
ndhanushkodi: Yup, great callout thanks!

cli/cmd/install/install.go
```diff
@@ -0,0 +1,496 @@
...
+	FlagDryRun    = "dry-run"
+	DefaultDryRun = false
+
+	FlagAutoApprove    = "auto-approve"
+	DefaultAutoApprove = false
+
+	FlagNamespace    = "namespace"
+	DefaultNamespace = "consul"
+
+	HelmRepository = "https://helm.releases.hashicorp.com"
```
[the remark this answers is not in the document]
ndhanushkodi: Nope, nice callout, unexported them!

cli/cmd/install/install.go
```diff
@@ -0,0 +1,496 @@
...
+
+	if err := c.validateFlags(args); err != nil {
+		c.UI.Output(err.Error())
+		return 1
+	}
+
+	// A hack to set namespace via the HELM_NAMESPACE env var until we merge a PR that will allow us to use the latest
+	// Helm templates.
+	prevHelmNSEnv := os.Getenv("HELM_NAMESPACE")
+	os.Setenv("HELM_NAMESPACE", c.flagNamespace)
```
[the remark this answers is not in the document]
ndhanushkodi: The PR that was merged fixes the latest templates. The PR we need for this is for the CLI itself to build in the latest templates. Currently this is using templates from releases.hashicorp.com.

cli/cmd/install/presets.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Preset map which maps preset name to a map from string
+// to interface{}. Basically just YAML.
+var presets = map[string]interface{}{
+	PresetDemo:   convert(demo),
+	PresetSecure: convert(secure),
+}
+
+// Below are the various presets in YAML.
+var demo = `
```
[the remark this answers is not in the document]
ndhanushkodi: good call, added an asana as well

cli/cmd/install/install.go
```diff
@@ -0,0 +1,496 @@
...
+		c.UI.Output(err.Error(), terminal.WithErrorStyle())
+		return 1
+	}
+
+	// Ensure there's no previous bootstrap secret lying around.
+	err = c.checkForPreviousSecrets()
+	if err != nil {
+		c.UI.Output(err.Error(), terminal.WithErrorStyle())
+		return 1
+	}
```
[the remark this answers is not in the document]
ndhanushkodi: Yup I see the pattern used in endpoints controller, thanks for the suggestion!

cli/cmd/install/install.go
ndhanushkodi: You're totally right, the error should be checked as per your previous comment, and not here, updated!

cli/cmd/install/install_test.go
ndhanushkodi: made a change similar to the previous comment, kept the struct but changed the order and spacing for readability

cli/cmd/install/install_test.go
ndhanushkodi: Great suggestion, this looks a lot nicer!

cli/cmd/install/install.go
```diff
@@ -0,0 +1,496 @@
...
+
+	c.help = c.set.Help()
+}
+
+func (c *Command) Run(args []string) int {
+	c.once.Do(c.init)
+	// Note that `c.init` and `c.Init` are NOT the same thing. One initializes the command struct,
+	// the other the UI. It looks similar because BaseCommand is embedded in Command.
+	c.Init()
```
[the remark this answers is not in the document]
ndhanushkodi: Yup, I love this suggestion, moved it and works great!

cli/cmd/install/install.go
ndhanushkodi: Yup, moved the initialization step down!

ndhanushkodi:
Thanks @thisisnotashwin!! I think I've addressed everything but the table tests, so I'm just going to go ahead and re-request your review to see if there's anything blocking after that!

cli/cmd/version/version.go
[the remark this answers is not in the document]
t-eckert: Hm, totally fine for the initial release. Might be worth rethinking how we are copying and pasting code throughout. It could get unmaintainable. For now, not a problem.

thisisnotashwin (approved):
Thanks so much for the quick turnaround on the PR Nitya! This looks great. I think we can migrate to table tests later! Such an excellent PR!!

cli/cmd/install/install.go
```diff
@@ -0,0 +1,500 @@
...
+			Name:    flagNameAutoApprove,
+			Target:  &c.flagAutoApprove,
+			Default: defaultAutoApprove,
+			Usage:   "Skip confirmation prompt.",
+		})
+		f.BoolVar(&flag.BoolVar{
+			Name:    flagNameDryRun,
+			Target:  &c.flagDryRun,
+			Default: defaultDryRun,
+			Usage:   "Validate installation and return summary of installation.",
```
t-eckert, line 92: This could be more explicit in telling the user that the dry run will not modify anything, maybe "Displays a summary of what would happen if installation were to run".

cli/cmd/install/install.go
```diff
@@ -0,0 +1,500 @@
...
+				"specified multiple times. Supports Consul Helm chart values.",
+		})
+		f.StringSliceVar(&flag.StringSliceVar{
+			Name:   flagNameSetStringValues,
+			Target: &c.flagSetStringValues,
+			Usage:  "Set a string value to customize. Can be specified multiple times. Supports Consul Helm chart values.",
+		})
+
+		f = c.set.NewSet("Global Options")
+		f.StringVar(&flag.StringVar{
```
t-eckert, line 130: Given that by default people keep the kubeconfig file at `~/.kube/config`, I think it would make sense to use that as the default.

cli/cmd/install/install.go
```diff
@@ -0,0 +1,500 @@
...
+
+	// Dry Run should exit here, no need to actual locate/download the charts.
+	if c.flagDryRun {
+		c.UI.Output("Dry run complete - installation can proceed.", terminal.WithInfoStyle())
+		return 0
+	}
+
+	if !c.flagAutoApprove {
+		confirmation, err := c.UI.Input(&terminal.Input{
+			Prompt: "Proceed with installation? (y/n)",
```
t-eckert, line 265: I would capitalize either "y" or "n" depending on which choice will be run by default.

t-eckert (approved):
Fantastic work, I think this will be a huge improvement for new users coming to Consul on K8s.

cli/cmd/install/install.go
ndhanushkodi: good idea!

cli/cmd/install/install.go
ndhanushkodi: I reworded to: Run pre-install checks and display summary of installation.

cli/cmd/install/install.go
ndhanushkodi: In the default case, for any command that uses helm, we'll want to respect Helm's order of evaluation to get the kube context. I think if we were to explicitly set this to ~/.kube/config, then we wouldn't be able to respect the Helm library's "backups" for finding a kubeconfig (i.e an env var, etc.).

t-eckert: Cool, then this is just me not knowing yet :) Thanks!

ndhanushkodi:
Thanks so much for the review and fast turnaround on a big PR @t-eckert and @thisisnotashwin!

PR #718, opened 2021-09-10
Bump Consul version to 1.10.2

lkysow:
-  Bump default Consul version to 1.10.2
- fix tests that were failing due to a change in how the
    DestinationUpstream field is set in Consul OSS vs Enterprise.
- set use_streaming_backend to false for Consul clients because
    streaming is not currently supported with mesh gateway federation

How I've tested this PR:
- acceptance tests

- code

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/client-config-configmap.yaml
+++ b/charts/consul/templates/client-config-configmap.yaml
@@ -24,7 +24,8 @@ data:
        in the UI. */}}
   config.json: |-
     {
-      "check_update_interval": "0s"
+      "check_update_interval": "0s",
+      "use_streaming_backend": false
     }
   {{- end }}
 {{- end }}
--- a/charts/consul/test/unit/client-daemonset.bats
+++ b/charts/consul/test/unit/client-daemonset.bats
@@ -543,7 +543,7 @@ load _helpers
 #--------------------------------------------------------------------
 # config-configmap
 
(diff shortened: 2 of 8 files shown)


brunorochadasilva:
The world needs this PR @lkysow 😄 

 👍 in this PR even in draft.

charts/consul/templates/client-config-configmap.yaml
```diff
@@ -24,7 +24,8 @@ data:
        in the UI. */}}
   config.json: |-
     {
-      "check_update_interval": "0s"
+      "check_update_interval": "0s",
+      "use_streaming_backend": false
```
lkysow, line 28: I put this here because it only causes issues when doing federation and consul connect

charts/consul/test/unit/client-daemonset.bats
```diff
@@ -543,7 +543,7 @@ load _helpers
 #--------------------------------------------------------------------
 # config-configmap
 
-@test "client/DaemonSet: adds config-checksum annotation when extraConfig is blank" {
```
lkysow, line 546: just re-worded the test titles a bit here because it *always* adds the config-checksum annotation so really what we're testing is that the checksum changes when the configmap is different

t-eckert (approved):

ishustava (approved):

PR #721, opened 2021-09-10
Only support default partition in server cluster

thisisnotashwin:
Changes proposed in this PR:
- Fail if the user attempts to update the admin partition name on the server cluster to a value that isn't "default"
- Create an immutable configmap with the partition name which will cause `helm upgrade` to fail if the admin partition name is updated between the install and the upgrade.

Questions:
Even though the helm upgrade fails when the partition name is changed, the partition-init job runs and the agents get reassigned. I could make the partition-init job a `pre-install` job and not a `pre-upgrade` job, BUT the drawback of doing so would be anyone trying to upgrade an existing non-partition cluster to a partitioned cluster cannot. Should we ignore this use case? Ignore the upgrade use-case will ensure that the partition-init job will not run thereby not creating a new partition.

FOLLOW UP:
After conversations with the team, it was decided that in the near term, we will not support updating to a partition that is non-default. Removing the `pre-upgrade` hook from `partition-init` job. This will ensure we don't create a partition when a user errantly updates the partition-name and performs a helm upgrade.

How I've tested this PR:
Manual testing.
Bats tests

Code review
Would love to hear thoughts on the above questions.

Checklist:
- [x] Tests added

### DIFF
--- a/charts/consul/test/unit/client-daemonset.bats
+++ b/charts/consul/test/unit/client-daemonset.bats
@@ -1402,7 +1402,20 @@ rollingUpdate:
       -s templates/client-daemonset.yaml  \
       --set 'global.adminPartitions.enabled=true' \
       --set 'global.adminPartitions.name=test' \
+      --set 'server.enabled=false' \
       . | tee /dev/stderr |
       yq -c -r '.spec.template.spec.containers[0].command | join(" ") | contains("partition = \"test\"")' | tee /dev/stderr)
   [ "${actual}" = "true" ]
+}
+
+@test "client/DaemonSet: partition name has to be default in server cluster" {
+  cd `chart_dir`
+  run helm template \
+      -s templates/client-daemonset.yaml  \
+      --set 'global.adminPartitions.enabled=true' \
+      --set 'global.adminPartitions.name=test' \
+      .
+
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "global.adminPartitions.name has to be \"default\" in the server cluster" ]]
 }
--- a/charts/consul/test/unit/partition-name-configmap.bats
+++ b/charts/consul/test/unit/partition-name-configmap.bats
@@ -1,0 +1,47 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "partitionName/ConfigMap: disabled by default" {
+  cd `chart_dir`
+  assert_empty helm template \
+      -s templates/partition-init-role.yaml  \
+      .
+}
+
+@test "partitionName/ConfigMap: enabled with global.adminPartitions.enabled=true and servers = false" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/partition-init-role.yaml  \
+      --set 'global.adminPartitions.enabled=true' \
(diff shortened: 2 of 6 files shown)


charts/consul/test/unit/partition-name-configmap.bats
```diff
@@ -0,0 +1,48 @@
...
+      --set 'global.adminPartitions.enabled=true' \
+      --set 'global.enabled=true' \
+      .
+}
+
+@test "partitionName/ConfigMap: disabled with global.adminPartitions.enabled=false" {
+  cd `chart_dir`
+  assert_empty helm template \
+      -s templates/partition-init-role.yaml  \
+      --set 'global.adminPartitions.enabled=true' \
```
kschoche, line 45: I think the description or the test is wrong here on whether or not `adminPartitions.enabled` is true or false!

charts/consul/test/unit/partition-name-configmap.bats
```diff
@@ -0,0 +1,48 @@
...
+@test "partitionName/ConfigMap: disabled with global.adminPartitions.enabled=true and servers = true" {
+  cd `chart_dir`
+  assert_empty helm template \
+      -s templates/partition-init-role.yaml  \
+      --set 'global.adminPartitions.enabled=true' \
+      --set 'server.enabled=true' \
+      .
+}
+
+@test "partitionName/ConfigMap: disabled with global.adminPartitions.enabled=true and global.enabled = true" {
```
kschoche, line 32: with `global.enabled = true` , servers should be enabled, and I *think* this test should instead enable the configMap?  Maybe I'm misreading?

charts/consul/test/unit/client-daemonset.bats
```diff
@@ -1351,7 +1351,20 @@ rollingUpdate:
...
+
+@test "client/DaemonSet: partition name has to be default in server cluster" {
+  cd `chart_dir`
+  run helm template \
+      -s templates/client-daemonset.yaml  \
+      --set 'global.adminPartitions.enabled=true' \
+      --set 'global.adminPartitions.name=test' \
+      .
+
+  [ "$status" -eq 1 ]
```
kschoche, line 1368: I like this test!

kschoche (approved):
Great work.
I left a couple comments before you merge but otherwise it's good to go!
On a meta note : I wish there was a better way to make the `$serverEnabled :=` easier to read, thx Helm!

charts/consul/test/unit/partition-name-configmap.bats
thisisnotashwin: We dont enable the configmap in the server cluster, because we would fail if they changed the name to being something non-default.

charts/consul/test/unit/partition-name-configmap.bats
thisisnotashwin: Good catch!

lkysow (approved):
The docs should be updated for the values to state that:
1. Only `default` is allowed for server cluster.
2. Can't change the partition name
3. Only supported on initial installation, can't enable partitions after initial installation.

PR #724, opened 2021-09-14
Update ConfigEntries to be Partition aware

thisisnotashwin:
Changes proposed in this PR:
- Update ServiceDefaults and ServiceIntentions to be partition aware.

How I've tested this PR:
- Unit tests
- Deploying to cluster and they get reconciled successfully.

- Code review

Notes: There is a validation on Service Intentions that fails if namespaces are not enabled but the namespace field has been populated. A similar check has not been made currently as this would require changing the method signature of the ConfigEntry interface which would leads to almost every file in the `/api/v1alpha1` directory getting impacted. Post the alpha release, I will do a refactor which is extract a bunch of fields required for validation and put them in a struct. This should ensure we don't have to change method signatures every time a new concept is introduced. We can update the struct and the config-entries that are impacted.

Checklist:
- [x] Tests added


ndhanushkodi:
> There is a validation on Service Intentions that fails if namespaces are not enabled but the namespace field has been populated.

I think this could be fine to keep around until alpha as you mentioned but I just wanted to clarify the following in case I'm misunderstanding:

Is this the failing ent unit test in CI? Would we keep the failing test until post-alpha? Would you be able to run acceptance tests on this PR by skipping the ent unit tests if those will be failing for alpha? Do we need to modify CI to skip the that unit test until post alpha and make a reminder to re-enable it?

thisisnotashwin:
> > There is a validation on Service Intentions that fails if namespaces are not enabled but the namespace field has been populated.
> 
> I think this could be fine to keep around until alpha as you mentioned but I just wanted to clarify the following in case I'm misunderstanding:
> 
> Is this the failing ent unit test in CI? Would we keep the failing test until post-alpha? Would you be able to run acceptance tests on this PR by skipping the ent unit tests if those will be failing for alpha? Do we need to modify CI to skip the that unit test until post alpha and make a reminder to re-enable it?

The failing enterprise test should stop failing with the latest Consul enterprise alpha image. The test that is failing is the partition-init job that currently depends on the existence of Admin Partitions. I could create a PR that skips enterprise unit tests and runs the acceptance tests to ensure that the acceptance tests are 🟢

ndhanushkodi (approved):
Looks great Ashwin!!

ishustava (approved):
Looks great! I left a couple of non-blocking, but otherwise looks good. I haven't tested it though. Should we add an acceptance test to test out partitions similar to how we test namespaces for config entries?

thisisnotashwin:
> Looks great! I left a couple of non-blocking, but otherwise looks good. I haven't tested it though. Should we add an acceptance test to test out partitions similar to how we test namespaces for config entries?

We definitely should update our acceptance tests. I was contemplating doing that as a part of the larger story of creating acceptance tests for Admin Partitions.

PR #725, opened 2021-09-15
cli: uninstall command

ndhanushkodi:
Based off of a demo branch @sadjamz worked on here: https://github.com/hashicorp/consul-k8s-cli/tree/uninstall

- `consul-k8s uninstall` uninstalls the Consul Helm installation with options to wipe all data
- If you don't provide the release name/namespace, it will find the release whose metadata.name matches "consul", and then prompt to uninstall that release and any pvcs/secrets based on that.
- You can provide the release name/namespace in a case where you have already `helm uninstall`'ed the consul helm installation, but still have PVCs, Secrets, etc to clean up.
- Note that this PR is difficult to unit test-- acceptance tests are coming as the next task after install/uninstall are complete. The Helm steps are difficult to mock out for reasons mentioned in the install PR:

  - https://github.com/hashicorp/consul-k8s/pull/713#discussion_r705794524

  - https://github.com/hashicorp/consul-k8s/blob/e77da79f99bbd1038f7b3dc203c6492127341e6b/cli/cmd/install/install.go#L338

How I've tested this PR:
Manual testing of the cases mentioned below, a few unit tests.

1) Code review

2) Test it out:
`cd cli`
`go build -o ./bin/consul-k8s`
`./bin/consul-k8s uninstall`
If you can test out the command with the following flags and let me know if you have behaviour suggestions. 
  a) no flags
  b) -name and -namespace set
  c) -auto-approve
  d) -wipe-all-data

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -109,12 +109,10 @@ Command Options:
       Namespace for the Consul installation.
 
   -wipe-data
-      This behaviour of this flag depends on the value of -auto-approve.
-      When -wipe-data=true, the CLI will delete PVCs and Secrets if
-      -auto-approve=true. When -wipe-data=false, the CLI will prompt to
-      wipe PVCs and Secrets if -auto-approve=false, and skip wiping them
-      if -auto-approve=true. Only set this to true when persisted data from
-      previous installations is no longer necessary. The default is false.
+      When used in combination with -auto-approve, all persisted data (PVCs
+      and Secrets) from previous installations will be deleted. Only set this
+      to true when data from previous installations is no longer necessary.
+      The default is false.
 
 Global Options:
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -4,7 +4,7 @@ This repository contains a CLI tool for installing and operating [Consul](https:
 
 ## Installation & Setup
 Currently the tool is not available on any releases page. Instead clone the repository and run `go build -o bin/consul-k8s`
-and proceed to run the binary.
+from this directory and proceed to run the binary.
 
 ## Commands
 * [consul-k8s install](#consul-k8s-install)
@@ -94,17 +94,16 @@ consul-k8s uninstall
 
 ```
 Usage: kubectl consul uninstall [options]
+Uninstall Consul with options to delete data and resources associated with Consul installation.
 
-Uninstall Consul and clean up all data.
-Any data store in Consul will not be recoverable.
 Command Options:
 
   -auto-approve
       Skip approval prompt for uninstalling Consul. The default is false.
 
   -name=<string>
-      Name of the installation. This will be prefixed to resources installed
-      on the cluster.
+      Name of the installation. This can be used to uninstall and/or delete
+      the resources of a specific Helm release.
 
   -namespace=<string>
       Namespace for the Consul installation.
@@ -127,6 +126,4 @@ Global Options:
 
   -kubeconfig=<string>
       Path to kubeconfig file. This is aliased as "-c".
-
-
 ```
--- a/cli/README.md
+++ b/cli/README.md
@@ -8,6 +8,7 @@ and proceed to run the binary.
 
 ## Commands
 * [consul-k8s install](#consul-k8s-install)
+* [consul-k8s uninstall](#consul-k8s-uninstall)
 
 ### consul-k8s install
 This command installs Consul on a Kubernetes cluster. It allows `demo` and `secure` installations via preset configurations
@@ -79,5 +80,53 @@ Global Options:
 
   -kubeconfig=<string>
       Path to kubeconfig file. This is aliased as "-c".
+
+```
+
+### consul-k8s uninstall
+This command uninstalls Consul on Kubernetes, while prompting whether to uninstall the release and whether to delete all
+related resources such as PVCs, Secrets, and ServiceAccounts.
+
+Get started with:
+```bash
+consul-k8s uninstall
+```
+
+```
+Usage: kubectl consul uninstall [options]
+
+Uninstall Consul and clean up all data.
+Any data store in Consul will not be recoverable.
+Command Options:
+
+  -auto-approve
+      Skip approval prompt for uninstalling Consul. The default is false.
+
+  -name=<string>
+      Name of the installation. This will be prefixed to resources installed
+      on the cluster.
+
+  -namespace=<string>
+      Namespace for the Consul installation.
+
+  -skip-wipe-data
+      Skip deleting all PVCs, Secrets, and Service Accounts associated with
+      Consul Helm installation without prompting for approval to delete. The
+      default is false.
+
+  -wipe-data
+      Delete all PVCs, Secrets, and Service Accounts associated with Consul
+      Helm installation without prompting for approval to delete. Only use
+      this when persisted data from previous installations is no longer
+      necessary. The default is false.
+
+Global Options:
+
+  -context=<string>
+      Kubernetes context to use.
+
+  -kubeconfig=<string>
+      Path to kubeconfig file. This is aliased as "-c".
+
 
 ```
--- a/cli/README.md
+++ b/cli/README.md
@@ -108,16 +108,13 @@ Command Options:
   -namespace=<string>
       Namespace for the Consul installation.
 
-  -skip-wipe-data
-      Skip deleting all PVCs, Secrets, and Service Accounts associated with
-      Consul Helm installation without prompting for approval to delete. The
-      default is false.
-
   -wipe-data
-      Delete all PVCs, Secrets, and Service Accounts associated with Consul
-      Helm installation without prompting for approval to delete. Only use
-      this when persisted data from previous installations is no longer
-      necessary. The default is false.
+      This behaviour of this flag depends on the value of -auto-approve.
+      When -wipe-data=true, the CLI will delete PVCs and Secrets if
+      -auto-approve=true. When -wipe-data=false, the CLI will prompt to
+      wipe PVCs and Secrets if -auto-approve=false, and skip wiping them
+      if -auto-approve=true. Only set this to true when persisted data from
+      previous installations is no longer necessary. The default is false.
 
 Global Options:
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -93,7 +93,7 @@ consul-k8s uninstall
 ```
 
 ```
-Usage: kubectl consul uninstall [options]
+Usage: consul-k8s uninstall [flags]
 Uninstall Consul with options to delete data and resources associated with Consul installation.
 
 Command Options:
@@ -107,6 +107,9 @@ Command Options:
 
   -namespace=<string>
       Namespace for the Consul installation.
+
+  -timeout=<string>
+      Timeout to wait for uninstall. The default is 10m.
 
   -wipe-data
       When used in combination with -auto-approve, all persisted data (PVCs
--- a/cli/cmd/common/utils.go
+++ b/cli/cmd/common/utils.go
@@ -1,0 +1,11 @@
+package common
+
+import "strings"
+
+func PrefixLines(prefix, lines string) string {
+	var prefixedLines string
+	for _, l := range strings.Split(lines, "\n") {
+		prefixedLines += prefix + l + "\n"
+	}
+	return strings.TrimSuffix(prefixedLines, "\n")
+}
--- a/cli/cmd/common/utils.go
+++ b/cli/cmd/common/utils.go
@@ -2,10 +2,11 @@ package common
 
 import "strings"
 
-func PrefixLines(prefix, lines string) string {
-	var prefixedLines string
-	for _, l := range strings.Split(lines, "\n") {
-		prefixedLines += prefix + l + "\n"
+// Abort returns true if the raw input string is not equal to "y" or "yes".
+func Abort(raw string) bool {
+	confirmation := strings.TrimSuffix(raw, "\n")
+	if !(strings.ToLower(confirmation) == "y" || strings.ToLower(confirmation) == "yes") {
+		return true
 	}
-	return strings.TrimSuffix(prefixedLines, "\n")
+	return false
 }
--- a/cli/cmd/common/utils.go
+++ b/cli/cmd/common/utils.go
@@ -10,6 +10,11 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
+const (
+	DefaultReleaseName      = "consul"
+	DefaultReleaseNamespace = "consul"
+)
+
 // Abort returns true if the raw input string is not equal to "y" or "yes".
 func Abort(raw string) bool {
 	confirmation := strings.TrimSuffix(raw, "\n")
--- a/cli/cmd/common/utils.go
+++ b/cli/cmd/common/utils.go
@@ -1,6 +1,14 @@
 package common
 
-import "strings"
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"helm.sh/helm/v3/pkg/action"
+	helmCLI "helm.sh/helm/v3/pkg/cli"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
 
 // Abort returns true if the raw input string is not equal to "y" or "yes".
 func Abort(raw string) bool {
@@ -9,4 +17,18 @@ func Abort(raw string) bool {
 		return true
 	}
 	return false
+}
+
+// InitActionConfig initializes a Helm Go SDK action configuration. This function currently uses a hack to override the
+// namespace field that gets set in the K8s client set up by the SDK.
+func InitActionConfig(actionConfig *action.Configuration, namespace string, settings *helmCLI.EnvSettings, logger action.DebugLog) (*action.Configuration, error) {
+	getter := settings.RESTClientGetter()
+	configFlags := getter.(*genericclioptions.ConfigFlags)
+	configFlags.Namespace = &namespace
+	err := actionConfig.Init(settings.RESTClientGetter(), namespace,
+		os.Getenv("HELM_DRIVER"), logger)
+	if err != nil {
+		return nil, fmt.Errorf("error setting up helm action configuration to find existing installations: %s", err)
+	}
+	return actionConfig, nil
 }
--- a/cli/cmd/install/install.go
+++ b/cli/cmd/install/install.go
@@ -86,82 +86,80 @@ func (c *Command) init() {
 	}
 
 	c.set = flag.NewSets()
-	{
-		f := c.set.NewSet("Command Options")
-		f.BoolVar(&flag.BoolVar{
-			Name:    flagNameAutoApprove,
-			Target:  &c.flagAutoApprove,
-			Default: defaultAutoApprove,
-			Usage:   "Skip confirmation prompt.",
-		})
-		f.BoolVar(&flag.BoolVar{
-			Name:    flagNameDryRun,
-			Target:  &c.flagDryRun,
-			Default: defaultDryRun,
-			Usage:   "Run pre-install checks and display summary of installation.",
-		})
-		f.StringSliceVar(&flag.StringSliceVar{
-			Name:    flagNameConfigFile,
-			Aliases: []string{"f"},
-			Target:  &c.flagValueFiles,
-			Usage:   "Path to a file to customize the installation, such as Consul Helm chart values file. Can be specified multiple times.",
-		})
-		f.StringVar(&flag.StringVar{
-			Name:    flagNameNamespace,
-			Target:  &c.flagNamespace,
-			Default: defaultNamespace,
-			Usage:   "Namespace for the Consul installation.",
-		})
-		f.StringVar(&flag.StringVar{
-			Name:    flagNamePreset,
-			Target:  &c.flagPreset,
-			Default: defaultPreset,
-			Usage:   fmt.Sprintf("Use an installation preset, one of %s. Defaults to none", strings.Join(presetList, ", ")),
-		})
-		f.StringSliceVar(&flag.StringSliceVar{
-			Name:   flagNameSetValues,
-			Target: &c.flagSetValues,
-			Usage:  "Set a value to customize. Can be specified multiple times. Supports Consul Helm chart values.",
-		})
-		f.StringSliceVar(&flag.StringSliceVar{
-			Name:   flagNameFileValues,
-			Target: &c.flagFileValues,
-			Usage: "Set a value to customize via a file. The contents of the file will be set as the value. Can be " +
-				"specified multiple times. Supports Consul Helm chart values.",
-		})
-		f.StringSliceVar(&flag.StringSliceVar{
-			Name:   flagNameSetStringValues,
-			Target: &c.flagSetStringValues,
-			Usage:  "Set a string value to customize. Can be specified multiple times. Supports Consul Helm chart values.",
-		})
-		f.StringVar(&flag.StringVar{
-			Name:    flagNameTimeout,
-			Target:  &c.flagTimeout,
-			Default: defaultTimeout,
-			Usage:   "Timeout to wait for installation to be ready.",
-		})
-		f.BoolVar(&flag.BoolVar{
-			Name:    flagNameWait,
-			Target:  &c.flagWait,
-			Default: defaultWait,
-			Usage:   "Determines whether to wait for resources in installation to be ready before exiting command.",
-		})
+	f := c.set.NewSet("Command Options")
+	f.BoolVar(&flag.BoolVar{
+		Name:    flagNameAutoApprove,
+		Target:  &c.flagAutoApprove,
+		Default: defaultAutoApprove,
+		Usage:   "Skip confirmation prompt.",
+	})
+	f.BoolVar(&flag.BoolVar{
+		Name:    flagNameDryRun,
+		Target:  &c.flagDryRun,
+		Default: defaultDryRun,
+		Usage:   "Run pre-install checks and display summary of installation.",
+	})
+	f.StringSliceVar(&flag.StringSliceVar{
+		Name:    flagNameConfigFile,
+		Aliases: []string{"f"},
+		Target:  &c.flagValueFiles,
+		Usage:   "Path to a file to customize the installation, such as Consul Helm chart values file. Can be specified multiple times.",
+	})
+	f.StringVar(&flag.StringVar{
+		Name:    flagNameNamespace,
+		Target:  &c.flagNamespace,
+		Default: defaultNamespace,
+		Usage:   "Namespace for the Consul installation.",
+	})
+	f.StringVar(&flag.StringVar{
+		Name:    flagNamePreset,
+		Target:  &c.flagPreset,
+		Default: defaultPreset,
+		Usage:   fmt.Sprintf("Use an installation preset, one of %s. Defaults to none", strings.Join(presetList, ", ")),
+	})
+	f.StringSliceVar(&flag.StringSliceVar{
+		Name:   flagNameSetValues,
+		Target: &c.flagSetValues,
+		Usage:  "Set a value to customize. Can be specified multiple times. Supports Consul Helm chart values.",
+	})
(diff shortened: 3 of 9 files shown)


cli/README.md
```diff
@@ -69,3 +70,51 @@ Global Flags:
...
+
+Get started with:
+```bash
+consul-k8s uninstall
+```
+
+```
+Usage: kubectl consul uninstall [options]
+
+Uninstall Consul and clean up all data.
```
kschoche, line 86:
```suggestion
Uninstall Consul and delete all data.
```

cli/README.md
```diff
@@ -69,3 +70,51 @@ Global Flags:
...
+Any data store in Consul will not be recoverable.
```
kschoche, line 87:
```suggestion
Any data stored in Consul will not be recoverable.
```

cli/README.md
```diff
@@ -69,3 +70,51 @@ Global Flags:
...
+Command Options:
+
+  -auto-approve
+      Skip approval prompt for uninstalling Consul. The default is false.
+
+  -name=<string>
+      Name of the installation. This will be prefixed to resources installed
+      on the cluster.
```
kschoche, line 95: I'm pretty sure this is a copy/pasta issue, but I'm not sure why we need the second sentence in the context of uninstall? :)

cli/README.md
```diff
@@ -69,3 +70,51 @@ Global Flags:
...
+  -skip-wipe-data
+      Skip deleting all PVCs, Secrets, and Service Accounts associated with
+      Consul Helm installation without prompting for approval to delete. The
+      default is false.
+
+  -wipe-data
+      Delete all PVCs, Secrets, and Service Accounts associated with Consul
+      Helm installation without prompting for approval to delete. Only use
+      this when persisted data from previous installations is no longer
+      necessary. The default is false.
```
kschoche, line 109: If skip-wipe-data is defaulted `false`  (meaning wipe the data)  and wipe-data is defaulted `false` (meaning dont wipe the data) what is the actual default behaviour? I'm assuming that we meant that skip-wipe-data is defaulted to `true`?

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,439 @@
...
+	flagKubeConfig  string
+	flagKubeContext string
+
+	once sync.Once
+	help string
+}
+
+func (c *Command) init() {
+	c.set = flag.NewSets()
+	{
```
kschoche, line 59: does this whole section need to be scoped in a `{ }` ?

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,439 @@
...
+
+	// c.Init() calls the embedded BaseCommand's initialization function.
+	c.Init()
+}
+
+func (c *Command) Run(args []string) int {
+	c.once.Do(c.init)
+
+	defer func() {
+		if err := c.Close(); err != nil {
```
kschoche, line 122: Just out of curiousity what does c.Close() here do? would c.UI.Output() still be usable if c.Close() failed?

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,439 @@
...
+		// Prompt for approval to uninstall Helm release.
+		if !c.flagAutoApprove {
+			confirmation, err := c.UI.Input(&terminal.Input{
+				Prompt: "Proceed with uninstall? (y/n)",
+				Style:  terminal.InfoStyle,
+				Secret: false,
+			})
+			if err != nil {
+				c.UI.Output(err.Error(), terminal.WithErrorStyle())
+				return 1
```
kschoche, line 203: Curious about the distinction of using `os.Exit(1)` and `return 1`, I know that at the beginning we're using Exit() as a means to highlight that the command failed due to flag parsing, but Is the exit behaviour any different from Run() when you return 1?

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,439 @@
...
+			return 1
+		} else if res != nil && res.Info != "" {
+			c.UI.Output("uninstall result: %s", res.Info, terminal.WithErrorStyle())
+			return 1
+		} else {
+			c.UI.Output("Successfully uninstalled Consul Helm release", terminal.WithSuccessStyle())
+		}
+	}
+
+	if c.flagSkipWipeData {
```
kschoche, line 230:
```suggestion
	if c.flagSkipWipeData || !c.flagWipeData {
```
Arent these flags implying the same thing (in reverse)?

kschoche: I'm suggesting that the user may pass in `-wipe-data=false`

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,439 @@
...
+		c.UI.Output(err.Error(), terminal.WithErrorStyle())
+		return 1
+	}
+
+	return 0
+}
+
+func (c *Command) Help() string {
+	c.once.Do(c.init)
+	s := "Usage: kubectl consul uninstall [options]" + "\n\n" + "Uninstall Consul and clean up all data." + "\n" +
```
kschoche, line 291: I'm now understanding why you worded it as "clean up all data" earlier because this can be used to "clean up" an old install. Wondering if switching the wording to "delete all data" like I mentioned earlier still makes sense? It is technically still doing that.

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,439 @@
...
+
+func (c *Command) Synopsis() string {
+	return "Uninstall Consul deployment."
+}
+
+const help = `
+Usage: kubectl consul uninstall [options]
+
+  Uninstall Consul and clean up all data.
+  This is a destructive action. Any data store in Consul will
```
kschoche, line 304:
```suggestion
  This is a destructive action. Any data stored in Consul will
```

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,439 @@
...
+		if strings.HasPrefix(secret.Name, foundReleaseName) {
+			err := c.kubernetes.CoreV1().Secrets(foundReleaseNamespace).Delete(c.Ctx, secret.Name, metav1.DeleteOptions{})
+			if err != nil {
+				return fmt.Errorf("deleteSecrets: error deleting Secret %q: %s", secret.Name, err)
+			}
+			secretNames = append(secretNames, secret.Name)
+		}
+	}
+	if len(secretNames) > 0 {
+		c.UI.Output("Consul secrets deleted.", terminal.WithSuccessStyle())
```
kschoche, line 410: This is a super nit comment, but if there are 3 secrets and 2 got deleted and the third exits the for loop because you cant delete it for some reason then neither of these Outputs get processed properly. I suppose it is okay since it "should work" in happy path.

kschoche: Maybe the correct thing to do is output the Error via `c.UI.Output()` instead of return the error, and then continue to process the other secrets?

I guess this logic applies to the other similar functions, if you choose.

kschoche (approved):
This is really great, I left some mostly minor comments but assuming you get time to look at them I think it's good to go!
I did not get a chance to test run the CLI unfortunately.. will get to that soon :)

cli/README.md
ndhanushkodi: Ended up removing this sentence because of the options we discussed in eng sync!

cli/cmd/uninstall/uninstall.go
ndhanushkodi: Nope! Thanks for catching that!

cli/cmd/uninstall/uninstall.go
ndhanushkodi: Yeah, I agree that delete all data is accurate and clear. The "clean up" refers more to us as devs using it for testing which is not our main audience.

cli/cmd/uninstall/uninstall.go
ndhanushkodi: Ended up removing this since we won't always remove all the data

cli/cmd/uninstall/uninstall.go
ndhanushkodi: Oh whoops, I had cleaned that up for install but not uninstall here, sorry about that! The only place we should have to os.Exit(1) is in the defer() function call. Everything else could return 1 for consistency. That is what we do in install.

cli/README.md
ndhanushkodi: Yeah good point! Updated that sentence to read: "This can be used to uninstall and/or delete the resources of a specific Helm release."

cli/cmd/install/install.go
```diff
@@ -86,82 +86,80 @@ func (c *Command) init() {
 	}
 
 	c.set = flag.NewSets()
-	{
```
ndhanushkodi, line 89: Note: this diff is just removing brackets

cli/README.md
```diff
@@ -81,3 +82,45 @@ Global Options:
...
+  -namespace=<string>
+      Namespace for the Consul installation.
+
+  -wipe-data
+      This behaviour of this flag depends on the value of -auto-approve.
+      When -wipe-data=true, the CLI will delete PVCs and Secrets if
+      -auto-approve=true. When -wipe-data=false, the CLI will prompt to
+      wipe PVCs and Secrets if -auto-approve=false, and skip wiping them
+      if -auto-approve=true. Only set this to true when persisted data from
+      previous installations is no longer necessary. The default is false.
```
ndhanushkodi, line 117: Note: the latest commit on this branch is the `-wipe-data` behaviour discussed in eng-sync. I realized when doing this that it's kind of confusing behaviour to explain in a flag description. 

The confusing part is that -wipe-data=false means 2 different things, based on what -auto-approve is.

What if we had `auto-approve` and `-auto-approve-wipe-data`, where the former would just uninstall with no prompts and the latter would uninstall and wipe data with no prompts. If both are false, the CLI will prompt both.

ndhanushkodi: Chatted with David and Luke at cli sync and Luke had a nice way to phrase this flag:

"When -wipe-data is set to true in combination with -auto-approve=true, uninstall will delete all persisted data (PVCs and Secrets) from previous installations."

Then we'd have flag validation for if you set it alone, it will error out. 

So I'll go ahead and update this!

ishustava (commented):
This is impressive work!!! I'm so excited about this ❤️  This is a lot of work, and I really appreciate how easy this was to read.

I've only reviewed the command (and not tests) and had some comments that I wanted to clarify before proceeding with the rest of the review. I've also left some minor edits.

I've also ran some manual tests and discussed the issues I ran into with Nitya already but wanted to document them here for posterity:
1. I think we need to change the helm list functionality to list any pending/uninstalled releases so that if your installation has failed, the uninstall will still find the release.
2. We need make sure that the kubernetes client is initialized with the correct namespace. With the 0.33.0 release currently, uninstall fails because tls-init-cleanup job doesn't have `Namespace` set in its template. We've fixed it now in the latest helm release, but we should still make sure that the client is initialized correctly.
3. It would be nice to default release name and namespace to `consul` for a case when I forgot to wipe-data in a previous uninstall. That way rerunning uninstall would still just work.

There is also another issue that we haven't discussed that I think would be nice to fix but doesn't have to be part of this PR. In my testing I kept running into errors saying that tls-init-cleanup job or service account etc already exist. It'd be nice if the uninstall command cleaned those up before calling helm delete.

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+			c.UI.Output(err.Error(), terminal.WithErrorStyle())
+			return 1
+		}
+		uninstaller := action.NewUninstall(actionConfig)
+		res, err := uninstaller.Run(foundReleaseName)
+		if err != nil {
+			c.UI.Output("unable to uninstall: %s", err, terminal.WithErrorStyle())
+			return 1
+		} else if res != nil && res.Info != "" {
+			c.UI.Output("uninstall result: %s", res.Info, terminal.WithErrorStyle())
```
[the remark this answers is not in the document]
ndhanushkodi: Thanks for catching the result being printed with error, that's not an error case you're right. I updated that. 

I have a bunch of helper functions throughout that return errors, and then if that function returns an error we print it out to the UI with error style. Since those errors will be lowercased, I was thinking it'd make more sense if any errors we output were lowercased. WDYT?

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+		// Actually call out to `helm delete`.
+		actionConfig, err = c.initActionConfig(actionConfig, foundReleaseNamespace, settings, uiLogger)
+		if err != nil {
+			c.UI.Output(err.Error(), terminal.WithErrorStyle())
+			return 1
+		}
+		uninstaller := action.NewUninstall(actionConfig)
+		res, err := uninstaller.Run(foundReleaseName)
+		if err != nil {
+			c.UI.Output("unable to uninstall: %s", err, terminal.WithErrorStyle())
```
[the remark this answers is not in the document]
ndhanushkodi: I have a bunch of helper functions throughout that return errors, and then if that function returns an error we print it out to the UI with error style. Since those errors will be lowercased, I was thinking it'd make more sense if any errors we output were lowercased. WDYT?

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	// At this point, even if no Helm release was found and uninstalled, there could
+	// still be PVCs, Secrets, and Service Accounts left behind from a previous installation.
+	// If there isn't a foundReleaseName and foundReleaseNamespace, we'll use the values of the
+	// flags c.flagReleaseName and c.flagNamespace. If those are empty we'll prompt the user that
+	// those should be provided to fully clean up the installation.
+	if !found {
+		if c.flagReleaseName == "" || c.flagNamespace == "" {
+			c.UI.Output("No existing Consul Helm installation was found. To search for existing PVCs, secrets, and service accounts left behind by a Consul installation, please provide -release-name and -namespace.", terminal.WithInfoStyle())
```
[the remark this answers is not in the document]
ndhanushkodi: This should happen after the logic on line 220 so we can print the right output in that case without printing the line here.
```
if c.flagAutoApprove && !c.flagWipeData {
        c.UI.Output("Skipping deleting PVCs, secrets, and service accounts.", terminal.WithSuccessStyle())
        return 0
}

File: cli/cmd/uninstall/uninstall.go [hunk 17]
```diff
@@ -0,0 +1,419 @@
...
+
+func (c *Command) initActionConfig(actionConfig *action.Configuration, namespace string, settings *helmCLI.EnvSettings, logger action.DebugLog) (*action.Configuration, error) {
+	var err error
+	if namespace == defaultAllNamespaces {
+		err = actionConfig.Init(settings.RESTClientGetter(), "",
+			os.Getenv("HELM_DRIVER"), logger)
+	} else {
+		err = actionConfig.Init(settings.RESTClientGetter(), namespace,
+			os.Getenv("HELM_DRIVER"), logger)
+	}
```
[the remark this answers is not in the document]
ndhanushkodi: Ah, nice simplification catch!

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+	for _, rel := range res {
+		if rel.Chart.Metadata.Name == "consul" {
+			if c.flagNamespace != defaultAllNamespaces {
+				if c.flagNamespace == rel.Name {
+					found = true
+					foundReleaseName = rel.Name
+					foundReleaseNamespace = rel.Namespace
+					break
+				}
+			}
```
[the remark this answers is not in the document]
ndhanushkodi: You're right! I modified the logic and added a comment to explain. if the -namespace flag is specified, we only want to say we found the release if the -namespace flag matches the release's namespace.

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+	err = backoff.Retry(func() error {
+		pvcs, err := c.kubernetes.CoreV1().PersistentVolumeClaims(foundReleaseNamespace).List(c.Ctx, pvcSelector)
+		if err != nil {
+			return fmt.Errorf("deletePVCs: %s", err)
+		}
+		if len(pvcs.Items) > 0 {
+			err = fmt.Errorf("deletePVCs: pvcs still exist")
+			return err
+		}
+		return err
```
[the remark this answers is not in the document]
ndhanushkodi: Yup I love it!

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+		return fmt.Errorf("deletePVCs: %s", err)
+	}
+	for _, pvc := range pvcs.Items {
+		err := c.kubernetes.CoreV1().PersistentVolumeClaims(foundReleaseNamespace).Delete(c.Ctx, pvc.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return fmt.Errorf("deletePVCs: error deleting PVC %q: %s", pvc.Name, err)
+		}
+		pvcNames = append(pvcNames, pvc.Name)
+	}
+	err = backoff.Retry(func() error {
```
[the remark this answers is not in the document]
ndhanushkodi: The idea is that when the command finishes, if install is run again immediately afterwards, that we've waited for the PVC to go through Bound --> Terminating --> deleted, so install wouldn't fail. 

If the PVC is deleted and someone had manually provisioned the volume with a "retain" policy they might be saving that data but they might be ok with provisioning new pvcs for a new installation. So I wasn't sure whether to base it off of PV in that case. 

I think we'll want to wait for secrets and service accounts as well, for a case where install is run programmatically, immediately after uninstall.

cli/cmd/uninstall/uninstall.go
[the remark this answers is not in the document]
ishustava: Oh I see, that makes sense! I think I have a slight bias toward consistency but that's fine too. I'm thinking about how to make it more obvious in the code in case someone else comes in and thinks the same thing. I think a comment before every error output is not feasible, but maybe we can create a helper function for error outputs? Doesn't have to be part of this PR though!

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+	var pvcNames []string
+	pvcSelector := metav1.ListOptions{LabelSelector: fmt.Sprintf("release=%s", foundReleaseName)}
+	pvcs, err := c.kubernetes.CoreV1().PersistentVolumeClaims(foundReleaseNamespace).List(c.Ctx, pvcSelector)
+	if err != nil {
+		return fmt.Errorf("deletePVCs: %s", err)
+	}
+	for _, pvc := range pvcs.Items {
+		err := c.kubernetes.CoreV1().PersistentVolumeClaims(foundReleaseNamespace).Delete(c.Ctx, pvc.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return fmt.Errorf("deletePVCs: error deleting PVC %q: %s", pvc.Name, err)
```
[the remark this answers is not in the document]
ndhanushkodi: I think it's ok for the user to re-run the command at this point. They'd have to manually resolve the error, and then they could just re-run.

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+
+// deleteSecrets deletes any secrets that have foundReleaseName in their name.
+func (c *Command) deleteSecrets(foundReleaseName, foundReleaseNamespace string) error {
+	var secretNames []string
+	secrets, err := c.kubernetes.CoreV1().Secrets(foundReleaseNamespace).List(c.Ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("deleteSecrets: %s", err)
+	}
+	for _, secret := range secrets.Items {
+		if strings.HasPrefix(secret.Name, foundReleaseName) {
```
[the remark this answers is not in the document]
ndhanushkodi: Yeah, thanks for the reminder. Tracking it here: https://app.asana.com/0/1200500259727746/1201031492148287/f

cli/cmd/uninstall/uninstall.go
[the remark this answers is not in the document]
ishustava: Oh got it! that makes a lot of sense. Thanks for clarifying.

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,419 @@
...
+	if len(secretNames) > 0 {
+		c.UI.Output("Consul secrets deleted.", terminal.WithSuccessStyle())
+	} else {
+		c.UI.Output("No Consul secrets found.", terminal.WithSuccessStyle())
+	}
+	return nil
+}
+
+// deleteServiceAccounts deletes service accounts that have foundReleaseName in their name.
+func (c *Command) deleteServiceAccounts(foundReleaseName, foundReleaseNamespace string) error {
```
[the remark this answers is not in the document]
ndhanushkodi: removing this function in favor of: https://app.asana.com/0/1200500259727746/1201031492148291/f

cli/cmd/uninstall/uninstall.go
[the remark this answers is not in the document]
ndhanushkodi: yeah great point, added an asana for this. I plan to address this while doing upgrade command https://app.asana.com/0/1200500259727746/1201031492148295/f

cli/cmd/common/utils.go
```diff
@@ -0,0 +1,11 @@
+package common
+
+import "strings"
+
+func PrefixLines(prefix, lines string) string {
```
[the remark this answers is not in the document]
ndhanushkodi: removed this function, since it was printing with weird indents anyways!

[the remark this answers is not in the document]
ndhanushkodi: It was prefixing multiple lines of output though, originally!

cli/cmd/uninstall/uninstall.go
ndhanushkodi: I think its ok because in an error case, it will exit the function at the point of the error without printing either of these outputs here, and just print the first error it got to. Then when a user re-runs uninstall after fixing their issue, it should succeed and delete the secrets properly.

cli/cmd/uninstall/uninstall.go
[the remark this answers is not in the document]
ndhanushkodi: I think re-run is ok, where they manually fix, then re-run the same command: https://github.com/hashicorp/consul-k8s/pull/725/files#r713606519

cli/cmd/uninstall/uninstall.go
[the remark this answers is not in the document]
ndhanushkodi: Actually in testing I am noticing the consul-tls-init role, rolebinding, and service account are sticking around if they are not manually deleted. I went ahead and added that logic

cli/cmd/uninstall/uninstall.go
ndhanushkodi: I switched this to using the logger instead. In our case, the UI doesn't implement the closer interface, so Close() would be a no-op. But we could potentially choose to implement it in the future if we say "upgrade" our UI.

cli/cmd/uninstall/uninstall_test.go
```diff
@@ -0,0 +1,187 @@
...
+	_, err := c.kubernetes.RbacV1().RoleBindings("default").Create(context.TODO(), rolebinding, metav1.CreateOptions{})
+	require.NoError(t, err)
+	_, err = c.kubernetes.RbacV1().RoleBindings("default").Create(context.TODO(), rolebinding2, metav1.CreateOptions{})
+	require.NoError(t, err)
+	err = c.deleteRoleBindings("consul", "default")
+	require.NoError(t, err)
+	rolebindings, err := c.kubernetes.RbacV1().RoleBindings("default").List(context.TODO(), metav1.ListOptions{})
+	require.NoError(t, err)
+	require.Len(t, rolebindings.Items, 0)
+}
```
ishustava, line 165: I think we can also add tests for the `findExistingInstallation` function.

ishustava (approved):
Looks good ❤️  Thank you so much for considering the feedback so thoughtfully 🙏 

I've added some minor comments, but they are not blocking.

cli/cmd/uninstall/uninstall_test.go
ndhanushkodi: This one is kind of tricky for similar reasons to https://github.com/hashicorp/consul-k8s/blob/e77da79f99bbd1038f7b3dc203c6492127341e6b/cli/cmd/install/install.go#L338. 

I think I can open a separate PR this week to refactor the helm logic out of `uninstall.findExistingInstallation` and `install.checkForPreviousInstallations`, and unit test the part that's unrelated to helm logic. The reason to pull it out into a separate PR is since I'll be focusing on testing within the acceptance tests PR we can add this additional testing there after a quick refactor.

cli/cmd/uninstall/uninstall_test.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	pvc2 := &v1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "consul-server-test2",
+			Labels: map[string]string{
+				"release": "consul",
+			},
+		},
+	}
+	_, err := c.kubernetes.CoreV1().PersistentVolumeClaims("default").Create(context.TODO(), pvc, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
ndhanushkodi: Thanks! I'll switch to that

cli/cmd/uninstall/uninstall_test.go
```diff
@@ -0,0 +1,187 @@
...
+	c.kubernetes = fake.NewSimpleClientset()
+	pvc := &v1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "consul-server-test1",
+			Labels: map[string]string{
+				"release": "consul",
+			},
+		},
+	}
+	pvc2 := &v1.PersistentVolumeClaim{
```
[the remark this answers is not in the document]
ndhanushkodi: great idea to test that!! Thanks!

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,509 @@
...
+
+	// At this point, even if no Helm release was found and uninstalled, there could
+	// still be PVCs, Secrets, and Service Accounts left behind from a previous installation.
+	// If there isn't a foundReleaseName and foundReleaseNamespace, we'll use the values of the
+	// flags c.flagReleaseName and c.flagNamespace. If those are empty we'll fall back to defaults "consul" for the
+	// installation name and "consul" for the namespace.
+	if !found {
+		if c.flagReleaseName == "" || c.flagNamespace == "" {
+			foundReleaseName = "consul"
+			foundReleaseNamespace = "consul"
```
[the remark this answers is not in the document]
ndhanushkodi: good idea!

cli/cmd/uninstall/uninstall.go
```diff
@@ -0,0 +1,509 @@
...
+
+	found := false
+	foundReleaseName := ""
+	foundReleaseNamespace := ""
+	for _, rel := range res {
+		if rel.Chart.Metadata.Name == "consul" {
+			if c.flagNamespace != defaultAllNamespaces {
+				// If we found a chart named "consul" and -namespace was specified, we only found the release if the
+				// release namespace matches the -namespace flag.
+				if c.flagNamespace == rel.Namespace {
```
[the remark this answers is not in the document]
ndhanushkodi: I like the idea to combine it for readability! I think it's possible for empty string value for `rel.Namespace` to mean default namespace, and we can't test our way around that since we can't test the call to List itself, so it makes sense to combine the check

Issue #715, opened 2021-09-09
Pods on EKS Nodes don't start after upgrade to 0.33


jamiegwatkin:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

I've tried repeated times to use 0.33 with a brand new AWS EKS cluster, when using 0.33 the core AWS pods are never schedulesd and the node never enters a ready state which stops any other pods starting.

After deleting both mutating webhooks finally the main `aws-node` pods start, the node then enters a ready state and all other pods startup.

Before deleting the mutating webhooks the AWS node error is:
<img width="1533" alt="image" src="https://user-images.githubusercontent.com/10600045/132723379-e0b1e77b-0240-4c1f-96d1-edfc65b69c9c.png">

### Reproduction Steps

1. When running helm install with the following `values.yml`:
```
global:
  datacenter: hashidc1
  enabled: true 
  metrics:    
    enabled: true
    enableGatewayMetrics: true
    enableAgentMetrics: true   

ui:
  service:
    type: 'LoadBalancer'
    annotations: 
      "service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0"
  metrics:
    enabled: true
    provider: "prometheus"
    baseURL: http://prometheus-server


syncCatalog:
  enabled: true

connectInject:
  enabled: true
  metrics:
    defaultEnableMerging: true 

controller:
  enabled: true

ingressGateways:
  enabled: true
  gateways:
    - name: ingress-gateway
      service:
        type: LoadBalancer
        ports:
          - port: 80
            nodePort: null
          - port: 443
            nodePort: null
dns:
  enabled: true

prometheus:
  enabled: true

```
2. Pods don't start
3. After deleting the webhooks:
```
kubectl delete MutatingWebhookConfiguration consul-cluster-consul-controller-mutating-webhook-configuration
kubectl delete MutatingWebhookConfiguration consul-cluster-consul-connect-injector-cfg
```
[...]

PR #726, opened 2021-09-15
Default to excluding system ns's from injection

lkysow:
kube-system is excluded because it's unlikely users will want to
provision Connect pods in that namespace and also because we don't want
to block pods being provisioned there if our webhook injector is down.

local-path-storage is excluded because this ns is used by kind to
provision PVCs and if ACLs are enabled then the install gets into a
deadlock where:
- PVC can't be provisioned because Kind needs to create a Pod
- Pod can't be created because injector webhook needs to be up
- injector webhook can't come up until its got an ACL token
- ACL token can't be provisioned because Consul server isn't up
- Consul server can't be started because it doesn't have a PVC

NOTE: This matching is only supported in Kube 1.21+ where they've added
these labels to namespaces automatically now.

Should fix https://github.com/hashicorp/consul-k8s/issues/715

How I've tested this PR:
- Ran helm install on Kind using Kube 1.21 with ACLs and connect inject enabled. Saw that the servers never start. Then ran the same install with these changes. They do start.
- acceptance tests pass

- code, can try it out if you'd like too

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


lkysow:
I think the tests are failing because they're not using the actual chart changes in this PR.

PR #728, opened 2021-09-15
Add partition support to ingress and terminating gateways

thisisnotashwin:
Changes proposed in this PR:
- Add partition support to ingress and terminating gateway

How I've tested this PR:
- unit tests
- deployed the gateways to helm and watch them spin up successfully with the correct config.

- code review
- deploy it in a partition if you'd like but it could be complicated. I can drop in screen shots.

Checklist:
- [x] Tests added


ishustava:
Looks good. Do we need to update config entries for ingress and term gateways too?

kschoche (approved):
🔥

thisisnotashwin:
> Looks good. Do we need to update config entries for ingress and term gateways too?

The config entries will be created in the correct partition as the client that will be used to create them will reside in the correct partition. 😄

ishustava (approved):

PR #738, opened 2021-09-23
Automatically Generate and Use Gossip Encryption Key

t-eckert:
Changes proposed in this PR:
- Add `global.gossipEncryption.autoGenerate` to `values.yaml`
- Change `client-daemonset.yaml` and `server-statefulset.yaml` to pickup the generated gossip encryption key
- Add `gossip-encryption-autogeneration-job.yaml` to generate and set up the gossip encryption key
- Add `gossip-encryption-autogeneration...`
  - `-podsecuritypolicy.yaml`
  - `-role.yaml`
  - `-rolebinding.yaml`
  - `-serviceaccount.yaml` 
- Add `bats` tests for the `client-daemonset`, `server-statefulset`, and `gossip-encryption-autogen-*`

How I've tested this PR:
- Running the new `bats` tests and confirming that no current `bats` tests are broken
- Deployed the new templates using Helm and verified that gossip encryption is working correctly

- Testing `bats` for the relevant changes.
- Testing gossip encryption:
  On a fresh Kubernetes instance, from `charts/consul/` directory
  ```bash
  helm install consul . --set global.gossipEncryption.autoGenerate=true
  ```
  Wait for install to complete. Check that the gossip encryption secret has been created:
  ```bash
   kubectl get secrets consul-consul-gossip-encryption-key
   ```
  Exec into a server pod to check gossip encryption:
  ```bash
  kubectl exec consul-consul-server-0 -it -- /bin/sh
  ```
  Check if Consul is using the secret:
  ```bash
  curl http://127.0.0.1:8500/v1/operator/keyring | jq
  ```
  If successful, the gossip encryption key should be listed in the key ring.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


charts/consul/templates/gossip-encryption-autogeneration-job.yaml
```diff
@@ -0,0 +1,76 @@
...
+              # Check if secret already exists
+              currentGossipSecret=$(curl -s -X GET --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+                https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/${NAMESPACE}/secrets/${secretName} \
+                -H "Authorization: Bearer $( cat /var/run/secrets/kubernetes.io/serviceaccount/token )" \
+                -H "Content-Type: application/json" \
+                -H "Accept: application/json")
+              doesCurrentSecretNameExist=$(echo $currentGossipSecret | jq '.code==200')
+              if [ "$doesCurrentSecretNameExist" = "true" ]; then
+                doesCurrentSecretKeyExist=$(echo $currentGossipSecret | jq '.data.key')
+              fi
```
thisisnotashwin, line 59: Can you elaborate on the scenario here? If I understand this correctly, this is a case where the secret exists, but the secret does not have a value set on the key `key`?  I'm a touch hesitant about modifying an existing secret that wasn't created by this job.

Additionally, is there a way for us to write a BATS test for this?

charts/consul/test/unit/gossip-encryption-autogeneration-job.bats
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "gossipEncryptionAutogeneration/Job: disabled by default" {
+  cd `chart_dir`
+  assert_empty helm template \
+      -s templates/gossip-encryption-autogen-job.yaml  \
```
thisisnotashwin, line 8: I think every test in this file needs to get updated with this change
```suggestion
      -s templates/gossip-encryption-autogeneration-job.yaml  \
```

thisisnotashwin (commented):
Hey @t-eckert !!

This looks good. I have a couple of questions and suggestions.
Can we write/update an acceptance test to use this to validate that it works as expected? Maybe update the basic test?

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
```diff
@@ -0,0 +1,76 @@
+{{- if .Values.global.gossipEncryption.autoGenerate }}
+{{- if (or .Values.global.gossipEncryption.secretName .Values.global.gossipEncryption.secretKey) }}
+  {{ fail "If global.gossipEncryption.autoGenerate is true, global.gossipEncryption.secretName and global.gossipEncryption.secretKey must not be set." }}
```
ishustava, line 3: ❤️

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
```diff
@@ -0,0 +1,76 @@
...
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
+spec:
+  template:
+    metadata:
+      name: {{ template "consul.fullname" . }}-gossip-encryption-autogenerate
+      labels:
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: gossip-encryption-autogen
```
ishustava, line 29: nit: could we make the name consistent?
```suggestion
        component: gossip-encryption-autogeneneration
```

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
ishustava: Yeah I think we need to exit 0 in this case since we don't want to modify an existing secret.

I think we can also use a POST request and ignore secrets already exists response code. I think it'll make the logic a bit simpler. It's what we used to do in the tls-init job: https://github.com/hashicorp/consul-helm/blob/c0722c63907fe22ca92f7b021caf0647de8c0028/templates/tls-init-job.yaml#L55-L58

charts/consul/templates/gossip-encryption-autogeneration-role.yaml
```diff
@@ -0,0 +1,33 @@
...
+    "helm.sh/hook-delete-policy": before-hook-creation
+rules:
+- apiGroups: [""]
+  resources:
+    - secrets
+  verbs:
+    - create
+    - update
+    - get
+    - list
```
ishustava, line 23: should it be just create and get?

charts/consul/test/unit/client-daemonset.bats
```diff
@@ -606,6 +606,16 @@ load _helpers
...
 }
 
+@test "client/DaemonSet: gossip encryption autogeneration properly sets secretName and secretKey" {
+  cd `chart_dir`
+  local actual=$(helm template \
+    -s templates/client-daemonset.yaml  \
+    --set 'global.gossipEncryption.autoGenerate=true' \
+    . | tee /dev/stderr |
+    yq '.spec.template.spec.containers[] | select(.name=="consul") | .env[] | select(.name == "GOSSIP_KEY") | .valueFrom.secretKeyRef | [.name=="RELEASE-NAME-consul-gossip-encryption-key", .key="key"] | all' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
```
ishustava, line 616: could you also add an assertion that the key is passed via `-encrypt` key too?

charts/consul/test/unit/gossip-encryption-autogeneration-job.bats
```diff
@@ -0,0 +1,74 @@
...
+      --set 'global.gossipEncryption.autoGenerate=true' \
+      --set 'global.gossipEncryption.secretName=name' \
+      --set 'global.gossipEncryption.secretKey=key' \
+      . 
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "If global.gossipEncryption.autoGenerate is true, global.gossipEncryption.secretName and global.gossipEncryption.secretKey must not be set." ]]
+}
+
+@test "gossipEncryptionAutogeneration/Job: fails if global.gossipEncryption.autoGenerate=true and global.gossipEncryption.secretName are set" {
```
ishustava, line 42:
```suggestion
@test "gossipEncryptionAutogeneration/Job: fails if global.gossipEncryption.autoGenerate=true and global.gossipEncryption.secretName is set" {
```

charts/consul/test/unit/gossip-encryption-autogeneration-job.bats
```diff
@@ -0,0 +1,74 @@
...
+  run helm template \
+      -s templates/gossip-encryption-autogen-job.yaml  \
+      --set 'global.gossipEncryption.autoGenerate=true' \
+      --set 'global.gossipEncryption.secretName=name' \
+      . 
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "If global.gossipEncryption.autoGenerate is true, global.gossipEncryption.secretName and global.gossipEncryption.secretKey must not be set." ]]
+}
+
+@test "gossipEncryptionAutogeneration/Job: fails if global.gossipEncryption.autoGenerate=true and global.gossipEncryption.secretKey are set" {
```
ishustava, line 53:
```suggestion
@test "gossipEncryptionAutogeneration/Job: fails if global.gossipEncryption.autoGenerate=true and global.gossipEncryption.secretKey is set" {
```

charts/consul/test/unit/gossip-encryption-autogeneration-job.bats
```diff
@@ -0,0 +1,74 @@
...
+
+@test "gossipEncryptionAutogeneration/Job: secretName and secretKey are generated" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/gossip-encryption-autogen-job.yaml \
+      --set 'global.gossipEncryption.autoGenerate=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | any(contains("secretName=RELEASE-NAME-consul-gossip-encryption-key\nsecretKey=key"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
ishustava, line 73: I don't think we can test this sufficiently in bats. Generally, if there's no templating logic in the command, then it won't be valuable to write a bats test for it. So I'd say remove this case altogether. 

It'd be nice to add this to at least our basic acceptance test. You'd just need to add it to helm values, setting
`global.gossipEncryption.autoGenerate` to `strconv.FormatBool(c.secure)` so that it's tested together with other secure features like TLS and ACLs:

https://github.com/hashicorp/consul-k8s/blob/9a974c1a9c9fe0594460352b25ef05fd4ba13734/charts/consul/test/acceptance/tests/basic/basic_test.go#L41-L45

charts/consul/test/unit/gossip-encryption-autogeneration-podsecurity.bats
```diff
@@ -0,0 +1,28 @@
...
+
+@test "gossipEncryptionAutogeneration/PodSecurityPolicy: disabled with global.gossipEncryption.autoGenerate=false" {
+  cd `chart_dir`
+  assert_empty helm template \
+      -s templates/gossip-encryption-autogeneration-podsecuritypolicy.yaml  \
+      --set 'global.gossipEncryption.autoGenerate=false' \
+      .
+}
+
+@test "gossipEncryptionAutogeneration/PodSecurityPolicy: enabled with global.gossipEncryption.autoGenerate=test" {
```
ishustava, line 20:
```suggestion
@test "gossipEncryptionAutogeneration/PodSecurityPolicy: enabled with global.gossipEncryption.autoGenerate=true" {
```

charts/consul/values.yaml
```diff
@@ -117,26 +117,19 @@ global:
...
   # ```
-  #
-  # To reference, use:
-  #
-  # ```yaml
-  # global:
-  #   gossipEncryption:
-  #     secretName: consul-gossip-encryption-key
-  #     secretKey: key
+  # $ kubectl create secret generic <secretName> --from-literal=<secretKey>=$(consul keygen)
```
ishustava, line 128: could we keep `kubectl` command the same? I think it's nice that you can just copy-paste it, and it'll work.

charts/consul/values.yaml
```diff
@@ -117,26 +117,19 @@ global:
...
   # ```
   gossipEncryption:
+    # automatically generate a gossip encryption key and load it as a Kubernetes secret
```
ishustava, line 131:
```suggestion
    # Automatically generate a gossip encryption key and save it to a Kubernetes secret.
```

ishustava (commented):
Great work, Thomas! I left some comments inline.

charts/consul/test/unit/gossip-encryption-autogeneration-job.bats
t-eckert: Great catch. I let this one slip by unchanged when I modified the filename.

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
t-eckert: I'll try to replicate the tls code. That is helpful. Then I don't need to do all the checks. I'll reach out after I implement it that way to make sure I did it right.

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
```diff
@@ -0,0 +1,76 @@
...
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: gossip-encryption-autogeneneration
+      annotations:
+        "consul.hashicorp.com/connect-inject": "false"
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ template "consul.fullname" . }}-gossip-encryption-autogenerate
+      containers:
+        - name: gossip-encryption-autogen
```
ishustava, line 36: I forgot to mention that we should add [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) to this pod so that it doesn't run as root. We're using the consul image here, and consul docker image runs as root by default unless you overwrite it. Here is an example of how we do it for the consul server:

https://github.com/hashicorp/consul-k8s/blob/9a974c1a9c9fe0594460352b25ef05fd4ba13734/charts/consul/templates/server-statefulset.yaml#L79

https://github.com/hashicorp/consul-k8s/blob/9a974c1a9c9fe0594460352b25ef05fd4ba13734/charts/consul/values.yaml#L508-L512

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
t-eckert: This works now by just POSTing, but it won't overwrite the secret if it's already there.

charts/consul/templates/gossip-encryption-autogeneration-role.yaml
t-eckert: Yes

charts/consul/values.yaml
t-eckert: Yep! That's a great point!

charts/consul/test/unit/client-daemonset.bats
t-eckert: Yes, great callout. Added

charts/consul/test/unit/gossip-encryption-autogeneration-job.bats
t-eckert: I started work on this, but ran into problems (I think I messed up my k8s cluster somehow). Can we sync on this and tackle it together next week?

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
t-eckert: Yes, and I'm going to just hardcode the security context because there is no reason this should ever need to run as root.

t-eckert: Done!

charts/consul/templates/gossip-encryption-autogeneration-podsecuritypolicy.yaml
```diff
@@ -0,0 +1,39 @@
...
+  name: {{ template "consul.fullname" . }}-gossip-encryption-autogenerate
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
thisisnotashwin, line 15: Does this need a `hook-succeeded` delete policy as well like the job? Is there a reason only the job has it?

charts/consul/templates/gossip-encryption-autogeneration-role.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: {{ template "consul.fullname" . }}-gossip-encryption-autogenerate
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
thisisnotashwin, line 14: Same question as above about the `hook-succeeded` policy.

charts/consul/templates/gossip-encryption-autogeneration-rolebinding.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: {{ template "consul.fullname" . }}-gossip-encryption-autogenerate
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
thisisnotashwin, line 14: Same question as above about the `hook-succeeded` policy.

charts/consul/templates/gossip-encryption-autogeneration-serviceaccount.yaml
```diff
@@ -0,0 +1,21 @@
...
+  name: {{ template "consul.fullname" . }}-gossip-encryption-autogenerate
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
thisisnotashwin, line 14: Same question as above about the `hook-succeeded` policy.

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
+spec:
+  template:
+    metadata:
+      name: {{ template "consul.fullname" . }}-gossip-encryption-autogenerate
+      labels:
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: gossip-encryption-autogeneneration
```
thisisnotashwin, line 29: Can we update this to `gossip-encryption-autogenerate`? Also, we seem to call the files `ecnryption-autogeneration-*` but the names of the jobs and related resources seems to use `encryption-autogenerate-*`. I think we should pick one and standardize the file names and resources name to it.

charts/consul/test/acceptance/tests/basic/basic_test.go
```diff
@@ -1,15 +1,18 @@
...
 	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/consul"
 	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/helpers"
 	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/logger"
 	"github.com/hashicorp/consul/api"
 	"github.com/stretchr/testify/require"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
thisisnotashwin, line 15: this is a nit, but can we rename this alias to `metav1`? I don't know if we have been consistent across the project but most kubernetes related imports follow the paradigm `/core/v1` -> `corev1`, `/meta/v1` -> `metav1`

charts/consul/test/acceptance/tests/basic/basic_test.go
```diff
@@ -63,6 +67,23 @@ func TestBasicInstallation(t *testing.T) {
 			kv, _, err := client.KV().Get(randomKey, nil)
 			require.NoError(t, err)
 			require.Equal(t, kv.Value, randomValue)
+
+			// Check that autogenerated gossip encryption key is being used
+			if c.secure {
+				secretName := releaseName + "-consul-gossip-encryption-key"
```
thisisnotashwin, line 73:
```suggestion
				secretName := fmt.Sprintf("%s-consul-gossip-encryption-key", releaseName)
```

charts/consul/test/acceptance/tests/basic/basic_test.go
```diff
@@ -63,6 +67,23 @@ func TestBasicInstallation(t *testing.T) {
...
+				keyring, err := client.Operator().KeyringList(nil)
+				require.NoError(t, err)
+
+				testContext := suite.Environment().DefaultContext(t)
+				secret, err := testContext.KubernetesClient(t).CoreV1().Secrets(testContext.KubectlOptions(t).Namespace).Get(context.Background(), secretName, v1.GetOptions{})
+				require.NoError(t, err)
+				gossipEncryptionKey := strings.TrimSpace(string(secret.Data[secretKey]))
+
+				require.Len(t, keyring, 2)
+				require.Contains(t, keyring[0].Keys, gossipEncryptionKey)
```
thisisnotashwin, line 85: ❤️ love the test!

charts/consul/test/unit/client-daemonset.bats
```diff
@@ -606,6 +606,26 @@ load _helpers
...
+@test "client/DaemonSet: gossip encryption key is passed in via the -encrypt flag" {
+  cd `chart_dir`
+  local actual=$(helm template \
+    -s templates/client-daemonset.yaml  \
+    --set 'global.gossipEncryption.autoGenerate=true' \
+    . | tee /dev/stderr |
+    yq '.spec.template.spec.containers[] | select(.name=="consul") | .command | any(contains("-encrypt=\"${GOSSIP_KEY}\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
```
thisisnotashwin, line 628: NICE!

charts/consul/test/unit/gossip-encryption-autogeneration-job.bats
```diff
@@ -0,0 +1,74 @@
...
+
+@test "gossipEncryptionAutogeneration/Job: secretName and secretKey are generated" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/gossip-encryption-autogeneration-job.yaml \
+      --set 'global.gossipEncryption.autoGenerate=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | any(contains("secretName=RELEASE-NAME-consul-gossip-encryption-key\nsecretKey=key"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
thisisnotashwin, line 73: This assertion feels arbitrary IMO. I think we can delete it.

charts/consul/test/unit/server-statefulset.bats
```diff
@@ -842,6 +842,16 @@ load _helpers
...
+@test "server/StatefulSet: gossip encryption autogeneration properly sets secretName and secretKey" {
+  cd `chart_dir`
+  local actual=$(helm template \
+    -s templates/server-statefulset.yaml  \
+    --set 'global.gossipEncryption.autoGenerate=true' \
+    . | tee /dev/stderr |
+    yq '.spec.template.spec.containers[] | select(.name=="consul") | .env[] | select(.name == "GOSSIP_KEY") | .valueFrom.secretKeyRef | [.name=="RELEASE-NAME-consul-gossip-encryption-key", .key="key"] | all' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
```
thisisnotashwin, line 854: Can we add a similar test here about testing the `-encrypt` flag being passed to the command?

thisisnotashwin (commented):
Hey @t-eckert !! This is looking really good. I left some more comments but this is looking close to done!! 👍

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
t-eckert: Yeah, I think I'm going to standardize on `gossip-encryption-autogenerate`. I think that's cleaner.

thisisnotashwin: SGTM

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+      annotations:
+        "consul.hashicorp.com/connect-inject": "false"
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ template "consul.fullname" . }}-gossip-encryption-autogenerate
+      securityContext:
+        runAsNonRoot: true
+        runAsGroup: 1000 
+        runAsUser: 100 
+        fsGroup: 1000
```
ishustava: ❤️ thank you!

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+      containers:
+        - name: gossip-encryption-autogen
+          image: "{{ .Values.global.image }}"
+          env:
+            - name: NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          workingDir: /tmp
```
ishustava: We don't need this since we're not saving any files.
```suggestion
```

charts/consul/templates/gossip-encryption-autogeneration-podsecuritypolicy.yaml
ishustava: If you add `hook-succeeded` to these, then helm will delete it immediately after creation, since for a psp to succeed you just need to create it. Basically, helm will delete it before the job even starts. This is because there's no way to group different resources into one logical hook in helm, so it actually treats all these as separate hooks, meaning the job and the psp are separate independent helm hooks.

Hope it makes sense.

charts/consul/templates/gossip-encryption-autogeneration-role.yaml
```diff
@@ -0,0 +1,31 @@
...
+rules:
+- apiGroups: [""]
+  resources:
+    - secrets
+  verbs:
+    - create
+    - get
```
ishustava: don't need get anymore since we're only doing `POST` requests.
```suggestion
```

charts/consul/test/unit/server-statefulset.bats
ishustava: +1

ishustava (approved):
Great work Thomas. I left some minor comments and also agree with Ashwin's comments. I'm approving though, assuming those get resolved.

charts/consul/templates/gossip-encryption-autogeneration-role.yaml
t-eckert: Adding it would kill the role before the job starts. I understand where the call out comes from though.

charts/consul/test/acceptance/tests/basic/basic_test.go
t-eckert: Thanks so much!

charts/consul/test/acceptance/tests/basic/basic_test.go
t-eckert: Nits are good, I'm still learning what idiomatic code looks like on the team.

charts/consul/templates/gossip-encryption-autogeneration-role.yaml
thisisnotashwin: aha!! that makes sense! thanks for the clarification!!

charts/consul/templates/gossip-encryption-autogenerate-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+          # return a 409 because these secrets would already exist;
+          # we are ignoring these response codes.
+          command:
+            - "/bin/sh"
+            - "-ec"
+            - |
+              secretName={{ template "consul.fullname" . }}-gossip-encryption-key
+              secretKey=key
+              keyValue=$(consul keygen | base64)
+              curl -s -X POST --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
```
lkysow, line 60: Right now if this errors out for a legit reason (like service account wrong) it will fail silently. I think instead we want it to fail and have kube re-run the job.

That would require parsing the response code to ignore 409s/200s which is getting a bit ugly so probably it does make sense to write this in Golang–as a follow-up PR.

charts/consul/test/acceptance/tests/basic/basic_test.go
```diff
@@ -1,15 +1,18 @@
...
 	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/consul"
 	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/helpers"
 	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/logger"
 	"github.com/hashicorp/consul/api"
 	"github.com/stretchr/testify/require"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
lkysow, line 15:
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

charts/consul/test/acceptance/tests/basic/basic_test.go
```diff
@@ -63,6 +67,23 @@ func TestBasicInstallation(t *testing.T) {
...
+				keyring, err := client.Operator().KeyringList(nil)
+				require.NoError(t, err)
+
+				testContext := suite.Environment().DefaultContext(t)
+				secret, err := testContext.KubernetesClient(t).CoreV1().Secrets(testContext.KubectlOptions(t).Namespace).Get(context.Background(), secretName, metaV1.GetOptions{})
+				require.NoError(t, err)
+				gossipEncryptionKey := strings.TrimSpace(string(secret.Data[secretKey]))
+
+				require.Len(t, keyring, 2)
+				require.Contains(t, keyring[0].Keys, gossipEncryptionKey)
```
lkysow, line 85: could test both are equal

charts/consul/values.yaml
```diff
@@ -117,26 +117,19 @@ global:
...
-  # ```yaml
-  # global:
-  #   gossipEncryption:
-  #     secretName: consul-gossip-encryption-key
-  #     secretKey: key
+  # $ kubectl create secret generic consul-gossip-encryption-key --from-literal=key=$(consul keygen)
   # ```
   gossipEncryption:
+    # Automatically generate a gossip encryption key and save it to a Kubernetes secret.
+    autoGenerate: false
```
lkysow, line 132: Should document that `secretName` and `secretKey` shouldn't be set if this is `true`.

charts/consul/values.yaml
```diff
@@ -117,26 +117,19 @@ global:
...
   enablePodSecurityPolicies: false
 
-  # Configures which Kubernetes secret to retrieve Consul's
-  # gossip encryption key from (see `-encrypt` (https://consul.io/docs/agent/options#_encrypt)). If secretName or
-  # secretKey are not set, gossip encryption will not be enabled. The secret must
-  # be in the same namespace that Consul is installed into.
+  # Configures Consul's gossip encryption key, set as a Kubernetes secret
+  # (see `-encrypt` (https://consul.io/docs/agent/options#_encrypt)).
+  # By default, gossip encryption is not enabled. The gossip encryption key may be set automatically or manually.
+  # To automatically generate and set a gossip encryption key, set autoGenerate to true.
```
lkysow, line 123: Maybe we could say "the recommended method is to automatically generate the key using autoGenerate=true`

charts/consul/test/unit/gossip-encryption-autogeneration-job.bats
t-eckert: Agreed

charts/consul/test/acceptance/tests/basic/basic_test.go
t-eckert: Maybe I will then...

charts/consul/templates/gossip-encryption-autogeneration-job.yaml
t-eckert: Michelle Tanner says "You got it, dude!"

charts/consul/templates/gossip-encryption-autogenerate-job.yaml
t-eckert: Sounds good! I will merge this when I get approval and then double back around immediately to write the process in Go.

charts/consul/test/unit/server-statefulset.bats
t-eckert: Great catch, added.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ IMPROVEMENTS:
 * Control Plane
   * Upgrade Docker image Alpine version from 3.13 to 3.14. [[GH-737](https://github.com/hashicorp/consul-k8s/pull/737)]
 * Helm Chart
+  * Add automatic generation of gossip encryption with `global.gossipEncryption.autoGenerate=true` [[GH-738](https://github.com/hashicorp/consul-k8s/pull/738)]
```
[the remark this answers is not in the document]
t-eckert: Yay!

charts/consul/test/acceptance/tests/basic/basic_test.go
```diff
@@ -63,6 +67,24 @@ func TestBasicInstallation(t *testing.T) {
...
+			// Check that autogenerated gossip encryption key is being used
+			if c.secure {
+				secretName := fmt.Sprintf("%s-consul-gossip-encryption-key", releaseName)
+				secretKey := "key"
+
+				keyring, err := client.Operator().KeyringList(nil)
+				require.NoError(t, err)
+
+				testContext := suite.Environment().DefaultContext(t)
+				secret, err := testContext.KubernetesClient(t).CoreV1().Secrets(testContext.KubectlOptions(t).Namespace).Get(context.Background(), secretName, metaV1.GetOptions{})
```
thisisnotashwin, line 80:
```suggestion
				secret, err := testContext.KubernetesClient(t).CoreV1().Secrets(testContext.KubectlOptions(t).Namespace).Get(context.Background(), secretName, metav1.GetOptions{})
```

thisisnotashwin (approved):
This looks good. Added the suggestion that should fix the tests. Nice job on the PR!

reminder to squash and merge the PR 😅

PR #740, opened 2021-09-23
consul-server: allow users to mount their own volumes

cabrinha:
Changes proposed in this PR:
- Allow users to mount their own volumes

How I've tested this PR: `helm template`

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


david-yu:
I'll go ahead and close as this PR has been open for a while,  and there has not been follow up since the last reply. Please do re-open if you are able to also find a way to enable the previous supported workflows.

PR #741, opened 2021-09-23
Add support for version command

lkysow:
How I've tested this PR:
- compiled and run

- compile and run

Checklist:
- Tests added: No tests added. This matches what we're doing in control-plane (no tests)
- CHANGELOG entry added


cli/cmd/version/version.go
```diff
@@ -2,49 +2,32 @@ package version
 
 import (
 	"fmt"
-	"strings"
-)
+	"sync"
 
-var (
```
lkysow, line 8: this code moved to `version/version` in top-level

control-plane/subcommand/version/command.go
```diff
@@ -12,7 +12,7 @@ type Command struct {
 }
 
 func (c *Command) Run(_ []string) int {
-	c.UI.Output(fmt.Sprintf("consul-k8s %s", c.Version))
+	c.UI.Output(fmt.Sprintf("consul-k8s-control-plane %s", c.Version))
```
lkysow, line 15: fix missing change from monorepo move

ndhanushkodi (approved):
Thanks for adding this, looks great! 

We did have one changelog entry that went into the alpha release (only install was merged at the time) https://github.com/hashicorp/consul-k8s/blob/main/CHANGELOG.md#0340-september-17-2021

So this could be added to the changelog as support for version command. 

I think the changelog is missing the uninstall command as well

cli/cmd/version/version.go
```diff
@@ -2,49 +2,32 @@ package version
...
+}
+
+func (c *Command) Run(_ []string) int {
+	c.once.Do(c.init)
+	c.UI.Output(fmt.Sprintf("consul-k8s %s", c.Version))
+	return 0
+}
+
+func (c *Command) Synopsis() string {
+	return "Prints the version."
```
t-eckert, line 28:
```suggestion
	return "Prints the version of the CLI."
```

cli/cmd/version/version.go
```diff
@@ -2,49 +2,32 @@ package version
...
+}
+
+func (c *Command) Help() string {
+	return ""
```
t-eckert, line 32: Should there be something returned here?

t-eckert (approved):
Just a few comments

cli/cmd/version/version.go
lkysow: yeah good call, I copied https://github.com/hashicorp/consul/blob/main/command/version/version.go#L90

PR #747, opened 2021-09-27
acceptance tests for admin partitions

thisisnotashwin:
Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


charts/consul/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: consul
 version: 0.34.1
-appVersion: 1.10.2
+appVersion: 1.11.0-alpha
```
thisisnotashwin, line 4: Given we need to update these values to 1.11.0-alpha for these tests to run, does it make sense to merge this PR? 🤔

thisisnotashwin: I guess i could explicitly set the image in the CI pipelines? that way only the PRs run against 1.11 but nightlies use 1.10

charts/consul/test/acceptance/tests/fixtures/cases/static-client-partition/patch.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: static-client
+spec:
+  template:
+    metadata:
+      annotations:
+        "consul.hashicorp.com/connect-inject": "true"
+        "consul.hashicorp.com/connect-service-upstreams": "static-server.default.secondary"
```
ishustava, line 10: I guess we can change it later since it's not tested right now, but I think it'll need a port too.

charts/consul/test/acceptance/tests/partitions/partitions_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/helpers"
+	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/k8s"
+	"github.com/hashicorp/consul-k8s/charts/consul/test/acceptance/framework/logger"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Test that Connect works in a default installation.
+// i.e. without ACLs because TLS is required for setting up Admin Partitions.
+func TestMeshGatewayDefault(t *testing.T) {
```
ishustava, line 19:
```suggestion
func TestPartitions(t *testing.T) {
```

charts/consul/test/acceptance/tests/partitions/partitions_test.go
```diff
@@ -0,0 +1,148 @@
...
+		"global.image":      "hashicorp/consul-enterprise:1.11.0-ent-alpha",
+
+		"global.adminPartitions.enabled": "true",
+		"global.enableConsulNamespaces":  "true",
+		"global.tls.enabled":             "true",
+
+		"server.exposeGossipAndRPCPorts": "true",
+
+		"connectInject.enabled": "true",
+		"controller.enabled":    "true",
```
ishustava, line 47: do we need controller?

charts/consul/test/acceptance/tests/partitions/partitions_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	if cfg.UseKind {
+		primaryHelmValues["global.adminPartitions.service.type"] = "NodePort"
+		primaryHelmValues["global.adminPartitions.service.nodePort.https"] = "30000"
+	}
+
+	releaseName := helpers.RandomName()
+
+	// Install the primary consul cluster in the default kubernetes context
```
ishustava, line 57: nit: Should we use other names for the clusters, i.e. not "primary"
```suggestion
	// Install the primary consul cluster in the default kubernetes context.
```

Also another nit: I think other go comments need to be updated (it's probably my fault since I wrote the mesh gw tests)

charts/consul/test/acceptance/tests/partitions/partitions_test.go
```diff
@@ -0,0 +1,148 @@
...
+	require.NoError(t, err)
+
+	var partitionSvcIP string
+	if !cfg.UseKind {
+		// Get the IP of the partition service to configure the external server address in the values file for the workload cluster.
+		partitionServiceName := fmt.Sprintf("%s-partition-secret", releaseName)
+		logger.Logf(t, "retrieving partition service to determine external IP for servers")
+		partitionsSvc, err := primaryContext.KubernetesClient(t).CoreV1().Services(primaryContext.KubectlOptions(t).Namespace).Get(ctx, partitionServiceName, metav1.GetOptions{})
+		require.NoError(t, err)
+		partitionSvcIP = partitionsSvc.Status.LoadBalancer.Ingress[0].IP
```
ishustava, line 85: Why do we have this logic if the tests only run on kind? I'm wondering if we should leave it out until it's needed, but maybe you wanted to keep it because we'll most likely need it in the future.

charts/consul/test/acceptance/tests/partitions/partitions_test.go
```diff
@@ -0,0 +1,148 @@
...
+	} else {
+		nodeList, err := primaryContext.KubernetesClient(t).CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		require.NoError(t, err)
+		for _, node := range nodeList.Items {
+			partitionSvcIP = node.Status.Addresses[0].Address
```
ishustava, line 90: since it's kind should we assume that there's only one node and skip the for loop? I think the for loop reads a bit confusing.

charts/consul/test/acceptance/tests/partitions/partitions_test.go
```diff
@@ -0,0 +1,148 @@
...
+	output, err := k8s.RunKubectlAndGetOutputE(t, secondaryContext.KubectlOptions(t), "logs", agentPodList.Items[0].Name, "-n", secondaryContext.KubectlOptions(t).Namespace)
+	require.NoError(t, err)
+	require.Contains(t, output, "Partition: 'secondary'")
+
+	// Check that we can connect services over the mesh gateways
+	logger.Log(t, "creating static-server in workload cluster")
+	k8s.DeployKustomize(t, secondaryContext.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-server-inject")
+
+	logger.Log(t, "creating static-client in server cluster")
+	k8s.DeployKustomize(t, primaryContext.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-client-partition")
```
ishustava, line 143: should we comment these out too?

ishustava (approved):
Great work, Ashwin!  I left some minor comments, but they're not blocking.

charts/consul/test/acceptance/tests/partitions/partitions_test.go
thisisnotashwin: 😓 oops

charts/consul/test/acceptance/tests/partitions/partitions_test.go
thisisnotashwin: good catch!

charts/consul/test/acceptance/tests/partitions/partitions_test.go
thisisnotashwin: you shouldn't take the blame for my lazy copy-pasting lol.

charts/consul/test/acceptance/tests/partitions/partitions_test.go
thisisnotashwin: That makes sense. I will remove the loop and just grab the 0th address and add a comment.

charts/consul/test/acceptance/tests/partitions/partitions_test.go
thisisnotashwin: Yeah. I added it for exactly that reason. We will need it in the long run.

ishustava: sounds good

charts/consul/test/acceptance/tests/fixtures/cases/static-client-partition/patch.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: static-client
+spec:
+  template:
+    metadata:
+      annotations:
+        "consul.hashicorp.com/connect-inject": "true"
+        "consul.hashicorp.com/connect-service-upstreams": "static-server.default.secondary:8080"
```
ishustava, line 10: Because the commented out assertion is doing `curl localhost:1234`. Or we can change the way around. 
```suggestion
        "consul.hashicorp.com/connect-service-upstreams": "static-server.default.secondary:1234"
```

charts/consul/test/acceptance/tests/partitions/partitions_test.go
ishustava: lol you are too kind

charts/consul/test/acceptance/tests/fixtures/cases/static-client-partition/patch.yaml
thisisnotashwin: i confused concepts :sigh:
Thank you!! 🙏

kschoche (approved):
Sorry for the late review, this looks great!

PR #752, opened 2021-09-29
acceptance-tests-framework: create ent image from values.yaml isntead of Chart.yaml

ishustava:
Changes proposed in this PR:
- ☝️ Previously we'd parse the consul version from Chart.yaml, but this is confusing if we're trying to test a consul version that hasn't been released yet when it wouldn't make sense to update chart.yaml with it. This could result in false positives if we run tests by setting `global.image` and pass `-enable-enterprise` flag, since that will overwrite `global.image` to the ent version of the Consul image from Chart.yaml.

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/test/acceptance/framework/config/config.go
+++ b/charts/consul/test/acceptance/framework/config/config.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


charts/consul/test/acceptance/framework/config/config.go
```diff
@@ -91,38 +90,48 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
...
-		split := strings.Split(appVersion, "-")
-		appVersion = split[0]
+	if strings.Contains(consulImageVersion, "-") {
+		split := strings.Split(consulImageVersion, "-")
+		consulImageVersion = split[0]
 		preRelease = fmt.Sprintf("-%s", split[1])
 	}
 
-	return fmt.Sprintf("hashicorp/consul-enterprise:%s-ent%s", appVersion, preRelease), nil
+	return fmt.Sprintf("hashicorp/consul-enterprise:%s-ent%s", consulImageVersion, preRelease), nil
```
thisisnotashwin, line 134: Can we add an exception here to images in the `repo/image@sha256:oioi2452345kjhlkh` form? And a test for it as well. 🙏

thisisnotashwin (commented):
Hey @ishustava !

This looks great. I have a suggestion which I realize might be a feature request so ignore it if is complicated. (but I would really like it if it is there, but can add that myself later)

charts/consul/test/acceptance/framework/config/config.go
ishustava: yeah sure! is there a use case for setting these in `values.yaml`? What should be the behavior in that case?

thisisnotashwin: In this case, i would expect it to just use the provided image. A specific use case would be me wanting to run acceptance tests for an enterprise feature against a consul image that hasn't been released.

thisisnotashwin: I guess it doesnt HAVE to be set in values.yaml but it is much easier to do as compared to updating the CI config, but that's me. Like I mentioned, there is a workaround for that so I wouldn't sweat it.

ishustava: that makes sense! It should be easy to add.

thisisnotashwin: youre very kind!! 🙏

ishustava: added!

charts/consul/test/acceptance/framework/config/config_test.go
```diff
@@ -116,39 +116,42 @@ func TestConfig_HelmValuesFromConfig(t *testing.T) {
...
 		},
 		{
-			appVersion: "1",
-			expErr:     "unable to cast chartMap.appVersion to string",
+			consulImage: "invalid",
+			expErr:      "could not determine consul version from global.image: invalid",
+		},
+		{
+			consulImage: "hashicorp/consul@sha256:oioi2452345kjhlkh",
+			expImage:    "hashicorp/consul@sha256:oioi2452345kjhlkh",
```
thisisnotashwin, line 141: ❤️

thisisnotashwin (approved):
greatest PR ever!

t-eckert (approved):
🚀🚀🚀

PR #743, opened 2021-09-25
CLI acceptance tests

ndhanushkodi:
Changes proposed in this PR:
- TODO: refactor to reuse test code

How I've tested this PR:

How I expect reviewers to test this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #757, opened 2021-10-04
Embed Helm chart templates in CLI

ndhanushkodi:
Changes proposed in this PR:
- Make the helm charts directory a go module where the helm chart files are embedded into an exported variable
- Import the variable into the CLI to embed it at compilation time. The replace directive makes it so that you don't need to bump the dependency on changes to the charts files.
- New functions to read those local templates and load the install from there

How I've tested this PR:
- unit tests and ran CLI smoke tests (#743) with this binary

- review

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


.circleci/config.yml
```diff
@@ -924,12 +924,12 @@ workflows:
...
             - dev-upload-docker
-            - unit-test-helm-templates
+#            - unit-test-helm-templates
             - unit-acceptance-framework
       - acceptance-tproxy:
           requires:
             - dev-upload-docker
-            - unit-test-helm-templates
+#            - unit-test-helm-templates
             - unit-acceptance-framework
```
ndhanushkodi, line 933: These will be reverted before merging.

charts/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/hashicorp/consul-k8s/charts
+
+go 1.16
```
t-eckert, line 3: Is this the minimal version? Should we be using 1.17 now that it's out?

charts/embed_chart.go
```diff
@@ -0,0 +1,15 @@
+package charts
+
+import (
```
t-eckert, line 3:
```suggestion
import "embed"
```

t-eckert, line 3: Can this just be 

``` go
import "embed"
```

cli/cmd/common/utils.go
```diff
@@ -1,20 +1,81 @@
...
 const (
 	DefaultReleaseName      = "consul"
 	DefaultReleaseNamespace = "consul"
+	chartFileName           = "Chart.yaml"
+	valuesFileName          = "values.yaml"
+	templatesDirName        = "templates"
+	TopLevelChartDirName    = "consul"
 )
 
+// ReadChartFiles reads the chart files from the embedded FS, and loads their contents into []*loader.BufferedFile. This
```
t-eckert, line 24: Can we say "file system" instead of FS? Just for clarity.

```suggestion
// ReadChartFiles reads the chart files from the embedded file system, and loads their contents into []*loader.BufferedFile. This
```

t-eckert: Overall, this is a great comment! Makes things super clear!

charts/go.mod
ndhanushkodi: I checked our modules, looks like control-plane and cli use 1.16, and acceptance uses 1.14. It should probably be a separate PR to bump them all. I would want to keep them in sync as much as possible too since I think we use specific go versions/images in our CI images

charts/embed_chart.go
ndhanushkodi: Yup!

cli/cmd/common/utils.go
```diff
@@ -1,20 +1,81 @@
...
+// is a format that the Helm Go SDK functions can read from to create a chart to install from. The names of these files
+// are important, as there are case statements in the Helm Go SDK looking for files named "Chart.yaml" or
+// "templates/<templatename>.yaml", which is why even though the embedded FS has them named "consul/Chart.yaml" we have
+// to strip the "consul" prefix out.
+func ReadChartFiles(chart embed.FS, chartDirName string) ([]*loader.BufferedFile, error) {
+	var chartFiles []*loader.BufferedFile
+
+	bytes, err := chart.ReadFile(fmt.Sprintf("%s/%s", chartDirName, chartFileName))
+	if err != nil {
+		return []*loader.BufferedFile{}, err
```
t-eckert, line 34: Here and below can you just return the empty `chartFiles` you instantiated at the top of the file?

cli/cmd/common/utils.go
```diff
@@ -1,20 +1,81 @@
...
+	bytes, err = chart.ReadFile(fmt.Sprintf("%s/%s", chartDirName, valuesFileName))
+	if err != nil {
+		return []*loader.BufferedFile{}, err
+	}
+	chartFiles = append(chartFiles,
+		&loader.BufferedFile{
+			Name: valuesFileName,
+			Data: bytes,
+		},
+	)
```
t-eckert, line 52: You could wrap this behavior in a separate function to reduce code duplication.

```go
func readChart(chartDirName, fileName) (*chartFile, error) {
        bytes, err = chart.ReadFile(fmt.Sprintf("%s/%s", chartDirName, fileName))
	if err != nil {
		return nil, err
	}
	return &loader.BufferedFile{
		Name: fileName,
		Data: bytes,
	}
}
```

cli/cmd/common/utils.go
ndhanushkodi: In L45 if we return `chartFiles`, it would have one file in it ("Chart.yaml"), and I don't think we should return an incomplete object in an error case

cli/cmd/common/utils.go
```diff
@@ -1,20 +1,81 @@
...
+				return []*loader.BufferedFile{}, err
+			}
+			chartFiles = append(chartFiles,
+				&loader.BufferedFile{
+					Name: fmt.Sprintf("%s/%s", templatesDirName, dir.Name()),
+					Data: bytes,
+				},
+			)
+		}
+	}
```
t-eckert, line 74: I also feel like this should be pulled out to a separate function. (lines 58 thru 74)

cli/cmd/common/utils_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed fixtures/consul/* fixtures/consul/templates/_helpers.tpl
+var testChart embed.FS
+
+func TestReadChartFiles(t *testing.T) {
+	files, err := ReadChartFiles(testChart, "fixtures/consul")
+	require.NoError(t, err)
+	foundChart := false
```
t-eckert, line 16: Kind of silly to suggest stylistic changes in tests, but just in case you want to inline this, they should all default to false.

```suggestion
	var foundChart, foundValues,  foundTemplate, foundHelper bool
```

t-eckert (approved):
This looks great! I had some stylistic nits, but nothing major. You are doing great with this!

charts/go.mod
t-eckert: Cool

cli/cmd/common/utils.go
t-eckert: Good call

cli/cmd/common/utils_test.go
t-eckert: They will all default to `false` confirmed: https://play.golang.org/p/K12S2kvhbks

cli/cmd/common/utils.go
lkysow: chiming in here: imo if you're return a non-nil error then all other return values should be considered invalid and so it's okay to return whatever you want as the other values–usually what the function writer finds easier.

I can't actually find this idiom defined anywhere though :)

So for this I'd probably do `return nil, err` because it's easier to write (and also in this case it's functionally the same as `[]*loader.BufferedFile{}`.

cli/cmd/common/utils.go
```diff
@@ -1,20 +1,81 @@
...
 
+// ReadChartFiles reads the chart files from the embedded file system, and loads their contents into
+// []*loader.BufferedFile. This is a format that the Helm Go SDK functions can read from to create a chart to install
+// from. The names of these files are important, as there are case statements in the Helm Go SDK looking for files named
+// "Chart.yaml" or "templates/<templatename>.yaml", which is why even though the embedded file system has them named
+// "consul/Chart.yaml" we have to strip the "consul" prefix out.
+func ReadChartFiles(chart embed.FS, chartDirName string) ([]*loader.BufferedFile, error) {
+	var chartFiles []*loader.BufferedFile
+
+	bytes, err := chart.ReadFile(fmt.Sprintf("%s/%s", chartDirName, chartFileName))
```
lkysow, line 32: nit: we can use `filepath.Join` for these kinds of expressions (`filepath.Join(chartDirName, chartFileName)`)

cli/cmd/common/utils.go
lkysow: I took a stab at a fun refactor using Thomas's suggestion and using `filepath.Rel` to do the removal of `charts/`:


```go
func ReadChartFiles(chart embed.FS, chartDirName string) ([]*loader.BufferedFile, error) {
	var chartFiles []*loader.BufferedFile

	// Load Chart.yaml and values.yaml first.
	for _, f := range []string{chartFileName, valuesFileName} {
		buffered, err := readFile(chart, filepath.Join(chartDirName, f), chartDirName)
		if err != nil {
			return nil, err
		}
		chartFiles = append(chartFiles, buffered)
	}

	// Now load everything under templates/.
	dirs, err := chart.ReadDir(filepath.Join(chartDirName, templatesDirName))
	if err != nil {
		return nil, err
	}
	for _, dir := range dirs {
		if dir.IsDir() {
			// We only need to package up files.
			continue
		}

		buffered, err := readFile(chart, filepath.Join(chartDirName, templatesDirName, dir.Name()), chartDirName)
		if err != nil {
			return nil, err
		}
		chartFiles = append(chartFiles, buffered)
	}

	return chartFiles, nil
}

func readFile(chart embed.FS, f string, basePath string) (*loader.BufferedFile, error) {
	bytes, err := chart.ReadFile(f)
	if err != nil {
		return nil, err
	}
	rel, err := filepath.Rel(basePath, f)
	if err != nil {
		return nil, err
	}
	return &loader.BufferedFile{
		Name: rel,
		Data: bytes,
	}, nil
}

```

lkysow (commented):
Nice work!

For future selves: Nitya and I discussed instead of doing this, pinning the chart version that helm downloads using `version/version.go`. This would reduce the complexity here but this work has the nice effect that there are no external calls to Helm/repositories and that everything is packaged up in the compiled binary so based on that we decided to continue with this approach.

PR #781, opened 2021-10-14
cli: bump golang.org/x/sys dependency

ndhanushkodi:
See https://github.com/golang/go/issues/46763. The CLI would error at
runtime after being compiled with go 1.17, unless the x/sys dependency
is bumped.

How I've tested this PR:
- upgraded to go 1.17.2
- Ran `go build . -o consul-k8s` 
- `./consul-k8s version`
- build and run the cli like above with go 1.17.x and check that it works

Checklist:
- [ ] CHANGELOG entry added 


ndhanushkodi:
@kschoche yes the go mod should still specify 1.16 but if people do build with 1.17 we should see it work!

PR #787, opened 2021-10-16
CONTRIBUTING.md: adding TOC

david-yu:
Adding a TOC since it was getting harder to read

Changes proposed in this PR:
- Added a TOC
- Added Line Breaks

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,30 @@
-# Contributing
+# Contributing to Consul on Kubernetes
+
+# Table of Contents
+1. [Contributing 101](#contributing-101)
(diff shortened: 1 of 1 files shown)


t-eckert:
This is a great idea! I had been thinking about it too. One thought to consider is that GitHub now automatically does this for all markdown files. It may add overhead to maintain the TOC. However, given that many people don't know about the feature in the first place or may not be reading the repo on GitHub, there is a definite benefit.

Screenshot of the GitHub automatic TOC feature.

PR #797, opened 2021-10-19
Set extra volumes and env vars for sync catalog deployment

cristianoalexandre:
Changes proposed in this PR:
- Add the possibility to specify extra volumes and extra environment variables to sync-catalog pods.

How I've tested this PR:
- Unit tests (`sync-catalog-deployment.bats`)

- Run the `sync-catalog-deployment` unit tests

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/sync-catalog-deployment.yaml
+++ b/charts/consul/templates/sync-catalog-deployment.yaml
@@ -33,8 +33,9 @@ spec:
         "consul.hashicorp.com/connect-inject": "false"
     spec:
       serviceAccountName: {{ template "consul.fullname" . }}-sync-catalog
-      {{- if .Values.global.tls.enabled }}
+      {{- if or .Values.global.tls.enabled .Values.syncCatalog.extraVolumes }}
       volumes:
+      {{- if .Values.global.tls.enabled }}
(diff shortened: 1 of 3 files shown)


cristianoalexandre:
Hi @ishustava !

Small initial disclaimer: I am not a K8s expert by any means, and most of my findings are based on what I could investigate so far, so please feel free to disagree with me or correct me. :)

My concern is not about any vulnerability in special - it's just that, from what I was able to investigate prior to opening this MR, by being directly in an env var, a secret may be accidentally exposed: for instance, while some tools treat these kind of env vars pointing to secrets in a "special" manner and not directly showing them (`kubectl` being one of them), many others just "dump" the whole environment as-is, which would leak the values.

To be fair, from what I also was able to investigate, having the secret in plain-text inside a file also has its risks and is prone to attack, but I do believe the risk of accidental leak (not due to an attack/exploitation) may be lessened using this approach.

ishustava:
Got it, thanks for this context. 

Consul and consul-k8s will not dump these and there are no other tools that are installed in the consul's or sync catalog's pod. 

Yeah I agree, having them in files doesn't necessarily mitigate the problem if some tool would just print out a secret to logs. Since this isn't the case with Consul or consul-k8s, would you feel comfortable using it as is?

PR #811, opened 2021-10-28
Bootstrap gossip encryption with Vault

kschoche:
Changes proposed in this PR:
- remove the extra logic in `vault_cluster` for waiting for pods as this is no longer needed now that we're using vault dev-mode (thx @ndhanushkodi) 
- Add a new Stanza in `values.yaml` :  `global.secretsBackend`  which holds the Roles and has some light documentation on its usage (to be improved yet)
- Add logic to bootstrap gossip encryption key to the server-statefulset+client-daemonset with Vault annotations.
- Add unit and acceptance tests to validate above logic.

How I've tested this PR:
1. Acceptance tests are added which bootstrap consul using a gossip encryption key
2. bats tests

Code Review + run the acceptance tests/bats tests.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


acceptance/framework/vault/vault_cluster.go
```diff
@@ -179,15 +176,8 @@ func (v *VaultCluster) Create(t *testing.T, ctx environment.TestContext) {
...
-	helpers.WaitForAllPodsToBeReady(t, v.kubernetesClient, v.vaultHelmOptions.KubectlOptions.Namespace, "app.kubernetes.io/name=vault-agent-injector")
-	// Wait for the server pod to be PodRunning, it will not be Ready because it has not been Init+Unseal'd yet.
-	// The vault server has health checks bound to unseal status, and Unseal is done as part of bootstrap (below).
-	retry.RunWith(&retry.Counter{Wait: 1 * time.Second, Count: 30}, t, func(r *retry.R) {
-		pod, err := v.kubernetesClient.CoreV1().Pods(v.vaultHelmOptions.KubectlOptions.Namespace).Get(context.Background(), fmt.Sprintf("%s-vault-0", v.vaultReleaseName), metav1.GetOptions{})
-		require.NoError(r, err)
-		require.Equal(r, pod.Status.Phase, corev1.PodRunning)
-	})
+	// Wait for the injector and vault server pods to become Ready.
+	helpers.WaitForAllPodsToBeReady(t, v.kubernetesClient, v.vaultHelmOptions.KubectlOptions.Namespace, fmt.Sprintf("helm.sh/chart=%s", v.vaultChartName))
```
kschoche, line 180: After switching to vault dev-mode the vault-server now becomes Ready right away.

charts/consul/test/unit/gossip-encryption-autogenerate-job.bats
```diff
@@ -38,26 +38,3 @@ load _helpers
   [ "$status" -eq 1 ]
   [[ "$output" =~ "If global.gossipEncryption.autoGenerate is true, global.gossipEncryption.secretName and global.gossipEncryption.secretKey must not be set." ]]
 }
-
-@test "gossipEncryptionAutogenerate/Job: fails if global.gossipEncryption.autoGenerate=true and global.gossipEncryption.secretName is set" {
```
kschoche, line 42: This test is now invalid due to requiring both secretKey + Name to be specified in server-statefulset

t-eckert (approved):
Thank you Kyle, very cool.

acceptance/tests/vault/vault_test.go
```diff
@@ -39,5 +44,90 @@ func TestVault_Create(t *testing.T) {
...
+	logger.Log(t, "Installing Consul")
+	consulCluster := consul.NewHelmCluster(t, consulHelmValues, ctx, cfg, consulReleaseName)
+	consulCluster.Create(t)
+
+	// Validate that the gossip encryption key is set correctly.
+	logger.Log(t, "Validating the gossip key has been set correctly.")
+	consulClient := consulCluster.SetupConsulClient(t, true)
+	keys, err := consulClient.Operator().KeyringList(nil)
+	require.NoError(t, err)
+	// we use keys[0] because KeyringList returns a list of keyrings for each dc, in this case there is only 1 dc.
```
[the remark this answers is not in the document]
kschoche: I'm not sure to be honest, but I switched it up to assert that it's equal to 1!

acceptance/tests/vault/vault_test.go
```diff
@@ -39,5 +44,89 @@ func TestVault_Create(t *testing.T) {
...
+
+	params = map[string]interface{}{
+		"bound_service_account_names":      consulServerServiceAccountName,
+		"bound_service_account_namespaces": "default",
+		"policies":                         "consul-gossip",
+		"ttl":                              "24h",
+	}
+	_, err = vaultClient.Logical().Write("auth/kubernetes/role/consul-server", params)
+	require.NoError(t, err)
+
```
kschoche, line 95: I will probably dry this up in the next PR for serverTLS as it will need to also create Policies and setup Roles

acceptance/framework/vault/vault_cluster.go
```diff
@@ -74,7 +71,14 @@ func NewVaultCluster(
 	helm.AddRepo(t, &helm.Options{}, "hashicorp/vault", "https://helm.releases.hashicorp.com")
 	// Ignoring the error from `helm repo update` as it could fail due to stale cache or unreachable servers and we're
 	// asserting a chart version on Install which would fail in an obvious way should this not succeed.
-	_, _ = helm.RunHelmCommandAndGetOutputE(t, &helm.Options{}, "repo", "update")
+	errStr, err := helm.RunHelmCommandAndGetOutputE(t, &helm.Options{}, "repo", "update")
```
kschoche, line 74: Looks like this step is super flakey:
```
=== RUN   TestVault_Create
TestVault_Create 2021-11-01T20:32:31Z logger.go:66: Running command helm with args [repo add hashicorp/vault https://helm.releases.hashicorp.com]
TestVault_Create 2021-11-01T20:32:31Z logger.go:66: "hashicorp/vault" has been added to your repositories
TestVault_Create 2021-11-01T20:32:31Z logger.go:66: Running command helm with args [repo update]
TestVault_Create 2021-11-01T20:32:31Z logger.go:66: Hang tight while we grab the latest from your chart repositories...
TestVault_Create 2021-11-01T20:32:31Z logger.go:66: ...Successfully got an update from the "hashicorp/vault" chart repository
TestVault_Create 2021-11-01T20:32:31Z logger.go:66: Update Complete. ⎈Happy Helming!⎈
TestVault_Create 2021-11-01T20:32:31Z logger.go:66: Running command helm with args [pull hashicorp/vault --version 0.17.0]
TestVault_Create 2021-11-01T20:32:31Z logger.go:66: Error: repo hashicorp not found
```

charts/consul/test/unit/gossip-encryption-autogenerate-job.bats
ishustava: why are they invalid? I don't think it's the same behavior. The server statefulset will fail if you've provided only one of secretName/secretKey. These tests are checking that you haven't set them when autoGenerate is true.

charts/consul/values.yaml
```diff
@@ -117,25 +117,44 @@ global:
   # created by this chart. See https://kubernetes.io/docs/concepts/policy/pod-security-policy/.
   enablePodSecurityPolicies: false
 
-  # Configures Consul's gossip encryption key, set as a Kubernetes secret
+  # secretsBackend is used to enable and disable the use of Vault as the secrets backend for the Consul on Kubernetes installation.
```
ishustava, line 120:
```suggestion
  # secretsBackend is used to configure Vault as the secrets backend for the Consul on Kubernetes installation.
```

charts/consul/values.yaml
```diff
@@ -117,25 +117,44 @@ global:
...
+  # In order to utilize this configuration it is required to have an accessible vault cluster which has been pre-configured
+  # by a Vault operator to have the Vault Kube Auth Method, KV2 + PKI secrets engines enabled as well as pre-created
+  # vault secrets, policies and roles.
```
ishustava, line 123:
```suggestion
  # The Vault cluster needs to have the Kubernetes Auth Method, 
  # KV2 and PKI secrets engines enabled and have necessary secrets, 
  # policies and roles created prior to installing Consul.
  # The Vault cluster should not have Consul as its storage backend.
```

charts/consul/values.yaml
```diff
@@ -117,25 +117,44 @@ global:
...
+  secretsBackend:
+    vault:
+      # Enabling the Vault secrets backend will cause Kubernetes secrets to no longer be used for the following Consul secrets:
+      # - gossip encryption key defined by `global.gossipEncryption.secretName/Key`
```
ishustava, line 127: Could you make this description more generic? I think this will a bit be hard to maintain going forward if for every secret we're enabling for vault, we'd also need to update this description

charts/consul/values.yaml
```diff
@@ -117,25 +117,44 @@ global:
...
+      enabled: false
+
+      # The consulServerRole must be created and have read capabilties via a Policy for the following secrets:
+      # - gossip encryption key defined by `global.gossipEncryption.secretName/Key`.
+      consulServerRole: ""  # e.g. "consul-server"
```
ishustava, line 132: could you move this example to the description block so that it'll be included in our helm ref docs?

charts/consul/values.yaml
```diff
@@ -117,25 +117,44 @@ global:
...
   gossipEncryption:
     # Automatically generate a gossip encryption key and save it to a Kubernetes secret.
     autoGenerate: false
-    # secretName is the name of the Kubernetes secret that holds the gossip
-    # encryption key. The secret must be in the same namespace that Consul is installed into.
+    # secretName is the name of the Kubernetes secret or Vault secret path that holds the gossip
+    # encryption key. The Kubernetes secret must be in the same namespace that Consul is installed into.
     secretName: ""
-    # secretKey is the key within the Kubernetes secret that holds the gossip
+    # secretKey is the key within the Kubernetes secret or Vault secret that holds the gossip
```
ishustava, line 157:
```suggestion
    # secretKey is the key within the Kubernetes secret or Vault secret key that holds the gossip
```

acceptance/framework/vault/vault_cluster.go
```diff
@@ -74,7 +71,14 @@ func NewVaultCluster(
...
+	if err != nil {
+		logger.Logf(t, "Unable to update helm repository: %s, %s", err, errStr)
+	}
+	errStr, err = helm.RunHelmCommandAndGetOutputE(t, &helm.Options{}, "pull", "hashicorp/vault", "--version", vaultChartVersion)
```
ishustava, line 78: should we always install latest? I think otherwise we'll have to remember to periodically update to the latest vault version.

acceptance/framework/vault/vault_cluster.go
```diff
@@ -74,7 +71,14 @@ func NewVaultCluster(
...
+	if err != nil {
+		logger.Logf(t, "Unable to pull helm repository: %s, %s", err, errStr)
```
ishustava, line 80: I think we need to `require.NoError(t, err)` instead of logging the error and continuing. Also, the same applies to above.

ishustava (commented):
Great work on this, Kyle! I'm still reviewing and finishing up some testing but wanted to leave the comments I have so far. Most of them are small edits, but there are a couple of questions inline as well, plus two comments below.

1. When I ran `TestVault_BootstrapConsulGossipEncryptionKey` test locally, it did not pass with vault-agent-init for the server erroring out with `2021-11-04T22:33:49.427Z [ERROR] auth.handler: error authenticating: error="context deadline exceeded" backoff=9.8s` and because of it the server-acl-init job could never finish.
2. I also have a larger UX question. I find it a bit counter-intuitive the way you need to set `secretKey` for a secret in vault that makes it inconsistent with how we do it in k8s. 
 
   For example, in Kubernetes when I create a secret with `kubectl create secret generic foo --from-literal=key=bar`, I'd then set `secretName` to `foo` and `secretKey` to `key`. The same thing is quite different in Vault. If I create a similar secret in vault with `vault kv put internal/foo key="bar"`, then I would need to set `secretKey` to `.Data.data.key` which is kind of confusing. I know this is how vault injector works under the hood, but I wonder if we should make the UX consistent since the fact that we're using vault injector is an implementation detail and the user should not care about it.

charts/consul/templates/client-daemonset.yaml
```diff
@@ -37,6 +37,16 @@ spec:
           {{- toYaml .Values.client.extraLabels | nindent 8 }}
         {{- end }}
       annotations:
+        {{- if .Values.global.secretsBackend.vault.enabled }}
+        "vault.hashicorp.com/agent-inject": "true"
+        "vault.hashicorp.com/agent-init-first": "true"
```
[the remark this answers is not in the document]
kschoche: It makes the vault init-container the first to run, ahead of things like tls/acl jobs. Maybe it's not needed? It is probably not necessary for gossip but will be for server-tls.

charts/consul/templates/client-daemonset.yaml
```diff
@@ -37,6 +37,16 @@ spec:
...
+        "vault.hashicorp.com/role": "{{ .Values.global.secretsBackend.vault.consulClientRole }}"
+        {{- with .Values.global.gossipEncryption }}
+        "vault.hashicorp.com/agent-inject-secret-gossip.txt": "{{ .secretName }}"
+        "vault.hashicorp.com/agent-inject-template-gossip.txt": '{{ "{{" }}- with secret "{{ .secretName }}" -{{ "}}" }} {{ "{{" }} {{ .secretKey }} {{ "}}" }} {{ "{{" }}- end -{{ "}}" }}'
```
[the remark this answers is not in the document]
kschoche: I *wish* this works, will try it out, I spent days on this in the past and maybe it will work now :D

charts/consul/templates/client-daemonset.yaml
```diff
@@ -37,6 +37,16 @@ spec:
...
+        "vault.hashicorp.com/template-static-secret-render-interval": "10s"
```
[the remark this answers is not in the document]
kschoche: Good catch, leftover code from a demo, thanks!

charts/consul/templates/client-daemonset.yaml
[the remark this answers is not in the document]
ishustava: Why is it needed for tls?

charts/consul/templates/client-daemonset.yaml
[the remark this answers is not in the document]
ishustava: FWIW I tested it and it works.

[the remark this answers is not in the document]
kschoche: nice! that is good news, these templates are a nightmare, thx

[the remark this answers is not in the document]
kschoche: Cool I think I got it to work, and also I'm going to move it into a helper function since it's used by client-daemonset and server-statefulset.

charts/consul/templates/client-daemonset.yaml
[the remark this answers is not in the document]
kschoche: You're totally right, idk what I was thinking, this is the wrong template anyway! :D

kschoche:
> Great work on this, Kyle! I'm still reviewing and finishing up some testing but wanted to leave the comments I have so far. Most of them are small edits, but there are a couple of questions inline as well, plus two comments below.
> 
> 1. When I ran `TestVault_BootstrapConsulGossipEncryptionKey` test locally, it did not pass with vault-agent-init for the server erroring out with `2021-11-04T22:33:49.427Z [ERROR] auth.handler: error authenticating: error="context deadline exceeded" backoff=9.8s` and because of it the server-acl-init job could never finish.
> 2. I also have a larger UX question. I find it a bit counter-intuitive the way you need to set `secretKey` for a secret in vault that makes it inconsistent with how we do it in k8s.
>    For example, in Kubernetes when I create a secret with `kubectl create secret generic foo --from-literal=key=bar`, I'd then set `secretName` to `foo` and `secretKey` to `key`. The same thing is quite different in Vault. If I create a similar secret in vault with `vault kv put internal/foo key="bar"`, then I would need to set `secretKey` to `.Data.data.key` which is kind of confusing. I know this is how vault injector works under the hood, but I wonder if we should make the UX consistent since the fact that we're using vault injector is an implementation detail and the user should not care about it.

Hi @ishustava thanks for the awesome/helpful review!
Your question number 1 was a bug in Vault that is fixed by modifying the issuer field of the kube auth config for vault. Once we start running against Vault 1.9 this wont be needed.
For your second question, I've updated the values.yaml file to include language which describes how the user will interact with consul and vault, in doing so I also removed the need for any vault api-specific wording in the key fields so now the user just specifies "key" and not ".Data.data.key" as this is implicit. I was not able to figure out a way to remove the implicit "data" field of the secret path, but I think this is less of an issue as the vault reference guides and docs mention that you need to add "data/" after your kv2 mount point when reading. Thoughts?

charts/consul/values.yaml
```diff
@@ -117,25 +117,56 @@ global:
...
+  # Note: When using Vault KV2 secrets engines the "data" field is implicitly required for Vault API calls,
+  # secretName should be in the form of  "vault-kv2-mount-path/data/secret-name".
+  # secretKey should be in the form of "key".
+  secretsBackend:
+    vault:
+      # Enabling the Vault secrets backend will replace Kubernetes secrets with referenced Vault secrets.
+      enabled: false
+
+      # The consulServerRole must be created and have read capabilties via a Policy for the following secrets:
+      # - gossip encryption key defined by `global.gossipEncryption.secretName`.
```
ishustava, line 134: Adding a bit more detail. If you like it, it'd be nice to do the same for the client role.
```suggestion
      # The Vault role for the Consul server.
      # The role must be connected to the Consul server's service account and
      # have have a policy with read capabilities for the following secrets:
      # - gossip encryption key defined by `global.gossipEncryption.secretName`.
      # To discover the service account name of the Consul server, run
      #     ```
      #     helm template -s templates/server-serviceaccount.yaml <release-name> charts/consul
      #     ```
      # and check the name of `metadata.name`.
```

charts/consul/values.yaml
```diff
@@ -117,25 +117,56 @@ global:
...
+  # Configures Consul's gossip encryption key.
   # (see `-encrypt` (https://consul.io/docs/agent/options#_encrypt)).
   # By default, gossip encryption is not enabled. The gossip encryption key may be set automatically or manually.
   # The recommended method is to automatically generate the key.
   # To automatically generate and set a gossip encryption key, set autoGenerate to true.
   # Values for secretName and secretKey should not be set if autoGenerate is true.
   # To manually generate a gossip encryption key, set secretName and secretKey and use Consul to generate
-  # a Kubernetes secret referencing these values.
+  # a key, saving this as a Kubernetes secret or Vault secret path and key.
+  # If `global.secretsBackend.vault.enabled=true` be sure to add the "data" field to the secretName as required by
```
ishustava, line 149:
```suggestion
  # If `global.secretsBackend.vault.enabled=true`, be sure to add the "data" field to the secretName as required by
```

ishustava (approved):
Good work, Kyle! I think everything looks good, I've added some minor edits.

My only other concern is about the issuer field for the vault auth method (see comment inline). I don't think it's blocking since for now it's passing on kind, but I think we should fix it soon because it'll cause problems when we merge the feature branch into `main` and run the nightly tests.

PR #827, opened 2021-11-02
update-changelog

thisisnotashwin:
Changes proposed in this PR:
- update readme

Checklist:
- [x] CHANGELOG entry added 


ndhanushkodi (approved):

t-eckert (approved):
Solid! When I'm looking at the CHANGELOG for final release, how much do we split hairs on "improvement" vs "feature"?

thisisnotashwin:
> Solid! When I'm looking at the CHANGELOG for final release, how much do we split hairs on "improvement" vs "feature"?

Not too much personally. If it is entirely new, then the feature column, but if it seems like it extends and existing feature, then improvement. But it isnt't a hill I'd die on.

PR #829, opened 2021-11-02
🚀🚀🚀 Release 0.36.0

t-eckert:
Let's release 0.36.0

Changes proposed in this PR:
- Update `imageK8S` in `values.yaml` to `0.36.0`
- Update `image` in `values.yaml` to `1.11.0`

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -105,7 +105,7 @@ global:
   # image that is used for functionality such as catalog sync.
   # This can be overridden per component.
   # @default: hashicorp/consul-k8s-control-plane:<latest version>
-  imageK8S: "hashicorp/consul-k8s-control-plane:0.35.1"
+  imageK8S: "hashicorp/consul-k8s-control-plane:0.36.0"
 
   # The name of the datacenter that the agents should
   # register as. This can't be changed once the Consul cluster is up and running
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -105,7 +105,7 @@ global:
   # image that is used for functionality such as catalog sync.
   # This can be overridden per component.
   # @default: hashicorp/consul-k8s-control-plane:<latest version>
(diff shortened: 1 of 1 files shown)


charts/consul/values.yaml
```diff
@@ -85,7 +85,7 @@ global:
   # image: "hashicorp/consul-enterprise:1.10.0-ent"
   # ```
   # @default: hashicorp/consul:<latest version>
-  image: "hashicorp/consul:1.10.3"
+  image: "hashicorp/consul:1.11.0"
```
ishustava, line 88: this should be the latest 1.10.x since there's no 1.11 GA yet

charts/consul/values.yaml
```diff
@@ -105,7 +105,7 @@ global:
   # image that is used for functionality such as catalog sync.
   # This can be overridden per component.
   # @default: hashicorp/consul-k8s-control-plane:<latest version>
-  imageK8S: "hashicorp/consul-k8s-control-plane:0.35.0"
+  imageK8S: "hashicorp/consul-k8s-control-plane:0.35.1"
```
ishustava: This should be 0.36. I think there are too many improvements in the changelog for a patch release.

ishustava (commented):
We also need to update chart.yaml

charts/consul/values.yaml
thisisnotashwin, line 88: This should still be consul 1.10.3
Consul 1.11.0 will only be available as a beta and users will have to explicitly update the values file to use the beta2 image.

t-eckert:
Closing this as I mistakenly thought the next version would be `0.35.1`. Version is actually `0.36.0` and I will open up a new branch/PR for that.

charts/consul/values.yaml
t-eckert: Cool, this was a bad assumption by me. I had mapped `0.35.0` to beta 1 so I figured `0.35.1` would be beta 1.

PR #833, opened 2021-11-03
Automate DNS redirection to Consul DNS

ndhanushkodi:
Changes proposed in this PR:
- When DNS is enabled, get the DNS Service IP and redirect DNS requests to the Consul DNS service.
- Add kubedns as a recursor for Consul DNS

Corresponding consul PR: https://github.com/hashicorp/consul/pull/11480

How I've tested this PR:
```bash
helm install -n consul consul ../charts/consul --set connectInject.enabled=true --set global.image=ashwinvenkatesh/consul@sha256:d02bf8201499a0a6be82fdf279bb1c5f9d2bc1eda2c5fac47ca832c879444292 --set global.imageK8S=ashwinvenkatesh/consul-k8s@sha256:ce817f69f109e7dcb27ba01f1cad29fcfedeff5ed424db21fa095e0e3d582fe9 --set dns.enableRedirection=true --set controller.enabled=true
```

```bash
k apply -f demo.yaml #included in this PR for testing, remove before merging
```

Then, exec onto the frontend pod, and
```bash
curl backend
```
```bash
nslookup backend.service.consul
```
and verify it returns a pod ip for backend. We can't curl `backend.service.consul` because envoy only knows to direct traffic based on VIPs not pod ips

- Code Review
- Re-try above steps

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/client-dns-clusterrole.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
```
ndhanushkodi, line 1: This is not currently necessary, this would only be necessary if we decided to use the kube api to get the kubedns service ip. Left it here in case we decide to go that route. The API request is also in a prev commit in this PR.

charts/consul/templates/client-dns-clusterrolebinding.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
```
ndhanushkodi, line 1: This file is not currently necessary, this would only be necessary if we decided to use the kube api to get the kubedns service ip. Left it here in case we decide to go that route. The API request is also in a prev commit in this PR.

charts/consul/templates/client-daemonset.yaml
```diff
@@ -202,6 +202,14 @@ spec:
             - |
               CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
+              {{- if .Values.dns.enabled }}
+              export kube_dns_service_ip=$(cat /etc/resolv.conf | grep nameserver | cut -d' ' -f2)
+              echo $kube_dns_service_ip > /consul/kubednsip
```
ndhanushkodi, line 207: echo was just for debugging

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -83,6 +83,13 @@ spec:
             - |
               CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
+              {{- if .Values.dns.enabled }}
+              consul_fullname_with_underscore=$(echo $CONSUL_FULLNAME | tr '-' '_')
+              dns_var_name=${consul_fullname_with_underscore}_dns_service_host
+              dns_var_name_in_caps=$(echo $dns_var_name | tr 'a-z' 'A-Z')
+              export dns_ip=$(eval echo \$${dns_var_name_in_caps})
```
ndhanushkodi, line 90: here, we are essentially templating the name of the environment variable that will have the Consul DNS IP. If `CONSUL_FULLNAME` is `consul-consul`, then the env var would be `CONSUL_CONSUL_DNS_SERVICE_HOST`, so we need to turn dashes into underscores, then capitalize all the letters. Then to dereference the variable once we've come up with the name, we use the `eval echo \$$` bit to do that

demo.yaml
```diff
@@ -0,0 +1,136 @@
...
+  selector:
+    matchLabels:
+      app: frontend
+  template:
+    metadata:
+      labels:
+        app: frontend
+      annotations:
+        "consul.hashicorp.com/connect-inject": "true"
+        "consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs": "151.101.2.133,151.101.66.133,151.101.130.133,151.101.194.133"
```
ndhanushkodi, line 38: will let us `apk update` && `apk add` stuff

charts/consul/templates/connect-inject-deployment.yaml
ndhanushkodi: in discussion, we think we could move this to inject-connect

charts/consul/templates/client-daemonset.yaml
```diff
@@ -202,6 +202,13 @@ spec:
             - |
               CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
+              {{- if (and .Values.dns.enabled .Values.dns.enableRedirection)}}
+              export kube_dns_service_ip=$(cat /etc/resolv.conf | grep nameserver | cut -d' ' -f2)
```
ishustava, line 206: should we rename this to be `nameserver_ip`?

charts/consul/templates/server-statefulset.yaml
```diff
@@ -192,6 +192,13 @@ spec:
             - |
               CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
+              {{- if (and .Values.dns.enabled .Values.dns.enableRedirection)}}
+              export kube_dns_service_ip=$(cat /etc/resolv.conf | grep nameserver | cut -d' ' -f2)
```
ishustava, line 196: we don't need export here since it's only used within this shell session.
```suggestion
              kube_dns_service_ip=$(cat /etc/resolv.conf | grep nameserver | cut -d' ' -f2)
```

charts/consul/templates/server-statefulset.yaml
```diff
@@ -254,6 +261,9 @@ spec:
                 {{- range $value := .Values.global.recursors }}
                 -recursor={{ quote $value }} \
                 {{- end }}
+                {{- if (and .Values.dns.enabled .Values.dns.enableRedirection)}}
+                -recursor="$kube_dns_service_ip" \
```
ishustava, line 265: this variable could contain multiple nameservers, so i think we need to iterate 😬

charts/consul/templates/server-statefulset.yaml
```diff
@@ -192,6 +192,13 @@ spec:
             - |
               CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
+              {{- if (and .Values.dns.enabled .Values.dns.enableRedirection)}}
+              export kube_dns_service_ip=$(cat /etc/resolv.conf | grep nameserver | cut -d' ' -f2)
+              if [[ -z $kube_dns_service_ip ]]; then
+                  echo "Unable to get kube-dns Service IP"
+              fi
```
ishustava, line 199: 🤔 It's a bit strange that we just echo here and don't do anything. 

I'm also thinking that it'll probably be ok to assume the nameserver(s) will always be there on k8s.

charts/consul/test/unit/connect-inject-deployment.bats
```diff
@@ -519,6 +519,30 @@ EOF
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/connect-inject-deployment.yaml \
+      --set 'connectInject.enabled=true' \
+      . | tee /dev/stderr |
+      yq -c -r '.spec.template.spec.containers[0].command | join(" ") | contains("-enable-consul-dns=true")' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "connectInject/Deployment: -enable-consul-dns is false if dns.enabled=false is disabled" {
```
ishustava, line 535:
```suggestion
@test "connectInject/Deployment: -enable-consul-dns is true if dns.enabled=true and dns.enableRedirection=true" {
```

charts/consul/templates/client-daemonset.yaml
ishustava: Also, should we only do that if servers are disabled?

charts/consul/values.yaml
```diff
@@ -1122,6 +1122,9 @@ dns:
   # @type: boolean
   enabled: "-"
 
+  # @type: boolean
```
ishustava, line 1125: needs docs

demo.yaml
```diff
@@ -0,0 +1,136 @@
+# Service to expose frontend
```
ishustava, line 1: we prob don't want to keep this

ishustava (commented):
great work Ashwin and Nitya! I've left some comments/questions inline.

charts/consul/templates/client-daemonset.yaml
thisisnotashwin: The dns-service points to both servers as well as clients (which I was surprised by but the label selector matches on clients and servers) so we might need to add a recursor here as well, unless I'm misunderstanding something

ishustava: Hmm i don't think it does atm. It will select only the pods that have the `hasDNS: "true"` label and clients actually don't have this label right now. I think we'll need to add it if servers are disabled, otherwise it should only point at servers I think.

thisisnotashwin: That's what I thought too: https://github.com/hashicorp/consul-k8s/blob/03d4b932256b68ba9fdccba4c00951bb3de275be/charts/consul/templates/client-daemonset.yaml#L35

ishustava: Oh you're right!

charts/consul/templates/_helpers.tpl
```diff
@@ -28,6 +28,18 @@ is passed to consul as a -config-file param on command line.
...
+{{/*
+Sets up a list of recusor flags for Consul agents by iterating over the IPs of every nameserver
+in /etc/resolv.conf and concatenating them into a string of arguments that can be passed directly
+to the consul agent command.
+*/}}
+{{- define "consul.recursors" -}}
+              recursor_flags=""
+              for ip in $(cat /etc/resolv.conf | grep nameserver | cut -d' ' -f2)
+              do
+                 recursor_flags=$recursor_flags" -recursor=$ip"
```
lkysow, line 40:
```suggestion
                 recursor_flags="$recursor_flags -recursor=$ip"
```

I think this is the same? Looks safer to me

charts/consul/templates/client-daemonset.yaml
```diff
@@ -202,6 +202,10 @@ spec:
             - |
               CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
+              {{- if (and .Values.dns.enabled .Values.dns.enableRedirection)}}
```
lkysow, line 205:
```suggestion
              {{- if (and .Values.dns.enabled .Values.dns.enableRedirection) }}
```

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -107,6 +107,9 @@ spec:
                 {{- else }}
                 -transparent-proxy-default-overwrite-probes=false \
                 {{- end }}
+                {{- if (and .Values.dns.enabled .Values.dns.enableRedirection)}}
```
lkysow, line 110:
```suggestion
                {{- if (and .Values.dns.enabled .Values.dns.enableRedirection) }}
```

lkysow: and below

control-plane/connect-inject/container_init.go
```diff
@@ -16,6 +17,7 @@ const (
 	envoyUserAndGroupID         = 5995
 	copyContainerUserAndGroupID = 5996
 	netAdminCapability          = "NET_ADMIN"
+	dnsServiceHostEnvSuffix     = "DNS_SERVICE_HOST"
```
lkysow, line 20: What if there's a service like `foo-dns`? I think we need to use the Consul DNS service name exactly.

control-plane/connect-inject/container_init.go
```diff
@@ -107,6 +112,21 @@ func (h *Handler) containerInit(namespace corev1.Namespace, pod corev1.Pod) (cor
...
+
+	var consulDNSIP string
+	if dnsEnabled {
+		for _, e := range os.Environ() {
+			if strings.Contains(e, dnsServiceHostEnvSuffix) {
+				dnsServiceHostEnv := strings.SplitN(e, "=", 2)
+				consulDNSIP = dnsServiceHostEnv[1]
+			}
+		}
+	}
```
lkysow, line 128: What if the Consul DNS service is in a different namespace? It won't be listed in the env vars.

charts/consul/values.yaml
```diff
@@ -1122,6 +1122,12 @@ dns:
   # @type: boolean
   enabled: "-"
 
+  # Used to determine if services using Consul Connect use Consul DNS
```
ishustava, line 1125:
```suggestion
  # If true, services using Consul Connect will use Consul DNS
```

lkysow (commented):
Still looking but wanted to get feedback on my question about namespaces.

ishustava (approved):

control-plane/connect-inject/container_init.go
ishustava: I don't think this is possible? connect injector and consul dns service are created in the same namespace if you're using the helm chart.

lkysow: ahh yeah sorry I thought this was the `connect-init` command

control-plane/connect-inject/container_init.go
```diff
@@ -66,6 +68,9 @@ type initContainerCommandData struct {
 	// TProxyExcludeUIDs is a list of additional user IDs to exclude from traffic redirection via
 	// the consul connect redirect-traffic command.
 	TProxyExcludeUIDs []string
+
+	// ConsulDNSIP is the IP of the Consul DNS Service.
+	ConsulDNSIP string
```
lkysow, line 73:
```suggestion
	// ConsulDNSIP is the IP of the Consul DNS Service.
	ConsulDNSClusterIP string
```

Maybe use cluster ip just so it's crystal?

control-plane/connect-inject/container_init.go
```diff
@@ -107,6 +112,21 @@ func (h *Handler) containerInit(namespace corev1.Namespace, pod corev1.Pod) (cor
...
 	}
 
+	dnsEnabled, err := consulDNSEnabled(namespace, pod, h.EnableConsulDNS)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+
+	var consulDNSIP string
+	if dnsEnabled {
+		for _, e := range os.Environ() {
```
lkysow, line 122: I found this kinda confusing. I think it could use a comment.

control-plane/connect-inject/container_init.go
lkysow: Can we take in `-release-prefix` and convert to uppercase/`_`'s?

lkysow (commented):
Looks mostly good but wondering about how we're finding the ClusterIP.

Also for acceptance tests what are we thinking? Is this gonna be a separate PR?

control-plane/connect-inject/container_init.go
thisisnotashwin: ✅

control-plane/connect-inject/container_init.go
thisisnotashwin: ✅

thisisnotashwin:
> Also for acceptance tests what are we thinking? Is this gonna be a separate PR?

@lkysow Acceptance tests are a great question. The approach I was considering for acceptance testing this change was indirectly using the changes the will be added for transparent proxy for x-partition communication. I will update the fixtures to support transparent proxying the upstream which should test the usage of Consul DNS we ultimately care about.

control-plane/connect-inject/container_init.go
```diff
@@ -117,12 +117,16 @@ func (h *Handler) containerInit(namespace corev1.Namespace, pod corev1.Pod) (cor
 		return corev1.Container{}, err
 	}
 
-	var consulDNSIP string
+	var consulDNSClusterIP string
 	if dnsEnabled {
+		// If Consul DNS is enabled, we loop through the environment variables
+		// to identify the one that has the value of the ClusterIP onf the Consul
```
lkysow, line 123:
```suggestion
		// to identify the one that has the value of the ClusterIP of the Consul
```

control-plane/connect-inject/handler.go
```diff
@@ -138,6 +138,10 @@ type Handler struct {
 	// from mesh services.
 	EnableConsulDNS bool
 
+	// ReleasePrefix is the prefix used for the installation which is used to determine the Service
+	// name of the Consul DNS service.
+	ReleasePrefix string
```
lkysow, line 143: Can you use `ResourcePrefix` to match server-acl-init, create-federation-job and others

control-plane/subcommand/inject-connect/command.go
```diff
@@ -94,6 +94,7 @@ type Command struct {
 
 	// Consul DNS flags.
 	flagEnableConsulDNS bool
+	flagReleasePrefix   string
```
lkysow, line 97: `resourcePrefix`

control-plane/connect-inject/container_init.go
```diff
@@ -117,12 +117,16 @@ func (h *Handler) containerInit(namespace corev1.Namespace, pod corev1.Pod) (cor
...
+		// DNS Service. constructDNSServiceHostName returns the name of the env
+		// variable whose value is the ClusterIP of the Consul DNS Service.
 		for _, e := range os.Environ() {
```
lkysow, line 126: Is this looping still necessary now that the env var is fully specified?

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -107,7 +107,8 @@ spec:
                 {{- else }}
                 -transparent-proxy-default-overwrite-probes=false \
                 {{- end }}
-                {{- if (and .Values.dns.enabled .Values.dns.enableRedirection)}}
+                {{- if (and .Values.dns.enabled .Values.dns.enableRedirection) }}
+                -release-prefix={{ template "consul.fullname" . }} \
```
lkysow, line 111: I think we should always set `-resource-prefix` because other developers might assume it's available.

Also I see there's a `CONSUL_FULLNAME` above. This is remnants from when we used to have a resource prefix flag I'm pretty sure. It can be removed.

lkysow (commented):
I think we can clean up the env var things now?

control-plane/connect-inject/container_init.go
thisisnotashwin: good catch!!

charts/consul/test/unit/connect-inject-deployment.bats
```diff
@@ -519,6 +519,51 @@ EOF
...
+@test "connectInject/Deployment: -release-prefix is true if dns.enabled=true and dns.enableRedirection=true" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/connect-inject-deployment.yaml \
+      --set 'connectInject.enabled=true' \
+      --set 'dns.enableRedirection=true' \
+      . | tee /dev/stderr |
+      yq -c -r '.spec.template.spec.containers[0].command | join(" ") | contains("-release-prefix=RELEASE-NAME-consul")' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
lkysow, line 565: don't need these now

control-plane/connect-inject/container_init.go
```diff
@@ -107,6 +112,19 @@ func (h *Handler) containerInit(namespace corev1.Namespace, pod corev1.Pod) (cor
...
+	if err != nil {
+		return corev1.Container{}, err
+	}
+
+	var consulDNSClusterIP string
+	if dnsEnabled {
+		// If Consul DNS is enabled, we find the environment variable that has the value
+		// of the ClusterIP of the Consul DNS Service. constructDNSServiceHostName returns
+		// the name of the env variable whose value is the ClusterIP of the Consul DNS Service.
+		consulDNSClusterIP = os.Getenv(h.constructDNSServiceHostName())
```
lkysow, line 125: this should error out if the env var is not found

lkysow (approved):

control-plane/connect-inject/container_init.go
```diff
@@ -107,6 +113,22 @@ func (h *Handler) containerInit(namespace corev1.Namespace, pod corev1.Pod) (cor
...
+		if consulDNSClusterIP == "" {
+			return corev1.Container{}, errors.New("failed to find ClusterIP for Consul DNS Service Host")
```
lkysow, line 128:
```suggestion
			return corev1.Container{}, fmt.Errorf("environment variable %s is not found", h.constructDNSServiceHostName())
```

lkysow (approved):

PR #835, opened 2021-11-03
cli: Create Secrets With A Label

sadjamz:
Changes proposed in this PR:
- Every-time a secret is created, add a label "managed-by" which is equal to "consul-k8s".
- On uninstall, only delete those secrets with said label.

How I've tested this PR:
Changed the uninstall test, and ensured any secret created with this label is deleted, while secrets lacking this label persist. Also some ad-hoc testing, running the `consul-k8s install` and `consul-k8s uninstall` command.

Code review.

Checklist:
- [X] Tests added
- [x] CHANGELOG entry added 


ishustava (commented):
Great work, Saad! 

A couple of comments/questions:
1. Should we also add label if the secret exists? I think some commands already update the secret if it exists (e.g webhook-cert-manager), while others (like `server-acl-init`) don't. Adding a label if secret exists is nice because if someone runs a helm upgrade, all secrets would get updated with a label.
2. I think the commands need unit tests to check that the label is added.

cli/cmd/install/install.go
```diff
@@ -371,8 +371,8 @@ func (c *Command) checkForPreviousSecrets() error {
 	}
 	for _, secret := range secrets.Items {
 		// future TODO: also check for federation secret
-		if strings.Contains(secret.Name, "consul-bootstrap-acl-token") {
-			return fmt.Errorf("found consul-acl-bootstrap-token secret from previous installations: %q in namespace %q. To delete, run kubectl delete secret %s --namespace %s",
+		if secret.ObjectMeta.Labels["managed-by"] == "consul-k8s" {
+			return fmt.Errorf("found consul secret from previous installations: %q in namespace %q. To delete, run kubectl delete secret %s --namespace %s",
```
t-eckert, line 375:
```suggestion
			return fmt.Errorf("found Consul secret from previous installations: %q in namespace %q. To delete, run kubectl delete secret %s --namespace %s",
```

cli/cmd/install/install_test.go
```diff
@@ -53,13 +53,14 @@ func TestCheckForPreviousSecrets(t *testing.T) {
...
-			Name: "test-consul-bootstrap-acl-token",
+			Name:   "test-consul-bootstrap-acl-token",
+			Labels: map[string]string{"managed-by": "consul-k8s"},
 		},
 	}
 	c.kubernetes.CoreV1().Secrets("default").Create(context.Background(), secret, metav1.CreateOptions{})
 	err := c.checkForPreviousSecrets()
 	require.Error(t, err)
-	require.Contains(t, err.Error(), "found consul-acl-bootstrap-token secret from previous installations: \"test-consul-bootstrap-acl-token\" in namespace \"default\". To delete, run kubectl delete secret test-consul-bootstrap-acl-token --namespace default")
+	require.Contains(t, err.Error(), "found consul secret from previous installations: \"test-consul-bootstrap-acl-token\" in namespace \"default\". To delete, run kubectl delete secret test-consul-bootstrap-acl-token --namespace default")
```
t-eckert, line 63: You might be better off only including the relevant snippet of the error you want to check. Maybe "found Consul secret from previous installations". The longer the string that you are checking, the higher the likelihood that something in your implementation changes which breaks this test even when the "behavior" of the code, throwing an error when a secret is found, hasn't changed.

In Go, it is somewhat necessary to check the text of the error because there is only one error type (I personally think this was a silly thing to do). However, when you work with languages that have different error types, it's usually better to check just the type of the error and not the text of the error.

cli/cmd/uninstall/uninstall_test.go
```diff
@@ -81,7 +82,8 @@ func TestDeleteSecrets(t *testing.T) {
 	require.NoError(t, err)
 	secrets, err := c.kubernetes.CoreV1().Secrets("default").List(context.Background(), metav1.ListOptions{})
 	require.NoError(t, err)
-	require.Len(t, secrets.Items, 0)
+	// Only secret1 should have been deleted, secret2 persists since it doesn't have the label.
```
t-eckert, line 85: That is a very good test! Great work!

t-eckert (commented):
Great work! Very clear code. I would echo what @ishustava said and change the `"consul-k8s"` string in both the CLI and Control Plane to be a constant that is passed around. That will make it easier for us to change later on if we need to.

ishustava (approved):
Great work Saad! I've left some minor comments, but won't block the merge.

cli/cmd/install/install.go
```diff
@@ -391,11 +390,12 @@ func (c *Command) checkForPreviousSecrets() error {
 	}
 	for _, secret := range secrets.Items {
 		// future TODO: also check for federation secret
-		if strings.Contains(secret.Name, "consul-bootstrap-acl-token") {
-			return fmt.Errorf("found consul-acl-bootstrap-token secret from previous installations: %q in namespace %q. To delete, run kubectl delete secret %s --namespace %s",
+		if secret.ObjectMeta.Labels[common.CLILabelKey] == common.CLILabelValue {
+			return fmt.Errorf("found Consul secret from previous installations: %q in namespace %q. To delete, run kubectl delete secret %s --namespace %s",
```
t-eckert, line 394:
```suggestion
			return fmt.Errorf("found Consul secret from previous installation: %q in namespace %q. To delete, run kubectl delete secret %s --namespace %s",
```

cli/cmd/uninstall/uninstall.go
```diff
@@ -392,13 +393,11 @@ func (c *Command) deleteSecrets(foundReleaseName, foundReleaseNamespace string)
...
-			err := c.kubernetes.CoreV1().Secrets(foundReleaseNamespace).Delete(c.Ctx, secret.Name, metav1.DeleteOptions{})
-			if err != nil {
-				return fmt.Errorf("deleteSecrets: error deleting Secret %q: %s", secret.Name, err)
-			}
-			secretNames = append(secretNames, secret.Name)
+		err := c.kubernetes.CoreV1().Secrets(foundReleaseNamespace).Delete(c.Ctx, secret.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return fmt.Errorf("deleteSecrets: error deleting Secret %q: %s", secret.Name, err)
 		}
+		secretNames = append(secretNames, secret.Name)
```
t-eckert, line 400: I would move the instantiation of `secretNames` to right above line 394, before you start the iteration. I would also change the name to `deletedSecrets` so it's more clear what you are listing.

control-plane/subcommand/common/common.go
```diff
@@ -26,6 +26,11 @@ const (
 	// ACLTokenSecretKey is the key that we store the ACL tokens in when we
 	// create Kubernetes secrets.
 	ACLTokenSecretKey = "token"
+
+	// CLILabelKey and CLILabelValue are added to each secret on creation so the CLI knows
+	// which key to delete on an uninstall.
```
t-eckert, line 31: I think you mean which "secret" instead of which "key".

t-eckert (approved):
Great work on this, Saad! You should feel very proud of the work you've done in your time here. It's been a pleasure working with you!

cli/cmd/uninstall/uninstall.go
ndhanushkodi: I lean towards secretNames rather than changing it to deletedSecrets because they are the names of secrets to delete, but haven't been deleted yet.

PR #836, opened 2021-11-03
Update CONTRIBUTING to reflect monorepo change

t-eckert:
Changes proposed in this PR:

- Update the CONTRIBUTING file to reflect the monorepo

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -794,30 +794,30 @@ Here are some things to consider before adding a test:
 
 ## Helm Reference Docs
  
-The helm reference docs (https://www.consul.io/docs/k8s/helm) are automatically
+The Helm reference docs (https://www.consul.io/docs/k8s/helm) are automatically
 generated from our `values.yaml` file.
 
 ### Generating Helm Reference Docs
  
 To generate the docs and update the `helm.mdx` file:
 
-1. Fork `hashicorp/consul` (https://github.com/hashicorp/consul) on GitHub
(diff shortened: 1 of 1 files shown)


david-yu (approved):
LGTM

ndhanushkodi (approved):

CONTRIBUTING.md
```diff
@@ -794,30 +794,30 @@ Here are some things to consider before adding a test:
...
 generated from our `values.yaml` file.
 
 ### Generating Helm Reference Docs
  
 To generate the docs and update the `helm.mdx` file:
 
-1. Fork `hashicorp/consul` (https://github.com/hashicorp/consul) on GitHub
-1. Clone your fork:
+1. Fork `hashicorp/consul` (https://github.com/hashicorp/consul) on GitHub.
+2. Clone your fork:
```
kschoche, line 805: @t-eckert : I believe the convention for numbering for .md files is to use `1.` for every line

lkysow: Agreed. Because when it renders it will automatically number and then when you edit the page to add a new point you don't need to re-number anything.

t-eckert: Gotcha. I will change it back and then merge the PR.

Just to share my thinking here, because this list is unlikely to change frequently, I thought I would take the chance to change the numbering so it's clearer for people reading the "source". But I'm not married to that idea, it's just something I've taken to doing in the past.

Issue #808, opened 2021-10-27
x509: consul-webhook-cert-manager is unable to put correct caBundle in consul-controller-mutating-webhook-configuration on consul update


amit106679:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

Upgrading any parameter in consul via helm in a kubernetes cluster starts giving error:  Internal error occurred: failed calling webhook "mutate-servicerouter.consul.hashicorp.com": Post "https://consul-consul-controller-webhook.consul.svc:443/mutate-v1alpha1-servicerouter?timeout=10s": x509: certificate signed by unknown authority.

### Reproduction Steps

Install consul 1.10.2 via helm.
Update any small parameter like affinity in the config yaml and update via helm. 

### Cause of the issue & Workaround

Updating consul resets caBundle value in consul-controller-mutating-webhook-configuration to the default value. Restarting the consul-webhook-cert-manager deployment mitigates the issue by putting correct value of caBundle again in the above mutatingwebhookconfiguration.

The expected behaviour should be that webhook-cert-manager reloads the caBundle correctly but on upgrade it is not able to till it is restarted.

lkysow:
I've reproduced this and we're working on a fix now.

amit106679:
@lkysow thanks for the fix - any timeline when this will be released on helm charts? 

Also we faced another issue -  x509: certificate has expired or is not yet valid: current time 2021-11-09T05:37:48Z is after 2021-11-08T19:40:44Z - please let me know if this is fixed or should I create another ticket for this?

lkysow:
We should have a release in the next two weeks.

For your other issue please open up a new ticket since the certs should
still be renewed even with the current release. Please also include which
platform you're running on. I've seen something similar when
running locally on Docker when I stop and restart the Docker daemon.

PR #837, opened 2021-11-03
Fix bug where webhookconfig would not be updated

lkysow:
Fixes an issue where if the webhookconfiguration is updated after the
webhook cert manager starts, then it is never reset to the correct
configuration until the manager is restarted or the certificate expires.

The issue was occurring because in the loop where we check the
webhookconfiguration we are only checking the webhookconfiguration of
the last updated certificate bundle. If there are multiple webhook
configurations then we are only watching a single one at a time since
the loop always operates on one bundle.

The fix is for each webhook to run its own loop.

In addition, the reason this was causing issues is because during a helm
upgrade we were resetting the webhookconfiguration's caBundle fields.
I've removed the caBundle setting (which already didn't exist in the
connect injector's webhook config) so that during a helm upgrade it
doesn't overwrite that field.

Fixes https://github.com/hashicorp/consul-k8s/issues/808

How I've tested this PR:
* unit tests (the test I added used to fail)
* manual test: install Consul and then edit both the connect inject webhook config and the controller webhook config and set caBundle to `""`. Wait a couple seconds. Now get the configs. Only _one_ will have been reset properly. Now do the same with this image (`ghcr.io/lkysow/consul-k8s-dev:nov03-3`)

* code/manual testing if interested

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ BUG FIXES:
     then on subsequent re-runs of server-acl-init the tokens are never set. [[GH-825](https://github.com/hashicorp/consul-k8s/issues/825)]
   * ACLs: Fix issue where if the number of Consul servers is increased, the new servers are never provisioned
     an ACL token. [[GH-677](https://github.com/hashicorp/consul-k8s/issues/677)]
+  * Fix issue where after a `helm upgrade`, users would see `x509: certificate signed by unknown authority.`
+    errors when modifying config entry resources. [[GH-837](https://github.com/hashicorp/consul-k8s/pull/837)]
 
 ## 0.36.0 (November 02, 2021)
 
--- a/control-plane/subcommand/webhook-cert-manager/command.go
+++ b/control-plane/subcommand/webhook-cert-manager/command.go
@@ -157,8 +157,6 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
-	certCh := make(chan cert.MetaBundle)
-
 	// Create the certificate notifier so we can update certificates,
 	// then start all the background routines for updating certificates.
 	var notifiers []*cert.Notify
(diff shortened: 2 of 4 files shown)


control-plane/subcommand/webhook-cert-manager/command.go
```diff
@@ -179,13 +177,14 @@ func (c *Command) Run(args []string) int {
 				Expiry: expiry,
 			}
 		}
+
+		certCh := make(chan cert.MetaBundle)
 		certNotify := &cert.Notify{Source: certSource, Ch: certCh, WebhookConfigName: config.Name, SecretName: config.SecretName, SecretNamespace: config.SecretNamespace}
 		notifiers = append(notifiers, certNotify)
 		go certNotify.Start(ctx)
+		go c.certWatcher(ctx, certCh, c.clientset, c.logger)
 	}
```
lkysow, line 187: The issue is on line 203. `bundle` is always set to the last bundle that came out of `ch`. Then we wake up every `defaultRetryDuration` to check things but we're only checking one `bundle`.

thisisnotashwin (approved):
This is excellent! Thanks for the fix @lkysow 🚀

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## UNRELEASED
 
+BUG FIXES:
+* Control Plane
+  * Fix issue where after a `helm upgrade`, users would see `x509: certificate signed by unknown authority.`
+    errors when modifying config entry resources. [[GH-808](https://github.com/hashicorp/consul-k8s/issues/808)]
```
kschoche, line 6: I see what you did here, but curious if this should link to the actual PR instead of the issue?

lkysow: done!

kschoche (approved):
Great work!!

PR #629, opened 2021-08-20
Re-enable psps in accceptance tests

ishustava:
Changes proposed in this PR:
- Allow privileged containers in the test PSP used by services deployed in tests

How I've tested this PR:
Ran GKE tests [here](https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/2435/workflows/5cbda3ca-0ab4-40e0-96b0-0e725cfc76bb/jobs/14252)

How I expect reviewers to test this PR:
code review

PR #839, opened 2021-11-04
Increase wait time to 10m

lkysow:
The wait time for pods to be ready was accidentally reduced to 3m here (https://github.com/hashicorp/consul-k8s/pull/629/files#diff-4e31c92ae1107b6f5baeb3ac84a55265254d58211a97d2ac12860edcd4414766R40) when the frequency was increased to 1s.

In one of my recent tests AKS took 3m to pull the image which then failed the test.

### DIFF
--- a/acceptance/framework/helpers/helpers.go
+++ b/acceptance/framework/helpers/helpers.go
@@ -34,10 +34,10 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
 
 	logger.Log(t, "Waiting for pods to be ready.")
 
-	// Wait up to 15m.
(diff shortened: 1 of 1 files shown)


lkysow:
The failures are unrelated.

ishustava (approved):
🤦‍♀️ I think when I decreased the `Wait` I didn't notice that the total time would be decreasing the total time. Do you think we should keep 15min wait like we intended?

t-eckert (approved):

lkysow:
> Do you think we should keep 15min wait like we intended?

It hasn't been failing very often so I think we can try 10 now. Maybe AKS disks sped up?

Issue #642, opened 2021-07-11
helm:UI metrics enabled when global.metrics.enabled=false


lkysow:
```
helm template consul hashicorp/consul --version 0.32.0 -s templates/server-config-configmap.yaml
---
# Source: consul/templates/server-config-configmap.yaml
# StatefulSet to run the actual Consul server cluster.
apiVersion: v1
kind: ConfigMap
metadata:
  name: consul-consul-server-config
  namespace: default
  labels:
    app: consul
    chart: consul-helm
    heritage: Helm
    release: consul
data:
  extra-from-values.json: |-
    {}

  ui-config.json: |-
    {
      "ui_config": {
        "enabled": true,
        "metrics_provider": "prometheus",
        "metrics_proxy": {
          "base_url": "http://prometheus-server"
        }
      }
    }
  central-config.json: |-
    {
      "enable_central_service_config": true
    }
```

The default of `global.metrics.enabled` is `false` and the default of `ui.metrics.enabled` is `-` so I'd expect the metrics to default to `false`.

I think what's happening is that `-` is being type-converted to boolean true in:

https://github.com/hashicorp/consul-helm/blob/9598a357b19184a73309077a536fefab629e41e1/templates/server-config-configmap.yaml#L30

PR #841, opened 2021-11-04
Fix logic for enabling ui metrics

lkysow:
Previously ui metrics would be enabled when
`global.metrics.enabled=false` and `ui.metrics.enabled=-`. Now that
would cause UI metrics to be disabled which makes sense since "-" means
inherit from `global.metrics.enabled`.

The logic is now to enable ui metrics when:

```
ui.enabled &&
  (
    (global.metrics.enabled == true && ui.metrics.enabled == "-") ||
    (ui.metrics.enabled == true)
  )
```

Fixes https://github.com/hashicorp/consul-k8s/issues/642

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,17 @@
 ## UNRELEASED
 
+BREAKING CHANGES:
+* Previously [UI metrics](https://www.consul.io/docs/connect/observability/ui-visualization) would be enabled when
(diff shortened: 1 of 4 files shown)


kschoche (approved):
nice!

t-eckert (approved):

Issue #574, opened 2021-07-27
Give better error when ACLs enabled and using the default service account name


lkysow:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Feature Description

If ACLs are enabled and services are using the default service account, you'll see this error in the init container logs:

```
2021-07-27T16:43:27.707Z [ERROR] Consul login failed; retrying: error="error logging in: Unexpected response code: 403 (rpc error making call: Permission denied)"
2021-07-27T16:43:28.720Z [ERROR] Consul login failed; retrying: error="error logging in: Unexpected response code: 403 (rpc error making call: Permission denied)"
2021-07-27T16:43:29.729Z [ERROR] Consul login failed; retrying: error="error logging in: Unexpected response code: 403 (rpc error making call: Permission denied)"
2021-07-27T16:43:30.795Z [ERROR] Consul login failed; retrying: error="error logging in: Unexpected response code: 403 (rpc error making call: Permission denied)"
2021-07-27T16:43:30.795Z [ERROR] Hit maximum retries for consul login: error="error logging in: Unexpected response code: 403 (rpc error making call: Permission denied)"
```

To fix, add a service account that matches the name of the service. But perhaps we could figure this out beforehand and give a better error on pod scheduling.

ishustava:
One thing to keep in mind here is that we don't always know whether the default service account is allowed in your installation unless we look at the binding rule. By default, the [binding rule](https://github.com/hashicorp/consul-helm/blob/860365145ea1a6c229f153d55085dfcd7fa29c62/values.yaml#L1627) we use does not allow that, but if someone overrides it in the helm chart, then we'd need a way to determine that.

PR #842, opened 2021-11-04
Give better error when using default service acct

lkysow:
If consul login fails when the service account name is `default` then
give an explicit warning that the reason it failed is because in
default installations that is not a support service account name.

We can't fail during injection because we support modifying the binding
rule such that `default` _is_ a valid svc account name.

Fixes https://github.com/hashicorp/consul-k8s/issues/574

How I've tested this PR:
* built container and looked at logs

Checklist:
- [x] CHANGELOG entry added 


ishustava (approved):

t-eckert (approved):
Nice UX

Issue #593, opened 2021-08-09
Add validation to ConnectInject Consul destination namespace


Ranjandas:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Is your feature request related to a problem? Please describe.

The ConnectInject Consul destination namespace currently doesn't have any validation. Setting it to `consul` will result in Sidecar injection fail with the following error:

```
[ERROR] handler: Error checking or creating namespace: err="Unexpected response code: 500 (rpc error making call: Namespace "consul" is reserved for future internal use)" Namespace=consul Request Name=
```

The root cause is that name `consul`  for namespaces is reserved and can't be created in Consul.

#### Feature Description

There should be a validation that fails the installation if the value is set to `consul`.

PR #846, opened 2021-11-05
Prevent using reserved names for ns's/partitions

lkysow:
Some names are reserved by Consul and can't be used for namespaces or
partitions. Catching this early before the helm install is a better UX
than having the install fail (in the case of partitions) or succeed (in
the case of namespaces) and then fail during syncing or injection.

This code uses some tricks:
1. The code for failing on reserved names is extracted into a helper template so it can be re-used in other templates
2. The test code uses a helper function so there's less duplication in the tests

Fixes https://github.com/hashicorp/consul-k8s/issues/593

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -119,3 +119,21 @@ This template is for an init container.
       memory: "50Mi"
       cpu: "50m"
 {{- end -}}
+
+{{/*
+Fails when a reserved name is passed in. This should be used to test against
+Consul namespaces and partition names.
+This template accepts an array that contains two elements. The first element
+is the name that's being checked and the second is the name of the values.yaml
+key that's setting the name.
+
+Usage: {{ template "consul.reservedNamesFailer" (list .Values.key "key") }}
+
+*/}}
+{{- define "consul.reservedNamesFailer" -}}
+{{- $name := index . 0 -}}
+{{- $key := index . 1 -}}
+{{- if or (eq "system" $name) (eq "universal" $name) (eq "consul" $name) (eq "operator" $name) (eq "root" $name) }}
+{{- fail (cat "The name" $name "set for key" $key "is reserved by Consul for future use." ) }}
+{{- end }}
+{{- end -}}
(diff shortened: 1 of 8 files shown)


charts/consul/templates/_helpers.tpl
```diff
@@ -119,3 +119,21 @@ This template is for an init container.
...
+Usage: {{ template "consul.reservedNamesFailer" (list .Values.key "key") }}
+
+*/}}
+{{- define "consul.reservedNamesFailer" -}}
+{{- $name := index . 0 -}}
+{{- $key := index . 1 -}}
+{{- if or (eq "system" $name) (eq "universal" $name) (eq "consul" $name) (eq "operator" $name) (eq "root" $name) }}
+{{- fail (cat "The name" $name "set for key" $key "is reserved by Consul for future use." ) }}
+{{- end }}
+{{- end -}}
```
lkysow, line 139: Review notes:
* See template docs for how it works
* I'm using `cat` inside `fail` to get those values interpolated into the error string.

ishustava: 🤯

ishustava (approved):

charts/consul/test/unit/connect-inject-deployment.bats
```diff
@@ -1512,4 +1512,41 @@ EOF
...
+  cd `chart_dir`
+  local -r name="$1"
+		run helm template \
+				-s templates/connect-inject-deployment.yaml  \
+				--set 'connectInject.enabled=true' \
+				--set "connectInject.consulNamespaces.consulDestinationNamespace=$name" .
+
+		[ "$status" -eq 1 ]
+		[[ "$output" =~ "The name $name set for key connectInject.consulNamespaces.consulDestinationNamespace is reserved by Consul for future use" ]]
+}
```
thisisnotashwin, line 1552: 💥

thisisnotashwin (approved):
NAICE!

PR #858, opened 2021-11-11
Ignore services in endpoint controller using `consul.hashicorp.com/service-ignore`

### LINKED ISSUE #592: If two Services select the same pods then pods won't start
### Community Note

---

### Overview of the Issue

If two kube services have the same selector of mesh pods then Consul will register 4 service instances (two for each pod, sidecar and app) and then connect-init will loop forever because it's waiting for exactly 2 service instances.

### Possible solutions

We can't register two service instances for one pod because the pod's sidecar can only receive traffic for one service name. So I think we need to provide a way to exclude one of the services from syncing. I checked and if you label the service, the labels are available at the endpoints level so we could do something basic like:

```
labels:
  consul.hashicorp.com/connect-ignore: true
```

t-eckert:
> Fixes #592 

Changes proposed in this PR:
- If a service has the label `consul.hashicorp.com/service-ignore: "true"`, it will not be registered by the endpoints controller. If is has been registered before, it will be deregistered.
- If the user does not have the correct number of services registered, they will get a log instructing them to add the "connect ignore" label to services which should not be routed.
- Some comments have been added where it took me time to understand what processes were doing.
- `mapAddresses` function created as a refactor. I spun off reading this "subprocess" in the `Reconcile` method and got sidetracked. I think it makes sense to abstract over it. Feel free to disagree.

How I've tested this PR:
- Unit tests.
- Manual test (detailed below).

- Running the unit tests.
- The manual testing setup I have been using is [here](https://github.com/t-eckert/boilerplates/tree/main/kubernetes/headless-service)
  To test the behavior run the `install` script which will install Consul onto the cluster; create two services, one with the `connect-ignore` label; and create a pod which will come online.

Here is a demo of the change, initially the services are deployed without the `connect-ignore` label on the "headless" service. The pod crashes as `consul-connect-inject-init` cannot register the service. When the label is added to the "headless" service, the pod comes up in a healthy state:

https://user-images.githubusercontent.com/29112081/142670205-671b7d73-0461-4c6b-bb47-69222c379540.mov

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ IMPROVEMENTS:
    * Pre-check in the `install` command to verify the correct license secret exists when using an enterprise Consul image. [[GH-875](https://github.com/hashicorp/consul-k8s/pull/875)]
 * Control Plane
    * Add a label "managed-by" to every secret the control-plane creates. Only delete said secrets on an uninstall. [[GH-835](https://github.com/hashicorp/consul-k8s/pull/835)]
-  * Add `consul.hashicorp.com/service-ignore` to prevent services from being registered in Consul. [[GH-858](https://github.com/hashicorp/consul-k8s/pull/858)]
+  * Add  support for labeling a Kubernetes service with `consul.hashicorp.com/service-ignore` to prevent services from being registered in Consul. [[GH-858](https://github.com/hashicorp/consul-k8s/pull/858)]
 
 ## 0.37.0 (November 18, 2021)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ IMPROVEMENTS:
    * Pre-check in the `install` command to verify the correct license secret exists when using an enterprise Consul image. [[GH-875](https://github.com/hashicorp/consul-k8s/pull/875)]
 * Control Plane
    * Add a label "managed-by" to every secret the control-plane creates. Only delete said secrets on an uninstall. [[GH-835](https://github.com/hashicorp/consul-k8s/pull/835)]
-  * Add  support for labeling a Kubernetes service with `consul.hashicorp.com/service-ignore` to prevent services from being registered in Consul. [[GH-858](https://github.com/hashicorp/consul-k8s/pull/858)]
+  * Add support for labeling a Kubernetes service with `consul.hashicorp.com/service-ignore` to prevent services from being registered in Consul. [[GH-858](https://github.com/hashicorp/consul-k8s/pull/858)]
 
 ## 0.37.0 (November 18, 2021)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ BREAKING CHANGES:
 IMPROVEMENTS:
 * CLI
    * Pre-check in the `install` command to verify the correct license secret exists when using an enterprise Consul image. [[GH-875](https://github.com/hashicorp/consul-k8s/pull/875)]
+* Control Plane
    * Add a label "managed-by" to every secret the control-plane creates. Only delete said secrets on an uninstall. [[GH-835](https://github.com/hashicorp/consul-k8s/pull/835)]
+  * Add `consul.hashicorp.com/service-ignore` to prevent services from being registered in Consul. [[GH-858](https://github.com/hashicorp/consul-k8s/pull/858)]
 
 ## 0.37.0 (November 18, 2021)
 
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -149,7 +149,7 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	// If the endpoints object has the label "connect-ignore" set to true, deregister all instances in Consul for this service.
 	// It is possible that the endpoints object has never been registered, in which case deregistration is a no-op.
-	if value, ok := serviceEndpoints.Labels[labelConnectIgnore]; ok && value == "true" {
+	if value, ok := serviceEndpoints.Labels[labelServiceIgnore]; ok && value == "true" {
 		return r.deregisterServiceOnAllAgents(ctx, req.Name, req.Namespace, nil, endpointPods)
 	}
 
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -140,7 +140,8 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 	if k8serrors.IsNotFound(err) {
 		// Deregister all instances in Consul for this service. The function deregisterServiceOnAllAgents handles
 		// the case where the Consul service name is different from the Kubernetes service name.
-		return r.deregisterServiceOnAllAgents(ctx, req.Name, req.Namespace, nil, endpointPods)
+		err = r.deregisterServiceOnAllAgents(ctx, req.Name, req.Namespace, nil, endpointPods)
+		return ctrl.Result{}, err
 	} else if err != nil {
 		r.Log.Error(err, "failed to get Endpoints", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, err
@@ -660,7 +661,7 @@ func getHealthCheckStatusReason(healthCheckStatus, podName, podNamespace string)
 // The argument endpointsAddressesMap decides whether to deregister *all* service instances or selectively deregister
 // them only if they are not in endpointsAddressesMap. If the map is nil, it will deregister all instances. If the map
 // has addresses, it will only deregister instances not in the map.
-func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, k8sSvcName, k8sSvcNamespace string, endpointsAddressesMap map[string]bool, endpointPods mapset.Set) (ctrl.Result, error) {
+func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, k8sSvcName, k8sSvcNamespace string, endpointsAddressesMap map[string]bool, endpointPods mapset.Set) error {
 	// Get all agents by getting pods with label component=client, app=consul and release=<ReleaseName>
 	agents := corev1.PodList{}
 	listOptions := client.ListOptions{
@@ -673,7 +674,7 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 	}
 	if err := r.Client.List(ctx, &agents, &listOptions); err != nil {
 		r.Log.Error(err, "failed to get Consul client agent pods")
-		return ctrl.Result{}, err
+		return err
 	}
 
 	// On each agent, we need to get services matching "k8s-service-name" and "k8s-namespace" metadata.
@@ -681,14 +682,14 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 		client, err := r.remoteConsulClient(agent.Status.PodIP, r.consulNamespace(k8sSvcNamespace))
 		if err != nil {
 			r.Log.Error(err, "failed to create a new Consul client", "address", agent.Status.PodIP)
-			return ctrl.Result{}, err
+			return err
 		}
 
 		// Get services matching metadata.
 		svcs, err := serviceInstancesForK8SServiceNameAndNamespace(k8sSvcName, k8sSvcNamespace, client)
 		if err != nil {
 			r.Log.Error(err, "failed to get service instances", "name", k8sSvcName)
-			return ctrl.Result{}, err
+			return err
 		}
 
 		// Deregister each service instance that matches the metadata.
@@ -702,7 +703,7 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 					r.Log.Info("deregistering service from consul", "svc", svcID)
 					if err = client.Agent().ServiceDeregister(svcID); err != nil {
 						r.Log.Error(err, "failed to deregister service instance", "id", svcID)
-						return ctrl.Result{}, err
+						return err
 					}
 					serviceDeregistered = true
 				}
@@ -710,7 +711,7 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 				r.Log.Info("deregistering service from consul", "svc", svcID)
 				if err = client.Agent().ServiceDeregister(svcID); err != nil {
 					r.Log.Error(err, "failed to deregister service instance", "id", svcID)
-					return ctrl.Result{}, err
+					return err
 				}
 				serviceDeregistered = true
 			}
@@ -720,12 +721,13 @@ func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, 
 				err = r.deleteACLTokensForServiceInstance(client, serviceRegistration.Service, k8sSvcNamespace, serviceRegistration.Meta[MetaKeyPodName])
 				if err != nil {
 					r.Log.Error(err, "failed to reconcile ACL tokens for service", "svc", serviceRegistration.Service)
-					return ctrl.Result{}, err
+					return err
 				}
 			}
 		}
 	}
-	return ctrl.Result{}, nil
+
+	return nil
 }
 
 // deleteACLTokensForServiceInstance finds the ACL tokens that belongs to the service instance and deletes it from Consul.
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -118,10 +118,14 @@ type EndpointsController struct {
 	context.Context
 }
 
+// Reconcile reads the state of the cluster's service endpoints when a request is received and makes changes
+// to the endpoints based on the state read. Requests are created when the endpoints are created, updated,
+// or deleted.
 func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var errs error
 	var serviceEndpoints corev1.Endpoints
 
+	// Ignore the request if the namespace of the endpoint is not allowed.
 	if shouldIgnore(req.Namespace, r.DenyK8sNamespacesSet, r.AllowK8sNamespacesSet) {
 		return ctrl.Result{}, nil
 	}
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -150,6 +150,7 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 	// If the endpoints object has the label "connect-ignore" set to true, deregister all instances in Consul for this service.
 	// It is possible that the endpoints object has never been registered, in which case deregistration is a no-op.
 	if value, ok := serviceEndpoints.Labels[labelServiceIgnore]; ok && value == "true" {
+		// We always deregister the service to handle the case where a user has registered the service, then added the label later.
 		err = r.deregisterServiceOnAllAgents(ctx, req.Name, req.Namespace, nil, endpointPods)
 		return ctrl.Result{}, err
 	}
@@ -647,11 +648,6 @@ func getHealthCheckStatusReason(healthCheckStatus, podName, podNamespace string)
 	}
 
 	return fmt.Sprintf("Pod \"%s/%s\" is not ready", podNamespace, podName)
-}
-
-// fetchServices returns a list of services from the given k8s client.
-func (r *EndpointsController) fetchServices(ctx context.Context, k8sSvcName, k8sSvcNamespace string) ([]*api.AgentService, error) {
-	return nil, nil
 }
 
 // deregisterServiceOnAllAgents queries all agents for service instances that have the metadata
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -161,17 +161,7 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	// Register all addresses of this Endpoints object as service instances in Consul.
 	for _, subset := range serviceEndpoints.Subsets {
-		// Combine all addresses to a map, with a value indicating whether an address is ready or not.
-		allAddresses := make(map[corev1.EndpointAddress]string)
-		for _, readyAddress := range subset.Addresses {
-			allAddresses[readyAddress] = api.HealthPassing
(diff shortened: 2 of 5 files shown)


control-plane/connect-inject/endpoints_controller.go
```diff
@@ -137,15 +140,21 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		return r.deregisterServiceOnAllAgents(ctx, req.Name, req.Namespace, nil, endpointPods)
 	} else if err != nil {
 		r.Log.Error(err, "failed to get Endpoints", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, err
 	}
 
+	// If the endpoints object has the label "connect-ignore" set to true, deregister all instances in Consul for this service.
+	// It is possible that the endpoints object has never been registered, in which case deregistration is a no-op.
+	if value, ok := serviceEndpoints.Labels[labelServiceIgnore]; ok && value == "true" {
+		if isRegistered() {
```
t-eckert, line 152: Right now this always returns `true` and deregistration is always run. This is... ok because deregistration is idempotent. However, it wastes compute time and I would like to short circuit it. Still figuring out how.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -535,6 +535,87 @@ func TestProcessUpstreams(t *testing.T) {
...
+		k8sObjects                 func() []runtime.Object
+		initialConsulSvcs          []*api.AgentServiceRegistration
+		expectedNumSvcInstances    int
+		expectedConsulSvcInstances []*api.CatalogService
+		expectedProxySvcInstances  []*api.CatalogService
+		expectedAgentHealthChecks  []*api.AgentCheck
+		expErr                     string
+	}{
+		// TODO: Test case where service is registered and should be deregistered.
+		// TODO: Test case where service is not registered and just gets ignored.
```
thisisnotashwin, line 555: I dont know if this is intentional but I think we do need to add tests here.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -535,6 +535,87 @@ func TestProcessUpstreams(t *testing.T) {
...
+				ReleaseName:           "consul",
+				ReleaseNamespace:      "default",
+				ConsulClientCfg:       cfg,
+			}
+			namespacedName := types.NamespacedName{
+				Namespace: "default",
+				Name:      "service-created",
+			}
+
+			_, _ = ep.Reconcile(context.Background(), ctrl.Request{NamespacedName: namespacedName})
```
thisisnotashwin, line 614: This test currently lacks assertions.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -661,7 +660,7 @@ func getHealthCheckStatusReason(healthCheckStatus, podName, podNamespace string)
 // The argument endpointsAddressesMap decides whether to deregister *all* service instances or selectively deregister
 // them only if they are not in endpointsAddressesMap. If the map is nil, it will deregister all instances. If the map
 // has addresses, it will only deregister instances not in the map.
-func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, k8sSvcName, k8sSvcNamespace string, endpointsAddressesMap map[string]bool, endpointPods mapset.Set) error {
+func (r *EndpointsController) deregisterServiceOnAllAgents(ctx context.Context, k8sSvcName, k8sSvcNamespace string, endpointsAddressesMap map[string]bool, endpointPods mapset.Set) (ctrl.Result, error) {
```
thisisnotashwin, line 663: Looks like this method always returns only `ctrl.Result{}`. IMO we should undo this refactor as it makes the method signature confusing.

thisisnotashwin (commented):
hey @t-eckert. I took a quick look and the approach looks sound. Left some comment inline.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -4761,6 +4842,29 @@ func TestGetTokenMetaFromDescription(t *testing.T) {
...
+	addresses := corev1.EndpointSubset{
+		Addresses: []corev1.EndpointAddress{
+			{Hostname: "host1"},
+			{Hostname: "host2"},
+		},
+		NotReadyAddresses: []corev1.EndpointAddress{
+			{Hostname: "host3"},
+			{Hostname: "host4"},
+		},
+	}
```
thisisnotashwin, line 4855: You can add a few more cases here where all addresses are ready or all addresses are not ready and compare them against the output.

control-plane/connect-inject/endpoints_controller.go
t-eckert: Makes sense. I like how clean this makes calling the method, but I get your concern.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -117,10 +117,13 @@ type EndpointsController struct {
 	context.Context
 }
 
+// Reconcile reads the state of the cluster's service endpoints when a request is received and makes changes
+// to the endpoints based on the state read. Reconcile is called when services are created, updated, or deleted.
```
kschoche, line 121:
```suggestion
// Reconcile reads the state of an Endpoints object for a Kubernetes Service and reconciles Consul services which
// correspond to the Kubernetes Service. These events are driven by changes to the Pods backing the Kube service.
```

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -137,15 +140,21 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
...
-		}
-		return ctrl.Result{}, nil
+		err = r.deregisterServiceOnAllAgents(ctx, req.Name, req.Namespace, nil, endpointPods)
+		return ctrl.Result{}, err
 	} else if err != nil {
 		r.Log.Error(err, "failed to get Endpoints", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, err
 	}
 
+	// If the endpoints object has the label "service-ignore" set to true, deregister all instances in Consul for this service.
```
kschoche, line 150:
```suggestion
	// If the endpoints object has the label "consul.hashicorp.com/service-ignore" set to true, deregister all instances in Consul for this service.
```

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -1004,10 +1004,22 @@ func (r *EndpointsController) consulNamespace(namespace string) string {
...
-	if anno, ok := pod.Annotations[keyInjectStatus]; ok {
-		if anno == injected {
-			return true
-		}
+	if anno, ok := pod.Annotations[keyInjectStatus]; ok && anno == injected {
+		return true
 	}
 	return false
 }
+
```
kschoche, line 1012: This function doesn't appear to be used by anyone other than the place it was removed from.
Would probably make sense to remove this change and its tests, unless there was some reason that I'm not understanding. Was there a bug that is fixed in this?

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -2717,6 +2717,128 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
 	}
 }
 
+// TestReconcileIgnoresServiceIgnoreLabel tests that the endpoints controller correctly ignores services
+// with the service-ignore label and deregisters services previously registered if the service-ignore
+// label is added. This is done by running Reconcile twice, once on an initial set of labels, then again
+// on a new set of labels with the service-ignore label. The number of services registered in Consul is
+// checked before updating the labels against the `svcsExpectedBeforeUpdate` and after updating the labels
+// where the expected number of services should be zero.
+func TestReconcileIgnoresServiceIgnoreLabel(t *testing.T) {
```
kschoche, line 2726: This looks good, but I think it can be simplified so that we're not running `Reconcile()` twice, which could introduce weird behaviour in failure cases.

You can set it up to do a table test which holds the initial configuration of the pods, and services, and then an expected result, and then it runs with the following behaviour:
1. Initialize kube + consul with all of the initial configurations and service registrations.
2. Run `Reconcile()` 
3. Read Consul's services to see the expected end state has occurred.

Then you could run this with an initial configuration where the service is registered AND does/not have the label, and again when the service isn't registered AND does/not have the label.

Let me know if this does not make sense and I'd be happy to hop on a call and walk through it!

control-plane/subcommand/connect-init/command.go
```diff
@@ -177,7 +177,12 @@ func (c *Command) Run(args []string) int {
 			// it is not "lost" to the user at the end of the retries when the pod enters a CrashLoop.
 			if registrationRetryCount%10 == 0 {
 				c.logger.Info("Check to ensure a Kubernetes service has been created for this application." +
-					" If your pod is not starting also check the connect-inject deployment logs.")
+					" If your pod is not starting also check the connect-inject deployment logs." +
+					"\nNote that only one service may be used to route requests to each pod." +
+					" If you need multiple services to point to the same pod, add the label" +
+					" `consul.hashicorp.com/connect-ignore: \"true\"` to services which should" +
```
kschoche, line 183:
```suggestion
					" `consul.hashicorp.com/service-ignore: \"true\"` to services which should" +
```

control-plane/subcommand/connect-init/command.go
```diff
@@ -177,7 +177,12 @@ func (c *Command) Run(args []string) int {
 			// it is not "lost" to the user at the end of the retries when the pod enters a CrashLoop.
 			if registrationRetryCount%10 == 0 {
 				c.logger.Info("Check to ensure a Kubernetes service has been created for this application." +
-					" If your pod is not starting also check the connect-inject deployment logs.")
+					" If your pod is not starting also check the connect-inject deployment logs." +
```
lkysow, line 180: @kschoche and I talked about this and we had an idea to log this connect-ignore recommendation only if `len(serviceList) > 2`.

So something like "Error: there are multiple Consul services registered for this pod when there should only be one. Check if there are multiple Kubernetes services pointing at this pod and add the label consul.hashicorp.com/connect-ignore: \"true\" to all but one of the services."

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -137,15 +140,21 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		err = r.deregisterServiceOnAllAgents(ctx, req.Name, req.Namespace, nil, endpointPods)
+		return ctrl.Result{}, err
 	} else if err != nil {
 		r.Log.Error(err, "failed to get Endpoints", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, err
 	}
 
+	// If the endpoints object has the label "consul.hashicorp.com/service-ignore" set to true, deregister all instances in Consul for this service.
+	// It is possible that the endpoints object has never been registered, in which case deregistration is a no-op.
+	if value, ok := serviceEndpoints.Labels[labelServiceIgnore]; ok && value == "true" {
```
ishustava, line 152: I think we should check the value with `strconv.ParseBool` the same way we do for true/false annotations like `consul.hashicorp.com/connect-inject`. This function will parse any truthy value.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -200,7 +199,7 @@ func (r *EndpointsController) SetupWithManager(mgr ctrl.Manager) error {
 		).Complete(r)
 }
 
-// registerServicesAndHealthCheck creates Consul registrations for the service and proxy and register them with Consul.
+// registerServicesAndHealthCheck creates Consul registrations for the service and proxy and registers them with Consul.
```
ishustava, line 202: 🙏

control-plane/connect-inject/endpoints_controller.go
ishustava: I think this was added for readability.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -2717,6 +2717,128 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
...
+			},
+			svcsExpectedBeforeUpdate: 0,
+		},
+		"Endpoint without labels is registered, then deregisterd when service-ignore is added": {
+			initialLabels: map[string]string{},
+			updatedLabels: map[string]string{
+				labelServiceIgnore: "true",
+			},
+			svcsExpectedBeforeUpdate: 1,
+		},
```
ishustava, line 2752: might be worth testing the update the other way around too: endpoints with label is not registered, then you update the label and check that it's registered.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -2717,6 +2717,128 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
...
+				},
+			}
+			pod1 := createPod("pod1", "1.2.3.4", true, true)
+			fakeClientPod := createPod("fake-consul-client", "127.0.0.1", false, true)
+			fakeClientPod.Labels = map[string]string{"component": "client", "app": "consul", "release": "consul"}
+			ns := corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
+			k8sObjects := []runtime.Object{endpoint, pod1, fakeClientPod, &ns}
+			fakeClient := fake.NewClientBuilder().WithRuntimeObjects(k8sObjects...).Build()
+
+			// Create test Consul server
```
ishustava, line 2787: nit: might need to update other comments to follow godoc style.
```suggestion
			// Create test Consul server.
```

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -2717,6 +2717,128 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
...
+
+			// Run the reconcile process the second time
+			resp, err = ep.Reconcile(context.Background(), ctrl.Request{NamespacedName: namespacedName})
+			require.NoError(t, err)
+			require.False(t, resp.Requeue)
+
+			// Check that no services are registered with Consul
+			serviceInstances, _, err = consulClient.Catalog().Service(serviceName, "", nil)
+			require.NoError(t, err)
+			require.Empty(t, serviceInstances)
```
ishustava, line 2837: Great work on this test! It was super easy to read!!

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -4761,6 +4883,74 @@ func TestGetTokenMetaFromDescription(t *testing.T) {
...
+			},
+		},
+	}
+
+	for name, c := range cases {
+		t.Run(name, func(t *testing.T) {
+			actual := mapAddresses(c.addresses)
+			require.Equal(t, c.expected, actual)
+		})
+	}
```
ishustava, line 4951: nice!

control-plane/subcommand/connect-init/command.go
ishustava: yeah I think that's a great idea.

manobi:
This looks very useful, but I would like to present a different use case.
I'm using Loki helm chart, which does not have the option to add annotations in headless-service (consul is trying to register loki and loki-headless):
https://github.com/grafana/helm-charts/blob/main/charts/loki/templates/service-headless.yaml

In this scenario I would prefere to use this annotations in pod annotation instead of service:

```
podAnnotations:
  consul.hashicorp.com/service-ignore: "loki-headless"
```

I can see many scenarios where "service-ignore" would work, but it would be cool to control it in pod annotations along with other consul pairs.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+  * Add `consul.hashicorp.com/service-ignore` to prevent services from being registered in Consul. [[GH-858](https://github.com/hashicorp/consul-k8s/pull/858)]
```
kschoche, line 4:
```suggestion
  * Add  support for labeling a Kubernetes service with `consul.hashicorp.com/service-ignore` to prevent services from being registered in Consul. [[GH-858](https://github.com/hashicorp/consul-k8s/pull/858)]
```
something like this?

control-plane/connect-inject/endpoints_controller.go
t-eckert: I think that's a great improvement. I made the change, but it does make the code a little longer because I have to handle the error from parsing the bool. Let me know if you would have handled it differently. I don't think it's complicated enough to separate into another function yet.

control-plane/subcommand/connect-init/command.go
t-eckert: Great idea!

control-plane/connect-inject/endpoints_controller_test.go
t-eckert: Cool, I made that modification. Let me know what you think.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -2717,6 +2717,137 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
...
+					Name:    serviceName,
+					Port:    0,
+					Address: "1.2.3.4",
+					Meta: map[string]string{
+						"k8s-namespace":    namespace,
+						"k8s-service-name": serviceName,
+						"managed-by":       "consul-k8s-endpoints-controller",
+						"pod-name":         "pod1",
+					},
+				})
```
kschoche, line 2819: For completeness sake you'd also usually have a proxy service associated with this service to simulate a happy path where a service has actually been registered. That being said I took at a look at the code and creating one here wouldn't necessarily exercise additional code because it just loops over a filtered list from Consul.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -2717,6 +2717,137 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
...
+			resp, err := ep.Reconcile(context.Background(), ctrl.Request{NamespacedName: namespacedName})
+			require.NoError(t, err)
+			require.False(t, resp.Requeue)
+
+			// Check that no services are registered with Consul
+			serviceInstances, _, err := consulClient.Catalog().Service(serviceName, "", nil)
+			require.NoError(t, err)
+			require.Len(t, serviceInstances, tt.expectedNumSvcInstances)
+		})
+	}
```
kschoche, line 2848: Great work on this test :-)

control-plane/subcommand/connect-init/command.go
```diff
@@ -179,6 +179,12 @@ func (c *Command) Run(args []string) int {
 				c.logger.Info("Check to ensure a Kubernetes service has been created for this application." +
 					" If your pod is not starting also check the connect-inject deployment logs.")
 			}
+			if len(serviceList) > 2 {
+				c.logger.Error("There are multiple Consul services registered for this pod when there should only be one." +
+					" Check if there are multiple Kubernetes services pointing at this pod and add the label" +
+					" `consul.hashicorp.com/service-ignore: \"true\"` to services which should not be used by Consul for handling requests.")
```
kschoche, line 185: I think it's good to be very explicit here so as not to leave room for users to shoot themselves in the feet.

```suggestion
				c.logger.Error("There are multiple Consul services registered for this pod when there must only be one." +
					" Check if there are multiple Kubernetes services selecting this pod and add the label" +
					" `consul.hashicorp.com/service-ignore: \"true\"` to all services except the one used by Consul for handling requests.")
```

kschoche (approved):
Looks great @t-eckert, great work on this one.
I left a couple comments, 👍🏽

control-plane/connect-inject/endpoints_controller_test.go
t-eckert: Thank you!

control-plane/connect-inject/endpoints_controller_test.go
t-eckert: I'm a fan of completeness. What do you think of what I did? I added the proxy service and I also added the check that it's treated properly. Is that the most efficient way to do it?

ishustava (approved):
Looks great Thomas!

nflaig:
@t-eckert is there a specific reason why this has to be set as a label? most/all helm charts support setting custom annotations through values but labels are sometimes not supported. Everything else also uses annotations this seems to be the only case where a label is used instead.

control-plane/subcommand/connect-init/command.go
```diff
@@ -179,6 +179,12 @@ func (c *Command) Run(args []string) int {
 				c.logger.Info("Check to ensure a Kubernetes service has been created for this application." +
 					" If your pod is not starting also check the connect-inject deployment logs.")
 			}
+			if len(serviceList) > 2 {
+				c.logger.Error("There are multiple Consul services registered for this pod when there must only be one." +
```
nflaig, line 183: @t-eckert I am running multiple stateful sets in kubernets (kafka, zookeeper, elasticsearch, redis) and all of those have two services normal + headless but I dont get this error? shouldnt this fail the pod scheduling?

lkysow:
> is there a specific reason why this has to be set as a label?

Because we're reacting on changes to the Endpoint object which gets its label from the Service but it doesn't get annotations from the Service.

Note that if you have ACLs enabled we're planning on automatically ignoring all Services that don't have the same name as the `serviceAccountName`. In which case you won't need to label the other services with ignore.

control-plane/subcommand/connect-init/command.go
lkysow: Hmm, yes it should be failing scheduling if both the headless and non-headless services select the same pods. Maybe open up an issue about this.

nflaig:
>  Note that if you have ACLs enabled we're planning on automatically ignoring all Services that don't have the same name as the serviceAccountName. In which case you won't need to label the other services with ignore.

It that a good idea? If you run stateful sets the headless service is used by most client-to-server connections. Or is there a solution for this? Maybe you could first test that with the most common stateful sets like kafka, zookeeper, etc.

What is recommendation, should stateful sets be used with the service mesh? I feel like the transparent proxy design does not consider stateful sets correctly, for example the issues mentioned in #932

lkysow:
> It that a good idea? 

Currently Consul only supports one service pointing at one set of pods. If you have two services that point at the same set of pods then none of the pods will start. In addition, when ACLs are enabled the `serviceAccountName` of the pod must match the service name otherwise the pods won't start. So our plan to ignore services that don't match the serviceAccountName simply means that you won't get in a broken situation. Those services would never have worked before, they would have caused the pods to not come up.

RE statefulsets we definitely want to support them. We need to triage and get to your issue.

nflaig:
>  statefulsets we definitely want to support them. We need to triage and get to your issue

ok  thanks, that is definitely good to hear and of course this is not an easy problem to solve. The work you guys are doing is great and I really appreciate it

PR #859, opened 2021-11-11
Flatten partitions tests into a single test

thisisnotashwin:
Changes proposed in this PR:
- Move the "in-partition" acceptance tests and "x-partition" acceptance tests into a single test to enable faster test times.

How I've tested this PR:
- running it in CI 🤞 
- The above has been tested and the test runtime has decreased from ~42 mins to ~28 mins.

- If the acceptance tests go 🟢  and the total time for running them is under 43 minutes for the partitions test, hit the like, comment, subscribe and merge button!!

### DIFF
--- a/acceptance/tests/partitions/partitions_test.go
+++ b/acceptance/tests/partitions/partitions_test.go
@@ -23,435 +23,8 @@ const staticServerName = "static-server"
 const staticServerNamespace = "ns1"
 const staticClientNamespace = "ns2"
 
-// Test that Connect works in a default installation.
-// i.e. without ACLs because TLS is required for setting up Admin Partitions.
-func TestPartitionsWithoutMesh(t *testing.T) {
-	env := suite.Environment()
-	cfg := suite.Config()
-
-	if !cfg.EnableEnterprise {
-		t.Skipf("skipping this test because -enable-enterprise is not set")
-	}
-
-	if !cfg.UseKind {
-		t.Skipf("skipping this test because Admin Partition tests are only supported in Kind for now")
-	}
-
-	if cfg.EnableTransparentProxy {
-		t.Skipf("skipping this test because -enable-transparent-proxy is true")
-	}
-
-	const defaultPartition = "default"
-	const secondaryPartition = "secondary"
-	const defaultNamespace = "default"
-	cases := []struct {
-		name                 string
-		destinationNamespace string
-		mirrorK8S            bool
-		secure               bool
-	}{
-		{
-			"default namespace",
-			defaultNamespace,
-			false,
-			false,
-		},
-		{
-			"default namespace; secure",
-			defaultNamespace,
-			false,
-			true,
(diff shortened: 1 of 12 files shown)


acceptance/tests/partitions/partitions_test.go
```diff
@@ -786,170 +359,280 @@ func TestPartitionsWithMesh(t *testing.T) {
 			helpers.Cleanup(t, cfg.NoCleanupOnFailure, func() {
 				k8s.KubectlDeleteK(t, clientClusterContext.KubectlOptions(t), kustomizeDir)
 			})
+			// This section of the tests run the in-partition networking tests.
+			t.Cleanup(func() {
```
ishustava, line 363: why is it part of the cleanup?

acceptance/tests/partitions/partitions_test.go
```diff
@@ -786,170 +359,280 @@ func TestPartitionsWithMesh(t *testing.T) {
...
+				// Client cluster.
+				services, _, err = consulClient.Catalog().Service(staticServerName, "", serverQueryClientOpts)
+				require.NoError(t, err)
+				require.Len(t, services, 1)
+
+				services, _, err = consulClient.Catalog().Service(staticClientName, "", clientQueryClientOpts)
+				require.NoError(t, err)
+				require.Len(t, services, 1)
+
+				defaultPartitionExports := &api.PartitionExportsConfigEntry{
```
ishustava, line 533: could we make it a CRD and put it in the fixture? I think this should be the workflow that we think users on k8s will use.

acceptance/tests/partitions/partitions_test.go
thisisnotashwin: If I didnt not put it inside the cleanup block, the resources from this run would hang around and would could the x-partition tests to fail.

acceptance/tests/partitions/partitions_test.go
thisisnotashwin: good call. yeah ill create a CRD fixture for this!

acceptance/tests/partitions/partitions_test.go
ishustava: 🤔  Now it seems that most of our test logic lives in cleanup functions, so I'm wondering if we can reorder the test somehow to make it work?

thisisnotashwin: I thought the job of the cleanup functions was to invoke a cleanup when we reached the end 🤔 
I can think about this some more

t-eckert (approved):

acceptance/tests/partitions/partitions_test.go
thisisnotashwin: swapping this for a `t.Run()` worked and I like it.

ishustava (approved):
🎉

PR #872, opened 2021-11-22
Add configuration for the vault Connect CA provider

ishustava:
Changes proposed in this PR:
- Adds helm values to configure vault connect ca provider
- Uses k8s auth method to configure connect CA provider in consul (hashicorp/consul#11573)
- This PR does not work if vault is configured with TLS (this change will come as a separate PR)

How I've tested this PR:
acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/framework/helpers/helpers.go
+++ b/acceptance/framework/helpers/helpers.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/gruntwork-io/terratest/modules/helm"
 	"os"
 	"os/signal"
 	"strings"
 	"syscall"
 	"testing"
 	"time"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
 
 	terratestk8s "github.com/gruntwork-io/terratest/modules/k8s"
 	"github.com/gruntwork-io/terratest/modules/random"
@@ -85,6 +86,7 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
 	retry.RunWith(counter, t, func(r *retry.R) {
 		pods, err := client.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{LabelSelector: podLabelSelector})
 		require.NoError(r, err)
+		require.NotEmpty(t, pods.Items)
 
 		var notReadyPods []string
 		for _, pod := range pods.Items {
--- a/acceptance/framework/helpers/helpers.go
+++ b/acceptance/framework/helpers/helpers.go
@@ -86,7 +86,7 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
 	retry.RunWith(counter, t, func(r *retry.R) {
 		pods, err := client.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{LabelSelector: podLabelSelector})
 		require.NoError(r, err)
-		require.NotEmpty(t, pods.Items)
+		require.NotEmpty(r, pods.Items)
 
 		var notReadyPods []string
 		for _, pod := range pods.Items {
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -20,6 +20,8 @@ path "consul/data/secret/gossip" {
   capabilities = ["read"]
 }`
 
+	// connectCAPolicy allows Consul to bootstrap all certificates for the service mesh in Vault.
+	// Adapted from https://www.consul.io/docs/connect/ca/vault#consul-managed-pki-paths.
 	connectCARules = `
 path "/sys/mounts" {
   capabilities = [ "read" ]
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -43,9 +43,8 @@ path "/connect_inter/*" {
 `
 )
 
-// todo: update
-// Installs Vault, bootstraps it with secrets, policies, and Kube Auth Method
-// then creates a gossip encryption secret and uses this to bootstrap Consul.
+// TestVault installs Vault, bootstraps it with secrets, policies, and Kube Auth Method.
+// It then configures Consul to use vault as the backend and checks that it works.
 func TestVault(t *testing.T) {
 	cfg := suite.Config()
 	ctx := suite.Environment().DefaultContext(t)
@@ -102,9 +101,8 @@ func TestVault(t *testing.T) {
 	}
 	_, err = vaultClient.Logical().Write("consul/data/secret/gossip", params)
 	require.NoError(t, err)
-
 	consulHelmValues := map[string]string{
-		"global.image": "ishustava/consul-dev@sha256:a873b4aa28f7511628281e852daab4c04918872bb86023cefb6197acabb43426",
+		"global.image": "docker.mirror.hashicorp.services/hashicorpdev/consul:latest",
 
 		"server.enabled":  "true",
 		"server.replicas": "1",
@@ -134,8 +132,8 @@ func TestVault(t *testing.T) {
 	consulClient := consulCluster.SetupConsulClient(t, true)
 	keys, err := consulClient.Operator().KeyringList(nil)
 	require.NoError(t, err)
-	// There should only be one key since there is only 1 dc.
-	require.Len(t, keys, 1)
+	// There are two identical keys for LAN and WAN since there is only 1 dc.
+	require.Len(t, keys, 2)
 	require.Equal(t, 1, keys[0].PrimaryKeys[gossipKey])
 
 	// Confirm that the Vault Connect CA has been bootstrapped correctly.
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -138,6 +138,7 @@ func TestVault(t *testing.T) {
 
 	// Confirm that the Vault Connect CA has been bootstrapped correctly.
 	caConfig, _, err := consulClient.Connect().CAGetConfig(nil)
+	require.NoError(t, err)
 	require.Equal(t, caConfig.Provider, "vault")
 
 	// Deploy two services and check that they can talk to each other.
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -15,14 +15,14 @@ import (
 )
 
 const (
-	gossipRules = `
+	gossipPolicy = `
 path "consul/data/secret/gossip" {
   capabilities = ["read"]
 }`
 
 	// connectCAPolicy allows Consul to bootstrap all certificates for the service mesh in Vault.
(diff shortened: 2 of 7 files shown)


acceptance/tests/vault/vault_test.go
```diff
@@ -4,34 +4,48 @@ import (
...
 	"github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/vault"
 	"github.com/stretchr/testify/require"
-	"testing"
 )
 
-// generateGossipSecret generates a random 32 byte secret returned as a base64 encoded string.
-func generateGossipSecret() (string, error) {
```
ishustava, line 16: I moved this to the bottom to make it a bit easier to see the test

acceptance/tests/vault/vault_test.go
```diff
@@ -4,34 +4,48 @@ import (
...
 
-	key := make([]byte, 32)
-	n, err := rand.Reader.Read(key)
-	if err != nil {
-		return "", fmt.Errorf("error reading random data: %s", err)
-	}
-	if n != 32 {
-		return "", fmt.Errorf("couldn't read enough entropy")
-	}
+	connectCARules = `
```
kschoche, line 23: Now that I'm reading this I wonder if `rules` is the appropriate word or if we should swap everything to `policies`.
No need to make any changes, I can do those changes in my next PR if we decide it makes sense.
Thoughts?

acceptance/tests/vault/vault_test.go
```diff
@@ -113,6 +132,49 @@ path "consul/data/secret/gossip" {
 	consulClient := consulCluster.SetupConsulClient(t, true)
 	keys, err := consulClient.Operator().KeyringList(nil)
 	require.NoError(t, err)
-	// We use keys[0] because KeyringList returns a list of keyrings for each dc, in this case there is only 1 dc.
+	// There are two identical keys for LAN and WAN since there is only 1 dc.
```
kschoche, line 135: Nice cleanup :)

charts/consul/templates/server-config-configmap.yaml
```diff
@@ -11,6 +11,36 @@ metadata:
...
+              }
+            }
+          ],
+          "ca_provider": "vault"
+        }
+      ]
+    }
+  {{- if .connectCA.additionalConfig }}
+  additional-connect-ca-config.json: |
+{{ tpl .connectCA.additionalConfig $ | trimAll "\"" | indent 4 }}
```
kschoche, line 40: It might be my browser but is the indentation off on this line?

acceptance/tests/vault/vault_test.go
ishustava: yeah, policy seems better!

charts/consul/templates/server-config-configmap.yaml
ishustava: It's not your browser 😄 I think it needs to be at the beginning of the line so that `indent 4` will indent it correctly. It the same as extraConfig below.

acceptance/framework/helpers/helpers.go
```diff
@@ -85,6 +86,7 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
 	retry.RunWith(counter, t, func(r *retry.R) {
 		pods, err := client.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{LabelSelector: podLabelSelector})
 		require.NoError(r, err)
+		require.NotEmpty(r, pods.Items)
```
kschoche, line 89: Nice catch, I think this might be a fix for a bug I'd been seeing! ❤️

kschoche (approved):
Fantastic work @ishustava!!

I just had one question which is not a blocker...

I see that you've added policies against `sys/mounts` and a bunch of others based on the consul.io page you linked (thx!), and it looks like this actually allows Consul to create/manage the vault pieces (which explains all the policy paths).

There is a second method where you enable the PKI paths from vault instead of consul like this, I'll be using it for Server-TLS and we could extend the approach to connect-ca as well: 
```
      // Enable the PKI Secrets engine.
       err = v.vaultClient.Sys().Mount("pki", &vapi.MountInput{
               Type:   "pki",
               Config: vapi.MountConfigInput{},
       })
       if err != nil {
               t.Fatal("unable to mount pki secrets engine to pki", "err", err)
       }
```

What I'm wondering is if this is a reasonable/realistic approach to expect that users would allow Consul to manage the vault paths, or would users more likely be expecting to do this from Vault instead of Consul. Does it really matter since it's just acceptance tests? (probably not). Thoughts?

charts/consul/values.yaml
```diff
@@ -152,6 +152,43 @@ global:
       # and check the name of `metadata.name`.
       consulClientRole: ""
 
+      # Configuration for the Vault Connect CA provider.
+      # The provider will be configured to use the Vault Kubernetes auth method
+      # and therefore requires the role provided by `consulServerRole`
+      # needs to have permissions to the root and intermediate PKI paths.
```
ishustava, line 158:
```suggestion
      # to have permissions to the root and intermediate PKI paths.
```

ishustava:
>There is a second method where you enable the PKI paths from vault instead of consul like this, I'll be using it for Server-TLS and we could extend the approach to connect-ca as well:

Ah that makes sense! yeah feel free to change this policy if we don't need those paths anymore once we have the PKI mounted.

acceptance/tests/vault/vault_test.go
```diff
@@ -48,16 +64,15 @@ func TestVault_BootstrapConsulGossipEncryptionKey(t *testing.T) {
...
+	logger.Log(t, "Creating policies")
+	err := vaultClient.Sys().PutPolicy("consul-gossip", gossipPolicy)
+	require.NoError(t, err)
+
+	err = vaultClient.Sys().PutPolicy("connect-ca", connectCAPolicy)
 	require.NoError(t, err)
 
-	// Create the Auth Roles for consul-server + consul-client.
-	logger.Log(t, "Creating the consul-server and consul-client-roles")
+	// Create the Auth Roles for consul-server and consul-client.
```
ndhanushkodi, line 74: Could we say "Create the Auth Roles for consul-server and consul-client which enables ___"? I'm not sure what auth roles do so maybe it'd be nice to add a tiny summary in the comment?

acceptance/tests/vault/vault_test.go
```diff
@@ -88,17 +103,23 @@ path "consul/data/secret/gossip" {
...
+		"controller.enabled":    "true",
 
 		"global.secretsBackend.vault.enabled":          "true",
 		"global.secretsBackend.vault.consulServerRole": "consul-server",
 		"global.secretsBackend.vault.consulClientRole": "consul-client",
 
+		"global.secretsBackend.vault.connectCA.address":             fmt.Sprintf("http://%s-vault:8200", vaultReleaseName),
+		"global.secretsBackend.vault.connectCA.rootPKIPath":         "connect_root",
+		"global.secretsBackend.vault.connectCA.intermediatePKIPath": "connect_inter",
+
```
ndhanushkodi, line 127: why are the connectCA keys under global.secretsBackend.vault but the gossip stuff is under global.gossipEncryption? Were we just trying to respect the previously used values for the gossipEncryption configuration? I guess I found it slightly unintuitive that you configure different vault things under different keys but this is not blocking the PR, just wanted to clarify if this is the case. cc @kschoche

acceptance/tests/vault/vault_test.go
```diff
@@ -88,17 +103,23 @@ path "consul/data/secret/gossip" {
 	}
```
ndhanushkodi, line 108: why don't we need to generate anything for the connect ca like we generated the gossip secret? is the only info vault needs in this case just the address and two paths and it will generate the certs itself? maybe its worth a comment here but also if this is just something that would be documented in user facing docs that's fine too.

charts/consul/templates/server-config-configmap.yaml
```diff
@@ -11,6 +11,36 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 data:
+  {{- $vaultConnectCAEnabled := and .Values.global.secretsBackend.vault.connectCA.address .Values.global.secretsBackend.vault.connectCA.rootPKIPath .Values.global.secretsBackend.vault.connectCA.intermediatePKIPath -}}
+  {{- if and .Values.global.secretsBackend.vault.enabled $vaultConnectCAEnabled }}
```
ndhanushkodi, line 15: so beautifully readable!!

charts/consul/values.yaml
```diff
@@ -152,6 +152,43 @@ global:
       # and check the name of `metadata.name`.
       consulClientRole: ""
 
+      # Configuration for the Vault Connect CA provider.
+      # The provider will be configured to use the Vault Kubernetes auth method
+      # and therefore requires the role provided by `consulServerRole`
```
ndhanushkodi, line 157:
```suggestion
      # and therefore requires the role provided by `secretsBackend.vault.consulServerRole`
```

ndhanushkodi (approved):
Really awesome work!! This is the first vault PR I'm reviewing so I just had some comments around where adding some clarifications could be nice based on where I wasn't sure of something. No worries if a more appropriate place to address some of this would be in official documentation though. I really love how the acceptance test kind of self documents for even someone like me coming back to looking at this project after a long time!

acceptance/tests/vault/vault_test.go
ishustava: Yeah, great call-out! I've added more info to this comment. Let me know if that makes sense!

acceptance/tests/vault/vault_test.go
ishustava: >is the only info vault needs in this case just the address and two paths and it will generate the certs itself?
Yep, except it's Consul that will generate it for us (as long as the policy allows it to do so).

From https://www.consul.io/docs/connect/ca/vault#root-and-intermediate-pki-paths:

> If either path does not exist, then Consul will attempt to mount and initialize it. This requires additional privileges by the Vault token in use. If the paths already exist, Consul will use them as configured.

acceptance/tests/vault/vault_test.go
ishustava: Good question. I actually suggested that we use existing configuration for gossip secret to make a consistent regardless of whether you use k8s secrets or vault secrets. I think this allows us to re-use existing configuration that is for the same purpose rather than duplicate it for vault specifically. So in this case, gossip key configuration is just a reference to some secret which could be in k8s or in vault.

Connect CA config is a bit different. First, it doesn't require any secrets since Consul can create all those secrets in Vault as long as the policy allows. Second, this configuration, unlike gossip key config, is specific to vault connect CA provider and so these options only make sense under `secretsBackend.vault`

ndhanushkodi: that makes sense, thank you!!

PR #874, opened 2021-11-23
Support Vault server running with TLS

ishustava:
Changes proposed in this PR:
- Change vault cluster in acceptance tests to only run with TLS. All tests will run against vault with TLS because that is the use case we think will be the most valuable for users to test
- Support adding Vault CA as a secret to pods that will be using vault agent. We need to add two annotations to pods:
   - `vault.hashicorp.com/agent-extra-secret` with the value of the vault CA secret name. The secret will be mounted to vault agent at `/vault/custom` path. See docs [here](https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-extra-secret)
   - `vault.hashicorp.com/ca-cert` - with the path of the ca file inside the vault agent container. This should be `/vault/custom/<secret key>`
- Most pods will only need those annotations. The server pods also need the Vault CA secret to be mounted as a volume because it needs the CA to be on the file system for the vault connect CA provider.

How I've tested this PR:
acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


acceptance/framework/vault/vault_cluster.go
```diff
@@ -75,35 +66,34 @@ func NewVaultCluster(
...
 	return &VaultCluster{
 		ctx:                ctx,
-		vaultHelmOptions:   vaultHelmOpts,
+		helmOptions:        vaultHelmOpts,
 		kubectlOptions:     kopts,
-		namespace:          cfg.KubeNamespace,
-		values:             helmValues,
 		kubernetesClient:   ctx.KubernetesClient(t),
-		kubeConfig:         cfg.Kubeconfig,
-		kubeContext:        cfg.KubeContext,
```
ishustava, line 84: the once that are deleted were unused.

acceptance/framework/vault/vault_cluster.go
```diff
@@ -13,55 +14,45 @@ import (
...
 
 // NewVaultCluster creates a VaultCluster which will be used to install Vault using Helm.
-func NewVaultCluster(
-	t *testing.T,
-	helmValues map[string]string,
-	ctx environment.TestContext,
-	cfg *config.TestConfig,
-	releaseName string,
-) *VaultCluster {
+func NewVaultCluster(t *testing.T, ctx environment.TestContext, cfg *config.TestConfig, releaseName string) *VaultCluster {
```
ishustava, line 48: `helmValues` was unused

acceptance/framework/helpers/helpers.go
```diff
@@ -77,7 +77,7 @@ func CheckForPriorInstallations(t *testing.T, client kubernetes.Interface, optio
 func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespace, podLabelSelector string) {
 	t.Helper()
 
-	logger.Log(t, "Waiting for pods to be ready.")
+	logger.Logf(t, "Waiting for pods with label %q to be ready.", podLabelSelector)
```
kschoche, line 80: 👏🏽

acceptance/framework/vault/vault_cluster.go
```diff
@@ -122,7 +112,9 @@ func (v *VaultCluster) SetupVaultClient(t *testing.T) *vapi.Client {
 		tunnel.Close()
 	})
 
-	config.Address = fmt.Sprintf("http://127.0.0.1:%d", localPort)
+	config.Address = fmt.Sprintf("https://127.0.0.1:%d", localPort)
+	err := config.ConfigureTLS(&vapi.TLSConfig{Insecure: true})
```
kschoche, line 116: Sanity check : is `Insecure: true` correct?

acceptance/framework/vault/vault_cluster.go
```diff
@@ -146,18 +137,28 @@ func (v *VaultCluster) bootstrap(t *testing.T, ctx environment.TestContext) {
...
 	}
-	// We need to kubectl exec this one on the vault server:
-	// This is taken from https://learn.hashicorp.com/tutorials/vault/kubernetes-google-cloud-gke?in=vault/kubernetes#configure-kubernetes-authentication
-	cmdString := fmt.Sprintf("VAULT_TOKEN=%s vault write auth/kubernetes/config disable_iss_validation=\"true\" token_reviewer_jwt=\"$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" kubernetes_host=\"https://${KUBERNETES_PORT_443_TCP_ADDR}:443\" kubernetes_ca_cert=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt", vaultRootToken)
 
 	v.logger.Logf(t, "updating vault kube auth config")
-	k8s.RunKubectl(t, ctx.KubectlOptions(t), "exec", "-i", fmt.Sprintf("%s-vault-0", v.vaultReleaseName), "--", "sh", "-c", cmdString)
+
+	// To configure the auth method, we need to read the token and the ca cert from the Vault's server
+	// service account token.
```
kschoche, line 149: Nice change! I had meant to get back around to this :)

acceptance/framework/vault/vault_cluster.go
```diff
@@ -171,34 +172,174 @@ func (v *VaultCluster) Create(t *testing.T, ctx environment.TestContext) {
...
+		},
+		Data: map[string][]byte{
+			corev1.TLSCertKey: []byte(caPem),
+		},
+		Type: corev1.SecretTypeOpaque,
+	}, metav1.CreateOptions{})
+	require.NoError(t, err)
+}
+
+func (v *VaultCluster) initAndUnseal(t *testing.T) {
```
kschoche, line 302: OK - so I take it TLS is not supported in dev mode?

acceptance/framework/vault/vault_cluster.go
```diff
@@ -171,34 +172,174 @@ func (v *VaultCluster) Create(t *testing.T, ctx environment.TestContext) {
...
+
+		v.vaultClient = v.SetupVaultClient(t)
+
+		initResp, err := v.vaultClient.Sys().Init(&vapi.InitRequest{
+			SecretShares:    1,
+			SecretThreshold: 1,
+		})
+		require.NoError(r, err)
+		v.vaultClient.SetToken(initResp.RootToken)
+
```
kschoche, line 321: Wondering if we can put just a couple more comments in this function for others to follow, this all looks super familiar to me because I've done it before ;-)

```suggestion
		// KeysB64[0] is the Unseal key that was created on `Sys().Init()`.
```

acceptance/framework/vault/vault_cluster.go
```diff
@@ -171,34 +172,174 @@ func (v *VaultCluster) Create(t *testing.T, ctx environment.TestContext) {
...
+			Name:      rootTokenSecret,
+			Namespace: namespace,
+		},
+		Data: map[string][]byte{
+			"token": []byte(v.vaultClient.Token()),
+		},
+		Type: corev1.SecretTypeOpaque,
+	}, metav1.CreateOptions{})
+	require.NoError(t, err)
+}
```
kschoche, line 345: I'm not entirely sure why we need to do this for the test since we're storing it in the vault client, is someone else going to be using the vault root token? that seems strange so I must be missing something :)

acceptance/framework/vault/vault_cluster.go
ishustava: This is for local traffic, and so we don't need to verify TLS for that.

kschoche: Ah! makes sense, thanks!

acceptance/framework/vault/vault_cluster.go
ishustava: yep, unfortunately there's no way to make it work with the dev server

acceptance/framework/vault/vault_cluster.go
ishustava: The reason i was thinking about saving it is when you run it locally with -no-cleanup-on-failure. Otherwise, if run tests with that flag, you will have no access to the vault cluster and so no ability to debug.

acceptance/framework/vault/vault_cluster.go
ishustava: Added

t-eckert (approved):
This looks great! Apologies for taking so long to get back to you on this review.

kschoche (approved):
Amazing work @ishustava thanks for getting to this!!!!!!!!!!!

PR #878, opened 2021-11-24
Bump to go 1.17

ndhanushkodi:
Changes proposed in this PR:
- See commit messages, in order. (ignore empty commit)
- related release pipeline PR: https://github.com/hashicorp/consul-k8s-releases/pull/8

How I've tested this PR:
- The pipeline build steps and acceptance tests pass

- review

Checklist:
- [ ] CHANGELOG entry added 


control-plane/catalog/to-consul/consul_node_services_client_ent_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build enterprise
```
kschoche, line 1: I'm sure there is a reason but why do we need the duplicate enterprise build declaration? Is the old one left for backward compatibility?

ndhanushkodi: This stack overflow was good context on how they interplay-- https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives?noredirect=1&lq=1

We can leave both which in theory would be backwards compatible but we have bumped the min go version to 1.17 anyways so we could probably just remove the old `// +build` lines. I think that's probably the cleanest thing to do, does that sound good?

kschoche: Thanks for the info and link!!

I was thinking that if we're bumping the min go version there would be no need for back compatibility lines? If its too much work there's very little to be gained from it, but might be cleaner. Will defer to to your judgement!

acceptance/go.mod
```diff
@@ -12,3 +12,61 @@ require (
 	k8s.io/apimachinery v0.19.3
 	k8s.io/client-go v0.19.3
 )
+
+require (
```
t-eckert, line 16: Could we merge these require blocks?

cli/go.mod
```diff
@@ -1,35 +1,164 @@
...
 	helm.sh/helm/v3 v3.6.1
 	k8s.io/api v0.21.2
 	k8s.io/apimachinery v0.21.2
 	k8s.io/cli-runtime v0.21.0
 	k8s.io/client-go v0.21.2
-	rsc.io/letsencrypt v0.0.3 // indirect
 	sigs.k8s.io/yaml v1.2.0
 )
 
+require (
```
t-eckert, line 25: Could we merge these require blocks?

control-plane/go.mod
```diff
@@ -40,4 +27,107 @@ require (
 	sigs.k8s.io/controller-runtime v0.10.2
 )
 
-go 1.16
+require (
```
t-eckert, line 30: Here too, please?

t-eckert (approved):
Looks great! I just added the same note three times about merging require blocks. I don't know if there is a reason they should be separate. If so, feel free to ignore.

acceptance/go.mod
ndhanushkodi: go 1.17 groups the require blocks as indirect and direct, and since this is generated I think we should leave it as it was generated via `go mod tidy`

control-plane/catalog/to-consul/consul_node_services_client_ent_test.go
ndhanushkodi: yup I agree!

kschoche (approved):
Great stuff, nothing blocking from me, thanks for taking care of this one!

Does it make sense to also update the CONTRIBUTNG.md? It currently says `go-1.11.4+` :-)

ndhanushkodi:
> Does it make sense to also update the CONTRIBUTNG.md? It currently says go-1.11.4+ :-)

yup definitely, thanks!!!

acceptance/go.mod
t-eckert: Thanks for explaining!

PR #882, opened 2021-11-29
add cd chart_dir so tests can be run from root level

kschoche:
Changes proposed in this PR:
- Add `cd chart_dir` to the ui-ingress tests that were missing it so they can be run cleanly from the root.
-

How I've tested this PR:
run bats test

code review

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


thisisnotashwin (approved):
💥

PR #887, opened 2021-12-01
Add a workaround to check that the ACL token is replicated to other Consul servers

### LINKED ISSUE #862: 403 (ACL not found) followed by successful deployment
### Community Note

---

### Overview of the Issue

_(this issue looks very similar to the one I posted [before](https://github.com/hashicorp/consul-k8s/issues/844) though the previous issue apparently revolved around two different problems -- one being incorrect service which prevented registration whilst the ACL issue remained)_

I am running a **private** kubernetes cluster on GKE. I deployed consul (app version `1.10.3`, chart version `0.36.0`) and I was able to access consul ui. I can see the servers and client agents (on each node) running. I disregard errors during bootstrap as these are most likely more warnings than errors, stating that something is not ready yet.

Every time I try to add a new pod with `connect-inject`, I get a bunch of `Unexpected response code: 403 (ACL not found)`
 errors followed by successful(?) registration. I wonder actually whether this deployment should be considered successful knowing that this ACL check probably timed-out after 30 retries yet all checks are green.

The application is responsive and I can reach it from other pods, yet the pod deployment is slowed down by those ACL checks and **surely something is not right** so I'd rather not keep it that way :)

### Reproduction Steps

After bootstrapping consul cluster (with `manageSystemACLs` set to `true`) I tried to deploy `hashicorp/http-echo:latest` pod with `'consul.hashicorp.com/connect-inject': 'true'` annotation (see yaml at the bottom of this issue).

### Logs

<details>
  <summary>Logs</summary>

```
[INFO]  Consul login complete
[ERROR] Unable to get Agent services: error="Unexpected response code: 403 (ACL not found)"
[ERROR] Unable to get Agent services: error="Unexpected response code: 403 (ACL not found)"
[ERROR] Unable to get Agent services: error="Unexpected response code: 403 (ACL not found)"
(..repeat 30 times..)
[INFO]  Registered service has been detected: service=static-server-xyz
[INFO]  Registered service has been detected: service=static-server-xyz-sidecar-proxy
[INFO]  Connect initialization completed
        Successfully applied traffic redirection rules
```

</details>

### Expected behavior

Those errors should not be present.

### Environment details

**values.yaml**

```yaml
global:
  image: "consul:1.10.3"
  datacenter: gcp
  recursors:
    - '1.1.1.1'
    - '8.8.8.8'
  acls:
    manageSystemACLs: true
  tls:
    enabled: false

server:
  replicas: 3

connectInject:
  enabled: true
  default: true
  replicas: 1

ui:
  metrics:
    enabled: false
```

**sample.yaml**

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: static-server-xyz
---
apiVersion: v1
kind: Service
metadata:
  name: static-server-xyz
spec:
  selector:
    app: static-server-xyz
  ports:
    - port: 80
      targetPort: 8080
---
apiVersion: v1
kind: Pod
metadata:
  name: static-server-xyz
  labels:
    app: static-server-xyz
  # annotations:
  #   'consul.hashicorp.com/connect-inject': 'true'
spec:
  containers:
    # This name will be the service name in Consul.
    - name: static-server-xyz
      image: hashicorp/http-echo:latest
      args:
        - -text="hello world"
        - -listen=:8080
      ports:
        - containerPort: 8080
          name: http
  # If ACLs are enabled, the serviceAccountName must match the Consul service name.
  serviceAccountName: static-server-xyz
```

ishustava:
Fixes #862

A consul client may reach out to a follower instead of a leader to resolve the token during the
call to get services. This is because clients talk to servers in the stale consistency mode
to decrease the load on the servers (see https://www.consul.io/docs/architecture/consensus#stale).
In that case, it's possible that the token isn't replicated
to that server instance yet. The client will then get an "ACL not found" error
and subsequently cache this not found response. Then our call
to get services from the agent will keep hitting the same "ACL not found" error
until the cache entry expires (determined by the `acl_token_ttl` which defaults to 30 seconds).
This is not great because it will delay app start up time by 30 seconds in most cases
(if you are running 3 servers, then the probability of ending up on a follower is close to 2/3).

To help with that, we try to first read the token in the stale consistency mode until we
get a successful response. This should not take more than 100ms because raft replication
should in most cases take less than that (see https://www.consul.io/docs/install/performance#read-write-tuning)
but we set the timeout to 2s to be sure.

Note though that this workaround does not eliminate this problem completely. It's still possible
for this call and the next call to reach different servers and those servers to have different
states from each other.
For example, this call can reach a leader and succeed, while the call below can go to a follower
that is still behind the leader and get an "ACL not found" error.
However, this is a pretty unlikely case because
clients have sticky connections to a server, and those connections get rebalanced only every 2-3min.
And so, this workaround should work in a vast majority of cases.

How I've tested this PR:
- unit tests
- manually in a cluster with 3 servers

- code review

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ IMPROVEMENTS:
 * CLI
    * Pre-check in the `install` command to verify the correct license secret exists when using an enterprise Consul image. [[GH-875](https://github.com/hashicorp/consul-k8s/pull/875)]
 
+BUG FIXES:
+* Control Plane:
+   * Add a workaround to check that the ACL token is replicated to other Consul servers. [[GH-862](https://github.com/hashicorp/consul-k8s/issues/862)]
+
 ## 0.37.0 (November 18, 2021)
 
 BREAKING CHANGES:
--- a/control-plane/subcommand/connect-init/command.go
+++ b/control-plane/subcommand/connect-init/command.go
@@ -26,6 +26,9 @@ const (
 	numLoginRetries = 3
 	// The number of times to attempt to read this service (120s).
 	defaultServicePollingRetries = 120
+
+	raftReplicationTimeout   = 2 * time.Second
+	tokenReadPollingInterval = 100 * time.Millisecond
 )
 
 type Command struct {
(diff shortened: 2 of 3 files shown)


t-eckert (approved):
Spectacular! I always learn something new reading through your code.

lkysow (approved):

PR #888, opened 2021-12-02
add a makefile target for parallel bats

kschoche:
Changes proposed in this PR:
- Adds a makefile target to run bats tests in parallel. This shaves off about 5 minutes from a full run of the bats tests.
- This is the same command we run in CI.

How I've tested this PR:
`make bats-tests` passes.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


lkysow (approved):
🎉

PR #892, opened 2021-12-02
Fail if partitions enabled with federation

thisisnotashwin:
Changes proposed in this PR:
- Fail if Admin Partitions and WAN federation are both enabled as they are mutually exclusive.
- Fixup changelog.

How I've tested this PR:
- BATS

- 👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/client-daemonset.yaml
+++ b/charts/consul/templates/client-daemonset.yaml
@@ -2,6 +2,7 @@
 {{- if (and (and .Values.global.tls.enabled .Values.global.tls.httpsOnly) (and .Values.global.metrics.enabled .Values.global.metrics.enableAgentMetrics))}}{{ fail "global.metrics.enableAgentMetrics cannot be enabled if TLS (HTTPS only) is enabled" }}{{ end -}}
 {{- $serverEnabled := (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) -}}
 {{- if (and .Values.global.adminPartitions.enabled $serverEnabled (ne .Values.global.adminPartitions.name "default"))}}{{ fail "global.adminPartitions.name has to be \"default\" in the server cluster" }}{{ end -}}
+{{- if and .Values.global.federation.enabled .Values.global.adminPartitions.enabled }}{{ fail "If global.federation.enabled is true, global.adminPartitions.enabled must be false because they are mutually exclusive" }}{{ end }}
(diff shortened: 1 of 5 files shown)


charts/consul/templates/client-daemonset.yaml
```diff
@@ -2,6 +2,7 @@
 {{- if (and (and .Values.global.tls.enabled .Values.global.tls.httpsOnly) (and .Values.global.metrics.enabled .Values.global.metrics.enableAgentMetrics))}}{{ fail "global.metrics.enableAgentMetrics cannot be enabled if TLS (HTTPS only) is enabled" }}{{ end -}}
 {{- $serverEnabled := (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) -}}
 {{- if (and .Values.global.adminPartitions.enabled $serverEnabled (ne .Values.global.adminPartitions.name "default"))}}{{ fail "global.adminPartitions.name has to be \"default\" in the server cluster" }}{{ end -}}
+{{- if and .Values.global.federation.enabled .Values.global.adminPartitions.enabled }}{{ fail "If global.federation.enabled is true, global.adminPartitions.enabled must be false because they are mutually exclusive" }}{{ end }}
```
ishustava, line 5: do we need this check on the clients? I think federation is only possible if servers are also enabled

ishustava (approved):

charts/consul/templates/client-daemonset.yaml
thisisnotashwin: That is true. I put a check on clients in case someone was using a values file with federation set to enabled on a `non-default` admin partition.

lkysow: > I think federation is only possible if servers are also enabled

I think actually we require federation: true if doing HCP federation. I can double check.

thisisnotashwin: So we have a situation where we might have federation enabled and partitions enabled? Consul Core intends on disabling this. https://github.com/hashicorp/consul-enterprise/pull/1391
Am I confusing specifics?

lkysow: I'm saying that I think we need to keep this check here rather than on the server statefulset since potentially someone has:

```yaml
server:
  enabled: false
global:
  federation:
    enabled: true
```

Although that being said I haven't had time to find out if we actually do use this for HCP fed.

lkysow (approved):

PR #903, opened 2021-12-06
cli: allow enterprise images to be named without the string "-ent"

ndhanushkodi:
In a case where users might have a Consul enterprise image that may not
contain the string "-ent" in the name, we don't want to fail the
installation. This is a fairly common case, so we've removed the check.

How I've tested this PR:
unit tests

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/cli/cmd/install/install.go
+++ b/cli/cmd/install/install.go
@@ -3,11 +3,12 @@ package install
 import (
 	"errors"
 	"fmt"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"os"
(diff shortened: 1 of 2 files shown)


thisisnotashwin (approved):
💥

kschoche (approved):
🔥

t-eckert (approved):

NicoletaPopoviciu (approved):

lkysow (approved):
Getting in on the party 🎉

Issue #650, opened 2021-03-17
helm:Envoy sidecar shutting down too early causes requests to fail


nflaig:
Hey guys,

during a rolling update (pod termination) we are getting 502 errors in nginx for upstream requests. This happens because the load balancer in k8s still sends requests to the terminating pod due to the fact that the endpoint deregistration happens asynchronously, see https://github.com/kubernetes/kubernetes/issues/43576.

For nginx itself, I was able to resolve this race condition issue by adding a simple sleep to the `preStop` hook of the container
```
lifecycle:
  preStop:
    exec:
      command: ["/bin/sh", "-c", "sleep 5 && /usr/sbin/nginx -s quit"]
```

but the problem is not fully resolved since the envoy sidecar is still shutting down too early which causes requests to fail.

### Current behavior

Nginx returns 502 errors because the upstream requests are send through the envoy sidecar proxy which is already shutting down due to the `SIGTERM` send by k8s.

### Expected behavior 

The envoy sidecar proxy should handle requests of the proxied service as long as the service is still running, e.g. if the service does some cleanup and still needs to send data when terminating or in case of nginx if further requests are still being routed to the pod.

### Suggestion

In the [envoy preStop command](https://github.com/hashicorp/consul-k8s/blob/master/connect-inject/envoy_sidecar.go#L194) a delay could be added to prevent instantly sending the `SIGTERM` signal to the container. Since a hardcoded sleep duration seems to  be too static maybe it would make sense to add something like this 
```
while [ $(netstat -plunt | grep tcp | grep -v envoy | wc -l | xargs) -ne 0 ]; do sleep 1; done
```
It would delay the shutdown process of envoy until there are no more TCP listeners, i.e. the proxied service is no longer running and it is save to also shutdown envoy without causing further requests to fail.

Another option could be to allow the user of the helm chart to customize the envoy `preStop` hook.

### Environment details

- `consul-k8s` version: 0.24.0
- `consul-helm` version: 0.30.0
- `consul` version: 1.9.3
- `envoy` version: 1.16.0

### Related

- https://github.com/istio/istio/issues/7136
- https://github.com/istio/istio/issues/10112
- https://github.com/istio/istio/issues/7665
- https://github.com/istio/istio/issues/12183
- https://github.com/istio/istio/pull/11298
- https://github.com/istio/istio/pull/12342
[...]

PR #911, opened 2021-12-08
Add configurable preStop delay to Envoy sidecar

hamishforbes:
Changes proposed in this PR:
- Add a new annotation `consul.hashicorp.com/sidecar-proxy-prestop-delay`
- Pods with this annotation have a [Lifecycle hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) added to delay the Envoy proxy sidecar shutdown by the specified number of seconds

Should probably be some validation on the value of the annotation, atm this would allow random command injection to the sidecar container on shutdown

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

Related: #650

### DIFF
--- a/control-plane/connect-inject/annotations.go
+++ b/control-plane/connect-inject/annotations.go
@@ -69,6 +69,9 @@ const (
 	annotationSidecarProxyMemoryLimit   = "consul.hashicorp.com/sidecar-proxy-memory-limit"
 	annotationSidecarProxyMemoryRequest = "consul.hashicorp.com/sidecar-proxy-memory-request"
 
+	// annotationSidecarProxyPreStopDelay is the number of seconds to delay Envoy Sidecar shutdown
+	annotationSidecarProxyPreStopDelay = "consul.hashicorp.com/sidecar-proxy-prestop-delay"
+
 	// annotations for metrics to configure where Prometheus scrapes
 	// metrics from, whether to run a merged metrics endpoint on the consul
 	// sidecar, and configure the connect service metrics.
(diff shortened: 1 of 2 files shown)


david-yu:
Thanks for the PR. We'd like to support a more graceful way of shutting down Envoy so that if the is still running Envoy shuts down after the app is shut down first. We will this into consideration for future development of this feature. 

I'm curious is this workaround working well in your environment?

hamishforbes:
Yeah for sure, it's not a particularly elegant solution!

I've been running a [fork](https://github.com/hashicorp/consul-k8s/compare/v0.40.0...hamishforbes:upstreamConfig) with this patch and a couple of other tweaks in production since opening the PR though, works perfectly for me.

dschaaff:
I'd also note that I'm running a fork with the changes in this pull request. This feature really needs to get prioritized. There is no way to do rolling deployments without 5xx errors due to the envoy sidecar stopping before the appliation container completes.

narendrapatel:
We are also testing this changes as we have some very high traffic apps and any issues gets quickly noticed and reported. A solution to this feature would be really helpful.

alt-dima:
Can't understand, how Consul Connect/Mesh can be used in production without this?

natitomattis:
We also have the same issue, we created (this fork)[https://github.com/team-telnyx/consul-k8s/pull/3. ] since our version (0.45.0) is behind main. We validated that adding the preStop hook fixed the problem.

david-yu:
Closing as this is now resolved by https://github.com/hashicorp/consul-k8s/pull/2233 and will be released in upcoming patch releases for 1.0.x, 1.1.x and 1.2.x.

PR #925, opened 2021-12-10
initial support for API Gateway

nickethier:
Changes proposed in this PR:
- Add support for installing the api-gateway controller from the [Consul API Gateway project](https://github.com/hashicorp/consul-api-gateway).
- Extends server-acl-init to include a new token for the api-gateway controller

How I've tested this PR:

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/api-gateway-gatewayclass.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if (or (and (ne (.Values.apiGateway.enabled | toString) "-") .Values.apiGateway.enabled) (and (eq (.Values.apiGateway.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if .Values.apiGateway.managedGatewayClass.enabled }}
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: GatewayClass
+metadata:
+  name: consul-api-gateway
```
ishustava, line 6: This and GatewayClassConfig below would probably need to be prefixed by the `{{ template "consul.fullname" . }}` like we do for other resources

charts/consul/values.yaml
```diff
@@ -2458,6 +2458,111 @@ terminatingGateways:
...
+  # Image to use for the api-gateway-controller pods and gateway instances
+  # @type: string
+  image: null
+
+
+  # Override global log verbosity level for api-gateway-controller pods. One of "debug", "info", "warn", or "error".
+  # @type: string
+  logLevel: info
+  managedGatewayClass:
+    enabled: true
```
ishustava, line 2475: We typically provide docs for every single helm value.

charts/consul/values.yaml
```diff
@@ -2458,6 +2458,111 @@ terminatingGateways:
...
+    # ```yaml
+    # nodeSelector: |
+    #   beta.kubernetes.io/arch: amd64
+    # ```
+    #
+    # @type: string
+    nodeSelector: null
+
+    # This value defines the type of service created for gateways (e.g. LoadBalancer, ClusterIP)
+    serviceType: LoadBalancer
```
ishustava, line 2490: If the service is of type NodePort, would you ever need to also accept a user-provided node port? If so, it might be better to convert it a `service` stanza and then have `type` under it. We do something like this for example for admin partitions:

https://github.com/hashicorp/consul-k8s/blob/ca980cbec061dd08bb0534deaa2b3352a891924c/charts/consul/values.yaml#L46-L73

charts/consul/values.yaml
```diff
@@ -2458,6 +2458,111 @@ terminatingGateways:
...
+
+    # This value toggles if the gateway ports should be mapped to host ports
+    useHostPorts: false
```
ishustava, line 2493: I think this is not used?

charts/consul/values.yaml
```diff
@@ -2458,6 +2458,111 @@ terminatingGateways:
...
+      # service: |
+      # - external-dns.alpha.kubernetes.io/hostname
+      # ```
+      #
+      # @type: string
+      service: null
+
+    # Override global log verbosity level for ateway pods. One of "debug", "info", "warn", or "error".
+    # @type: string
+    logLevel: info
```
ishustava, line 2510: Is this a duplicate of `logLevel` on line 2473?

ishustava (commented):
Looking really good, Nick!! I'm excited about this work to be merged 🎉 

I'm leaving some comments I have so far mostly about our Helm chart conventions and also some questions/suggestions. I didn't look at the bats tests though.

charts/consul/templates/api-gateway-controller-service.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if (or (and (ne (.Values.apiGateway.enabled | toString) "-") .Values.apiGateway.enabled) (and (eq (.Values.apiGateway.enabled | toString) "-") .Values.global.enabled)) }}
+apiVersion: v1
+kind: Service
```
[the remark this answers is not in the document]
nickethier: Correct, it is just used so that the gateway instances can dial back to the controller.

charts/consul/templates/api-gateway-gatewayclass.yaml
nickethier: Maybe the class config,  but I'm not sure that the class should. We want the gateway class thats provided by consul-k8s to be the same across installations so our documentation can have a deterministic name to reference. When a gateway is defined by an operator is must reference the gateway class which I don't think there should ever be more than one of for `consul-api-gateway`.

charts/consul/templates/api-gateway-controller-clusterrole.yaml
```diff
@@ -0,0 +1,211 @@
...
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
nickethier: Since gateway certificates are provided as a secret we do need access to all secrets in any namespace. Potentially an operator could limit the scope if they have all certificates in a specific namespace for example.

charts/consul/templates/api-gateway-controller-clusterrole.yaml
```diff
@@ -0,0 +1,211 @@
+{{- if (or (and (ne (.Values.apiGateway.enabled | toString) "-") .Values.apiGateway.enabled) (and (eq (.Values.apiGateway.enabled | toString) "-") .Values.global.enabled)) }}
```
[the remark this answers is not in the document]
nickethier: Ah yeah. Since we default to not enabled. It doesn't really matter what the global value is. Good catch

charts/consul/templates/api-gateway-controller-clusterrole.yaml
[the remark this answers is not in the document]
ishustava: Ah got it! I thought it might be the certs, but wanted to clarify. Thanks!

>Potentially an operator could limit the scope if they have all certificates in a specific namespace for example.

I don't think you can limit a scope for a cluster role to a namespace (only a specific resource). To limit it to a single ns, you'd have to create a role instead in that namespace. Maybe something we can look into later, as we had users that don't have perms to create Cluster* RBAC objects, and so we might need to think of a solution for our other components as well.

charts/consul/templates/api-gateway-gatewayclass.yaml
ishustava: Ah that makes a lot of sense!

charts/consul/templates/api-gateway-controller-clusterrole.yaml
```diff
@@ -0,0 +1,211 @@
...
+  - watch
+- apiGroups:
+  - gateway.networking.k8s.io
+  resources:
+  - gatewayclasses/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - gateway.networking.k8s.io
+  resources:
```
kschoche, line 127: This is a super minor suggestion, but you could do something like this for resources under the same `apiGroups` that require the same `verbs`:

```
apiGroups:
  - gateway.networking.k8s.io
  resources: ["gateways", "httproutes", "gatewayclasses"] 
  verbs: 
   <snip>
```
It might be easier to read, might not be, but if you find that you have a bunch of duplicate verb requirements under the same apiGroup you could dry it up.

charts/consul/templates/api-gateway-gatewayclassconfig.yaml
```diff
@@ -0,0 +1,47 @@
...
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: api-gateway
+spec:
+  consul:
+    authentication:
+      {{- if .Values.global.acls.manageSystemACLs }}
+      managed: true
+      method: {{ template "consul.fullname" . }}-k8s-auth-method
```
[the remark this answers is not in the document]
nickethier: This is read by the api gateway controller. Managed means that the acl system is managed so we create a service account for each logical gateway that matches the name we register the service as so that each gateway instance can auth to consul via the configured ACL auth method. By default managed is false and the controller expects an `account` to be set to the name of a operator managed ServiceAccount.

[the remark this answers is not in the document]
kschoche: Awesome, thanks for the explanation!

charts/consul/templates/api-gateway-controller-clusterrole.yaml
nickethier: I agree on readability, however I'd like to keep the format here in sync with the role we have defined in our repo: https://github.com/hashicorp/consul-api-gateway/blob/main/config/rbac/role.yaml

I may take this suggestion though and apply it there at some point. Thanks!

charts/consul/values.yaml
nickethier: In this case no. It is driving the configuration of our GatewayClassConfig. The user provides the port in the Gateway definition. We simply have a toggle for if those ports should be mapped to node ports of the same value. 

Reference:
https://github.com/hashicorp/consul-api-gateway/blob/c82e707fb30bebf1a88980568e0f67e318f65496/pkg/apis/v1alpha1/types.go#L61-L67

charts/consul/values.yaml
nickethier: Nice catch 🙌🏼  I plumbed this correctly into the gatewayclassconfig

charts/consul/templates/api-gateway-controller-clusterrole.yaml
kschoche: Sounds good!!

control-plane/subcommand/server-acl-init/rules.go
```diff
@@ -138,6 +138,34 @@ partition_prefix "" {
 	return c.renderRules(anonTokenRulesTpl)
 }
 
+func (c *Command) apiGatewayControllerRules() (string, error) {
```
lkysow, line 141: needs corresponding update to rules_test.go

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -527,6 +531,19 @@ func (c *Command) Run(args []string) int {
...
+			c.log.Error("Error templating api gateway rules", "err", err)
+			return 1
+		}
+		err = c.createLocalACL("api-gateway-controller", apigwRules, consulDC, isPrimary, consulClient)
+		if err != nil {
+			c.log.Error(err.Error())
+			return 1
+		}
+	}
+
```
lkysow, line 546: Needs test updates to:
* `TestRun_TokensPrimaryDC`
* `TestRun_TokensReplicatedDC`
* `TestRun_TokensWithProvidedBootstrapToken`

Sorry we should have the "adding a new token" process documented somewhere.

ishustava (commented):
Looks good, just a couple of minor comments.

+1 To Luke's comment about the unit tests for the ACL token. 

I know we talked about adding an acceptance test, but I'm curious if you decided not to do it or do it in a separate PR?

nickethier:
Thanks @ishustava.

As far as acceptance test I had planned to follow up once we have release so there is a stable binary to reference from the test.

ishustava (approved):
Looks good 🎉  I'm approving assuming the unit tests Luke mentioned are added. All of my recent comments are non-blocking

charts/consul/templates/api-gateway-controller-clusterrolebinding.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.apiGateway.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "consul.fullname" . }}-api-gateway-controller-role-binding
```
thisisnotashwin, line 5: Hey @nickethier! I've been working on #993 and just making those updates here as well!
```suggestion
  name: {{ template "consul.fullname" . }}-api-gateway-controller
```

charts/consul/templates/api-gateway-controller-clusterrolebinding.yaml
```diff
@@ -0,0 +1,20 @@
...
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: api-gateway-controller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "consul.fullname" . }}-api-gateway-controller
+subjects:
+- kind: ServiceAccount
+  name: {{ template "consul.fullname" . }}-api-gateway-controller-svc-account
```
thisisnotashwin, line 18:
```suggestion
  name: {{ template "consul.fullname" . }}-api-gateway-controller
```

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -0,0 +1,148 @@
...
+        "vault.hashicorp.com/ca-cert": "/vault/custom/{{ .Values.global.secretsBackend.vault.ca.secretKey }}"
+        {{- end }}
+        {{- end }}
+      labels:
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: api-gateway-controller
+    spec:
+      serviceAccountName: {{ template "consul.fullname" . }}-api-gateway-controller-svc-account
```
thisisnotashwin, line 45:
```suggestion
      serviceAccountName: {{ template "consul.fullname" . }}-api-gateway-controller
```

charts/consul/templates/api-gateway-controller-serviceaccount.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if .Values.apiGateway.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "consul.fullname" . }}-api-gateway-controller-svc-account
```
thisisnotashwin, line 5:
```suggestion
  name: {{ template "consul.fullname" . }}-api-gateway-controller
```

charts/consul/templates/api-gateway-gatewayclassconfig.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if (and .Values.apiGateway.enabled .Values.apiGateway.managedGatewayClass.enabled) }}
+apiVersion: api-gateway.consul.hashicorp.com/v1alpha1
+kind: GatewayClassConfig
+metadata:
+  name: consul-api-gateway-class-config
```
thisisnotashwin, line 5:
```suggestion
  name: consul-api-gateway
```

thisisnotashwin: Let me know if this is a breaking change for your team! Ignore it if that is the case.

charts/consul/templates/api-gateway-gatewayclass.yaml
```diff
@@ -0,0 +1,18 @@
...
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: api-gateway-controller
+spec:
+  controllerName: hashicorp.com/consul-api-gateway-controller
+  parametersRef:
+    group: api-gateway.consul.hashicorp.com
+    kind: GatewayClassConfig
+    name: consul-api-gateway-class-config
```
thisisnotashwin, line 17:
```suggestion
    name: consul-api-gateway
```

thisisnotashwin (approved):
Just a bunch of drive-by suggestions to sync this PR with #993

PR #928, opened 2021-12-13
Revert go mod changes from #917

ndhanushkodi:
Changes proposed in this PR:
- revert go mod changes that only fail in cloud acceptance tests: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/3755/workflows/a8f17d55-8b0b-4b80-bf76-f14cf739d8e0/jobs/32826

How I've tested this PR:
- gke acceptance was able to run on this branch

👀


.circleci/config.yml
```diff
@@ -969,6 +969,10 @@ workflows:
             - unit-test-helm-templates
             - unit-acceptance-framework
             - unit-cli
+      - cleanup-gcp-resources
```
ndhanushkodi, line 972: This will be reverted before merging

thisisnotashwin:
Quick question: Are all those dependencies that were fetched in the go.sum fetched by the `ginkgo` and `gomega` dependencies? that seems surprising. Also acceptance test on gke on this branch appear to be failing 🤔

t-eckert (approved):

thisisnotashwin (approved):
nicely done!

ndhanushkodi:
>Are all those dependencies that were fetched in the go.sum fetched by the ginkgo and gomega dependencies? that seems surprising. Also acceptance test on gke on this branch appear to be failing 🤔
I don't think theyre related to ginkgo and gomega, I'm not sure how they came to be, and go mod tidy before and after this commit doesn't change anything which is very strange, so I'm just simply reverting those files here.

You're right! running eks and aks now to see if those pass, in which case we can at least still be sure that vault works on clouds without psps

PR #934, opened 2021-12-14
Show diff when upgrading

t-eckert:

- Show the user the difference between the current installation and the upgraded installation on upgrade.
- Some refactoring of common behaviors.
- Added a `LoadChart` function for loading Helm charts.

> Skip to 1 minute in to see the diff.

https://user-images.githubusercontent.com/29112081/148441040-ff31efdb-41cd-4538-8afe-aa224afe48de.mov

## How I've tested this PR:

- Added tests for diffing and loading Helm charts.

The best way to quickly see the change is to use the CLI to install the `demo` preset, then upgrade to the `secure` preset.

From `./cli`,

``` bash
go run . install -preset demo
```

Wait for install to complete. Then

``` bash
go run . upgrade -preset secure
```

You will see the difference between the two displayed as a diff:

``` diff
  connectInject:
    enabled: true
-   metrics:
-     defaultEnableMerging: true
-     defaultEnabled: true
-     enableGatewayMetrics: true
  controller:
    enabled: true
  global:
+   acls:
+     manageSystemACLs: true
+   gossipEncryption:
+     autoGenerate: true
-   metrics:
-     enableAgentMetrics: true
-     enabled: true
    name: consul
+   tls:
+     enableAutoEncrypt: true
+     enabled: true
- prometheus:
-   enabled: true
  server:
    replicas: 1
- ui:
-   enabled: true
-   service:
-     enabled: true
```

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


cli/cmd/common/diff_test.go
```diff
@@ -0,0 +1,142 @@
...
+				b: map[string]interface{}{
+					"foo": "bar",
+					"baz": []string{"qux"},
+					"qux": map[string]string{
+						"quux": "corge",
+					},
+				},
+			},
+			expected: "+ baz:\n+ - qux\n  foo: bar\n+ qux:\n+   quux: corge\n",
+		},
```
ndhanushkodi, line 72: Maybe we could include a simple string removal case and a deep map removal case as analogs to the new elements cases? wdyt?

ndhanushkodi (commented):
This is looking great Thomas, thanks for writing the diffing logic to be so clear and readable!

Is it worth noting somewhere (docs, comments, etc) that the install overrides and the the diff output will always only show the difference between user-set overrides in the installation and user-set overrides in the upgrade, rather than only the values that are different from the defaults? (For example in the screenshots below, I explicitly set global.enabled=true during install and didn't set it while upgrading, but the default value of global.enabled is true anyways, but it still shows up in the diff). 

One slight nit (also in the screenshots below) is that the values output for install is indented one tab over from the values output for upgrade. Also, install calls those values "Overrides". I think it makes sense to also use the title "Overrides" for the upgrade diff as well.

Screen Shot 2022-01-09 at 10 47 49 PM
Screen Shot 2022-01-09 at 10 48 14 PM

cli/cmd/common/chart_test.go
```diff
@@ -0,0 +1,49 @@
...
+		"key": "value",
+	}
+
+	actual, err := LoadChart(testChartFiles, directory)
+	require.NoError(t, err)
+	require.Equal(t, expectedApiVersion, actual.Metadata.APIVersion)
+	require.Equal(t, expectedName, actual.Metadata.Name)
+	require.Equal(t, expectedVersion, actual.Metadata.Version)
+	require.Equal(t, expectedDescription, actual.Metadata.Description)
+	require.Equal(t, expectedValues, actual.Values)
```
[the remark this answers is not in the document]
t-eckert: Sure! I had removed them as I was changing the values of the fixtures to be parsed by the loader.

cli/cmd/common/utils_test.go
```diff
@@ -1,39 +1,28 @@
...
+	cases := []struct {
+		name     string
+		label    string
+		expected bool
+	}{
+		{"Valid label", "such-a-good-label", true},
+		{"Invalid label empty", "", false},
+		{"Invalid label too long", "a-very-very-very-long-label-that-is-more-than-63-characters-long", false},
+		{"Invalid label starts with a dash", "-invalid-label", false},
+		{"Invalid label ends with a dash", "invalid-label-", false},
```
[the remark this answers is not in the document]
t-eckert: Yes, that was just an oversight as I ported tests over.

cli/cmd/common/chart_test.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func TestReadChartFiles(t *testing.T) {
+	directory := "fixtures/consul"
+	expectedFileNames := []string{"Chart.yaml", "values.yaml", "templates/_helpers.tpl", "templates/foo.yaml"}
```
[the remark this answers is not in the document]
t-eckert: I actually just changed the test back to checking the contents. I removed the contents check when I changed the contents of the fixtures.

cli/cmd/common/diff_test.go
t-eckert: 100%

cli/cmd/common/chart_test.go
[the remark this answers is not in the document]
t-eckert: I added that check back to ReadChartFiles

ishustava (commented):
Amazing work on this PR Thomas!! I really love the refactor and all the work on tests! I'm continuing to review today (and apologies for taking long) but wanted to leave comments I have so far. 

I noticed a few things when running the command with `-dry-run` (full output below):

1. It displayed the summary twice, and only one of them had the diffing. I think we probably only need to display one with diffing.
2. It said `Performing dry run upgrade` twice. Once in the beginning without the `===>` and then once after the first summary.
3. It kind hand a bit at the ` --> preparing upgrade for consul` line, which seemed strange because the info it displayed wasn't that different from the one displayed by `==> Consul Upgrade Summary`. I'm not sure if this is something that is expected.

Here is the full output:
```
$ go run . upgrade -preset demo -set server.replicas=3 -dry-run
    Performing dry run upgrade.

==> Pre-Upgrade Checks
 ✓ Existing installation found to be upgraded.
    Name: consul
    Namespace: consul
 ✓ Loaded charts

==> Consul Upgrade Summary
    Installation name: consul
    Namespace: consul
  connectInject:
    enabled: true
    metrics:
      defaultEnableMerging: true
      defaultEnabled: true
      enableGatewayMetrics: true
  controller:
    enabled: true
  global:
    metrics:
      enableAgentMetrics: true
      enabled: true
    name: consul
  prometheus:
    enabled: true
  server:
-   replicas: 1
+   replicas: 3
  ui:
    enabled: true
    service:
      enabled: true


==> Performing Dry Run Upgrade
 --> preparing upgrade for consul
 --> performing update for consul
 --> dry run for consul
    Dry run complete - upgrade can proceed.
    Config:
    connectInject:
      enabled: true
      metrics:
        defaultEnableMerging: true
        defaultEnabled: true
        enableGatewayMetrics: true
    controller:
      enabled: true
    global:
      metrics:
        enableAgentMetrics: true
        enabled: true
      name: consul
    prometheus:
      enabled: true
    server:
      replicas: 3
    ui:
      enabled: true
      service:
        enabled: true
```

cli/cmd/common/chart_test.go
```diff
@@ -0,0 +1,56 @@
...
+	actual, err := LoadChart(testChartFiles, directory)
+	require.NoError(t, err)
+	require.Equal(t, expectedApiVersion, actual.Metadata.APIVersion)
+	require.Equal(t, expectedName, actual.Metadata.Name)
+	require.Equal(t, expectedVersion, actual.Metadata.Version)
+	require.Equal(t, expectedDescription, actual.Metadata.Description)
+	require.Equal(t, expectedValues, actual.Values)
+}
+
+func TestReadChartFiles(t *testing.T) {
```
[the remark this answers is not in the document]
t-eckert: Thank you!

cli/cmd/common/chart_test.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, f := range files {
+		actualFiles[f.Name] = string(f.Data)
+	}
+
+	for expectedName, expectedContents := range expectedFiles {
+		actualContents, ok := actualFiles[expectedName]
+		require.True(t, ok, "Expected file %s not found", expectedName)
+		require.Equal(t, expectedContents, actualContents)
+	}
+}
```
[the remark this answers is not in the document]
t-eckert: Yes

cli/cmd/common/terminal/ui.go
```diff
@@ -137,7 +139,7 @@ func WithErrorStyle() Option {
 	}
 }
 
-// WithWarningStyle styles the output as an error message.
+// WithWarningStyle styles the output as an warning message.
```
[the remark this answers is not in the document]
t-eckert: Ariana Grande flips her hair

cli/cmd/common/diff_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			actual, err := Diff(tt.args.a, tt.args.b)
+			require.NoError(t, err)
+			require.Equal(t, tt.expected, actual)
+		})
+	}
+}
```
ishustava, line 171: great work on this test!

cli/go.mod
```diff
@@ -113,12 +113,13 @@ require (
...
-	github.com/prometheus/client_golang v1.7.1 // indirect
+	github.com/prometheus/client_golang v1.11.0 // indirect
 	github.com/prometheus/client_model v0.2.0 // indirect
-	github.com/prometheus/common v0.10.0 // indirect
-	github.com/prometheus/procfs v0.2.0 // indirect
+	github.com/prometheus/common v0.26.0 // indirect
+	github.com/prometheus/procfs v0.6.0 // indirect
 	github.com/rubenv/sql-migrate v0.0.0-20200616145509-8d140a17f351 // indirect
 	github.com/russross/blackfriday v1.5.2 // indirect
+	github.com/sergi/go-diff v1.2.0 // indirect
```
ishustava, line 122: are we using this library?

ishustava (commented):
Finishing my review. A few more comments.

cli/go.mod
t-eckert: Oh, not anymore. I’ll remove it.

cli/cmd/common/diff.go
```diff
@@ -0,0 +1,144 @@
...
+
+	buf := new(strings.Builder)
+	err := diffRecursively(a, b, 0, buf)
+
+	return buf.String(), err
+}
+
+// diffRecursively iterates over both maps and writes the differences to the given buffer.
+func diffRecursively(a, b map[string]interface{}, recurseDepth int, buf *strings.Builder) error {
+	// Get the union of keys in a and b sorted alphabetically.
```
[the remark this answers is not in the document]
t-eckert: Just for consistency of output. Otherwise the diff could come out in a different order every time. I figure alphabetically is a fair enough consistent sorting. 

With some additional work, I could get it outputting in the same order as values.yaml, but it didn’t seem like the right choice for now.

Without some sorting it would be harder (not impossible) to test as well.

cli/cmd/common/diff.go
```diff
@@ -0,0 +1,144 @@
...
+				if err != nil {
+					return err
+				}
+
+				writeWithPrepend("  ", string(asYaml), recurseDepth, buf)
+				continue
+			}
+
+			// If the map slices are different and there is another level of depth to the map, recurse.
+			if !isMaxDepth(aSlice) && !isMaxDepth(bSlice) {
```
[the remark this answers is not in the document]
t-eckert: Yes, but that would require reflection as far as I can tell. I might be wrong though.

cli/cmd/common/chart_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"embed"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Embed a test chart to test against.
+//go:embed fixtures/consul/* fixtures/consul/templates/_helpers.tpl
```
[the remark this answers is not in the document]
t-eckert: @ndhanushkodi did it first! It’s new as of Go 1.16. 

I’ll check on that. I have some vague sense that Go is weird about the wildcard in this case. It has special rules here too like not being able to access parent directories.

cli/cmd/common/diff.go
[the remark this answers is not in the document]
ishustava: yeah you can do it with reflection (I think it'd be justified in this case). Another way is to try to cast to a map, e.g. 

```
valueInAMap, ok := valueInA.(map[string]interface{})
```
If casting is successful (`ok` is `true`), it's map. Otherwise it's not.

cli/cmd/common/diff.go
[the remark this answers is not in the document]
ishustava: Oh got it, thanks for clarifying that.

t-eckert:
> Is it worth noting somewhere (docs, comments, etc) that the install overrides and the the diff output will always only show the difference between user-set overrides in the installation and user-set overrides in the upgrade, rather than only the values that are different from the defaults? (For example in the screenshots below, I explicitly set global.enabled=true during install and didn't set it while upgrading, but the default value of global.enabled is true anyways, but it still shows up in the diff).

This is such an interesting question. I think the solution for now is what you said where there is a note "Diffing user overrides..." something like that. I think we should discuss the right approach for future iterations as a team. Should we show the whole diff with the whole helm chart or just the overrides? I could go either way.

cli/cmd/common/diff_test.go
t-eckert: Thank you!

cli/cmd/common/diff.go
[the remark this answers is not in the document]
t-eckert: Brilliant!

cli/cmd/common/chart_test.go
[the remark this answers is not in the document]
t-eckert: Yeah, you have to explicitly embed that helpers file too. 🤷🏻

cli/cmd/common/diff.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	buf := new(strings.Builder)
+	err := diffRecursively(a, b, 0, buf)
+
+	return buf.String(), err
+}
+
+// diffRecursively iterates over both maps and writes the differences to the given buffer.
+func diffRecursively(a, b map[string]interface{}, recurseDepth int, buf *strings.Builder) error {
```
thisisnotashwin, line 25: 🚗 drive by comment: you could potentially change the signature of diffRecursively to (a, b map[string]interface{}) (string, err)

Then Diff() can return diffRecursively(a, b, 0) and you wouldn't need to create the stringBuilder inside Diff. IMO it reads more idiomatic.

cli/cmd/common/chart_test.go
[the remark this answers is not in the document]
t-eckert: So I took some time to look into this test. It would require setting up a mock K8s client and loading it with configuration as part of the test setup. I think this would be a good test to do, but I feel we should first do some work to abstract over Helm in the code. Right now, the code is coupled to the Helm CLI code in a way that makes it hard to test. I have an idea for how to improve this. Are you ok with skipping this test for now?

cli/cmd/common/diff.go
t-eckert: I definitely see the appeal of that. I tried that model too. There is a way to do it with each recursive call creating a string buffer and a way where concatenated strings are just returned. 

The former can be wasteful with generating new buffers. Not really a problem, but easily avoided.

The latter gets messy because the recursion isn't purely linear. It's recursing down a tree structure so you get all these concatenations to the current diff plus where you need to put the values from the next recursion.

I'd be willing to give the second model one more try and see if I can get it clean.

thisisnotashwin: hmm. you can still create just the single buffer but create it inside `diffRecursively()`, if that would make the execution in other places simpler.
Also, given we are parsing our helm config which we know does not go very deep in terms of nesting, it should not involve that much recursion so I think we can get away with having multiple buffers.

t-eckert: I'll think about it some more and come back to the idea later. I think what is there is ok for now. But I do appreciate this take.

t-eckert: I tried to integrate those thoughts here. Let me know what you think! I always appreciate your drive-bys.

thisisnotashwin: Just gave it a read! I really like it. Thanks for addressing it :)

cli/common/utils.go
```diff
@@ -0,0 +1,108 @@
...
+		if rel.Chart.Metadata.Name == "consul" {
+			return rel.Name, rel.Namespace, nil
+		}
+	}
+	return "", "", errors.New("couldn't find consul installation")
+}
+
+// MergeMaps merges two maps giving b precedent.
+// @source: https://github.com/helm/helm/blob/main/pkg/cli/values/options.go
+func MergeMaps(a, b map[string]interface{}) map[string]interface{} {
```
ishustava, line 58: This needs to be tested

cli/common/utils_test.go
```diff
@@ -0,0 +1,30 @@
...
+		label    string
+		expected bool
+	}{
+		{"Valid label", "such-a-good-label", true},
+		{"Invalid label empty", "", false},
+		{"Invalid label contains capital letters", "Peppertrout", false},
+		{"Invalid label contains underscores", "this_is_not_python", false},
+		{"Invalid label too long", "a-very-very-very-long-label-that-is-more-than-63-characters-long", false},
+		{"Invalid label starts with a dash", "-invalid-label", false},
+		{"Invalid label ends with a dash", "invalid-label-", false},
```
ishustava, line 21: I like how broke down previous test cases, but it looks like we lost `Standard name with leading numbers works.` test case

ishustava (approved):
I think this looks good! Thanks for your patience as we were working through the UX and having long review cycles.

Leaving a few minor comments, but won't block.

cli/cmd/upgrade/upgrade.go
```diff
@@ -170,11 +168,18 @@ func (c *Command) Run(args []string) int {
...
+	if err != nil {
 		c.UI.Output(err.Error())
 		return 1
 	}
 
+	c.timeoutDuration, err = time.ParseDuration(c.flagTimeout)
+	if err != nil {
+		c.UI.Output(fmt.Sprintf("Invalid timeout: %s", err))
+		return 1
+	}
```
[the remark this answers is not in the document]
t-eckert: I moved it out of `validateFlags` because I felt it was a side-effect that didn't meet the intention of that function. IMHO it should just validate flags and not set values on the struct.

cli/cmd/upgrade/upgrade.go
```diff
@@ -170,11 +168,18 @@ func (c *Command) Run(args []string) int {
 
 	defer common.CloseWithError(c.BaseCommand)
 
-	if err := c.validateFlags(args); err != nil {
+	err := c.validateFlags(args)
+	if err != nil {
```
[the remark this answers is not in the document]
t-eckert: This is kind of a knock-on change from the change below at line 177. If I use `:=` on line 177 the compiler gets mad because `c.timeoutDuration` is already instantiated (even though err isn't). So I needed to instantiate `err` in the function scope to fix that error. I could do that either with the `var` keyword or just by moving this check. Both added a new line and I felt that between the two, this was the cleaner option. Otherwise there would be an `err` var shadowing an `err` in the enclosing scope that wouldn't be instantiated until later.

cli/cmd/upgrade/upgrade.go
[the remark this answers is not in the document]
ishustava: Ah, I see what you mean. The downside though is that we have the parsing and error handling logic twice. Maybe just documenting that it will also set the timeout in the function description or renaming it to `validateAndParseFlags` would work better?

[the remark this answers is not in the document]
ishustava: You know, I actually realized that we can just change the type of `flagTimeout` from `string` to `time.Duration` and then we don't need validation at all. For example:

https://github.com/hashicorp/consul-k8s/blob/c33d3bc62c396cf83749e52bdf445bcb06077b52/control-plane/subcommand/server-acl-init/command.go#L212

cli/cmd/upgrade/upgrade.go
```diff
@@ -374,9 +331,6 @@ func (c *Command) validateFlags(args []string) error {
 		}
 	}
 
-	if c.flagDryRun {
-		c.UI.Output("Performing dry run upgrade.", terminal.WithInfoStyle())
```
ndhanushkodi, line 378: Do we want to print something like this at the top before anything else so the user knows all their output is from a dry run?

ndhanushkodi (approved):
Looks good to me Thomas, I didn't try manually testing this time around, but let me know if you'd like me to!

cli/cmd/upgrade/upgrade.go
t-eckert: Sure, I just deleted it here because it was printing twice. Once here and once later on. I can shift that single printing to be earlier.

cli/common/utils_test.go
t-eckert: Agreed, just an oversight from me. Thank you for catching it!

cli/common/utils.go
t-eckert: Done :)

cli/cmd/upgrade/upgrade.go
t-eckert: I think it would ease user anxiety to show this earlier so I will.

t-eckert: I added it on line 178 to print "Dry Run Upgrade: No changes will be made to the cluster."

I didn't want it to be in `validateFlags` because it's a side effect.

PR #937, opened 2021-12-20
Add new value ui.dashboardURLTemplates

lkysow:
This value is used to set Consul's `ui_config.dashboard_url_templates`
config. I'm pulling it out as its own value rather than getting users
to just use `server.extraConfig` because these template strings always
use the characters `{{` and when used within `server.extraConfig`,
users need to escape those characters.

Before:

```yaml
server:
  extraConfig: |
    {
      "ui_config": {
        "dashboard_url_templates": {
          "service": "http://grafana/service/{{ "{{" }}Service.Name}}"
        }
      }
    }
```

Now:

```yaml
ui:
  dashboardURLTemplates:
    service: "http://grafana/service/{{Service.Name}}"
```

How I've tested this PR:
- tests and manual deployment

- code

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/charts/consul/test/unit/server-statefulset.bats
+++ b/charts/consul/test/unit/server-statefulset.bats
@@ -1785,3 +1785,20 @@ load _helpers
     yq -r '.containers[0].volumeMounts[] | select(.name == "consul-ca-key")' | tee /dev/stderr)
   [ "${actual}" = "" ]
 }
+
+#--------------------------------------------------------------------
+# ui.dashboardURLTemplates.service
+
+@test "server/StatefulSet: ui.dashboardURLTemplates.service sets the template" {
+  cd `chart_dir`
+
+  local expected='-hcl='\''ui_config { dashboard_url_templates { service = \"http://localhost:3000/d/WkFEBmF7z/services?orgId=1&var-Service={{Service.Name}}\" } }'
+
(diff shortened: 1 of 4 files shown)


lkysow:
https://github.com/hashicorp/consul-helm/issues/1006 is an issue where this caused problems.

charts/consul/test/unit/server-statefulset.bats
```diff
@@ -1785,3 +1785,20 @@ load _helpers
     yq -r '.containers[0].volumeMounts[] | select(.name == "consul-ca-key")' | tee /dev/stderr)
   [ "${actual}" = "" ]
 }
+
+#--------------------------------------------------------------------
+# ui.dashboardURLTemplates.service
+
+@test "server/StatefulSet: ui.dashboardURLTemplates.service sets the template" {
```
ishustava, line 1792: do we need to add a test for default behavior?

ishustava (approved):

charts/consul/test/unit/server-statefulset.bats
lkysow: done.

ndhanushkodi (approved):
This looks great!

PR #940, opened 2021-12-23
fix: add license to snapshot-agent

TJM:
Changes proposed in this PR:
- Add the consul enterprise license to the snapshot agent regardless of `global.acls.manageSystemACLs` 

How I've tested this PR:
helm upgrade with our existing values file

* values:

```yaml
global:
  acls:
    manageSystemACLs: true
  enterpriseLicense:
    enableLicenseAutoload: true
    secretKey: license-key
    secretName: consul-enterprise-license
client:
  snapshotAgent:
    configSecret:
      secretKey: agent_config
      secretName: consul-snapshot-agent
    enabled: true
```

You will need the requisite secrets for `consul-enterprise-license` and `consul-snapshot-agent`.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

Related to ZenDesk Support Ticket # 62914

### DIFF
--- a/charts/consul/templates/client-snapshot-agent-deployment.yaml
+++ b/charts/consul/templates/client-snapshot-agent-deployment.yaml
@@ -62,12 +62,11 @@ spec:
         {{- if .Values.global.acls.manageSystemACLs }}
         - name: aclconfig
           emptyDir: {}
-        {{- else }}
+        {{- end }}
         {{- if (and .Values.global.enterpriseLicense.secretName .Values.global.enterpriseLicense.secretKey .Values.global.enterpriseLicense.enableLicenseAutoload) }}
         - name: consul-license
           secret:
             secretName: {{ .Values.global.enterpriseLicense.secretName }}
-        {{- end }}
         {{- end }}
         {{- if .Values.global.tls.enabled }}
         {{- if not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots) }}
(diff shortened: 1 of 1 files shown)


TJM:
I can't tell whether the test (that failed above) is incorrect for Consul versions > 1.10.0 or if the ACL needs to be modified to handle this version of Consul? If we have "manageSystemACLs" set to true, snapshot agent fails on 1.10.x and 1.11.x.

TJM:
I realize I submitted this over the holidays, but I am hoping for some feedback. Do we need to remove/revert the test, maybe do some sort of version checks to make the test conditional? or does something in the ACL need updated for 1.10.x?

david-yu:
Thanks @TJM our team will take a look!

lkysow:
If ACLs are enabled then the snapshot agent should be pulling its own license. If this isn't happening then it's probably a Consul bug. Will need to investigate.

david-yu:
Hi @TJM what version of Consul are you using?

TJM:
@david-yu - We have been stuck on 1.9.x due to this bug. We are currently on `1.9.9+ent` (and I know we could go to a newer 1.9.x, as I did that in a test env)

ishustava:
Hey @TJM 

Thanks for this PR! As Luke mentioned, it **should** work when you provide the ACL token and it is a bug in Consul which is going to be worked on soon. 

The reason providing a license like you did in this PR works is because it bypasses the auto-retrieval logic and instead just uses the provided license.

This change isn't something we'd want there to be all the time in the helm chart because we would want the snapshot agent to auto-retrieve the license instead of relying on the provided one. This is useful because that means you only need to provide the license to servers, and then clients and snapshot agent should auto-retrieve that license in the case when ACLs are enabled. This would also handle updates to the license if it has expired, for example.

We're going to ask the Consul team to fix this.

TJM:
@ishustava - Yep, that is what I expected, once I saw that there was a test specifically laying out what I tried to circumvent. I suggest that we can close this PR, since it will not be merged. Perhaps I should have just started with an issue, but when I saw the way the conditionals were laid out in the templates, I thought it was just a "bug" :) 

Will we use our zendesk ticket or should I file an issue in lieu of this PR?

TJM:
FYI: As per above, this PR will not be merged, but if you need this workaround, I have published it as of today's main...

https://github.com/TJM/consul-k8s/releases/download/v0.40.0-pr940/consul-0.40.0-pr940.tgz

Use at your own risk, of course.

PR #953, opened 2022-01-06
Setup and run partitions tests on EKS

ndhanushkodi:
Changes proposed in this PR:
- Run partitions tests on EKS
- For cross partition communication, the networking setup needs to be such that the node ips of all clients need to be able to talk with the node ips of all servers and vice versa (traffic should be routeable and allowed). Clients across partitions do not need to be able to route to each other.
- To make this work on EKS, the terraform has been modified to:
  - Set up a 2 clusters in 2 VPCs without overlapping IPs
  - Set up peering between those VPCs
  - Add security group ingress rules to each set of nodes' security group, to allow all TCP traffic from the other VPC
  - Add route table entries such that traffic going to the other VPC is routed through the peer
- A few tweaks to partitions acceptance tests so that they pass
- Outside of this PR, I added permissions to the ci role used for helm tests to be able to do anything on EC2 (ec2*) using the aws cli: 
```
aws iam put-role-policy --role-name service-circleci-consul-helm-tests --policy-name CircleCIManageEC2 --policy-document file://ec2policy.json
```
where ec2policy.json looks like this:
```
{
   "Version": "2012-10-17",
   "Statement": [
    {
      "Effect": "Allow",
      "Action": "ec2:*",
      "Resource": "*"
    }
   ]
}
```

How I've tested this PR:
- Running EKS acceptance tests in the pipeline: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/3871/workflows/854039f8-5c8c-4b29-8e43-6a45445025c3 (without tproxy)
- EKS acc tests with tproxy passes here: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/3875/workflows/7b4c2445-4848-44f9-b3e0-01fca56cd9e2/jobs/34507

- Review-- I'd like especially like feedback on:
  - ~This currently relies on always deploying 2 clusters, and isn't conditional on any variables for whether to set up the peering. Since the pipeline always deploys 2 clusters, I thought this could be alright. We could even turn cluster_count from a var to a local, to make it clearer the config is meant for deploying 2 clusters. In a case where someone wants to deploy 1 cluster, they'd need to remove the config from line 92 down, and change the local cluster_count to 1.~ Edited to work with cluster_count = 1 or 2.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,7 @@ commands:
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
+                          -run TestPartitions \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
                       echo "Tests in ${pkg} failed, aborting early"
@@ -134,6 +135,7 @@ commands:
                       -enable-multi-cluster \
                       ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                       -debug-directory="$TEST_RESULTS/debug" \
+                      -run TestPartitions \
                       -consul-k8s-image=<< parameters.consul-k8s-image >>
 
 jobs:
@@ -764,7 +766,7 @@ jobs:
       - run: mkdir -p $TEST_RESULTS
 
       - run-acceptance-tests:
-          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig"
+          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig" -enable-transparent-proxy
 
       - store_test_results:
           path: /tmp/test-results
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,6 @@ commands:
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
-                          -run TestPartitions \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
                       echo "Tests in ${pkg} failed, aborting early"
@@ -135,7 +134,6 @@ commands:
                       -enable-multi-cluster \
                       ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                       -debug-directory="$TEST_RESULTS/debug" \
-                      -run TestPartitions \
                       -consul-k8s-image=<< parameters.consul-k8s-image >>
 
 jobs:
@@ -971,10 +969,6 @@ workflows:
             - unit-test-helm-templates
             - unit-acceptance-framework
             - unit-cli
-      - cleanup-eks-resources
-      - acceptance-eks-1-19:
-          requires:
-            - cleanup-eks-resources
   nightly-acceptance-tests:
     triggers:
       - schedule:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,6 @@ commands:
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
-                          -run TestPartitions \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
                       echo "Tests in ${pkg} failed, aborting early"
@@ -135,7 +134,6 @@ commands:
                       -enable-multi-cluster \
                       ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                       -debug-directory="$TEST_RESULTS/debug" \
-                      -run TestPartitions \
                       -consul-k8s-image=<< parameters.consul-k8s-image >>
 
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -972,7 +972,7 @@ workflows:
             - unit-acceptance-framework
             - unit-cli
       - cleanup-eks-resources
-      - acceptance-eks-1-18:
+      - acceptance-eks-1-19:
           requires:
             - cleanup-eks-resources
   nightly-acceptance-tests:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,7 @@ commands:
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
+                          -run TestPartitions \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
                       echo "Tests in ${pkg} failed, aborting early"
@@ -134,6 +135,7 @@ commands:
                       -enable-multi-cluster \
                       ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                       -debug-directory="$TEST_RESULTS/debug" \
+                      -run TestPartitions \
                       -consul-k8s-image=<< parameters.consul-k8s-image >>
 
 jobs:
@@ -764,7 +766,7 @@ jobs:
       - run: mkdir -p $TEST_RESULTS
 
       - run-acceptance-tests:
-          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig" -enable-transparent-proxy
+          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig"
 
       - store_test_results:
           path: /tmp/test-results
@@ -775,7 +777,7 @@ jobs:
           name: terraform destroy
           working_directory: *eks-terraform-path
           command: |
-            terraform destroy -auto-approve
+            terraform destroy -var cluster_count=2 -auto-approve
           when: always
 
       - slack/status:
@@ -969,6 +971,10 @@ workflows:
             - unit-test-helm-templates
             - unit-acceptance-framework
             - unit-cli
+      - cleanup-eks-resources
+      - acceptance-eks-1-18:
+          requires:
+            - cleanup-eks-resources
   nightly-acceptance-tests:
     triggers:
       - schedule:
--- a/acceptance/tests/partitions/partitions_test.go
+++ b/acceptance/tests/partitions/partitions_test.go
@@ -32,9 +32,9 @@ func TestPartitions(t *testing.T) {
 		t.Skipf("skipping this test because -enable-enterprise is not set")
 	}
 
-	if !cfg.UseKind {
-		t.Skipf("skipping this test because Admin Partition tests are only supported in Kind for now")
-	}
+	//if !cfg.UseKind {
+	//	t.Skipf("skipping this test because Admin Partition tests are only supported in Kind for now")
+	//}
 
 	const defaultPartition = "default"
 	const secondaryPartition = "secondary"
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -103,6 +103,7 @@ commands:
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
+                          -run TestPartitions \
```
ndhanushkodi, line 106: To be removed before merging

.circleci/config.yml
```diff
@@ -969,6 +971,10 @@ workflows:
             - unit-test-helm-templates
             - unit-acceptance-framework
             - unit-cli
+      - cleanup-eks-resources
+      - acceptance-eks-1-18:
+          requires:
+            - cleanup-eks-resources
```
ndhanushkodi, line 977: To be removed before merging

acceptance/tests/partitions/partitions_test.go
```diff
@@ -32,10 +32,6 @@ func TestPartitions(t *testing.T) {
 		t.Skipf("skipping this test because -enable-enterprise is not set")
 	}
 
-	if !cfg.UseKind {
-		t.Skipf("skipping this test because Admin Partition tests are only supported in Kind for now")
-	}
-
```
ndhanushkodi, line 38: Should I add a temporary config value for EKS? To enable this only on Kind and EKS until we get all the clouds running?

acceptance/tests/partitions/partitions_test.go
```diff
@@ -258,7 +251,6 @@ func TestPartitions(t *testing.T) {
 			// Ensure consul client are created.
 			agentPodList, err := clientClusterContext.KubernetesClient(t).CoreV1().Pods(clientClusterContext.KubectlOptions(t).Namespace).List(ctx, metav1.ListOptions{LabelSelector: "app=consul,component=client"})
 			require.NoError(t, err)
-			require.Len(t, agentPodList.Items, 1)
```
ndhanushkodi, line 261: In a cloud env, there will be more than one client agent bc there will be multiple nodes. I could add back this check just for Kind but it seemed not essential

acceptance/tests/partitions/partitions_test.go
```diff
@@ -181,13 +177,10 @@ func TestPartitions(t *testing.T) {
...
-				// The Kubernetes AuthMethod IP for Kind is read from the endpoint for the Kubernetes service. On other clouds,
-				// this can be identified by reading the cluster config.
-				kubernetesEndpoint, err := clientClusterContext.KubernetesClient(t).CoreV1().Endpoints(defaultNamespace).Get(ctx, "kubernetes", metav1.GetOptions{})
-				require.NoError(t, err)
-				k8sAuthMethodHost = fmt.Sprintf("%s:%d", kubernetesEndpoint.Subsets[0].Addresses[0].IP, kubernetesEndpoint.Subsets[0].Ports[0].Port)
-			}
+			// The Kubernetes AuthMethod IP for Kind is read from the endpoint for the Kubernetes service.
+			kubernetesEndpoint, err := clientClusterContext.KubernetesClient(t).CoreV1().Endpoints(defaultNamespace).Get(ctx, "kubernetes", metav1.GetOptions{})
+			require.NoError(t, err)
+			k8sAuthMethodHost = fmt.Sprintf("%s:%d", kubernetesEndpoint.Subsets[0].Addresses[0].IP, kubernetesEndpoint.Subsets[0].Ports[0].Port)
```
ndhanushkodi, line 183: In EKS theres a service called "kubernetes" which points to the API as well, so I removed the `if cfg.UseKind`.

acceptance/tests/partitions/partitions_test.go
thisisnotashwin: hmm.. maybe we can explicitly skip AKS and GKE? that way one can peel away those checks as those clouds are configured for these tests? is that possible?

acceptance/tests/partitions/partitions_test.go
```diff
@@ -168,11 +164,11 @@ func TestPartitions(t *testing.T) {
...
 			if !cfg.UseKind {
 				// Get the IP of the partition service to configure the external server address in the values file for the workload cluster.
-				partitionSecretName := fmt.Sprintf("%s-partition-secret", releaseName)
+				partitionServiceName := fmt.Sprintf("%s-consul-partition-service", releaseName)
 				logger.Logf(t, "retrieving partition service to determine external IP for servers")
-				partitionsSvc, err := serverClusterContext.KubernetesClient(t).CoreV1().Services(serverClusterContext.KubectlOptions(t).Namespace).Get(ctx, partitionSecretName, metav1.GetOptions{})
+				partitionsSvc, err := serverClusterContext.KubernetesClient(t).CoreV1().Services(serverClusterContext.KubectlOptions(t).Namespace).Get(ctx, partitionServiceName, metav1.GetOptions{})
 				require.NoError(t, err)
-				partitionSvcIP = partitionsSvc.Status.LoadBalancer.Ingress[0].IP
+				partitionSvcIP = partitionsSvc.Status.LoadBalancer.Ingress[0].Hostname
```
thisisnotashwin, line 171: NICE!

acceptance/tests/partitions/partitions_test.go
thisisnotashwin: i agree..it's not essential!!

thisisnotashwin (approved):
THIS LOOKS GREAT!! EXCELLENTLY DONE NITYA!

charts/consul/test/terraform/eks/main.tf
```diff
@@ -84,4 +87,86 @@ data "aws_eks_cluster" "cluster" {
 data "aws_eks_cluster_auth" "cluster" {
   count = var.cluster_count
   name  = module.eks[count.index].cluster_id
+}
+
+# The following resources are meant to apply when cluster_count=2 to set up vpc peering and the appropriate routes and
+# security groups so traffic between vpc's is allowed. Since the pipeline only ever deploys with a cluster_count of 2,
+# these resources are hardcoded to work in an environment with 2 clusters.
```
thisisnotashwin, line 94: 🙏 thanks for the clarification. I was about to ask about it 😁

thisisnotashwin: @ishustava is this true for AKS as well? i noticed in your PR that is doesn't assume that the cluster_count is 2 and computes it.

ishustava: 🤔 yeah I didn't assume that. I left a comment about it above.

ndhanushkodi: Yeah for AKS, the cluster_count is allowed to be 1 or 2. In the current state of this EKS PR, cluster_count is only allowed to be 2, unless you make some changes to the terraform (like commenting out the whole vpc block). Because these EKS resources are like unidirectional, so I needed to set them up such that ingress rule from cluster 0 --> 1 works, and vice versa, so I couldn't just rely on the count variable for these resources.

charts/consul/test/terraform/eks/main.tf
```diff
@@ -84,4 +87,86 @@ data "aws_eks_cluster" "cluster" {
...
+
+# Get the route table ids that are associated with eks cluster 0, so we can create routes in those route tables.
+data "aws_route_tables" "rtseks0" {
+  vpc_id = module.vpc[0].vpc_id
+
+  filter {
+    name   = "tag:Name"
+    values = [format("%s-*", module.eks[0].cluster_id)]
+  }
+}
```
ishustava, line 146: I'm curious why do we need a data resource if we have the vpc module created in this terraform. In my experience, I usually use data resource for resources that exist in a different terraform state.

Should this be a local variable instead?

charts/consul/test/terraform/eks/main.tf
ndhanushkodi: Yeah for AKS, the cluster_count is allowed to be 1 or 2. In the current state of this EKS PR, cluster_count is only allowed to be 2, unless you make some changes to the terraform (like commenting out the whole vpc block). Because these EKS resources are like unidirectional, so I needed to set them up such that ingress rule from cluster 0 --> 1 works, and vice versa, so I couldn't just rely on the count variable for these resources.

ishustava (commented):
Amazing work, Nitya!!

I left some comments inline. Sorry for taking so long to review!

charts/consul/test/terraform/eks/main.tf
ndhanushkodi: But it looks like @ishustava 's suggestion for conditionalizing the count may work! terraform apply is chugging along, so 🤞 that destroy works alright too

acceptance/tests/partitions/partitions_test.go
ndhanushkodi: should we just consider partitions-eks the feature branch? i saw the aks pr was made against this branch

charts/consul/test/terraform/eks/main.tf
```diff
@@ -84,4 +87,86 @@ data "aws_eks_cluster" "cluster" {
 data "aws_eks_cluster_auth" "cluster" {
   count = var.cluster_count
   name  = module.eks[count.index].cluster_id
+}
+
+# The following resources are meant to apply when cluster_count=2 to set up vpc peering and the appropriate routes and
+# security groups so traffic between vpc's is allowed. Since the pipeline only ever deploys with a cluster_count of 2,
```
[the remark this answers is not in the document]
ndhanushkodi: Edited with your suggestion! Passed here: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/3906/workflows/62c03cdd-aaf4-49e5-bb65-3f7335ddef8f/jobs/34923

charts/consul/test/terraform/eks/main.tf
ishustava: I think you can use vpc module output so you can combine public and private route table ids into one array: `[module.vpc[count.index].public_route_table_ids[0], module.vpc[count.index].private_route_table_ids[0]]`

ishustava: I think the vpc module creates two route tables that are shared between subnets. So there should be one private and one public route table per VPC

acceptance/tests/partitions/partitions_test.go
ndhanushkodi: Ya that makes sense, I changed the base to cloud-partitions-tests as the feature branch

ishustava (approved):
🎉 🎉 

I'm curious what you think about this comment: https://github.com/hashicorp/consul-k8s/pull/953#discussion_r782523964

Otherwise, looks good! I won't hold up my approval on that as I don't think it's blocking

charts/consul/test/terraform/eks/main.tf
```diff
@@ -84,4 +87,93 @@ data "aws_eks_cluster" "cluster" {
...
+# Get the route table ids that are associated with eks cluster 1, so we can create routes in those route tables.
+data "aws_route_tables" "rtseks1" {
+  count  = var.cluster_count > 1 ? 1 : 0
+  vpc_id = module.vpc[1].vpc_id
+
+  filter {
+    name   = "tag:Name"
+    values = [format("%s-*", module.eks[1].cluster_id)]
+  }
+}
```
thisisnotashwin, line 161: will the below work? or will it interact poorly with the `aws_route` in the next step? it could potentially work as rtseks[0]
```suggestion
# Get the route table ids that are associated with eks cluster 0, so we can create routes in those route tables.
data "aws_route_tables" "rtseks" {
  count  = var.cluster_count > 1 ? 2 : 0
  vpc_id = module.vpc[count.index].vpc_id

  filter {
    name   = "tag:Name"
    values = [format("%s-*", module.eks[count.index].cluster_id)]
  }
}
```

ndhanushkodi: oh yeah that could possibly work! but I think i'll end up removing these per iryna's suggestion to use the vpc module to grab the route tables instead!

thisisnotashwin: sounds good!! 👍

ndhanushkodi:
Thanks all! Sorry for missing that comment earlier Iryna! I'm testing that suggestion here: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s?branch=partitions-eks&filter=all

charts/consul/test/terraform/eks/main.tf
ndhanushkodi: sorry for missing this earlier! your suggestion worked for me and I was able to remove the route table data block!

PR #956, opened 2022-01-10
control-plane: Add possibility to manage consul sidecar resources with annotations

rrondeau:
Hi

We are currently migrating our metrics polling to use merged metrics and we have one issue.
Some of our pods return lots of metrics (~150k) and the consul sidecar get OOM killed.
We bump the memory of the sidecar but dont want to waste memory where we dont need it.

Changes proposed in this PR:
- default resources are set via flags of the injector
- can be overriden with annotations on pods

I renamed the injector flags to mirror the envoy sidecar configuration.
The code is heavily inspired by envoy sidecar configuration but i kept the `consulSidecarResources` object in the subcommand.
If you want that i replicate the envoy stuff and split the resources into 4 vars when passing the settings to the handler, i can do it.

I am open to any ideas / modifications :)

How I've tested this PR:
- Added tests
- Run the tests ( control-plane and charts )

¯\\_(ツ)_/¯

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -201,16 +201,16 @@ spec:
                 {{- if .Values.global.consulSidecarContainer }}
                 {{- $consulSidecarResources := .Values.global.consulSidecarContainer.resources }}
                 {{- if not (kindIs "invalid" $consulSidecarResources.limits.memory) }}
-                -consul-sidecar-memory-limit={{ $consulSidecarResources.limits.memory }} \
+                -default-consul-sidecar-memory-limit={{ $consulSidecarResources.limits.memory }} \
                 {{- end }}
                 {{- if not (kindIs "invalid" $consulSidecarResources.requests.memory) }}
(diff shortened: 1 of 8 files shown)


david-yu:
Thanks for the PR @rrondeau this sounds very useful. I'm curious how is it that your pod emits 150k metrics? That is definitely something that we did not consider before!

rrondeau:
Yeah we have 2 types of pod that need more than 50Mi of memory.
- nodejs app + envoy metrics with ~ 20 upstreams
```bash
$ curl -s 127.0.0.1:20200/metrics | grep ^envoy_ | wc -l
11461
```
- nodejs app with lots of custom metrics for our business/internal usage
```bash
/ $ curl -s 127.0.0.1:20200/metrics | grep -v -e ^envoy_ -e ^# -e ^$ | wc -l
138533
```

ishustava (approved):
Looks good to me!

ndhanushkodi (approved):
This looks great!

PR #963, opened 2022-01-12
Add partitions test to GKE

thisisnotashwin:
Changes proposed in this PR:
- Update GKE terraform templates for partitions tests
  - we create firewall rules that allow nodes and pods from each cluster to reach the nodes of the other.
- Update server podsercuritypolicy to allow hostports on specific ports if `server.exportGossipAndRPCPorts` is set to `true`

How I've tested this PR:
- ran tests: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/3926/workflows/1aeed2ff-40b4-4a7a-99cb-f84f751048a0

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,7 @@ commands:
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
+                          -run TestPartitions \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
                       echo "Tests in ${pkg} failed, aborting early"
@@ -134,6 +135,7 @@ commands:
                       -enable-multi-cluster \
                       ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                       -debug-directory="$TEST_RESULTS/debug" \
+                      -run TestPartitions \
                       -consul-k8s-image=<< parameters.consul-k8s-image >>
 
 jobs:
@@ -764,7 +766,7 @@ jobs:
       - run: mkdir -p $TEST_RESULTS
 
       - run-acceptance-tests:
-          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig"
+          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig" -enable-transparent-proxy
 
       - store_test_results:
           path: /tmp/test-results
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,6 @@ commands:
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
-                          -run TestPartitions \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
                       echo "Tests in ${pkg} failed, aborting early"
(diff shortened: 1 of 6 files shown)


thisisnotashwin:
The PR is ready. Waiting for it to go 🟢 on CI before marking it ready for review.

.circleci/config.yml
```diff
@@ -971,8 +971,8 @@ workflows:
             - unit-test-helm-templates
             - unit-acceptance-framework
             - unit-cli
-      - cleanup-eks-resources
-      - acceptance-aks-1-21
+      - cleanup-gcp-resources
+      - acceptance-gke-1-20
```
thisisnotashwin, line 975: These will be removed before merging!

charts/consul/templates/server-podsecuritypolicy.yaml
```diff
@@ -27,6 +27,15 @@ spec:
     - 'downwardAPI'
     - 'persistentVolumeClaim'
   hostNetwork: false
+  hostPorts:
+  {{- if .Values.server.exposeGossipAndRPCPorts }}
+  - min: 8300
+    max: 8300
+  - min: {{ .Values.server.ports.serflan.port }}
+    max: {{ .Values.server.ports.serflan.port }}
```
thisisnotashwin, line 35: Im not sure what happens here if the value of the `serflan.port` is greater than 8302. Does the next entry then clobber it?

charts/consul/test/terraform/gke/variables.tf
```diff
@@ -19,10 +19,18 @@ variable "init_cli" {
...
   default     = 1
   description = "The number of Kubernetes clusters to create."
+
+  // We currently cannot support more than 2 cluster
+  // because setting up peering is more complicated if cluster count is
+  // more than two.
+  validation {
+    condition     = var.cluster_count < 3 && var.cluster_count > 0
+    error_message = "The cluster_count value must be 1 or 2."
+  }
```
thisisnotashwin, line 29: Always a good idea to steal from @ishustava

charts/consul/templates/server-podsecuritypolicy.yaml
thisisnotashwin: This definitely works with the default values. A touch worried about non-default ones.

ishustava: I don't think the next entry would clobber it. I think each array item is an independent port range.

charts/consul/test/terraform/gke/main.tf
```diff
@@ -24,14 +24,33 @@ resource "google_container_cluster" "cluster" {
...
   }
 
   resource_labels = var.labels
 }
 
+resource "google_compute_firewall" "firewall-rules" {
+  project     = var.project
+  name        = "consul-k8s-acceptance-firewall-${random_id.suffix[count.index].dec}"
+  network     = "default"
+  description = "Creates firewall rule targeting tagged instances"
```
ishustava, line 41: need to update description I think to reflect what it does
```suggestion
  description = "Creates firewall rule allowing traffic from nodes and pods of the ${random_id.suffix[count.index == 0 ? 1 : 0].dec} Kubernetes cluster."
```

charts/consul/test/terraform/gke/variables.tf
ishustava: Great artists steal... 😄

ishustava (approved):
Looks great! A couple of minor comments, but non-blocking.

ndhanushkodi (approved):
Looks great Ashwin!!

PR #965, opened 2022-01-14
Add - separated names for config entry crds

lkysow:
For example allow both kubectl get ingressgateways and
kubectl get ingress-gateways.

When specified in .hcl or .json the config entries are dash-separated so
this makes it consistent if someone uses a dash by mistake after reading
our docs.

How I've tested this PR:
- edited a crd in my cluster to match and tested out the aliasname

- make sure I didn't typo or use the wrong name for a config entry
- also review if this is a good idea

Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/control-plane/api/v1alpha1/exportedservices_types.go
+++ b/control-plane/api/v1alpha1/exportedservices_types.go
@@ -29,6 +29,7 @@ func init() {
 // +kubebuilder:printcolumn:name="Synced",type="string",JSONPath=".status.conditions[?(@.type==\"Synced\")].status",description="The sync status of the resource with Consul"
 // +kubebuilder:printcolumn:name="Last Synced",type="date",JSONPath=".status.lastSyncedTime",description="The last successful synced time of the resource with Consul"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="The age of the resource"
+// +kubebuilder:resource:shortName="exported-services"
 type ExportedServices struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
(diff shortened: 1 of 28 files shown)


control-plane/api/v1alpha1/exportedservices_types.go
```diff
@@ -29,6 +29,7 @@ func init() {
 // +kubebuilder:printcolumn:name="Synced",type="string",JSONPath=".status.conditions[?(@.type==\"Synced\")].status",description="The sync status of the resource with Consul"
 // +kubebuilder:printcolumn:name="Last Synced",type="date",JSONPath=".status.lastSyncedTime",description="The last successful synced time of the resource with Consul"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="The age of the resource"
+// +kubebuilder:resource:shortName="exported-services"
```
thisisnotashwin, line 32: it's funny that it is the "short" name even though it is a character longer.

thisisnotashwin (commented):
Have the CRDs intentionally not been migrated to the helm chart as well? This change looks good!

kschoche (approved):
I did not catch any typos :D 
Looks great!

control-plane/api/v1alpha1/exportedservices_types.go
lkysow: 😂 I know I'm stretching the recommended usage... but they don't have an `aliases` field!

lkysow:
> Have the CRDs intentionally not been migrated to the helm chart as well? This change looks good!

Right... no I just forgot them!

thisisnotashwin (approved):
Love the "short" names ;) 
Great job @lkysow !! Hitting approve assuming the CRDs will be migrated into the helm charts too!!!

PR #981, opened 2022-01-19
Move make targets to top-level

lkysow:
* Move make targets from control-plane Makefile to top-level Makefile
* leave targets that are needed for CI
* remove charts/consul/Makefile because its test-docker target was never
  used
* add help text to all Makefile targets. You can see the help by running
  `make` or `make help`
* Add `control-plane-` prefix to control plane targets
* Modify targets that download Go tools to use `go install` instead of
  `go get` because using go get to download tools is deprecated
* modify the `ctrl-gen` target to also copy the generated CRDs to the chart

How I've tested this PR:
* ran each make target

* code

### DIFF
--- a/charts/consul/Makefile
+++ b/charts/consul/Makefile
@@ -1,6 +1,0 @@
-TEST_IMAGE?=consul-helm-test
-
-test-docker:
-	@docker build --rm -t '$(TEST_IMAGE)' -f $(CURDIR)/test/docker/Test.dockerfile $(CURDIR)
-
-.PHONY: test-docker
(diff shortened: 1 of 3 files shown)


charts/consul/Makefile
```diff
@@ -1,6 +0,0 @@
-TEST_IMAGE?=consul-helm-test
-
-test-docker:
-	@docker build --rm -t '$(TEST_IMAGE)' -f $(CURDIR)/test/docker/Test.dockerfile $(CURDIR)
-
-.PHONY: test-docker
```
lkysow, line 6: I don't think we ever use this

thisisnotashwin (approved):
bellissimo

t-eckert (approved):
LGTM

kschoche (approved):
Awesome, I can't wait to use this from the correct root directory :)

PR #982, opened 2022-01-20
Support the value `$POD_NAME` for the annotation

thisisnotashwin:
Changes proposed in this PR:
- Support the annotation `consul.hashicorp.com/service-meta-pod_name: $POD_NAME` where $POD_NAME will be replaced with the Pod's name. This mimics the support we had before the endpoints controller where environment variables were interpolated for tags.
- This PR only supports $POD_NAME for now because it's the only environment variable we've been asked to support.

How I've tested this PR:
- Unit tests.

- Code Review

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ IMPROVEMENTS:
   * Allow customization of `terminationGracePeriodSeconds` on the ingress gateways. [[GH-947](https://github.com/hashicorp/consul-k8s/pull/947)]
   * Support `ui.dashboardURLTemplates.service` value for setting [dashboard URL templates](https://www.consul.io/docs/agent/options#ui_config_dashboard_url_templates_service). [[GH-937](https://github.com/hashicorp/consul-k8s/pull/937)]
   * Allow using dash-separated names for config entries when using `kubectl`. [[GH-965](https://github.com/hashicorp/consul-k8s/pull/965)]
+* Control Plane
+  * Support the value `$POD_NAME` for the annotation `consul.hashicorp.com/service-meta-pod_name` that will now be interpolated and set to the pod name in the service's metadata. [[GH-982](https://github.com/hashicorp/consul-k8s/pull/982)]
 
 BUG FIXES:
 * Helm
--- a/control-plane/connect-inject/endpoints_controller_test.go
+++ b/control-plane/connect-inject/endpoints_controller_test.go
@@ -886,6 +886,7 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 				pod1.Annotations[annotationService] = "different-consul-svc-name"
 				pod1.Annotations[fmt.Sprintf("%sname", annotationMeta)] = "abc"
 				pod1.Annotations[fmt.Sprintf("%sversion", annotationMeta)] = "2"
+				pod1.Annotations[fmt.Sprintf("%spod_name", annotationMeta)] = "$POD_NAME"
 				pod1.Annotations[annotationTags] = "abc,123,$POD_NAME"
 				pod1.Annotations[annotationConnectTags] = "def,456,$POD_NAME"
 				pod1.Annotations[annotationUpstreams] = "upstream1:1234"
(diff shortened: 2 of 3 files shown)


control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -886,6 +886,7 @@ func TestReconcileCreateEndpoint(t *testing.T) {
 				pod1.Annotations[annotationService] = "different-consul-svc-name"
 				pod1.Annotations[fmt.Sprintf("%sname", annotationMeta)] = "abc"
 				pod1.Annotations[fmt.Sprintf("%sversion", annotationMeta)] = "2"
+				pod1.Annotations[fmt.Sprintf("%spod_name", annotationMeta)] = "$POD_NAME"
 				pod1.Annotations[annotationTags] = "abc,123,$POD_NAME"
```
kschoche, line 890: I see two tests referencing `$POD_NAME` did this never work to begin with or are the tests bogus?

thisisnotashwin: I think there was a `pod-name` vs `pod_name` situation with the keys. Additionally, i think the provide the env var `$POD_NAME` use different annotations and we are supporting both.

kschoche: ahhh, that makes sense!

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ IMPROVEMENTS:
   * Allow customization of `terminationGracePeriodSeconds` on the ingress gateways. [[GH-947](https://github.com/hashicorp/consul-k8s/pull/947)]
   * Support `ui.dashboardURLTemplates.service` value for setting [dashboard URL templates](https://www.consul.io/docs/agent/options#ui_config_dashboard_url_templates_service). [[GH-937](https://github.com/hashicorp/consul-k8s/pull/937)]
   * Allow using dash-separated names for config entries when using `kubectl`. [[GH-965](https://github.com/hashicorp/consul-k8s/pull/965)]
+* Control Plane
+  * Support the value `$POD_NAME` for the annotation `consul.hashicorp.com/service-meta-pod_name` that will now be interpolated and set to the pod name. [[GH-982](https://github.com/hashicorp/consul-k8s/pull/982)]
```
kschoche, line 13:
```suggestion
  * Support the value `$POD_NAME` for the annotation `consul.hashicorp.com/service-meta-pod_name` that will now be interpolated and set to the pod name in the service's metadata. [[GH-982](https://github.com/hashicorp/consul-k8s/pull/982)]
```

kschoche: something like the above to explain what actually happens?

kschoche (approved):
Looks great, left a minor comment which is 100% not blocking :)

ishustava (approved):
:shipit:

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ IMPROVEMENTS:
   * Allow customization of `terminationGracePeriodSeconds` on the ingress gateways. [[GH-947](https://github.com/hashicorp/consul-k8s/pull/947)]
   * Support `ui.dashboardURLTemplates.service` value for setting [dashboard URL templates](https://www.consul.io/docs/agent/options#ui_config_dashboard_url_templates_service). [[GH-937](https://github.com/hashicorp/consul-k8s/pull/937)]
   * Allow using dash-separated names for config entries when using `kubectl`. [[GH-965](https://github.com/hashicorp/consul-k8s/pull/965)]
+* Control Plane
+  * Support the value `$POD_NAME` for the annotation `consul.hashicorp.com/service-meta-pod_name` that will now be interpolated and set to the pod name in the service's metadata. [[GH-982](https://github.com/hashicorp/consul-k8s/pull/982)]
```
lkysow, line 13: @thisisnotashwin it's supported for any meta annotation right? Not just `service-meta-pod_name` but also `service-meta-foo` for example.

thisisnotashwin: not really. I asked @david-yu and the ask from the customer is to only support `$POD_NAME`.

jeenx: or would it make sense to add another value for `$POD_NAMESPACE`?

lkysow: It's also important to note that this isn't actually an env var. We're doing this in the endpoints controller that doesn't have access to the actual pod's environment variables. So we can only support data we actually have when looking at the pod yaml.

jeenx: Got it, would it be possible to add all that data that is available on the pod (e.g, `namespace`)?

Issue #719, opened 2021-09-07
Service tags containing a comma 


AndyL4N:
Consul uses comma-separated service tags notation.
But using traefik proxy with consul catalog I found some complications to setup multiple middlewares or entrypoints, required comma in value. For example:
```
'consul.hashicorp.com/service-tags':  |- 
      traefik.http.routers.router1.middlewares=middleware-redirect-https,middleware-ipwhitelist
```
Here Consul will split the second middleware to another tag
Of course, in consul terminology, this does not match the architecture. But the flexibility of the traefik-consul-kubernetes stack is very efficient. Maybe there is some workаround?

david-yu:
Hi @AndyL4N Could you tell us more about your use case? Have you enabled Connect or are you using mainly Catalog Sync? I believe likely do expect commas to separate tags, but I wonder if you attempted to escape that whole tag somehow with with a single or double quote?

AndyL4N:
Hi @david-yu. Currently we are only using Service Sync without Connect.  Sure, we tried many options to escape commas and tags with quotes, double quotes, braces and bracets, backslashes and even replacing with HTML code symbols. Same for escaping only tag values and whole tag line. As I see in listing above it has no sense because whole annotation tag string splits only by commas without any other logic. 

For example this annotations:

```  annotations:
    'consul.hashicorp.com/service-sync': 'true'
    'consul.hashicorp.com/service-tags':  |- 
     env=prod,
      solution=SupperApp,
      traefik.enable=true,
      traefik.http.middlewares.app-ipwhitelist.ipwhitelist.sourcerange=192.168.1.0/24, 
      traefik.http.middlewares.app-redirect-https.redirectscheme.scheme=https,
      traefik.http.routers.app-http.middlewares=app-redirect-https,app-ipwhitelist,
```
will be splitted to something like this:
<pre>
consul catalog services -tags | grep app
my-service-in-consul      <b>app-ipwhitelist</b>,env=prod,solution=SupperApp,traefik.enable=true,
                          traefik.http.middlewares.app-ipwhitelist.ipwhitelist.sourcerange=192.168.1.1/24,
                          traefik.http.middlewares.app-redirect-https.redirectscheme.scheme=https,
                          <b>traefik.http.routers.app-http.middlewares=app-redirect-https</b>
</pre>

Trying to escape whole tag with double quotes doing a similar thing:
<pre>
[...]

PR #983, opened 2022-01-20
Support commas in tag annotation.

lkysow:
Commas can be set by escaping with backslash. Fixes https://github.com/hashicorp/consul-k8s/issues/719

Example:

	annotation: "this,supports,commas\,now"

Becomes:

	["this", "supports", "commas,now"]

How I've tested this PR:
* unit tests

* code

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


t-eckert (approved):
Very clean! I like it!

ndhanushkodi (approved):
Looks good!!

erdanzhang:
@lkysow Thanks for fixing!!
However, when I tried to add service tags into my k8s services like 
```
template:
    metadata:
      annotations:
        consul.hashicorp.com/service-tags: mid\,dle,com\,ma
```

When I use curl to get the service tags, http://127.0.0.1:8500/v1/catalog/services
```
{
    "test-svc": [
        "dle",
        "com\\",
        "ma",
        "mid\\"
    ],
}
```

You can see the tag is incorrectly parsed. I suspect it may be related to the double slash \\, wondering if that is a known bug/issue, can I resolve it by upgrading the version or is there some clever workaround?

Thanks for the helping, this is current the blocker for our team to onboard using Consul :)

PR #993, opened 2022-01-25
Standardize resource names across the helm chart

thisisnotashwin:
Changes proposed in this PR:
- Rename helm resources that have redundant resource-name suffixes.
- Update `server-acl-init` as the secret name was hard-coded to use the `-svc-aaccount` suffix.

How I've tested this PR:
- Pipelines.

- 👀 

Checklist:
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/connect-inject-authmethod-clusterrole.yaml
+++ b/charts/consul/templates/connect-inject-authmethod-clusterrole.yaml
@@ -3,7 +3,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ template "consul.fullname" . }}-connect-injector-authmethod-role
+  name: {{ template "consul.fullname" . }}-connect-injector-authmethod
   labels:
     app: {{ template "consul.name" . }}
     chart: {{ template "consul.chart" . }}
--- a/charts/consul/templates/connect-inject-authmethod-clusterrole.yaml
+++ b/charts/consul/templates/connect-inject-authmethod-clusterrole.yaml
@@ -3,7 +3,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ template "consul.fullname" . }}-connect-injector-authmethod
+  name: {{ template "consul.fullname" . }}-connect-injector
   labels:
     app: {{ template "consul.name" . }}
     chart: {{ template "consul.chart" . }}
--- a/charts/consul/templates/connect-inject-authmethod-clusterrolebinding.yaml
+++ b/charts/consul/templates/connect-inject-authmethod-clusterrolebinding.yaml
@@ -3,7 +3,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "consul.fullname" . }}-connect-injector-authmethod-authdelegator
+  name: {{ template "consul.fullname" . }}-connect-injector-authdelegator
   labels:
     app: {{ template "consul.name" . }}
     chart: {{ template "consul.chart" . }}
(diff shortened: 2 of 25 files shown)


charts/consul/templates/webhook-cert-manager-configmap.yaml
```diff
@@ -17,10 +17,10 @@ data:
...
         "name": "{{ template "consul.fullname" . }}-connect-injector-cfg",
         "tlsAutoHosts": [
-          "{{ template "consul.fullname" . }}-connect-injector-svc",
-          "{{ template "consul.fullname" . }}-connect-injector-svc.{{ .Release.Namespace }}",
-          "{{ template "consul.fullname" . }}-connect-injector-svc.{{ .Release.Namespace }}.svc",
-          "{{ template "consul.fullname" . }}-connect-injector-svc.{{ .Release.Namespace }}.svc.cluster.local"
+          "{{ template "consul.fullname" . }}-connect-injector",
+          "{{ template "consul.fullname" . }}-connect-injector.{{ .Release.Namespace }}",
+          "{{ template "consul.fullname" . }}-connect-injector.{{ .Release.Namespace }}.svc",
+          "{{ template "consul.fullname" . }}-connect-injector.{{ .Release.Namespace }}.svc.cluster.local"
```
lkysow, line 23: need to think about what will happen during an upgrade with these certs and if there will be any downtime. Might be safer to keep this one as is.

thisisnotashwin: Wouldn't the upgrade cause both, the connect container as well and the cert-manager container to get recreated, in which case the upgrade should occur without downtime? Am I missing something?

lkysow: right so I guess:
- before upgrade svc and certs match the mutatingwebhookconfig
- upgrade modifies mutatingwebhookconfig to point at new svc object
- new svc has no pods yet (restarting) but that's okay
- inject pods and cert manager start
- cert manager mints new cert, updates mutatingwebhook config and inject pods pick up new secret

So it should all be fine, just a bit of downtime as those new connect inject pods start.

thisisnotashwin: yup! i expect it to be the same amount of downtime as any usual `upgrade`. Maybe a touch longer? But it should be fine.

charts/consul/templates/connect-inject-authmethod-clusterrole.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ template "consul.fullname" . }}-connect-injector-authmethod-role
+  name: {{ template "consul.fullname" . }}-connect-injector-authmethod
```
lkysow, line 6: I think these can all be `-connect-injector`

charts/consul/templates/connect-inject-authmethod-clusterrolebinding.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "consul.fullname" . }}-connect-injector-authmethod-authdelegator-role-binding
+  name: {{ template "consul.fullname" . }}-connect-injector-authmethod-authdelegator
```
lkysow, line 6: this can be `-connect-injector-authdelegator` since this is a special clusterrolebinding where we're binding to a preset role vs how we're usually binding to a role we've created. So this can be with the special name and the rest can be just the name of the component, i.e. `-connect-injector`

charts/consul/templates/connect-inject-authmethod-clusterrolebinding.yaml
```diff
@@ -16,13 +16,13 @@ roleRef:
...
   - kind: ServiceAccount
-    name: {{ template "consul.fullname" . }}-connect-injector-authmethod-svc-account
+    name: {{ template "consul.fullname" . }}-connect-injector-authmethod
     namespace: {{ .Release.Namespace }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "consul.fullname" . }}-connect-injector-authmethod-serviceaccount-role-binding
+  name: {{ template "consul.fullname" . }}-connect-injector-authmethod-serviceaccount
```
lkysow, line 25: and then this can just be `-connect-injector`

lkysow (commented):
* `enterprise-license-job` name should be `-enterprise-licence`

lkysow (commented):
I'm also thinking we should list this under breaking changes with a link to the PR and saying something like "if you're not referencing any consul components by names externally this won't be a breaking change"

lkysow (approved):
I'm irrationally stoked about this.

t-eckert (approved):
Beautifully done. No notes.

PR #1002, opened 2022-01-27
Fix Makefile Contrib Instructions

t-eckert:
Changes proposed in this PR:
- Fix the instructions for using the makefile now that the directives are in the root
- Change some wording to be more declarative
- Remove `$` from snippets so that the copy button gives the user something they can paste into their shell

How I've tested this PR:

Boy looks at computer approvingly and gives a thumbs up

Boy looks at computer approvingly and gives a thumbs up

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,31 +39,25 @@ You will also need to install the Docker engine:
 Clone the repository:
 
 ```shell
-$ git clone https://github.com/hashicorp/consul-k8s.git
+git clone https://github.com/hashicorp/consul-k8s.git
 ```
 
-Change directories into the appropriate folder:
+Compile the `consul-k8s-control-plane` binary for your local machine:
 
 ```shell
-$ cd control-plane
+make control-plane-dev
 ```
 
-To compile the `consul-k8s-control-plane` binary for your local machine:
-
-```shell
-$ make dev
-```
-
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -39,31 +39,25 @@ You will also need to install the Docker engine:
 Clone the repository:
 
 ```shell
-$ git clone https://github.com/hashicorp/consul-k8s.git
+git clone https://github.com/hashicorp/consul-k8s.git
```
thisisnotashwin, line 42: I think consul docs does start commandline commands with a `$`. I don't know if we have been consistent, but the desired pattern is to retain them.

thisisnotashwin (commented):
The changes look good. Just have one comment.

CONTRIBUTING.md
t-eckert: I talked about this with David a few months back. I think it would be best to move away from doing that because GitHub added a feature where code blocks have a button which will copy their contents. If we have the `$` in the code block then the user can't paste the command into their shell easily.

t-eckert: Now, for the Consul docs site this is different because it has a feature which omits the `$` when you copy the command.

thisisnotashwin: oh cool! i didnt know that. The rich diff view does not show it but it looks like it does.

thisisnotashwin (approved):

david-yu (approved):

PR #1007, opened 2022-01-28
Use kube svc dns for api gateway sds

lkysow:
Previously it was using the pod IP of the controller and then when
gateways were created they were hardcoded to use that pod IP.
This was an issue because if the controller pod dies then there will be
a new pod IP and the gateway will lose its connection to the SDS
service. Instead, this PR changes it to use the kube dns url.

How I've tested this PR:
- bats
- manually deployed gateway following https://learn.hashicorp.com/tutorials/consul/kubernetes-api-gateway and looked at sds url

- code

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -59,10 +59,6 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: status.hostIP
-        - name: IP
-          valueFrom:
-            fieldRef:
-              fieldPath: status.podIP
         {{- if .Values.global.acls.manageSystemACLs }}
         - name: CONSUL_HTTP_TOKEN
           valueFrom:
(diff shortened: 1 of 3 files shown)


andrewstucki (approved):

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -59,10 +59,6 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: status.hostIP
-        - name: IP
```
lkysow, line 62: @nickethier @andrewstucki  can you just double check this. I'm removing this env var because I think it was only used for `-sds-server-host`, just wanted to make sure it's not read inside the `consul-api-gateway server` command.

andrewstucki: Yep, it should be unused at this point.

PR #754, opened 2021-09-30
Lkysow/multiport

lkysow:
Changes proposed in this PR:
-
-

How I've tested this PR:

How I expect reviewers to test this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #1012, opened 2022-02-01
Multiport

ndhanushkodi:
Based off of research and POC by @lkysow https://github.com/hashicorp/consul-k8s/pull/754 Thank you!

Note to reviewers:
I recommend reading the acceptance test first, then looking at the implementation. In the acceptance tests, I've changed the signature for `CheckStaticServerConnection*` functions, so there's a lot of 1 line changes all over the acceptance tests due to that.

Changes proposed in this PR:

Support a workaround for multi port pods by registering a Consul service per port, and by injecting init containers and envoy sidecars per port.

- Acceptance test for multiport
  - The multiport app is just http-echo listening on different ports in 2 containers
  - First, the test deploys `static-client` and `multiport`, with static-client having `multiport` and `multiport-admin` as upstreams. It checks that `static-client` can make connections to each service-- `multiport`, and `multiport-admin`. This is to test inbound connections for a multiport app.
  - Then the test deploys `static-server`, and checks that `multiport` can make a connection to `static-server`. This is to test outbound connections from a multiport app. Note that there is only 1 intention because all upstream connections are configured and go through the 1st service's envoy proxy.
  
- Implementation for multiport
  - `handler`
    - uses the service name annotation to figure out if this is a multiport pod or not
    - inject a connect-init container and an envoy-sidecar container per port/service
    - mount all service account tokens not already on the pod. For any service whose service account is not already attached to the pod, this mounts a volume for that service account token. In container-init, these volumes are added as a container volumeMount so the service account token file can be passed to the connect-init command via `-bearer-token-file`
  - `container_init` -- update the template to pass in arguments specific to the multiport service
  - `envoy_sidecar` -- allow the bootstrap file path to have the service name in it for multiport services, and add `--base-id`  since there could be multiple envoy sidecars
  - `endpoints_controller`: the listener port should be 20000 for the first service, 20001 for the second, etc.
  - `connect_init` command: when its a multiport app, add a filter to poll by service name, use the service specific bearertoken, acl-token-sink, and proxyid files
  
How I've tested this PR:
acceptance tests and unit tests

see note above, and code review.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


acceptance/framework/k8s/deploy.go
```diff
@@ -16,8 +16,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
```
ndhanushkodi, line 18: Changed the function signature for CheckStaticServer* to allow passing in an `app` which is the app you're connecting from, and an `expectedSuccessOutput` which is the expected message back from the server. I needed to be able to make connections from the multiport app and not just static client, and each of the multiport services had a different success output.

acceptance/tests/connect/connect_helper.go
```diff
@@ -19,6 +19,7 @@ import (
 
 const staticClientName = "static-client"
```
ndhanushkodi, line 20: This file is function signature changes

acceptance/tests/connect/connect_inject_namespaces_test.go
```diff
@@ -177,9 +177,9 @@ func TestConnectInjectNamespaces(t *testing.T) {
 			if c.secure {
 				logger.Log(t, "checking that the connection is not successful because there's no intention")
 				if cfg.EnableTransparentProxy {
-					k8s.CheckStaticServerConnectionFailing(t, staticClientOpts, fmt.Sprintf("http://static-server.%s", staticServerNamespace))
```
ndhanushkodi, line 180: function signature changes

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -148,8 +150,169 @@ func TestConnectInject_RestartConsulClients(t *testing.T) {
...
 	logger.Log(t, "checking that connection is still successful")
 	if cfg.EnableTransparentProxy {
-		k8s.CheckStaticServerConnectionSuccessful(t, ctx.KubectlOptions(t), "http://static-server")
+		k8s.CheckStaticServerConnectionSuccessful(t, ctx.KubectlOptions(t), staticClientName, "http://static-server")
 	} else {
-		k8s.CheckStaticServerConnectionSuccessful(t, ctx.KubectlOptions(t), "http://localhost:1234")
+		k8s.CheckStaticServerConnectionSuccessful(t, ctx.KubectlOptions(t), staticClientName, "http://localhost:1234")
+	}
+}
+
```
ndhanushkodi, line 158: Adds the multiport acceptance test

acceptance/tests/fixtures/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM kschoche/http-echo:latest
```
ndhanushkodi, line 1: This adds curl to the http-echo image, I plan to swap out this for the actual http-echo image, so this can be ignored for now in review-- the only change would be to delete this file and swap the image.

acceptance/tests/fixtures/bases/multiport-app/anyuid-scc-rolebinding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
ndhanushkodi, line 1: Added a new multiport app-- this was much easier to read than trying to patch the existing static-server app. It just runs http-echo on two different ports with 2 different messages.

acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
```diff
@@ -128,7 +128,7 @@ func TestIngressGatewaySingleNamespace(t *testing.T) {
 				// via the bounce pod. It should fail to connect with the
```
ndhanushkodi, line 128: function signature changes

acceptance/tests/ingress-gateway/ingress_gateway_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
```
ndhanushkodi, line 15: Function signature changes

acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -127,7 +127,7 @@ func TestMeshGatewayDefault(t *testing.T) {
 	k8s.DeployKustomize(t, primaryContext.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-client-multi-dc")
 
 	logger.Log(t, "checking that connection is successful")
```
ndhanushkodi, line 129: Function signature changes

acceptance/tests/partitions/partitions_test.go
```diff
@@ -415,11 +415,11 @@ func TestPartitions(t *testing.T) {
 				if c.ACLsAndAutoEncryptEnabled {
 					logger.Log(t, "checking that the connection is not successful because there's no intention")
 					if cfg.EnableTransparentProxy {
```
ndhanushkodi, line 417: Function signature changes

acceptance/tests/terminating-gateway/terminating_gateway_namespaces_test.go
```diff
@@ -121,7 +121,7 @@ func TestTerminatingGatewaySingleNamespace(t *testing.T) {
```
ndhanushkodi, line 121: Function signature changes

acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -93,7 +93,7 @@ func TestTerminatingGateway(t *testing.T) {
 
 			// Test that we can make a call to the terminating gateway.
```
ndhanushkodi, line 94: Function signature changes

acceptance/tests/vault/vault_test.go
```diff
@@ -15,7 +15,8 @@ import (
 )
```
ndhanushkodi, line 16: Function signature changes

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -538,6 +538,383 @@ func TestProcessUpstreams(t *testing.T) {
...
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+
+			svcName := getServiceName(*tt.pod(), *tt.endpoint)
+			require.Equal(t, tt.expSvcName, svcName)
+
+		})
+	}
+}
+
```
ndhanushkodi, line 603: Kept this TestReconcileCreateEndpoint_MultiportService test separate from the existing create endpoint test because of the changes to the case structure that wouldn't be needed by the non-multiport tests. Also because this is a workaround, it might be nice to keep the code as separate as possible.

ndhanushkodi (commented):
Adding some comments for reviewers since theres a lot of files

control-plane/connect-inject/handler.go
```diff
@@ -225,21 +231,92 @@ func (h *Handler) Handle(ctx context.Context, req admission.Request) admission.R
...
+	annotatedSvcNames := h.annotatedServiceNames(pod)
+	multiPort := len(annotatedSvcNames) > 1
 
-	// Add the Envoy sidecar.
-	envoySidecar, err := h.envoySidecar(*ns, pod)
-	if err != nil {
-		h.Log.Error(err, "error configuring injection sidecar container", "request name", req.Name)
-		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("error configuring injection sidecar container: %s", err))
+	// For single port pods, add the single init container and envoy sidecar.
+	if len(annotatedSvcNames) == 0 || len(annotatedSvcNames) == 1 {
```
ishustava, line 240: I think that's equivalent?
```suggestion
	if len(annotatedSvcNames) < 2 {
```

ishustava: or maybe this could be `if !multiPort` and then the following `if multiPort` could be an `else`?

ishustava (commented):
This looks fantastic!!! Thanks you so much for comments throughout the code and the PR - this made it so easy to review ❤️ 

I left some comments and questions inline, mostly to clarify my own understanding and correct some minor things.

Great work on this!!

charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -49,5 +49,10 @@ rules:
   - {{ template "consul.fullname" . }}-connect-inject-acl-token
   verbs:
   - get
+- apiGroups: [""]
+  resources:
+    - serviceaccounts
+  verbs:
+    - get
```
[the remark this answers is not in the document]
ndhanushkodi: So since we're mounting the service account secret on the volume, the reason we're getting the service account in the handler is to [get the secret name](https://github.com/hashicorp/consul-k8s/pull/1012/files/9e7920439cc3b8f2e3aae447e06ef787dab2b480#diff-9edb67e94492a07a3800e8add0f97513e33e559038ec532ced787d9af7b63f09R284) (and to do validation that the service account exists). The secret name would be something like: `web-token-bvbk5` so we can't infer the whole name from the service account name, so I think we need this perm for that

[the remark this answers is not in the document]
ishustava: Ah ok! that makes sense. I thought for a second that we may be able to just mount the secret but you're right we can't infer it. I guess the only question is whether we should put it behind a feature flag since we only need it for multiport? I think it's also OK to leave as is and change if someone raises an issue.

acceptance/framework/k8s/deploy.go
```diff
@@ -77,10 +75,10 @@ func DeployKustomize(t *testing.T, options *k8s.KubectlOptions, noCleanupOnFailu
 // to be "hello world" in a case of success.
 // If expectSuccess is true, it will expect connection to succeed,
 // otherwise it will expect failure due to intentions.
-func CheckStaticServerConnection(t *testing.T, options *k8s.KubectlOptions, expectSuccess bool, failureMessages []string, curlArgs ...string) {
+func CheckStaticServerConnection(t *testing.T, options *k8s.KubectlOptions, app string, expectSuccess bool, failureMessages []string, expectedSuccessOutput string, curlArgs ...string) {
```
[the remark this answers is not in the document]
ndhanushkodi: yeah definitely good idea!

acceptance/framework/k8s/deploy.go
```diff
@@ -77,10 +75,10 @@ func DeployKustomize(t *testing.T, options *k8s.KubectlOptions, noCleanupOnFailu
 // to be "hello world" in a case of success.
```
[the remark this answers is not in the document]
ndhanushkodi: yup, updated!

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -148,8 +150,169 @@ func TestConnectInject_RestartConsulClients(t *testing.T) {
...
+const multiport = "multiport"
+const multiportAdmin = "multiport-admin"
+
+// Test that Connect works in a default and a secure installation.
```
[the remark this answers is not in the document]
ndhanushkodi: yup thanks for catching!

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -148,8 +150,169 @@ func TestConnectInject_RestartConsulClients(t *testing.T) {
...
+					Name: multiport,
+					Sources: []*api.SourceIntention{
+						{
+							Name:   staticClientName,
+							Action: api.IntentionActionAllow,
+						},
+					},
+				}, nil)
+				require.NoError(t, err)
+				logger.Log(t, fmt.Sprintf("creating intention for %s", staticServerAdminName))
```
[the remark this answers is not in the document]
ndhanushkodi: yup!

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -148,8 +150,169 @@ func TestConnectInject_RestartConsulClients(t *testing.T) {
...
+	cases := []struct {
+		secure      bool
+		autoEncrypt bool
+	}{
+		{false, false},
+		{true, false},
+		{true, true},
+	}
+
+	for _, c := range cases {
```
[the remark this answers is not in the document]
ndhanushkodi: Yeah, it was totally just running in the tproxy acceptance tests but using explicit upstreams-- added the skip, thanks!

control-plane/connect-inject/container_init.go
```diff
@@ -142,12 +149,42 @@ func (h *Handler) containerInit(namespace corev1.Namespace, pod corev1.Pod) (cor
 		TProxyExcludeUIDs:          splitCommaSeparatedItemsFromAnnotation(annotationTProxyExcludeUIDs, pod),
 		ConsulDNSClusterIP:         consulDNSClusterIP,
 		EnvoyUID:                   envoyUserAndGroupID,
+		MultiPort:                  multiPort,
+		EnvoyAddressPort:           20000 + mpi.serviceIndex,
```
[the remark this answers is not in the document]
ndhanushkodi: I think you are right, it might be redundant from what we're doing in endpoint controller for the proxy service registration, I'll try removing this and running acceptance tests!

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -538,6 +538,383 @@ func TestProcessUpstreams(t *testing.T) {
...
+							DestinationServiceID:   "pod1-web-admin",
+							LocalServiceAddress:    "127.0.0.1",
+							LocalServicePort:       9090,
+							Upstreams: []api.Upstream{
+								{
+									DestinationType: api.UpstreamDestTypeService,
+									DestinationName: "upstream1",
+									LocalBindPort:   1234,
+								},
+							},
```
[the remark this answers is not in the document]
ndhanushkodi: you're totally right-- I think it was still working as expected with both proxies configuring the same set of upstreams, but its redundant and a weird edge case and not as we planned to implement in the RFC. Great catch!! I'll update the test and implementation

control-plane/connect-inject/envoy_sidecar.go
```diff
@@ -10,19 +10,25 @@ import (
...
-func (h *Handler) envoySidecar(namespace corev1.Namespace, pod corev1.Pod) (corev1.Container, error) {
+func (h *Handler) envoySidecar(namespace corev1.Namespace, pod corev1.Pod, mpi multiPortInfo) (corev1.Container, error) {
 	resources, err := h.envoySidecarResources(pod)
 	if err != nil {
 		return corev1.Container{}, err
 	}
 
-	cmd, err := h.getContainerSidecarCommand(pod)
+	multiPort := mpi != multiPortInfo{}
+	cmd, err := h.getContainerSidecarCommand(pod, mpi.serviceName, mpi.serviceIndex)
```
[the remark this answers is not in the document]
ndhanushkodi: yup, it checks whether mpi.ServiceName is empty to determine how to configure the command!

control-plane/connect-inject/endpoints_controller_test.go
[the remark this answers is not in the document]
ndhanushkodi: Updated the implementation and test!

acceptance/framework/k8s/deploy.go
```diff
@@ -90,19 +88,24 @@ func CheckStaticServerConnection(t *testing.T, options *k8s.KubectlOptions, expe
 // If expectSuccess is true, it will expect connection to succeed,
 // otherwise it will expect failure due to intentions. If multiple failureMessages are provided it will assert
 // on the existence of any of them.
-func CheckStaticServerConnectionMultipleFailureMessages(t *testing.T, options *k8s.KubectlOptions, expectSuccess bool, failureMessages []string, curlArgs ...string) {
+func CheckStaticServerConnectionMultipleFailureMessages(t *testing.T, options *k8s.KubectlOptions, app string, expectSuccess bool, failureMessages []string, expectedSuccessOutput string, curlArgs ...string) {
```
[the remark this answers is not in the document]
ndhanushkodi: Nearly every call is setting "" and using the default, except for CheckStaticServerConnectionSuccessfulWithMessage, which is what we're using in the multiport test

charts/consul/templates/connect-inject-clusterrole.yaml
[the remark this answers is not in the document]
ndhanushkodi: Yeah, that sounds good, I'll go ahead and leave it then and see if it becomes an issue to allow this extra perm

control-plane/connect-inject/envoy_sidecar.go
```diff
@@ -10,19 +10,25 @@ import (
...
 
-func (h *Handler) envoySidecar(namespace corev1.Namespace, pod corev1.Pod) (corev1.Container, error) {
+func (h *Handler) envoySidecar(namespace corev1.Namespace, pod corev1.Pod, mpi multiPortInfo) (corev1.Container, error) {
 	resources, err := h.envoySidecarResources(pod)
 	if err != nil {
 		return corev1.Container{}, err
 	}
 
-	cmd, err := h.getContainerSidecarCommand(pod)
+	multiPort := mpi != multiPortInfo{}
```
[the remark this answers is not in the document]
ndhanushkodi: Updated!

control-plane/connect-inject/container_init.go
```diff
@@ -72,6 +72,11 @@ type initContainerCommandData struct {
 
 	// ConsulDNSClusterIP is the IP of the Consul DNS Service.
 	ConsulDNSClusterIP string
+
+	MultiPort        bool
+	EnvoyAdminPort   int
+	EnvoyAddressPort int
+	BearerTokenFile  string
```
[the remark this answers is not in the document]
ndhanushkodi: Yup thanks!

control-plane/connect-inject/handler.go
ndhanushkodi: made this an if !multiport + else!

acceptance/tests/fixtures/Dockerfile
ndhanushkodi: Swapped and deleted

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -148,8 +150,169 @@ func TestConnectInject_RestartConsulClients(t *testing.T) {
...
+				t.Cleanup(func() {
+					retrier := &retry.Timer{Timeout: 5 * time.Minute, Wait: 1 * time.Second}
+					retry.RunWith(retrier, t, func(r *retry.R) {
+						tokens, _, err := consulClient.ACL().TokenList(nil)
+						require.NoError(r, err)
+						for _, token := range tokens {
+							require.NotContains(r, token.Description, multiport)
+							require.NotContains(r, token.Description, multiportAdmin)
+							require.NotContains(r, token.Description, staticClientName)
+							require.NotContains(r, token.Description, staticServerName)
```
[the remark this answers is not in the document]
ishustava: Ah!!! that makes sense!

ndhanushkodi:
WANFed tests for vault are failing but it might be due to my image? I'll try rebasing and using a new image for acceptance tests, but since everything else is passing and this doesn't touch vault this is ready for a second round of review

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -1072,3 +1093,12 @@ func consulTags(pod corev1.Pod) []string {
...
 	return interpolatedTags
 }
+
+func getMultiPortIdx(pod corev1.Pod, serviceEndpoints corev1.Endpoints) int {
+	for i, name := range strings.Split(pod.Annotations[annotationService], ",") {
+		if name == getServiceName(pod, serviceEndpoints) {
+			return i
+		}
+	}
+	return -1
```
ishustava, line 1103: nit: I think we could also return an error here (it feels a bit more idiomatic to go) but definitely not a blocker!

ishustava (approved):
🎉 🎉 🎉 

A couple of minor comments, but they're not blocking!

control-plane/subcommand/connect-init/command.go
```diff
@@ -66,21 +67,16 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagConsulServiceNamespace, "consul-service-namespace", "", "Consul destination namespace of the service.")
 	c.flagSet.StringVar(&c.flagServiceAccountName, "service-account-name", "", "Service account name on the pod.")
 	c.flagSet.StringVar(&c.flagServiceName, "service-name", "", "Service name as specified via the pod annotation.")
+	c.flagSet.StringVar(&c.flagBearerTokenFile, "bearer-token-file", defaultBearerTokenFile, "Path to service account token file.")
+	c.flagSet.StringVar(&c.flagACLTokenSink, "acl-token-sink", defaultTokenSinkFile, "Service name as specified via the pod annotation.")
```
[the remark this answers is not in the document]
ndhanushkodi: omg good catch

control-plane/connect-inject/envoy_sidecar.go
```diff
@@ -62,7 +68,7 @@ func (h *Handler) envoySidecar(namespace corev1.Namespace, pod corev1.Pod) (core
 		// has only injected init containers so all containers defined in pod.Spec.Containers are from the user.
 		for _, c := range pod.Spec.Containers {
 			// User container and Envoy container cannot have the same UID.
-			if c.SecurityContext != nil && c.SecurityContext.RunAsUser != nil && *c.SecurityContext.RunAsUser == envoyUserAndGroupID {
+			if c.SecurityContext != nil && c.SecurityContext.RunAsUser != nil && *c.SecurityContext.RunAsUser == envoyUserAndGroupID && c.Image != h.ImageEnvoy {
```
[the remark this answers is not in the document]
ndhanushkodi: When we're running multiple sidecars in the pod, they clash with the same uid/gid so this check allows using the same uid/gid for 2 envoy sidecars. Since tproxy isn't supported it seems ok to let them run with the same uid/gid.

PR #1014, opened 2022-02-02
Allow API Gateway controller to update k8s Deployments + Services

nathancoleman:
The k8s client abstraction in consul-api-gateway [expects](https://github.com/hashicorp/consul-api-gateway/blob/c82e707fb30bebf1a88980568e0f67e318f65496/internal/k8s/gatewayclient/gatewayclient.go#L67-L68) to be able to create **or update** k8s `Deployments` and `Services`; however, the `ClusterRole` as defined today does not allow for updates to these resources. This prevents the Consul API Gateway controller from successfully installing a new gateway to the cluster.

Changes proposed in this PR:
- Modify `ClusterRole` to allow controller to update k8s `Deployment`
- Modify `ClusterRole` to allow controller to update k8s `Service`

How I've tested this PR:
Applied changes to deployment on GKE

Follow the [Learn tutorial](https://learn.hashicorp.com/tutorials/consul/kubernetes-api-gateway) for Consul API Gateway; however, the Consul Helm chart install needs to use the branch in this PR.

I believe the easiest way to do that is to check this branch out locally and `$ helm install ...` from there (replacing `$ helm install --values consul/config.yaml consul hashicorp/consul --version "0.40.0"` with the following):
```shell
# git checkout the branch from this PR locally
...
$ helm install --values consul/config.yaml consul <path/to>/consul-k8s/charts/consul
```

Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 BUG FIXES:
 * API Gateway
   * Fix issue where if the API gateway controller pods restarted, gateway pods would become disconnected from the secret discovery service. [[GH-1007](https://github.com/hashicorp/consul-k8s/pull/1007)]
+  * Fix issue where the API gateway controller could not update existing Deployments or Services. [[GH-1014](https://github.com/hashicorp/consul-k8s/pull/1014)]
 
 ## 0.40.0 (January 27, 2022)
 
--- a/charts/consul/templates/api-gateway-controller-clusterrole.yaml
+++ b/charts/consul/templates/api-gateway-controller-clusterrole.yaml
@@ -34,6 +34,7 @@ rules:
   - create
   - get
   - list
+  - update
   - watch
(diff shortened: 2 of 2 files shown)


andrewstucki (approved):
Clarification that this does allow for new gateways to be deployed, but subsequent modifications to the gateway that affect the deployment/service configurations such as an annotation change on the gateway or listener port changes while the gateway has an associated service deployment should fail without these permission updates.

Aside from me commenting on the scope of this, LGTM. Also linking in the related PR for the source of the generated RBAC policy: https://github.com/hashicorp/consul-api-gateway/pull/86

ishustava (approved):
LGTM. Would you mind also updating the Changelog?

nickethier (approved):
Thanks for catching this!

PR #1018, opened 2022-02-03
api-gateway: add node:read policy

mikemorris:
To avoid silently filtering all nodes when ACLs are enabled and returning an empty set when attempting to map routes to Consul services registered in consul-k8s

- Add `node_prefix "" { policy = "read" }` to the API Gateway template

**How I've tested this PR:**
- Manually applying this change fixed an error loop when deploying Consul API Gateway in GKE with ACLs enabled.

- Deploy Consul API Gateway via the Helm chart with the following config set
```
global:
  gossipEncryption:
    autoGenerate: true 
  tls:
    enabled: true
    enableAutoEncrypt: true
  acls:
    manageSystemACLs: true
```
- Attempt to apply a route to the gateway - it should fail to attach before this fix, but succeed with this change.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/control-plane/subcommand/server-acl-init/rules.go
+++ b/control-plane/subcommand/server-acl-init/rules.go
@@ -154,6 +154,9 @@ namespace_prefix "" {
     policy = "write"
     intentions = "write"
   }
+  node_prefix "" {
+    policy = "read"
+  }
 {{- if .EnableNamespaces }}
 }
 {{- end }}


andrewstucki (approved):
From the api gateway side of things, this LGTM, not sure if we want other k8s folks to sign off first though, also will need a Changelog entry.

control-plane/subcommand/server-acl-init/rules.go
```diff
@@ -154,6 +154,9 @@ namespace_prefix "" {
     policy = "write"
     intentions = "write"
   }
+  node_prefix "" {
+    policy = "read"
+  }
 {{- if .EnableNamespaces }}
```
andrewstucki, line 160: Just for posterity sake, this enables [this particular api call](https://github.com/hashicorp/consul-api-gateway/blob/4c6ee19207474ea318a96bf96bce6cbe5b8248b4/internal/k8s/service/resolver.go#L288). Without the ACL, the call returns an empty node set and the controller is unable to resolve any Consul services needed in an HTTPRoute configuration.

nathancoleman:
I've requested review from @hashicorp/consul-eco-platform since that's who I was pointed to yesterday

nathancoleman (approved):

ndhanushkodi (approved):
LGTM!

PR #1019, opened 2022-02-03
Smoke tests for the CLI `upgrade` command

t-eckert:
## Changes proposed in this PR

- Modify the "Upgrade" method of the `CLICluster` to use the CLI instead of Helm
- Add some common methods to the `CLICluster` to reduce code duplication between install and upgrade
- Change "self" reference in `CLICluster` from `h` to `c` to help differentiate it from the `HelmCluster` struct.
- Change `ConnectInjectConnectivityCheck` from a function to the struct `ConnectHelper` (explained below).
- Move `MergeMaps` from `consul` to `helpers` so it can be found more easily.
- Merge CLI tests for connect injection into existing Helm-based connect injection tests.

## Why change `ConnectInjectConnectivityCheck` to `ConnectHelper`

The function `ConnectInjectConnectivityCheck` takes a configuration for ACLs and TLS and tests how the Consul cluster behaves when `connectInject.enabled` is `true`. It both creates the cluster/client and runs a series of checks against it. For the upgrade tests, I want to create the cluster/client, run the checks, upgrade the cluster/client, then run the checks again.

For this case, it was useful for me to separate creating and upgrading the cluster/client from running the checks. I could have done this by separating the "checks" part of the function into another function and having `ConnectInjectConnectivityCheck` call that. While I did separate out the "checks" part of the function into `testConnectInject`, I also inverted the `ConnectInjectConnectivityCheck` to be a `struct` which takes the configuration previously passed into the function. I did this to reduce the arity of the functions called by tests and the `testConnectInject` function. There is a lot of shared state between the two functions (which is likely why they were a single function before) and it made sense to me to contain that within a shared struct.

## How I've tested this PR

Running the existing tests repeatedly and carefully to ensure changing them does not break the testing suite.

## How I expect reviewers to test this PR

The related tests in the acceptance test suite can be run with the following commands.

``` bash
go test -timeout 15m -run ^TestConnectInject$ github.com/hashicorp/consul-k8s/acceptance/tests/connect -v
```

``` bash
go test -timeout 15m -run ^TestConnectInjectOnUpgrade$ github.com/hashicorp/consul-k8s/acceptance/tests/connect -v
```

Checklist:
- [x] Tests added
- [ ] ~~CHANGELOG entry added~~


acceptance/tests/connect/connect_inject_test.go
```diff
@@ -16,25 +18,147 @@ import (
...
+	}{
+		"CLI upgrade changes nothing": {
+			clusterGen:  consul.NewCLICluster,
+			releaseName: consul.CLIReleaseName,
+		},
+		"CLI upgrade from 1.10 to 1.11": {
+			clusterGen:  consul.NewCLICluster,
+			releaseName: consul.CLIReleaseName,
+			initial: map[string]string{
+				"global.image": "hashicorp/consul:1.10.7",
```
ndhanushkodi, line 102: I think we need to not hardcode these versions-- you could automatically fetch the two most recent consul versions to perform the upgrade. The versions could be fetched in a step of the pipeline using the github api, and passed into the acceptance tests as flags.

Another option we had discussed before for this test was grabbing the n-1 version of the CLI itself to install with and upgrading to the latest version of the CLI. This would possibly involve a pipeline step like above, to fetch the n-1 version of the CLI. This may not even need the github api, and could parse a previous version from the changelog, and use `apt` to fetch the CLI at that version in the pipeline step. Whenever we upgrade the CLI itself, we'd be bumping the consul-k8s image as part of that upgrade so it'd trigger a decent change to test.

t-eckert: I added a way to fetch the versions so I can make this test dynamic. I also will need to write that test that uses 1 version of the CLI then another to upgrade. I think that is a separate PR. I'll make a task for it.

ndhanushkodi: awesome thanks for working on this! I think the upgrade case with CLI versions could be a part of this PR since it's an upgrade smoke test case, but if not, I think it'd be the highest priority subtask within upgrade tests since I'd consider the upgrade smoke tests task to encompass both these cases.

t-eckert: Should I have it exercise the same connect inject behavior as this collection of tests?

t-eckert: I'll close this conversation and add a note to do the upgrade of the CLI itself.

t-eckert:
- [ ] Outstanding task:

> Another option we had discussed before for this test was grabbing the n-1 version of the CLI itself to install with and upgrading to the latest version of the CLI. This would possibly involve a pipeline step like above, to fetch the n-1 version of the CLI. This may not even need the github api, and could parse a previous version from the changelog, and use apt to fetch the CLI at that version in the pipeline step. Whenever we upgrade the CLI itself, we'd be bumping the consul-k8s image as part of that upgrade so it'd trigger a decent change to test.

acceptance/tests/connect/connect_inject_test.go
ndhanushkodi: Given eng sync today, I think it makes sense to not include upgrading the CLI itself (bc it could result in a consul image change). I thought eventually clients and servers could come up with a one-shot upgrade but it seems thats not the case! Sorry about that!

acceptance/framework/consul/cli_cluster.go
```diff
@@ -102,92 +105,94 @@ func NewCLICluster(
...
-	for k, v := range h.values {
-		args = append(args, "-set", fmt.Sprintf("%s=%s", k, v))
+	k8s.WaitForAllPodsToBeReady(t, c.kubernetesClient, consulNS, fmt.Sprintf("release=%s", c.releaseName))
+}
+
+// Upgrade uses the `consul-k8s upgrade` command to upgrade a Consul cluster.
+func (c *CLICluster) Upgrade(t *testing.T, helmValues map[string]string) {
+	t.Helper()
+
+	k8s.WritePodsDebugInfoIfFailed(t, c.kubectlOptions, c.debugDirectory, "release="+c.releaseName)
```
ndhanushkodi, line 145: If there are failed pods at this point, it'd be from the install before an upgrade. Should we potentially abort the upgrade at this point and return out of this function and fail the test?

acceptance/tests/connect/connect_helper.go
```diff
@@ -17,40 +17,83 @@ import (
...
+	switch c.ClusterKind {
+	case consul.Helm:
+		c.consulCluster = consul.NewHelmCluster(c.T, c.helmValues(), c.Ctx, c.Cfg, c.ReleaseName)
+	case consul.CLI:
+		c.consulCluster = consul.NewCLICluster(c.T, c.helmValues(), c.Ctx, c.Cfg, c.ReleaseName)
 	}
+}
 
-	consulCluster.Create(t)
+// Install creates a new cluster using the ClusterGenerator function then runs its Create method
```
ndhanushkodi, line 65: What's the ClusterGenerator this is referring to? Was just wondering if the name changed!

acceptance/tests/connect/connect_helper.go
```diff
@@ -97,31 +148,62 @@ func ConnectInjectConnectivityCheck(t *testing.T, ctx environment.TestContext, c
...
-		k8s.CheckStaticServerConnectionMultipleFailureMessages(t, ctx.KubectlOptions(t), false, []string{"curl: (56) Recv failure: Connection reset by peer", "curl: (52) Empty reply from server", "curl: (7) Failed to connect to static-server port 80: Connection refused"}, "http://static-server")
+	logger.Log(c.T, "checking that connection is unsuccessful")
+	if c.Cfg.EnableTransparentProxy {
+		k8s.CheckStaticServerConnectionMultipleFailureMessages(c.T, c.Ctx.KubectlOptions(c.T), false, []string{"curl: (56) Recv failure: Connection reset by peer", "curl: (52) Empty reply from server", "curl: (7) Failed to connect to static-server port 80: Connection refused"}, "http://static-server")
 	} else {
-		k8s.CheckStaticServerConnectionMultipleFailureMessages(t, ctx.KubectlOptions(t), false, []string{"curl: (56) Recv failure: Connection reset by peer", "curl: (52) Empty reply from server"}, "http://localhost:1234")
+		k8s.CheckStaticServerConnectionMultipleFailureMessages(c.T, c.Ctx.KubectlOptions(c.T), false, []string{"curl: (56) Recv failure: Connection reset by peer", "curl: (52) Empty reply from server"}, "http://localhost:1234")
+	}
+
+	// Return the static-server to a "healthy state
```
ndhanushkodi, line 186:
```suggestion
	// Return the static-server to a "healthy state".
```

ndhanushkodi (commented):
Looks great Thomas! Just had a few minor questions/clarifications

acceptance/framework/consul/cli_cluster.go
t-eckert: Great thought! I added the check below this to abort if the test is failing. Let me know what you think.

ndhanushkodi: Yeah that looks good!

ndhanushkodi (approved):
Nice work Thomas!!

acceptance/tests/connect/connect_helper.go
```diff
@@ -17,40 +17,85 @@ import (
...
+	logger.Log(c.T, "Installing Consul cluster")
+	c.consulCluster.Create(c.T)
+	c.consulClient = c.consulCluster.SetupConsulClient(c.T, c.Secure)
+}
+
+// Upgrade uses the existing Consul cluster and upgrades it using Helm values
+// set by the Secure, AutoEncrypt, and HelmValues fields.
+func (c *ConnectHelper) Upgrade() {
+	require.True(c.T, c.consulCluster != nil, "consulCluster must be set before calling Upgrade (Call Install first).")
+	require.True(c.T, c.consulClient != nil, "consulClient must be set before calling Upgrade (Call Install first).")
```
thisisnotashwin, line 79: you can use `require.NotNil` here instead of true. It will make the assertion more readable.

acceptance/tests/connect/connect_helper.go
```diff
@@ -17,40 +17,85 @@ import (
...
+	// HelmValues are the additional helm values to use when installing or
+	// upgrading the cluster beyond connectInject.enabled, global.tls.enabled,
+	// global.tls.enableAutoEncrypt, global.acls.mangageSystemACLs which are
+	// set by the Secure and AutoEncrypt fields.
+	HelmValues map[string]string
+
+	// RelaseName is the name of the Consul cluster.
+	ReleaseName string
+
+	T   *testing.T
```
thisisnotashwin, line 46: Can `testing.T` be something that each of the methods receives as an argument? This way, one could send in a `retry.R` which adheres to the same interface in-case one of the methods needs to be run until it succeeds. Also, it allows for some context-based de-coupling of `T`.
eg. Setup(t *testing.T)

acceptance/tests/connect/connect_helper.go
```diff
@@ -97,31 +153,76 @@ func ConnectInjectConnectivityCheck(t *testing.T, ctx environment.TestContext, c
...
+
+	helpers.MergeMaps(helmValues, c.HelmValues)
+
+	return helmValues
+}
+
+// deployFixtureCase looks up the fixture case by name in the `../fixtures/cases`
+// directory and deploys it to the Kubernetes cluster.
+func (c *ConnectHelper) deployFixtureCase(name string) {
+	k8s.DeployKustomize(c.T, c.Ctx.KubectlOptions(c.T), c.Cfg.NoCleanupOnFailure, c.Cfg.DebugDirectory, "../fixtures/cases/"+name)
```
thisisnotashwin, line 227: I think this can be inlined. IMO given all it is doing is deploying Kustomize, it is easier to read as it rather than extracting it into a method.

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -18,25 +18,135 @@ import (
...
+	}
+}
+
+// TestConnectInjectOnUpgrade tests that Connect works before and after an
+// upgrade is performed on the cluster.
+func TestConnectInjectOnUpgrade(t *testing.T) {
+	cases := map[string]struct {
+		clusterKind         consul.ClusterKind
+		releaseName         string
+		secure, autoEncrypt bool
```
thisisnotashwin, line 94: These seem to be unused in the tests below

thisisnotashwin (commented):
Hey @t-eckert. Great job on the tests. I have left some comments to address!!

acceptance/tests/connect/connect_helper.go
t-eckert: Yeah, I like that better. I think it's more consistent with existing code too.

acceptance/tests/connect/connect_inject_test.go
t-eckert: Good catch

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -18,25 +18,130 @@ import (
...
+				Ctx:         ctx,
+				Cfg:         cfg,
+			}
+
+			connHelper.Setup(t)
+
+			connHelper.Install(t)
+			connHelper.DeployClientAndServer(t)
+			connHelper.TestConnectionFailureWithoutIntention(t)
+			connHelper.CreateIntention(t)
```
thisisnotashwin, line 80: Can we extract the `if c.secure()` logic from inside create intention to outside it?
ie 
if c.secure {
  connHepler.CreateIntention(t)
}

I assumed that we were always creating intentions when I initially read it.

thisisnotashwin: Similarly, move the above assertion about connection failure without intention as well. It is an assertion that only runs when secure is true, but the way it is structured, it feels like it always runs that assertion.

thisisnotashwin (commented):
This looks great! have one more suggestion for how we do secure tests but we should be good after that.

acceptance/tests/connect/connect_inject_test.go
t-eckert: I thought about the same thing too and was on the fence. Looking at it again, I think you are right on this.

thisisnotashwin: yeah! my rule of thumb is usually to frontload any information that someone working with this code might need to know without having to dive into the method definition or differently put, avoid surprises.

thisisnotashwin (approved):
💥 nicely done

PR #1023, opened 2022-02-07
Correcting connect-init documentation that said it creates the service rather than listening for so it can extract the proxy id for bootstrapping envoy.

jmurret:
Changes proposed in this PR:
- modified comments to reflect that connect init does not register the service but rather polls for it soit can use the proxy-id for Envoy bootstrapping

How I've tested this PR:
N/A

N/A

### DIFF
--- a/control-plane/connect-inject/container_init.go
+++ b/control-plane/connect-inject/container_init.go
@@ -104,8 +104,8 @@ func (h *Handler) initCopyContainer() corev1.Container {
 	return container
 }
 
-// containerInit returns the init container spec for registering the Consul
-// service, setting up the Envoy bootstrap, etc.
+// containerInit returns the init container spec for that polls for the service and the connect proxy service to be registered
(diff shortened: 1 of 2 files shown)


control-plane/connect-inject/container_init.go
```diff
@@ -104,8 +104,8 @@ func (h *Handler) initCopyContainer() corev1.Container {
 	return container
 }
 
-// containerInit returns the init container spec for registering the Consul
-// service, setting up the Envoy bootstrap, etc.
+// containerInit returns the init container spec for that polls for the service and the connect proxy service to be registered
+// so that it can save theproxy service id to the shared volume and boostrap Envoy with the proxy-id
```
ishustava, line 108:
```suggestion
// so that it can save the proxy service ID to the shared volume and boostrap Envoy with that proxy-id.
```

ishustava (approved):
🙏

PR #1025, opened 2022-02-08
vault: add support for wan federation with Vault when ACLs are enabled

ishustava:
Changes proposed in this PR:
- This PR adds support for storing ACL replication token in Vault

How I've tested this PR:
acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


.circleci/config.yml
```diff
@@ -70,7 +70,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:latest"
+        default: "ishustava/consul-k8s-dev:02-08-2022-d21554d8"
```
ishustava, line 73: todo: remove before merging

charts/consul/templates/server-acl-init-job.yaml
```diff
@@ -4,7 +4,9 @@
 {{- if and .Values.global.acls.createReplicationToken (not .Values.global.acls.manageSystemACLs) }}{{ fail "if global.acls.createReplicationToken is true, global.acls.manageSystemACLs must be true" }}{{ end -}}
 {{- if .Values.global.bootstrapACLs }}{{ fail "global.bootstrapACLs was removed, use global.acls.manageSystemACLs instead" }}{{ end -}}
 {{- if .Values.global.acls.manageSystemACLs }}
-{{- /* We don't render this job when server.updatePartition > 0 because that
+{{- if and .Values.global.secretsBackend.vault.enabled .Values.global.acls.replicationToken.secretName (not .Values.global.secretsBackend.vault.manageSystemACLsRole) }}{{ fail "global.secretsBackend.vault.manageSystemACLsRole must be set if global.secretsBackend.vault.enabled is true and global.acls.replicationToken is provided" }}{{ end -}}
+{{- if or (and .Values.global.acls.replicationToken.secretName (not .Values.global.acls.replicationToken.secretKey))  (and .Values.global.acls.replicationToken.secretKey (not .Values.global.acls.replicationToken.secretName))}}{{ fail "both global.acls.replicationToken.secretKey and global.acls.replicationToken.secretName must be set if one of them is provided" }}{{ end -}}
```
ishustava, line 8: These simplify logic and testing across our templates since we don't have to test for the combinations when replication token secret name is provided but not secret key and vice versa.

kschoche (approved):
🔥

acceptance/framework/consul/cli_cluster.go
```diff
@@ -51,7 +51,7 @@ func NewCLICluster(
 	ctx environment.TestContext,
 	cfg *config.TestConfig,
 	releaseName string,
-) Cluster {
+) *CLICluster {
```
t-eckert, line 54: 🙏🏻 Thank you!

acceptance/framework/consul/consul_cluster.go
```diff
@@ -47,6 +47,8 @@ type HelmCluster struct {
 	noCleanupOnFailure bool
 	debugDirectory     string
 	logger             terratestLogger.TestLogger
+
+	ACLToken string
```
t-eckert, line 51: Can we move this public field to the top of the struct?

t-eckert: Oh, also a nit, but what do you think of `ACLTokenOverride` to communicate that there is behavior whether this is set or not. Maybe add a comment that this is optional. Right now, someone using this would have to read the implementation in order to know that the cluster has behavior with and without this value being set.

ishustava: I think I prefer to keep it as `ACLToken` because it's also not an override. It's more of an optional value that you can set if you don't want us to get the token from the kube secret. 

I've added docs to clarify what it should be used for.

t-eckert: Cool, I dig that. I'll use that pattern in my work in the codebase in the future too.

t-eckert (approved):
Spectacular! I always enjoy how clear your code is, Iryna. Great work.

PR #1032, opened 2022-02-14
Add support for Enterprise License to be configured in Vault

jmurret:
Changes proposed in this PR:
For server-statefulset, client-agent-snapshot-deployment, and client-daemonset:
- when vault is enabled as secrets backend, allow default behavior of retrieving kubernetes secret for enterprise license
- when vault is enabled as secrets backend
  - use vault injectors to get retrieve the license from vault and write it to disk
  - use the license from disk to write the license value to the environment variable for the license.  (server-statefulset and client-daemonset use CONSUL_LICENSE_PATH and client-agent-snapshot-deployment uses ENTERPRISE_LICENSE).

How I've tested this PR:
- [x] unit tests
- [ ] acceptance tests
- [ ] manual verification

- review code and test changes

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added
- [ ] Acceptance tests added
- [ ] Docs PR todescribe how to enable this behavior

### DIFF
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -15,7 +15,7 @@ as well as the global.name setting.
 {{- end -}}
 {{- end -}}
 
-{{- define "consul.vaultGossipTemplate" -}}
+{{- define "consul.vaultSecretTemplate" -}}
  |
             {{ "{{" }}- with secret "{{ .secretName }}" -{{ "}}" }}
             {{ "{{" }}- {{ printf ".Data.data.%s" .secretKey }} -{{ "}}" }}
--- a/charts/consul/templates/server-statefulset.yaml
+++ b/charts/consul/templates/server-statefulset.yaml
@@ -66,7 +66,7 @@ spec:
         {{- if .Values.global.gossipEncryption.secretName }}
         {{- with .Values.global.gossipEncryption }}
         "vault.hashicorp.com/agent-inject-secret-gossip.txt": "{{ .secretName }}"
-        "vault.hashicorp.com/agent-inject-template-gossip.txt": {{ template "consul.vaultGossipTemplate" . }}
+        "vault.hashicorp.com/agent-inject-template-gossip.txt": {{ template "consul.vaultSecretTemplate" . }}
         {{- end }}
         {{- end }}
         {{- if .Values.server.serverCert.secretName }}
@@ -79,6 +79,12 @@ spec:
         {{- end }}
         {{- if .Values.global.secretsBackend.vault.agentAnnotations }}
         {{ tpl .Values.global.secretsBackend.vault.agentAnnotations . | nindent 8 | trim }}
+        {{- end }}
+        {{- if .Values.global.enterpriseLicense.secretName }}
+        {{- with .Values.global.enterpriseLicense }}
+        "vault.hashicorp.com/agent-inject-secret-enterpriselicense.txt": "{{ .secretName }}"
+        "vault.hashicorp.com/agent-inject-template-enterpriselicense.txt": {{ template "consul.vaultSecretTemplate" . }}
+        {{- end }}
         {{- end }}
         {{- end }}
         "consul.hashicorp.com/connect-inject": "false"
@@ -133,7 +139,7 @@ spec:
             secretName: {{ template "consul.fullname" . }}-server-cert
             {{- end }}
         {{- end }}
-        {{- if (and .Values.global.enterpriseLicense.secretName .Values.global.enterpriseLicense.secretKey .Values.global.enterpriseLicense.enableLicenseAutoload) }}
+        {{- if (and .Values.global.enterpriseLicense.secretName .Values.global.enterpriseLicense.secretKey .Values.global.enterpriseLicense.enableLicenseAutoload (not .Values.global.secretsBackend.vault.enabled)) }}
         - name: consul-license
           secret:
             secretName: {{ .Values.global.enterpriseLicense.secretName }}
@@ -215,7 +221,7 @@ spec:
             - name: CONSUL_CACERT
               value: /consul/tls/ca/tls.crt
             {{- end }}
-            {{- if (and .Values.global.enterpriseLicense.secretName .Values.global.enterpriseLicense.secretKey .Values.global.enterpriseLicense.enableLicenseAutoload) }}
+            {{- if (and .Values.global.enterpriseLicense.secretName .Values.global.enterpriseLicense.secretKey .Values.global.enterpriseLicense.enableLicenseAutoload (not .Values.global.secretsBackend.vault.enabled)) }}
             - name: CONSUL_LICENSE_PATH
               value: /consul/license/{{ .Values.global.enterpriseLicense.secretKey }}
             {{- end }}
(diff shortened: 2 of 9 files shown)


jmurret:
I'm most curious to talk through the VolumeMounts, how I should assume they will work, and whether those need to be modified.

jmurret:
I've also found these [failure validations for gossipEncryption key](https://github.com/hashicorp/consul-k8s/blob/f3428c23cf010b08e20397add4a1cf3cbdae3bfd/charts/consul/test/unit/server-statefulset.bats#L916) that I also implemented for [enterpriseLicense in this PR](https://github.com/hashicorp/consul-k8s/pull/1032/files#diff-823c35dc294a7bcf137ed1d5231e19c1dbe4dfe0ceacc9afc0cfb632effde5aeR16-R17).

This has caused these [two tests](https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/test/unit/server-acl-init-job.bats#L219-L239) to fail.  So, I amwondering:
- should I modify the old tests?  modify the added logic? just forgo this logic in the PR?
- should these failure validations also be in client-daemonset and client-snapshot-agent-deployment? (likewise client-daemonset does not have the validations for gossip key).

charts/consul/templates/_helpers.tpl
```diff
@@ -15,7 +15,7 @@ as well as the global.name setting.
 {{- end -}}
 {{- end -}}
 
-{{- define "consul.vaultGossipTemplate" -}}
+{{- define "consul.vaultSecretTemplate" -}}
```
jmurret, line 18: Made this name more generic because we are also using it for enterpriseLicense.

charts/consul/test/unit/enterprise-license-job.bats
```diff
@@ -29,24 +29,6 @@ load _helpers
       .
 }
 
-@test "enterpriseLicense/Job: disabled when secretName is missing" {
```
jmurret, line 32: We have a hard failure in [server-statefulset](https://github.com/hashicorp/consul-k8s/pull/1032/files#diff-3b78b7152b597d177763b2ffdad9fab0bde0b6e1a0c1080b439a5fa7121ab794R1378-R1398) when only one of `global.enterpriseLicense.secretKey` or `global.enterpriseLicense.secretName` are provided.
These tests are redundant and were only soft failing by not including that configuration.  Now the overall helm chart install/upgrade/apply will fail if these conditions are true.

charts/consul/test/unit/enterprise-license-podsecuritypolicy.bats
```diff
@@ -30,24 +30,6 @@ load _helpers
       .
 }
 
-@test "enterpriseLicense/PodSecurityPolicy: disabled when ent secretName missing" {
```
jmurret, line 33: See above comment about removing redundant tests.

charts/consul/test/unit/enterprise-license-role.bats
```diff
@@ -29,24 +29,6 @@ load _helpers
       .
 }
 
-@test "enterpriseLicense/Role: disabled when ent secretName missing" {
```
jmurret, line 32: See above comment about removing redundant tests.

charts/consul/test/unit/enterprise-license-rolebinding.bats
```diff
@@ -29,24 +29,6 @@ load _helpers
       .
 }
 
-@test "enterpriseLicense/RoleBinding: disabled when ent secretName missing" {
```
jmurret, line 32: See above comment about removing redundant tests.

charts/consul/test/unit/enterprise-license-serviceaccount.bats
```diff
@@ -29,24 +29,6 @@ load _helpers
       .
 }
 
-@test "enterpriseLicense/ServiceAccount: disabled when ent secretName missing" {
```
jmurret, line 32: See above comment about removing redundant tests.

charts/consul/test/unit/server-acl-init-job.bats
```diff
@@ -216,28 +216,6 @@ load _helpers
   [ "${actual}" = "true" ]
 }
 
-@test "serverACLInit/Job: ent license acl option disabled missing global.enterpriseLicense.secretName" {
```
jmurret, line 219: See above comment about removing redundant tests.

acceptance/tests/vault/vault_test.go
```diff
@@ -163,9 +186,6 @@ func TestVault(t *testing.T) {
 	require.NoError(t, err)
 
 	consulHelmValues := map[string]string{
-		// TODO: Update the global image once 1.11 is GA.
```
jmurret, line 166: Removed as this is no longer needed and also prevents us from running -enable-enterprise tests with the proper image.

acceptance/tests/vault/vault_test.go
```diff
@@ -20,6 +21,11 @@ path "consul/data/secret/gossip" {
   capabilities = ["read"]
 }`
 
+	enterpriseLicensePolicy = `
+path "consul/data/secret/enterpriselicense" {
```
kschoche, line 25: Super minor nit, but I was thinking that this might read better as `enterprise-license` && `consul-enterprise-license` throughout. No changes required unless you feel the need :)

kschoche: ^ I can see why you made this choice though since all of the previous examples are single words like `consul-server`, etc :)

charts/consul/templates/server-statefulset.yaml
```diff
@@ -13,6 +13,8 @@
...
 {{- if (and (and .Values.global.secretsBackend.vault.enabled .Values.global.tls.enabled) (not .Values.global.tls.enableAutoEncrypt)) }}{{ fail "global.tls.enableAutoEncrypt must be true if global.secretsBackend.vault.enabled=true and global.tls.enabled=true" }}{{ end -}}
 {{- if (and (and .Values.global.secretsBackend.vault.enabled .Values.global.tls.enabled) (not .Values.global.secretsBackend.vault.consulCARole)) }}{{ fail "global.secretsBackend.vault.consulCARole must be provided if global.secretsBackend.vault.enabled=true and global.tls.enabled=true" }}{{ end -}}
+{{- if (and .Values.global.enterpriseLicense.secretName (not .Values.global.enterpriseLicense.secretKey)) }}{{fail "enterpriseLicense.secretKey and secretName must both be specified." }}{{ end -}}
+{{- if (and (not .Values.global.enterpriseLicense.secretName) .Values.global.enterpriseLicense.secretKey) }}{{fail "enterpriseLicense.secretKey and secretName must both be specified." }}{{ end -}}
```
ishustava, line 17: I think we might want to add these failures to the consul client daemon set as well because it's possible to deploy consul clients without servers (sorry! I think I might have said otherwise in our 1:1)

charts/consul/templates/server-statefulset.yaml
```diff
@@ -131,7 +139,7 @@ spec:
             secretName: {{ template "consul.fullname" . }}-server-cert
             {{- end }}
         {{- end }}
-        {{- if (and .Values.global.enterpriseLicense.secretName .Values.global.enterpriseLicense.secretKey .Values.global.enterpriseLicense.enableLicenseAutoload) }}
+        {{- if (and .Values.global.enterpriseLicense.secretName .Values.global.enterpriseLicense.secretKey .Values.global.enterpriseLicense.enableLicenseAutoload (not .Values.global.secretsBackend.vault.enabled)) }}
```
ishustava, line 142: could now only check for `.Values.global.enterpriseLicense.secretName` instead of both because we'll now fail if both of them are not provided?

ishustava (commented):
A couple more comments about tests, but otherwise looks good!

ishustava (approved):
:shipit:

PR #1038, opened 2022-02-18
Allow configuring primary DC and gateways via designated Helm values

ishustava:
Changes proposed in this PR:
- Add `global.federation.primaryDatacenter` and `global.federation.primaryGateway` to allow users to specify these values directly rather than via `server.extraConfig` or `server.extraVolumes`

Note that we will keep the existing behavior of `create-federation-secret` that generates the serverExtraConfig with those values for the time being and will re-assess it when the ACL work to switch out tokens for auth methods will be complete as it will likely affect it. So at the moment we will allow using both methods, and if the user provides these values via any of the `server.extra*` values, they will take precedence.

How I've tested this PR:
vault acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/framework/config/config.go
+++ b/acceptance/framework/config/config.go
@@ -63,11 +63,11 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
 			return nil, err
 		}
 		setIfNotEmpty(helmValues, "global.image", entImage)
-	}
 
-	if t.EnterpriseLicense != "" {
-		setIfNotEmpty(helmValues, "global.enterpriseLicense.secretName", LicenseSecretName)
-		setIfNotEmpty(helmValues, "global.enterpriseLicense.secretKey", LicenseSecretKey)
+		if t.EnterpriseLicense != "" {
+			setIfNotEmpty(helmValues, "global.enterpriseLicense.secretName", LicenseSecretName)
+			setIfNotEmpty(helmValues, "global.enterpriseLicense.secretKey", LicenseSecretKey)
+		}
 	}
 
 	if t.EnableOpenshift {
--- a/acceptance/tests/vault/vault_wan_fed_test.go
+++ b/acceptance/tests/vault/vault_wan_fed_test.go
@@ -199,11 +199,12 @@ func TestVault_WANFederationViaGateways(t *testing.T) {
 
 	// Get the address of the mesh gateway.
 	primaryMeshGWAddress := meshGatewayAddress(t, cfg, primaryCtx, consulReleaseName)
-	serverExtraConfig := fmt.Sprintf(`"{\"primary_gateways\":[\"%s\"]\,\"primary_datacenter\":\"dc1\"}"`, primaryMeshGWAddress)
 	secondaryConsulHelmValues := map[string]string{
 		"global.datacenter": "dc2",
 
-		"global.federation.enabled": "true",
+		"global.federation.enabled":            "true",
+		"global.federation.primaryDatacenter":  "dc1",
+		"global.federation.primaryGateways[0]": primaryMeshGWAddress,
 
 		// TLS config.
 		"global.tls.enabled":           "true",
@@ -229,7 +230,6 @@ func TestVault_WANFederationViaGateways(t *testing.T) {
 		"server.extraVolumes[0].type": "secret",
 		"server.extraVolumes[0].name": vaultCASecretName,
 		"server.extraVolumes[0].load": "false",
-		"server.extraConfig":          serverExtraConfig,
 
 		// Vault config.
 		"global.secretsBackend.vault.enabled":                       "true",
--- a/charts/consul/test/unit/server-config-configmap.bats
+++ b/charts/consul/test/unit/server-config-configmap.bats
@@ -493,3 +493,41 @@ load _helpers
       yq '.data["connect-ca-config.json"] | contains("\"ca_file\": \"/consul/vault-ca/tls.crt\"")' | tee /dev/stderr)
   [ "${actual}" = "true" ]
 }
+
+@test "server/ConfigMap: doesn't add federation config by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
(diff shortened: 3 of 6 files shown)


acceptance/framework/config/config.go
```diff
@@ -63,11 +63,11 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
...
 		setIfNotEmpty(helmValues, "global.image", entImage)
-	}
 
-	if t.EnterpriseLicense != "" {
-		setIfNotEmpty(helmValues, "global.enterpriseLicense.secretName", LicenseSecretName)
-		setIfNotEmpty(helmValues, "global.enterpriseLicense.secretKey", LicenseSecretKey)
+		if t.EnterpriseLicense != "" {
+			setIfNotEmpty(helmValues, "global.enterpriseLicense.secretName", LicenseSecretName)
+			setIfNotEmpty(helmValues, "global.enterpriseLicense.secretKey", LicenseSecretKey)
+		}
```
ishustava, line 70: we only want to set it when enterprise is enabled. Otherwise, we will always we will always set enterpriseLicense helm values which will now break when running with Vault.

acceptance/tests/vault/vault_wan_fed_test.go
```diff
@@ -229,7 +230,6 @@ func TestVault_WANFederationViaGateways(t *testing.T) {
 		"server.extraVolumes[0].type": "secret",
 		"server.extraVolumes[0].name": vaultCASecretName,
 		"server.extraVolumes[0].load": "false",
-		"server.extraConfig":          serverExtraConfig,
```
t-eckert, line 232: I was wondering if it would be worthwhile to keep a test where this is set via extra config, but I don't really think it is because at that point it would just be testing something that should be covered by the `bats` tests. So I think it's the right move to just modify the test.

charts/consul/values.yaml
```diff
@@ -430,6 +430,13 @@ global:
     # `<helm-release-name>-consul-federation`.
     createFederationSecret: false
 
+    # The name of the primary datacenter.
+    primaryDatacenter: ""
```
t-eckert, line 434: Depending on how much context a user would have by the time they are setting this value, it may be valuable to note what the significance of the primary datacenter is to the behavior of federation. In the past, users have been capable of figuring this out and adding it via extra config, so it may not be a problem.

t-eckert: It also may be worthwhile to add further context to the parent of this item linking the user to the federation documentation: https://www.consul.io/docs/k8s/installation/multi-cluster

charts/consul/values.yaml
```diff
@@ -430,6 +430,13 @@ global:
...
+
+    # A list of addresses of the primary mesh gateways in the form <ip>:<port>.
```
t-eckert, line 436: Might be extra helpful to see an example.
```suggestion
    # A list of addresses of the primary mesh gateways in the form <ip>:<port>.
    # (e.g. ["1.1.1.1:443", "2.3.4.5:443"] 
```

I think this could be helpful because I'm not sure what the YAML parsing behavior would be if the user entered `[1.1.1.1:443]`. I'm not sure if it might think it's parsing a float, get confused, and throw a hard to parse error.

charts/consul/test/unit/server-config-configmap.bats
```diff
@@ -492,4 +492,42 @@ load _helpers
...
+      --set 'global.federation.primaryDatacenter=dc1' \
+      --set 'global.federation.primaryGateways[0]=1.1.1.1:443' \
+      --set 'global.federation.primaryGateways[1]=2.2.2.2:443' \
+      --set 'global.tls.enabled=true' \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.data["federation-config.json"]' | tee /dev/stderr)
+  [ "${actual}" = '"{\n  \"primary_datacenter\": \"dc1\",\n  \"primary_gateways\": [\"1.1.1.1:443\",\"2.2.2.2:443\"]\n}"' ]
 }
```
t-eckert, line 533: What is the behavior if the user sets this configuration using `serverExtraConfig` _and_ sets it using these flags? Is it worth testing that scenario?

t-eckert (approved):
This looks fantastic, Iryna. I had some hot cocoa and free time so I went ahead and gave this a look tonight.

My only concern is making sure we understand the behavior if the user sets `primary_datacenter` and `primary_gateways` using both `serverExtraConfig` _and_ the new keys.

I made the relevant comment in the bats test, but it really makes more sense in the integration test. All the same, I think it's an edge case.

charts/consul/test/unit/server-config-configmap.bats
ishustava: The server extraConfig will override these settings (this is also the case in the general with extraConfig: it'll override any settings we set in the helm chart). Good question about testing - I don't think so and it'll also be hard to test with helm templates because these are provided via separate flags to the `consul agent` command. And so we'll be testing that one flag is provided after the other, and I don't know if there's a lot of value in that.

acceptance/tests/vault/vault_wan_fed_test.go
ishustava: We do use server extra config in the mesh_gateway_test which doesn't use vault:

https://github.com/hashicorp/consul-k8s/blob/ae206c9009a1860391b3cb55dfaeea394654234c/acceptance/tests/mesh-gateway/mesh_gateway_test.go#L76-L80

charts/consul/values.yaml
ishustava: I think most users are familiar with the concept of datacenter in Consul and also that is the value they provide via `global.datacenter` Helm value in the primary. I don't know if it makes sense to reference this value here because their primary datacenter could be on VMs. wdyt?

curtbushko (approved):

PR #1044, opened 2022-02-22
Expose optional metrics for API Gateway controller

nathancoleman:
Changes proposed in this PR:
- Expose optional metrics port for Consul API Gateway controller

How I've tested this PR:
- Installed this chart to a local kind cluster with and without the following values set:
  - `global.metrics.enabled=true`
  - `global.metrics.gateway.enableGatewayMetrics=true`
- Verified that the annotations and command on the deployment are update appropriately
- Verified that the controller logs include the following when enabled:

     ```
     2022-02-22T20:21:58.439Z [INFO]  manager/internal.go:383: consul-api-gateway-server.controller-runtime: starting metrics server: path=/metrics
     ```

See above

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -36,6 +36,11 @@ spec:
         "vault.hashicorp.com/ca-cert": "/vault/custom/{{ .Values.global.secretsBackend.vault.ca.secretKey }}"
         {{- end }}
         {{- end }}
+        {{- if (and .Values.global.metrics.enabled .Values.global.metrics.enableGatewayMetrics) }}
+        "prometheus.io/scrape": "true"
+        "prometheus.io/path": "/metrics"
+        "prometheus.io/port": "20200"
+        {{- end }}
       labels:
         app: {{ template "consul.name" . }}
         chart: {{ template "consul.chart" . }}
@@ -80,6 +85,9 @@ spec:
               -sds-server-host {{ template "consul.fullname" . }}-api-gateway-controller.{{ .Release.Namespace }}.svc \
               -k8s-namespace {{ .Release.Namespace }} \
               -log-level {{ default .Values.global.logLevel .Values.apiGateway.logLevel }} \
+              {{- if (and .Values.global.metrics.enabled .Values.global.metrics.enableGatewayMetrics) }}
+              -metrics-port 20200 \
+              {{ end }}
         volumeMounts:
           {{- if .Values.global.tls.enabled }}
           {{- if .Values.global.tls.enableAutoEncrypt }}


charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -36,6 +36,11 @@ spec:
         "vault.hashicorp.com/ca-cert": "/vault/custom/{{ .Values.global.secretsBackend.vault.ca.secretKey }}"
         {{- end }}
         {{- end }}
+        {{- if (and .Values.global.metrics.enabled .Values.global.metrics.enableGatewayMetrics) }}
```
nathancoleman, line 39: @Jeff-Apple the flags that we're keying off of can be found here: https://github.com/hashicorp/consul-k8s/blob/842452602717097c930e6fe6978f2c0416e31183/charts/consul/values.yaml#L441-L465

Question: Do we want to re-use `enableGatewayMetrics` for the purpose of this PR? If so, does it make sense to just add "API gateways" to the list of things that will expose metrics endpoints?

david-yu: I think that should be ok, since API GW is a superset of the Ingress GW uses the same metrics port.

nathancoleman: Committed minor docstring adjustment

andrewstucki: @david-yu So my question would be - do we have a pattern where our current `consul-k8s` controllers are exposing metrics via one of these flags? This particular change would only be exposing **_controller_** metrics, not individual gateway metrics, as we don't currently have some of the groundwork done for those yet (i.e. there's some work around proxying and metrics collection at the gateway deployment level that would need to be done to allow for that)

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -80,6 +85,9 @@ spec:
               -sds-server-host {{ template "consul.fullname" . }}-api-gateway-controller.{{ .Release.Namespace }}.svc \
               -k8s-namespace {{ .Release.Namespace }} \
               -log-level {{ default .Values.global.logLevel .Values.apiGateway.logLevel }} \
+              {{- if (and .Values.global.metrics.enabled .Values.global.metrics.enableGatewayMetrics) }}
```
andrewstucki, line 88: Not sure if `global.metrics.enableGatewayMetrics` makes sense here since we're dealing with **_controller_** metrics and not _**gateway**_ metrics per se.

nathancoleman: This is the main thing I'm looking for feedback on here. My understanding shifted after initially talking to @Jeff-Apple , but it sounded like we still wanted this in ahead of GA

andrewstucki: Yeah, so I'm ok exposing the metrics if we so desire, but my thought is that we could just key off of `global.metrics.enabled` rather than both it and `global.metrics.enableGatewayMetrics` since we're talking about controller-level metrics. I would think that as far as deployed api gateways are concerned we'd key off of both, but for the controller, at least IMO (which isn't super informed here 😅 ) the one makes sense, but not necessarily the other.

I can totally envision someone wanting to scrape controller metrics but not all deployed gateways, and tying the two options together this way would preclude that possibility without adding yet another flag/values option down the line.

charts/consul/templates/api-gateway-controller-deployment.yaml
ishustava: @andrewstucki at the moment we don't expose controller metrics. This flag is for envoy metrics from those gateways. So in your case I think this flag would control the metrics emitted by the ingress gateways proxies that get created via CRDs.

I think if this is only for the controller metrics, then re-using this flag could be confusing. We don't have a flag for this at the moment, but maybe this could be something like  `controllerMetrics` which will then enable it for all controllers. We (the k8s team) will then need to also add this to our controllers.

nathancoleman: @ishustava great idea! I've added `enableControllerMetrics`. Can you please check the docstring for it and make sure that contract makes sense going foward?

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -80,6 +85,9 @@ spec:
...
+              -metrics-port 20200 \
```
ishustava, line 89: I think controller uses port `8080` by default. Would it make sense to use the default that way we don't need additional flag for this.

charts/consul/templates/api-gateway-controller-deployment.yaml
ishustava: Looks good! thanks for updating so quickly. I've added one question about the port. I think we also need bats tests for this (sorry if this is something you're already aware of!)

charts/consul/templates/api-gateway-controller-deployment.yaml
nathancoleman: Are you referring to the API Gateway controller or a different one? API Gateway defaults to a port of `0`/disabled if this flag isn't present.

nathancoleman:
@ishustava and I discussed on Zoom a bit and agreed that there are too many outstanding questions here to attempt merge before release tomorrow. Primary points:
- The controller runtime exposes its own Prometheus metrics on port `8080`
- We're exposing some different Prometheus metrics on port `x` when the `-metrics-port` flag is provided
- There's no way to scrape Prometheus metrics from 2 different ports, so if we wanted both then there's some complicated metrics merging that has to happen. It sounds like the consul-k8s controller spins up a separate server inside the container to handle this merging. If we need to do that, then hopefully we could re-use something instead of writing another implementation.
- As an additional note: Metrics from the gateway deployments themselves are likely much higher priority than metrics from the controller.

charts/consul/templates/api-gateway-controller-deployment.yaml
andrewstucki: Keying off of the comment below about controller-runtime metrics, we'll likely, in the long run, want to see how simple it is to do something like register a pre-existing prometheus registry within a `github.com/armon/go-metrics` prometheus sink so we can use our standard library for metrics without requiring doing some scraping/aggregation.

PR #1052, opened 2022-02-23
Fix an issue with the issue-context-bot trigger.

t-eckert:
This fixes an issue that was causing the issue-context-bot to not trigger.


david-yu (approved):

kschoche (approved):
🚒

PR #1059, opened 2022-02-25
Use kube auth method to provision ACL client token

jmurret:
*NOTE: This draft PR is for a merge into a dev branch -`controller-acl-token-refactor`.  The intention is just to get feedback of concerns before I write more tests.  When `controller-acl-token-refactor` is merged into `acls-refactor-base-branch`, I will close this Draft PR and open  a proper one against `acls-refactor-base-branch`.*

Changes proposed in this PR:
- `control-plane/subcommand/acl-init/command.go` - when using ACLs, retrieve the token secret from consul login rather than k8s secrets
-  `control-plane/subcommand/common/common.go` 
    - modify `ConsulLogin()` to also return the token (consul core code does this as well), in addition to storing it in the `tokenSinkFile`, so that it can be used in ACL init to write out the 
    - modify `ConsulLogin()`to write the `tokenSinkFile` with 0644 permissions that was original used before ACLs refactor, in part to support [this bug fix] (https://github.com/hashicorp/consul-k8s/pull/248/files#diff-29d4f970a3786e8c721fe5472dc71f9f94482513a3e6042f03c7481e6cec2195L108).
`acl-config.json` file that is also used in the k8s secrets workflow
- `control-plane/subcommand/connect-init/command.go` - minor changes to support return type change to`ConsulLogin`
- `control-plane/subcommand/server-acl-init/create_or_update.go` - make change to create BindingRule if it does not exist rather than error when there are other existing BindingRules associated with the auth method

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/control-plane/subcommand/common/common.go
+++ b/control-plane/subcommand/common/common.go
@@ -31,6 +31,8 @@ const (
 	// which secrets to delete on an uninstall.
 	CLILabelKey   = "managed-by"
 	CLILabelValue = "consul-k8s"
+
+	ComponentAuthMethod = "consul-k8s-component-auth-method"
 )
 
 // Logger returns an hclog instance with log level set and JSON logging enabled/disabled, or an error if level is invalid.
@@ -80,9 +82,9 @@ func ValidateUnprivilegedPort(flagName, flagValue string) error {
 // ConsulLogin issues an ACL().Login to Consul and writes out the token to tokenSinkFile.
 // The logic of this is taken from the `consul login` command.
 func ConsulLogin(client *api.Client, bearerTokenFile, authMethodName, tokenSinkFile, namespace string, meta map[string]string) error {
-	if meta == nil {
-		return fmt.Errorf("invalid meta")
-	}
+	//if meta == nil {
+	//	return fmt.Errorf("invalid meta")
+	//}
 	data, err := ioutil.ReadFile(bearerTokenFile)
 	if err != nil {
 		return fmt.Errorf("unable to read bearerTokenFile: %v, err: %v", bearerTokenFile, err)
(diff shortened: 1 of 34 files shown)


control-plane/subcommand/common/common.go
```diff
@@ -88,29 +88,32 @@ func ValidateUnprivilegedPort(flagName, flagValue string) error {
...
 			BearerToken: bearerToken,
 			Meta:        meta,
 		}
-		tok, _, err := client.ACL().Login(req, &api.WriteOptions{Namespace: namespace})
+		tok, _, err = client.ACL().Login(req, &api.WriteOptions{Namespace: namespace})
 		if err != nil {
 			return fmt.Errorf("error logging in: %s", err)
 		}
 		// Write out the resultant token file.
-		if err := WriteFileWithPerms(tokenSinkFile, tok.SecretID, 0444); err != nil {
```
ishustava, line 113: I wonder if we should also make the writing to the file optional (maybe based on whether `tokenSinkFile` is provided or not). I'm thinking that when we're using it for consul client, we don't need the token to be written both to a file and to config json.

jmurret: I think you are also saying "Are we writing to a Sink file even though when this optional flag is not set?"???

jmurret: An additional question for context....Will consul agent merge the token in the sink file with the tokens specified in this type of config that is written on the client?
```
{
  "acl": {
    "enabled": true,
    "default_policy": "deny",
    "down_policy": "extend-cache",
    "tokens": {
      "agent": "{{ . }}"
    }
  }
```

Namely, we always need to write this part if `init-type="client"` because it not only sets the token but also enables ACls, sets deny as default, etc. ?

ishustava: >Are we writing to a Sink file even though when this optional flag is not set?

yep exactly. The consul client agent will only look at the token in the config and so writing it to a file as well is not necessary.

>Namely, we always need to write this part if init-type="client" because it not only sets the token but also enables ACls, sets deny as default, etc. ?

Yes.

jmurret:
Closing this and will re-open a proper rebasedPR

PR #1061, opened 2022-02-25
Support storing bootstrap token in Vault

ishustava:
Changes proposed in this PR:
Support providing a bootstrap token secret that is stored in Vault. Previously in our code, we assumed that the bootstrap token secret would only need to be provided when running with external servers (i.e. consul servers running outside of *this* Kubernetes cluster). This is no longer the case. 

To accomodate for the case when servers are running externally, there's now a new flag `-set-server-tokens` for the server-acl-init command. When servers are running in this cluster, we will set this flag to true so that we set server tokens. When servers are running externally, we don't want to set server tokens because we don't manage those servers.

To configure the bootstrap token for the server, we provide it via config (`initial_manangement` token: https://www.consul.io/docs/agent/options#acl_tokens_initial_management)

How I've tested this PR:
acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,8 +70,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        #default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:latest"
-        default: "kyleschochenmaier/consul-k8s-acls"
+        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:$(git rev-parse --short HEAD)"
       go-path:
         type: string
         default: "/home/circleci/.go_workspace"
--- a/acceptance/tests/vault/helpers.go
+++ b/acceptance/tests/vault/helpers.go
@@ -18,6 +18,13 @@ const (
 path "consul/data/secret/gossip" {
   capabilities = ["read"]
 }`
+
+	bootstrapTokenPolicy = `
+path "consul/data/secret/bootstrap" {
+  capabilities = ["read"]
+}`
+
+	// todo: should this policy be only "read"?
 	replicationTokenPolicy = `
 path "consul/data/secret/replication" {
   capabilities = ["read", "update"]
@@ -145,8 +152,9 @@ func configureKubernetesAuthRoles(t *testing.T, vaultClient *vapi.Client, consul
 	params = map[string]interface{}{
 		"bound_service_account_names":      consulServerServiceAccountName,
 		"bound_service_account_namespaces": ns,
-		"policies":                         fmt.Sprintf(sharedPolicies+",connect-ca-%s,consul-server-%s,consul-replication-token", datacenter, datacenter),
-		"ttl":                              "24h",
+		// todo: should we only allow the primary server have access to the bootstrap token? and only the secondary to have access to the replication token
+		"policies": fmt.Sprintf(sharedPolicies+",connect-ca-%s,consul-server-%s,consul-replication-token,consul-bootstrap-token", datacenter, datacenter),
+		"ttl":      "24h",
 	}
 	_, err = vaultClient.Logical().Write(fmt.Sprintf("auth/%s/role/consul-server", authPath), params)
 	require.NoError(t, err)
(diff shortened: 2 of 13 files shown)


acceptance/tests/vault/helpers.go
```diff
@@ -18,9 +18,15 @@ const (
...
+
+	bootstrapTokenPolicy = `
+path "consul/data/secret/bootstrap" {
+  capabilities = ["read"]
+}`
+
 	replicationTokenPolicy = `
 path "consul/data/secret/replication" {
-  capabilities = ["read", "update"]
+  capabilities = ["read"]
```
ishustava, line 29: this policy didn't need to have `update` perms

acceptance/tests/vault/helpers.go
```diff
@@ -160,6 +178,22 @@ func configureKubernetesAuthRoles(t *testing.T, vaultClient *vapi.Client, consul
 	}
 	_, err = vaultClient.Logical().Write(fmt.Sprintf("auth/%s/role/consul-ca", authPath), params)
 	require.NoError(t, err)
+
+	logger.Log(t, "Creating kubernetes auth role for the server-acl-init job")
```
ishustava, line 182: we now need to setup the role for server-acl-init here (instead of `configureReplicationTokenVaultSecret`) because it's needs both replication and bootstrap tokens policies

charts/consul/templates/_helpers.tpl
```diff
@@ -73,6 +73,13 @@ as well as the global.name setting.
           {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
 
+{{- define "consul.vaultBootstrapTokenConfigTemplate" -}}
+|
+          {{ "{{" }}- with secret "{{ .Values.global.acls.bootstrapToken.secretName }}" -{{ "}}" }}
+          acl { tokens { initial_management = "{{ "{{" }}- {{ printf ".Data.data.%s" .Values.global.acls.bootstrapToken.secretKey }} -{{ "}}" }}" }}
+          {{ "{{" }}- end -{{ "}}" }}
+{{- end -}}
```
ishustava, line 81: The consul servers are configured to have the `initial_management` token that is stored in vault. This is an alternative to calling the `/v1/acl/bootstrap` api to get the bootstrap token. I think in our case this is safe because vault will store this config in tmpfs.

charts/consul/templates/server-statefulset.yaml
```diff
@@ -286,14 +290,6 @@ spec:
...
                 -client=0.0.0.0 \
                 -config-dir=/consul/config \
-                {{- /* Always include the extraVolumes at the end so that users can
-                    override other Consul settings. The last -config-dir takes
-                    precedence. */}}
-                {{- range .Values.server.extraVolumes }}
-                {{- if .load }}
-                -config-dir=/consul/userconfig/{{ .name }} \
-                {{- end }}
-                {{- end }}
```
ishustava, line 296: moving this to be at the end to behave like the comment says

charts/consul/test/unit/server-acl-init-job.bats
```diff
@@ -1483,64 +1610,36 @@ load _helpers
 #--------------------------------------------------------------------
 # global.acls.bootstrapToken
 
-@test "serverACLInit/Job: -bootstrap-token-file is not set by default" {
+@test "serverACLInit/Job: bootstrapToken.secretKey is required when bootstrapToken.secretName is set" {
```
ishustava, line 1613: these tests have changed because we now fail if only one of secretKey or secretName has been provided

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -1659,196 +1659,245 @@ func TestRun_ClientTokensRetry(t *testing.T) {
...
-	k8s := fake.NewSimpleClientset()
-	setUpK8sServiceAccount(t, k8s, ns)
-
-	type APICall struct {
-		Method string
-		Path   string
+	cases := map[string]bool{
+		"token saved in k8s secret": true,
+		"token provided via file":   false,
 	}
```
ishustava, line 1665: this test is mostly the same, except it now needs to test two cases instead of one (when token is stored as a secret and when it's provided via file)

.circleci/config.yml
```diff
@@ -70,8 +70,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        #default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:latest"
-        default: "kyleschochenmaier/consul-k8s-acls"
+        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:$(git rev-parse --short HEAD)"
```
ishustava, line 73: This will allow us to use whatever the image is for this commit instead of `latest`. This can go into the `main` branch as it will allow us to not have to set the image for features we're working on.

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -319,12 +323,7 @@ func (c *Command) Run(args []string) int {
 
 	var bootstrapToken string
 
-	if c.flagBootstrapTokenFile != "" {
-		// If bootstrap token is provided, we skip server bootstrapping and use
-		// the provided token to create policies and tokens for the rest of the components.
-		c.log.Info("Bootstrap token is provided so skipping Consul server ACL bootstrapping")
-		bootstrapToken = providedBootstrapToken
-	} else if c.flagACLReplicationTokenFile != "" && !c.flagCreateACLReplicationToken {
```
ishustava, line 327: previously, we'd skip server bootstrapping (including setting server agent tokens) because we assumed you're running with external servers. We no longer want to do when the bootstrap token is stored in vault.

.circleci/config.yml
kschoche: hooray, thanks for doing this, I was trying to figure out how it was possible.
🥇

thisisnotashwin (approved):
This is an excellent PR! thanks for the detailed comments. made reviewing it super easy.

kschoche (approved):
most excellent work!

PR #1065, opened 2022-02-28
Increase wait for all pods to be ready from 10 to 11 minutes

t-eckert:
Changes proposed in this PR:
- Increase the timeout for pods to be in a ready state from 10 minutes to 11 minutes in acceptance tests

We've increased this wait time before, but I think it makes sense to increase it again after seeing this
test failure: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/4505/workflows/fe2c423d-47b5-493e-b3b2-9707655f77f6/jobs/43239/tests 

How I've tested this PR:

I haven't I'm hoping this reduces some flakiness I've observed in the nightly tests.

We'll see if this helps with the flakiness. 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/framework/consul/consul_cluster.go
+++ b/acceptance/framework/consul/consul_cluster.go
@@ -85,11 +85,11 @@ func NewHelmCluster(
 
 	logger := terratestLogger.New(logger.TestLogger{})
 
-	// Wait up to 20 min for K8s resources to be in a ready state. Increasing
+	// Wait up to 15 min for K8s resources to be in a ready state. Increasing
 	// this from the default of 5 min could help with flakiness in environments
 	// like AKS where volumes take a long time to mount.
 	extraArgs := map[string][]string{
-		"install": {"--timeout", "20m"},
+		"install": {"--timeout", "15m"},
 	}
 
 	opts := &helm.Options{
--- a/acceptance/framework/consul/consul_cluster.go
+++ b/acceptance/framework/consul/consul_cluster.go
@@ -85,11 +85,11 @@ func NewHelmCluster(
 
 	logger := terratestLogger.New(logger.TestLogger{})
 
-	// Wait up to 15 min for K8s resources to be in a ready state. Increasing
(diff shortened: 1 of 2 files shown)


acceptance/framework/consul/consul_cluster.go
```diff
@@ -85,11 +85,11 @@ func NewHelmCluster(
 
 	logger := terratestLogger.New(logger.TestLogger{})
 
-	// Wait up to 15 min for K8s resources to be in a ready state. Increasing
+	// Wait up to 20 min for K8s resources to be in a ready state. Increasing
 	// this from the default of 5 min could help with flakiness in environments
 	// like AKS where volumes take a long time to mount.
 	extraArgs := map[string][]string{
-		"install": {"--timeout", "15m"},
+		"install": {"--timeout", "20m"},
```
ishustava, line 92: I'm not sure if this will help for the test you linked since for the non-secure case helm exists immediately and we use the `WaitForAllPodsToBeReady` function to wait

ndhanushkodi: @t-eckert here's the timeout for WaitForAllPodsToBeReady: https://github.com/hashicorp/consul-k8s/blob/3ad09f72f9df397e163abcc24b36feffbb35d454/acceptance/framework/k8s/helpers.go#L46

The ingress gateway pod's logs do show it was registered with consul pretty close to when the test cleaned everything up, which could mean that maybe increasing that timeout a little could help

t-eckert: That's what I was thinking. It was very close when the test timed out.

t-eckert: So I set this back and I changed the retry counter to timeout at 11 minutes instead of 10. I also fixed the comment to reflect the change.

ndhanushkodi (approved):
Should change commit message and go fmt before merging but I think this could be worth trying!

ishustava (approved):

PR #1070, opened 2022-03-02
Use kube auth method to provision ACL client token

jmurret:
*Prior Art: https://github.com/hashicorp/consul-k8s/pull/1059*

Changes proposed in this PR:
- `control-plane/subcommand/acl-init/command.go` - when using ACLs, retrieve the token secret from consul login rather than k8s secrets
-  `control-plane/subcommand/common/common.go` 
    - modify `ConsulLogin()` to also return the token (consul core code does this as well), in addition to storing it in the `tokenSinkFile`, so that it can be used in ACL init to write out the 
    - modify `ConsulLogin()`to write the `tokenSinkFile` with 0644 permissions that was original used before ACLs refactor, in part to support [this bug fix] (https://github.com/hashicorp/consul-k8s/pull/248/files#diff-29d4f970a3786e8c721fe5472dc71f9f94482513a3e6042f03c7481e6cec2195L108).
`acl-config.json` file that is also used in the k8s secrets workflow
- `control-plane/subcommand/connect-init/command.go` - minor changes to support return type change to`ConsulLogin`
- `control-plane/subcommand/server-acl-init/create_or_update.go` - make change to create BindingRule if it does not exist rather than error when there are other existing BindingRules associated with the auth method

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/control-plane/subcommand/partition-init/command.go
+++ b/control-plane/subcommand/partition-init/command.go
@@ -28,11 +28,10 @@ type Command struct {
 	flagPartitionName string
 
 	// Flags to configure Consul connection
-	flagServerAddresses     []string
-	flagServerPort          uint
-	flagConsulCACert        string
-	flagConsulTLSServerName string
-	flagUseHTTPS            bool
+	flagServerAddresses []string
+	flagServerPort      uint
+	flagConsulCACert    string
+	flagUseHTTPS        bool
 
 	flagLogLevel string
 	flagLogJSON  bool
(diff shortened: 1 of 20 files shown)


control-plane/subcommand/partition-init/command.go
```diff
@@ -134,14 +131,11 @@ func (c *Command) Run(args []string) int {
 	}
 	// For all of the next operations we'll need a Consul client.
 	serverAddr := fmt.Sprintf("%s:%d", serverAddresses[0], c.flagServerPort)
-	consulClient, err := consul.NewClient(&api.Config{
-		Address: serverAddr,
-		Scheme:  scheme,
-		TLSConfig: api.TLSConfig{
```
jmurret, line 140: TLS COnfig will get set anyway by `c.http.MergeOntoConfig()` call.

PR #1072, opened 2022-03-03
rename create-controller-token flag to enable-controller

kschoche:
Since `server-acl-init` no longer creates a controller token rename the flag to `-enable-controller`.

Changes proposed in this PR:
- renames `-create-controller-token` flag to `-enable-controller` for `server-acl-init`.

How I've tested this PR:
unit tests still pass

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/server-acl-init-job.yaml
+++ b/charts/consul/templates/server-acl-init-job.yaml
@@ -251,7 +251,7 @@ spec:
                 {{- end }}
 
                 {{- if .Values.controller.enabled }}
-                -create-controller-token=true \
+                -enable-controller=true \
                 {{- end }}
 
                 {{- if .Values.apiGateway.enabled }}
--- a/charts/consul/templates/server-acl-init-job.yaml
+++ b/charts/consul/templates/server-acl-init-job.yaml
@@ -251,7 +251,7 @@ spec:
                 {{- end }}
 
                 {{- if .Values.controller.enabled }}
-                -enable-controller=true \
+                -controller=true \
                 {{- end }}
 
                 {{- if .Values.apiGateway.enabled }}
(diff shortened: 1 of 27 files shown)


control-plane/subcommand/server-acl-init/command.go
```diff
@@ -147,7 +147,7 @@ func (c *Command) init() {
 	c.flags.StringVar(&c.flagBindingRuleSelector, "acl-binding-rule-selector", "",
 		"Selector string for connectInject ACL Binding Rule.")
 
-	c.flags.BoolVar(&c.flagCreateControllerPoliciesAndBindings, "create-controller-token", false,
+	c.flags.BoolVar(&c.flagCreateControllerPoliciesAndBindings, "enable-controller", false,
```
ishustava, line 150: Need to rename flag var name

ishustava (approved):

charts/consul/templates/server-acl-init-job.yaml
```diff
@@ -251,7 +251,7 @@ spec:
                 {{- end }}
 
                 {{- if .Values.controller.enabled }}
-                -create-controller-token=true \
+                -enable-controller=true \
```
thisisnotashwin, line 254: how about `-controller=true`? It seems similar to patterns we have for partitions.

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -47,7 +47,7 @@ type Command struct {
 	flagInjectAuthMethodHost string
 	flagBindingRuleSelector  string
 
-	flagCreateControllerPoliciesAndBindings bool
+	flagEnableController bool
```
thisisnotashwin, line 50: This can still stay the same though. Just the name of the flag can be `-controller`

ishustava: I think we usually try to match the var name with flag name. IMO it's kind of nice to know what the flag name is by looking at the command var as you're reading through the code.

PR #1073, opened 2022-03-03
updating make control-plane-dev to work ondarwin/arm64.  Update helm version for bats tests

jmurret:
Changes proposed in this PR:
- Update docs to require `gox` as a pre-requisite install so you can run `make control-plane-dev` and `make-control-plane-dev-docker`
- Update local build scripts to not exclude darwin/arm64, which was set 4 years ago, when it was probably not supported by gox at that time, but works now.
- update `helm` pre-requisited to denote pinned version of 3.6.3, otherwise helm changes its output in bats test to be non-cpaitalized and they fail.

How I've tested this PR:
- `make control-plane-dev` and `make-control-plane-docker` failed for me on my darwin/arm64 laptop.
- implemented changes and they worked
- 
👀

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -582,7 +582,7 @@ The acceptance tests require a Kubernetes cluster with a configured `kubectl`.
   ```bash
   brew install python-yq
   ```
-* [Helm 3](https://helm.sh) (Helm 2 is not supported)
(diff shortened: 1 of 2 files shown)


kschoche (approved):

curtbushko (approved):

jmurret:
re-opening as this is still valid for local development.

PR #1074, opened 2022-03-03
Set DOCKER_DEFAULT_PLATFORM for make control-plane-dev-docker so that images built on M1/arm64machines do not have failures in k8s.

jmurret:
Changes proposed in this PR:
-  Update Makefile for `control-plane-dev-docker` so that it sets DOCKER_DEFAULT_PLATFORM=linux/amd64.  This sets the proper architecture in the docker manifest. Without this if you build adocker image on an arm64 machine, it will set the docker manifest to arm64.  This then causes errors in kubernetes running the image.

How I've tested this PR:
- completely blocked from publishing images that i could use in acceptance tests running locally
- implemented this
- unblocked

👀


t-eckert (approved):
Nice! This is something I haven't run into running on Intel (of course) so I'm glad you caught it.

PR #1075, opened 2022-03-04
Global auth method

thisisnotashwin:
Changes proposed in this PR:
- Update server-acl-init to create authmethods in the primary datacenter when the job is run in a secondary datacenter during federation. This authmethod allows us to issue logins for global policies.
- Update the controller workflow in server-acl-init to use this global authmethod when run in a secondary DC.
- Update the mesh-gateway acceptance tests to create proxy defaults in the secondary DC to test above behavior works successfully.
- Updated logout to not pass in the partition flag as it is not required.
- Update server acl init tests to migrate from `require := require.New(t)` to `require.xyz(t, ...)` patterns.

How I've tested this PR:
- Unit tests
- Acceptance tests

- Code Review

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/mesh-gateway/mesh_gateway_test.go
+++ b/acceptance/tests/mesh-gateway/mesh_gateway_test.go
@@ -33,8 +33,9 @@ func TestMeshGatewayDefault(t *testing.T) {
 		"global.federation.enabled":                "true",
 		"global.federation.createFederationSecret": "true",
 
-		"connectInject.enabled": "true",
-		"controller.enabled":    "true",
+		"connectInject.enabled":  "true",
+		"connectInject.replicas": "1",
+		"controller.enabled":     "true",
 
 		"meshGateway.enabled":  "true",
 		"meshGateway.replicas": "1",
@@ -79,7 +80,9 @@ func TestMeshGatewayDefault(t *testing.T) {
 		"server.extraVolumes[0].items[0].key":  "serverConfigJSON",
 		"server.extraVolumes[0].items[0].path": "config.json",
 
-		"connectInject.enabled": "true",
+		"connectInject.enabled":  "true",
+		"connectInject.replicas": "1",
+		"controller.enabled":     "true",
 
 		"meshGateway.enabled":  "true",
 		"meshGateway.replicas": "1",
@@ -164,8 +167,9 @@ func TestMeshGatewaySecure(t *testing.T) {
 				"global.federation.enabled":                "true",
 				"global.federation.createFederationSecret": "true",
 
-				"connectInject.enabled": "true",
-				"controller.enabled":    "true",
+				"connectInject.enabled":  "true",
+				"connectInject.replicas": "1",
+				"controller.enabled":     "true",
 
 				"meshGateway.enabled":  "true",
 				"meshGateway.replicas": "1",
@@ -191,6 +195,19 @@ func TestMeshGatewaySecure(t *testing.T) {
 			_, err = secondaryContext.KubernetesClient(t).CoreV1().Secrets(secondaryContext.KubectlOptions(t).Namespace).Create(context.Background(), federationSecret, metav1.CreateOptions{})
 			require.NoError(t, err)
 
+			var k8sAuthMethodHost string
+			// When running on kind, the kube API address in kubeconfig will have a localhost address
+			// which will not work from inside the container. That's why we need to use the endpoints address instead
+			// which will point the node IP.
+			if cfg.UseKind {
+				// The Kubernetes AuthMethod host is read from the endpoints for the Kubernetes service.
+				kubernetesEndpoint, err := secondaryContext.KubernetesClient(t).CoreV1().Endpoints("default").Get(context.Background(), "kubernetes", metav1.GetOptions{})
+				require.NoError(t, err)
+				k8sAuthMethodHost = fmt.Sprintf("%s:%d", kubernetesEndpoint.Subsets[0].Addresses[0].IP, kubernetesEndpoint.Subsets[0].Ports[0].Port)
+			} else {
+				k8sAuthMethodHost = k8s.KubernetesAPIServerHostFromOptions(t, secondaryContext.KubectlOptions(t))
+			}
+
 			// Create secondary cluster
 			secondaryHelmValues := map[string]string{
 				"global.datacenter": "dc2",
@@ -207,7 +224,9 @@ func TestMeshGatewaySecure(t *testing.T) {
 				"global.acls.replicationToken.secretName": federationSecretName,
 				"global.acls.replicationToken.secretKey":  "replicationToken",
 
-				"global.federation.enabled": "true",
+				"global.federation.enabled":           "true",
+				"global.federation.k8sAuthMethodHost": k8sAuthMethodHost,
+				"global.federation.primaryDatacenter": "dc1",
 
 				"server.extraVolumes[0].type":          "secret",
 				"server.extraVolumes[0].name":          federationSecretName,
@@ -215,7 +234,9 @@ func TestMeshGatewaySecure(t *testing.T) {
 				"server.extraVolumes[0].items[0].key":  "serverConfigJSON",
 				"server.extraVolumes[0].items[0].path": "config.json",
 
-				"connectInject.enabled": "true",
+				"connectInject.enabled":  "true",
+				"connectInject.replicas": "1",
+				"controller.enabled":     "true",
 
 				"meshGateway.enabled":  "true",
 				"meshGateway.replicas": "1",
@@ -248,9 +269,9 @@ func TestMeshGatewaySecure(t *testing.T) {
 			// gateways.
 			logger.Log(t, "creating proxy-defaults config")
 			kustomizeDir := "../fixtures/bases/mesh-gateway"
-			k8s.KubectlApplyK(t, primaryContext.KubectlOptions(t), kustomizeDir)
+			k8s.KubectlApplyK(t, secondaryContext.KubectlOptions(t), kustomizeDir)
 			helpers.Cleanup(t, cfg.NoCleanupOnFailure, func() {
-				k8s.KubectlDeleteK(t, primaryContext.KubectlOptions(t), kustomizeDir)
+				k8s.KubectlDeleteK(t, secondaryContext.KubectlOptions(t), kustomizeDir)
 			})
 
 			// Check that we can connect services over the mesh gateways
--- a/charts/consul/templates/controller-clusterrole.yaml
+++ b/charts/consul/templates/controller-clusterrole.yaml
@@ -57,15 +57,6 @@ rules:
   - get
   - list
   - update
-{{- if .Values.global.acls.manageSystemACLs }}
-- apiGroups: [""]
-  resources:
-  - secrets
-  resourceNames:
-  - {{ template "consul.fullname" . }}-controller-acl-token
-  verbs:
-  - get
-{{- end }}
 {{- if .Values.global.enablePodSecurityPolicies }}
 - apiGroups: ["policy"]
   resources: ["podsecuritypolicies"]
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -85,7 +85,7 @@ global:
   # image: "hashicorp/consul-enterprise:1.10.0-ent"
   # ```
   # @default: hashicorp/consul:<latest version>
-  image: "hashicorp/consul:1.11.3"
+  image: "hashicorp/consul:1.11.4"
 
   # Array of objects containing image pull secret names that will be applied to each service account.
   # This can be used to reference image pull secrets if using a custom consul or consul-k8s-control-plane Docker image.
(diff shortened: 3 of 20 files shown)


control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2353,6 +2353,82 @@ func TestRun_PoliciesAndBindingRulesACLLoginInSecondaryDatacenterDuringFederatio
...
+			_, jwtToken := setUpK8sServiceAccount(t, k8s, ns)
+			k8sMockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("content-type", "application/json")
+				if r != nil && r.URL.Path == "/apis/authentication.k8s.io/v1/tokenreviews" && r.Method == "POST" {
+					w.Write([]byte(tokenReviewsResponse(c.ComponentName, ns)))
+				}
+				if r != nil && r.URL.Path == fmt.Sprintf("/api/v1/namespaces/%s/serviceaccounts/%s", ns, c.ComponentName) &&
+					r.Method == "GET" {
+					w.Write([]byte(serviceAccountGetResponse(c.ComponentName, ns)))
+				}
```
ishustava, line 2390: These need to be tied to the service account name instead of the `c.ComponentName`. We add a prefix to the service account for the binding rule (i.e. `fmt.Sprintf("%s-%s", resourcePrefix, c.ComponentName)` so probably consul returns 403 because there's no matching binding rule?

thisisnotashwin: GENIUS! I got the tests working! ill finish up the PR and have it ready for review on Monday!

Thank you so much @ishustava and @kschoche !! I couldn't have done this without your help.

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -125,7 +124,7 @@ func TestRun_Defaults(t *testing.T) {
 		"-resource-prefix=" + resourcePrefix,
 	}
 	responseCode := cmd.Run(args)
-	require.Equal(0, responseCode, ui.ErrorWriter.String())
+	require.Equal(t, 0, responseCode, ui.ErrorWriter.String())
```
thisisnotashwin, line 127: updated these tests to no longer use the `require := require.New(t)` pattern and use the `require.xyz(t, ...)` pattern

control-plane/subcommand/server-acl-init/create_or_update.go
```diff
@@ -11,12 +11,63 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// addRoleAndBindingRule adds an ACLRole and ACLBindingRule which reference the authMethod.
```
thisisnotashwin, line 14: I moved the order of the method around so that the methods that are invoked from another are defined below the caller and hence the diff looks "big". The only actual change is adding support for global tokens.

acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -112,9 +115,9 @@ func TestMeshGatewayDefault(t *testing.T) {
 	// gateways.
 	logger.Log(t, "creating proxy-defaults config")
 	kustomizeDir := "../fixtures/bases/mesh-gateway"
-	k8s.KubectlApplyK(t, primaryContext.KubectlOptions(t), kustomizeDir)
+	k8s.KubectlApplyK(t, secondaryContext.KubectlOptions(t), kustomizeDir)
 	helpers.Cleanup(t, cfg.NoCleanupOnFailure, func() {
-		k8s.KubectlDeleteK(t, primaryContext.KubectlOptions(t), kustomizeDir)
+		k8s.KubectlDeleteK(t, secondaryContext.KubectlOptions(t), kustomizeDir)
 	})
```
ishustava, line 121: should we keep this test using the primary for more coverage? I think the main one we care about is the secure test anyways

charts/consul/templates/controller-clusterrole.yaml
```diff
@@ -57,15 +57,6 @@ rules:
...
   - update
-{{- if .Values.global.acls.manageSystemACLs }}
-- apiGroups: [""]
-  resources:
-  - secrets
-  resourceNames:
-  - {{ template "consul.fullname" . }}-controller-acl-token
-  verbs:
-  - get
-{{- end }}
```
ishustava, line 68: great catch!

control-plane/subcommand/acl-init/command.go
```diff
@@ -73,6 +74,7 @@ func (c *Command) init() {
 
 	// Flags related to using consul login to fetch the ACL token.
 	c.flags.StringVar(&c.flagNamespace, "k8s-namespace", "", "Name of Kubernetes namespace where the token Kubernetes secret is stored.")
+	c.flags.StringVar(&c.flagPrimaryDatacenter, "primary-datacenter", "", "Name of the primary datacenter when federation is enabled and the command is run in a secondary partition.")
```
ishustava, line 77: partitions are haunting you 😄 
```suggestion
	c.flags.StringVar(&c.flagPrimaryDatacenter, "primary-datacenter", "", "Name of the primary datacenter when federation is enabled and the command is run in a secondary datacenter.")
```

charts/consul/values.yaml
```diff
@@ -431,13 +431,29 @@ global:
...
+    #
+    # You could retrieve this value from your `kubeconfig` by running:
+    #
+    # ```shell-session
+    # $ kubectl config view \
+    #   -o jsonpath="{.clusters[?(@.name=='<your cluster name>')].cluster.server}"
+    # ```
+    #
+    # @type: string
+    k8sAuthMethodHost: null
```
ishustava, line 455: do you think it'll make sense to clarify that this does not affect an auth method used for the service mesh?

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -1168,16 +1173,266 @@ func TestRun_PoliciesAndBindingRulesForACLLoginNamespacesEnabled(t *testing.T) {
...
+				for x := range role.Policies {
+					if role.Policies[x].Name == policy.Name {
+						found = true
+						break
+					}
+				}
+				require.True(t, found)
+
+				// Check that there exists a BindingRule that references this Role.
+				rb, _, err := consul.ACL().BindingRuleList(fmt.Sprintf("release-name-%s-%s", componentAuthMethod, secondaryDatacenter), &api.QueryOptions{Datacenter: primaryDatacenter})
```
ishustava, line 1257: could we use `resourcePrefix` variable that we're passing to the command?

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -1168,16 +1173,266 @@ func TestRun_PoliciesAndBindingRulesForACLLoginNamespacesEnabled(t *testing.T) {
...
+						break
+					}
+				}
+				require.True(t, found)
+			}
+		})
+	}
+}
+
+// Test that server-acl-init used the global auth method to create the desired token in the primary datacenter.
```
ishustava, line 1273:
```suggestion
// Test that server-acl-init used the local auth method to create the desired token in the primary datacenter.
```

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -1168,16 +1173,266 @@ func TestRun_PoliciesAndBindingRulesForACLLoginNamespacesEnabled(t *testing.T) {
...
+// The test works by running the login command and then ensuring that the token
+// returned has the correct role for the component.
+func TestRun_NamespaceEnabled_ValidateLoginToken_PrimaryDatacenter(t *testing.T) {
```
ishustava, line 1276: why do we need these tests for ent? is there any ns logic that we're testing here?

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2238,6 +2221,425 @@ func TestRun_GatewayErrors(t *testing.T) {
...
+				for x := range role.Policies {
+					if role.Policies[x].Name == policy.Name {
+						found = true
+						break
+					}
+				}
+				require.True(t, found)
+
+				// Check that there exists a BindingRule that references this Role.
+				rb, _, err := consul.ACL().BindingRuleList("release-name-"+componentAuthMethod, &api.QueryOptions{})
```
ishustava, line 2298: could we use `resourcePrefix` var here?

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2238,6 +2221,425 @@ func TestRun_GatewayErrors(t *testing.T) {
...
+				AuthMethod:  authMethodName,
+				BearerToken: jwtToken,
+				Meta:        map[string]string{},
+			}, &api.WriteOptions{})
+			require.NoError(t, err)
+
+			require.Equal(t, len(tok.Roles), len(c.Roles))
+			for _, role := range tok.Roles {
+				require.Contains(t, c.Roles, role.Name)
+			}
```
ishustava, line 2479: could we add an assertion for token locality here and in the below test?

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2632,5 +2906,50 @@ func policyExists(t require.TestingT, name string, client *api.Client) *api.ACLP
 	return policy
 }
 
+func tokenReviewsResponse(name, ns string) string {
```
ishustava, line 2909: could we re-use the ones from test_util?

ishustava (approved):
Looks fantastic!! A few minor comments, but nothing major.

control-plane/subcommand/server-acl-init/command_ent_test.go
thisisnotashwin: I set these up for future tests that have components that need global/local tokens based on whether namespaces are enabled.

control-plane/subcommand/acl-init/command.go
thisisnotashwin: lolol

control-plane/subcommand/server-acl-init/command_ent_test.go
thisisnotashwin: It is tricky for this test because the `resourcePrefix` is `release-name-consul` and the authmethod name is `consul-k8s-component-authmethod` and the `consul` word will get repeated. Ill see if there is something else that could potentially clean this up.

ishustava: could we change the `consul-k8s-component-authmethod` const? I think it's only used in tests anyways

control-plane/subcommand/server-acl-init/command_ent_test.go
ishustava: Maybe we should only have them for the components that have NS logic? This test and the test above, don't check anything namespace-related, so I'm not sure if they could ever fail.

control-plane/subcommand/server-acl-init/command_ent_test.go
thisisnotashwin: thats true. On it!

control-plane/subcommand/server-acl-init/command_ent_test.go
thisisnotashwin: Hmm. That makes sense. These are not related to controller logic but make sense for connect-inject and sync. I can remove them from here and keep these tests only for tokens that are impacted by namespaces being enabled.

charts/consul/templates/controller-clusterrole.yaml
```diff
@@ -57,15 +57,6 @@ rules:
   - get
   - list
   - update
-{{- if .Values.global.acls.manageSystemACLs }}
```
kschoche, line 60: nice catch!

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -1168,16 +1174,256 @@ func TestRun_PoliciesAndBindingRulesForACLLoginNamespacesEnabled(t *testing.T) {
...
+				"-server-port", strings.Split(consulHTTPAddr, ":")[1],
+			}, c.TokenFlags...)
+			cmd.init()
+			responseCode := cmd.Run(cmdArgs)
+			require.Equal(t, 0, responseCode, ui.ErrorWriter.String())
+
+			// Check that the Role exists + has correct Policy and is associated with a BindingRule.
+			for i := range c.Roles {
+				// Check that the Policy exists.
+
```
kschoche, line 1237:
```suggestion
```

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2238,6 +2221,431 @@ func TestRun_GatewayErrors(t *testing.T) {
...
+				"-server-port", strings.Split(consulHTTPAddr, ":")[1],
+			}, c.TokenFlags...)
+			cmd.init()
+			responseCode := cmd.Run(cmdArgs)
+			require.Equal(t, 0, responseCode, ui.ErrorWriter.String())
+
+			// Check that the Role exists + has correct Policy and is associated with a BindingRule.
+			for i := range c.Roles {
+				// Check that the Policy exists.
+
```
kschoche, line 2371:
```suggestion
```

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2238,6 +2221,431 @@ func TestRun_GatewayErrors(t *testing.T) {
...
+	require.Equal(t, 0, responseCode, ui.ErrorWriter.String())
+
+	// Check that the expected local auth-method was created.
+	localAuthMethod, _, err := consul.ACL().AuthMethodRead(resourcePrefix+"-k8s-component-auth-method", &api.QueryOptions{})
+	require.NoError(t, err)
+	require.NotNil(t, localAuthMethod)
+	// Check that the expected global auth-method was created.
+	globalAuthMethod, _, err := consul.ACL().AuthMethodRead(resourcePrefix+"-k8s-component-auth-method-dc2", &api.QueryOptions{Datacenter: "dc1"})
+	require.NoError(t, err)
+	require.NotNil(t, globalAuthMethod)
```
kschoche, line 2646: does it make sense to assert that the globalAuthMethod's datacenter is correct, or are we cool bc of the QueryOption?

kschoche (approved):
🔥 
Great work, I love all the new tests as well!!

control-plane/subcommand/server-acl-init/command_test.go
thisisnotashwin: The query option should be sufficient here! good question though.

PR #1076, opened 2022-03-06
Connect inject acl refactor

thisisnotashwin:
Changes proposed in this PR:
- Refactor connect-injector to use the new auth-method workflow when ACLs are enabled so that Kubernetes secrets are not used.

How I've tested this PR:
- Unit tests.
- Acceptance tests.

- Code Review

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -66,6 +66,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            {{- if .Values.global.acls.manageSystemACLs }}
+            - name: CONSUL_HTTP_TOKEN_FILE
+              value: "/consul/login/acl-token"
+            {{- end }}
             {{- if .Values.global.tls.enabled }}
             - name: CONSUL_CACERT
               value: /consul/tls/ca/tls.crt
@@ -80,12 +84,6 @@ spec:
                 secretKeyRef:
                   name: {{ .Values.connectInject.aclInjectToken.secretName }}
                   key: {{ .Values.connectInject.aclInjectToken.secretKey }}
-            {{- else if .Values.global.acls.manageSystemACLs }}
-            - name: CONSUL_HTTP_TOKEN
-              valueFrom:
-                secretKeyRef:
-                  name: "{{ template "consul.fullname" . }}-connect-inject-acl-token"
-                  key: "token"
             {{- end }}
             - name: CONSUL_HTTP_ADDR
               {{- if .Values.global.tls.enabled }}
@@ -216,6 +214,16 @@ spec:
                 -default-consul-sidecar-cpu-request={{ $consulSidecarResources.requests.cpu }} \
                 {{- end }}
                 {{- end }}
+          {{- if .Values.global.acls.manageSystemACLs }}
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                - "/bin/sh"
+                - "-ec"
+                - |
+                  consul-k8s-control-plane consul-logout
+          {{- end }}
           startupProbe:
             httpGet:
               path: /readyz/ready
@@ -246,6 +254,9 @@ spec:
           - name: certs
             mountPath: /etc/connect-injector/certs
             readOnly: true
+          - mountPath: /consul/login
+            name: consul-data
+            readOnly: true
           {{- if .Values.global.tls.enabled }}
           {{- if .Values.global.tls.enableAutoEncrypt }}
           - name: consul-auto-encrypt-ca-cert
@@ -264,6 +275,9 @@ spec:
         secret:
           defaultMode: 420
           secretName: {{ template "consul.fullname" . }}-connect-inject-webhook-cert
+      - name: consul-data
+        emptyDir:
+          medium: "Memory"
       {{- if .Values.global.tls.enabled }}
       {{- if not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots) }}
       - name: consul-ca-cert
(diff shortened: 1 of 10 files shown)


charts/consul/templates/server-acl-init-job.yaml
```diff
@@ -183,7 +183,7 @@ spec:
                 {{- end }}
 
                 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
-                -create-inject-token=true \
+                -create-inject-policy=true \
```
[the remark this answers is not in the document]
thisisnotashwin: yup! we will no longer be creating tokens, so we wanted to rename them to be more representative.

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -285,16 +299,57 @@ spec:
...
+          {{- if .Values.global.tls.enabled }}
+        - name: CONSUL_CACERT
+          value: /consul/tls/ca/tls.crt
+          {{- end }}
+        - name: CONSUL_HTTP_ADDR
+            {{- if .Values.global.tls.enabled }}
+          value: https://$(HOST_IP):8501
+            {{- else }}
+          value: http://$(HOST_IP):8500
+            {{- end }}
```
ishustava, line 321: need to fix indentation pls 🙏

charts/consul/templates/server-acl-init-job.yaml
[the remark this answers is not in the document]
ishustava: I thought we're renaming them to be `-enable-x` or `-x` from https://github.com/hashicorp/consul-k8s/pull/1072

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -527,14 +526,22 @@ func (c *Command) Run(args []string) int {
...
 		// to be allowed to create namespaces.
 		if c.flagEnableNamespaces {
-			err = c.createGlobalACL("connect-inject", injectRules, consulDC, primary, consulClient)
+			// Create the controller ACL Policy, Role and BindingRule but do not issue any ACLTokens or create Kube Secrets.
+			// ConnectInjector token must be global when namespaces are enabled. This means secondary datacenters need
+			// a token that is known by the primary datacenters.
+			if !primary {
+				componentAuthMethodName = globalComponentAuthMethodName
+			}
+			err = c.createACLPolicyRoleAndBindingRule("connect-inject", injectRules, consulDC, primaryDC, globalToken, primary, componentAuthMethodName, serviceAccountName, consulClient)
```
ishustava, line 541: we could also just pass the `globalComponentAuthMethodName` here and avoid the `if` above:
```suggestion
			err = c.createACLPolicyRoleAndBindingRule("connect-inject", injectRules, consulDC, primaryDC, globalToken, primary, globalComponentAuthMethodName, serviceAccountName, consulClient)
```

Also do we still need to pass `globalToken` and `localToken`? If we do, maybe we should rename those since we're not creating tokens and it reads a bit confusing IMO

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -139,12 +139,11 @@ func (c *Command) init() {
 	var unused bool
 	c.flags.BoolVar(&unused, "create-inject-namespace-token", false,
 		"Toggle for creating a connect injector token. Only required when namespaces are enabled. "+
```
ishustava, line 141: should we remove this flag? I think it was added for backward compatibility but at this point we don't support different versions of consul-k8s-control-plane and the helm chart

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -139,12 +139,11 @@ func (c *Command) init() {
...
-			"Deprecated: set -enable-namespaces and -create-inject-token instead.")
-
-	c.flags.BoolVar(&c.flagCreateInjectToken, "create-inject-auth-method", false,
-		"Toggle for creating a connect inject auth method. Deprecated: use -create-inject-token instead.")
-	c.flags.BoolVar(&c.flagCreateInjectToken, "create-inject-token", false,
-		"Toggle for creating a connect inject auth method and an ACL token.")
+			"Deprecated: set -enable-namespaces and -create-inject-policy instead.")
+	c.flags.BoolVar(&c.flagCreateInjectPolicy, "create-inject-auth-method", false,
+		"Toggle for creating a connect inject auth method. Deprecated: use -create-inject-policy instead.")
```
ishustava, line 144: same here: we prob can remove the deprecated flag since probably no one can run helm chart with that version of consul-k8s anyway

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -1259,8 +1254,8 @@ func TestRun_SecondaryDatacenter_PoliciesAndBindingRulesForACLLogin_NamespacesEn
 				require.NoError(t, err)
 				require.NotNil(t, rb)
 				found = false
-				for x := range rb {
-					if rb[x].BindName == c.Roles[i] {
+				for j := range rb {
+					if rb[j].BindName == c.Roles[i] {
 						found = true
```
ishustava, line 1259: this test is the same as when namespaces are disabled (OSS). Is there anything ns-specific we should assert on here? should we check here that e.g. these are always created in the default ns because the binding rule creation ignores namespace rules like mirroring and destination ns.

ishustava (commented):
Awesome ❤️  I left some questions in-line

charts/consul/templates/server-acl-init-job.yaml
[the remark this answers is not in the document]
thisisnotashwin: let me fix that to be `-x`. 🙏

control-plane/subcommand/server-acl-init/command.go
thisisnotashwin: renamed these to `globalPolicy` and `localPolicy`

control-plane/subcommand/server-acl-init/command_ent_test.go
thisisnotashwin: That is a good point. I dont mind adding a different assertion here, but maybe the best idea is to not have this test duplicated and only have it run in OSS

ishustava: yeah maybe we don't need the roles checking since they are always set up the same with and w/o namespaces, but I think we can check that we skip namespace flags for the binding rule creation since there's logic there to `skipNamespacing`. I think that also means that we need to add multiple cases here for mirroring and destination ns to check that we always ignore those settings. wdyt?

thisisnotashwin: ooh. that is true. let me take a crack at this!

ishustava (approved):
:shipit:

jmurret (approved):
👏 Nice work!

PR #1083, opened 2022-03-08
Refactor Consul API Gateway Controller to use AuthMethod workflow.

thisisnotashwin:
Changes proposed in this PR:
- Refactor API gateway controller to use new authmethod ACL workflow.

How I've tested this PR:
- Unit tests.
- Acceptance tests.

- Code review.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -73,34 +70,47 @@ spec:
...
+            -consul-destination-namespace={{ .Values.apiGateway.consulNamespaces.consulDestinationNamespace }} \
+            {{- end }}
+            {{- if .Values.apiGateway.consulNamespaces.mirroringK8S }}
+            -mirroring-k8s=true \
+            {{- if .Values.apiGateway.consulNamespaces.mirroringK8SPrefix }}
+            -mirroring-k8s-prefix={{ .Values.apiGateway.consulNamespaces.mirroringK8SPrefix }} \
+            {{- end }}
+            {{- end }}
+            {{- end }}
+            -log-level {{ default .Values.global.logLevel .Values.apiGateway.logLevel }} \
```
kschoche, line 90: I know you didnt change this bit, but out of curiosity I wonder if apigw supports -log-json?

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -121,18 +131,57 @@ spec:
...
+        - "/bin/sh"
+        - "-ec"
+        - |
+          consul-k8s-control-plane acl-init \
+            -component-name=api-gateway-controller \
+            -acl-auth-method={{ template "consul.fullname" . }}-k8s-component-auth-method \
+            {{- if .Values.global.adminPartitions.enabled }}
+            -partition={{ .Values.global.adminPartitions.name }} \
+            {{- end }}
+            -log-level={{ default .Values.global.logLevel .Values.controller.logLevel }} \
```
kschoche, line 183: I wonder if `.Values.controller.logLevel` is the correct value for this controller?

thisisnotashwin: fixed it here: https://github.com/hashicorp/consul-k8s/pull/1085/files

charts/consul/templates/api-gateway-controller-deployment.yaml
kschoche: Yeah looks like they support `-log-json` could we add it?

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -73,34 +70,47 @@ spec:
...
+          readOnly: true
+        {{- if .Values.global.acls.manageSystemACLs }}
+        lifecycle:
+          preStop:
+            exec:
+              command:
+              - "/bin/sh"
+              - "-ec"
+              - |
+                consul-k8s-control-plane consul-logout
```
ishustava, line 112: 🚗 this binary is not available in this container. 🤔 I wonder how we should handle it here: maybe copy the consul binary?

ishustava: yeah, that makes sense

jmurret: I am using consul binary like [this](https://github.com/hashicorp/consul-k8s/pull/1070/files#diff-9b21e8a4c735f78620378655874a74263f5196a648e0e4c34e308150faf68f07R198).  Should I switch?

ishustava: I don't think so. In your case, the prestop hook will run in a container that's using the consul image so you're doing the right thing using the consul binary. 

For the api gateway, the image used is the gateway controller which doesn't have either consul-k8s-control-plane or the consul binaries, and so we need to copy either one.

charts/consul/templates/api-gateway-controller-deployment.yaml
thisisnotashwin: done!

charts/consul/templates/api-gateway-controller-deployment.yaml
thisisnotashwin: done! thanks for looking out!

PR #1086, opened 2022-03-08
Migrate most of build-and-test CircleCI jobs to Github Actions

curtbushko:
Changes proposed in this PR:
- Migrated most of the CircleCI jobs to Github Actions. Github Actions are missing several features that are standard in CircleCI such as test splitting, nice test outputs and being able to choose the size of the VM that you run tests on.  
- 3 jobs were not migrated: dev-docker-upload, acceptance-tproxy and acceptance. This is because the default runners that Github gives us are underpowered to run the acceptance tests. We will have to wait until the Summer of 2022 for Github to provide faster VMs.
- The above 3 jobs are in place, code wise, but they are commented out.
- The dev-docker-upload, acceptance-tproxy and acceptance jobs have been left intact in CircleCI.
- All duplicate jobs have been removed from CircleCI
- Nightly acceptance tests are left in place in CircleCI 
- We were missing gofmt and vet for the cli so I added that.

How I've tested this PR:

- I've run these jobs 100s of times.

👀


.circleci/config.yml
```diff
@@ -872,57 +872,14 @@ workflows:
   version: 2
   test-and-build:
     jobs:
-      # Fmt, vet, lint control plane and helm code
```
curtbushko, line 875: I removed all of these jobs because they were duplicates of what is now in build-and-test.yml

.circleci/config.yml
```diff
@@ -872,57 +872,14 @@ workflows:
...
-          ARCH: "386"
-          name: build-distros-386
-          requires:
-            - test-control-plane
-            - test-enterprise-control-plane
+     # Build this one control-plane binary so that acceptance and acceptance-tproxy will run
+     # The rest of these CircleCI jobs have been migrated to Github Actions. We need to wait until
+     # the summer of 2022 for larger puplic Github Action VMs be available before the acceptance tests can
+     # be moved
       - build-distro:
```
curtbushko, line 879: This is the only distro we need to run the acceptance-tproxy and acceptance tests.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,289 @@
+name: test-and-build 
+on:
+  push:
+
+env:
```
curtbushko, line 5: There is a limitation in Github Actions where you cannot pass an environment variable to a reusable workflows. You will the GOTESTSUM_VERSION being duplicated in several files because I cannot pass ${{ env.GOTESTSUM_VERSION }} into the inputs of reusable workflows. It is the same with go-version. I would have rather set a variable for that but it isn't possible 👎

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,289 @@
...
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: Validate helm gen 
+        working-directory: hack/helm-reference-gen 
+        run: |
+          go run ./... -validate
+
+  go-fmt-vet-helm-gen:
+    uses: hashicorp/consul-k8s/.github/workflows/reusable-go-fmt-vet.yml@crt-build-and-tests
```
curtbushko, line 39: I will change `@crt-build-and-tests` will to `@main` before I merge this PR.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,289 @@
...
+              rm consul_${{env.CONSUL_VERSION}}_linux_amd64.zip
+            chmod +x $HOME/bin/consul
+            PACKAGE_NAMES=$(go list ./...)
+            gotestsum --junitfile ${{env.TEST_RESULTS}}/gotestsum-report.xml -- -p 4 $PACKAGE_NAMES
+
+  test-enterprise-control-plane:
+    needs: [lint-control-plane, go-fmt-vet-control-plane]
+    runs-on: ubuntu-latest
+    env:
+      CONSUL_LICENSE: ${{secrets.CONSUL_LICENSE}}
```
curtbushko, line 134: You have to specifically pull in the secrets you want into the workflow, unlike CircleCI where it is just available to you as an environment variable.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,289 @@
...
+              rm consul_${{env.CONSUL_ENT_VERSION}}_linux_amd64.zip
+            chmod +x $HOME/bin/consul
+            PACKAGE_NAMES=$(go list ./...)
+            gotestsum --junitfile ${{env.TEST_RESULTS}}/gotestsum-report.xml -- -tags=enterprise -p 4 $PACKAGE_NAMES
+
+  build-distros:
+    needs: [test-control-plane, test-enterprise-control-plane]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
curtbushko, line 178: A matrix is how Github Actions lets you run parallel jobs. It is quite flexible.

.github/workflows/reusable-acceptance.yml
```diff
@@ -0,0 +1,128 @@
...
+        required: true
+        type: string
+      gotestsum-version:
+        required: true
+        type: string
+      kind-version:
+        required: false
+        type: string
+        default: "v1.22.4"
+    secrets:
```
curtbushko, line 30: You have to enable the secrets on both sides of the action, as the caller and the callee.

.github/workflows/reusable-acceptance.yml
```diff
@@ -0,0 +1,128 @@
...
+env:
+  TEST_RESULTS: /tmp/test-results # path to where test results are saved
+  CONSUL_ENT_LICENSE: ${{ secrets.CONSUL_ENT_LICENSE }} 
+
+jobs:
+  a:
+    runs-on: ubuntu-latest
+    strategy: 
+      matrix:
+        include: # I am really sorry for this but I could not find a way to automatically split our tests into several runners. For now, split manually.
```
curtbushko, line 44: Github Actions do not have a way to split tests like CircleCI does with `circlecI test split` so I had to split the tests manually so that they would run in parallel. 

This reusable workflow is not called right now but it is in place for the summer when larger VMs are available.

.github/workflows/reusable-acceptance.yml
```diff
@@ -0,0 +1,128 @@
...
+        run: |
+          kind create cluster --name dc1 --image kindest/node:${{ inputs.kind-version }}
+          kind create cluster --name dc2 --image kindest/node:${{ inputs.kind-version }}
+
+      # We have to run the tests for each package separately so that we can
+      # exit early if any test fails (-failfast only works within a single
+      # package).
+      - name: Run acceptance tests ${{ matrix.runner }}
+        working-directory: ${{ inputs.directory }}
+        if: github.repository_owner == 'hashicorp' # This prevents running on forks
```
curtbushko, line 93: `if: github.repository_owner == 'hashicorp'` prevents the tests from running on forks.

.github/workflows/reusable-acceptance.yml
```diff
@@ -0,0 +1,128 @@
...
+            then
+              echo "Tests in ${pkg} failed, aborting early"
+              exit_code=1
+              break
+            fi
+          done
+          gotestsum --raw-command --junitfile "${{ env.TEST_RESULTS }}/gotestsum-report.xml" -- cat jsonfile*
+          exit $exit_code
+    
+      - name: Upload tests
```
curtbushko, line 116: Github Actions does not have a pretty test page that you can look at so we will have to look at the artifacts for tests.

Debug output is added as an artifact if the tests fail.

jmurret (approved):
Nice work.  Thank you for the comments.  I'm not sure what I can eyeball that is going to be super helpful.  I think the proof is in the passing pipelines on GH Actions and CircleCI.

.circleci/config.yml
ishustava: Q1: we have some lint jobs defined in GHA already (e.g. https://github.com/hashicorp/consul-k8s/blob/main/.github/workflows/golangci-lint-control-plane.yml). Should we move them into one pipeline? 

Q2: The golangci linter also runs go fmt and so you'd get two failures in our pipeline: one from linter and one from the `go-fmt-and-vet*` jobs. Should we just have the linter? I think we can configure it to also run go vet (https://pkg.go.dev/cmd/vet). I'm not sure if it already runs by default. This question maybe is a bit out-of-scope so it could make sense to address it in a separate PR if we decide to do so.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,289 @@
...
+    with:
+      directory: hack/helm-reference-gen
+      go-version: 1.17.2
+
+  unit-test-helm-templates:
+    needs: [unit-helm-gen]
+    runs-on: ubuntu-latest
+    container: 
+      image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.10.0
+      options: --user root
```
ishustava, line 56: I'm curious about this one - do we need to be root to run these tests?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,289 @@
...
+          rm gotestsum_${{env.GOTESTSUM_VERSION}}_linux_amd64.tar.gz
+
+      - run:  mkdir -p ${{env.TEST_RESULTS}}
+      - run:  echo "$HOME/bin" >> $GITHUB_PATH
+
+      - name: Run go tests 
+        working-directory: control-plane 
+        run: |
+            mkdir -p $HOME/bin
+            wget https://releases.hashicorp.com/consul/${{env.CONSUL_VERSION}}/consul_${{env.CONSUL_VERSION}}_linux_amd64.zip && \
```
ishustava, line 123: nit: It might be worth splitting downloading the binary into a separate step. I'm seeing a lot of download output in this step, and it'd be nice to be able to scroll faster to the test output. Not sure if there's a limitation to doing so in GHA though

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,289 @@
...
+            chmod +x $HOME/bin/consul
+            PACKAGE_NAMES=$(go list ./...)
+            gotestsum --junitfile ${{env.TEST_RESULTS}}/gotestsum-report.xml -- -p 4 $PACKAGE_NAMES
+
+  test-enterprise-control-plane:
+    needs: [lint-control-plane, go-fmt-vet-control-plane]
+    runs-on: ubuntu-latest
+    env:
+      CONSUL_LICENSE: ${{secrets.CONSUL_LICENSE}}
+    steps:
```
ishustava, line 135: we need to prevent this job from running on forks

.github/workflows/reusable-go-fmt-vet.yml
```diff
@@ -0,0 +1,52 @@
...
+    inputs:
+      directory:
+        required: true
+        type: string
+      go-version:
+        required: true
+        type: string
+
+jobs:
+  a:
```
ishustava, line 14: I'm curious if we can name it a bit more descriptively? maybe `go-fmt-and-vet`?

control-plane/Makefile
```diff
@@ -27,22 +27,22 @@ CONSUL_K8S_IMAGE_VERSION?=latest
...
 	@docker pull hashicorp/consul-k8s:$(CONSUL_K8S_IMAGE_VERSION) >/dev/null #todo change this back after pulling it the first time since the dockerfile is FROM this image
 	@echo "Building consul-k8s Development container - $(CI_DEV_DOCKER_IMAGE_NAME)"
 	@docker build -t '$(CI_DEV_DOCKER_NAMESPACE)/$(CI_DEV_DOCKER_IMAGE_NAME):$(GIT_COMMIT)' \
 	--build-arg CONSUL_K8S_IMAGE_VERSION=$(CONSUL_K8S_IMAGE_VERSION) \
-	--label COMMIT_SHA=$(CIRCLE_SHA1) \
-	--label PULL_REQUEST=$(CIRCLE_PULL_REQUEST) \
-	--label CIRCLE_BUILD_URL=$(CIRCLE_BUILD_URL) \
+	--label COMMIT_SHA=$(GITHUB_SHA) \
+	--label PULL_REQUEST=$(GITHIB_PULL_REQUEST) \
+	--label GITHUB_BUILD_URL=$(GITHUB_SERVER_URL)/$(GITHUB_REPOSITORY)/actions/runs/$(GITHUB_RUN_ID) \
```
ishustava, line 40: Should we revert this change since we're keeping `dev-upload-docker` in circle for now?  Maybe we can define another makefile target so we don't forget how to define these correctly?

ishustava (commented):
Looks fantastic, Curt!! I really appreciate the refactoring you did to pull out common jobs into reusable workflows and all the comments throughout!

I'm leaving some questions in-line, but once we resolve them, this is good to go!

.github/workflows/build-and-test.yml
curtbushko: That's a good nit. I have separated it into 2 steps for both control-plane tests.

.github/workflows/reusable-go-fmt-vet.yml
curtbushko: This one is funky and I couldn't think of a nice way to do it. If you take a look at the visual diagram of the workflows (https://github.com/hashicorp/consul-k8s/actions/runs/1953797104) you will see that several of the job names are:

`go-fmt-vet-cli / a` or `go-fmt-vet-helm-gen / a`. This is because it is concatenating `<workflow job name> / <reusable job name>`. The constraints are that <workflow job name> has to be unique and I could not do something like `cli / go-fmt-vet` as the next `cli` job name in the workflow would be in conflict. Also, the reusable workflow has to have a job name or it will not work...

control-plane/Makefile
curtbushko: Good catch! I have reverted, added another make target `ci.dev-docker-github` and updated the commented out section of build-and-test.yml

.circleci/config.yml
curtbushko: @kschoche I followed the old workflow as much as possible. I lint early and try to parallel the jobs as much as possible. We can have more dependencies if we want to make sure all things stop right away if we want...


See: https://github.com/hashicorp/consul-k8s/actions/runs/1964296498

.github/workflows/build-and-test.yml
curtbushko: Good catch. I changed it to user 1001.

.github/workflows/reusable-go-fmt-vet.yml
ishustava: Yeah, naming is hard 😄 

What if we name it `go-fmt-and-vet` or `go-fmt-vet` or `fmt-vet`? Then it will show as:
- `go-fmt-vet-cli/go-fmt-and-vet`
- `go-fmt-vet-cli/go-fmt-vet`
- `go-fmt-vet-cli/fmt-vet` 

Another idea: what if we name them something also generic like `job` but slightly more specific than `a`? Then it'll be `go-fmt-vet-cli/job`.

I think it'd be nice to have more specific/descriptive names if at all possible. WDYT?

curtbushko: Naming is terribly hard! 

I think 'job' may be a key word and will cause GHA to fail.  How about `lint` since I moved go-fmt-vet to use golangci-lint?

ishustava (approved):

PR #1087, opened 2022-03-08
Give error if imageK8s is used

lkysow:
Give error for folks setting `imageK8s` instead of `imageK8S` (an easy mistake)

How I've tested this PR:
* unit tests

* unit tests

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added: not sure this warrants it?


kschoche:
🤣 🍭

ishustava:
🤔 maybe we should accept both instead of giving an error?

ndhanushkodi (approved):
I'm ok with this or accepting both names as Iryna mentioned

thisisnotashwin:
@lkysow lol @ the branch name! 

> 🤔 maybe we should accept both instead of giving an error?
I'm a touch worried about accepting both because it increases the likelihood of us introducing errors into our templates if we forget to template one of the vaules.

kschoche:
> @lkysow lol @ the branch name!
> 
> > 🤔 maybe we should accept both instead of giving an error?
> 
> I'm a touch worried about accepting both because it increases the likelihood of us introducing errors into our templates if we forget to template one of the vaules.

ditto

lkysow:
We looked at supporting both but there are other Helm values that use `K8S` so we wanted to keep things consistent and so are going with the error.

Issue #1037, opened 2022-02-17
Unable to launch consul client with podsecuritypolicy.


kevinlmadison:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

We are running in a secure environment with RKE2 and thus need to manually set permissions with podsecuritypolicies. Enabling the podsecurity policy with helm allows us to use  [8500,8502,8301] but not 8600 so the daemonset is failing to deploy. In the client psp I don't see any option to allow port 8600, does this need to be added to the template?

### Logs

```
  Warning  FailedCreate  1s (x10 over 22s)  daemonset-controller  Error creating: pods "consul-client-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.containers[0].hostPort: Invalid value: 8600: Host port 8600 is not allowed to be used. Allowed ports: [8500,8502,8301] spec.containers[0].hostPort: Invalid value: 8600: Host port 8600 is not allowed to be used. Allowed ports: [8500,8502,8301] spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.containers[0].hostPort: Invalid value: 8500: Host port 8500 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 8502: Host port 8502 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 8301: Host port 8301 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 8301: Host port 8301 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 8600: Host port 8600 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 8600: Host port 8600 is not allowed to be used. Allowed ports: []]

```

This goes away if I manually add port 8600 to the psp

### Environment details

If not already included, please provide the following:
- `consul-k8s` version: 1.11.2
- `values.yaml`:
```
global:
  name: consul
[...]

PR #1090, opened 2022-03-09
Support hostNetwork with pod security policies

lkysow:
Changes proposed in this PR:
- fixes https://github.com/hashicorp/consul-k8s/issues/1037
- when `hostNetwork` is true, kube sets all containerPorts as hostPorts. We then need to allow those ports in our PSP.

How I've tested this PR:
- bats only

- bats

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ IMPROVEMENTS:
 * Control Plane
   * Upgrade Docker image Alpine version from 3.14 to 3.15. [[GH-1058](https://github.com/hashicorp/consul-k8s/pull/1058)]
 
+BUG FIXES:
+* Helm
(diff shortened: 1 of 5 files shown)


charts/consul/templates/server-podsecuritypolicy.yaml
```diff
@@ -28,14 +28,26 @@ spec:
     - 'persistentVolumeClaim'
   hostNetwork: false
   hostPorts:
-  {{- if .Values.server.exposeGossipAndRPCPorts }}
+  {{- if (or .Values.server.exposeGossipAndRPCPorts .Values.server.hostNetwork) }}
```
ishustava, line 31: We currently don't support hostNetwork mode for the servers

ishustava (commented):
I think we also need to do it for the mesh gateway PSP.

ishustava (approved):
Thank you for fixing it!! :shipit:

curtbushko (approved):

PR #1101, opened 2022-03-16
Refactor logic in common.Login

ishustava:
Changes proposed in this PR:
- convert function args to a struct
- add some missing tests
- move logic that is only relevant for connect out

How I've tested this PR:
unit tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/subcommand/common/common.go
+++ b/control-plane/subcommand/common/common.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/cenkalti/backoff"
 	"github.com/go-logr/logr"
-	"github.com/hashicorp/consul-k8s/control-plane/consul"
 	godiscover "github.com/hashicorp/consul-k8s/control-plane/helper/go-discover"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-discover"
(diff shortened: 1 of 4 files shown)


control-plane/subcommand/common/common.go
```diff
@@ -86,60 +85,72 @@ func ValidateUnprivilegedPort(flagName, flagValue string) error {
...
 		return err
-	}, backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), numLoginRetries))
+	}, backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), params.numRetries))
 	if err != nil {
-		if serviceAccountName == "default" {
-			log.Warn("The service account name for this Pod is \"default\"." +
-				" In default installations this is not a supported service account name." +
-				" The service account name must match the name of the Kubernetes Service" +
-				" or the consul.hashicorp.com/connect-service annotation.")
-		}
```
ishustava, line 132: I've moved this back into `connect-init` as I think it's only relevant there, and so it doesn't make sense to put it in `common`.

control-plane/subcommand/common/common.go
```diff
@@ -86,60 +85,72 @@ func ValidateUnprivilegedPort(flagName, flagValue string) error {
...
 		log.Error("Hit maximum retries for consul login", "error", err)
 		return "", err
 	}
-	// Now update the client so that it will read the ACL token we just fetched.
-	cfg.TokenFile = tokenSinkFile
-	client, err = consul.NewClient(cfg)
-	if err != nil {
-		log.Error("Unable to update client connection", "error", err)
-		return "", err
-	}
```
ishustava, line 142: I've removed the client config side-effect because we can instead set the token below via query params and so there's no need to pass config pointer.

control-plane/subcommand/common/common_test.go
```diff
@@ -199,15 +299,16 @@ func TestGetResolvedServerAddresses(t *testing.T) {
 // startMockServer starts an httptest server used to mock a Consul server's
 // /v1/acl/login endpoint. apiCallCounter will be incremented on each call to /v1/acl/login.
 // It returns a consul client pointing at the server.
-func startMockServer(t *testing.T, apiCallCounter *int) (*api.Client, *api.Config) {
-
+func startMockServer(t *testing.T) *api.Client {
```
ishustava, line 302: we don't need the counter anymore I think. Previously, the mock server always returned `testLoginResponse` regardless of the URL path and method, and so we instead relied on the counter to make sure this call is made. If we only return successful response for the right path+method, we don't need the counter.

control-plane/subcommand/common/common_test.go
```diff
@@ -54,54 +54,154 @@ func TestValidateUnprivilegedPort(t *testing.T) {
...
+	require.NoError(t, err)
 	// Validate that the token file was written to disk.
 	data, err := ioutil.ReadFile(tokenFile)
-	require.NoError(err)
-	require.Equal(string(data), "b78d37c7-0ca7-5f4d-99ee-6d9975ce4586")
+	require.NoError(t, err)
+	require.Equal(t, string(data), "b78d37c7-0ca7-5f4d-99ee-6d9975ce4586")
+}
+
+// TestConsulLogin_Retries tests we retry /v1/acl/login call if it fails.
```
jmurret, line 79: Thank you for these tests!

control-plane/subcommand/connect-init/command.go
```diff
@@ -123,11 +123,25 @@ func (c *Command) Run(args []string) int {
...
+		loginParams := common.LoginParams{
+			AuthMethod:      c.flagACLAuthMethod,
+			Namespace:       c.flagAuthMethodNamespace,
+			BearerTokenFile: c.flagBearerTokenFile,
+			TokenSinkFile:   c.flagACLTokenSink,
+			Meta:            loginMeta,
+		}
+		token, err := common.ConsulLogin(consulClient, loginParams, c.logger)
 		if err != nil {
+			if c.flagServiceAccountName == "default" {
```
jmurret, line 135: Do we need to wait for an error to make this kind of validation? Or can we put it with the validateFlag-type guard clauses at the beginning of `Run()`?  
(If so, it would also make writing a test for this easier.)

thisisnotashwin (approved):
Thanks so much for doing this Iryna!! This is a work of art! Left a review that isn't blocking, but more food for thought.

control-plane/subcommand/connect-init/command.go
ishustava: That's a great question. I had to dig up git blame on this to find out why. Looks like we don't want to fail early or fail in general because https://github.com/hashicorp/consul-k8s/commit/8a63540547ababee7d0ff245a9b18308fa5e1dcd

We also can't give this warning message earlier because we only want to do it if there's an error so we can help the user find if there's a misconfiguration.

control-plane/subcommand/common/common.go
```diff
@@ -86,71 +85,83 @@ func ValidateUnprivilegedPort(flagName, flagValue string) error {
...
+	Meta map[string]string
+
+	// numRetries is only used in tests to make them run faster.
+	numRetries uint64
+}
+
 // ConsulLogin issues an ACL().Login to Consul and writes out the token to tokenSinkFile.
 // The logic of this is taken from the `consul login` command.
-func ConsulLogin(client *api.Client, cfg *api.Config, authMethodName, datacenter, namespace, bearerTokenFile, serviceAccountName, tokenSinkFile string, meta map[string]string, log hclog.Logger) (string, error) {
+func ConsulLogin(client *api.Client, params LoginParams, log hclog.Logger) (string, error) {
```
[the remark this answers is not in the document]
ishustava: Oh that's an interesting idea! My only reservation is that `ConsulLogin` doesn't make so much sense as a method on `LoginParams`. Something like `params.ConsulLogin(client, logger)` reads a bit counter-intuitive to me because I'm not seeing params as a struct that typically has a lot of methods as it's usually something you pass around. WDYT?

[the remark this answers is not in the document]
thisisnotashwin: hmm. we could add the api.Client to the struct, call it LoginHelper (name needs to be workshopped), and then have Login on that struct? It would feel more cohesive then IMO

jmurret (approved):
🙏  Thank you.  This looks great.  :shipit:

control-plane/subcommand/common/common.go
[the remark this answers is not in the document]
ishustava: I think I'm going to keep it as is for now. I thought about adding maybe a sort of a wrapped consul client but then it wouldn't make sense to use it just for login and maybe we'd want to re-use it for something else, in which case it also doesn't make sense to add params to that struct. It might be a bigger rewrite also.

PR #1102, opened 2022-03-16
Enable terminating gateways to use ACL Auth Method

jmurret:
Changes proposed in this PR:
- change terminating gateway deployment to pass `acl-auth-method` so it can properly call Consul login.
- add prestop for logout to terminating gateway container.
- refactor logic from `server-acl-init` command for terminating gateway so that it can also be used for upcoming ingress gateway work.

How I've tested this PR:
- unit tests
- acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/acceptance/tests/terminating-gateway/terminating_gateway_test.go
+++ b/acceptance/tests/terminating-gateway/terminating_gateway_test.go
@@ -69,11 +69,11 @@ func TestTerminatingGateway(t *testing.T) {
 			// Register the external service
 			registerExternalService(t, consulClient, "")
 
-			// If ACLs are enabled we need to update the token of the terminating gateway
+			// If ACLs are enabled we need to update the role of the terminating gateway
 			// with service:write permissions to the static-server service
 			// so that it can can request Connect certificates for it.
 			if c.secure {
-				updateTerminatingGatewayToken(t, consulClient, staticServerPolicyRules)
+				updateTerminatingGatewayRole(t, consulClient, staticServerPolicyRules)
 			}
 
 			// Create the config entry for the terminating gateway.
@@ -133,32 +133,32 @@ func registerExternalService(t *testing.T, consulClient *api.Client, namespace s
 	require.NoError(t, err)
 }
 
-func updateTerminatingGatewayToken(t *testing.T, consulClient *api.Client, rules string) {
+func updateTerminatingGatewayRole(t *testing.T, consulClient *api.Client, rules string) {
 	t.Helper()
 
-	// Create a write policy for the static-server.
+	logger.Log(t, "creating a write policy for the static-server")
 	_, _, err := consulClient.ACL().PolicyCreate(&api.ACLPolicy{
 		Name:  "static-server-write-policy",
 		Rules: rules,
 	}, nil)
 	require.NoError(t, err)
 
-	// Get the terminating gateway token.
-	tokens, _, err := consulClient.ACL().TokenList(nil)
+	logger.Log(t, "getting the terminating gateway role")
+	roles, _, err := consulClient.ACL().RoleList(nil)
 	require.NoError(t, err)
-	var termGwTokenID string
-	for _, token := range tokens {
-		if strings.Contains(token.Description, "token created via login: {\"component\":\"terminating-gateway\"}") {
-			termGwTokenID = token.AccessorID
+	terminatingGatewayRoleID := ""
+	for _, role := range roles {
+		if strings.Contains(role.Name, "terminating-gateway") {
+			terminatingGatewayRoleID = role.ID
 			break
 		}
 	}
-	termGwToken, _, err := consulClient.ACL().TokenRead(termGwTokenID, nil)
-	require.NoError(t, err)
 
-	// Add policy to the token and update it
-	termGwToken.Policies = append(termGwToken.Policies, &api.ACLTokenPolicyLink{Name: "static-server-write-policy"})
-	_, _, err = consulClient.ACL().TokenUpdate(termGwToken, nil)
+	logger.Log(t, "update role with policy")
+	termGwRole, _, err := consulClient.ACL().RoleRead(terminatingGatewayRoleID, nil)
+	require.NoError(t, err)
+	termGwRole.Policies = append(termGwRole.Policies, &api.ACLTokenPolicyLink{Name: "static-server-write-policy"})
+	_, _, err = consulClient.ACL().RoleUpdate(termGwRole, nil)
 	require.NoError(t, err)
 }
 
--- a/acceptance/tests/terminating-gateway/terminating_gateway_test.go
+++ b/acceptance/tests/terminating-gateway/terminating_gateway_test.go
@@ -148,7 +148,7 @@ func updateTerminatingGatewayToken(t *testing.T, consulClient *api.Client, rules
 	require.NoError(t, err)
 	var termGwTokenID string
 	for _, token := range tokens {
-		if strings.Contains(token.Description, "terminating-gateway-terminating-gateway-token") {
+		if strings.Contains(token.Description, "token created via login: {\"component\":\"terminating-gateway\"}") {
 			termGwTokenID = token.AccessorID
 			break
 		}
--- a/charts/consul/templates/terminating-gateways-deployment.yaml
+++ b/charts/consul/templates/terminating-gateways-deployment.yaml
@@ -24,7 +24,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ template "consul.fullname" $root }}-{{ .name }}
+  name: {{ template "consul.fullname" $root }}-{{ .name }}-terminating-gateway
   namespace: {{ $root.Release.Namespace }}
   labels:
     app: {{ template "consul.name" $root }}
@@ -32,7 +32,7 @@ metadata:
     heritage: {{ $root.Release.Service }}
     release: {{ $root.Release.Name }}
     component: terminating-gateway
-    terminating-gateway-name: {{ template "consul.fullname" $root }}-{{ .name }}
+    terminating-gateway-name: {{ template "consul.fullname" $root }}-{{ .name }}-terminating-gateway
 spec:
   replicas: {{ default $defaults.replicas .replicas }}
   selector:
@@ -42,7 +42,7 @@ spec:
       heritage: {{ $root.Release.Service }}
       release: {{ $root.Release.Name }}
       component: terminating-gateway
-      terminating-gateway-name: {{ template "consul.fullname" $root }}-{{ .name }}
+      terminating-gateway-name: {{ template "consul.fullname" $root }}-{{ .name }}-terminating-gateway
   template:
     metadata:
       labels:
@@ -51,7 +51,7 @@ spec:
         heritage: {{ $root.Release.Service }}
         release: {{ $root.Release.Name }}
         component: terminating-gateway
-        terminating-gateway-name: {{ template "consul.fullname" $root }}-{{ .name }}
+        terminating-gateway-name: {{ template "consul.fullname" $root }}-{{ .name }}-terminating-gateway
       annotations:
         {{- if (and $root.Values.global.secretsBackend.vault.enabled $root.Values.global.tls.enabled) }}
         "vault.hashicorp.com/agent-init-first": "true"
--- a/charts/consul/templates/terminating-gateways-deployment.yaml
+++ b/charts/consul/templates/terminating-gateways-deployment.yaml
@@ -179,13 +179,13 @@ spec:
             - name: CONSUL_HTTP_ADDR
               value: http://$(HOST_IP):8500
             {{- end }}
-          command:
+          command:  
             - "/bin/sh"
             - "-ec"
             - |
                 {{- if $root.Values.global.acls.manageSystemACLs }}
                 consul-k8s-control-plane acl-init \
-                  -component-name=terminating-gateway \
+                  -component-name=terminating-gateway/{{ template "consul.fullname" $root }}-{{ .name }} \
                   -acl-auth-method={{ template "consul.fullname" $root }}-k8s-component-auth-method \
                   {{- if $root.Values.global.adminPartitions.enabled }}
                   -partition={{ $root.Values.global.adminPartitions.name }} \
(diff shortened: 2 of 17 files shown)


charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -97,6 +97,9 @@ spec:
         - name: consul-service
           emptyDir:
             medium: "Memory"
+        - name: consul-data
```
jmurret, line 100: We have used `consul-data` in some existing helm charts like mesh gateway and others and in other work like client token others, we have used `consul-login`. I will look at trying to align those outside of this PR.

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -666,58 +666,12 @@ func (c *Command) Run(args []string) int {
 	}
 
 	if len(c.flagTerminatingGatewayNames) > 0 {
-		// Create a token for each terminating gateway name. Each gateway needs a
```
jmurret, line 669: extracting this code into a private helper function so that it can also be used for upcoming ingress gateway work.

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2192,6 +2153,18 @@ func TestRun_PoliciesAndBindingRulesForACLLogin_PrimaryDatacenter(t *testing.T)
...
 		},
+		{
+			TestName: "Terminating Gateway",
+			TokenFlags: []string{"-terminating-gateway-name=terminating",
+				"-terminating-gateway-name=gateway",
+				"-terminating-gateway-name=another-gateway"},
+			PolicyNames: []string{"terminating-terminating-gateway-policy",
+				"gateway-terminating-gateway-policy",
+				"another-gateway-terminating-gateway-policy"},
+			Roles: []string{"release-name-consul-terminating-acl-role",
```
jmurret, line 2164: Are there any scenarios besides additional data centers where someone can define a duplicate gateway name? For example, could I define:
- terminating gateway named foo (where role would be `release-name-consul-foo-acl-role`)
- ingress gateway named foo (where role would be `release-name-consul-foo-acl-role`)

This seems like a possible issue?

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -252,6 +260,9 @@ spec:
           volumeMounts:
           - name: consul-bin
             mountPath: /consul-bin
+          - mountPath: /consul/service
```
jmurret, line 263: We have used `consul-data` in some existing helm charts like mesh gateway and others and in other work like client token others, we have used `consul-login`. I will look at trying to align those outside of this PR.

thisisnotashwin: Thanks @jmurret !! I did realize that while I was working on the mesh-gateway and I should have fixed it up, but I ran into some issues during my testing and was addressing bugs one failure at a time and forgot to then come back and rename all the things! this would be really appreciated 🙏

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -345,6 +350,10 @@ spec:
                       -partition={{ $root.Values.global.adminPartitions.name }} \
                       {{- end }}
                       -id="${POD_NAME}"
+                  {{- if $root.Values.global.acls.manageSystemACLs }}
+                  - |
+                    consul-k8s-control-plane consul-logout
```
thisisnotashwin, line 355: The `consul-k8s-control-plane binary` will not be available on this container as this is an envoy container. you will have to issues a `/consul-bin/consul logout`. Something similar is done in the mesh gateway. Also, I like the way it is done here as a new entry into the slice. If this templating does work, can we update the mesh gateway to do something similar? I have used some `&&` logic there but this looks cleaner.

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -804,6 +758,73 @@ func (c *Command) createAuthMethod(consulClient *api.Client, authMethod *api.ACL
 		})
 }
 
+type gatewayRulesGenerator func(name, namespace string) (string, error)
+
+func (c *Command) configureGateway(gatewayType string, gatewayNames []string,
+	authMethodName string, ruleGenerator gatewayRulesGenerator, consulDC string,
+	primaryDC string, primary bool, consulClient *api.Client) error {
```
thisisnotashwin, line 765: can we move some of the gateway-specific params here into a `GatewayConfig` struct that we can pass into this method to simplify the params?

control-plane/subcommand/server-acl-init/command_test.go
thisisnotashwin: I think it makes sense to update the role name logic to be similar to how the policy names are generated. This could be a problem AFAIR.

minor nit: can we swap out `release-name-consul` for `resourcePrefix` like in the other fixtures? 🙏

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2510,18 +2506,20 @@ func TestRun_ValidateLoginToken_PrimaryDatacenter(t *testing.T) {
 			})
 			require.NoError(t, err)
 
-			tok, _, err := client.ACL().Login(&api.ACLLoginParams{
-				AuthMethod:  authMethodName,
-				BearerToken: jwtToken,
-				Meta:        map[string]string{},
-			}, &api.WriteOptions{})
-			require.NoError(t, err)
+			if len(c.Roles) == 1 {
```
thisisnotashwin, line 2509: Can we create an assertion to ensure the roles created for terminating gateways are what we have supplied in the above? We are skipping that assertion right now right?

thisisnotashwin (commented):
This looks really good @jmurret !! Left some responses to some of the questions and added some comments

control-plane/subcommand/server-acl-init/command_test.go
jmurret: agg.  that was a remnant.  thanks for catching this!

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -804,6 +765,91 @@ func (c *Command) createAuthMethod(consulClient *api.Client, authMethod *api.ACL
...
+	//PrimaryDC is the name of the Primary Data Center
+	PrimaryDC string
+	//Primary specifies whether the ConsulDC is the Primary Data Center
+	Primary bool
+}
+
+func (c *Command) configureGateway(gatewayParams ConfigureGatewayParams, consulClient *api.Client) error {
+	// Create a token for each gateway name. Each gateway needs a
+	// separate token because users may need to attach different policies
+	// to each gateway token depending on what the services it represents
```
ishustava, line 791:
```suggestion
	// Each gateway needs to be configured
	// separately because users may need to attach different policies
	// to each gateway role depending on what services it represents.
```

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -804,6 +765,91 @@ func (c *Command) createAuthMethod(consulClient *api.Client, authMethod *api.ACL
...
+		}
+
+		// Define the gateway rules
+		rules, err := gatewayParams.RulesGenerator(name, namespace)
+		if err != nil {
+
+			errMessage := fmt.Sprintf("Error templating %s gateway rules",
+				gatewayParams.GatewayType)
+			c.log.Error(errMessage, "gateway-name", name,
+				"namespace", namespace, "err", err)
```
ishustava, line 833: should we also include gateway type in the log message?

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -804,6 +765,91 @@ func (c *Command) createAuthMethod(consulClient *api.Client, authMethod *api.ACL
...
+					namespace = parts[1]
+				} else {
+					namespace = consulDefaultNamespace
+				}
+			} else {
+				// Use the default Consul namespace
+				namespace = consulDefaultNamespace
+			}
+		} else if strings.ContainsAny(name, ".") {
+			errMessage := "Gateway names shouldn't include a namespace if Consul namespaces aren't enabled"
```
ishustava, line 821: I think error messages start with lowercase in go so that they can be wrapped.
```suggestion
			errMessage := "gateway names shouldn't include a namespace if Consul namespaces aren't enabled"
```

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2645,20 +2653,22 @@ func TestRun_ValidateLoginToken_SecondaryDatacenter(t *testing.T) {
...
-				}, &api.WriteOptions{})
-				require.NoError(r, err)
-
-				require.Equal(r, len(tok.Roles), len(c.Roles))
-				for _, role := range tok.Roles {
-					require.Contains(r, c.Roles, role.Name)
-				}
-				require.Equal(r, !c.GlobalToken, tok.Local)
-			})
+			if len(c.Roles) == 1 {
```
ishustava, line 2656: I'm a bit confused by this logic. Why do we only do it if we have one role?

control-plane/subcommand/server-acl-init/command_test.go
```diff
@@ -2588,6 +2585,17 @@ func TestRun_ValidateLoginToken_SecondaryDatacenter(t *testing.T) {
...
+		{
+			ComponentName: "terminating gateway",
+			TokenFlags: []string{"-terminating-gateway-name=terminating",
+				"-terminating-gateway-name=gateway",
+				"-terminating-gateway-name=another-gateway"},
+			Roles: []string{resourcePrefix + "-terminating-acl-role-dc2",
+				resourcePrefix + "-gateway-acl-role-dc2",
+				resourcePrefix + "-another-gateway-acl-role-dc2"},
+			GlobalAuthMethod: false,
+			GlobalToken:      false,
```
ishustava, line 2597: for the login test we don't really benefit from providing multiple gateway names since we're stubbing out k8s call, so I think we can test it with just one name/role

control-plane/subcommand/server-acl-init/command_test.go
jmurret: Re-typing for completeness of other viewers. This was a remnant.

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -804,6 +764,90 @@ func (c *Command) createAuthMethod(consulClient *api.Client, authMethod *api.ACL
 		})
 }
 
+type gatewayRulesGenerator func(name, namespace string) (string, error)
+
+// ConfigureGatewayParams are parameters used to configure Ingress and Terminating Gateways.
+type ConfigureGatewayParams struct {
+	//GatewayType specifies whether it is an ingress or terminating gateway
```
ishustava, line 771: nit
```suggestion
	// GatewayType specifies whether it is an ingress or terminating gateway.
```

I'm surprised it didn't get `go fmt`'d, but maybe it doesn't look at struct fields?

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -804,6 +764,90 @@ func (c *Command) createAuthMethod(consulClient *api.Client, authMethod *api.ACL
...
+	Primary bool
+}
+
+func (c *Command) configureGateway(gatewayParams ConfigureGatewayParams, consulClient *api.Client) error {
+	// Each gateway needs to be configured
+	// separately because users may need to attach different policies
+	// to each gateway role depending on what services it represents.
+	for _, name := range gatewayParams.GatewayNames {
+		if name == "" {
+			errMessage := fmt.Sprintf("%s gateway names cannot be empty",
```
ishustava, line 793:
```suggestion
			errMessage := fmt.Sprintf("%s gateway name cannot be empty",
```

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -804,6 +764,90 @@ func (c *Command) createAuthMethod(consulClient *api.Client, authMethod *api.ACL
...
+			errMessage := "gateway names shouldn't include a namespace if Consul namespaces aren't enabled"
+			c.log.Error(errMessage, "gateway-name", name)
+			return errors.New(errMessage)
+		}
+
+		// Define the gateway rules
+		rules, err := gatewayParams.RulesGenerator(name, namespace)
+		if err != nil {
+
+			errMessage := fmt.Sprintf("Error templating %s gateway rules",
```
ishustava, line 829: err messages should start with lowercase in golang
```suggestion
			errMessage := fmt.Sprintf("error templating %s gateway rules",
```

ishustava (approved):
Looks great, John! I left a couple of edits, but nothing blocking! ❤️ the refactor to configure gateways.

acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -69,11 +69,11 @@ func TestTerminatingGateway(t *testing.T) {
 			// Register the external service
 			registerExternalService(t, consulClient, "")
 
-			// If ACLs are enabled we need to update the token of the terminating gateway
+			// If ACLs are enabled we need to update the role of the terminating gateway
 			// with service:write permissions to the static-server service
 			// so that it can can request Connect certificates for it.
 			if c.secure {
-				updateTerminatingGatewayToken(t, consulClient, staticServerPolicyRules)
+				updateTerminatingGatewayRole(t, consulClient, staticServerPolicyRules)
```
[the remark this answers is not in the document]
jmurret: 👍 When Luke asked me about what I was working on last week, he mentioned [this part of that doc](https://www.consul.io/docs/k8s/connect/terminating-gateways#update-terminating-gateway-acl-token-if-acls-are-enabled) would need to change to reflect updating role rather than the token.  I used what he described to do the acceptance test change and plan on updating this doc now based on the acceptance test change.

acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -133,32 +133,32 @@ func registerExternalService(t *testing.T, consulClient *api.Client, namespace s
...
 	_, _, err := consulClient.ACL().PolicyCreate(&api.ACLPolicy{
 		Name:  "static-server-write-policy",
 		Rules: rules,
 	}, nil)
 	require.NoError(t, err)
 
-	// Get the terminating gateway token.
-	tokens, _, err := consulClient.ACL().TokenList(nil)
+	logger.Log(t, "getting the terminating gateway role")
+	roles, _, err := consulClient.ACL().RoleList(nil)
```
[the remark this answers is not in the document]
jmurret: I will follow up with this.  I originally tried it, but I don't think I was passing in the correct name.

PR #1103, opened 2022-03-16
bug: exec'ing into consul server container no longer works when tls is enabled

curtbushko:
Vault secrets backend.

When TLS is enabled we set env vars inside consul server container so that we can run consul cli command when exec'ing into it. With vault providing TLS certs/CAs, we should now point to the  correct file where the CA is stored instead of the old location where the kube CA cert is normally stored: https://github.com/hashicorp/consul-k8s/blob/93cc206579578cd467420f7f92f67811271efd21/charts/consul/templates/server-statefulset.yaml#L226

We might have to do it for consul clients as well.

Changes proposed in this PR:
- changed the helm templates for client-daemonset and server-statefulset to set CONSUL_CACERT to the correct vault secrets location (/vaul/secrets/serverca.crt)
- added a check in the vault acceptance test to verify that CONSUL_CACERT is set in the <release name>-consul-server-0 container using the kubeapi
- added bats checks to both client-daemonset and server-statefulset to check that CONSUL_CACERT is set during helm template 

Note:

There are other places in the helm templates that are also using CONSUL_CACERT. I will create a ticket for those and add it to the project backlog as I am not sure if they are valid or not.

How I've tested this PR:
- verified that exec'ing into the server container and running `consul info` works
- acceptance test
- bat tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -1,6 +1,8 @@
 package vault
 
 import (
+	"context"
+	"fmt"
 	"testing"
 
 	"github.com/hashicorp/consul-k8s/acceptance/framework/consul"
@@ -9,6 +11,7 @@ import (
 	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/vault"
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const staticClientName = "static-client"
@@ -110,6 +113,20 @@ func TestVault(t *testing.T) {
 	caConfig, _, err := consulClient.Connect().CAGetConfig(nil)
 	require.NoError(t, err)
 	require.Equal(t, caConfig.Provider, "vault")
+
+	// Validate that tls.crt is not set to the default value
+	logger.Log(t, "Validating that CONSUL_CACERT has been set correctly")
+	serverPod, err := ctx.KubernetesClient(t).CoreV1().Pods(ns).Get(context.Background(), fmt.Sprintf("%s-consul-server-0", consulReleaseName), metav1.GetOptions{})
+	require.NoError(t, err)
+
+	// Environment variables are buried in the pod spec
+	containerVars := serverPod.Spec.Containers[0].Env
+	for k := range containerVars {
+		if containerVars[k].Name == "CONSUL_CACERT" {
+			require.NotEqual(t, containerVars[k].Value, "/consul/tls/ca/tls.crt")
+			require.Equal(t, containerVars[k].Value, "/vault/secrets/serverca.crt")
+		}
+	}
 
 	if cfg.EnableEnterprise {
 		// Validate that the enterprise license is set correctly.
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -1,6 +1,7 @@
 package vault
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
@@ -10,6 +11,7 @@ import (
 	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/vault"
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const staticClientName = "static-client"
@@ -66,7 +68,7 @@ func TestVault(t *testing.T) {
 		"global.secretsBackend.vault.connectCA.rootPKIPath":         "connect_root",
 		"global.secretsBackend.vault.connectCA.intermediatePKIPath": "dc1/connect_inter",
 
-		"global.acls.manageSystemACLs":       "false",
+		"global.acls.manageSystemACLs":       "true",
 		"global.tls.enabled":                 "true",
 		"global.gossipEncryption.secretName": "consul/data/secret/gossip",
 		"global.gossipEncryption.secretKey":  "gossip",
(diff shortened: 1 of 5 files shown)


acceptance/tests/vault/vault_test.go
```diff
@@ -111,6 +114,20 @@ func TestVault(t *testing.T) {
...
+	logger.Log(t, "Validating that CONSUL_CACERT has been set correctly")
+	serverPod, err := ctx.KubernetesClient(t).CoreV1().Pods(ns).Get(context.Background(), fmt.Sprintf("%s-consul-server-0", consulReleaseName), metav1.GetOptions{})
+	require.NoError(t, err)
+
+	// Environment variables are buried in the pod spec
+	containerVars := serverPod.Spec.Containers[0].Env
+	for k := range containerVars {
+		if containerVars[k].Name == "CONSUL_CACERT" {
+			require.NotEqual(t, containerVars[k].Value, "/consul/tls/ca/tls.crt")
+			require.Equal(t, containerVars[k].Value, "/vault/secrets/serverca.crt")
```
thisisnotashwin, line 127: ❤️

charts/consul/templates/client-daemonset.yaml
```diff
@@ -231,13 +231,16 @@ spec:
             {{- if .Values.global.tls.enabled }}
             - name: CONSUL_HTTP_ADDR
               value: https://localhost:8501
-            {{- if .Values.global.tls.enableAutoEncrypt }}
+              {{- if .Values.global.tls.enableAutoEncrypt }}
```
thisisnotashwin, line 234: This `if` and `end` need to get unintended. I think they were accidentally tabbed.

thisisnotashwin (approved):
This is excellent!! Great job Curt

acceptance/tests/vault/vault_test.go
ishustava: I wonder if the assertion should be more behavioral. This feels similar to what we already do in bats where we check that the rendered template has the right env vars.

What if we change it to exercise a user workflow instead? In this case, it'll be:
```
When I exec into consul server container, I should be able to run consul CLI commands and not see errors.
```

Note: We will also need to provide an ACL token to the consul CLI command if we were to do that.

charts/consul/templates/client-daemonset.yaml
```diff
@@ -234,8 +234,11 @@ spec:
             {{- if .Values.global.tls.enableAutoEncrypt }}
             - name: CONSUL_HTTP_SSL_VERIFY
               value: "false"
-            {{- else }}
+            {{- end }}
```
ishustava, line 237: I'm curious why this was switched to an `end` instead of an `else`? With auto-encrypt, we still want to skip ssl verification because the CA of the client != the CA of the server.

ishustava: Oh and since vault only works with auto-encrypt, we don't need this change for the client then! Sorry for confusing it in the task; I only realized this after looking at your PR!

ishustava (commented):
Great work, Curt!!!

I left some comments in-line.

charts/consul/templates/client-daemonset.yaml
curtbushko: No problem. So just reset client-daemonset back to what is on main?

acceptance/tests/vault/vault_test.go
curtbushko: Behavioural would probably catch a lot more as it would ensure that everything works...

charts/consul/templates/client-daemonset.yaml
ishustava: yeah, we don't need to change the client

acceptance/tests/vault/vault_test.go
```diff
@@ -65,7 +66,7 @@ func TestVault(t *testing.T) {
 		"global.secretsBackend.vault.connectCA.rootPKIPath":         "connect_root",
 		"global.secretsBackend.vault.connectCA.intermediatePKIPath": "dc1/connect_inter",
 
-		"global.acls.manageSystemACLs":       "true",
+		"global.acls.manageSystemACLs":       "false",
```
curtbushko, line 69: @ishustava When I talked to Ashwin the other day he was telling me that the ACLs were not needed because it doesn't change the TLS behaviour.

I did spend the time to look into ACLs and how to do it for my learning. So, if you want, I could still do it with ACLs and use the bootstrap token.. etc.

acceptance/tests/vault/vault_test.go
```diff
@@ -111,6 +114,16 @@ func TestVault(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, caConfig.Provider, "vault")
 
+	// Validate that consul sever is running correctly and the consul members command works
```
curtbushko, line 117: @ishustava If this would be useful somewhere else as a generic helper then I could put it somewhere?

acceptance/tests/vault/vault_test.go
```diff
@@ -111,6 +114,16 @@ func TestVault(t *testing.T) {
...
+	tokenSecret, err := ctx.KubernetesClient(t).CoreV1().Secrets(ns).Get(context.Background(), fmt.Sprintf("%s-consul-bootstrap-acl-token", consulReleaseName), metav1.GetOptions{})
+	require.NoError(t, err)
+	token := string(tokenSecret.Data["token"])
+
+	membersOutput, err := k8s.RunKubectlAndGetOutputE(t, ctx.KubectlOptions(t), "exec", fmt.Sprintf("%s-consul-server-0", consulReleaseName), "-c", "consul", "--", "sh", "-c", fmt.Sprintf("CONSUL_HTTP_TOKEN=%s consul members", token))
```
ishustava, line 122: I think we'd want to run it with the discard logger so the token is not printed to logs (this needs an import `terratestLogger "github.com/gruntwork-io/terratest/modules/logger"`):
```suggestion
	logger.Log(t, Confirming that we can run Consul commands when exec'ing into server container")
	membersOutput, err := k8s.RunKubectlAndGetOutputWithLoggerE(t, ctx.KubectlOptions(t), terratestLogger.Discard, "exec", fmt.Sprintf("%s-consul-server-0", consulReleaseName), "-c", "consul", "--", "sh", "-c", fmt.Sprintf("CONSUL_HTTP_TOKEN=%s consul members", token))
```

acceptance/tests/vault/vault_test.go
```diff
@@ -111,6 +114,16 @@ func TestVault(t *testing.T) {
...
+	logger.Log(t, "Members: ", membersOutput)
```
ishustava, line 123: Slightly nicer output:
```suggestion
	logger.Logf(t, "Members: \n%s", membersOutput)
```

acceptance/tests/vault/vault_test.go
ishustava: I think it's ok for now. If we end up using it in multiple places, then we can put it in helpers

ishustava (approved):
Looks great, Curt! I've made a couple of suggestions, but they're not blocking. Thanks for fixing this!

PR #1104, opened 2022-03-16
Don't set TTL for server certs when using Vault

ishustava:
Changes proposed in this PR:
- When using vault secrets backend, don't set TTL for server certs explicitly so that they don't expire too early. This should be controlled via the [PKI role](https://www.vaultproject.io/api-docs/secret/pki#ttl-1) rather than hard-coded in the template so that the users can control the TTL that is suitable for them.

How I've tested this PR:
unit tests

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ IMPROVEMENTS:
 * Helm
   * API Gateway: Allow controller to read Kubernetes namespaces in order to determine if route is allowed for gateway. [[GH-1092](https://github.com/hashicorp/consul-k8s/pull/1092)]
 
+BUG FIXES:
+* Helm
+  * Don't set TTL for server certificates when using Vault as the secrets backend. [[GH-1104](https://github.com/hashicorp/consul-k8s/pull/1104)]
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ IMPROVEMENTS:
 * Helm
   * API Gateway: Allow controller to read Kubernetes namespaces in order to determine if route is allowed for gateway. [[GH-1092](https://github.com/hashicorp/consul-k8s/pull/1092)]
 
+BUG FIXES:
+* Helm
+  * Don't set TTL for server certificates when using Vault as the secrets backend. [[GH-1104](https://github.com/hashicorp/consul-k8s/pull/1104)]
+
```
curtbushko, line 16: Just to clarify, the TTL isn't needed for regular/non-vault setups?

ishustava: yep!

curtbushko (approved):

jmurret (approved):
Sorry I just saw this.  Nice work!

PR #1112, opened 2022-03-23
[RelAPI Onboarding] Add release API metadata file

mdeggies:
👋  This PR adds a `.release/release-metadata.hcl` file to the repo. This contains static metadata that will be processed and sent as part of the payload in RelAPI POST requests, which will be sent when staging and production releases are triggered.  

This can be merged now, but will not have any effect until after the RelAPI launch. Similar additions are being added across all projects that publish to releases.hashicorp.com.


jeanneryan (approved):
LGTM

.release/release-metadata.hcl
```diff
@@ -0,0 +1,4 @@
+url_docker_registry_dockerhub = "https://hub.docker.com/r/hashicorp/consul-k8s"
```
ishustava, line 1: @mdeggies we use registry today for the docker images: https://hub.docker.com/r/hashicorp/consul-k8s-control-plane/tags

PR #1113, opened 2022-03-23
Enable snapshot agent configuration to be retrieved from vault

jmurret:
Changes proposed in this PR:
- `charts/consul/values.yaml` - Add `global.secretsBackend.vault.consulSnapshotAgentRole` helm value so users can preconfigure the auth role in Vault with this service account and have snapshot agent retrieve the snapshot agent config secret from vault using this service account.
- `charts/consul/templates/client-snapshot-agent-deployment.yaml`
  - configure snapshot agent with vault role annotation for new `global.secretsBackend.vault.consulSnapshotAgentRole` helm value
  - set up vault annotations so injector downloads snapshot agent config to`/vault/secrets/snapshot-agent-config.txt`
  - configure command for `consul-snapshot-agent` container to:
    - base64 decode the contents of `/vault/secrets/snapshot-agent-config.txt` into `/vault/secrets/snapshot-agent-config.json`
    - pass `-config-file=/vault/secrets/snapshot-agent-config.json` flag to snapshot agent command
  - pass `-token-sink-file=/consul/login/acl-token` to acl-init command so that it saves the login token where `CONSUL_HTTP_TOKEN` file expects it and that the `consul-snapshot-agent` container will use.
How I've tested this PR:

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ IMPROVEMENTS:
   * Upgrade Docker image Alpine version from 3.14 to 3.15. [[GH-1058](https://github.com/hashicorp/consul-k8s/pull/1058)]
 * Helm
   * API Gateway: Allow controller to read Kubernetes namespaces in order to determine if route is allowed for gateway. [[GH-1092](https://github.com/hashicorp/consul-k8s/pull/1092)]
+* Vault: Add support for Snapshot Agent to be configured in Vault. [[GH-XXXX](https://github.com/hashicorp/consul-k8s/pull/XXXX)]
 
 BUG FIXES:
 * Helm
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ IMPROVEMENTS:
   * Upgrade Docker image Alpine version from 3.14 to 3.15. [[GH-1058](https://github.com/hashicorp/consul-k8s/pull/1058)]
 * Helm
   * API Gateway: Allow controller to read Kubernetes namespaces in order to determine if route is allowed for gateway. [[GH-1092](https://github.com/hashicorp/consul-k8s/pull/1092)]
-* Vault: Enable snapshot agent configuration to be retrieved from vault. [[GH-1113](https://github.com/hashicorp/consul-k8s/pull/1113)]
+  * Vault: Enable snapshot agent configuration to be retrieved from vault. [[GH-1113](https://github.com/hashicorp/consul-k8s/pull/1113)]
 
 BUG FIXES:
 * Helm
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ IMPROVEMENTS:
   * Upgrade Docker image Alpine version from 3.14 to 3.15. [[GH-1058](https://github.com/hashicorp/consul-k8s/pull/1058)]
 * Helm
   * API Gateway: Allow controller to read Kubernetes namespaces in order to determine if route is allowed for gateway. [[GH-1092](https://github.com/hashicorp/consul-k8s/pull/1092)]
-* Vault: Add support for Snapshot Agent to be configured in Vault. [[GH-XXXX](https://github.com/hashicorp/consul-k8s/pull/XXXX)]
+* Vault: Enable snapshot agent configuration to be retrieved from vault. [[GH-1113](https://github.com/hashicorp/consul-k8s/pull/1113)]
 
 BUG FIXES:
 * Helm
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -22,6 +22,13 @@ as well as the global.name setting.
             {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
 
+{{- define "consul.vaultDecodedSecretTemplate" -}}
+ |
+            {{ "{{" }}- with secret "{{ .secretName }}" -{{ "}}" }}
+            {{ "{{" }}- {{ printf "base64Decode .Data.data.%s" .secretKey }} -{{ "}}" }}
+            {{ "{{" }}- end -{{ "}}" }}
+{{- end -}}
+
 {{- define "consul.serverTLSCATemplate" -}}
  |
             {{ "{{" }}- with secret "{{ .Values.global.tls.caCert.secretName }}" -{{ "}}" }}
--- a/charts/consul/templates/client-snapshot-agent-deployment.yaml
+++ b/charts/consul/templates/client-snapshot-agent-deployment.yaml
@@ -31,7 +31,7 @@ spec:
         "consul.hashicorp.com/connect-inject": "false"
         {{- if .Values.global.secretsBackend.vault.enabled }}
         {{- if and .Values.global.tls.enabled .Values.client.snapshotAgent.configSecret.secretName }}
-        "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulCARole }},{{ .Values.global.secretsBackend.vault.consulSnapshotAgentRole }}
+        "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulSnapshotAgentRole }}
         {{- else if and .Values.global.tls.enabled (not .Values.client.snapshotAgent.configSecret.secretName) }}
         "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulCARole }}
         {{- else if and (not .Values.global.tls.enabled) .Values.client.snapshotAgent.configSecret.secretName }}
--- a/charts/consul/templates/client-snapshot-agent-deployment.yaml
+++ b/charts/consul/templates/client-snapshot-agent-deployment.yaml
@@ -30,12 +30,10 @@ spec:
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
         {{- if .Values.global.secretsBackend.vault.enabled }}
-        {{- if and .Values.global.tls.enabled .Values.client.snapshotAgent.configSecret.secretName }}
+        {{- if .Values.client.snapshotAgent.configSecret.secretName }}
         "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulSnapshotAgentRole }}
-        {{- else if and .Values.global.tls.enabled (not .Values.client.snapshotAgent.configSecret.secretName) }}
+        {{- else if and .Values.global.tls.enabled  }}
         "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulCARole }}
-        {{- else if and (not .Values.global.tls.enabled) .Values.client.snapshotAgent.configSecret.secretName }}
-        "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulSnapshotAgentRole }}
         {{- end }}
         {{- if .Values.global.tls.enabled }}
         "vault.hashicorp.com/agent-init-first": "true"
--- a/charts/consul/templates/client-snapshot-agent-deployment.yaml
+++ b/charts/consul/templates/client-snapshot-agent-deployment.yaml
@@ -1,4 +1,5 @@
 {{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if or (and .Values.client.snapshotAgent.configSecret.secretName (not .Values.client.snapshotAgent.configSecret.secretKey)) (and (not .Values.client.snapshotAgent.configSecret.secretName) .Values.client.snapshotAgent.configSecret.secretKey) }}{{fail "client.snapshotAgent.configSecret.secretKey and secretName must both be specified." }}{{ end -}}
 {{- if .Values.client.snapshotAgent.enabled }}
 apiVersion: apps/v1
 kind: Deployment
@@ -49,6 +50,15 @@ spec:
         "vault.hashicorp.com/agent-inject-template-enterpriselicense.txt": {{ template "consul.vaultSecretTemplate" . }}
         {{- end }}
         {{- end }}
+        {{- if .Values.client.snapshotAgent.configSecret.secretName }}
+        {{- with .Values.client.snapshotAgent.configSecret }}
+        "vault.hashicorp.com/agent-inject-secret-snapshot-agent-config.txt": "{{ .secretName }}"
+        "vault.hashicorp.com/agent-inject-template-snapshot-agent-config.txt": {{ template "consul.vaultSecretTemplate" . }}
+        {{- end }} 
+        {{- end }}
+        {{- if .Values.client.snapshotAgent.configSecret.secretName }}
+        "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulSnapshotAgentRole }}
+        {{- end }}
         {{- end }}
     spec:
       {{- if .Values.client.tolerations }}
@@ -65,7 +75,7 @@ spec:
       - name: consul-data
         emptyDir:
           medium: "Memory"
-      {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
+      {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey (not .Values.global.secretsBackend.vault.enabled)) }}
       - name: snapshot-config
         secret:
           secretName: {{ .Values.client.snapshotAgent.configSecret.secretName }}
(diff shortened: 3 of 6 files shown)


charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -49,6 +50,15 @@ spec:
...
         {{- end }}
         {{- end }}
+        {{- if .Values.client.snapshotAgent.configSecret.secretName }}
+        {{- with .Values.client.snapshotAgent.configSecret }}
+        "vault.hashicorp.com/agent-inject-secret-snapshot-agent-config.txt": "{{ .secretName }}"
+        "vault.hashicorp.com/agent-inject-template-snapshot-agent-config.txt": {{ template "consul.vaultSecretTemplate" . }}
+        {{- end }} 
+        {{- end }}
+        {{- if .Values.client.snapshotAgent.configSecret.secretName }}
+        "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulSnapshotAgentRole }}
```
jmurret, line 60: role that has been preconfigured in a policy in Vault that has read access to the snapshot agent config. will be used in the request to download `snapshot-agent-config.txt` that results from the vault injector annotations above.

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -137,10 +147,17 @@ spec:
           {{- .Values.client.snapshotAgent.caCert | nindent 14 }}
           EOF
           {{- end }}
+          {{- if  .Values.global.secretsBackend.vault.enabled }}
+          cat /vault/secrets/snapshot-agent-config.txt | base64 -d > /vault/secrets/snapshot-agent-config.json
```
jmurret, line 151: consul snapshot command will look for only files with a .json extension and the contents from vault need to be base64 decoded.

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -137,10 +147,17 @@ spec:
...
+          {{- end }}
           exec /bin/consul snapshot agent \
             {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
+            {{- if  .Values.global.secretsBackend.vault.enabled }}
+            -config-file=/vault/secrets/snapshot-agent-config.json \
```
jmurret, line 156: above json transformation used here specify the configuration.

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -65,7 +75,7 @@ spec:
       - name: consul-data
         emptyDir:
           medium: "Memory"
-      {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
+      {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey (not .Values.global.secretsBackend.vault.enabled)) }}
```
jmurret, line 78: when using vault, no need for k8s secret volume

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -156,7 +173,7 @@ spec:
                 /bin/consul logout
         {{- end }}
         volumeMounts:
-        {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
+        {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey (not .Values.global.secretsBackend.vault.enabled)) }}
```
jmurret, line 176: when using vault, no need for k8s secret volume mount

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -229,6 +246,7 @@ spec:
               {{- if .Values.global.adminPartitions.enabled }}
               -partition={{ .Values.global.adminPartitions.name }} \
               {{- end }}
+              -token-sink-file=/consul/login/acl-token \
```
jmurret, line 249: consul login will save the acl token to this sink file location.
snapshot agent needs an acl token and if not passed on through the `-token` flag, it will look `CONSUL_HTTP_TOKEN_FILE`...which is mapped to this location in that container.

charts/consul/values.yaml
```diff
@@ -162,6 +162,17 @@ global:
...
+      # The Vault role for the Consul client snapshot agent.
+      # The role must be connected to the Consul client snapshot agent's service account and
+      # have a policy with read capabilities for the following secrets:
+      # - gossip encryption key defined by `client.snapshotAgent.configSecret.secretName`.
+      # To discover the service account name of the Consul client, run
+      #     ```shell-session
+      #     $ helm template --show-only templates/client-snapshot-agent- serviceaccount.yaml <release-name> charts/consul
+      #     ```
+      # and check the name of `metadata.name`.
+      consulSnapshotAgentRole: ""
```
jmurret, line 174: User has to set up a vault policy and auth role that maps the service account to the policy that has access to the snapshot agent config secret.  Configuring it here will assign it in a vault annotation in the snapshot agent deployment so that the vault injector for the snapshot agent config will be authorized.

charts/consul/templates/client-snapshot-agent-deployment.yaml
curtbushko, line 151: I don't think that `base64 -d` is platform agnostic. I am sure it is `base64 -D` on linux. Who knows what Windows is.

It might be possible to use the built in helm functions to do this. The function is something like `b64enc`. If the vault file is already a json encode, can you get as snapshot-agent-config.json and do something like:

{{ "/vault/secrets/snapshot-agent-config.json" | b64enc }}

I am not 100% sure that works.

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if or (and .Values.client.snapshotAgent.configSecret.secretName (not .Values.client.snapshotAgent.configSecret.secretKey)) (and (not .Values.client.snapshotAgent.configSecret.secretName) .Values.client.snapshotAgent.configSecret.secretKey) }}{{fail "client.snapshotAgent.configSecret.secretKey and secretName must both be specified." }}{{ end -}}
```
curtbushko, line 2: Nit. I would change the fail message to be name/key and a little more specific to:

`fail "client.snapshotAgent.configSecret.secretName and client.snapshotAgent.configSecret.secretKey must both be specified."`

charts/consul/test/unit/client-snapshot-agent-deployment.bats
```diff
@@ -39,6 +39,83 @@ load _helpers
       .
 }
 
+@test "client/SnapshotAgentDeployment: when client.snapshotAgent.configSecret.secretKey!=null and client.snapshotAgent.configSecret.secretName=null, fail" {
```
curtbushko, line 42: I love that you did these 2 bats tests the opposite of eachother!

curtbushko (commented):
Nice job! I especially like all the bats tests!  

I only had one small nit and maybe a better way to do the base64 encoding. I am not 100% sure if it can be done that way though.

charts/consul/templates/client-snapshot-agent-deployment.yaml
jmurret: great call!  🙏

curtbushko (approved):

charts/consul/templates/client-snapshot-agent-deployment.yaml
jmurret: I tried a few stabs at this and could never get this working and found a solution that achieves what we want.  We can use consul template's base64 decode in the vault agent injector annotation and have it write directly to a .json file that is decode.  So there is no need to copy from a .txt file to a .json file.  It's in my latest commit.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ IMPROVEMENTS:
   * Upgrade Docker image Alpine version from 3.14 to 3.15. [[GH-1058](https://github.com/hashicorp/consul-k8s/pull/1058)]
 * Helm
   * API Gateway: Allow controller to read Kubernetes namespaces in order to determine if route is allowed for gateway. [[GH-1092](https://github.com/hashicorp/consul-k8s/pull/1092)]
+* Vault: Enable snapshot agent configuration to be retrieved from vault. [[GH-1113](https://github.com/hashicorp/consul-k8s/pull/1113)]
```
ishustava, line 12: I think we typically group them under either `Helm`, `Control plane` or `CLI` top-level sections. So I think this change should be under `Helm`
```suggestion
  * Vault: Enable snapshot agent configuration to be retrieved from vault. [[GH-1113](https://github.com/hashicorp/consul-k8s/pull/1113)]
```

charts/consul/templates/_helpers.tpl
```diff
@@ -22,6 +22,13 @@ as well as the global.name setting.
             {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
 
+{{- define "consul.vaultDecodedSecretTemplate" -}}
+ |
+            {{ "{{" }}- with secret "{{ .secretName }}" -{{ "}}" }}
+            {{ "{{" }}- {{ printf "base64Decode .Data.data.%s" .secretKey }} -{{ "}}" }}
+            {{ "{{" }}- end -{{ "}}" }}
+{{- end -}}
```
ishustava, line 30: Why do we need to base64 decode it? Does vault kv engine not work with not-encoded config? Or does it encode it for you if similar to k8s secrets?

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if or (and .Values.client.snapshotAgent.configSecret.secretName (not .Values.client.snapshotAgent.configSecret.secretKey)) (and (not .Values.client.snapshotAgent.configSecret.secretName) .Values.client.snapshotAgent.configSecret.secretKey) }}{{fail "client.snapshotAgent.configSecret.secretKey and client.snapshotAgent.configSecret.secretName must both be specified." }}{{ end -}}
```
ishustava, line 2: I think we should move this below `if .Values.client.snapshotAgent.enabled` so that it only fails if you have snapshot agent enabled.

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -49,6 +50,15 @@ spec:
...
         {{- end }}
         {{- end }}
+        {{- if .Values.client.snapshotAgent.configSecret.secretName }}
+        {{- with .Values.client.snapshotAgent.configSecret }}
+        "vault.hashicorp.com/agent-inject-secret-snapshot-agent-config.json": "{{ .secretName }}"
+        "vault.hashicorp.com/agent-inject-template-snapshot-agent-config.json": {{ template "consul.vaultDecodedSecretTemplate" . }}
+        {{- end }} 
+        {{- end }}
+        {{- if .Values.client.snapshotAgent.configSecret.secretName }}
+        "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulSnapshotAgentRole }}
```
ishustava, line 60: should we put it inside the same `if` condition on line 53?

Also, we have a role already provided on line 36 so I think it's possible to end up with two role annotations if you have TLS enabled and also have provided snapshot config secret.

charts/consul/templates/client-snapshot-agent-deployment.yaml
ishustava: This is also the default in the `acl-init ` command but I think it's good to set this explicitly:

https://github.com/hashicorp/consul-k8s/blob/acls-refactor-base-branch/control-plane/subcommand/acl-init/command.go#L30

charts/consul/test/unit/client-snapshot-agent-deployment.bats
```diff
@@ -746,6 +823,78 @@ MIICFjCCAZsCCQCdwLtdjbzlYzAKBggqhkjOPQQDAjB0MQswCQYDVQQGEwJDQTEL' \
...
+  [ "${actual}" = "path/to/secret" ]
+
+  actual=$(echo $object |
+      yq -r '.annotations["vault.hashicorp.com/agent-inject-template-snapshot-agent-config.json"]' | tee /dev/stderr)
+  local expected=$'{{- with secret \"path/to/secret\" -}}\n{{- base64Decode .Data.data.config -}}\n{{- end -}}'
+  [ "${actual}" = "${expected}" ]
+
+  actual=$(echo $object | jq -r '.annotations["vault.hashicorp.com/role"]' | tee /dev/stderr)
+  [ "${actual}" = "bar" ]
+}
```
ishustava, line 851: I think we need a test case for when tls is enabled and also snapshot config secret is provided

charts/consul/values.yaml
```diff
@@ -162,6 +162,17 @@ global:
       # and check the name of `metadata.name`.
       consulClientRole: ""
 
+      # The Vault role for the Consul client snapshot agent.
+      # The role must be connected to the Consul client snapshot agent's service account and
+      # have a policy with read capabilities for the following secrets:
+      # - gossip encryption key defined by `client.snapshotAgent.configSecret.secretName`.
+      # To discover the service account name of the Consul client, run
+      #     ```shell-session
+      #     $ helm template --show-only templates/client-snapshot-agent- serviceaccount.yaml <release-name> charts/consul
```
ishustava, line 171:
```suggestion
      #     $ helm template --show-only templates/client-snapshot-agent-serviceaccount.yaml --set client.snapshotAgent.enabled=true <release-name> hashicorp/consul
```

charts/consul/values.yaml
```diff
@@ -162,6 +162,17 @@ global:
       # and check the name of `metadata.name`.
       consulClientRole: ""
 
+      # The Vault role for the Consul client snapshot agent.
```
ishustava, line 165:
```suggestion
      # [Enterprise Only] The Vault role for the Consul client snapshot agent.
```

charts/consul/values.yaml
```diff
@@ -162,6 +162,17 @@ global:
...
+      # The role must be connected to the Consul client snapshot agent's service account and
+      # have a policy with read capabilities for the following secrets:
+      # - gossip encryption key defined by `client.snapshotAgent.configSecret.secretName`.
```
ishustava, line 168:
```suggestion
      # have a policy with read capabilities for the snapshot agent config defined by `client.snapshotAgent.configSecret.secretName`.
```

ishustava (commented):
Looking good, John! Sorry for the delay!

I left some comments in-line. The main blocker is about making sure we only provide one vault role annotation. Let me know if I'm missing something though!

charts/consul/templates/_helpers.tpl
jmurret: Great question!  Let me go back and check that I did not encode as part of the test set up.

jmurret: I totally was.  I used json.Marshall() and then saving that to Vault instead of wrapping that result in string() and then saving it.  I have a PR with the acceptance test and made the modification there ...since I can test it.  This is the commit: https://github.com/hashicorp/consul-k8s/pull/1125/commits/634164a3b2db74148249293ceb78238187ab6a13

charts/consul/templates/client-snapshot-agent-deployment.yaml
```diff
@@ -29,10 +30,16 @@ spec:
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
         {{- if .Values.global.secretsBackend.vault.enabled }}
+        {{- if and .Values.global.tls.enabled .Values.client.snapshotAgent.configSecret.secretName }}
+        "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulCARole }},{{ .Values.global.secretsBackend.vault.consulSnapshotAgentRole }}
```
ishustava, line 34: I think vault can only have one role, so in this case it would be the `consulSnapshotAgentRole`. 

Note that the user would not need to allow access to the CA for the `consulSnapshotAgentRole` if they have configured it like we recommend in our docs allowing all service accounts (`*`) to bind to it : https://www.consul.io/docs/k8s/installation/vault/server-tls. The CA role is really only needed when you need access to the CA only because vault agent will not start without a role provided. Hope it makes sense!

ishustava (approved):
Looks great!!

PR #1115, opened 2022-03-23
Fix "consul-k8s install fails: now can't re-install or upgrade"

### LINKED ISSUE #1005: consul-k8s install fails: now can't re-install or upgrade
### Community Note

---

### Overview of the Issue

When `consul-k8s install` fails there's no way to force a helm upgrade

### Reproduction Steps

1. `consul-k8s install` with something that will error. In my example I had a previous `servicedefaults` config in a previous installation that when I uninstalled consul couldn't get deleted because the controller was gone. On a re-install the queued-up crd delete succeeded which then cause the install to fail:
```
 --> beginning wait for 45 resources with timeout of 10m0s
 ! customresourcedefinitions.apiextensions.k8s.io "servicedefaults.consul.hashicorp.com" not found
```
1. Now I'm in a situation where I have a failed install:

```
helm list
NAME  	NAMESPACE	REVISION	UPDATED                             	STATUS	CHART        	APP VERSION
consul	default  	1       	2022-01-27 13:03:41.014162 -0800 PST	failed	consul-0.40.0	1.11.2
```

But I can't re-install:

```
consul-k8s install -config-file=consul/config.yaml -namespace=default

==> Pre-Install Checks
 ! existing Consul installation found (name=consul, namespace=default) - run consul-k8s uninstall if you wish to re-install
```

Nor upgrade:

```
consul-k8s upgrade -config-file=consul/config.yaml

==> Pre-Upgrade Checks
 ✓ Existing installation found to be upgraded.
    Name: consul
    Namespace: default
 ✓ Loaded charts
 ! release: not found
```

But `helm upgrade` works:

```
 helm upgrade consul hashicorp/consul -f consul/config.yaml
Release "consul" has been upgraded. Happy Helming!
NAME: consul
```

t-eckert:
Fixes #1005 

Changes proposed in this PR:
- Testing
  - Extend `HelmCluster` and `CLICluster` to accept arguments on `Create`, `Update`, and `Delete` actions.
  - Allow tests to set the namespace used for CLI acceptance tests.
  - Add a scenario test to exercise upgrading Consul after a failed install.
  - Add a scenario test to exercise upgrading Consul installed in the `default` namespace.
- Bug fix
  - Fix an issue in the upgrade case where the cluster was being found by namespace instead of its name. 

How I've tested this PR:

Both using scenario tests and by debugging the upgrade command when an install was present in the `default` namespace.

The scenario tests may be run from the `acceptance` directory with

``` bash
go test -timeout 15m github.com/hashicorp/consul-k8s/acceptance/tests/cli -v
```

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


jmurret:
Just starting to look through this.  There are a lot of places where you are looking for args and if they don't exist you use default args.  Would it be easier or more correct to merge the provided args onto the default args and use that merged map?

t-eckert:
> Just starting to look through this. There are a lot of places where you are looking for args and if they don't exist you use default args. Would it be easier or more correct to merge the provided args onto the default args and use that merged map? (I am not saying it is easier or more correct, but just that with default values and supplied values, that's the common pattern I see.)

Interesting question. I can definitely look at that. I can give that a shot with an `args map[string]string` being passed in and get your opinion on that.

jmurret:
ok, i said map, but now that you mention it and i look again you are setting the extra args for each function and its an array.  I don't think you need to bother changing that structure.  What I was getting at is you may want to take default args and append the passed in args to get the working set of args and that would represent better how default and user specified options are merged together.
You can do an append and if a passed in arg was already in the default set, it would result in two of the same arg.  Maybe that does not cause issues..but you could also quickly check if it exists in the default set of args and then skip appending it.
anyway, i would not spend a lot of time on this.  It was just more of an inquiry or a nit than anything.  Since this is infrastructure code for tests, if it works, we can also just move ahead and fix anything that comes up in the future.  I think you could try out the contains/append with the default and passed in args pretty quickly tho, so if you do anything, you may time box that and then call it good.

t-eckert:
Hm, if I do that, I will actually use a map and then flatten it because some flags take a value and others do not. I can map `string` to `*string` and implement this kind of merging without having to hard-code logic of which args take values and which do not.

t-eckert:
So, when I did that I remembered why I went with `...string` as an argument. Otherwise, I have to change the call to these methods in a few dozen or so places to pass in an empty map.

acceptance/framework/consul/cli_cluster.go
```diff
@@ -305,3 +303,26 @@ func (c *CLICluster) runCLI(args []string) ([]byte, error) {
...
+		if !contains(args, k) {
+			args = append(args, k)
+			if v != "" {
+				args = append(args, v)
+			}
+		}
+	}
+
+	return args
+}
```
t-eckert, line 328: @jmurret what do you think about this?

jmurret: yea, i think that's about right.  Do you have an example of defaults?  I am just curious what these look like because you are appending the key and the value from default args....which if the k/v pair is something like enable/true, you will append "enable" and "true".  But if it is enable/"", then you only append "enable". Is that desired?

acceptance/tests/cli/scenario_test.go
```diff
@@ -0,0 +1,38 @@
+package cli
```
jmurret, line 1: [nit] would naming this file `upgrade_test.go` be more descriptive?

jmurret (approved):
🎉 🙏 awesome find and fix!  Minor nit on file name and wont' block on the ongoing conversation about merging args since everything works and don't want to draw this out too much.

acceptance/framework/consul/cli_cluster.go
jmurret: 👍

acceptance/tests/cli/scenario_test.go
t-eckert: Yes, I thought the same thing. Now that you mention it too, I think it's the right thing.

acceptance/tests/cli/upgrade_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"testing"
+
+	"github.com/hashicorp/consul-k8s/acceptance/framework/consul"
+)
+
+// TestUpgradeAfterFailedInstall exercises the upgrade command after a failed
+// install. This scenario tests this issue: https://github.com/hashicorp/consul-k8s/issues/1005
+func TestUpgradeAfterFailedInstall(t *testing.T) {
```
ishustava, line 11: Generally, acceptance tests should be added for happy case scenarios because they're kind of heavy-handed. These scenarios feel more appropriate as unit tests but I know that we don't have unit tests set up. I've tried mocking it out on this branch, but it still needs more work (and we've synced up on it over zoom): https://github.com/hashicorp/consul-k8s/compare/ihustava/wip-cli-unit-test. 

This is not a blocker but if we do decide to merge it with the acceptance tests, then I think we should at least add a TODO here.

ishustava (commented):
Nice fix, Thomas!

I'm leaving some comments and questions in-line.

acceptance/framework/consul/cli_cluster.go
```diff
@@ -26,10 +26,14 @@ import (
 )
 
 const (
-	consulNS       = "consul"
```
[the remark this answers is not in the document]
t-eckert: The reason I removed it is to steer future engineers away from referencing it and not the `cfg.KubeNamespace`. I could see a hidden bug appearing in the future where the test sets up `cfg.KubeNamespace` but a part of the cluster code is ignoring it and using `consulNS`. What do you think in that case?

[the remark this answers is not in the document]
ishustava: Could we set `cfg.KubeNamespace` to `consulNS` const? This const is scoped to the `cli_cluster` so you shouldn't be able to reference it from the tests.

acceptance/framework/consul/cli_cluster.go
```diff
@@ -118,29 +127,24 @@ func (c *CLICluster) Create(t *testing.T) {
...
-	args = append(args, "-timeout", "15m")
-	args = append(args, "-auto-approve")
+	args = append(args, "-namespace", c.namespace)
 
 	out, err := c.runCLI(args)
 	if err != nil {
 		c.logger.Logf(t, "error running command `consul-k8s %s`: %s", strings.Join(args, " "), err.Error())
 		c.logger.Logf(t, "command stdout: %s", string(out))
 	}
-	require.NoError(t, err)
```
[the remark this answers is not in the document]
ishustava: See my comment below where I realized why you were doing it 😄 

In addition to that, there are other tests for CLI that rely on this error checking. If we remove it, there's no way to make sure that this function succeeded because it does not return an error. This is another sign that acceptance tests framework is not meant for testing failing installations. So in your case, you'd have to either add another method that returns an error to the interface and implementations (e.g. `CreateE(t *testing.T) error`) or not test it in acceptance tests like I mentioned in another comment.

acceptance/framework/consul/cli_cluster.go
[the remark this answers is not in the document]
t-eckert: Right, but I was hoping people wouldn't modify the code in `cli_cluster` to reference the namespace not set by the user.

acceptance/tests/cli/upgrade_test.go
t-eckert: Definitely, I will add a TODO here and circle back in the future to remove these tests and cover the behavior in unit tests in the CLI.

t-eckert: I'm actually going to remove these tests now and handle unit testing in this PR.

acceptance/framework/consul/cli_cluster.go
```diff
@@ -108,7 +117,7 @@ func NewCLICluster(
 
 // Create uses the `consul-k8s install` command to create a Consul cluster. The command itself will fail if there are
 // prior installations of Consul in the cluster so it is sufficient to run the install command without a pre-check.
-func (c *CLICluster) Create(t *testing.T) {
+func (c *CLICluster) Create(t *testing.T, args ...string) {
```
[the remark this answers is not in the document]
t-eckert: Given that I am removing the acceptance tests. Should I remove this change too?

acceptance/framework/consul/cli_cluster.go
```diff
@@ -305,3 +303,26 @@ func (c *CLICluster) runCLI(args []string) ([]byte, error) {
...
+// mergeArgs merges the given args with the default args.
+func mergeArgs(defaults map[string]string, args ...string) []string {
+	contains := func(s []string, e string) bool {
+		for _, a := range s {
+			if a == e {
+				return true
+			}
+		}
+		return false
+	}
```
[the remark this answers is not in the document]
t-eckert: Yeah, I can change that. I will wait on your comment as to whether or not I should include the `args` being passed in anyway. If so, I'll make this fix. If we ditch `args`, I'll just cut all this.

cli/helm/chart_test.go
```diff
@@ -31,6 +32,20 @@ func TestLoadChart(t *testing.T) {
...
+	logger := mock.CreateLogger(t)
+
+	expected := map[string]interface{}{}
+
+	actual, err := FetchChartValues(namespace, name, settings, logger)
+	require.NoError(t, err)
+
+	require.Equal(t, expected, actual)
+}
+
```
t-eckert, line 48: @ishustava, this test provides coverage for the fundamental bug being fixed. We definitely need a better foundation for testing behavior of installs, etc. but once this test is finished, I think it will be enough to cover what is necessary.

acceptance/framework/consul/cli_cluster.go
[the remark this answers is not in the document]
t-eckert: I'm going to remove all acceptance test changes.

ishustava (approved):
Looks good!

PR #1118, opened 2022-03-23
Enable ingress gateways to use ACL Auth Method

jmurret:
Changes proposed in this PR:
- change ingress gateway deployment to pass `acl-auth-method` so it can properly call Consul login.
- add prestop for logout to ingress gateway container.

How I've tested this PR:
- unit tests
- acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added


charts/consul/templates/ingress-gateways-deployment.yaml
```diff
@@ -26,15 +26,15 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ template "consul.fullname" $root }}-{{ .name }}
+  name: {{ template "consul.fullname" $root }}-{{ .name }}-ingress-gateway
```
jmurret, line 29: throughout this PR (and the terminating gateway one), we add `-terminating-gateway` to the end of all resources because unlike other components, users can launch multiple terminating gateways and multiple ingress gateways that they can name themselves.
One, we want to be able to identify the component type if the name it `foo`.  Two, if they have an ingress gateway named `foo` and a terminating gateway named `foo`, this suffix will prevent name collisions / confusion.

charts/consul/templates/ingress-gateways-deployment.yaml
```diff
@@ -171,9 +171,14 @@ spec:
...
                 consul-k8s-control-plane acl-init \
-                  -secret-name="{{ template "consul.fullname" $root }}-{{ .name }}-ingress-gateway-acl-token" \
-                  -k8s-namespace={{ $root.Release.Namespace }} \
-                  -token-sink-file=/consul/service/acl-token
+                  -component-name=ingress-gateway/{{ template "consul.fullname" $root }}-{{ .name }} \
+                  -acl-auth-method={{ template "consul.fullname" $root }}-k8s-component-auth-method \
+                  {{- if $root.Values.global.adminPartitions.enabled }}
+                  -partition={{ $root.Values.global.adminPartitions.name }} \
+                  {{- end }}
+                  -token-sink-file=/consul/service/acl-token \
```
jmurret, line 179: save the acl token to this sink file location.  Then we will use it later in setting the `CONSUL_HTTP_TOKEN_FILE` on ingress gateway container so that it can run with this ACL token.

charts/consul/templates/ingress-gateways-deployment.yaml
lkysow: Not sure about this. We'll then have everything called `consul-ingress-gateway-ingress-gateway`

lkysow: I don't think most folks use multiple igw's so not sure we should optimize for that. What's the impact if they accidentally set both names to `foo`? Does it fail the helm install/upgrade?

lkysow: By default we have:

```yaml
ingressGateways:
  gateways:
  - name: ingress-gateway
```

so all you need to do in your values is:

```
ingressGateways:
  enabled: true
```

Now this would cause you to have `consul-ingress-gateway-ingress-gateway` everywhere which looks pretty bad imo.

charts/consul/templates/ingress-gateways-deployment.yaml
```diff
@@ -399,6 +404,9 @@ spec:
                         -partition={{ $root.Values.global.adminPartitions.name }} \
                         {{- end }}
                         -id="${POD_NAME}"
+                  {{- if $root.Values.global.acls.manageSystemACLs }}
+                  - "/consul-bin/consul logout"
+                  {{- end}}
```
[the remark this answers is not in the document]
jmurret: Also, whatever the answer is, this will be the case with all components that are now using consul login to get an acl token.  They are all defining this preStop logout.

charts/consul/templates/ingress-gateways-deployment.yaml
jmurret: Looks like helm install fails before anything...as you inquired about.  Getting errors related to duplicate service accounts.  Getting `serviceaccounts "test-nsy8zi-consul-foo" already exists`.

@lkysow what do you think is the best option? I'm going to look into a helm check at the moment.  I agree with you that this PR is over optimizing for a rare case.  The check is probably going to involve checking:
- look for duplicate terminating gateways 
- look for duplicate ingress gateways
- look for duplicates across all ingress and terminating gateway names.

I'm guessing each of these checks will involve nested loops. So, for us the code will be 🙈 , but I think it will be worth it for a better user UX.  If there are other options or ideas, please let me know what you think.

lkysow: Yeah I think the helm check is a good idea and I apologize in advance for what is likely to be quite un-fun.

Can definitely be a separate PR at least.

jmurret: Cool.  I address it in this PR: https://github.com/hashicorp/consul-k8s/pull/1120

ishustava: That sounds good! Feels like a much better solution than repeated names

charts/consul/templates/ingress-gateways-deployment.yaml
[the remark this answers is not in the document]
ishustava: Yeah we decided that we won't cleanup tokens in the case of a force-shutdown. The other option was to run a controller in k8s but it felt a bit heavy-handed for this and we thought we'll wait for user feedback before we implement it.

ishustava (approved):
Looks great!

Issue #657, opened 2021-01-22
helm: Support a pre-set bootstrap ACL token


lkysow:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Description 

We currently support the following use case with Consul K8s: 

If I have a Consul cluster, I can manually bootstrap ACLs and extract the boostrap token into a Kubernetes secret (i.e. create a Kubernetes secret manually after `consul acl bootstrap`). 

However, the following use case is what this feature request is tracking: 

I don't have a Consul cluster yet and I want to start one with bootstrap token defined via a String value. Could we support setting the master token ahead of time prior to the installation? For example if `global.acls.bootstrapToken` was set to a secret we would use the value in the secret.

david-yu:
Closing as this is partially addressed via https://github.com/hashicorp/consul-k8s/pull/1061, where Vault is used as the secrets backend to specify a ores-set bootstrap ACL token. This support was built for Vault since setting a pre-set bootstrap token via String is what works well with a KV secrets engine like Vault. If a more generic solution is needed please re-open a new issue.

PR #1125, opened 2022-03-28
Ability to set initial_management token when using k8s secret store.  Snapshot agent acceptance tests

jmurret:
Changes proposed in this PR:
- Snapshot agent acceptance tests
- Ability to set `initial_management` token when using k8s secret store.  
  - We currently support this with vault.
  - We had [this issue](https://github.com/hashicorp/consul-k8s/issues/657) that we closed with supporting it in Vault but this would fully support it.
  - This change was required to get the snapshot acceptance test working.

How I've tested this PR:
- Unit tests
- Acceptance tests

- 👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added


CHANGELOG.md
```diff
@@ -9,7 +9,9 @@ IMPROVEMENTS:
   * Upgrade Docker image Alpine version from 3.14 to 3.15. [[GH-1058](https://github.com/hashicorp/consul-k8s/pull/1058)]
 * Helm
   * API Gateway: Allow controller to read Kubernetes namespaces in order to determine if route is allowed for gateway. [[GH-1092](https://github.com/hashicorp/consul-k8s/pull/1092)]
-* Vault: Enable snapshot agent configuration to be retrieved from vault. [[GH-1113](https://github.com/hashicorp/consul-k8s/pull/1113)]
+  * Support a pre-set bootstrap ACL token. [[GH-1125](https://github.com/hashicorp/consul-k8s/pull/1125)]
```
t-eckert, line 12: Worthy of the Oscar for nittiest nit, but I kind of like `pre-configured`.

```suggestion
  * Support a pre-configured bootstrap ACL token. [[GH-1125](https://github.com/hashicorp/consul-k8s/pull/1125)]
```

acceptance/framework/consul/helm_cluster.go
```diff
@@ -470,3 +452,25 @@ func defaultValues() map[string]string {
 	}
 	return values
 }
+
+func CreateOrUpdateK8sSecret(t *testing.T, client kubernetes.Interface, cfg *config.TestConfig, namespace string, secretName string, secretKey string, secret string) {
```
t-eckert, line 456: In Go, you can omit repeated types in function signatures which can save space on these longer lines:

```suggestion
func CreateOrUpdateK8sSecret(t *testing.T, client kubernetes.Interface, cfg *config.TestConfig, namespace, secretName, secretKey, secret string) {
```

acceptance/framework/consul/helm_cluster.go
```diff
@@ -470,3 +452,25 @@ func defaultValues() map[string]string {
...
+	_, err := client.CoreV1().Secrets(namespace).Get(context.Background(), secretName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		_, err := client.CoreV1().Secrets(namespace).Create(context.Background(), &corev1.Secret{
```
t-eckert, line 459: Is there value in passing in `context` and using that here?

jmurret: Great question!

I don't think at the moment, personally.  For me, dependency injection is valuable when:
- you have multiple types of an interface and you want to exchange them (we don't here)
- you want to pass in a mock for testing purposes (a big topic that I'll hint at below)

Code can always be refactored so if these two needs change, I would prefer to inject it then.

We try to push for testing workflows with little mocking and stubbing and so DI becomes less of a concern.  Some people prefer to write all of their code this way and at one point I did, but I found that unless one of the two above things were true (with the first item being more important) creating DI patterns provided less value...given that I could easily refactor when needed.

When we also look at this file and context is not injected anywhere, would I just change only this function? Would I modify all the other references to context in this file and possibly their callers? If I am injecting this are there other things I should then inject?  I think it just muddies up the code change and the conversation and I am brought back to the `why?` and the two points I mentioned above.

Thoughts?

t-eckert: Excellent reasoning! I totally agree. I just saw the opportunity and wanted to ask if you had considered it. I think your argument totally makes sense.

charts/consul/test/unit/server-statefulset.bats
```diff
@@ -1228,6 +1228,68 @@ load _helpers
...
+
+@test "server/StatefulSet: when global.acls.bootstrapToken.secretKey!=null and global.acls.bootstrapToken.secretName=null, fail" {
+    cd `chart_dir`
+    run helm template \
+        -s templates/server-statefulset.yaml \
+        --set 'global.acls.bootstrapToken.secretName=' \
+        --set 'global.acls.bootstrapToken.secretKey=enterpriselicense' \
+        .
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "both global.acls.bootstrapToken.secretKey and global.acls.bootstrapToken.secretName must be set if one of them is provided." ]]
```
t-eckert, line 1242: I hadn't encountered the `=~` in BATS tests before. I assume it's some kind of approximately equal sign. However, it's a little difficult to Google for. Is my assumption correct?

t-eckert (approved):

acceptance/framework/consul/helm_cluster.go
```diff
@@ -470,3 +452,25 @@ func defaultValues() map[string]string {
 	}
 	return values
 }
+
+func CreateOrUpdateK8sSecret(t *testing.T, client kubernetes.Interface, cfg *config.TestConfig, namespace, secretName, secretKey, secret string) {
```
ndhanushkodi, line 456:
```suggestion
func CreateOrCheckK8sSecret(t *testing.T, client kubernetes.Interface, cfg *config.TestConfig, namespace, secretName, secretKey, secret string) {
```
I think its not updating a secret, just checking if it exists?

acceptance/framework/vault/helpers.go
```diff
@@ -116,9 +121,24 @@ func configureEnterpriseLicenseVaultSecret(t *testing.T, vaultClient *vapi.Clien
 	require.NoError(t, err)
 }
 
-// configureKubernetesAuthRole configures a role for the component for the Kubernetes auth method
+// ConfigureEnterpriseLicenseVaultSecret stores it in vault as a secret and configures a policy to access it.
```
ndhanushkodi, line 124:
```suggestion
// ConfigureSnapshotAgentSecret stores it in vault as a secret and configures a policy to access it.
```

acceptance/framework/vault/helpers.go
```diff
@@ -116,9 +121,24 @@ func configureEnterpriseLicenseVaultSecret(t *testing.T, vaultClient *vapi.Clien
...
+		},
+	}
+	_, err := vaultClient.Logical().Write("consul/data/secret/snapshot-agent-config", params)
+	require.NoError(t, err)
+
+	err = vaultClient.Sys().PutPolicy("snapshot-agent-config", snapshotAgentPolicy)
+	require.NoError(t, err)
+}
+
+// ConfigureKubernetesAuthRole configures a role for the component for the Kubernetes auth method
```
ndhanushkodi, line 139: for any component? Just wasn't sure how to read the comment! Maybe:
```suggestion
// ConfigureKubernetesAuthRole configures a role for a component for the Kubernetes auth method
```

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
```diff
@@ -0,0 +1,138 @@
...
+	vaultReleaseName := helpers.RandomName()
+
+	vaultCluster := vault.NewVaultCluster(t, ctx, cfg, vaultReleaseName, nil)
+	vaultCluster.Create(t, ctx)
+	// Vault is now installed in the cluster.
+
+	// Now fetch the Vault client so we can create the policies and secrets.
+	vaultClient := vaultCluster.VaultClient(t)
+
+	// gossipKey := configureGossipVaultSecret(t, vaultClient)
```
ndhanushkodi, line 42:
```suggestion
```

acceptance/framework/vault/helpers.go
ndhanushkodi: can we add to the comment that this adds a role to vault? When reading the test I was trying to recall what this function does and I think adding that to the comment could help :)

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	bootstrapToken := vault.ConfigureACLTokenVaultSecret(t, vaultClient, "bootstrap")
+
+	vault.ConfigureSnapshotAgentSecret(t, vaultClient, cfg, generateSnapshotAgentConfig(t, bootstrapToken))
+
+	serverPolicies := "gossip,connect-ca-dc1,server-cert-dc1,bootstrap-token"
+	if cfg.EnableEnterprise {
+		serverPolicies += ",license"
+	}
+	vault.ConfigureKubernetesAuthRole(t, vaultClient, consulReleaseName, ns, "kubernetes", "server", serverPolicies)
```
ndhanushkodi, line 57: These roles are for these components to be able to get their secrets out of vault right?

ndhanushkodi (approved):
looks good!! just asking some questions for my understanding and a few clarifications!

PR #1127, opened 2022-03-30
refactor: move server config from command flags to configmap

ishustava:
Changes proposed in this PR:
Move as much of the server config into config map as possible. Things that could not be moved into config maps are flags that contain env variables and that are being scripted inside the command (e.g.  DNS `recursors`). 

The motivation for this change is to make the configuration more readable and consolidate it in one place as much as possible. For historical context, we were not doing that from the beginning because when you update the config map, the server  stateful set would not be automatically updated and so we were using command flags instead to ensure that servers will be restarted. However, this has changed with https://github.com/hashicorp/consul-helm/pull/550, but we never went back and updated it.

This change could be easier to read via commits:
- [refactor: move server TLS config to configmap](https://github.com/hashicorp/consul-k8s/pull/1127/commits/266a4c314b506dbdb931e46424f56709fecc4e53)
- [refactor: consolidate all ui config into server configmap](https://github.com/hashicorp/consul-k8s/pull/1127/commits/74be34879d049eb77aaba40bf6d7106dbd1397c3)
- [Use single DNS name for server's retry-join](https://github.com/hashicorp/consul-k8s/pull/1127/commits/51240a2ce9279b2e1e17701b6ae04c17ecf1ef34)
- [refactor: move general server config from cmd flags to configmap](https://github.com/hashicorp/consul-k8s/pull/1127/commits/5be3b40f8809ab21cbcc84b331f50fcabfbf23a0)
- [refactor: move recursors to server configmap](https://github.com/hashicorp/consul-k8s/pull/1127/commits/70b78d2a9560aba8831a5068fe91a5c5d904078c)
- [refactor: move server connect wan fed config to configmap](https://github.com/hashicorp/consul-k8s/pull/1127/commits/b6d9bab681566f234f4bac6e8cbf72174210ae4b)
- [refactor: move server telemetry config to configmap](https://github.com/hashicorp/consul-k8s/pull/1127/commits/69995ea0ea6706717507093306fd4ec46caf0ca3)

Note: the client configmap and flags will be potentially updated in a future PR

How I've tested this PR:
- unit & acceptance tests
- some manual testing to ensure that servers get rolled when configmap changes

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/server-statefulset.yaml
```diff
@@ -302,10 +302,7 @@ spec:
                 -hcl="acl { tokens { agent = \"${ACL_REPLICATION_TOKEN}\", replication = \"${ACL_REPLICATION_TOKEN}\" } }" \
                 {{- end }}
                 {{- end }}
-                {{- $serverSerfLANPort  := .Values.server.ports.serflan.port -}}
-                {{- range $index := until (.Values.server.replicas | int) }}
-                -retry-join="${CONSUL_FULLNAME}-server-{{ $index }}.${CONSUL_FULLNAME}-server.${NAMESPACE}.svc:{{ $serverSerfLANPort }}" \
-                {{- end }}
+                -retry-join="${CONSUL_FULLNAME}-server.${NAMESPACE}.svc:{{ .Values.server.ports.serflan.port }}" \
```
jmurret: 👏

jmurret (approved):
Nice!  What is the glue that binds these config maps to where consul reads them?  Does the config-checksum annotation lay these out similar to having .json files in the -config-dir?

ishustava:
>What is the glue that binds these config maps to where consul reads them? 

Configmap is mounted as a volume to pods (similar to secrets):

https://github.com/hashicorp/consul-k8s/blob/a713f85e35ebef641254e521f49e1cd6f6a12be6/charts/consul/templates/server-statefulset.yaml#L130-L132

https://github.com/hashicorp/consul-k8s/blob/a713f85e35ebef641254e521f49e1cd6f6a12be6/charts/consul/templates/server-statefulset.yaml#L363-L364

All files from the configmap will be in that `/consul/config` directory we mount to the server pod. So by default it will be:

```
/consul/consul-config:
- server.json
- central-config.json
- ui-config.json
etc
```

Then we pass the entire config dir to consul via `-config-dir` flag:

https://github.com/hashicorp/consul-k8s/blob/a713f85e35ebef641254e521f49e1cd6f6a12be6/charts/consul/templates/server-statefulset.yaml#L305

Consul will consume all files in alphabetical order and merge them all into one configuration.

>Does the config-checksum annotation lay these out similar to having .json files in the -config-dir?

The config-checksum is there to detect that update server statefulset when configmap changes. For example, let's say you had an installation without TLS and now you want to enable TLS. 

Without this annotation, if you run `helm upgrade consul --set global.tls.enabled=true hashicorp/consul`, it will just update the configmap, but the user will have to upgrade the servers themselves to apply the configmap changes. 

With an annotation, when run helm upgrade, the checksum of the configmap will change and so the helm upgrade will need to update the checksum annotation on your server statefulset pods, which will trigger a restart of the pods.

jmurret:
@ishustava Thank you.  This is cool.

ndhanushkodi (approved):
This looks great! The commit by commit readability made it so nice to review!

PR #1132, opened 2022-03-31
[CRT] Build docker images, rpms and debs

curtbushko:
Changes proposed in this PR:
- Added a multi-target Dockerfile for building all of our images
- Updated makefile to create control-plane dev image with Dockerfile
- Split build-and-test.yml into build.yml and test.yml to support CRT (CRT expects a build.yml file)
- build.yml
   - Build all of the architectures that we support for both cli and control-plane
   - Builld docker images for control-plane (and test that they work)
   - Build the rpms and debs for the cli

How I've tested this PR:
- CI pipleines
- The images are tested to see if control-plane displays the correct version
- Validated that the cli works

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,6 +89,8 @@ jobs:
           - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "windows", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
           - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "amd64", component: "cli", pkg_name: "consul-k8s" }
           - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "arm64", component: "cli", pkg_name: "consul-k8s" }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "arm64", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
       fail-fast: true
 
     name: Go ${{ matrix.go }} ${{ matrix.goos }} ${{ matrix.goarch }} ${{ matrix.component }} build    
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,6 +87,8 @@ jobs:
           - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "solaris", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane" } 
           - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "windows", goarch: "386", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
           - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "windows", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "amd64", component: "cli", pkg_name: "consul-k8s" }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
       fail-fast: true
 
     name: Go ${{ matrix.go }} ${{ matrix.goos }} ${{ matrix.goarch }} ${{ matrix.component }} build    
@@ -116,7 +118,7 @@ jobs:
           path: ${{ matrix.component}}/out/${{ matrix.pkg_name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
             
       - name: Package rpm and deb files 
-        if: ${{ matrix.goos == 'linux' && matrix.component == 'cli' }}
+        if: ${{ matrix.goos == 'linux' && matrix.component == 'cli' && matrix.goarch == 'amd64'}}
         uses: hashicorp/actions-packaging-linux@v1
         with:
           name: consul-k8s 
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,259 @@
...
+          GCO_ENABLED=0 go build -o dist/${{ matrix.pkg_name }} .
+          zip -r -j out/${{ matrix.pkg_name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip dist/
+
+      - name: Upload built binaries 
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.pkg_name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+          path: ${{ matrix.component}}/out/${{ matrix.pkg_name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+            
+      - name: Package rpm and deb files
```
curtbushko, line 118: Only build rpms and debs for the cli.

jmurret (approved):
Nice work! 🎉

.github/workflows/build.yml
```diff
@@ -0,0 +1,259 @@
...
+          path: out/${{ env.RPM_PACKAGE }}
+
+      - name: Upload debian packages 
+        uses: actions/upload-artifact@v2
+        if: ${{ matrix.goos == 'linux' && matrix.component == 'cli' }}
+        with:
+          name: ${{ env.DEB_PACKAGE }}
+          path: out/${{ env.DEB_PACKAGE }}
+
+  build-darwin:
```
ishustava, line 153: could it be part of the `build` job?

control-plane/Dockerfile
```diff
@@ -0,0 +1,136 @@
...
+COPY dist/${TARGETOS}/${TARGETARCH}/${BIN_NAME} /bin/
+
+USER 100
+CMD /bin/${BIN_NAME}
+
+# alternate release image, just for the sake of example. In this case we're using
+# debian as the base image just to make the image different from the default alpine one.
+#
+# The use cases for alternate images are things like defining an additional UBI compatible
+# image, or an image with a different function than the main image, e.g. Waypoint's ODR images.
```
ishustava, line 77: should we change this comment?

control-plane/Dockerfile
```diff
@@ -0,0 +1,136 @@
...
+# TARGETOS and TARGETARCH are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+# This is the location of the releases.
+ENV HASHICORP_RELEASES=https://releases.hashicorp.com
+
+# Copy license for Red Hat certification.
+COPY LICENSE.md /licenses/mozilla.txt
+
+RUN microdnf install -y shadow-utils
```
ishustava, line 117: Why do we need this? I don't think it was in our current UBI image

control-plane/Dockerfile
```diff
@@ -0,0 +1,136 @@
...
+# OpenShift.
+#
+# This Dockerfile creates a production release image for the project. This
+# downloads the release from releases.hashicorp.com and therefore requires that
+# the release is published before building the Docker image.
+#
+# We don't rebuild the software because we want the exact checksums and
+# binary signatures to match the software and our builds aren't fully
+# reproducible currently.
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5 as release-alternate
```
ishustava, line 89: Could we call it `ubi`?
```suggestion
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5 as ubi
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,259 @@
...
+            echo "Test PASSED"
+          version: ${{ env.version }}
+          target: release-alternate
+          arch: ${{ matrix.arch }}
+          pkg_name: consul-k8s-control-plane_${{ env.version }}
+          bin_name: consul-k8s-control-plane 
+          workdir: control-plane
+          tags: |
+            docker.io/hashicorp/${{ env.repo }}-control-plane-ubi:${{ env.version }}
+            986891699432.dkr.ecr.us-east-1.amazonaws.com/hashicorp/${{ env.repo }}-control-plane-ubi:${{ env.version }}
```
ishustava, line 255: I think this would have to go to redhat image registry

.github/workflows/build.yml
```diff
@@ -0,0 +1,259 @@
...
+          dev_tags: |
+            hashicorppreview/${{ env.repo }}-control-plane-ubi:${{ env.version }}-dev
+            docker.io/hashicorppreview/${{ env.repo }}-control-plane-ubi:${{ env.version }}-${{ github.sha }}
```
ishustava, line 258: should it go to the `hashicorpdev` docker repo/account? I'm not familiar with the `hashicorppreview`

ishustava (commented):
Great work, Curt!!! This is awesome!

I left some questions in-line, mainly to clear up on my understanding of things

.github/workflows/build.yml
curtbushko: Oh yes, I think it could. I was following the pattern that I had seen other teams use but there really isn't any reason for it to be separate.

I will move it over.

control-plane/Dockerfile
curtbushko: Sure!

.github/workflows/build.yml
curtbushko: I changed it to `registry.connect.redhat.com/hashicorp/consul-k8s-control-plane:${{env.version}}-ubi` which is listed at: https://catalog.redhat.com/software/containers/hashicorp/consul-k8s-control-plane/61524326bc74e0be7ca66c58?container-tabs=gtiv

control-plane/Dockerfile
curtbushko: For some reason 'groupadd' was missing from the ubi image. I pulled it locally and yes it was missing. Very weird.

ishustava: Oh I've missed it in the original image, we do install it there too. I think we also need to install some of those other packages then too:

https://github.com/hashicorp/consul-k8s/blob/d2ce3750c554779139efc2671092f1a8a326db04/control-plane/build-support/docker/Release.ubi.dockerfile#L38

From looking at them, I think some are only needed for installation (like `gnupg`, `wget`, `unzip` and `tar`) so we can probably skip them. But I feel like others we would need baked into the image for sure:
- `iptables` we need this for tproxy
- `ca-certificates` we probably need this to have the latest trusted ca certs 
- `curl` is used by the `tls-init-cleanup` job 
- `libcap` - not sure why we need it but we have it in our release image too
-  `openssl` - not sure what uses it but we have it in our release image too

I think we need to test these images before we merge, wdyt?

ishustava: Also, we probably need to do a similar audit of the utils for the release image

control-plane/Dockerfile
```diff
@@ -0,0 +1,131 @@
...
+
+# Set ARGs as ENV so that they can be used in ENTRYPOINT/CMD
+ENV NAME=${BIN_NAME}
+ENV VERSION=${VERSION}
+
+# TARGETOS and TARGETARCH are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+# This is the location of the releases.
+ENV HASHICORP_RELEASES=https://releases.hashicorp.com
```
ishustava, line 107: this is no longer used

control-plane/Dockerfile
curtbushko: That makes sense to me. This means that I will need to get an OpenShift environment running somewhere?

ishustava: For the release and dev images, it should be sufficient to build them, push them to some docker repo and kick off acceptance tests in a draft PR.

For openshift, it's a bit more complicated. I'd still push the openshift image to dockerhub. You can then use [Azure RHO](https://docs.microsoft.com/en-us/azure/openshift/tutorial-create-cluster) to set up a cluster and run tests against it.

.github/workflows/build.yml
```diff
@@ -0,0 +1,262 @@
...
+          zip -r -j out/${{ matrix.pkg_name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip dist/
+
+      - name: Upload built binaries 
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.pkg_name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+          path: ${{ matrix.component}}/out/${{ matrix.pkg_name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+            
+      - name: Package rpm and deb files 
+        if: ${{ matrix.goos == 'linux' && matrix.component == 'cli' && matrix.goarch == 'amd64'}}
```
curtbushko, line 121: It turns out that we only ship AMD64 rpm and debs (verified through the rpm upload website and in our docs). Most HC products only support AMD64 for rpm and debs

.github/workflows/build.yml
```diff
@@ -0,0 +1,262 @@
...
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "linux", goarch: "arm64", component: "control-plane", pkg_name: "consul-k8s-control-plane" } 
+            # solaris is only built for the control plane
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "solaris", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane" } 
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "windows", goarch: "386", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "windows", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "amd64", component: "cli", pkg_name: "consul-k8s" }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane" }
```
lkysow, line 91: are we gonna do arm64 as well? (apple silicon)

curtbushko: Sure!  (I just added it)

control-plane/Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+# 
+#   Non-release images.
+#
+# ===================================
+
+# dev copies the binary from a local build
+# -----------------------------------
+# BIN_NAME is a requirement in the hashicorp docker github action 
+
+FROM alpine:3.15 AS dev
```
curtbushko, line 18: @ishustava Dockerfile dev target is 90% the same as as default-release target. The only difference is where it copies the binary from. 

I am leaving some tech debt here that we should remove when we get rid of CircleCI. That debt is the local build places the built binary in /pkg/bin/<arch>/<binary> and this is the location that CircleCI depends on to do acceptance tests. When we fully move to Github Actions we can move everything to /dist

control-plane/Makefile
```diff
@@ -31,15 +31,13 @@ ci.dev-tree:
 # In CircleCI, the linux binary will be attached from a previous step at pkg/bin/linux_amd64/. This make target
 # should only run in CI and not locally.
 ci.dev-docker:
-	@echo "Pulling consul-k8s container image - $(CONSUL_K8S_IMAGE_VERSION)"
-	@docker pull hashicorp/consul-k8s:$(CONSUL_K8S_IMAGE_VERSION) >/dev/null #todo change this back after pulling it the first time since the dockerfile is FROM this image
 	@echo "Building consul-k8s Development container - $(CI_DEV_DOCKER_IMAGE_NAME)"
 	@docker build -t '$(CI_DEV_DOCKER_NAMESPACE)/$(CI_DEV_DOCKER_IMAGE_NAME):$(GIT_COMMIT)' \
-	--build-arg CONSUL_K8S_IMAGE_VERSION=$(CONSUL_K8S_IMAGE_VERSION) \
+	--target=dev \
```
curtbushko, line 36: CircleCI will now use the dev target that is pretty close to the release-default target.

control-plane/Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+      version=${VERSION} \
+      release=${VERSION} \
+      summary="consul-k8s-control-plane provides first-class integrations between Consul and Kubernetes." \
+      description="consul-k8s-control-plane provides first-class integrations between Consul and Kubernetes."
+
+# Set ARGs as ENV so that they can be used in ENTRYPOINT/CMD
+ENV BIN_NAME=${BIN_NAME}
+ENV VERSION=${VERSION}
+
+RUN apk add --no-cache ca-certificates curl gnupg libcap openssl su-exec iputils libc6-compat iptables
```
curtbushko, line 37: @ishustava Added missing packages and acceptance tests succeeded

.github/workflows/build.yml
```diff
@@ -0,0 +1,264 @@
...
+          bin_name: consul-k8s-control-plane 
+          workdir: control-plane
+          tags: |
+            docker.io/hashicorp/${{ env.repo }}-control-plane:${{ env.version }}
+            986891699432.dkr.ecr.us-east-1.amazonaws.com/hashicorp/${{ env.repo }}-control-plane:${{ env.version }}
+          dev_tags: |
+            hashicorppreview/${{ env.repo }}-control-plane:${{ env.version }}-dev
+            docker.io/hashicorppreview/${{ env.repo }}-control-plane:${{ env.version }}-${{ github.sha }}
+
+  build-docker-alternate:
```
ishustava, line 228: let's comment this out for now so that we don't inadvertently push a ubi image that we haven't tested yet. wdyt?

ishustava (approved):
Awesome work!

.github/workflows/build.yml
curtbushko: Sounds good

PR #991, opened 2022-01-25
connect: Avoid making unnecessary calls to Consul in the endpoints controller

ishustava:
Fixes #779 
Based on work done in #795 

Changes proposed in this PR:
When Consul client is unavailable, it could slow down events processing in the endpoints controller because we always make *some* calls to consul (because `deregisterServiceOnAllAgents` is called on every `Reconcile`). When we can't reach a Consul client agent, it can take quite long (20-30sec) to process a single event because of the i/o timeout. This could cause your service mesh applications to be stuck in the init phase for longer than you'd expect.

This PR addresses the two cases when making those calls are unnecessary: 
1. When we are processing a service not on the mesh, i.e. when endpoints pods are not injected, we can skip any further processing of that endpoints object.
2. When we are deregistering services on all agents, we can skip agents that are not ready (work proposed in #795). This is OK because when/if an agent becomes ready again, we'll trigger another reconciliation of all services.

How I've tested this PR:
manually with reproduction steps from #779. In my testing I saw a decrease of `consul-connect-inject-init` runtime from ~2min to ~3sec.

How I expect reviewers to test this PR:
👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #1136, opened 2022-04-01
Fix bug when upgrading from non-mesh to mesh

lkysow:
Changes proposed in this PR:
- fixes https://github.com/hashicorp/consul-k8s/issues/1143
- reverts some of https://github.com/hashicorp/consul-k8s/pull/991
- what was happening was that if any endpoint was a non-mesh pod we would short circuit and not look at the rest of the pods
- in addition, when you're removing the mesh annotation, if we saw a pod that wasn't injected we would short circuit and not try to delete any service instances

How I've tested this PR:
- unit tests
- built an image and tested

- code

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


t-eckert (approved):

ishustava (approved):
❤️

PR #1139, opened 2022-04-04
Fix cli to work on windows

lkysow:
Previously would error about missing file `consul\Chart.yaml`. This
is because the go embedded filesystem must be accessed using `/`
delimiters but we were using `filepath` functions which on windows
use `\`.

How I've tested this PR:
- tested on windows

- code

Checklist:
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ BUG FIXES:
 * Helm
   * Don't set TTL for server certificates when using Vault as the secrets backend. [[GH-1104](https://github.com/hashicorp/consul-k8s/pull/1104)]
   * Fix PodSecurityPolicies for clients/mesh gateways when hostNetwork is used. [[GH-1090](https://github.com/hashicorp/consul-k8s/pull/1090)]
+* CLI
+  * Fix `install` and `upgrade` commands for Windows. [[GH-1139](https://github.com/hashicorp/consul-k8s/pull/1139)]
 
 ## 0.41.1 (February 24, 2022)
 
--- a/cli/helm/chart.go
+++ b/cli/helm/chart.go
@@ -2,7 +2,8 @@ package helm
 
 import (
 	"embed"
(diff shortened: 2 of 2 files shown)


cli/helm/chart.go
```diff
@@ -86,11 +93,7 @@ func readFile(chart embed.FS, f string, pathPrefix string) (*loader.BufferedFile
 	if err != nil {
 		return nil, err
 	}
-	// Remove the path prefix.
```
lkysow, line 89: `Rel()` doesn't exist in the `path` package. I looked at copying it from `filepath` but it uses a lot of internal stuff to `filepath` so not an easy copy. I think we're okay just trimming the prefix here because we're always passing in pathPrefix (hence variable name), not something we need to properly calculate the relative directory of.

t-eckert (approved):
This looks really good. Thank you for catching and fixing it.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ BUG FIXES:
 * Helm
   * Don't set TTL for server certificates when using Vault as the secrets backend. [[GH-1104](https://github.com/hashicorp/consul-k8s/pull/1104)]
   * Fix PodSecurityPolicies for clients/mesh gateways when hostNetwork is used. [[GH-1090](https://github.com/hashicorp/consul-k8s/pull/1090)]
+* CLI
+  * Fix `install` command for Windows. [[GH-1139](https://github.com/hashicorp/consul-k8s/pull/1139)]
```
ishustava, line 23: would this fix also work for other commands because the `chart.go` logic is common?

ishustava (approved):
🙏

CHANGELOG.md
t-eckert: Yes, it also fixes the `upgrade` command.

lkysow: updated!

PR #1146, opened 2022-04-05
Upload docker image in nightly tests

t-eckert:
Changes proposed in this PR:
- Nightly tests will upload their own Docker image instead of relying on the image existing from a previous `test-and-build` run.

How I've tested this PR:

I will see how this performs tonight and see if there are any adjustments I need to make.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


t-eckert:
I'm fairly certain that if you upload an image with a tag that already exists, it will just be overwritten. This change depends on that behavior. If that is not the case, please let me know and I will create a workaround. I tested it and it seemed to work how I expect.

ndhanushkodi:
Can you run this on GKE on this PR and see it go green?

PR #1150, opened 2022-04-07
Add annotation for explicitly specifying which service you want registered

andrewstucki:
So, I was helping someone go through the process of setting up an Elasticsearch instance on the service mesh so it could be routed to via the API Gateway. All of the pods and services are created via a CRD specified by [this project](https://github.com/elastic/cloud-on-k8s).

One of the pain points of this is that the CRD + controller create a single deployment that has multiple services targeting it. You can override parts of the pod spec template in the CRD and some of the service templates, but not all of them. This makes things difficult because our only way of properly specifying which service we want to use in Consul registration when a pod is using `connect-inject` and has multiple services targeting it is through the service label [consul.hashicorp.com/service-ignore](https://www.consul.io/docs/k8s/annotations-and-labels#consul-hashicorp-com-service-ignore). Since these services are managed by a third-party controller it means you have to patch that label onto each service (and who knows if the controller just resets it in a reconciliation run).

In the case of a vanilla setup without all of these additional CRDs and controllers, you would have to go through the cumbersome process of ensuring _every_ service that targets the deployment is labeled with `service-ignore` other than the one you want registered.

As an alternative to doing that, my thought is that we should just allow for an annotation on the pod that says, "here's the Kubernetes service that I want to use for registration" and, when specified, ignore any endpoints that don't have that explicit name. I think apart from increasing the likelihood of compatibility with third-party systems, this is an overall better user experience than having to label a bunch of services.

Changes proposed in this PR:
- Add a `"consul.hashicorp.com/kubernetes-service"` annotation for pods

How I've tested this PR:
- Unit tested - can probably end-to-end test this later today with a custom Docker build.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -178,7 +178,7 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 				serviceName, ok := pod.Annotations[annotationKubernetesService]
 				if ok && serviceEndpoints.Name != serviceName {
 					r.Log.Info("ignoring endpoint because it doesn't match explicit service annotation", "name", serviceEndpoints.Name, "ns", serviceEndpoints.Namespace)
-					// deregistration happens later because we don't add this pod to the endpointAddressMap
+					// deregistration for service instances that don't match the annotation happens later because we don't add this pod to the endpointAddressMap.
 					continue
 				}
 
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -175,6 +175,12 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 					continue
 				}
 
+				serviceName, ok := pod.Annotations[annotationKubernetesService]
+				if ok && serviceEndpoints.Name != serviceName {
+					r.Log.Info("ignoring endpoint because it doesn't match explicit service annotation", "name", serviceEndpoints.Name, "ns", serviceEndpoints.Namespace)
+					continue
+				}
+
 				if hasBeenInjected(pod) {
 					endpointPods.Add(address.TargetRef.Name)
 					if err := r.registerServicesAndHealthCheck(pod, serviceEndpoints, healthStatus, endpointAddressMap); err != nil {
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -178,6 +178,7 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 				serviceName, ok := pod.Annotations[annotationKubernetesService]
 				if ok && serviceEndpoints.Name != serviceName {
 					r.Log.Info("ignoring endpoint because it doesn't match explicit service annotation", "name", serviceEndpoints.Name, "ns", serviceEndpoints.Namespace)
+					// deregistration happens later because we don't add this pod to the endpointAddressMap
 					continue
 				}
 
--- a/control-plane/connect-inject/endpoints_controller_test.go
+++ b/control-plane/connect-inject/endpoints_controller_test.go
@@ -3491,7 +3491,7 @@ func TestReconcile_podSpecifiesExplicitService(t *testing.T) {
 
 	// Run the reconcile again with the service we want to register.
 	serviceName = endpoint.Name
-	namespacedName = types.NamespacedName{Namespace: badEndpoint.Namespace, Name: serviceName}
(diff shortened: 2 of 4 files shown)


andrewstucki:
I also just end-to-end tested this and things seem peachy, here's what I did.

Went through the [API Gateway](https://learn.hashicorp.com/tutorials/consul/kubernetes-api-gateway) tutorial with the following changes (where `public.ecr.aws/d0j3t2j0/foobar:3` is an image built with `make control-plane-dev-docker DEV_IMAGE=public.ecr.aws/d0j3t2j0/foobar:3`):

```diff
diff --git a/api-gateway/consul/config.yaml b/api-gateway/consul/config.yaml
index 205b470..a510871 100644
--- a/api-gateway/consul/config.yaml
+++ b/api-gateway/consul/config.yaml
@@ -1,6 +1,7 @@
 global:
   name: consul
   datacenter: dc1
+  imageK8S: public.ecr.aws/d0j3t2j0/foobar:3
   tls:
     enabled: true
 server:
diff --git a/api-gateway/two-services/echo-1.yaml b/api-gateway/two-services/echo-1.yaml
index 3b68724..40935c7 100644
--- a/api-gateway/two-services/echo-1.yaml
+++ b/api-gateway/two-services/echo-1.yaml
@@ -22,6 +22,21 @@ spec:
     app: echo-1
 ---
 apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: echo-1
+  name: bad-echo-1
+spec:
+  ports:
+  - port: 8081
+    name: high
+    protocol: TCP
+    targetPort: 8081
+  selector:
+    app: echo-1
+---
+apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: echo-1
@@ -44,12 +59,15 @@ spec:
         app: echo-1
       annotations:
         'consul.hashicorp.com/connect-inject': 'true'
+        'consul.hashicorp.com/connect-service-port': '8080'
+        'consul.hashicorp.com/kubernetes-service': 'echo-1'
     spec:
       serviceAccountName: echo-1
       containers:
       - image: gcr.io/kubernetes-e2e-test-images/echoserver:2.2
         name: echo-1
         ports:
+        - containerPort: 8081
         - containerPort: 8080
         env:
           - name: NODE_NAME
```

I took turns deleting and recreating both `echo-1` and `bad-echo-1` to make sure only `echo-1` was ever registered. I also tried removing the new annotation and almost immediately ran into this initialization error:

> 2022-04-07T19:08:32.501Z [ERROR] There are multiple Consul services registered for this pod when there must only be one. Check if there are multiple Kubernetes services selecting this pod and add the label `consul.hashicorp.com/service-ignore: "true"` to all services except the one used by Consul for handling requests.

Here's a screenshot of the registered service:

Screen Shot 2022-04-07 at 3 01 38 PM

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -3373,6 +3373,138 @@ func TestReconcile_endpointsIgnoredWhenNotInjected(t *testing.T) {
...
+	serviceInstances, _, err = consulClient.Catalog().Service(serviceName, "", nil)
+	require.NoError(t, err)
+	require.Len(t, serviceInstances, 0)
+	proxyServiceInstances, _, err := consulClient.Catalog().Service(serviceName+"-sidecar-proxy", "", nil)
+	require.NoError(t, err)
+	require.Len(t, proxyServiceInstances, 0)
+
+	// Run the reconcile again with the service we want to register.
+	serviceName = endpoint.Name
+	namespacedName = types.NamespacedName{Namespace: badEndpoint.Namespace, Name: serviceName}
```
ndhanushkodi, line 3494:
```suggestion
	namespacedName = types.NamespacedName{Namespace: endpoint.Namespace, Name: serviceName}
```

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -175,6 +175,13 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
 					continue
 				}
 
+				serviceName, ok := pod.Annotations[annotationKubernetesService]
+				if ok && serviceEndpoints.Name != serviceName {
+					r.Log.Info("ignoring endpoint because it doesn't match explicit service annotation", "name", serviceEndpoints.Name, "ns", serviceEndpoints.Namespace)
+					// deregistration happens later because we don't add this pod to the endpointAddressMap
```
ndhanushkodi, line 181:
```suggestion
					// deregistration for service instances that don't match the annotation happens later because we don't add this pod to the endpointAddressMap.
```

ndhanushkodi (approved):
This looks great! Thanks for making this change!!

curtbushko (approved):
Looks great! Thanks Andrew!

PR #1151, opened 2022-04-07
Improve Nightly Test Debugging and Reliability with Docker Uploads

t-eckert:
Changes proposed in this PR:
- Add checks to verify deletion of resources in Destroy method on Helm Cluster

How I've tested this PR:
I set up this PR run nightly-acceptance-tests.

Watch the nightly-acceptance-tests and see if they succeed.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -891,6 +891,23 @@ workflows:
       - acceptance-tproxy:
           requires:
             - dev-upload-docker
+      - cleanup-gcp-resources
+      - cleanup-azure-resources
+      - cleanup-eks-resources
+      # Disable until we can use UBI images.
+      #      - acceptance-openshift:
+      #          requires:
+      #          - cleanup-azure-resources
+      - acceptance-gke-1-20:
+          requires:
+            - cleanup-gcp-resources
+      - acceptance-eks-1-19:
+          requires:
+            - cleanup-eks-resources
+      - acceptance-aks-1-21:
+          requires:
+            - cleanup-azure-resources
+      - acceptance-kind-1-23
   nightly-acceptance-tests:
     triggers:
       - schedule:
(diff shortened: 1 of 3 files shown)


control-plane/Makefile
```diff
@@ -46,6 +46,25 @@ ifeq ($(CIRCLE_BRANCH), main)
 	@docker push $(CI_DEV_DOCKER_NAMESPACE)/$(CI_DEV_DOCKER_IMAGE_NAME):latest
 endif
 
+# TODO: Remove this ci.nightly-dev-docker target once we move the acceptance tests to Github Actions.
+# In CircleCI, the linux binary will be attached from a previous step at pkg/bin/linux_amd64/. This make target
+# should only run in CI and not locally.
```
ndhanushkodi, line 51:
```suggestion
# should only run in CI and not locally. This is a guardrail for a case where the latest commit on main does not succeed to upload a docker image, so with this step the nightly test will upload its own docker image off of main before starting.
```

.circleci/config.yml
```diff
@@ -891,6 +905,23 @@ workflows:
       - acceptance-tproxy:
           requires:
             - dev-upload-docker
+      - cleanup-gcp-resources
```
ndhanushkodi, line 908: reminder to remove 908-924 before merging :)

ndhanushkodi (approved):
These look good Thomas! They'll help us realize the underlying issues a little sooner rather than having to dig for them. Thanks!!

thisisnotashwin (approved):
This looks good! Could you update the commit to say debug nightly tests? I don't know if this strictly fixes our tests.

t-eckert:
> Could you update the commit to say debug nightly tests?

Is what I changed it to good? I didn't update the commit, but when it squashes it will have the title of the PR as the commit.

I'm OOO today so if you think it's good, go ahead and squash it into `main`.

PR #1154, opened 2022-04-08
Get federation secret from any of the places it is set.

t-eckert:
Changes proposed in this PR:
- When installing a federated cluster, the install command will look in all of the places the name of the federation secret could be set and return the first set value it finds.

How I've tested this PR:

Added unit tests, ran the install tests which exercise the presence of the fed secret.

Running the unit tests as well if you like.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ BUG FIXES:
 * CLI
   * Fix issue where clusters not in the same namespace as their deployment name could not be upgraded. [[GH-1115](https://github.com/hashicorp/consul-k8s/pull/1115)]
   * Fix issue where the CLI was looking for secrets in namespaces other than the namespace targeted by the release. [[GH-1156](https://github.com/hashicorp/consul-k8s/pull/1156)]
+  * Fix issue where the federation secret was not being found in certain configurations. [[GH-1154](https://github.com/hashicorp/consul-k8s/issue/1154)]
 
 IMPROVEMENTS:
 * Helm
--- a/cli/release/release.go
+++ b/cli/release/release.go
@@ -23,3 +23,27 @@ func (r *Release) ShouldExpectFederationSecret() bool {
 		r.Configuration.Global.Datacenter != r.Configuration.Global.Federation.PrimaryDatacenter &&
 		!r.Configuration.Global.Federation.CreateFederationSecret
 }
+
+// FedSecret fetches the federation secret for the release or an empty string
+// if the federation secret is not set. The federation secret could be set
+// in several places in the configuration. This method returns the first
+// secret name that is set.
+func (r *Release) FedSecret() string {
+	if s := r.Configuration.Global.TLS.CaCert.SecretName; s != "" {
+		return s
+	}
+
+	if s := r.Configuration.Global.TLS.CaKey.SecretName; s != "" {
+		return s
+	}
+
+	if s := r.Configuration.Global.Acls.ReplicationToken.SecretName; s != "" {
+		return s
+	}
+
+	if s := r.Configuration.Global.GossipEncryption.SecretName; s != "" {
+		return s
+	}
+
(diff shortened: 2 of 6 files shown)


cli/release/release.go
```diff
@@ -23,3 +23,27 @@ func (r *Release) ShouldExpectFederationSecret() bool {
...
+func (r *Release) FedSecret() string {
+	if s := r.Configuration.Global.TLS.CaCert.SecretName; s != "" {
+		return s
+	}
+
+	if s := r.Configuration.Global.TLS.CaKey.SecretName; s != "" {
+		return s
+	}
+
+	if s := r.Configuration.Global.Acls.ReplicationToken.SecretName; s != "" {
```
curtbushko, line 40: Good job! This should find the right secret! I only have one question/comment:

In the helm chart and RFC CSL-119, it looks like acls.enabled can be false when using WAN federation is enabled. TLS and Gossip  Encryption need to be true.  

Should you check to see if r.Configuration.Global.Acls.Enabled = true before checking for r.Configuration.GlobalAcls.ReplicationToken.SecretName ?

jmurret: Is this right? Isn't the secret `RELEASE-NAME-federation` in the secondary data center?  

I know this is not gossip key, but the createFederationSecret adds gossip key,replication token, ca cert and keyinto the federation secret. The [docs](https://www.consul.io/docs/k8s/operations/gossip-encryption-key-rotation) for rotating gossip keys also show in step 5 that in secondary data centers you need get the gossip key from `consul-federation` and it uses the key of `gossipEncryptionKey`.  So I would assume that acl replication token works the same way and gets it from the same place where [it is put](https://github.com/hashicorp/consul-k8s/blob/d708a65db4875f689fb304e77a81e577923cee8b/control-plane/subcommand/create-federation-secret/command.go#L196) ?

t-eckert: I was basing the work off of [these docs](https://www.consul.io/docs/k8s/installation/multi-cluster/kubernetes#secondary-cluster-s) and just looking at any place that the user would set the secret to `consul-federation`.

jmurret: In the docs, secondary datacenters are set that way because that secret name is hardcoded here: https://github.com/hashicorp/consul-k8s/blob/d708a65db4875f689fb304e77a81e577923cee8b/control-plane/subcommand/create-federation-secret/command.go#L129

So these docs are just saying "You need to set these this way because it is a hard coded name and this consul-federation secret will have keys for gossipkey, ca cert, ca key, and replication token."

I think you want to use the hardcoded `<prefix>-federation` to check if the secret exists.  Not use this fallback pattern.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 BUG FIXES:
   * Fix issue where clusters not in the same namespace as their deployment name could not be upgraded. [[GH-1115](https://github.com/hashicorp/consul-k8s/issue/1005)]
+  * Fix issue where the federation secret was not being found in certain configurations. [[GH-1154](https://github.com/hashicorp/consul-k8s/issue/1154)]
```
ishustava, line 5: 🚗 :homerfade:
```suggestion
  * Fix issue where the federation secret was not being found in certain configurations. [[GH-1154](https://github.com/hashicorp/consul-k8s/pull/1154)]
```

cli/release/release.go
t-eckert: That's a great point. Thank you for finding this. I will update the method to just return the release name with `"-federation"` appended. Now that it's so simple, it doesn't _need_ to be a method, but I like having it so that if this ever changes, we can update it in just one place.

curtbushko (approved):

jmurret (approved):
🎉 nice work!

PR #1156, opened 2022-04-08
Only lookup secrets in the namespace where the install is targeting.

t-eckert:
Changes proposed in this PR:
- The install command will only look for secrets in the namespace the install is targeting.
  This fixes the issue raised by @lkysow.
  > another one, it's complaining I have a secret in my default ns but I'm installing consul into the consul ns:
  > ``` bash
  > consul-k8s install
  > ==> Checking if Consul can be installed
  > ✓ No existing Consul installations found.
  > ✓ No existing Consul persistent volume claims found
  > ! Found Consul secrets, possibly from a previous installation.
  > Delete existing Consul secrets from Kubernetes:
  > 
  > kubectl delete secret consul-federation --namespace default
  > ``` 

How I've tested this PR:

Extended existing unit tests to cover this case.

:eyes: and running unit tests.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 BUG FIXES:
 * CLI
   * Fix issue where clusters not in the same namespace as their deployment name could not be upgraded. [[GH-1115](https://github.com/hashicorp/consul-k8s/pull/1115)]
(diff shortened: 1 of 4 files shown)


jmurret (approved):
Nice!

lkysow (approved):

PR #1169, opened 2022-04-13
Re-use connectInject.consulNamespaces for API Gateway

nathancoleman:
**Background:**
If ACLs are enabled and a user doesn't have identical mappings for the helm chart under `connectInject.consulNamespaces` and `apiGateway.consulNamespaces`, everything breaks. This is because the API Gateway uses a combination of both internally which results in a gateway trying to register itself under a different Consul namespace than it is authorized to.

After talking with @lkysow, we determined that these two `consulNamespaces` options should really be one and the same. Letting them diverge breaks many uses cases. In addition, it's cumbersome for the practitioner to have to specify both with the exact same value.

- Drop `apiGateway.consulNamespaces` and instead re-use `connectInject.consulNamespaces` for the API Gateway installation

**How I've tested this PR:**
Installed chart from this branch with various `connectInject.consulNamespaces` configurations, verifying that those values are piped through to the API Gateway controller Deployment spec just as they were previously from `apiGateway.consulNamespaces`.

<details>
<summary>Example</summary>

```yaml
connectInject:
  enabled: true
  consulNamespaces:
    mirroringK8S: true
    mirroringK8SPrefix: "nate-"
  default: true
```
image

</details>

**Checklist:**
- [ ] Tests added
- [x] CHANGELOG entry added 


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ BUG FIXES:
 IMPROVEMENTS:
 * Helm
   * API Gateway: Allow controller to read ReferencePolicy in order to determine if route is allowed for backend in different namespace. [[GH-1148](https://github.com/hashicorp/consul-k8s/pull/1148)]
+  * API Gateway: Re-use connectInject.consulNamespaces instead of requiring apiGateway.consulNamespaces have the same value. [[GH-1169](https://github.com/hashicorp/consul-k8s/pull/1169)]
```
lkysow, line 18: It should be a breaking change I think?

nathancoleman: I agree it _feels_ like a breaking change, but I'm still thinking on it.My understanding is that there's no valid case where `apiGateway.consulNamespaces` differs from `connectInject.consulNamespaces`, and it doesn't break if you do something like `helm install ... --set apiGateway.consulNamespaces...` even after removing from `values.yaml`.

lkysow: Overall I think better safe than sorry. I'm sure there's a contrived case where this would be a breaking change. For example, I've got `gateway.destNamespace=default` and I've got my gateway in kube ns `foo` and I've got `inject.mirroring=true`. Now with this change my gateway will move from `default` to `foo`.

nathancoleman: Good point! Addressed

andrewstucki:
@nathancoleman looks like some stray empty waypoint server files accidentally got checked in.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+BREAKING CHANGES:
+* Helm
+  * API Gateway: Re-use connectInject.consulNamespaces instead of requiring apiGateway.consulNamespaces have the same value. [[GH-1169](https://github.com/hashicorp/consul-k8s/pull/1169)]
```
andrewstucki, line 5: nit:
```suggestion
  * API Gateway: Re-use connectInject.consulNamespaces instead of requiring that apiGateway.consulNamespaces have the same value when ACLs are enabled. [[GH-1169](https://github.com/hashicorp/consul-k8s/pull/1169)]
```

andrewstucki (approved):

ishustava (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+BREAKING CHANGES:
+* Helm
+  * API Gateway: Re-use connectInject.consulNamespaces instead of requiring that apiGateway.consulNamespaces have the same value when ACLs are enabled. [[GH-1169](https://github.com/hashicorp/consul-k8s/pull/1169)]
```
lkysow, line 5: Not a big deal but for BC's it's nice to tell the user what to do. So in this case if their `apiGateway.consulNamespaces` is the same as their `connectInject.consulNamespaces` configuration, they don't have to do anything. If it's different, then they need to be aware that the `connectInject.consulNamespaces` will now be used.

PR #1173, opened 2022-04-14
Cleanup EKS VPC peering connections

ndhanushkodi:
Changes proposed in this PR:
- Currently if you run terraform apply with cluster count of 2, and then run our cleanup script, it will fail because VPCs can't get deleted since they have a dependency on peering connections
- This allows you to run the cleanup script successfully, by deleting peering conns

How I've tested this PR:
- Ran script after terraforming 2 eks clusters, and saw it succeed.

👀


thisisnotashwin (approved):
Thanks so much for doing this!!

ishustava (approved):
🙏

PR #1180, opened 2022-04-20
Add support for ACLs for sync catalog and partitions.

thisisnotashwin:
Changes proposed in this PR:
- Update rules.go to support partitions when Sync catalog is used.
- Add an acceptance test to validate sync catalog works when partitions are enabled in the default and non-default partitions.

How I've tested this PR:
- Units tests.
- Acceptance tests.
- Code Review.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/partitions/partitions_sync_test.go
+++ b/acceptance/tests/partitions/partitions_sync_test.go
@@ -1,0 +1,297 @@
+package partitions
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"testing"
+	"time"
+
+	terratestk8s "github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/consul"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/environment"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Test that Sync Catalog works in a default and ACLsAndAutoEncryptEnabled installations for partitions.
+func TestPartitions_Sync(t *testing.T) {
+	env := suite.Environment()
+	cfg := suite.Config()
+
+	if !cfg.EnableEnterprise {
+		t.Skipf("skipping this test because -enable-enterprise is not set")
+	}
+
+	const defaultPartition = "default"
+	const secondaryPartition = "secondary"
+	const defaultNamespace = "default"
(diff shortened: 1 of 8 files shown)


control-plane/subcommand/server-acl-init/rules_test.go
```diff
@@ -496,111 +498,251 @@ func TestSyncRules(t *testing.T) {
...
+    node_prefix "" {
+      policy = "read"
+    }
+    service_prefix "" {
+      policy = "write"
+    }`,
+		},
+		{
+			Name:                           "Namespaces are disabled, non-default node name",
+			ConsulSyncDestinationNamespace: "sync-namespace",
```
curtbushko, line 510: Even though there are defaults, should you be specific about all settings in the test. For example, Add 

```
EnablePartitions: "false",
PartitionName: "",
EnableNamespaces: "false",
```

to this test.

The only reason I bring this up is that I could see us or docs in the future grabbing one of these testcases and saying 'this matrix of settings, results in this'

curtbushko (changes requested):
Super nice! Acceptance tests are so great!

I only have one nit/question below but don't let it block the PR.

control-plane/subcommand/server-acl-init/rules_test.go
thisisnotashwin: good call!! I think i fixed that if youd like to take a look.

curtbushko (approved):

acceptance/tests/partitions/partitions_sync_test.go
```diff
@@ -0,0 +1,297 @@
...
+				"dns.enableRedirection": strconv.FormatBool(cfg.EnableTransparentProxy),
+			}
+
+			serverHelmValues := map[string]string{
+				"server.exposeGossipAndRPCPorts": "true",
+			}
+
+			// On Kind, there are no load balancers but since all clusters
+			// share the same node network (docker bridge), we can use
+			// a NodePort service so that we can access node(s) in a different Kind cluster.
```
t-eckert, line 114: This is a great clarifying comment. 👍🏻

acceptance/tests/partitions/partitions_sync_test.go
```diff
@@ -0,0 +1,297 @@
...
+			// Install the consul cluster with servers in the default kubernetes context.
+			serverConsulCluster := consul.NewHelmCluster(t, serverHelmValues, serverClusterContext, cfg, releaseName)
+			serverConsulCluster.Create(t)
+
+			// Get the TLS CA certificate and key secret from the server cluster and apply it to the client cluster.
+			caCertSecretName := fmt.Sprintf("%s-consul-ca-cert", releaseName)
+			caKeySecretName := fmt.Sprintf("%s-consul-ca-key", releaseName)
+
+			logger.Logf(t, "retrieving ca cert secret %s from the server cluster and applying to the client cluster", caCertSecretName)
+			moveSecret(t, serverClusterContext, clientClusterContext, caCertSecretName)
```
t-eckert, line 133: I know you didn't add this function, but I'm wondering if it might be clearer if it were called `copySecret`. "Move" makes me think that the original instance is no longer there.

t-eckert (approved):
LGTM

PR #1191, opened 2022-04-27
Webhook Certs Using Vault

jmurret:
Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -41,7 +41,7 @@ func TestVault(t *testing.T) {
 
 	bootstrapToken := vault.ConfigureACLTokenVaultSecret(t, vaultClient, "bootstrap")
 
-	serverPolicies := "gossip,connect-ca-dc1,server-cert-dc1,bootstrap-token,connect-webhook-cert-dc1"
+	serverPolicies := "gossip,connect-ca-dc1,server-cert-dc1,bootstrap-token"
 	if cfg.EnableEnterprise {
 		serverPolicies += ",license"
 	}
@@ -55,6 +55,9 @@ func TestVault(t *testing.T) {
 	pathForConnectInjectWebookCerts :=
 		vault.ConfigurePKICertificatesForConnectInjectWebhook(t, vaultClient,
 			consulReleaseName, ns, "dc1", "1h")
+	pathForControllerWebookCerts :=
+		vault.ConfigurePKICertificatesForControllerWebhook(t, vaultClient,
+			consulReleaseName, ns, "dc1", "1h")
 
 	vaultCASecret := vault.CASecretName(vaultReleaseName)
 
@@ -67,6 +70,7 @@ func TestVault(t *testing.T) {
 		"connectInject.replicas":           "1",
 		"connectInject.tlsCert.secretName": pathForConnectInjectWebookCerts,
 		"controller.enabled":               "true",
+		"controller.tlsCert.secretName":    pathForControllerWebookCerts,
 
 		"global.secretsBackend.vault.enabled":              "true",
 		"global.secretsBackend.vault.consulServerRole":     "server",
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -41,7 +41,7 @@ func TestVault(t *testing.T) {
 
 	bootstrapToken := vault.ConfigureACLTokenVaultSecret(t, vaultClient, "bootstrap")
 
-	serverPolicies := "gossip,connect-ca-dc1,server-cert-dc1,bootstrap-token"
+	serverPolicies := "gossip,connect-ca-dc1,server-cert-dc1,bootstrap-token,connect-webhook-cert-dc1"
 	if cfg.EnableEnterprise {
 		serverPolicies += ",license"
 	}
@@ -63,10 +63,10 @@ func TestVault(t *testing.T) {
 		"server.extraVolumes[0].name": vaultCASecret,
 		"server.extraVolumes[0].load": "false",
 
-		"connectInject.enabled":    "true",
-		"connectInject.replicas":   "1",
-		"connectInject.tlsCertDir": pathForConnectInjectWebookCerts,
-		"controller.enabled":       "true",
+		"connectInject.enabled":            "true",
+		"connectInject.replicas":           "1",
+		"connectInject.tlsCert.secretName": pathForConnectInjectWebookCerts,
+		"controller.enabled":               "true",
 
 		"global.secretsBackend.vault.enabled":              "true",
 		"global.secretsBackend.vault.consulServerRole":     "server",
--- a/acceptance/tests/vault/vault_test.go
+++ b/acceptance/tests/vault/vault_test.go
@@ -52,6 +52,9 @@ func TestVault(t *testing.T) {
 
 	vault.ConfigurePKICA(t, vaultClient)
 	certPath := vault.ConfigurePKICertificates(t, vaultClient, consulReleaseName, ns, "dc1", "1h")
+	pathForConnectInjectWebookCerts :=
+		vault.ConfigurePKICertificatesForConnectInjectWebhook(t, vaultClient,
+			consulReleaseName, ns, "dc1", "1h")
 
 	vaultCASecret := vault.CASecretName(vaultReleaseName)
 
@@ -60,9 +63,10 @@ func TestVault(t *testing.T) {
 		"server.extraVolumes[0].name": vaultCASecret,
 		"server.extraVolumes[0].load": "false",
 
-		"connectInject.enabled":  "true",
-		"connectInject.replicas": "1",
-		"controller.enabled":     "true",
+		"connectInject.enabled":    "true",
+		"connectInject.replicas":   "1",
+		"connectInject.tlsCertDir": pathForConnectInjectWebookCerts,
+		"controller.enabled":       "true",
 
 		"global.secretsBackend.vault.enabled":              "true",
 		"global.secretsBackend.vault.consulServerRole":     "server",
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -45,6 +45,22 @@ as well as the global.name setting.
             {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
 
+{{- define "consul.connectInjectWebhookTLSCertTemplate" -}}
+ |
+            {{ "{{" }}- with secret "{{ .Values.connectInject.tlsCert.secretName }}" "{{ printf "common_name=server.%s.%s" .Values.global.datacenter .Values.global.domain }}"
+            "alt_names={{ include "consul.serverTLSAltNames" . }}" "ip_sans=127.0.0.1{{ include "consul.serverAdditionalIPSANs" . }}" -{{ "}}" }}
+            {{ "{{" }}- .Data.certificate -{{ "}}" }}
+            {{ "{{" }}- end -{{ "}}" }}
+{{- end -}}
+
+{{- define "consul.connectInjectWebhookTLSKeyTemplate" -}}
+ |
+            {{ "{{" }}- with secret "{{ .Values.connectInject.tlsCert.secretName }}" "{{ printf "common_name=server.%s.%s" .Values.global.datacenter .Values.global.domain }}"
+            "alt_names={{ include "consul.serverTLSAltNames" . }}" "ip_sans=127.0.0.1{{ include "consul.serverAdditionalIPSANs" . }}" -{{ "}}" }}
+            {{ "{{" }}- .Data.private_key -{{ "}}" }}
+            {{ "{{" }}- end -{{ "}}" }}
+{{- end -}}
+
 {{- define "consul.serverTLSAltNames" -}}
 {{- $name := include "consul.fullname" . -}}
 {{- $ns := .Release.Namespace -}}
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -47,16 +47,16 @@ as well as the global.name setting.
 
 {{- define "consul.connectInjectWebhookTLSCertTemplate" -}}
  |
-            {{ "{{" }}- with secret "{{ .Values.connectInject.tlsCert.secretName }}" "{{ printf "common_name=server.%s.%s" .Values.global.datacenter .Values.global.domain }}"
-            "alt_names={{ include "consul.serverTLSAltNames" . }}" "ip_sans=127.0.0.1{{ include "consul.serverAdditionalIPSANs" . }}" -{{ "}}" }}
+            {{ "{{" }}- with secret "{{ .Values.connectInject.tlsCert.secretName }}" "{{ printf "common_name=connect-injector.%s.%s" .Values.global.datacenter .Values.global.domain }}"
+            "alt_names={{ include "consul.connectInjectorTLSAltNames" . }}" "ip_sans=127.0.0.1" -{{ "}}" }}
             {{ "{{" }}- .Data.certificate -{{ "}}" }}
             {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
 
 {{- define "consul.connectInjectWebhookTLSKeyTemplate" -}}
  |
-            {{ "{{" }}- with secret "{{ .Values.connectInject.tlsCert.secretName }}" "{{ printf "common_name=server.%s.%s" .Values.global.datacenter .Values.global.domain }}"
-            "alt_names={{ include "consul.serverTLSAltNames" . }}" "ip_sans=127.0.0.1{{ include "consul.serverAdditionalIPSANs" . }}" -{{ "}}" }}
+            {{ "{{" }}- with secret "{{ .Values.connectInject.tlsCert.secretName }}" "{{ printf "common_name=connect-injector.%s.%s" .Values.global.datacenter .Values.global.domain }}"
+            "alt_names={{ include "consul.connectInjectorTLSAltNames" . }}" "ip_sans=127.0.0.1" -{{ "}}" }}
             {{ "{{" }}- .Data.private_key -{{ "}}" }}
             {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
@@ -73,6 +73,12 @@ as well as the global.name setting.
 
 {{- define "consul.serverAdditionalIPSANs" -}}
 {{- if .Values.global.tls -}}{{- if .Values.global.tls.serverAdditionalIPSANs -}}{{- range $ipsan := .Values.global.tls.serverAdditionalIPSANs }},{{ $ipsan }} {{- end -}}{{- end -}}{{- end -}}
+{{- end -}}
+
+{{- define "consul.connectInjectorTLSAltNames" -}}
+{{- $name := include "consul.fullname" . -}}
+{{- $ns := .Release.Namespace -}}
+{{ printf "localhost,%s-connect-injector,*.%s-connect-injector,*.%s-connect-injector.%s,%s-connect-injector.%s,*.%s-connect-injector.%s.svc,%s-connect-injector.%s.svc,*.connect-injector.%s.%s" $name $name $name $ns $name $ns $name $ns $name $ns (.Values.global.datacenter ) (.Values.global.domain) }}
 {{- end -}}
 
 {{- define "consul.vaultReplicationTokenTemplate" -}}
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -261,7 +261,7 @@ spec:
           volumeMounts:
           {{- if not .Values.global.secretsBackend.vault.enabled }}
           - name: certs
-            mountPath: {{ .Values.connectInject.tlsCert.dir }}
+            mountPath: /etc/connect-injector/certs
             readOnly: true
           {{- end}}
           - mountPath: /consul/login
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -44,6 +44,10 @@ spec:
         "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulCARole }}
         "vault.hashicorp.com/agent-inject-secret-serverca.crt": {{ .Values.global.tls.caCert.secretName }}
         "vault.hashicorp.com/agent-inject-template-serverca.crt": {{ template "consul.serverTLSCATemplate" . }}
+        "vault.hashicorp.com/agent-inject-secret-injectwebhookcert.crt": {{ .Values.connectInject.tlsCert.secretName }}
+        "vault.hashicorp.com/agent-inject-template-injectwebhookcert.crt": {{ include "consul.connectInjectWebhookTLSCertTemplate" . }}
+        "vault.hashicorp.com/agent-inject-secret-injectwebhookcert.key": {{ .Values.connectInject.tlsCert.secretName }}
+        "vault.hashicorp.com/agent-inject-template-injectwebhookcert.key": {{ include "consul.connectInjectWebhookTLSCertTemplate" . }}
         {{- if and .Values.global.secretsBackend.vault.ca.secretName .Values.global.secretsBackend.vault.ca.secretKey }}
         "vault.hashicorp.com/agent-extra-secret": "{{ .Values.global.secretsBackend.vault.ca.secretName }}"
         "vault.hashicorp.com/ca-cert": "/vault/custom/{{ .Values.global.secretsBackend.vault.ca.secretKey }}"
@@ -167,7 +171,7 @@ spec:
                 -consul-cross-namespace-acl-policy=cross-namespace-policy \
                 {{- end }}
                 {{- end }}
-                -tls-cert-dir={{ .Values.connectInject.tlsCertDir }} \
+                -tls-cert-dir={{ .Values.connectInject.tlsCert.dir }} \
                 {{- $resources := .Values.connectInject.sidecarProxy.resources }}
                 {{- /* kindIs is used here to differentiate between null and 0 */}}
                 {{- if not (kindIs "invalid" $resources.limits.memory) }}
@@ -253,7 +257,7 @@ spec:
           volumeMounts:
           {{- if not .Values.global.secretsBackend.vault.enabled }}
           - name: certs
-            mountPath: {{ .Values.connectInject.tlsCertDir }}
+            mountPath: {{ .Values.connectInject.tlsCert.dir }}
             readOnly: true
           {{- end}}
           - mountPath: /consul/login
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -167,7 +167,7 @@ spec:
                 -consul-cross-namespace-acl-policy=cross-namespace-policy \
                 {{- end }}
                 {{- end }}
-                -tls-cert-dir=/etc/connect-injector/certs \
+                -tls-cert-dir={{ .Values.connectInject.tlsCertDir }} \
                 {{- $resources := .Values.connectInject.sidecarProxy.resources }}
                 {{- /* kindIs is used here to differentiate between null and 0 */}}
                 {{- if not (kindIs "invalid" $resources.limits.memory) }}
@@ -251,9 +251,11 @@ spec:
             successThreshold: 1
             timeoutSeconds: 5
           volumeMounts:
+          {{- if not .Values.global.secretsBackend.vault.enabled }}
           - name: certs
-            mountPath: /etc/connect-injector/certs
+            mountPath: {{ .Values.connectInject.tlsCertDir }}
             readOnly: true
+          {{- end}}
           - mountPath: /consul/login
             name: consul-data
             readOnly: true
@@ -271,10 +273,12 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
       volumes:
+      {{- if not .Values.global.secretsBackend.vault.enabled }}
       - name: certs
         secret:
           defaultMode: 420
           secretName: {{ template "consul.fullname" . }}-connect-inject-webhook-cert
+      {{- end }}
       - name: consul-data
         emptyDir:
           medium: "Memory"
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -171,7 +171,11 @@ spec:
                 -consul-cross-namespace-acl-policy=cross-namespace-policy \
                 {{- end }}
                 {{- end }}
+                {{- if .Values.global.secretsBackend.vault.enabled }}
                 -tls-cert-dir=/vault/secrets \
+                {{- else }}
+                -tls-cert-dir=/etc/connect-injector/certs \
+                {{- end }}
                 {{- $resources := .Values.connectInject.sidecarProxy.resources }}
                 {{- /* kindIs is used here to differentiate between null and 0 */}}
                 {{- if not (kindIs "invalid" $resources.limits.memory) }}
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -44,10 +44,10 @@ spec:
         "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulCARole }}
         "vault.hashicorp.com/agent-inject-secret-serverca.crt": {{ .Values.global.tls.caCert.secretName }}
         "vault.hashicorp.com/agent-inject-template-serverca.crt": {{ template "consul.serverTLSCATemplate" . }}
-        "vault.hashicorp.com/agent-inject-secret-injectwebhookcert.crt": {{ .Values.connectInject.tlsCert.secretName }}
-        "vault.hashicorp.com/agent-inject-template-injectwebhookcert.crt": {{ include "consul.connectInjectWebhookTLSCertTemplate" . }}
-        "vault.hashicorp.com/agent-inject-secret-injectwebhookcert.key": {{ .Values.connectInject.tlsCert.secretName }}
-        "vault.hashicorp.com/agent-inject-template-injectwebhookcert.key": {{ include "consul.connectInjectWebhookTLSCertTemplate" . }}
+        "vault.hashicorp.com/agent-inject-secret-tls.crt": {{ .Values.connectInject.tlsCert.secretName }}
+        "vault.hashicorp.com/agent-inject-template-tls.crt": {{ include "consul.connectInjectWebhookTLSCertTemplate" . }}
+        "vault.hashicorp.com/agent-inject-secret-tls.key": {{ .Values.connectInject.tlsCert.secretName }}
+        "vault.hashicorp.com/agent-inject-template-tls.key": {{ include "consul.connectInjectWebhookTLSKeyTemplate" . }}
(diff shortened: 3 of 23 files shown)


acceptance/tests/vault/vault_webhook_certs_test.go
```diff
@@ -0,0 +1,294 @@
...
+		KubernetesNamespace: ns,
+		DataCenter:          "dc1",
+		ServiceAccountName:  fmt.Sprintf("%s-consul-%s", consulReleaseName, "server"),
+		AllowedSubdomain:    fmt.Sprintf("%s-consul-%s", consulReleaseName, "server"),
+		MaxTTL:              "1h",
+		AuthMethodPath:      "kubernetes",
+	}
+	serverPKIConfig.ConfigurePKIAndAuthRole(t, vaultClient)
+
+	// Configure controller webhook PKI
```
jmurret, line 64: This test is the same as our original `TestVault()` except we set up the controller and connectInjector webhook PKI here and configure it in the helm values.

acceptance/tests/vault/vault_webhook_certs_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+	consulHelmValues := map[string]string{
+		"server.extraVolumes[0].type": "secret",
+		"server.extraVolumes[0].name": vaultCASecret,
+		"server.extraVolumes[0].load": "false",
+
+		"connectInject.enabled":  "true",
+		"connectInject.replicas": "1",
+		"controller.enabled":     "true",
+		"global.secretsBackend.vault.connectInject.tlsCert.secretName": connectInjectorWebhookPKIConfig.CertPath,
```
jmurret, line 191: these 4 lines are the added configuration to turn off web cert manager and update the certs and CA bundle in the controller and connect injector directly.

acceptance/tests/vault/vault_webhook_certs_test.go
```diff
@@ -0,0 +1,294 @@
...
+		"global.secretsBackend.vault.connectInject.caCert.secretName":  connectInjectorWebhookPKIConfig.CAPath,
+		"global.secretsBackend.vault.controller.tlsCert.secretName":    controllerWebhookPKIConfig.CertPath,
```
jmurret, line 193: these two lines are the last two additions.  These roles have role access to the vault path where the ca bundle and the tls certsare.

control-plane/subcommand/webhook-cert-manager/command.go
```diff
@@ -3,7 +3,6 @@ package webhookcertmanager
 import (
```
jmurret, line 3: Update webhook-cert-manager to use new comon code in `mutatingwebhookconfiguration.UpdateWithCABundle()`

control-plane/subcommand/inject-connect/command.go
```diff
@@ -34,19 +35,20 @@ import (
...
+	flagDefaultInject         bool   // True to inject by default
+	flagConsulImage           string // Docker image for Consul
+	flagEnvoyImage            string // Docker image for Envoy
+	flagConsulK8sImage        string // Docker image for consul-k8s
+	flagACLAuthMethod         string // Auth Method to use for ACLs, if enabled
+	flagWriteServiceDefaults  bool   // True to enable central config injection
+	flagDefaultProtocol       string // Default protocol for use with central config
+	flagConsulCACert          string // [Deprecated] Path to CA Certificate to use when communicating with Consul clients
+	flagEnvoyExtraArgs        string // Extra envoy args when starting envoy
+	flagEnableWebhookCAUpdate bool
```
jmurret, line 49: - Expose this `enable-webhook-ca-update` which basically says "webhook-cert-manager is turned off so we are going to configure webhook certs directly here
- it will use common code in `mutatingwebhookconfiguration.UpdateWithCABundle()` and will use the `ca.crt` in the `flagCertDir` directory

control-plane/subcommand/controller/command.go
```diff
@@ -1,23 +1,28 @@
 package controller
 
 import (
+	"context"
```
jmurret, line 4: Changes in here:

- Expose this `enable-webhook-ca-update` which basically says "webhook-cert-manager is turned off so we are going to configure webhook certs directly here
- it will use common code in `mutatingwebhookconfiguration.UpdateWithCABundle()` and will use the `ca.crt` in the `flagWebhookTLSCertDir ` directory

control-plane/helper/mutating-webhook-configuration/mutating_webhook_configuration.go
```diff
@@ -0,0 +1,51 @@
+package mutatingwebhookconfiguration
```
jmurret, line 1: this is repeated code that existed in and has been refactored out to here:
- control-plane/subcommand/controller/command.go
- control-plane/subcommand/inject-connect/command.go
- control-plane/subcommand/webhook-cert-manager/command.go

charts/consul/test/unit/webhook-cert-manager-clusterrole.bats
```diff
@@ -40,6 +40,93 @@ load _helpers
   [ "${actual}" = "true" ]
 }
 
+#--------------------------------------------------------------------
+# rules
+
```
jmurret, line 45: added some missing tests

charts/consul/templates/webhook-cert-manager-clusterrole.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if or .Values.connectInject.enabled .Values.controller.enabled}}
```
jmurret, line 1: For all webhook-cert-manager helm resources, we change this to not render when the vault is enable and the following are provided:

- global.secretsBackend.vault.enabled, .global.secretsBackend.vault.enabled
- global.secretsBackend.vault.consulConnectInjectCARole
- global.secretsBackend.vault.connectInject.tlsCert.secretName
- global.secretsBackend.vault.connectInject.caCert.secretName
- global.secretsBackend.vault.consulControllerCARole
- global.secretsBackend.vault.controller.tlsCert.secretName
- global.secretsBackend.vault.controller.caCert.secretName

Since webhook cert manager manages webhook certs for controller and conect injector all of the above need to be configured together to replace the functionality.  Then webhook certs will be managed locally within controller and connect injector.

charts/consul/templates/_helpers.tpl
```diff
@@ -22,13 +22,17 @@ as well as the global.name setting.
             {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
 
-{{- define "consul.serverTLSCATemplate" -}}
+{{- define "consul.vaultCATemplate" -}}
```
jmurret, line 25: extracted `serverTLSCATemplate` to a more generic name. 

modified `serverTLSCATemplate` to call `vaultCATemplate` because there are lots of existing references that would each require a few line code change.

charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -33,6 +33,17 @@ rules:
   - get
   - list
   - update
+{{- if (and .Values.global.secretsBackend.vault.enabled .Values.global.secretsBackend.vault.consulConnectInjectCARole .Values.global.secretsBackend.vault.connectInject.tlsCert.secretName  .Values.global.secretsBackend.vault.connectInject.caCert.secretName)}}
+- apiGroups:
```
jmurret, line 37: k8s permissions needed when web-cert-manager is turned off and connect injector manages webhook certs itself with vault.

charts/consul/test/unit/connect-inject-clusterrole.bats
```diff
@@ -29,6 +29,94 @@ load _helpers
       .
 }
 
+#--------------------------------------------------------------------
+# rules
+
+@test "connectInject/ClusterRole: sets get, list, and watch access to pods, endpoints, services, and namespaces in all api groups" {
```
jmurret, line 35: added these two tests that were missing from existing code, but are unrelated to webhook certs.

charts/consul/test/unit/controller-clusterrole.bats
```diff
@@ -18,6 +18,149 @@ load _helpers
       yq 'length > 0' | tee /dev/stderr)
   [ "${actual}" = "true" ]
 }
+#--------------------------------------------------------------------
+# rules
+
+@test "controller/ClusterRole: sets create, delete, get, list, patch, update and watch access to all CRDs in the consul.hashicorp.com api group" {
```
jmurret, line 24: added these three tests that were missing from existing code, but are unrelated to webhook certs.

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -41,9 +42,25 @@ spec:
...
         "vault.hashicorp.com/agent-inject": "true"
+        {{- if .Values.global.secretsBackend.vault.consulConnectInjectCARole }}
+        "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulConnectInjectCARole }}
+        {{ else }}
         "vault.hashicorp.com/role": {{ .Values.global.secretsBackend.vault.consulCARole }}
+        {{ end }}
         "vault.hashicorp.com/agent-inject-secret-serverca.crt": {{ .Values.global.tls.caCert.secretName }}
         "vault.hashicorp.com/agent-inject-template-serverca.crt": {{ template "consul.serverTLSCATemplate" . }}
+        {{- if .Values.global.secretsBackend.vault.connectInject.caCert.secretName }}
+        {{- with .Values.global.secretsBackend.vault.connectInject.caCert }}
```
jmurret, line 53: we need the CA Cert from the PKI we setup so we can update the mutating webhook before connect injector starts running.

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -252,9 +274,11 @@ spec:
             successThreshold: 1
             timeoutSeconds: 5
           volumeMounts:
+          {{- if not (and .Values.global.secretsBackend.vault.enabled .Values.global.secretsBackend.vault.connectInject.tlsCert.secretName) }}
```
jmurret, line 277: don't need to mount these when using vault because they are used in webhook-cert-manager mode.  vault downloads the certs to `/vault/secrets`.

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -41,9 +42,25 @@ spec:
         {{- if (and .Values.global.secretsBackend.vault.enabled .Values.global.tls.enabled) }}
         "vault.hashicorp.com/agent-init-first": "true"
         "vault.hashicorp.com/agent-inject": "true"
+        {{- if .Values.global.secretsBackend.vault.consulConnectInjectCARole }}
```
jmurret, line 45: we have a separate role because someone can choose to mount PKI for connect injector webhooks at different path than the connect CA PKI

charts/consul/values.yaml
```diff
@@ -263,6 +263,46 @@ global:
...
+        # deletion, and update, to get CA certificates used issued from vault
+        # to send webhooks to the controller.
+        caCert:
+          # @type: string
+          secretName: null
+      
+      # The Vault role for consul controller to read the Consul's controllers's CA Certificate (unauthenticated).
+      # A Vault policy must be created which grants read capabilities to 
+      # `global.secretsBackend.vault.controller.tlsCert.secretName`.
+      consulControllerCARole: ""
```
jmurret, line 284: I'd prefer to put this role inside of the `controller` stanza above and name it something more succinct.  I left this outside of that stanza because all of the previously defined roles for like `consulCARole`,`adminPartitionsRole`, manageSystemACLsRole`, etc are at this level.

Same for `consulConnectInjectCARole` below.

acceptance/tests/vault/vault_webhook_certs_test.go
```diff
@@ -0,0 +1,307 @@
...
+
+	if cfg.EnableEnterprise {
+		// Validate that the enterprise license is set correctly.
+		logger.Log(t, "Validating the enterprise license has been set correctly.")
+		license, licenseErr := consulClient.Operator().LicenseGet(nil)
+		require.NoError(t, licenseErr)
+		require.True(t, license.Valid)
+	}
+
+	// Check that webhook-cert-manager is not deployed.
```
jmurret, line 282: This validation that webhook-cert-manager is not running is also an additional check on top of what `TestVault()`is doing

charts/consul/templates/webhook-cert-manager-clusterrole.yaml
```diff
@@ -1,4 +1,4 @@
...
+{{- if (and (or .Values.connectInject.enabled .Values.controller.enabled) (not (and .Values.global.secretsBackend.vault.enabled .Values.global.secretsBackend.vault.consulConnectInjectCARole .Values.global.secretsBackend.vault.connectInject.tlsCert.secretName .Values.global.secretsBackend.vault.connectInject.caCert.secretName .Values.global.secretsBackend.vault.consulControllerCARole .Values.global.secretsBackend.vault.controller.tlsCert.secretName  .Values.global.secretsBackend.vault.controller.caCert.secretName))) }}
```
ishustava, line 1: might be worth extracting the second argument to `and` into a variable for readability.

charts/consul/test/unit/connect-inject-clusterrole.bats
```diff
@@ -29,6 +29,94 @@ load _helpers
...
+  [ "${actual}" != null ]
+
+  local actual=$(echo $object | yq -r '.verbs | index("list")' | tee /dev/stderr)
+  [ "${actual}" != null ]
+
+  local actual=$(echo $object | yq -r '.verbs | index("update")' | tee /dev/stderr)
+  [ "${actual}" != null ]
+}
+
+@test "connectInject/ClusterRole: sets get access to serviceaccounts when manageSystemACLSis true" {
```
ishustava, line 99:
```suggestion
@test "connectInject/ClusterRole: sets get access to serviceaccounts when manageSystemACLS is true" {
```

charts/consul/test/unit/connect-inject-clusterrole.bats
```diff
@@ -53,3 +141,48 @@ load _helpers
       yq -r '.rules | map(select(.resources[0] == "podsecuritypolicies")) | length' | tee /dev/stderr)
   [ "${actual}" = "1" ]
 }
+
+#--------------------------------------------------------------------
+# vault
+
+@test "connectInject/ClusterRole: vault sets get, list, watch, and patch access to mutatingwebhookconfigurationswhen the following are configured - global.secretsBackend.vault.enabled, global.secretsBackend.vault.consulConnectInjectCARole, global.secretsBackend.vault.connectInject.tlsCert.secretName, and global.secretsBackend.vault.connectInject.caCert.secretName." {
```
ishustava, line 148:
```suggestion
@test "connectInject/ClusterRole: sets get, list, watch, and patch access to mutatingwebhookconfigurations when the following are configured - global.secretsBackend.vault.enabled, global.secretsBackend.vault.consulConnectInjectCARole, global.secretsBackend.vault.connectInject.tlsCert.secretName, and global.secretsBackend.vault.connectInject.caCert.secretName." {
```

charts/consul/test/unit/connect-inject-deployment.bats
```diff
@@ -1813,6 +1813,46 @@ EOF
...
+      --set 'connectInject.enabled=true' \
+      --set 'global.tls.enabled=true' \
+      --set 'global.tls.enableAutoEncrypt=true' \
+      --set 'global.tls.caCert.secretName=foo' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | any(contains("-enable-webhook-ca-update"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "connectInject/Deployment: enable-webhook-ca-update flag is not set on command when using vault" {
```
ishustava, line 1832:
```suggestion
@test "connectInject/Deployment: enable-webhook-ca-update flag is set on command when using vault" {
```

charts/consul/test/unit/controller-clusterrole.bats
ishustava: Typically, we don't add tests for YAML that doesn't have any templating logic around it (the case for these rules I think) because we don't see they add that much value that would justify the effort of writing them (writing bats is kind of tedious). I think it's fine to keep these here though if you found they were valuable!

charts/consul/test/unit/controller-clusterrole.bats
```diff
@@ -43,3 +186,46 @@ load _helpers
       yq '.rules | map(select(.resources[0] == "podsecuritypolicies")) | length' | tee /dev/stderr)
   [ "${actual}" = "1" ]
 }
+
+#--------------------------------------------------------------------
+# vault
+
+@test "controller/ClusterRole: vault sets get, list, watch, and patch access to mutatingwebhookconfigurationswhen the following are configured - global.secretsBackend.vault.enabled, global.secretsBackend.vault.consulControllerCARole, global.secretsBackend.vault.controller.tlsCert.secretName, and global.secretsBackend.vault.controller.caCert.secretName." {
```
ishustava, line 193:
```suggestion
@test "controller/ClusterRole: sets get, list, watch, and patch access to mutatingwebhookconfigurations when the following are configured - global.secretsBackend.vault.enabled, global.secretsBackend.vault.consulControllerCARole, global.secretsBackend.vault.controller.tlsCert.secretName, and global.secretsBackend.vault.controller.caCert.secretName." {
```

charts/consul/test/unit/webhook-cert-manager-podsecuritypolicy.bats
```diff
@@ -28,7 +28,7 @@ load _helpers
   [ "${actual}" = "true" ]
 }
 
-@test "webhookCertManager/Configmap: enabled with connectInject.enabled=true, controller.enabled=false and global.enablePodSecurityPolicies=true" {
+@test "webhookCertManager/PodSecurityPolicy: enabled with connectInject.enabled=true, controller.enabled=false and global.enablePodSecurityPolicies=true" {
```
ishustava, line 31: nice, thank you!

charts/consul/values.yaml
```diff
@@ -263,6 +263,46 @@ global:
         additionalConfig: |
           {}
 
+      controller:
+        # The directory that Kubernetes will use on Kubernetes CRD creation, 
+        # deletion, and update, to get TLS certificates used issued from vault
+        # to send webhooks to the controller.
```
ishustava, line 269: Isn't this the vault secret of the webhook cert?

charts/consul/values.yaml
```diff
@@ -263,6 +263,46 @@ global:
...
+        tlsCert:
+          # @type: string
+          secretName: null
```
ishustava, line 272: could you add docs for these too?

charts/consul/values.yaml
```diff
@@ -263,6 +263,46 @@ global:
...
+          
+        # The directory that Kubernetes will use on Kubernetes CRD creation, 
+        # deletion, and update, to get CA certificates used issued from vault
+        # to send webhooks to the controller.
+        caCert:
+          # @type: string
+          secretName: null
+      
+      # The Vault role for consul controller to read the Consul's controllers's CA Certificate (unauthenticated).
```
ishustava, line 281: this role is needed for reading CA as well as webhook cert/private key, correct?

control-plane/subcommand/controller/command.go
```diff
@@ -81,6 +88,10 @@ func (c *Command) init() {
 		"Directory that contains the TLS cert and key required for the webhook. The cert and key files must be named 'tls.crt' and 'tls.key' respectively.")
 	c.flagSet.BoolVar(&c.flagEnableWebhooks, "enable-webhooks", true,
 		"Enable webhooks. Disable when running locally since Kube API server won't be able to route to local server.")
+	c.flagSet.StringVar(&c.flagResourcePrefix, "resource-prefix", "",
+		"Release prefix of the Consul installation used to determine Consul DNS Service name.")
```
ishustava, line 92: A bit confused by the description. Why do we need to determine consul dns service name?

control-plane/subcommand/controller/command.go
```diff
@@ -81,6 +88,10 @@ func (c *Command) init() {
...
+	c.flagSet.BoolVar(&c.flagEnableWebhookCAUpdate, "enable-webhook-ca-update", false,
+		"Enables updating the CABundle on the webhook within this controller rather than using the web cert manager.")
```
ishustava, line 94:
```suggestion
		"Enables updating the CABundle on the webhook within this controller rather than using the webhook-cert-manager.")
```

control-plane/subcommand/controller/command.go
```diff
@@ -329,6 +348,32 @@ func (c *Command) Run(args []string) int {
...
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return err
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	webhookConfigName := fmt.Sprintf("%s-%s", c.flagResourcePrefix, "controller")
```
ishustava, line 364:
```suggestion
	webhookConfigName := fmt.Sprintf("%s-controller", c.flagResourcePrefix)
```

control-plane/subcommand/controller/command.go
```diff
@@ -329,6 +348,32 @@ func (c *Command) Run(args []string) int {
...
+	caPath := fmt.Sprintf("%s/%s", c.flagWebhookTLSCertDir, "ca.crt")
```
ishustava, line 365: should we make `ca.crt` a const?

control-plane/subcommand/controller/command.go
```diff
@@ -329,6 +348,32 @@ func (c *Command) Run(args []string) int {
 	return 0
 }
 
+func (c *Command) configureCABundleUpdate() error {
```
ishustava, line 351: nit: maybe rename this function to `updateWebhookCABundle`? I think `configure` reads a bit counter-intuitive

ishustava (commented):
Looking good so far, John! Thank you so much for doing this work!

I'm leaving my comments in-line.

charts/consul/test/unit/connect-inject-deployment.bats
```diff
@@ -1892,7 +1932,7 @@ EOF
     --set 'global.secretsBackend.vault.enabled=true' \
     --set 'global.secretsBackend.vault.consulClientRole=foo' \
     --set 'global.secretsBackend.vault.consulServerRole=test' \
-    --set 'global.secretsBackend.vault.consulCARole=test' \
+    --set 'global.secretsBackend.vault.consulCARole=test2' \
```
kschoche, line 1935: 👍🏽

charts/consul/templates/connect-inject-deployment.yaml
jmurret: Done.  Thank you.  I also renamed `consulControllerCARole` to `consulControllerRole`.

charts/consul/values.yaml
jmurret: yes.  great call.

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -168,7 +185,12 @@ spec:
                 -consul-cross-namespace-acl-policy=cross-namespace-policy \
                 {{- end }}
                 {{- end }}
+                {{- if and .Values.global.secretsBackend.vault.enabled .Values.global.secretsBackend.vault.connectInject.tlsCert.secretName }}
+                -tls-cert-dir=/vault/secrets \
```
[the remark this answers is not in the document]
jmurret: This has been fixed.  Thanks for the tip.

charts/consul/templates/_helpers.tpl
```diff
@@ -45,6 +49,38 @@ as well as the global.name setting.
             {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
 
+{{- define "consul.connectInjectWebhookTLSCertTemplate" -}}
+ |
+            {{ "{{" }}- with secret "{{ .Values.global.secretsBackend.vault.connectInject.tlsCert.secretName }}" "{{ printf "common_name=connect-injector.%s.%s" .Values.global.datacenter .Values.global.domain }}"
```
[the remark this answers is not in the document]
jmurret: we are not.  The common name currently in use is "Consul Webhook Certificates". I have tried to configure this including setting up the right role permissions in Vault, but still get `* common name Consul Webhook Certificates not allowed by this role` in `vault-agent-init` for both of those pods.  I'll have to revisit this.

charts/consul/templates/_helpers.tpl
```diff
@@ -59,6 +95,18 @@ as well as the global.name setting.
 {{- if .Values.global.tls -}}{{- if .Values.global.tls.serverAdditionalIPSANs -}}{{- range $ipsan := .Values.global.tls.serverAdditionalIPSANs }},{{ $ipsan }} {{- end -}}{{- end -}}{{- end -}}
 {{- end -}}
 
+{{- define "consul.connectInjectorTLSAltNames" -}}
+{{- $name := include "consul.fullname" . -}}
+{{- $ns := .Release.Namespace -}}
+{{ printf "localhost,%s-connect-injector,*.%s-connect-injector,*.%s-connect-injector.%s,%s-connect-injector.%s,*.%s-connect-injector.%s.svc,%s-connect-injector.%s.svc,*.connect-injector.%s.%s" $name $name $name $ns $name $ns $name $ns $name $ns (.Values.global.datacenter ) (.Values.global.domain) }}
```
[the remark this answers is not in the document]
jmurret: Done.  🙏

charts/consul/templates/_helpers.tpl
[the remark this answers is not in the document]
jmurret: nvm. this is working.

kschoche (approved):
This looks good, thanks for all of the great updates and tests.
I've left a few comments here and there and I think we're good to merge once they're addressed.
One more thought, does it make sense to move the controller test which uses vault into the vault test package, so far everything related to Vault has resided there.

control-plane/subcommand/inject-connect/command.go
```diff
@@ -464,14 +470,34 @@ func (c *Command) Run(args []string) int {
...
+		err := c.updateWebhookCABundle(ctx)
+		if err != nil {
+			setupLog.Error(err, "problem getting CA Cert")
+			return 1
+		}
+	}
+
 	if err := mgr.Start(ctx); err != nil {
 		setupLog.Error(err, "problem running manager")
-		return 1
```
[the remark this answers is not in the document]
jmurret: oh, nice call.  I did not see that was removed. 🙏

control-plane/subcommand/inject-connect/command.go
```diff
@@ -464,14 +470,34 @@ func (c *Command) Run(args []string) int {
 			ConsulAPITimeout:              c.http.ConsulAPITimeout(),
 		}})
 
+	if c.flagEnableWebhookCAUpdate {
+		err := c.updateWebhookCABundle(ctx)
```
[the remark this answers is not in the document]
jmurret: The Run function does not have many tests around except flag validation.  I tried to add tests, but the fact that we set LeaderElection to true causes the leader election code to get the namespace from the `/var/run/secrets/kubernetes.io/serviceaccount/namespace` in [this section](https://github.com/kubernetes-sigs/controller-runtime/blob/f46919744bee01060c9084a285e049afffd38c9d/pkg/leaderelection/leader_election.go#L115) that fails in tests.  I have tried creating this file during the unit test but creating anything in `/var/run` fails with secrets errors.

Some options:
- create an `enable-leader-election` flag that defaults to true and we can set it to false in tests
- add `leaderElection` to the command similar to `clientSet` and do similarly as above where we set it to false in tests
- make `updateWebhookCABundle()` public and test that.  Don't worry about the logic that just looks at the flag and calls this when it is true.
- do nothing.  (Related to the last option above making `updateWebhookCABundle()` public.  It just reads a file and calls `mutatingwebhookconfiguration.UpdateWithCABundle()` which already is tested pretty well. )

Thoughts?

charts/consul/templates/_helpers.tpl
```diff
@@ -45,6 +49,38 @@ as well as the global.name setting.
             {{ "{{" }}- end -{{ "}}" }}
 {{- end -}}
 
+{{- define "consul.connectInjectWebhookTLSCertTemplate" -}}
+ |
+            {{ "{{" }}- with secret "{{ .Values.global.secretsBackend.vault.connectInject.tlsCert.secretName }}" "{{ printf "common_name=connect-injector.%s.%s" .Values.global.datacenter .Values.global.domain }}"
+            "alt_names={{ include "consul.connectInjectorTLSAltNames" . }}" "ip_sans=127.0.0.1" -{{ "}}" }}
```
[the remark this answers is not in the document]
ishustava: I still see the localhost IP SAN.

ishustava (commented):
I started reviewing but for some of my previous comments, I see that the code is the same so not if you're keeping is as or if you meant to accept the suggestion. I'm leaving my comments so far, but let me know if I misunderstood.

control-plane/subcommand/inject-connect/command.go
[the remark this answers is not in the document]
jmurret: I actually tried option #2 above and made changes that look a lot like `changing my production code to add in test code` which I have mixed feelings about.  Once it got to the point that the Controller manager was started, the execution of the test just hung as this separate process ran.  It never came back and the test timed out.  I feel like this may not be worth the effort to make sure that mutatingwebhookconfiguration.UpdateWithCABundle() gets called since we have unit tests that assert that works properly and we have acceptance tests that this works as a whole.

charts/consul/templates/_helpers.tpl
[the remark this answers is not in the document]
jmurret: There is some logic in Vault that seems to tie `allowed_domains` with `common_name`.  I have set up a policy for `Consul Webhook Certificates Service` but still get this error.  It feels like there may be some logic that expects `common_name` to be something like the top level domain and interplays with `allowed_domains`.  A couple questions:
- Is `Consul Webhook Certificates Service` truly what is used as the common name with cert manager?
- If so, do we want to continue to use that across both controller and connect-injector?

Basically, what is our ideal common name pattern here?  It feels like I am running into a Vault configuration I need to set, so I want to make sure I trying to make this work with the ideal common name we are striving for because something that does not conform to a domain like pattern is going to be harder to troubleshoot.  So I just want to confirm and avoid unnecessary pain if possible.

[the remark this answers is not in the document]
ishustava: I think I might have mentioned it somewhere but could we make common name to be the name of the kubernetes service for the webhook. This will also be in the SANs and so should be allowed in the `allowed_domains`.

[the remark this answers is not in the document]
jmurret: Thank you.  Sorry I missed that.

I've made the name to match these two entries that are in the webhook-cert-manager configmap, since we changed the sans to match those in a previous comment.  
https://github.com/hashicorp/consul-k8s/blob/aa05d3db2158e0887f49a0f5354893603ff6b6f2/charts/consul/templates/webhook-cert-manager-configmap.yaml#L20
and
https://github.com/hashicorp/consul-k8s/blob/aa05d3db2158e0887f49a0f5354893603ff6b6f2/charts/consul/templates/webhook-cert-manager-configmap.yaml#L32

is this ok?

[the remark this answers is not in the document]
jmurret: Also, what is the reason that we make the sans and common name for server TLS include datacenter in it and not other certs like this?  Do we have concerns with mesh fed for server tls that does not apply to these other certs?

charts/consul/values.yaml
```diff
@@ -263,6 +275,44 @@ global:
...
+        # deletion, and update, to get CA certificates used issued from vault
+        # to send webhooks to the controller.
+        caCert:
+          # The Vault secret path that contains the CA certificate for controller
+          # webhooks.
+          # @type: string
+          secretName: null
+      
+      connectInject:
+        # The directory that Kubernetes will use on Kubernetes CRD creation,
```
ishustava, line 298: why is it the derectory? Isn't it just pointing to the secret?

charts/consul/values.yaml
```diff
@@ -263,6 +275,44 @@ global:
...
+        # to send webhooks to ConnectInject
+        caCert: 
+          # The Vault secret path that contains the CA certificate for
+          # Connect Inject webhooks.
+          # @type: string
+          secretName: null
+        
+        # The directory that Kubernetes will use on Kubernetes CRD creation, 
+        # deletion, and update, to get TLS certificates used issued from vault
+        # to send webhooks to Connect Inject
```
ishustava, line 309: This description also doesn't seem accurate. This value is for the secret, correct?

control-plane/subcommand/controller/command.go
```diff
@@ -81,6 +90,10 @@ func (c *Command) init() {
 		"Directory that contains the TLS cert and key required for the webhook. The cert and key files must be named 'tls.crt' and 'tls.key' respectively.")
 	c.flagSet.BoolVar(&c.flagEnableWebhooks, "enable-webhooks", true,
 		"Enable webhooks. Disable when running locally since Kube API server won't be able to route to local server.")
+	c.flagSet.StringVar(&c.flagResourcePrefix, "resource-prefix", "",
+		"Release prefix of the Consul installation used to prepend on the webhook name that will have its CA bundle updated.")
+	c.flagSet.BoolVar(&c.flagEnableWebhookCAUpdate, "enable-webhook-ca-update", false,
+		"Enables updating the CABundle on the webhook within this controller rather than using the web cert manager.")
```
ishustava, line 96:
```suggestion
		"Enables updating the CABundle on the webhook within this controller rather than using the webhook-cert-manager.")
```

acceptance/tests/vault/vault_test.go
```diff
@@ -210,6 +257,21 @@ func TestVault(t *testing.T) {
...
+	client := environment.KubernetesClientFromOptions(t, kubectlOptions)
+	podList, err := client.CoreV1().Pods(kubectlOptions.Namespace).List(context.Background(),
+		metav1.ListOptions{LabelSelector: fmt.Sprintf("app=consul,component=connect-injector,release=%s", consulReleaseName)})
+	require.NoError(t, err)
+	require.NotEmpty(t, podList.Items)
+	connectInjectorPodName := podList.Items[0].Name
+	connectInjectorPodAddress := consulCluster.CreatePortForwardTunnelToResourcePort(t, connectInjectorPodName, 8080)
+	connectInjectorCert, err := getCertificate(t, connectInjectorPodAddress)
+	require.NoError(t, err)
+	logger.Logf(t, "RPC expiry: %s \n", connectInjectorCert.NotAfter.String())
```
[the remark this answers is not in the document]
jmurret: ahh, its copy and paste code from the auto config reload where we test rpc, https,etc.  I'll fix this.  🙏

charts/consul/values.yaml
jmurret: Let me see.  I'm pretty sure I copied an existing description.  I can fix this.

jmurret: changed.  🙏

charts/consul/values.yaml
jmurret: done

charts/consul/values.yaml
```diff
@@ -195,6 +195,24 @@ global:
...
+      # CA and issue a certificate and private key.
+      # A Vault policy must be created which grants issue capabilities to 
+      # `global.secretsBackend.vault.controller.tlsCert.secretName`.
+      consulControllerRole: ""
+
+      # The Vault role to read Consul connect-injector webhook's CA
+      # and issue a certificate and private key.
+      # A Vault policy must be created which grants issue capabilities to 
+      # `global.secretsBackend.vault.connectInject.tlsCert.secretName`.
+      consulConnectInjectRole: ""
```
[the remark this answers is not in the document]
jmurret: I've changed this to drop the prefix.

PR #1195, opened 2022-04-29
Enable configuration of server readinessProbe

rescDev:
Problem statement to better understand the use case:
While trying to deploy the consul helm chart on GKE and configuring the Ingress for incoming traffic, the health check path of the GCP Load Balancer backend was set to `/` because no path is specified by default (only `.exec.command`). The health check never succeeded because of this and the LB backend has been marked as unhealthy so it could not serve any traffic. This was due to the fact that when accessing `/` from outside, Consul does some redirects so calls to `/` do not return an HTTP 200.

GCP LB backends can derive the health check path from the Kubernetes manifest if it is set so I thought it would make sense to make this configurable. I also (for no particular reason) made all other `readinessProbe` attributes configurable.

Let me know if this makes sense to you. This is my first contribution/suggestion to a Helm chart, so please also let me know if I missed something

Changes proposed in this PR:
- Being able to overwrite defaults for the `readinessProbe` of the server pods

How I've tested this PR:
unit tests

skip them ;)

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


rescDev:
hey, is there something missing in this PR ? I would also be fine if I'm told that this feature is not wanted (for now or at all), just let me know. Otherwise I would really appreciate a review :)

lkysow:
Hi, we use the readiness probe at /status/leader to ensure that during server roll out, a new server is elected leader before shutting down the subsequent server.

If we were to merge this change, then users could unexpectedly have all of their Consul servers down without a leader being elected during a rollout, so I don't think we can safely merge it as is.

Is there another way to configure the GCP health checks?

rescDev:
Hi @lkysow , thanks for your answer. This sounds reasonable, I did not think of that :)

Yes, there's a way to configure this for GCP, so in case someone follows this thread: you can use GKE BackendConfig resources for this: https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#direct_health

I will close the PR

PR #1211, opened 2022-05-05
Remove timeouts from EKS vpc tf module definition

jmurret:
Background:
- I added this timeouts section last night when we had a nightly EKS job fail when trying to destroy and it timeout on destroying the VPC.
- It turns out that this module does not expose timeouts
- Also the [aws_vpc resource does not either](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/main.tf#L20) compared to [others that do](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/41da6881e295ff5e94bbf97b41018e7c550c7285/main.tf#L148).

Changes proposed in this PR:
-  Removing timeouts as it is not a valid option

How I've tested this PR:
- re-run tf apply locally

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -919,6 +919,7 @@ workflows:
           ARCH: "amd64"
           name: build-distros-amd64
       - nightly-dev-upload-docker:
+          context: consul-ci
           requires:
             - build-distros-amd64
       - cleanup-gcp-resources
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -919,6 +919,7 @@ workflows:
           ARCH: "amd64"
           name: build-distros-amd64
       - nightly-dev-upload-docker:
+          context: consul-ci
```
curtbushko, line 922: This will cause a failure. Context isn't allowed in scheduled jobs.

kschoche (approved):

PR #1215, opened 2022-05-06
Exempt ACL Bootstrap logic from`global.consulAPITimeout` by giving it a 5 timeout

jmurret:
Changes proposed in this PR:
- give httpclient that calls bootstrap ACLs a 5 minute timeout since this is one area where we have to wait and we can't retry.
- our nightly acceptance tests have gone from basically green to close to a dozen failures per cloud provider.

How I've tested this PR:
- CI runs of acceptance tests

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/subcommand/server-acl-init/servers.go
+++ b/control-plane/subcommand/server-acl-init/servers.go
@@ -72,6 +72,14 @@ func (c *Command) bootstrapACLs(firstServerAddr string, scheme string, bootToken
 		Address: c.flagConsulTLSServerName,
 		CAFile:  c.flagConsulCACert,
 	}
+	// Exempting this particular use of the http client from using global.consulAPITimeout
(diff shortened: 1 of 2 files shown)


curtbushko (approved):
Code looks good to me if the reasoning is correct.

control-plane/subcommand/server-acl-init/servers.go
```diff
@@ -70,6 +72,9 @@ func (c *Command) bootstrapACLs(firstServerAddr string, scheme string, bootToken
 		Address: c.flagConsulTLSServerName,
 		CAFile:  c.flagConsulCACert,
 	}
+	clientConfig.HttpClient = &http.Client{
```
lkysow, line 75: Can you add a comment why?

jmurret: 👍 good call.

PR #1216, opened 2022-05-06
Use discoveryv1 EndpointSlice instead of corev1 Endpoints in the endpoints controller

kschoche:
Changes proposed in this PR:
- Convert the endpoints controller to use `discoveryv1.EndpointSlice` instead of `corev1.Endpoints`
- Change the connect-inject clusterrole to use `EndpointSlice` resources instead of `Endpoints` resources.
- Adds a new MetaKey/Value to each service which holds the name of the ServiceEndpoint that it was registered with. This is used to ensure we do not Reconcile/deregister endpoints of the same Kubernetes service that reside in different EndpointSlices. Ex: a k8s service with 2 endpoint slices and one endpointslice is deleted, we should not deregister the service for the still-valid-endpointslice's service instances.
- Update all of the unit tests accordingly. Much of this involves just building up EndpointSlice's instead, but it also requires us to change a lot of the expected object names since `EndpointSlice` objects do not have the exact same name as their `Service`, which was expected for `Endpoints`. 
- This simply means adding something like `-123456` to the end of the `EndpointSlice` objects Name and then parsing it out in the controller.
Here is an example :
```
demo $ k get endpoints
NAME                                  ENDPOINTS                                                        AGE
test-pa49y1-consul-connect-injector   10.244.0.20:8080,10.244.0.21:8080                                3m19s

demo $ k get svc
NAME                                  TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                                                                   AGE
test-pa49y1-consul-connect-injector   ClusterIP   10.96.139.240   <none>        443/TCP                                                                   3m25s

demo $ k get endpointslice
NAME                                        ADDRESSTYPE   PORTS                        ENDPOINTS                 AGE
test-pa49y1-consul-connect-injector-th6fx   IPv4          8080                         10.244.0.21,10.244.0.20   3m31s

```

How I've tested this PR:
Manual testing, acceptance tests and unit tests.

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	corev1 "k8s.io/api/core/v1"
 	discv1 "k8s.io/api/discovery/v1"
+	discv1beta1 "k8s.io/api/discovery/v1beta1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -59,10 +60,6 @@ const (
 	// exposedPathsStartupPortsRangeStart is the start of the port range that we will use as
 	// the ListenerPort for the Expose configuration of the proxy registration for a startup probe.
 	exposedPathsStartupPortsRangeStart = 20500
-
-	// annotationKubeService is the Label that the EndpointSlice is assigned by Kubernetes that
-	// references it's Kube Service.
-	annotationKubeService = "kubernetes.io/service-name"
 )
 
 type EndpointsController struct {
@@ -1066,7 +1063,7 @@ func (r *EndpointsController) consulNamespace(namespace string) string {
 }
 
 func getServiceNameFromEndpointSlice(endpointSlice discv1.EndpointSlice) string {
-	return endpointSlice.Labels[annotationKubeService]
+	return endpointSlice.Labels[discv1beta1.LabelServiceName]
 }
 
 // hasBeenInjected checks the value of the status annotation and returns true if the Pod has been injected.
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -123,7 +123,7 @@ type EndpointsController struct {
 	context.Context
 }
 
-// Reconcile reads the state of an Endpoints object for a Kubernetes Service and reconciles Consul services which
+// Reconcile reads the state of an EndpointSlice object for a Kubernetes Service and reconciles Consul services which
 // correspond to the Kubernetes Service. These events are driven by changes to the Pods backing the Kube service.
 func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var errs error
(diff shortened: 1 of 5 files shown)


control-plane/connect-inject/endpoints_controller_ent_test.go
```diff
@@ -90,37 +91,41 @@ func TestReconcileCreateEndpointWithNamespaces(t *testing.T) {
 			k8sObjects: func() []runtime.Object {
 				pod1 := createPodWithNamespace("pod1", test.SourceKubeNS, "1.2.3.4", true, true)
 				pod2 := createPodWithNamespace("pod2", test.SourceKubeNS, "2.2.3.4", true, true)
-				endpointWithTwoAddresses := &corev1.Endpoints{
+				endpointslice := &v1.EndpointSlice{
 					ObjectMeta: metav1.ObjectMeta{
-						Name:      "service-created",
-						Namespace: test.SourceKubeNS,
+						Name:      "service-created-123456",
```
kschoche, line 96: Most of the test changes are simply creating an EndpointSlice object instead of Endpoints objects.
Careful attention will also show that the Name of the EndpointSlice object is no longer synonymous with the Kube Service it is serving. This uses the full DNS subdomain name (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).

The controller will then use either the OwnerReference (which is the actual Kube Service), or will parse this dns name in the event that there is no OwnerReference (such as in the case of delete).

control-plane/connect-inject/handler_ent_test.go
```diff
@@ -30,7 +30,7 @@ func TestHandler_MutateWithNamespaces(t *testing.T) {
 
 	basicSpec := corev1.PodSpec{
 		Containers: []corev1.Container{
-			corev1.Container{
+			{
```
kschoche, line 33: linter cleanup

jmurret (changes requested):
This is great work.  I'm glad it came together so nicely.  I did leave one request for change that we get the service name from the `kubernetes.io/service-name` label on the endpointslice rather than parsing the endpointslice name itself.  A link is in the comment.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-multierror"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/discovery/v1"
```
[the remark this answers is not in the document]
kschoche: Yeah totally! I went with the default that golang suggested but not a problem, thanks!

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -122,86 +123,106 @@ type EndpointsController struct {
...
+	// endpointPods holds a set of all pods this endpoint slice object is currently pointing to.
 	// We use this later when we reconcile ACL tokens to decide whether an ACL token in Consul
 	// is for a pod that no longer exists.
 	endpointPods := mapset.NewSet()
 
-	// If the endpoints object has been deleted (and we get an IsNotFound
+	// If the endpoint slice object has been deleted (and we get an IsNotFound
 	// error), we need to deregister all instances in Consul for that service.
 	if k8serrors.IsNotFound(err) {
 		// Deregister all instances in Consul for this service. The function deregisterServiceOnAllAgents handles
```
lkysow, line 147: If my service has 2000 pods then I'll have >= 1 endpoint slice. If I scale that down to 100 pods then I'll have only one endpoint slice. At that point we'll get an event hat the endpoint slice has been deleted and we'll run through this function.

Now we'll attempt to deregister _all_ instances of this service but there's still 100 running.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -122,86 +123,106 @@ type EndpointsController struct {
...
 		// the case where the Consul service name is different from the Kubernetes service name.
-		err = r.deregisterServiceOnAllAgents(ctx, req.Name, req.Namespace, nil)
+		// The following grabs the k8s servicename from the name of the endpointslice by trimming the trailing
+		// hypen text from the endpointslice object name: `my-endpointslice-qxrzz` yields a service name of `my-endpointslice`.
+		reqName := strings.Split(req.Name, "-")
```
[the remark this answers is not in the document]
jmurret: gotcha, yea that makes sense.  not sure I have better ideas on this.  I'd still propose using the label over the OwnerReferences array for the reasons I outlined in that comment as well.

control-plane/connect-inject/endpoints_controller.go
thisisnotashwin: Maybe we should have a meta tags on all the service registrations which maps to the name of the endpoint-slice object? We could use that to determine which service instances to deregister.

control-plane/connect-inject/endpoints_controller_ent_test.go
```diff
@@ -18,18 +18,19 @@ import (
 	"github.com/hashicorp/consul/sdk/testutil"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/discovery/v1"
```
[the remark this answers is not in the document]
thisisnotashwin: `discv1` maybe a reasonable middle ground so that we arent very verbose

control-plane/connect-inject/endpoints_controller.go
lkysow: Is each pod guaranteed to always be on the same ep slice?

abhiroop93:
@kschoche 
Can you confirm if Consul uses the endpointslice controller now? Or is it still using endpoints only?
If so, then how are pods > 1000 tackled in the K8s versions after 1.22?

kschoche:
> @kschoche Can you confirm if Consul uses the endpointslice controller now? Or is it still using endpoints only? If so, then how are pods > 1000 tackled in the K8s versions after 1.22?

HI! Thanks for following up on this. As of right now we still use endpoint objects and not endpoint slices. We have support for endpoint slices in our roadmap. Stay tuned!

abhiroop93:
Understood, thanks @kschoche ! 
I'm guessing the new topology-aware hints will also follow after that?
Ref: https://kubernetes.io/docs/concepts/services-networking/topology-aware-hints/

abhiroop93:
@kschoche 
Do you know if consul supports zone/topology aware routing?
Ref: https://kubernetes.io/docs/concepts/services-networking/topology-aware-hints/

abhiroop93:
Any clues if consul supports topology/zone aware hints/routing?
Could not find any documentation on that.
Ref: https://kubernetes.io/docs/concepts/services-networking/topology-aware-hints/

kschoche:
Hi @abhiroop93  - sorry for the delayed response, I didn't notice your follow-up question!
Right now we do not have support for topology aware routing but plan on doing this in the future!

PR #1218, opened 2022-05-10
Revert 1.12 TLS config to ensure 1.11 image is also supported.

thisisnotashwin:
- Revert this commit when we want to stop supporting 1.11 style TLS
  config.

Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


jmurret (approved):
👍 Nice work ashwin!

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 BREAKING CHANGES:
 * Helm
-  * Using the Vault integration requires Consul 1.12.0+. [[GH-1213](https://github.com/hashicorp/consul-k8s/pull/1213)]
+  * Using the Vault integration requires Consul 1.12.0+. [[GH-1213](https://github.com/hashicorp/consul-k8s/pull/1213)], [[GH-1218](https://github.com/hashicorp/consul-k8s/pull/1218)]
```
kschoche, line 5: Sorry not trying to be nitpicky here, but should we specify if this is Vault as the connect-ca, or the vault secrets backend or both?

charts/consul/templates/server-config-configmap.yaml
```diff
@@ -86,33 +86,25 @@ data:
...
-          "cert_file": "/vault/secrets/servercert.crt",
-          "key_file": "/vault/secrets/servercert.key"
-          {{- else }}
-          "ca_file": "/consul/tls/ca/tls.crt",
-          "cert_file": "/consul/tls/server/tls.crt",
-          "key_file": "/consul/tls/server/tls.key"
-          {{- end }}
-        }
-      },
+       {{- if .Values.global.secretsBackend.vault.enabled }}
```
kschoche, line 89: This might be my browser but it looks like the indentation is off for these two blocks. I don't think it makes a functional difference tho with JSON?

CHANGELOG.md
thisisnotashwin: I think most users who are using Vault for either will be ok migrating to 1.12 which is why this was left generic.

kschoche (approved):
Looks great!

PR #1221, opened 2022-05-12
Refactor Vault tests to allow more flexibility in mounting PKI and also be more representative of the integration experience of the user

jmurret:
### Easy to follow configuration paths

Over time, the Vault acceptance test have organically grown where multiple helper functions would encapsulate what should be shared knowledge.  For example, all of the tests have helm chart configurations that look like this:

```
"global.secretsBackend.vault.connectCA.rootPKIPath":         "connect_root",
"global.secretsBackend.vault.connectCA.intermediatePKIPath": "dc1/connect_inter",
"global.acls.bootstrapToken.secretName": "consul/data/secret/bootstrap",
"global.acls.bootstrapToken.secretKey":  "token",
"global.gossipEncryption.secretName":    "consul/data/secret/gossip",
"global.gossipEncryption.secretKey":     "gossip",
"global.tls.caCert.secretName": "pki/cert/ca",
```
For each of these, it can be confusing to know where and how this get set previously to arriving here.  It is also easy to misconfigure and can lead to taking longer to troubleshoot.  For the above, you would would have to know:
- rootPKIPath and intermediatePKIPath were set up and encapsulated inside the `vault.CreateConnectCAPolicy(t, vaultClient, "dc1")` call
- bootrapToken secret name and key were setup and encapsulated inside the `vault.ConfigureACLTokenVaultSecret(t, vaultClient, "bootstrap")` call
- gossipEncryption secret name and key were setup and encapsulated inside the `vault.ConfigureGossipVaultSecret(t, vaultClient)` call
- `pki/cert/ca` was configured when `vault.NewVaultCluster(t, ctx, cfg, vaultReleaseName, nil)` and the pki engine was mounted three function calls deep and set at the `pki` path

There are more examples when you look at policies and roles. 

The main thing here is that the existing test code is hard to follow where something was originally configured and where it needed to be mapped downstream.

This PR means to make that more explicit.  While more verbose than the previous tests, one can follow when something is set up and when it is later mapped. This is representative of the UX of the user that is configuring this integration.
 
### Setting up the ability to mount multiple PKI engines at different paths
This work came about while working on https://github.com/hashicorp/consul-k8s/pull/1191.  In that PR, we need the ability to configure: 
- a CA for controller webhook certs
- a CA for connect injector webhook certs
and have these CAs be different from the server tls CA.  

In the current helper code this is difficult to achieve.  This work originally started in that PR, but has been moved here as a pre-requisite for that PR since that PR is already 1k+ lines of changes and so is this one.  This was the easiest way to split it up.

## How to read this PR:
- view `acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go` and see the comments I have made there.
- view all of the other tests.  it should be rinse and repeat concepts.
- view `acceptance/framework/vault/helpers.go` last.  The diff is hard to read because so much changed.  Having viewed everything else beforehand, you will get a good sense of the new functions and how they work.

## How I've tested this PR:
- integration tests locally and in CI

👀 

## Checklist:
- [x] Tests added
- ~~[ ] CHANGELOG entry added~~

### DIFF
--- a/acceptance/framework/vault/helpers.go
+++ b/acceptance/framework/vault/helpers.go
@@ -254,3 +254,181 @@ func CreateConnectCAPolicy(t *testing.T, vaultClient *vapi.Client, datacenter st
 		fmt.Sprintf(connectCAPolicyTemplate, datacenter, datacenter))
 	require.NoError(t, err)
 }
+
+// ConfigurePKICerts configures roles in Vault so
+// that controller webhook TLS certificates can be issued by Vault.
+func ConfigurePKICerts(t *testing.T,
+	vaultClient *vapi.Client, baseUrl, allowedSubdomain, roleName, ns, datacenter,
+	maxTTL string) string {
+	allowedDomains := fmt.Sprintf("%s.consul,%s,%s.%s,%s.%s.svc", datacenter,
+		allowedSubdomain, allowedSubdomain, ns, allowedSubdomain, ns)
+	params := map[string]interface{}{
+		"allowed_domains":    allowedDomains,
+		"allow_bare_domains": "true",
+		"allow_localhost":    "true",
+		"allow_subdomains":   "true",
+		"generate_lease":     "true",
+		"max_ttl":            maxTTL,
+	}
+
+	_, err := vaultClient.Logical().Write(
+		fmt.Sprintf("%s/roles/%s", baseUrl, roleName), params)
+	require.NoError(t, err)
+
+	certificateIssuePath := fmt.Sprintf("%s/issue/%s", baseUrl, roleName)
+	policy := fmt.Sprintf(`
+		path %q {
+		capabilities = ["create", "update"]
+		}`, certificateIssuePath)
+
+	// Create the server policy.
+	err = vaultClient.Sys().PutPolicy(roleName, policy)
+	require.NoError(t, err)
+
+	return certificateIssuePath
+}
+
+// ConfigurePKI generates a CA in Vault at a given path with a given policyName.
+func ConfigurePKI(t *testing.T, vaultClient *vapi.Client, baseUrl, policyName, commonName string) {
+	// Mount the PKI Secrets engine at the baseUrl.
+	err := vaultClient.Sys().Mount(baseUrl, &vapi.MountInput{
+		Type:   "pki",
+		Config: vapi.MountConfigInput{},
+	})
+	require.NoError(t, err)
+	// Create root CA to issue Consul server certificates and the `consul-server` PKI role.
+	// See https://learn.hashicorp.com/tutorials/consul/vault-pki-consul-secure-tls.
+	// Generate the root CA.
+	params := map[string]interface{}{
+		"common_name": commonName,
+		"ttl":         "24h",
+	}
+	_, err = vaultClient.Logical().Write(fmt.Sprintf("%s/root/generate/internal", baseUrl), params)
+	require.NoError(t, err)
+
+	policy := fmt.Sprintf(`path "%s/cert/ca" {
+		capabilities = ["read"]
+	  }`, baseUrl)
+	err = vaultClient.Sys().PutPolicy(policyName, policy)
+	require.NoError(t, err)
+}
+
+type KubernetesAuthRoleConfiguration struct {
+	ServiceAccountName  string
+	KubernetesNamespace string
+	PolicyNames         string
+	AuthMethodPath      string
+	RoleName            string
+}
+
+// ConfigureKubernetesAuthRole configures a role in Vault for the component for the Kubernetes auth method
+// that will be used by the test Helm chart installation.
+func ConfigureK8SAuthRole(t *testing.T, vaultClient *vapi.Client, config *KubernetesAuthRoleConfiguration) {
+	// Create the Auth Roles for the component.
+	// Auth roles bind policies to Kubernetes service accounts, which
+	// then enables the Vault agent init container to call 'vault login'
+	// with the Kubernetes auth method to obtain a Vault token.
+	// Please see https://www.vaultproject.io/docs/auth/kubernetes#configuration
+	// for more details.
+	logger.Logf(t, "Creating the %q", config.ServiceAccountName)
+	params := map[string]interface{}{
+		"bound_service_account_names":      config.ServiceAccountName,
+		"bound_service_account_namespaces": config.KubernetesNamespace,
+		"policies":                         config.PolicyNames,
+		"ttl":                              "24h",
+	}
+	_, err := vaultClient.Logical().Write(fmt.Sprintf("auth/%s/role/%s", config.AuthMethodPath, config.RoleName), params)
+	require.NoError(t, err)
+}
+
+type PKIAndAuthRoleConfiguration struct {
+	ServiceAccountName  string
+	BaseURL             string
+	PolicyName          string
+	RoleName            string
+	CommonName          string
+	CAPath              string
+	CertPath            string
+	KubernetesNamespace string
+	DataCenter          string
+	MaxTTL              string
+	AuthMethodPath      string
+	AllowedSubdomain    string
+}
+
+func ConfigurePKIAndAuthRole(t *testing.T, vaultClient *vapi.Client, config *PKIAndAuthRoleConfiguration) {
+	config.CAPath = fmt.Sprintf("%s/cert/ca", config.BaseURL)
+	// Configure role with read access to <baseURL>/cert/ca
+	ConfigurePKI(t, vaultClient, config.BaseURL, config.PolicyName,
+		config.CommonName)
+	// Configure role with create and update access to issue certs at
+	// <baseURL>/issue/<roleName>
+	config.CertPath = ConfigurePKICerts(t, vaultClient, config.BaseURL,
+		config.AllowedSubdomain, config.PolicyName, config.KubernetesNamespace,
+		config.DataCenter, config.MaxTTL)
+	// Configure AuthMethodRole that will map the service account name
+	// to the Vault role
+	authMethodRoleConfig := &KubernetesAuthRoleConfiguration{
+		ServiceAccountName:  config.ServiceAccountName,
+		KubernetesNamespace: config.KubernetesNamespace,
+		AuthMethodPath:      config.AuthMethodPath,
+		RoleName:            config.RoleName,
+		PolicyNames:         config.PolicyName,
+	}
+	ConfigureK8SAuthRole(t, vaultClient, authMethodRoleConfig)
+}
+
+type SaveVaultSecretConfiguration struct {
+	Path       string
+	Key        string
+	PolicyName string
+	Value      string
+}
+
+func SaveSecret(t *testing.T, vaultClient *vapi.Client, config *SaveVaultSecretConfiguration) {
+	policy := fmt.Sprintf(`
+	path "%s" {
+	  capabilities = ["read"]
+	}`, config.Path)
+	// Create the Vault Policy for the gossip key.
+	logger.Log(t, "Creating policy")
+	err := vaultClient.Sys().PutPolicy(config.PolicyName, policy)
+	require.NoError(t, err)
+
+	// Create the gossip secret.
+	logger.Log(t, "Creating the gossip secret")
+	params := map[string]interface{}{
+		"data": map[string]interface{}{
+			config.Key: config.Value,
+		},
+	}
+	_, err = vaultClient.Logical().Write(config.Path, params)
+	require.NoError(t, err)
+}
+
+// CreateConnectCAPolicyForDatacenter creates the Vault Policy for the connect-ca in a given datacenter.
+func CreateConnectCARootAndIntermediatePIKPolicy(t *testing.T, vaultClient *vapi.Client, policyName, rootPath, intermediatePath string) {
+	// connectCAPolicy allows Consul to bootstrap all certificates for the service mesh in Vault.
+	// Adapted from https://www.consul.io/docs/connect/ca/vault#consul-managed-pki-paths.
+	connectCAPolicyTemplate := `
+path "/sys/mounts" {
+  capabilities = [ "read" ]
+}
+path "/sys/mounts/%s" {
+  capabilities = [ "create", "read", "update", "delete", "list" ]
+}
+path "/sys/mounts/%s" {
+  capabilities = [ "create", "read", "update", "delete", "list" ]
+}
+path "/%s/*" {
+  capabilities = [ "create", "read", "update", "delete", "list" ]
+}
+path "/%s/*" {
+  capabilities = [ "create", "read", "update", "delete", "list" ]
+}
+`
+	err := vaultClient.Sys().PutPolicy(
+		policyName,
+		fmt.Sprintf(connectCAPolicyTemplate, rootPath, intermediatePath, rootPath, intermediatePath))
+	require.NoError(t, err)
+}
--- a/acceptance/framework/vault/helpers.go
+++ b/acceptance/framework/vault/helpers.go
@@ -407,7 +407,7 @@ func SaveSecret(t *testing.T, vaultClient *vapi.Client, config *SaveVaultSecretC
 }
 
 // CreateConnectCAPolicyForDatacenter creates the Vault Policy for the connect-ca in a given datacenter.
-func CreateConnectCARootAndIntermediatePIKPolicy(t *testing.T, vaultClient *vapi.Client, policyName, rootPath, intermediatePath string) {
+func CreateConnectCARootAndIntermediatePKIPolicy(t *testing.T, vaultClient *vapi.Client, policyName, rootPath, intermediatePath string) {
 	// connectCAPolicy allows Consul to bootstrap all certificates for the service mesh in Vault.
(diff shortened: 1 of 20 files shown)


acceptance/framework/vault/helpers.go
```diff
@@ -6,63 +6,11 @@ import (
 	"fmt"
 	"testing"
```
jmurret, line 8: This diff is not the easiest to follow unfortunately.  The best way to review this is to first gain an understanding of the vault helper functions are used in the tests.  So look to the test files first and look for `vault.*` function calls.

acceptance/framework/vault/vault_cluster.go
```diff
@@ -147,13 +147,6 @@ func (v *VaultCluster) bootstrap(t *testing.T, vaultNamespace string) {
 	})
 	require.NoError(t, err)
 
-	// Enable the PKI Secrets engine.
-	err = v.vaultClient.Sys().Mount("pki", &vapi.MountInput{
```
jmurret, line 151: This has moved to `ConfigurePKI()` in the above helpers file because we will need to mount more than one PKI engine and it will not always be at `pki` base path.

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
```diff
@@ -40,28 +41,136 @@ func TestSnapshotAgent_Vault(t *testing.T) {
...
+	// -------------------------
+	// Configure Service Mesh CA
+	connectCAPolicy := "connect-ca-dc1"
+	connectCARootPath := "connect_root"
+	connectCAIntermediatePath := "dc1/connect_inter"
+	// Configure Policy for Connect CA
+	vault.CreateConnectCARootAndIntermediatePKIPolicy(t, vaultClient, connectCAPolicy, connectCARootPath, connectCAIntermediatePath)
+
+	// Configure Server PKI
+	serverPKIConfig := &vault.PKIAndAuthRoleConfiguration{
```
jmurret, line 55: this object will get re-used in the auth role creation as well as in helm charts.

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
```diff
@@ -40,28 +41,136 @@ func TestSnapshotAgent_Vault(t *testing.T) {
...
+	}
+	vault.ConfigurePKIAndAuthRole(t, vaultClient, serverPKIConfig)
+
+	// -------------------------
+	// KV2 secrets
+	// -------------------------
+	// Gossip key
+	gossipKey, err := vault.GenerateGossipSecret()
+	require.NoError(t, err)
+	gossipSecret := &vault.SaveVaultSecretConfiguration{
```
jmurret, line 74: for these secrets, we know have a reference to the configuration we will need when we need to also set the `secretName` and `secretKey` in the helm chart.

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
```diff
@@ -75,36 +184,36 @@ func TestSnapshotAgent_Vault(t *testing.T) {
 		"controller.enabled":     "true",
 
 		"global.secretsBackend.vault.enabled":              "true",
-		"global.secretsBackend.vault.consulServerRole":     "server",
```
jmurret, line 78: all strings that can be linked to a previous configuration are.

acceptance/framework/vault/helpers.go
```diff
@@ -80,177 +28,183 @@ func GenerateGossipSecret() (string, error) {
...
+	AuthMethodPath      string
+	RoleName            string
+}
+
 // ConfigureKubernetesAuthRole configures a role in Vault for the component for the Kubernetes auth method
 // that will be used by the test Helm chart installation.
-func ConfigureKubernetesAuthRole(t *testing.T, vaultClient *vapi.Client, consulReleaseName, ns, authPath, component, policies string) {
-	componentServiceAccountName := fmt.Sprintf("%s-consul-%s", consulReleaseName, component)
-
+func ConfigureK8SAuthRole(t *testing.T, vaultClient *vapi.Client, config *KubernetesAuthRoleConfiguration) {
```
thisisnotashwin, line 101: Should this be defined on the config?

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
```diff
@@ -40,28 +41,136 @@ func TestSnapshotAgent_Vault(t *testing.T) {
...
-	vault.ConfigureKubernetesAuthRole(t, vaultClient, consulReleaseName, ns, "kubernetes", "client", "gossip")
-	vault.ConfigureKubernetesAuthRole(t, vaultClient, consulReleaseName, ns, "kubernetes", "server-acl-init", "bootstrap-token")
-	vault.ConfigureKubernetesAuthRole(t, vaultClient, consulReleaseName, ns, "kubernetes", "snapshot-agent", "snapshot-agent-config,license")
-	vault.ConfigureConsulCAKubernetesAuthRole(t, vaultClient, ns, "kubernetes")
 
-	vault.ConfigurePKICA(t, vaultClient)
-	certPath := vault.ConfigurePKICertificates(t, vaultClient, consulReleaseName, ns, "dc1", "1h")
+	// server
+	consulServerRole := "server"
+	vault.ConfigureK8SAuthRole(t, vaultClient, &vault.KubernetesAuthRoleConfiguration{
```
thisisnotashwin, line 125: hmm.. maybe the K8s auth role need not be defined on the struct. it looks like a different one each time.

thisisnotashwin (approved):
This makes the tests very very readable! love the approach!!

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
jmurret: can you elaborate?  do you mean don't have `ConfigureK8SAuthRole()` on the struct?

thisisnotashwin: i think it is fine either way. while combing through the tests, I observed that we had 4-6 different structs per test for the Kubernetes authroleconfig so I could understand not wanting to define it on the struct, but I do prefer defining the method on the struct than accepting it as a param. thanks for doing this John!

jmurret: yea, I see what you mean. i agree your suggestion of moving these things to the structs rather than a broad global vault service is nicer.  thanks for the thoughtful feedback!

acceptance/framework/vault/helpers.go
```diff
@@ -80,177 +28,183 @@ func GenerateGossipSecret() (string, error) {
...
-
+func (config *KubernetesAuthRoleConfiguration) ConfigureK8SAuthRole(t *testing.T, vaultClient *vapi.Client) {
 	// Create the Auth Roles for the component.
 	// Auth roles bind policies to Kubernetes service accounts, which
 	// then enables the Vault agent init container to call 'vault login'
 	// with the Kubernetes auth method to obtain a Vault token.
 	// Please see https://www.vaultproject.io/docs/auth/kubernetes#configuration
 	// for more details.
-	logger.Logf(t, "Creating the %q", componentServiceAccountName)
+	logger.Logf(t, "Creating the %q", config.ServiceAccountName)
```
kschoche, line 108:
```suggestion
	logger.Logf(t, "Creating the Vault Auth Role for %q", config.ServiceAccountName)
```

acceptance/framework/vault/helpers.go
```diff
@@ -80,177 +28,183 @@ func GenerateGossipSecret() (string, error) {
...
-	// Create the Vault Policy for the token.
-	logger.Logf(t, "Creating %s token policy", tokenName)
-	policyName := fmt.Sprintf("%s-token", tokenName)
-	tokenPolicy := fmt.Sprintf(tokenPolicyTemplate, tokenName)
-	err := vaultClient.Sys().PutPolicy(policyName, tokenPolicy)
-	require.NoError(t, err)
-
-	// Generate the token secret.
-	token, err := uuid.GenerateUUID()
+func (config *SaveVaultSecretConfiguration) Save(t *testing.T, vaultClient *vapi.Client) {
```
kschoche, line 164: Could you add some documentation on the `Save` function please?
It is not clear by the name what it does.
My understanding is that it creates a KV2 secret and a policy with read access?
Maybe it could be named something more descriptive 🤷🏽

acceptance/framework/vault/helpers.go
```diff
@@ -80,177 +28,183 @@ func GenerateGossipSecret() (string, error) {
...
+	policy := fmt.Sprintf(`
+	path "%s" {
+	  capabilities = ["read"]
+	}`, config.Path)
+	// Create the Vault Policy for the gossip key.
```
kschoche, line 169:
```suggestion
	// Create a Vault Policy for the secret with read capabilities.
```

acceptance/framework/vault/helpers.go
```diff
@@ -80,177 +28,183 @@ func GenerateGossipSecret() (string, error) {
...
+	logger.Log(t, "Creating policy")
+	err := vaultClient.Sys().PutPolicy(config.PolicyName, policy)
 	require.NoError(t, err)
 
-	// Create the replication token secret.
-	logger.Logf(t, "Creating the %s token secret", tokenName)
+	// Create the gossip secret.
```
kschoche, line 174:
```suggestion
	// Create the KV2 secret.
```

acceptance/framework/vault/helpers.go
```diff
@@ -80,177 +28,183 @@ func GenerateGossipSecret() (string, error) {
...
-	_, err = vaultClient.Logical().Write(fmt.Sprintf("consul/data/secret/%s", tokenName), params)
+	_, err = vaultClient.Logical().Write(config.Path, params)
 	require.NoError(t, err)
-
-	return token
 }
 
-// CreateConnectCAPolicy creates the Vault Policy for the connect-ca in a given datacenter.
-func CreateConnectCAPolicy(t *testing.T, vaultClient *vapi.Client, datacenter string) {
+// CreateConnectCAPolicyForDatacenter creates the Vault Policy for the connect-ca in a given datacenter.
```
kschoche, line 185:
```suggestion
// CreateConnectCARootAndIntermediatePKIPolicy creates the Vault Policy for the connect-ca in a given datacenter.
```

acceptance/framework/vault/helpers.go
```diff
@@ -80,177 +28,183 @@ func GenerateGossipSecret() (string, error) {
...
+}
+path "/%s/*" {
+  capabilities = [ "create", "read", "update", "delete", "list" ]
+}
+`
 	err := vaultClient.Sys().PutPolicy(
-		fmt.Sprintf("connect-ca-%s", datacenter),
-		fmt.Sprintf(connectCAPolicyTemplate, datacenter, datacenter))
+		policyName,
+		fmt.Sprintf(connectCAPolicyTemplate, rootPath, intermediatePath, rootPath, intermediatePath))
```
kschoche, line 208: This is an interesting way to do `Sprintf()`! 
While this is just a nit and you don't need to change anything, IMHO its a bit harder to read than simply filling in the parameters when defining the template variable as I have to go back and fourth jumping lines to figure out where the parameters are being used/insert.

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
```diff
@@ -40,28 +41,141 @@ func TestSnapshotAgent_Vault(t *testing.T) {
 	// Now fetch the Vault client so we can create the policies and secrets.
 	vaultClient := vaultCluster.VaultClient(t)
 
-	vault.CreateConnectCAPolicy(t, vaultClient, "dc1")
+	// -------------------------
+	// PKI
+	// -------------------------
+	// Configure Service Mesh CA
```
kschoche, line 47: I hate to be the grammar police, but all comments need to end with proper punctuation, could you run through and double check them? Thanks!

```suggestion
	// Configure Service Mesh CA.
```

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
```diff
@@ -40,28 +41,141 @@ func TestSnapshotAgent_Vault(t *testing.T) {
...
+	// Snapshot Agent config
+	snapshotAgentConfig := generateSnapshotAgentConfig(t, bootstrapToken)
+	require.NoError(t, err)
+	snapshotAgentConfigSecret := &vault.SaveVaultSecretConfiguration{
+		Path:       "consul/data/secret/snapshot-agent-config",
+		Key:        "config",
+		Value:      snapshotAgentConfig,
+		PolicyName: "snapshot-agent-config",
+	}
+	snapshotAgentConfigSecret.Save(t, vaultClient)
```
kschoche, line 113: OK I'm starting to see how you're using `Save()` .. maybe `WriteKV2Secret` or something like this would be appropriately descriptive if it cannot be used for any other Vault secret type.

acceptance/tests/vault/vault_namespaces_test.go
```diff
@@ -48,26 +49,118 @@ func TestVault_VaultNamespace(t *testing.T) {
...
+		ServiceAccountName:  consulClientServiceAccountName,
+		KubernetesNamespace: ns,
+		AuthMethodPath:      "kubernetes",
+		RoleName:            consulClientRole,
+		PolicyNames:         gossipSecret.PolicyName,
+	}
+	clientAuthRoleConfig.ConfigureK8SAuthRole(t, vaultClient)
+
+	// manageSystemACLs
+	manageSystemACLsRole := "server-acl-init"
```
kschoche, line 144: nit : it might be worth making `const`'s for all of the role names if they are re-used more than once/twice.

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
kschoche: Just to add I suppose this needs to be updated on every file sorry! :(

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
jmurret: 👍 Would it better to just rename the struct from `SaveVaultSecretConfiguration` to `KV2SecretConfiguration` or `KV2Secret`?

acceptance/framework/vault/helpers.go
jmurret: I'll change this.  This is just existing code where the template used to be a const.  So you really had to scroll!

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
jmurret: Is this used in a doc string or something?

kschoche: No I don't think it parsed anywhere but it's part of our code style.

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
kschoche: Sure!

acceptance/framework/vault/helpers.go
kschoche: Ah of course, sorry 🙏

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
jmurret: Done.

acceptance/framework/vault/helpers.go
jmurret: Done.  🙏

acceptance/framework/vault/helpers.go
jmurret: Changed the struct name to `KV2Secret` and changed the function name to `SaveSecretAndAddReadPolicy`. 🙏

acceptance/framework/vault/helpers.go
jmurret: Updated.  🙏

kschoche (approved):
Awesome work @jmurret !

I really like how this came together and how much more readable the tests are now, excellent!

acceptance/tests/vault/vault_namespaces_test.go
jmurret: Done. 🙏

acceptance/tests/snapshot-agent/snapshot_agent_vault_test.go
jmurret: Some context:
This is not a comment to explain the logic as we would see in comments in source code.

This comment in the test file is really just `there is so much manual set up related to vault that I put these headers in here so its easier to follow`.

So proceeding to put periods in this type of comment seems a little mismatched with the original intention of the code style.

I think I'll either move this code to helper functions or something else to break them up...or we revisit the need for an ending period for a comment that is not a sentence and is more of header.

kschoche: Yeha I think that is totally reasonable to not have periods on blocks like this. Just the other comments for the code!

PR #1223, opened 2022-05-13
update the flags library to prevent overflows

kschoche:
This is intended to fix alerts flagged by CodeQL: https://github.com/hashicorp/consul-k8s/security/code-scanning
Whereby we are using `strconv.ParseInt` with 64bits length and then casting it to a 32bit value without error checking.

Changes proposed in this PR:
- Updates `flags_int.go`'s use of `parse_Int` and `parse_Uint` to specify 32bit length for `strconv.ParseUint/Int` so that it will return error on overflow instead of actually overflow.
When `ParseInt` overflows it will return :
`value out of range`
Example:
```
func main() {
	a := "100000000000000000000000000000000000000"
	v, err := strconv.ParseInt(a, 0, 32)
	fmt.Println("%v", err)
}
```
Results in : `strconv.ParseInt: parsing "100000000000000000000000000000000000000": value out of range`

How I've tested this PR:
👀
Should I add a unit test?
 
👀

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/cli/common/flag/flag_int.go
+++ b/cli/common/flag/flag_int.go
@@ -23,7 +23,7 @@ type IntVar struct {
 func (f *Set) IntVar(i *IntVar) {
 	initial := i.Default
 	if v, exist := os.LookupEnv(i.EnvVar); exist {
-		if i, err := strconv.ParseInt(v, 0, 64); err == nil {
+		if i, err := strconv.ParseInt(v, 0, 32); err == nil {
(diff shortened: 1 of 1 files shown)


ishustava (approved):
Thanks for fixing it!

thisisnotashwin (approved):
nice!

PR #1237, opened 2022-05-25
Support new annotated format for upstreams

ndhanushkodi:
Changes proposed in this PR:

The proxy configuration upstreams set during service registration needs to support peer names. This is for explicit upstreams configured via the connect-service-upstreams annotation, not Tproxy.

Currently, upstreams are specified in the following format:
 
```
"[service-name].[service-namespace].[service-partition]:[port]:[optional datacenter], prepared_query:[query name]:[port]"
```

See [processUpstreams](https://github.com/hashicorp/consul-k8s/blob/8ab4d0cd5a0a3514b3f9e1971e6c29d3af5a0d1c/control-plane/connect-inject/endpoints_controller.go#L839) function for detail on how it’s parsed.

The cases for upstream services are:

* Local DC, cross partition:
  * Currently: you’d specify partition
  * With peering: you’d specify peer
* Remote DC, remote partition:
  * Currently: Not possible ATM
  * With peering: you’d specify peer
* Remote DC, default partition:
  * Currently: Possible with WAN fed. You’d specify datacenter
  * With peering: you’d specify peer

We want to allow the currently supported upstream annotation format, for potential migration cases and to not make a breaking change. When you specify a peer for an upstream, you don’t also want to include the partition or datacenter because it’s not needed and could make the upstream ambiguous. So in addition to the currently supported formats, we can have a format where these would be valid:

[service-name].svc:[port]
[service-name].svc.[service-namespace].ns.[service-peer].peer:[port]
[service-name].svc.[service-namespace].ns.[service-partition].ap:[port]
[service-name].svc.[service-namespace].ns.[service-datacenter].dc:[port]

When namespaces/partitions are NOT enabled, these would be valid:
[service-name].svc:[port]
[service-name].svc.[service-peer].peer:[port]
[service-name].svc.[service-dc].dc:[port]

First, we split on “,”. Each upstream will be checked for if it fits the old structure, prepared query structure, or new structure. For the new structure, called the "annotated format" in this PR, once we split the upstream on “.”, if the second part is the string “svc”, we’ll know that it’s the new structure. Only one of peer, ap, or dc should be set, we can return an error otherwise.

How I've tested this PR:
unit tests

👀 
pipeline is red because the base branch is peering-dialer-controller whose tests are red. However this was unit tested on its own and those tests are green.

Checklist:
- [x] Tests added
- [-] CHANGELOG entry added 

### DIFF
--- a/control-plane/connect-inject/endpoints_controller.go
+++ b/control-plane/connect-inject/endpoints_controller.go
@@ -834,6 +834,176 @@ func serviceInstancesForK8SServiceNameAndNamespace(k8sServiceName, k8sServiceNam
 			MetaKeyKubeServiceName, k8sServiceName, MetaKeyKubeNS, k8sServiceNamespace, MetaKeyManagedBy, managedByValue))
 }
 
+// processPreparedQueryUpstream processes an upstream in the format:
+// prepared_query:[query name]:[port].
+func processPreparedQueryUpstream(pod corev1.Pod, rawUpstream string) api.Upstream {
+	var preparedQuery string
+	var port int32
+	parts := strings.SplitN(rawUpstream, ":", 3)
+
+	port, _ = portValue(pod, strings.TrimSpace(parts[2]))
+	preparedQuery = strings.TrimSpace(parts[1])
+	var upstream api.Upstream
+	if port > 0 {
+
+		upstream = api.Upstream{
+			DestinationType: api.UpstreamDestTypePreparedQuery,
+			DestinationName: preparedQuery,
+			LocalBindPort:   int(port),
+		}
+	}
+	return upstream
+}
+
+// processNonAnnotatedUpstream processes an upstream in the format:
+// [service-name].[service-namespace].[service-partition]:[port]:[optional datacenter].
+func (r *EndpointsController) processNonAnnotatedUpstream(pod corev1.Pod, rawUpstream string) (api.Upstream, error) {
+	var datacenter, serviceName, namespace, partition, peer string
+	var port int32
+	var upstream api.Upstream
+
+	parts := strings.SplitN(rawUpstream, ":", 3)
+
+	port, _ = portValue(pod, strings.TrimSpace(parts[1]))
+
+	// If Consul Namespaces or Admin Partitions are enabled, attempt to parse the
+	// upstream for a namespace.
+	if r.EnableConsulNamespaces || r.EnableConsulPartitions {
+		pieces := strings.SplitN(parts[0], ".", 3)
+		switch len(pieces) {
+		case 3:
+			partition = strings.TrimSpace(pieces[2])
+			fallthrough
+		case 2:
+			namespace = strings.TrimSpace(pieces[1])
+			fallthrough
+		default:
+			serviceName = strings.TrimSpace(pieces[0])
+		}
+	} else {
+		serviceName = strings.TrimSpace(parts[0])
+	}
+
+	// parse the optional datacenter
+	if len(parts) > 2 {
+		datacenter = strings.TrimSpace(parts[2])
+
+		// Check if there's a proxy defaults config with mesh gateway
+		// mode set to local or remote. This helps users from
+		// accidentally forgetting to set a mesh gateway mode
+		// and then being confused as to why their traffic isn't
+		// routing.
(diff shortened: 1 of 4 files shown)


control-plane/connect-inject/endpoints_controller.go
```diff
@@ -834,6 +835,168 @@ func serviceInstancesForK8SServiceNameAndNamespace(k8sServiceName, k8sServiceNam
...
+			end := strings.TrimSpace(pieces[5])
+			switch end {
+			case "peer":
+				peer = strings.TrimSpace(pieces[4])
+			case "ap":
+				partition = strings.TrimSpace(pieces[4])
+			case "dc":
+				datacenter = strings.TrimSpace(pieces[4])
+			default:
+				return api.Upstream{}, errors.New(fmt.Sprintf("upstream structured incorrectly: %s", rawUpstream))
```
thisisnotashwin, line 946: I wonder if this should be a part of the Validation logic that we do in the connect webhook? that way, if the label is incorrect we can ensure pod creation fails with an accurate error message.

thisisnotashwin (commented):
This looks great! My only question is should we run this past the team in terms of the new structure. I think as far as the demo is concerned this is perfect. I can work on the RFC after where we formalize this. Great work on this Nitya!!

control-plane/connect-inject/endpoints_controller.go
ndhanushkodi: Yeah, I think in order to do that we'd need to have the entire logic for this function live in the webhook since you can't catch these cases without parsing the entire annotation. I think it'd work, and it'd probably be nice to have the pod creation fail, so do you think we should have the logic in both places? The prepared query case and the "non-annotated" upstreams are just less likely to have parsing issues compared to this one due to the number of pieces to parse.

ndhanushkodi:
For posterity, the decisions are being recorded and reviewed in this [doc](https://docs.google.com/document/d/1j0IgebZGoJz2rymiBnFtQCm0BGtaBvEHEfELo9elGCw/edit#) which will turn into an RFC.

control-plane/connect-inject/endpoints_controller.go
thisisnotashwin: If the two are in the same package, we should be able to extract this behavior in a way that can be reused potentially. I also think it is something we can do down the line as a nice to have. Let's get this merged in and create an asana task for the validation if that works for you as well!

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -834,6 +835,168 @@ func serviceInstancesForK8SServiceNameAndNamespace(k8sServiceName, k8sServiceNam
...
+			DestinationName:      serviceName,
+			Datacenter:           datacenter,
+			LocalBindPort:        int(port),
+		}
+	}
+	return upstream, nil
+
+}
+
+func (r *EndpointsController) processAnnotatedUpstream(pod corev1.Pod, rawUpstream string) (api.Upstream, error) {
```
thisisnotashwin, line 921: Can we copy the comment where you have described the formatting of the annotated upstream and add that above each of these method to describe the "shape" of the upstream that they will each process

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -202,6 +202,164 @@ func TestProcessUpstreams(t *testing.T) {
 		consulNamespacesEnabled bool
 		consulPartitionsEnabled bool
 	}{
+		{
```
thisisnotashwin, line 205: Do we test the error cases here? Should we add tests to validate the behavior on a few error cases?

thisisnotashwin (approved):
Great work on this @ndhanushkodi !! I have left a few comments. Feel free to merge this after you get a chance to look into those and the linting issues. It LGTM after that!! Love how this looks btw.

control-plane/connect-inject/endpoints_controller.go
ndhanushkodi: Yup totally makes sense, added an Asana

control-plane/connect-inject/endpoints_controller_test.go
ndhanushkodi: Added tests for error cases, great callout!

control-plane/connect-inject/endpoints_controller.go
ndhanushkodi: Yup added!

PR #1238, opened 2022-05-25
Add acceptance tests matrix tests for consul nightlies

kschoche:
Changes proposed in this PR:
- Adds 2 new acceptance tests which run nightly tests of `consul-k8s-control-plane:0.44` + nightly builds of `release/1.11.x` and `release/1.12.x` of Consul. 
- These are not actually nightly builds of Consul, but most recent commit builds off of the release branches, as the build pipeline for Consul doesnt have a nightly build yet.
- Adds a new flag to the acceptance test framework `-consul-major-version` which takes in the consul major version number and uses this to determine which tests to run. Right now this is used to disable Vault tests for anything older than `consul-1.12` 

It should be possible in the future when github adds large runners to the public to add these tests as a matrix test of our acceptance tests, but for now I didnt see a great way to dry this up or make it substantially more maintainable. I think in the meantime until that happens this approach isnt too bad and will give us additional test coverage.

We will need to bump the latest release of consul-k8s in acceptance tests each time we release unless there is a tag that I'm unaware of that we could use.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -505,7 +505,7 @@ jobs:
       - run: mkdir -p $TEST_RESULTS
       - run-acceptance-tests:
           failfast: true
-          additional-flags: -use-kind -kubecontext="kind-dc1" -secondary-kubecontext="kind-dc2" -consul-image=hashicorp/consul:1.11.6 -consul-k8s-image="docker.mirror.hashicorp.services/hashicorp/consul-k8s-control-plane:0.44.0" -consul-major-version=11
+          additional-flags: -use-kind -kubecontext="kind-dc1" -secondary-kubecontext="kind-dc2" -consul-image=hashicorp/consul:1.11.6 -consul-k8s-image="docker.mirror.hashicorp.services/hashicorp/consul-k8s-control-plane:0.44.0" -consul-version="1.11.6"
       - store_test_results:
           path: /tmp/test-results
       - store_artifacts:
@@ -897,7 +897,7 @@ jobs:
             - ~/.go_workspace/pkg/mod
       - run: mkdir -p $TEST_RESULTS
       - run-acceptance-tests:
-          additional-flags: -use-kind -kubecontext="kind-dc1" -secondary-kubecontext="kind-dc2" -enable-transparent-proxy -consul-k8s-image=$CONSUL_K8S_IMAGE -consul-image=$CONSUL_IMAGE -consul-major-version=11
+          additional-flags: -use-kind -kubecontext="kind-dc1" -secondary-kubecontext="kind-dc2" -enable-transparent-proxy -consul-k8s-image=$CONSUL_K8S_IMAGE -consul-image=$CONSUL_IMAGE -consul-version="1.11-dev"
       - store_test_results:
           path: /tmp/test-results
       - store_artifacts:
@@ -932,7 +932,7 @@ jobs:
             - ~/.go_workspace/pkg/mod
       - run: mkdir -p $TEST_RESULTS
       - run-acceptance-tests:
-          additional-flags: -use-kind -kubecontext="kind-dc1" -secondary-kubecontext="kind-dc2" -enable-transparent-proxy -consul-k8s-image=$CONSUL_K8S_IMAGE -consul-image=$CONSUL_IMAGE -consul-major-version=12
+          additional-flags: -use-kind -kubecontext="kind-dc1" -secondary-kubecontext="kind-dc2" -enable-transparent-proxy -consul-k8s-image=$CONSUL_K8S_IMAGE -consul-image=$CONSUL_IMAGE -consul-version="1.12-dev"
       - store_test_results:
           path: /tmp/test-results
       - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,9 +66,13 @@ commands:
         default: false
       additional-flags:
         type: string
+      consul-image:
+        type: string
+        default: "hashicorp/consul:1.11.6"
       consul-k8s-image:
         type: string
-        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:$(git rev-parse --short HEAD)"
+        default: "hashicorp/consul-k8s-control-plane:0.44.0"
+        #default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:$(git rev-parse --short HEAD)"
       go-path:
         type: string
         default: "/home/circleci/.go_workspace"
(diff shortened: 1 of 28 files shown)


.circleci/config.yml
```diff
@@ -864,7 +864,112 @@ jobs:
           path: /tmp/test-results
       - slack/status:
           fail_only: true
-          failure_message: "Acceptance tests against Kind with Kubernetes v1.22 failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+          failure_message: "Acceptance tests against Kind with Kubernetes v1.23 failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+
+  acceptance-kind-1-23-consul-nightly-1-10:
+    environment:
+      - TEST_RESULTS: /tmp/test-results
+      - CONSUL_IMAGE: "https://hub.docker.com/r/hashicorppreview/consul:1.10.0-dev"
```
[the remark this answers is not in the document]
kschoche: Updated with the correct images thanks! copy/pasta problems.

jmurret:
Can we utilize Circle's Matrix jobs for this instead: https://circleci.com/blog/circleci-matrix-jobs/ ?

kschoche:
> Can we utilize Circle's Matrix jobs for this instead: https://circleci.com/blog/circleci-matrix-jobs/ ?
> 
> GH actions also has this functionality when we move the acceptance tests over: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

My understanding is that we cant use the Matrix jobs because the github runners arent heavy duty enough to run our tests yet.

jmurret:
sorry I meant doing using CircleCI's matrix job to do this now since you are making a CircleCI change.  Maybe that's a v2 thing or wait for GH Actions.  I just think this 1k line config file could be more like 500 lines if we used a matrix job for consul-k8s acceptance and consul acceptance test.  It's more of a nice to have and can be done later.

jmurret (approved):

curtbushko (approved):
Looks good! Bring on the extra nightly tests!

.circleci/config.yml
```diff
@@ -864,7 +864,112 @@ jobs:
...
+      - store_artifacts:
+          path: /tmp/test-results
+      - slack/status:
+          fail_only: true
+          failure_message: "Acceptance tests against Kind with Kubernetes v1.23 with Consul 1.10.x nightly failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+
+  acceptance-kind-1-23-consul-nightly-1-11:
+    environment:
+      - TEST_RESULTS: /tmp/test-results
+      - CONSUL_IMAGE: "docker.mirror.hashicorp.services/hashicorppreview/consul:1.12.0-dev"
```
ishustava, line 907: should this be 1.11 image?

kschoche: Oh great catch, copy/pasta error :)

```suggestion
      - CONSUL_IMAGE: "docker.mirror.hashicorp.services/hashicorppreview/consul:1.11.0-dev"
```

.circleci/config.yml
```diff
@@ -864,7 +864,112 @@ jobs:
...
+          path: /tmp/test-results
+      - slack/status:
+          fail_only: true
+          failure_message: "Acceptance tests against Kind with Kubernetes v1.23 with Consul 1.10.x nightly failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+
+  acceptance-kind-1-23-consul-nightly-1-11:
+    environment:
+      - TEST_RESULTS: /tmp/test-results
+      - CONSUL_IMAGE: "docker.mirror.hashicorp.services/hashicorppreview/consul:1.11.0-dev"
+      - CONSUL_K8S_IMAGE: "hashicorp/consul-k8s-control-plane:0.44.0"
```
kschoche, line 908: I could not find a copy of the `0.44.0` release tag in our docker mirror, if anyone knows the correct URL I can move it to that instead, but since its an actual released image I dont think its as big of a deal?

kschoche:
> @kschoche did you try to run these tests? Right now some of our tests will fail against those versions

Thats a good question, I did not. I was planning on coming back and addressing them once they failed. I know that for example some of the Vault tests will not work with older Consul versions, etc. I'm sure there are others.

kschoche:
> @kschoche did you try to run these tests? Right now some of our tests will fail against those versions

Thats a good question, I did not. I was planning on coming back and addressing them once they failed. I know that for example some of the Vault tests will not work with older Consul versions, etc. I'm sure there are others.

.circleci/config.yml
lkysow: `docker pull docker.mirror.hashicorp.services/hashicorp/consul-k8s-control-plane:0.44.0` worked for me

thisisnotashwin:
> > @kschoche did you try to run these tests? Right now some of our tests will fail against those versions
> 
> Thats a good question, I did not. I was planning on coming back and addressing them once they failed. I know that for example some of the Vault tests will not work with older Consul versions, etc. I'm sure there are others.

I think Vault tests will fail against 1.11. The `-auto-reload-config` flag that was introduced in 1.12 gets enabled when Vault tests run and that is not compatible with 1.11 and will lead to errors. This is going to be tricky because we need to ensure acceptance tests for features added in 1.12 are not run against the 1.11 image. this might need some tinkering with flags on the acceptance tests.

acceptance/framework/flags/flags.go
```diff
@@ -55,6 +56,7 @@ func (t *TestFlags) init() {
 	flag.StringVar(&t.flagNamespace, "namespace", "", "The Kubernetes namespace to use for tests.")
 
 	flag.StringVar(&t.flagConsulImage, "consul-image", "", "The Consul image to use for all tests.")
+	flag.IntVar(&t.flagConsulMajorVersion, "consul-major-version", 0, "The Consul major version.")
```
kschoche, line 59: Alternatively I could use the `go-version` library to track the whole consul version throughout, but since this is just for acceptance tests so I'm on the fence about it.

.circleci/config.yml
```diff
@@ -864,7 +868,77 @@ jobs:
           path: /tmp/test-results
       - slack/status:
           fail_only: true
-          failure_message: "Acceptance tests against Kind with Kubernetes v1.22 failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+          failure_message: "Acceptance tests against Kind with Kubernetes v1.23 failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+
+  acceptance-kind-1-23-consul-nightly-1-11:
+    environment:
+      - TEST_RESULTS: /tmp/test-results
+      - CONSUL_IMAGE: "docker.mirror.hashicorp.services/hashicorppreview/consul:1.11.0-dev"
```
kschoche, line 876: This URL (and the one for 1.12) is subject to change before merging due to a rel-eng PR

curtbushko (approved):
Looks good!

PR #1245, opened 2022-05-26
`consul-k8s proxy list` command

t-eckert:
https://user-images.githubusercontent.com/29112081/171655291-3ac64e0d-9a12-4c38-846c-f958f262db06.mov

Changes proposed in this PR:
- Adds a command to the CLI which allows users to get a table of all Pods a given namespace (or all namespaces) which run an Envoy container managed by Consul.
- Changes the table output style to be consistent with the Consul CLI.
- Renames `TableEntry` to `Cell`.
- Renames `Rich` to `AddRow`.

How I've tested this PR:
- Unit tests for parsing arguments and returning the correct pods.
- Manual testing with an EKS cluster.

- Running unit tests (please let me know if there is more coverage needed).
- Running manually using `go install && alias consul-k8s=cli`

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,218 @@
+package list
```
t-eckert, line 1: This file is called `command` instead of `list` to match the pattern of the Consul-K8s control plane code. I am going to start shifting the code to this pattern so I can separate out "command" code from business logic in other commands down the line too.

t-eckert:
I am planning on adding the CHANGELOG entries when the feature branch merges to avoid merge conflicts.

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+
+	if c.flagAllNamespaces {
+		c.flagNamespace = ""
+	}
+
+	pods, err := c.fetchPods()
+	if err != nil {
+		c.UI.Output("Error fetching pods:\n%v", err, terminal.WithErrorStyle())
```
kschoche, line 105: Is there a reason why we're using `terminal.WithErrorStyle()` here but not elsewhere?

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,218 @@
...
+
+// Help returns a description of the command and how it is used.
+func (c *ListCommand) Help() string {
+	c.once.Do(c.init)
+	return c.Synopsis() + "\n\nUsage: consul-k8s proxy list [flags]\n\n" + c.help
+}
+
+// Synopsis returns a one-line command summary.
+func (c *ListCommand) Synopsis() string {
+	return "List all Pods running Consul proxies."
```
kschoche, line 121:
```suggestion
	return "List all Pods running an Envoy proxy."
```
I'm not sure on the wording here, these definitely arent "Consul" proxies, but also we technically support using a proxy that isnt Envoy. Thoughts?

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// validateFlags ensures that the flags passed in by the can be used.
+func (c *ListCommand) validateFlags() error {
+	if len(c.set.Args()) > 0 {
+		return errors.New("should have no non-flag arguments")
+	}
+	if !common.IsValidLabel(c.flagNamespace) {
```
kschoche, line 129: `IsValidLabel` appears to be meant for labels, does it make sense to rename this function to something specific about checking for "RFC1123" validity?

Also RFC1123 is a bit tough to find as an end user in google since it is used for many things like DNS and datetime specs.
 Maybe move that error message into the logic of `IsValidLabel()` so we dont have it duplicated elsewhere ?

Two alternatives:
*  we could use this library to replace all of the logic: https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/core/validation
* Not duplicate the check logic and let Kubernetes return the error back up the stack

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,218 @@
...
+	c.kubernetes, err = kubernetes.NewForConfig(restConfig)
+
+	return err
+}
+
+// fetchPods fetches all pods in flagNamespace which run Consul proxies.
+func (c *ListCommand) fetchPods() ([]v1.Pod, error) {
+	var pods []v1.Pod
+
+	gatewaypods, err := c.kubernetes.CoreV1().Pods(c.flagNamespace).List(c.Ctx, metav1.ListOptions{
```
kschoche, line 153: I feel like using just this LabelSelector could bring in other applications that may not have proxies attached.
We also have the following Labels that get attached to the deployments:
```
  selector:
    matchLabels:  
       release: {{ .Release.Name }}
       component: mesh-gateway
```

```
consul-mesh-gateway-f6d566cd8-x75bn            2/2     Running   0               3d17h   app=consul,chart=consul-helm,component=mesh-gateway,pod-template-hash=f6d566cd8,release=consul
```

Since we only support installing one Consul installation per k8s cluster does it make sense to read the helm secret to figure out the release name and use that to be more explicit on the Selectors throughout the entire `fetchPods` function?

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,218 @@
...
+	apigatewaypods, err := c.kubernetes.CoreV1().Pods(c.flagNamespace).List(c.Ctx, metav1.ListOptions{
+		LabelSelector: "api-gateway.consul.hashicorp.com/managed=true",
+	})
+	if err != nil {
+		return nil, err
+	}
+	pods = append(pods, apigatewaypods.Items...)
+
+	sidecarpods, err := c.kubernetes.CoreV1().Pods(c.flagNamespace).List(c.Ctx, metav1.ListOptions{
+		LabelSelector: "consul.hashicorp.com/connect-inject-status=injected",
```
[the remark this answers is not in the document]
kschoche: Woops, I meant for this to be on line `162` because I don't understand where the api-gateway label selector comes from.

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,218 @@
...
+
+// init sets up flags and help text for the command.
+func (c *ListCommand) init() {
+	c.set = flag.NewSets()
+
+	f := c.set.NewSet("Command Options")
+	f.StringVar(&flag.StringVar{
+		Name:    "namespace",
+		Target:  &c.flagNamespace,
+		Default: "default",
```
kschoche, line 43: I think the idea that the Default namespace is `default` makes sense on the surface, but shouldn't it actually be the default namespace defined by the user's kubecontext?

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,218 @@
...
+		c.UI.Output("Namespace: %s\n", c.flagNamespace)
+	}
+
+	var tbl *terminal.Table
+	if c.flagNamespace == "" {
+		tbl = terminal.NewTable("Namespace", "Name", "Type")
+	} else {
+		tbl = terminal.NewTable("Name", "Type")
+	}
+
```
kschoche, line 194: CLI  output formatting code is so challenging to review :(

cli/cmd/proxy/list/command_test.go
```diff
@@ -0,0 +1,211 @@
...
+					},
+				},
+			},
+		},
+		"Sidecar Pods": {
+			namespace: "default",
+			pods: []v1.Pod{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "sidecar",
```
kschoche, line 125: The `name` here is confusing. The sidecar is a container that runs alongside the workload container.
Can we call it `pod1` or something?

cli/cmd/proxy/list/command_test.go
```diff
@@ -0,0 +1,211 @@
...
+						Namespace: "default",
+						Labels: map[string]string{
+							"consul.hashicorp.com/connect-inject-status": "injected",
+						},
+					},
+				},
+			},
+		},
```
kschoche, line 188: Does it make sense to add a test case that validates pods which do NOT have the correct LabelSelectors do not get added?

cli/cmd/proxy/list/command_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			client := fake.NewSimpleClientset()
+
+			// Add the pods to the client.
+			for _, pod := range tc.pods {
+				client.CoreV1().Pods(pod.ObjectMeta.Namespace).Create(context.Background(), &pod, metav1.CreateOptions{})
+			}
+
```
kschoche, line 199: `fake.NewSimpleClientset` will take in objects so you don't need to do the corev1 Create()!
```suggestion
			client := fake.NewSimpleClientset(tc.pods...)
```
(double-check my syntax here please!)

cli/cmd/proxy/list/command.go
t-eckert: Yeah! I will grab that library.

cli/cmd/proxy/list/command.go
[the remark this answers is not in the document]
t-eckert: Sure! It's just to get Consul API gateways. They work differently from the other types of gateways. They don't have a `component: api-gateway` label. There is a controller that deploys API Gateway pods with `api-gateway.consul.hashicorp.com/managed=true`.

cli/common/terminal/table.go
```diff
@@ -5,10 +5,28 @@ import (
 	"github.com/olekukonko/tablewriter"
 )
 
+const (
```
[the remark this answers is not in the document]
t-eckert: The changes in this file rename `TableEntry` to `Cell` just to shorten the name and improve clarity. Also, I changed the table formatting to match the Consul CLI formatting. I changed `Rich` to `AddRow` for clarity. I also moved constants to the top of the file.

cli/cmd/proxy/list/command_test.go
t-eckert: Oh yeah! I forgot about that.

cli/cmd/proxy/list/command.go
t-eckert: Just a mistake. Good catch.

cli/cmd/proxy/list/command.go
t-eckert: Yes, that sounds possible. I would need to fetch the latest secret created by Helm. I see the concern though. A team could have their own code in a pod that has the `component: mesh-gateway` label.

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,228 @@
...
+// initKubernetes initializes the Kubernetes client and sets the namespace based
+// on the user-provided arguments.
+func (c *ListCommand) initKubernetes() error {
+	settings := helmCLI.New()
+
+	restConfig, err := settings.RESTClientGetter().ToRESTConfig()
+	if err != nil {
+		return fmt.Errorf("error retrieving Kubernetes authentication %v", err)
+	}
+	c.kubernetes, err = kubernetes.NewForConfig(restConfig)
```
jmurret, line 142: error check here and fail fast?

cli/cmd/proxy/list/command_test.go
jmurret: +1

jmurret (changes requested):
Nice.  This is looking good.  Mostly some nits.

cli/cmd/proxy/list/command.go
t-eckert: Yep! That was a good call. I made that adjustment and tested that it uses the namespace set by kubeconfig.

cli/cmd/proxy/list/command.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// Use the default namespace if not provided, unless the all-namespaces flag is set.
+	if c.flagNamespace == "" && !c.flagAllNamespaces {
```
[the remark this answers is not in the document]
t-eckert: Such a good improvement, thank you!

kschoche (approved):

cli/cmd/proxy/list/command.go
t-eckert: I will add this as a bug to fix before the final release.

t-eckert: Here is the bug to track this: https://app.asana.com/0/1200500259727746/1202437872167459/f

jmurret (approved):
🎉 Nice work!  Looks good.

PR #1248, opened 2022-05-31
Update api module and exported services validation

ndhanushkodi:
Changes proposed in this PR:
- Update api module and corresponding function signature changes for peering API endpoints
- Update exported services CRD validation: it is no longer an ENT-only CRD, so the validation behavior now accounts for each of those cases, and won't reject the CR in OSS.
  - Only when namespaces are enabled, you can specify the namespace of an exported service
  - Enforces that the name of the CR is "default" in OSS or in the default partition, or equal to the partition name if partitions are enabled.

How I've tested this PR:
- unit tests

- 👀 

Checklist:
- [x] Tests added
- [-] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -146,7 +146,7 @@ jobs:
         working-directory: control-plane
         run: |
           mkdir -p $HOME/bin
-          wget https://github.com/ndhanushkodi/binaries/releases/download/v1oss/consul -O consulbin && \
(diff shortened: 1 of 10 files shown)


ndhanushkodi:
This is ready for another round of review

thisisnotashwin (approved):
Nice work on the tests and also extending consulNamespace() to return `""` when namespaces are disabled!! 🥇

Issue #879, opened 2021-11-29
Add support for chaining root and intermediate CA certificates to server certificate


andreas-ahman:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Is your feature request related to a problem? Please describe.

We have a setup where we're using intermediate CAs, one per datacenter. For our VM's this is not an issue since we manually retrieve the certificates and can append the root and intermediate ca certificates to the server cert.

In kubernetes however we're using the root and intermediate ca chain and the intermediate ca key to sign new certificates. But since these certificates lacks the entire trust chain it fails when trying to communicate with other datacenters.

#### Feature Description

We would like the possibility to specify not only the root ca but the intermediate ca certificate and have the deployment automatically create the certificate chain for signin the certificates and to automatically append the root and intermediate ca certificates to the server certificates to achieve trust. 

#### Use Case(s)

The use case is when you're using intermediate CA's for servers in the Service Mesh and want to add a kubernetes cluster without having to sign the certificates using the root CA.

#### Contributions

Unfortunately I would not be able to help.

t-eckert:
Hi @andreas-ahman,

Thank you for suggesting this feature. I can see how it would be useful. We will triage it and keep you in the loop.

brucec5:
This is also a limitation when using Vault as the secrets backend.

david-yu:
Closing as we now have a solution where we present the intermediate as part of the chain when provisioning the server cert with the Vault secrets backend.

PR #1251, opened 2022-06-01
Present Consul server cert chain when using Vault secrets backend

brucec5:
Changes proposed in this PR:
- Present full cert chain (minus root cert) when using Vault as the secrets backend
- Update some bats tests to get them to pass

I was told by some HC support/CSM folks to open this as a PR. I don't know if this is the best way to accomplish this, or if the testing is sufficient, but hopefully this is useful. This was initially just meant to be a POC to show the engineering team what I did to address the issue on my end.

Currently, only the leaf cert gets presented by the Consul server (both UI and API). This works fine if you have a self-signed cert, but if you're using Vault as a secrets backend with an intermediate CA (and perhaps an offline root CA), then you can't currently verify the full chain.

The fix is to loop over the CA chain and append the intermediate cert(s) to the end of the server cert. This lets Consul present the full chain, which allows for proper validation.

This is related to #879, but only addresses the Vault secrets backend. Different work will need to be done to support the Kubernetes secrets backend.

How I've tested this PR:

I set up an offline root CA cert and used that to sign an L1 ICA in Vault, and used the L1 ICA to sign an L2 ICA. I then used that L2 ICA to issue certs for Consul. I added the root CA cert to my system's trust store and was able to access the UI in the browser without running into a TLS verification error.

It's been a while, but I'm pretty sure I also tested setting up a root CA directly in Vault and using that to issue certs for Consul.

Please feel free to suggest additional testing.

- Deploy Vault to a local K8s cluster
- Set up a root CA in Vault
- Set up an L1 ICA in Vault signed by the root CA
- Optionally set up an L2 ICA (and so on)
- Configure a values.yaml for Consul that uses the Vault secrets backend for TLS certs with TLS enabled
- Either add the root CA cert to your system trust store and load the UI via a web browser, or use openssl to verify the chain via `openssl s_client -connect localhost:8501 -CAfile path/to/root-ca.pem`, which should return successfully
- Uninstall Consul and reinstall it to get certs issued directly from the Vault root CA, which shouldn't result in any extra certs being presented in the chain

Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 


jmurret (approved):
@brucec5 Thank you so much for raising this issue, giving us detailed information, and submitting this solution.  I have validated through a manual script i created in this repo: https://github.com/jmurret/consul-vault-offline-ca-root.  
As mentioned in that README, the browser on a mac gives some fits related to the leaf cert not being standards compliant, but the trust chain is properly displayed in the browser and it passes openssl verification.  I think the mac browser issues are more related to the strictness of this development, self-signed cert, so I'm not worried about it.  

Also, I copied these changes and ran them through our acceptance tests and nothing failed.

I think for the moment manual verification plus the bats unit tests are sufficient.  I'll talk to the team if we want to create an acceptance for it, but since this is an area of code that won't change much and would not really be affected too much other Consul configurations or any timing conditions or anything similar that we see in Kubernetes environments, we may just leave the BATS tests as sufficient, but we'll see.

Once again, thank you so much.  I'll get another team to review.  Cheers.  -John

t-eckert (approved):
@brucec5, this looks fantastic! Thank you for putting the work in to solve this issue. And thank you to @jmurret for the validation script.

jmurret:
@brucec5 can you please sign the CLA? It's the last hurdle to getting this merged and ready for release.  🙏

brucec5:
Sorry for the delay on getting back to this one! I got distracted by some other stuff and kinda forgot about this PR. The CLA is signed now.

PR #1254, opened 2022-06-02
Add consul-version flag to acceptance test framework

kschoche:
Changes proposed in this PR:
- Adds a new flag "-consul-version" to the acceptance test framework that will be used to prevent tests from running against incompatible consul releases.
- Forces Vault acceptance tests to be skipped if consul-1.12+ is not used.
- In the follow-on PR which uses this feature, the acceptance tests will be enhanced to have 2 new jobs which each target specific consul versions and will pass this flag to indicate that.

How I've tested this PR:
* Manually validated the version comparison code.
* Manually tested that the Vault tests pass (are skipped) when consul-1.11.6 is used and "1.11.6" is passed in as a flag.

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/framework/flags/flags.go
+++ b/acceptance/framework/flags/flags.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/hashicorp/consul-k8s/acceptance/framework/config"
+	"github.com/hashicorp/go-version"
 )
 
 type TestFlags struct {
@@ -30,6 +31,7 @@ type TestFlags struct {
 
 	flagConsulImage    string
 	flagConsulK8sImage string
+	flagConsulVersion  string
 
 	flagNoCleanupOnFailure bool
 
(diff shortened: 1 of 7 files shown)


acceptance/framework/flags/flags.go
```diff
@@ -113,6 +116,9 @@ func (t *TestFlags) Validate() error {
 func (t *TestFlags) TestConfigFromFlags() *config.TestConfig {
 	tempDir := t.flagDebugDirectory
 
+	// if the Version is empty consulVersion will be nil
+	consulVersion, _ := version.NewVersion(t.flagConsulVersion)
```
curtbushko, line 120: I was going to say something about 'nil' but I cannot think of a better value for it. 😄

acceptance/tests/vault/vault_namespaces_test.go
```diff
@@ -24,6 +25,13 @@ func TestVault_VaultNamespace(t *testing.T) {
 	cfg := suite.Config()
 	ctx := suite.Environment().DefaultContext(t)
 	ns := ctx.KubectlOptions(t).Namespace
+
+	ver, err := version.NewVersion("1.12.0")
```
curtbushko, line 29: Does this mean that your other PR will change and you will pass in consul-version to the circle CI tests to match?

curtbushko (commented):
Nice job!  I only have a couple little questions but nothing blocking.

acceptance/tests/vault/vault_namespaces_test.go
kschoche: Correct, when we add the new circleci jobs they will each now pass `-consul-image=<hashicorpdevpreview>/consul:1.11-dev` and `-consul-version=1.11-dev` and those will be both hard coded and not have to change. (and 1.12-* respectively)

thisisnotashwin (approved):
💥

PR #1265, opened 2022-06-08
Backfill unit tests for peering dialler

thisisnotashwin:
Changes proposed in this PR:
- Clean up some of the logic in peering dialler and acceptor
- Rename handler -> connectWebhook

How I've tested this PR:
- Unit tests
- 👀 

Checklist:
- [x] Tests added


control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -55,112 +54,106 @@ func (r *PeeringDialerController) Reconcile(ctx context.Context, req ctrl.Reques
...
 		// If the peering in Consul does not exist, initiate peering.
-		if !peeringExists {
-			r.Log.Info("status.secret exists, but the peering doesn't exist in Consul; establishing peering with the existing spec.peer.secret", "secret-name", peeringDialer.Spec.Peer.Secret.Name, "secret-namespace", peeringDialer.Namespace)
-			peeringToken := existingSpecSecret.Data[peeringDialer.Spec.Peer.Secret.Key]
-			_, err := r.initiatePeering(ctx, peeringDialer.Name, string(peeringToken))
-			if err != nil {
-				_ = r.updateStatusError(ctx, peeringDialer, err)
+
+		// Read the peering from Consul.
+		// TODO(peering): do we need to pass in partition?
```
kschoche, line 126: Would this not already be set inside the `ConsulClient` ?

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -169,66 +162,60 @@ func (r *PeeringDialerController) Reconcile(ctx context.Context, req ctrl.Reques
...
 }
 
-func (r *PeeringDialerController) getExistingSecret(ctx context.Context, name string, namespace string) (bool, *corev1.Secret, error) {
-	existingSecret := &corev1.Secret{}
-	namespacedName := types.NamespacedName{
-		Name:      name,
-		Namespace: namespace,
-	}
-	err := r.Client.Get(ctx, namespacedName, existingSecret)
+func (r *PeeringDialerController) getSecret(ctx context.Context, name string, namespace string) (*corev1.Secret, error) {
```
kschoche, line 207: nice work on this cleanup

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -18,6 +23,462 @@ import (
...
+					},
+				},
+			},
+			peeringExists: true,
+		},
+	}
+	for name, tt := range cases {
+		t.Run(name, func(t *testing.T) {
+
+			// Create test consul server
```
kschoche, line 217:
```suggestion
			// Create test consul server.
```

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -18,6 +23,462 @@ import (
...
+				addr := strings.Split(acceptorPeer.HTTPAddr, ":")[0]
+				// Generate expected token for Peering Initiate.
+				tokenString := fmt.Sprintf(`{"CA":null,"ServerAddresses":["%s:8300"],"ServerName":"%s","PeerID":"%s"}`, addr, token.ServerName, token.PeerID)
+				// Create peering initiate secret in Kubernetes.
+				encodedPeeringToken = base64.StdEncoding.EncodeToString([]byte(tokenString))
+				k8sObjects = append(k8sObjects, tt.peeringSecret(encodedPeeringToken))
+			}
+
+			// Create test consul server
+			diallerPeer, err := testutil.NewTestServerConfigT(t, func(c *testutil.TestServerConfig) {
```
kschoche, line 264: `diallerPeer` is misspelled in this function, should be `dialerPeer` ?

kschoche: same with client. honestly its so nit maybe not worth bothering :D

kschoche: If you do go back and change it. Can we call it `dialerPeerServer` or something?

control-plane/connect-inject/peering_dialer_controller.go
ndhanushkodi: Yup it should be, it'll use the partition of the local client agent which is fine because we do want to use the local partition we're in for the controllers. It was a leftover comment from when I wasn't 100% sure. So I think this comment can be removed!

ndhanushkodi, line 126:
```suggestion
```

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -55,112 +54,106 @@ func (r *PeeringDialerController) Reconcile(ctx context.Context, req ctrl.Reques
...
 		// Or, if the peering in Consul does exist, compare it to the contents of the spec's secret. If there's any
 		// differences, initiate peering.
-		if r.specStatusSecretsDifferent(peeringDialer, existingSpecSecret) {
-			r.Log.Info("status.secret exists and is different from spec.peer.secret; establishing peering with the existing spec.peer.secret", "secret-name", peeringDialer.Spec.Peer.Secret.Name, "secret-namespace", peeringDialer.Namespace)
-			peeringToken := existingSpecSecret.Data[peeringDialer.Spec.Peer.Secret.Key]
-			_, err := r.initiatePeering(ctx, peeringDialer.Name, string(peeringToken))
-			if err != nil {
-				_ = r.updateStatusError(ctx, peeringDialer, err)
+		if r.specStatusSecretsDifferent(dialer, specSecret) {
+			r.Log.Info("status.secret exists and is different from spec.peer.secret; establishing peering with the existing spec.peer.secret", "secret-name", dialer.Secret().Name, "secret-namespace", dialer.Namespace)
```
ndhanushkodi, line 150:
```suggestion
			r.Log.Info("the secret in status.secretRef exists and is different from spec.peer.secret; establishing peering with the existing spec.peer.secret", "secret-name", dialer.Secret().Name, "secret-namespace", dialer.Namespace)
```

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -18,6 +23,462 @@ import (
...
+					},
+				}
+				return []runtime.Object{dialer}
+			},
+			expectedConsulPeerings: &api.Peering{
+				Name:  "peering",
+				State: api.PeeringStateActive,
+			},
+			peeringSecret: func(token string) *corev1.Secret {
+				return createSecret("dialer-token", "default", "token", token)
```
ndhanushkodi, line 106: Could it make sense to set the resource version on the secret here, and assert on that resource version in the expectedStatus? We might be able to do it here since we control the peering secret.

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -18,6 +23,462 @@ import (
...
+			// of the Consul server as the generated token currently does not have that set on it.
+			var encodedPeeringToken string
+			if tt.peeringName != "" {
+				var token struct {
+					CA              string
+					ServerAddresses []string
+					ServerName      string
+					PeerID          string
+				}
+				// Create the initial token,
```
ndhanushkodi, line 246:
```suggestion
				// Create the initial token.
```

control-plane/connect-inject/peering_dialer_controller_test.go
ndhanushkodi: +1 to `dialerPeerServer` :)

ndhanushkodi (approved):
This looks amazing and the unit tests are so easy to compare to the code cases and it's so thorough! I love how you've reduced the verbosity also!! Just a few small nits!

control-plane/api/v1alpha1/peeringacceptor_types.go
```diff
@@ -76,12 +76,15 @@ type ReconcileErrorStatus struct {
 }
 
 type SecretRefStatus struct {
-	// Name is the name of the secret generated.
-	Name string `json:"name,omitempty"`
-	// Key is the key of the secret generated.
-	Key string `json:"key,omitempty"`
-	// Backend is where the generated secret is stored. Currently supports the value: "kubernetes".
-	Backend string `json:"backend,omitempty"`
+	Secret `json:",inline"`
```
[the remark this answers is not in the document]
thisisnotashwin: yes yes!! externally it will have the same interface as before but we get to reuse an existing type because all it's fields are the same.

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -18,6 +23,465 @@ import (
...
+					Spec: v1alpha1.PeeringDialerSpec{
+						Peer: &v1alpha1.Peer{
+							Secret: nil,
+						},
+					},
+				}
+				return []runtime.Object{dialer}
+			},
+			expErr:        "PeeringDialer spec.peer.secret was not set",
+			peeringSecret: func(_ string) *corev1.Secret { return nil },
```
kschoche, line 57: these are so cool :)

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -18,6 +23,465 @@ import (
...
+							Key:     "token",
+							Backend: "vault",
+						},
+					},
+				},
+			},
+			secret:      nil,
+			isDifferent: true,
+		},
+		"different secret ref in status and saved secret": {
```
kschoche, line 416: Is it possible for the versions to end up out of order, i.e. this version is older than what is stored in the secret?
 if so should we test for that?

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -18,6 +23,465 @@ import (
...
+						},
+					},
+					Status: v1alpha1.PeeringDialerStatus{
+						SecretRef: &v1alpha1.SecretRefStatus{
+							Secret: v1alpha1.Secret{
+								Name:    "dialer-token-old",
+								Key:     "token",
+								Backend: "kubernetes",
+							},
+							ResourceVersion: "test-version",
```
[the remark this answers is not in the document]
thisisnotashwin: If there is a version mismatch, there will be an additional reconcile, the outcome of which will ensure the correct resource version is written into the status.

control-plane/connect-inject/peering_dialer_controller_test.go
thisisnotashwin: if the secret changes, it's resource version would change. I don't think there is a way where the spec secret version is older than the one in the status if that makes sense. am I misunderstanding the question?

kschoche: I was just wondering if it were possible to get out of order here due to caching effects or a failed secret update, etc. It's more of a question about how the resourceVersion is compared I suppose. But sounds like this isnt something we'd need to worry about!

thisisnotashwin: yeah! i think that aspect should be safe. we will never be in a situation where the status "gets ahead" of the spec. If the status is not updated for whatever reason, re-running the steps will be a no-op and the status will then get updated.

kschoche (approved):
Looks great, thanks for taking the time to chat about it with me!

PR #1267, opened 2022-06-09
Initial support for agentless

ishustava:
Changes proposed in this PR:
Add initial support for the basic service mesh running with agentless. 
- Update connect-inject deployment to talk to consul servers (currently only in-cluster servers are supported; external servers support will come in a later PR)
- Update endpoints controller to register and deregister services and health checks in the catalog and stop using agent APIs
- Update connect-init command to take the `-node-name` flag so that it can search for services within a given node using catalog APIs
- Add k8s probes to the envoy container
- Provide `-node-name` flag to the `consul connect envoy` command so that we can generate correct configuration
- Selectively disable acceptance tests that don't yet work in this configuration

How I've tested this PR:
acceptance tests

👀 
TODO:
- [x] add k8s probes to envoy container
- [x] fix acceptance tests
- [ ] ~~add support for external servers~~ (will come in a separate PR as this one is already big)

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


.circleci/config.yml
```diff
@@ -887,10 +887,10 @@ workflows:
...
           requires:
             - dev-upload-docker
-      - acceptance-tproxy:
-          context: consul-ci
-          requires:
-            - dev-upload-docker
+#      - acceptance-tproxy:
+#          context: consul-ci
+#          requires:
+#            - dev-upload-docker
```
ishustava, line 893: tproxy is not yet supported

charts/consul/test/unit/client-config-configmap.bats
```diff
@@ -2,30 +2,23 @@
...
-      . | tee /dev/stderr |
-      yq 'length > 0' | tee /dev/stderr)
-  [ "${actual}" = "true" ]
-}
-
 @test "client/ConfigMap: enable with global.enabled false" {
   cd `chart_dir`
   local actual=$(helm template \
       -s templates/client-config-configmap.yaml  \
+      --set 'client.enabled=true' \
```
ishustava, line 9: Clients are now disabled by default (previously they were enabled), and so lots of bats tests needed to be changed to enable clients explicitly

charts/consul/test/unit/connect-inject-deployment.bats
```diff
@@ -617,83 +575,36 @@ EOF
...
       --set 'global.tls.enabled=true' \
       . | tee /dev/stderr |
-      yq '.spec.template.spec.containers[0].command | any(contains("-tls-cert-dir=/etc/connect-injector/certs"))' | tee /dev/stderr)
-  [ "${actual}" != "" ]
+      yq -r '.spec.template.spec.containers[0].env[] | select(.name == "CONSUL_HTTP_ADDR").value' | tee /dev/stderr)
+  [ "${actual}" = "https://release-name-consul-server.default.svc:8501" ]
 }
 
-#--------------------------------------------------------------------
-# global.tls.enableAutoEncrypt
```
ishustava, line 632: Auto-encrypt is no longer supported for the connect-inject deployment as this is a setting used only when talking to consul clients.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -1089,92 +969,6 @@ func shouldIgnore(namespace string, denySet, allowSet mapset.Set) bool {
 	return false
 }
 
-// filterAgentPods receives meta and object information for Kubernetes resources that are being watched,
-// which in this case are Pods. It only returns true if the Pod is a Consul Client Agent Pod. It reads the labels
-// from the meta of the resource and uses the values of the "app" and "component" label to validate that
-// the Pod is a Consul Client Agent.
-func (r *EndpointsController) filterAgentPods(object client.Object) bool {
```
ishustava, line 1096: This code might come back to better handle upgrades but removing it for now as I'm not sure yet how upgrades will be handled.

control-plane/connect-inject/envoy_sidecar.go
```diff
@@ -27,25 +28,27 @@ func (w *MeshWebhook) envoySidecar(namespace corev1.Namespace, pod corev1.Pod, m
...
-					FieldRef: &corev1.ObjectFieldSelector{FieldPath: "status.hostIP"},
-				},
+	probe := &corev1.Probe{
+		Handler: corev1.Handler{
+			TCPSocket: &corev1.TCPSocketAction{
+				Port: intstr.FromInt(EnvoyInboundListenerPort + mpi.serviceIndex),
 			},
 		},
+		InitialDelaySeconds: 1,
+	}
```
ishustava, line 38: This probe is to replace our existing TCP health check that we register with consul. I didn't add startup probe as we don't have anything like that for now and so I'm not sure if adding a startup probe will cause problems because we haven't see envoy being slow to start up.

control-plane/subcommand/connect-init/command_ent_test.go
```diff
@@ -9,111 +9,100 @@ import (
...
 	"github.com/stretchr/testify/require"
+
+	connectinject "github.com/hashicorp/consul-k8s/control-plane/connect-inject"
+	"github.com/hashicorp/consul-k8s/control-plane/helper/test"
+	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
+	"github.com/hashicorp/consul-k8s/control-plane/subcommand/common"
 )
 
-func TestRun_ServicePollingWithACLsAndTLSWithNamespaces(t *testing.T) {
+func TestRun_WithNamespaces(t *testing.T) {
```
ishustava, line 23: I've removed TLS test cases from this test and the command_test.go because I don't think there's any logic in the command that is tls-related  as it just takes whatever client  we configure from http flags. As a result, I don't think it'll be possible for these tests to fail if the code changes, and so I don't think there's any reason to keep them.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -446,21 +334,37 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
 	}
 	tags := consulTags(pod)
```
ishustava, line 336: @mkeeler would love if you could take a look at this function. Here it is without the diff to make it a bit easier to read: https://github.com/hashicorp/consul-k8s/blob/53a26e2f2206fbb782b97e5ce1bc6266474bfada/control-plane/connect-inject/endpoints_controller.go#L293

For context, this is the function that will get called every time when there's any update to the kubernetes service's endpoints (e.g. when a new service is created or when a new pod added etc). 

My main question is about node registration. I think you mentioned before that we'd need to register the node before we register services. Currently in this code, I didn't do this but still provided `SkipNodeUpdate: true` and all tests (unit and acceptance) seem to be happy. I'm wondering if I'm missing something.

Let me know if this is enough context to review!

.circleci/config.yml
```diff
@@ -99,8 +99,8 @@ commands:
                     if ! gotestsum --no-summary=all --jsonfile=jsonfile-${pkg////-} -- $pkg -p 1 -timeout 2h -failfast \
                           << parameters.additional-flags >> \
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
-                          -enable-multi-cluster \
```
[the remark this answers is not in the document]
ishustava: This will be eventually supported but not for tech preview. That's why I'm removing it here and also disabling a bunch of other tests

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -1,7 +1,5 @@
 {{- if and .Values.global.peering.enabled (not .Values.connectInject.enabled) }}{{ fail "setting global.peering.enabled to true requires connectInject.enabled to be true" }}{{ end }}
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
-{{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled for connect injection" }}{{ end }}
```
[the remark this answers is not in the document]
ishustava: Helm doesn't support warnings so we can only throw an error.

>the .Values.client.* are all removed as per the RFC.

It's not yet removed - just setting it to default `false` for now, but it'll be removed eventually.

charts/consul/test/unit/connect-inject-deployment.bats
```diff
@@ -1862,33 +1717,44 @@ EOF
...
 
-@test "connectInject/Deployment: get-auto-encrypt-client-ca uses server's stateful set address by default and passes ca cert" {
+@test "connectInject/Deployment: CONSUL_CACERT env variable is set points to vault secrets when TLS and vault are enabled" {
   cd `chart_dir`
-  local command=$(helm template \
+  local actual=$(helm template \
       -s templates/connect-inject-deployment.yaml  \
       --set 'connectInject.enabled=true' \
       --set 'global.tls.enabled=true' \
       --set 'global.tls.enableAutoEncrypt=true' \
```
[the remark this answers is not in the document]
ishustava: good catch! I think all those need to be removed.

charts/consul/values.yaml
```diff
@@ -1067,9 +1067,8 @@ client:
   # If true, the chart will install all
   # the resources necessary for a Consul client on every Kubernetes node. This _does not_ require
   # `server.enabled`, since the agents can be configured to join an external cluster.
-  # @default: global.enabled
   # @type: boolean
-  enabled: "-"
+  enabled: false
```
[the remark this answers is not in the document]
ishustava: I'll add support for it after tech preview as this is needed for multi-cluster scenario which is not going to be supported yet

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -1240,7 +1032,7 @@ func consulTags(pod corev1.Pod) []string {
 
 func getMultiPortIdx(pod corev1.Pod, serviceEndpoints corev1.Endpoints) int {
 	for i, name := range strings.Split(pod.Annotations[annotationService], ",") {
-		if name == getServiceName(pod, serviceEndpoints) {
+		if name == serviceName(pod, serviceEndpoints) {
```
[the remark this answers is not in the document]
ishustava: Yeah I took this suggestion from Effective Go: https://go.dev/doc/effective_go#Getters

thisisnotashwin (commented):
This looks really good! Overall, the complexity that this shaves genuinely leads me to believe this is a better design!

charts/consul/test/unit/connect-inject-deployment.bats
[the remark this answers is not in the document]
ishustava: Removed the rest of the auto encrypt references in this test! thanks for the catch!

curtbushko (approved):
So much cleanup is so great!

thisisnotashwin (approved):
❤️ this is excellent!! great work Iryna!!

control-plane/connect-inject/endpoints_controller.go
mkeeler: `SkipNodeUpdate` only has an effect when the node is already registered. When its not already registered Consul will still register the node. This sort of makes sense because its a create and not an update that is happening.

I suspect your tests pass because they are either not using ACLs or the token used has the necessary `node:write` privileges.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -57,19 +50,29 @@ const (
...
+	// ConsulNodeAddress is the address of the consul node (defined by ConsulNodeName).
+	// This address does not need to be routable as this node is ephemeral, and we're only providing it because
+	// Consul's API currently requires node address to be provided when registering a node.
+	ConsulNodeAddress = "127.0.0.1"
+
+	// ConsulKubernetesCheckType is the type of health check in Consul for Kubernetes readiness status.
+	ConsulKubernetesCheckType = "kubernetes-readiness"
+
+	// ConsulKubernetesCheckName is the name of health check in Consul for Kubernetes readiness status.
+	ConsulKubernetesCheckName = "Kubernetes Readiness Check"
```
ndhanushkodi, line 66: in a custom health check like this, is the status of the check all consul would care about to decide whether to route traffic to the app?

ndhanushkodi (approved):
This looks so great!! Thanks for the helpful comments all around and also reviewing this made me realize how well the k8s agentless RFC is written!!!

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -326,45 +320,32 @@ spec:
...
+      initContainers:
       - name: connect-injector-acl-init
         env:
-        - name: HOST_IP
-          valueFrom:
-            fieldRef:
-              fieldPath: status.hostIP
-          {{- if .Values.global.tls.enabled }}
-        - name: CONSUL_CACERT
-          value: /consul/tls/ca/tls.crt
```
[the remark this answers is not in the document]
ishustava: yeah exactly! and more specifically, when using auto-encrypt the CA for the client will be different (connect CA) and so we needed to mount a client-specific CA in that case. Without clients, we'd only need the server CA.

control-plane/connect-inject/container_init.go
```diff
@@ -36,9 +36,17 @@ type initContainerCommandData struct {
 	ConsulNamespace           string
 	NamespaceMirroringEnabled bool
 
-	// The PEM-encoded CA certificate to use when
+	// ConsulCACert is the PEM-encoded CA certificate to use when
 	// communicating with Consul clients
```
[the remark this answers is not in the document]
ishustava: good catch!

control-plane/connect-inject/endpoints_controller.go
ishustava: yes!

ndhanushkodi: thanks!!

charts/consul/templates/connect-inject-deployment.yaml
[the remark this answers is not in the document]
ndhanushkodi: ahhh thank you for the clarification!!

PR #1269, opened 2022-06-10
Add finalizers to peering acceptor and dialer

thisisnotashwin:
Changes proposed in this PR:
- Add finalizers to peering acceptor and dialer
- Remove ownerRef from secret created by acceptor.
- go mod tidy

How I've tested this PR:
- Unit tests
- 👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/connect-inject/peering_acceptor_controller.go
+++ b/control-plane/connect-inject/peering_acceptor_controller.go
@@ -28,6 +28,8 @@ type PeeringAcceptorController struct {
 	context.Context
 }
 
+const FinalizerName = "finalizers.consul.hashicorp.com"
+
 //+kubebuilder:rbac:groups=consul.hashicorp.com,resources=peeringacceptors,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=consul.hashicorp.com,resources=peeringacceptors/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=apps,resources=secrets,verbs=get;list;watch;create;update;patch;delete
@@ -49,17 +51,41 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
 	acceptor := &consulv1alpha1.PeeringAcceptor{}
 	err := r.Client.Get(ctx, req.NamespacedName, acceptor)
 
-	// If the PeeringAcceptor resource has been deleted (and we get an IsNotFound
-	// error), we need to delete it in Consul.
+	// This can be safely ignored as a resource will only ever be not found if it has never been reconciled
+	// since we add finalizers to our resources.
 	if k8serrors.IsNotFound(err) {
-		r.Log.Info("PeeringAcceptor was deleted, deleting from Consul", "name", req.Name, "ns", req.Namespace)
-		if err := r.deletePeering(ctx, req.Name); err != nil {
-			return ctrl.Result{}, err
-		}
+		r.Log.Info("PeeringAcceptor resource not found. Ignoring resource", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, nil
 	} else if err != nil {
 		r.Log.Error(err, "failed to get PeeringAcceptor", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, err
+	}
+
+	// The DeletionTimestamp is zero when the object has not been marked for deletion. The finalizer is added
+	// in case it does not exist to all resources. If the DeletionTimestamp is non-zero, the object has been
+	// marked for deletion and goes into the deletion workflow.
+	if acceptor.GetDeletionTimestamp().IsZero() {
+		if !controllerutil.ContainsFinalizer(acceptor, FinalizerName) {
+			controllerutil.AddFinalizer(acceptor, FinalizerName)
+			if err := r.Update(ctx, acceptor); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	} else {
+		if containsString(acceptor.Finalizers, FinalizerName) {
+			r.Log.Info("PeeringAcceptor was deleted, deleting from Consul", "name", req.Name, "ns", req.Namespace)
+			if err := r.deletePeering(ctx, req.Name); err != nil {
+				return ctrl.Result{}, err
+			}
+			if acceptor.Secret().Backend == "kubernetes" {
+				if err := r.deleteK8sSecret(ctx, acceptor); err != nil {
+					return ctrl.Result{}, err
+				}
+			}
+			controllerutil.RemoveFinalizer(acceptor, FinalizerName)
+			err := r.Update(ctx, acceptor)
+			return ctrl.Result{}, err
+		}
 	}
 
 	statusSecretSet := acceptor.SecretRef() != nil
@@ -104,7 +130,7 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
 			return ctrl.Result{}, err
 		}
 		if acceptor.Secret().Backend == "kubernetes" {
-			secretResourceVersion, err = r.createK8sPeeringTokenSecretWithOwner(ctx, acceptor, resp)
+			secretResourceVersion, err = r.createK8sSecret(ctx, acceptor, resp)
 			if err != nil {
 				r.updateStatusError(ctx, acceptor, err)
 				return ctrl.Result{}, err
@@ -142,7 +168,7 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
 			return ctrl.Result{}, err
 		}
 		if acceptor.Secret().Backend == "kubernetes" {
-			secretResourceVersion, err = r.createK8sPeeringTokenSecretWithOwner(ctx, acceptor, resp)
+			secretResourceVersion, err = r.createK8sSecret(ctx, acceptor, resp)
 			if err != nil {
 				return ctrl.Result{}, err
 			}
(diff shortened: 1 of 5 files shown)


control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -52,16 +54,32 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
 	// If the PeeringAcceptor resource has been deleted (and we get an IsNotFound
 	// error), we need to delete it in Consul.
 	if k8serrors.IsNotFound(err) {
-		r.Log.Info("PeeringAcceptor was deleted, deleting from Consul", "name", req.Name, "ns", req.Namespace)
-		if err := r.deletePeering(ctx, req.Name); err != nil {
-			return ctrl.Result{}, err
-		}
+		r.Log.Info("PeeringAcceptor resource not found. Ignoring.", "name", req.Name, "ns", req.Namespace)
```
thisisnotashwin, line 57: We do not need to worry about the not found error here as this would imply the CRD was never reconciled. If it was, it would have a finalizer so it is guaranteed to go through the workflow that has just been added.

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -43,16 +44,32 @@ func (r *PeeringDialerController) Reconcile(ctx context.Context, req ctrl.Reques
 	// If the PeeringDialer resource has been deleted (and we get an IsNotFound
 	// error), we need to delete it in Consul.
 	if k8serrors.IsNotFound(err) {
-		r.Log.Info("PeeringDialer was deleted, deleting from Consul", "name", req.Name, "ns", req.Namespace)
-		if err := r.deletePeering(ctx, req.Name); err != nil {
-			return ctrl.Result{}, err
-		}
+		r.Log.Info("PeeringDialer resource not found. Ignoring.", "name", req.Name, "ns", req.Namespace)
```
thisisnotashwin, line 47: We do not need to worry about the not found error here as this would imply the CRD was never reconciled. If it was, it would have a finalizer so it is guaranteed to go through the workflow that has just been added.

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -251,9 +269,6 @@ func (r *PeeringAcceptorController) createK8sPeeringTokenSecretWithOwner(ctx con
 	secretName := acceptor.Secret().Name
 	secretNamespace := acceptor.Namespace
 	secret := createSecret(secretName, secretNamespace, acceptor.Secret().Key, resp.PeeringToken)
-	if err := controllerutil.SetControllerReference(acceptor, secret, r.Scheme); err != nil {
```
thisisnotashwin, line 254: Removed this as it causes the secret to be deleted as soon as it is migrated between cluster. We will add watchers for secrets in a subsequent PR.

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -52,16 +54,32 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
...
+		if !controllerutil.ContainsFinalizer(acceptor, FinalizerName) {
+			controllerutil.AddFinalizer(acceptor, FinalizerName)
+			if err := r.Update(ctx, acceptor); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	} else {
+		if containsString(acceptor.Finalizers, FinalizerName) {
+			r.Log.Info("PeeringAcceptor was deleted, deleting from Consul", "name", req.Name, "ns", req.Namespace)
+			if err := r.deletePeering(ctx, req.Name); err != nil {
```
ndhanushkodi, line 74: Could we also delete the generated K8s secret specified in the status when the PeeringAcceptor is deleted?

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -52,16 +54,32 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
...
+				return ctrl.Result{}, err
+			}
+			controllerutil.RemoveFinalizer(acceptor, FinalizerName)
+			err := r.Update(ctx, acceptor)
+			return ctrl.Result{}, err
```
ndhanushkodi, line 79: After returning at this point, does the Acceptor resource still exist, just without a finalizer? And then K8s will delete the Acceptor automatically?

control-plane/connect-inject/peering_acceptor_controller_test.go
```diff
@@ -459,8 +454,24 @@ func TestReconcileDeletePeeringAcceptor(t *testing.T) {
...
+					Peer: &v1alpha1.Peer{
+						Secret: &v1alpha1.Secret{
+							Name:    "acceptor-deleted-secret",
+							Key:     "data",
+							Backend: "kubernetes",
+						},
+					},
+				},
+			}
+			k8sObjects := []runtime.Object{&ns, acceptor}
```
ndhanushkodi, line 475: Would it make sense to assert that the Acceptor's finalizer is removed at the bottom of this test? Since after calling reconcile we'd expect that to be the case? I guess in real K8s env, the resource is probably being deleted but in this test we're just calling reconcile so we might be able to make the assertion that the finalizer is removed.

control-plane/connect-inject/peering_dialer_controller.go
ndhanushkodi, line 47:
```suggestion
		r.Log.Info("PeeringDialer resource not found. Ignoring resource", "name", req.Name, "ns", req.Namespace)
```
super nit: just cause we don't have a . at the end of any other logs :)

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -43,16 +44,32 @@ func (r *PeeringDialerController) Reconcile(ctx context.Context, req ctrl.Reques
...
+	if dialer.GetDeletionTimestamp().IsZero() {
+		if !controllerutil.ContainsFinalizer(dialer, FinalizerName) {
+			controllerutil.AddFinalizer(dialer, FinalizerName)
+			if err := r.Update(ctx, dialer); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	} else {
+		if containsString(dialer.Finalizers, FinalizerName) {
+			r.Log.Info("PeeringAcceptor was deleted, deleting from Consul", "name", req.Name, "ns", req.Namespace)
```
ndhanushkodi, line 63:
```suggestion
			r.Log.Info("PeeringDialer was deleted, deleting from Consul", "name", req.Name, "ns", req.Namespace)
```

control-plane/connect-inject/peering_acceptor_controller.go
ndhanushkodi: I think it would be reasonable to do for the Acceptor but not the Dialer, since the Acceptor is the one that generates that secret, vs on the Dialer the secret isn't managed by the controller.

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -502,10 +504,24 @@ func TestReconcileDeletePeeringDialer(t *testing.T) {
...
+					Namespace:         "default",
+					DeletionTimestamp: &metav1.Time{Time: time.Now()},
+					Finalizers:        []string{FinalizerName},
+				},
+				Spec: v1alpha1.PeeringDialerSpec{
+					Peer: &v1alpha1.Peer{
+						Secret: nil,
+					},
+				},
+			}
```
ndhanushkodi, line 522: Similar to my comment above, could we make an assertion on the finalizer key being removed?

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -52,16 +54,32 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
...
-			return ctrl.Result{}, err
-		}
+		r.Log.Info("PeeringAcceptor resource not found. Ignoring.", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, nil
 	} else if err != nil {
 		r.Log.Error(err, "failed to get PeeringAcceptor", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, err
 	}
 
+	if acceptor.GetDeletionTimestamp().IsZero() {
```
ndhanushkodi, line 64: This will add the finalizer to any resource that doesn't already have it right? Not just recently deleted resources? Could we add a comment for what this case is describing?

ndhanushkodi (commented):
Looks great!!! Could you also try manually adding an acceptor or dialer and seeing that when you delete it you generally see what you think should happen? TYSM!

Also for the acceptor, I think we should delete the peering secret in the finalizer handling, wdyt?

control-plane/connect-inject/peering_acceptor_controller.go
thisisnotashwin: great catch!!

control-plane/connect-inject/peering_acceptor_controller.go
thisisnotashwin: yup. at this point, the acceptor resource will exist without the finalizer and get immediately GC-ed by kubernetes.

control-plane/connect-inject/peering_acceptor_controller.go
thisisnotashwin: Let me know if the comment makes sense!

control-plane/connect-inject/peering_dialer_controller_test.go
thisisnotashwin: So the object gets deleted from the cluster but I wrote a test to validate that.

control-plane/connect-inject/peering_acceptor_controller.go
ndhanushkodi: Yep its great, tysm!!

control-plane/connect-inject/peering_dialer_controller_test.go
ndhanushkodi: Perfect!

ndhanushkodi (approved):
Looks great!

PR #1273, opened 2022-06-13
Cluster Peering

ndhanushkodi:
Changes proposed in this PR:
- Best reviewed by commit, plan is to rebase and merge (not squash) this branch into main
- Adds PeeringAcceptor controller run as part of the connect-inject deployment
- Adds PeeringDialer controller run as part of the connect-inject deployment
- Adds PeeringAcceptor and Peering Dialer CRs
- Adds Peer field to Exported Services CR and Intentions CR
- Allows ExportedServices CR to be used in OSS (previously it was used only for cross-partition in ENT)
- Adds Helm value `global.peering.enabled` which will deploy the controllers and CRs for peering, if connectInject.enabled is also true. 

How I've tested this PR:
unit tests and manual testing using steps [here](https://github.com/ndhanushkodi/cluster-peering-demo/tree/main/servicemesh-meshgw-intention-unpeer-demo)

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


.github/workflows/test.yml
```diff
@@ -142,14 +142,13 @@ jobs:
       - run:  mkdir -p ${{env.TEST_RESULTS}}
       - run:  echo "$HOME/bin" >> $GITHUB_PATH
```
ndhanushkodi, line 144: The changes in this file will need to be reverted before merging-- but we need the Consul release to happen first. Then we need to bump the CONSUL_VERSION variable to be the latest Consul release

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+	peering, _, err := r.ConsulClient.Peerings().Read(ctx, acceptor.Name, nil)
+	if err != nil {
+		r.Log.Error(err, "failed to get Peering from Consul", "name", req.Name)
+		return ctrl.Result{}, err
+	}
+
+	// If the peering doesn't exist in Consul, generate a new token, and store it in the specified backend. Store the
+	// current state in the status.
+	if peering == nil {
+		r.Log.Info("peering doesn't exist in Consul", "name", acceptor.Name)
```
ishustava, line 115:
```suggestion
		r.Log.Info("peering doesn't exist in Consul; creating new peering", "name", acceptor.Name)
```

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+
+		if statusSecretSet {
+			if existingStatusSecret != nil {
+				err := r.Client.Delete(ctx, existingStatusSecret)
```
ishustava, line 119: should we log this?

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+		err := r.updateStatus(ctx, acceptor, secretResourceVersion)
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// shouldGenerateToken returns whether a token should be generated, and whether the name of the secret has changed. It
+// compares the spec secret's name/key/backend and contents to the status secret's name/key/backend and contents. The
+// contents are compared by taking a SHA256 sum of the secret.
```
ishustava, line 197:
```suggestion
// shouldGenerateToken returns whether a token should be generated, and whether the name of the secret has changed. It
// compares the spec secret's name/key/backend and resource version with the name/key/backend and resource version of the status secret's.
```

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+		return ctrl.Result{}, err
+	} else if err != nil {
+		r.Log.Error(err, "failed to get Peering from Consul", "name", req.Name)
+		return ctrl.Result{}, err
+	}
+
+	// TODO(peering): Verify that the existing peering in Consul is an acceptor peer. If it is a dialing peer, an error should be thrown.
+
+	// If the peering does exist in Consul, figure out whether to generate and store a new token by comparing the secret
+	// in the status to the actual contents of the secret. If no secret is specified in the status, shouldGenerate will
```
ishustava, line 150:
```suggestion
	// If the peering does exist in Consul, figure out whether to generate and store a new token by comparing the secret
	// in the status to the resource version of the secret. If no secret is specified in the status, shouldGenerate will
```

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+	if acceptor.SecretRef().Key != acceptor.Secret().Key {
+		return true, false, nil
+	}
+	// TODO(peering): remove this when validation webhook exists.
+	if acceptor.SecretRef().Backend != acceptor.Secret().Backend {
+		return false, false, errors.New("PeeringAcceptor backend cannot be changed")
+	}
+	// Compare the existing secret hash.
+	// Get the secret specified by the status, make sure it matches the status' secret.latestHash.
```
ishustava, line 214:
```suggestion
	// Compare the existing secret resource version.
	// Get the secret specified by the status, make sure it matches the status' secret.ResourceVersion.
```

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+		// The secret was deleted.
+		return nil, nil
+	} else if err != nil {
+		r.Log.Error(err, "couldn't get secret", "name", name, "namespace", namespace)
+		return nil, err
+	}
+	return existingSecret, nil
+}
+
+// createK8sSecret creates a secret and uses the controller's K8s client to apply the secret. It checks if
```
ishustava, line 273: nit: should this function be called `createOrUpdateK8sSecret`?

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+func (r *PeeringAcceptorController) deletePeering(ctx context.Context, peerName string) error {
+	_, err := r.ConsulClient.Peerings().Delete(ctx, peerName, nil)
+	if err != nil {
+		r.Log.Error(err, "failed to delete Peering from Consul", "name", peerName)
+		return err
+	}
+	return nil
+}
+
+// createSecret is a helper function that creates a corev1.SecretRef when provided inputs.
```
ishustava, line 348:
```suggestion
// createSecret is a helper function that creates a corev1.Secret when provided inputs.
```

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		StringData: map[string]string{
+			key: value,
+		},
+		Data: map[string][]byte{
+			key: []byte(value),
+		},
```
ishustava, line 360: why do we need to provide both?

control-plane/connect-inject/peering_acceptor_controller_test.go
```diff
@@ -0,0 +1,962 @@
...
+	cases := []struct {
+		name                  string
+		initialConsulPeerName string
+		expErr                string
+	}{
+		{
+			name:                  "PeeringAcceptor ",
+			initialConsulPeerName: "acceptor-deleted",
+		},
+	}
```
ishustava, line 451: nit: if there's just one case, might want to consider to inline the test

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	// Get the status secret and the spec secret.
+	// Cases need to handle secretRefSet, statusSecret, secretSet, specSecret.
+	// no secretSet --> error bc spec needs to be set.
+	// no specSecret --> error bc waiting for spec secret to exist.
+	// no secretRefSet, yes secretSet, no statusSecret, yes specSecret --> initiate peering.
+	// yes secretRefSet, yes secretSet, no statusSecret, yes specSecret --> initiate peering.
+	// yes secretRefSet, yes secretSet, yes statusSecret, yes specSecret --> compare contents, if
+	// different initiate peering.
```
ishustava, line 82: should we remove this comment?

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -0,0 +1,268 @@
...
+		r.updateStatusError(ctx, dialer, err)
+		return ctrl.Result{}, err
+	}
+
+	// If specSecret doesn't exist, error because we can only initiate peering if we have a token to initiate with.
+	if specSecret == nil {
+		err = errors.New("PeeringDialer spec.peer.secret does not exist")
+		r.updateStatusError(ctx, dialer, err)
+		return ctrl.Result{}, err
+	}
```
ishustava, line 108: should `getSecret` return an error for this case?

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -0,0 +1,268 @@
...
+			if err := r.establishPeering(ctx, dialer.Name, string(peeringToken)); err != nil {
+				r.updateStatusError(ctx, dialer, err)
+				return ctrl.Result{}, err
+			} else {
+				err := r.updateStatus(ctx, dialer, specSecret.ResourceVersion)
+				return ctrl.Result{}, err
+			}
+		}
+
+		// Or, if the peering in Consul does exist, compare it to the contents of the spec's secret. If there's any
```
ishustava, line 165:
```suggestion
		// Or, if the peering in Consul does exist, compare it to the the spec's secret. If there's any
```

ishustava:
@ndhanushkodi will changelog be added separately or in this PR?

control-plane/api/v1alpha1/exportedservices_types.go
```diff
@@ -69,6 +68,8 @@ type ExportedService struct {
 type ServiceConsumer struct {
 	// Partition is the admin partition to export the service to.
 	Partition string `json:"partition,omitempty"`
+	// [Experimental] PeerName is the name of the peer to export the service to.
+	PeerName string `json:"peerName,omitempty"`
```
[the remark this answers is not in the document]
thisisnotashwin: we have matched the name from the consul config entry so far but I actually do prefer this.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -834,6 +834,176 @@ func serviceInstancesForK8SServiceNameAndNamespace(k8sServiceName, k8sServiceNam
...
+		upstream = api.Upstream{
+			DestinationType: api.UpstreamDestTypePreparedQuery,
+			DestinationName: preparedQuery,
+			LocalBindPort:   int(port),
+		}
+	}
+	return upstream
+}
+
+// processNonAnnotatedUpstream processes an upstream in the format:
```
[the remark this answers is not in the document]
thisisnotashwin: i think the implication is that it is not annotated with the type of each part. but I think it is confusing as annotation is being overloaded here.

.github/workflows/test.yml
thisisnotashwin: 1. yup

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -847,75 +1017,45 @@ func (r *EndpointsController) processUpstreams(pod corev1.Pod, endpoints corev1.
...
-						fallthrough
-					default:
-						serviceName = strings.TrimSpace(pieces[0])
-					}
-				} else {
-					serviceName = strings.TrimSpace(parts[0])
-				}
+			// parts separates out the port, and determines whether it's a prepared query or not, since parts[0] would
+			// be "prepared_query" if it is.
+			parts := strings.SplitN(raw, ":", 3)
```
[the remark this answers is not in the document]
thisisnotashwin: similarly, we should use regex but lets punt on this till the next release?

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -0,0 +1,377 @@
...
+			r.Log.Info("PeeringAcceptor was deleted, deleting from Consul", "name", req.Name, "ns", req.Namespace)
+			if err := r.deletePeering(ctx, req.Name); err != nil {
+				return ctrl.Result{}, err
+			}
+			if acceptor.Secret().Backend == "kubernetes" {
+				if err := r.deleteK8sSecret(ctx, acceptor); err != nil {
+					return ctrl.Result{}, err
+				}
+			}
+			controllerutil.RemoveFinalizer(acceptor, FinalizerName)
```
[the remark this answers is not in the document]
thisisnotashwin: maybe we only return an error at the end of the operations?
i can check explicitly for a not found error.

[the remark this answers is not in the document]
thisisnotashwin: soo..turns out delete is idempotent in consul and deleting a non existent acceptor does not error out (which is surprising)

control-plane/connect-inject/peering_acceptor_controller.go
thisisnotashwin: This was only for tests. good catch!

control-plane/connect-inject/consul_sidecar.go
```diff
@@ -11,13 +11,13 @@ import (
 // the metrics merging server when metrics merging feature is enabled.
 // It always disables service registration because for connect we no longer
 // need to keep services registered as this is handled in the endpoints-controller.
-func (h *Handler) consulSidecar(pod corev1.Pod) (corev1.Container, error) {
-	metricsPorts, err := h.MetricsConfig.mergedMetricsServerConfiguration(pod)
+func (w *MeshWebhook) consulSidecar(pod corev1.Pod) (corev1.Container, error) {
```
ishustava, line 14: ❤️

control-plane/api/v1alpha1/serviceintentions_types.go
```diff
@@ -455,14 +458,16 @@ func (in *ServiceIntentions) validateNamespaces(namespacesEnabled bool) field.Er
...
-			if source.Partition != "" {
-				errs = append(errs, field.Invalid(path.Child("sources").Index(i).Child("partition"), source.Partition, `Consul Enterprise Admin Partitions must be enabled to set source.partition`))
-			}
+	for i, source := range in.Spec.Sources {
+		if source.Partition != "" && !partitionsEnabled {
+			errs = append(errs, field.Invalid(path.Child("sources").Index(i).Child("partition"), source.Partition, `Consul Enterprise Admin Partitions must be enabled to set source.partition`))
+		}
+
+		if source.Peer != "" && source.Partition != "" {
+			errs = append(errs, field.Invalid(path.Child("sources").Index(i), source, `Both source.peer and source.partition cannot be set.`))
```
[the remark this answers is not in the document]
ishustava: Oh yes I do mean that! Thank you for reading my mind!

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -834,6 +834,176 @@ func serviceInstancesForK8SServiceNameAndNamespace(k8sServiceName, k8sServiceNam
...
+	var port int32
+	var upstream api.Upstream
+
+	parts := strings.SplitN(rawUpstream, ":", 3)
+
+	port, _ = portValue(pod, strings.TrimSpace(parts[1]))
+
+	service := parts[0]
+
+	pieces := strings.Split(service, ".")
```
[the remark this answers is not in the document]
ishustava: yep, sounds good!

control-plane/connect-inject/peering_acceptor_controller.go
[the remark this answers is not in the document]
ishustava: Oh yeah that is surprising, but I suppose is a good thing for this case since this will just work.

control-plane/connect-inject/peering_dialer_controller.go
ishustava: 🤔 In this case though we're getting the secret from the spec not status. And the call above on line 88 gets the secret. Then here we're checking if secret is nil and returning an error, so I'm wondering why not combine these two calls into one and just make `getSecret` return an error when the secret doesn't exist instead of returning an error here.

Maybe I'm misunderstanding something.

ishustava (approved):
Looks great! A couple of follow-up comments, but won't block approval as everything looks great!

❤️ ❤️ ❤️

control-plane/connect-inject/peering_dialer_controller.go
thisisnotashwin: Aah! sorry. I should have been specific. We use the method to get the status secret (L110) too. And if it is the first reconcile, it is ok if the secret doesn't exist, so it would be ok for getSecret to not return an error if the secret does not exist.

control-plane/connect-inject/endpoints_controller.go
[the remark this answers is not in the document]
ndhanushkodi: love it!!

control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -0,0 +1,780 @@
...
+	// Reconcile a resource that is not in K8s, but is still in Consul.
+	resp, err := pdc.Reconcile(context.Background(), ctrl.Request{
+		NamespacedName: namespacedName,
+	})
+	require.NoError(t, err)
+	require.False(t, resp.Requeue)
+
+	// After reconciliation, Consul should have a peering with its DeletedAt set.
+	peering, _, err := consulClient.Peerings().Read(context.Background(), "dialer-deleted", nil)
+	require.NotNil(t, peering.DeletedAt)
```
ndhanushkodi, line 554: Seeing the tests fail with a nil pointer exception here. I think its because the deletion logic was implemented in Consul, and the peering is sometimes being deleted before we can read it here. So I don't think we can make this assertion anymore. Maybe we can do a retry to check that the peering is nil? Since it's a deferred delete?

thisisnotashwin: I think you need to build a newer Consul image. I ran into the same issue but turns out, I had an older Consul binary.

ndhanushkodi: Hmm, I tried the retry to check that the peering is nil with a timer that waits for 5 seconds, and that also fails.

thisisnotashwin: THE DEFERRED DELETION!!! ok ok. that makes sense. thank you so much Nitya! that was an excellent catch.

thisisnotashwin (approved):
Woot woot! lets get this puppy merged today!!

control-plane/api/v1alpha1/serviceintentions_types.go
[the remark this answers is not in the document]
thisisnotashwin: heh! of course. but in this case, it is absolutely fine, because you could have an intention for a service in the same peer/partition as well in which case you would not need to provide that information. so I think there is a valid case for neither to be set.

PR #1277, opened 2022-06-14
[wip] Add support for configurable envoy proxy concurrency

kschoche:
Changes proposed in this PR:
- Adds a new flag to connect-inject deployment to take in a default concurrency value for every envoy sidecar.
- Adds support for a new annotation on connect injected pods `consul.hashicorp.com/consul-envoy-proxy-concurrency` which will override the default for any given pod.

I chose not to just force the user to add `--concurrency 42` to their `connectInject.sidecarProxy.extraArgs` because I thought it was a bit messy to have to edit strings in deployments vs just adding a new annotation to your deployment.

How I've tested this PR:
unit tests + acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -2069,7 +2069,7 @@ connectInject:
     # advised to keep this number low for sidecars and high for edge proxies.
     # This will control the `--concurrency` flag to Envoy.
     # This setting can be overridden on a per-pod basis via this annotation:
-    # - `consul.hashicorp.com/sidecar-concurrency`
+    # - `consul.hashicorp.com/consul-envoy-proxy-concurrency`
     # See: also https://blog.envoyproxy.io/envoy-threading-model-a8d44b922310
     # @type: string
     concurrency: 2
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -2063,6 +2063,16 @@ connectInject:
     secretKey: null
 
   sidecarProxy:
+    # The number of threads to be used by the Envoy proxy to manage worker threads. idecar proxy.
+    # By default the threading model of Envoy will use one thread per CPU core per envoy proxy. This
+    # leads to unnecessary thread and memory usage and leaves unnecessary idle connections open. It is
+    # advised to keep this number low for sidecars and high for edge proxies.
+    # This will control the `--concurrency` flag to Envoy.
+    # This setting can be overridden on a per-pod basis via this annotation:
+    # - `consul.hashicorp.com/sidecar-concurrency`
+    # See: also https://blog.envoyproxy.io/envoy-threading-model-a8d44b922310
+    # @type: string
+    concurrency: 2
     # Set default resources for sidecar proxy. If null, that resource won't
     # be set.
     # These settings can be overridden on a per-pod basis via these annotations:
--- a/control-plane/connect-inject/envoy_sidecar.go
+++ b/control-plane/connect-inject/envoy_sidecar.go
@@ -96,6 +96,20 @@ func (h *Handler) getContainerSidecarCommand(pod corev1.Pod, multiPortSvcName st
 		cmd = append(cmd, "--base-id", fmt.Sprintf("%d", multiPortSvcIdx))
 	}
 
+	// Check to see if the user has overriden concurrency via an annotation.
+	if pod.Annotations[annotationEnvoyProxyConcurrency] != "" {
+		_, err := strconv.ParseInt(pod.Annotations[annotationEnvoyProxyConcurrency], 10, 64)
+		if err != nil {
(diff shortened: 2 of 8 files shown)


charts/consul/values.yaml
```diff
@@ -2063,6 +2063,18 @@ connectInject:
...
+    # The number of worker threads to be used by the Envoy proxy.
+    # By default the threading model of Envoy will use one thread per CPU core per envoy proxy. This
+    # leads to unnecessary thread and memory usage and leaves unnecessary idle connections open. It is
+    # advised to keep this number low for sidecars and high for edge proxies.
+    # This will control the `--concurrency` flag to Envoy.
+    # For additional information see also: https://blog.envoyproxy.io/envoy-threading-model-a8d44b922310
+    #
+    # This setting can be overridden on a per-pod basis via this annotation:
+    # - `consul.hashicorp.com/consul-envoy-proxy-concurrency`
+    # @type: string
```
kschoche, line 2075:
```suggestion
    # @type: integer
```

control-plane/connect-inject/envoy_sidecar.go
```diff
@@ -96,6 +96,22 @@ func (h *Handler) getContainerSidecarCommand(pod corev1.Pod, multiPortSvcName st
 		cmd = append(cmd, "--base-id", fmt.Sprintf("%d", multiPortSvcIdx))
 	}
 
+	// Check to see if the user has overriden concurrency via an annotation.
+	if pod.Annotations[annotationEnvoyProxyConcurrency] != "" {
+		val, err := strconv.ParseInt(pod.Annotations[annotationEnvoyProxyConcurrency], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse annotation: %s", annotationEnvoyProxyConcurrency)
+		}
+		if val < 0 {
```
curtbushko, line 105: I looked in the envoy code and they are not setting an upper limit for the concurrency value (it is a uint32_t) so I think that only checking for >=0 is good enough.

control-plane/connect-inject/envoy_sidecar_test.go
```diff
@@ -280,6 +315,7 @@ func TestHandlerEnvoySidecar_EnvoyExtraArgs(t *testing.T) {
 			expectedContainerCommand: []string{
 				"envoy",
 				"--config-path", "/consul/connect-inject/envoy-bootstrap.yaml",
+				"--concurrency", "0",
```
curtbushko, line 318: Non-blocking nit: Should these all be set to 2 since that is the default?

curtbushko (commented):
Looks good! I only had some minor questions/nits but I think it all comes down to naming.

I prefer sidecar instead of envoy for consistency reasons. I think it is concurrency that is the word that makes it confusing as I immediately thought of replicas when I read the summary.

Would thread-concurrency make it clearer? -default-sidecar-proxy-thread-concurrency?

Naming is hard. 🥲

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ IMPROVEMENTS:
   * Enable configuring the pod topologySpreadConstraints for mesh, terminating, and ingress gateways. [[GH-1257](https://github.com/hashicorp/consul-k8s/pull/1257)]
   * Present Consul server CA chain when using Vault secrets backend. [[GH-1251](https://github.com/hashicorp/consul-k8s/pull/1251)]
   * API Gateway: Enable configuring of the new High Availability feature (requires Consul API Gateway v0.3.0+). [[GH-1261](https://github.com/hashicorp/consul-k8s/pull/1261)]
+  * Enable the configuration of Envoy proxy concurrency via `connectInject.sidecarProxy.concurrency` which can
```
kschoche, line 11:
```suggestion
  * Enable the configuration of Envoy proxy worker threads (concurrency) via `connectInject.sidecarProxy.concurrency` which can
```

kschoche: @curtbushko how's that look?

curtbushko: Looks good to me!

control-plane/connect-inject/envoy_sidecar_test.go
kschoche: An alternative approach would be for me to do something like this in every test:
```
	h := Handler{
				ImageConsul:    "hashicorp/consul:latest",
				ImageEnvoy:     "hashicorp/consul-k8s:latest",
				EnvoyExtraArgs: tc.envoyExtraArgs,
                                 EnvoyDefaultEnvoyConcurrency: 2,
			}
```
and then assert on it being "2" but its also kinda hacky, and I have a test already which validates being able to set the values and its defaults at the top of this test file.
Thoughts?

thisisnotashwin (approved):
NICE

curtbushko (approved):

PR #1282, opened 2022-06-15
Update mesh CRD with MeshHTTPConfig.

thisisnotashwin:
Changes proposed in this PR:
- Update the Mesh CRD with Mesh HTTP Config.
- Update acceptance API version.

How I've tested this PR:
- Unit tests.
- Deploying on local cluster.
- 👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


david-yu:
Thank you for doing this!

ndhanushkodi (approved):
Looks great!!

jmurret (approved):
Nice!  Thank you for doing this!

thisisnotashwin:
> Why does acceptance use API v1.10.something, when it was using v1.12.0? It went back 2 versions?

That is an excellent question. The versioning on the API project has gotten messed up which is why even though that SHA is from a commit that was pushed yesterday, the version still says 1.10. hopefully this is something we do fix after 1.13 but it looks pretty bad. we still are using the latest API changes though.

PR #1283, opened 2022-06-15
Allow building arm64 dev docker images

ishustava:
Changes proposed in this PR:

When using kind on an arm64 architecture locally
and installing consul-k8s with tproxy,
the service-mesh init container will fail to apply iptables rules.
This is because kind is running on an arm64 architecture but uses
amd64 docker image for consul-k8s. If we switch to arm64 images instead,
the error goes away.

How I've tested this PR:
built a dev image and deployed to kind running arm64. Then tried to deploy a service with tproxy enabled and saw that iptables are applied correctly

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


lkysow (approved):

thisisnotashwin (approved):

curtbushko (approved):
Thanks for fixing this! It works in my environment.

japui:
Permitir a criação de imagens do docker arm64 dev #

PR #1284, opened 2022-06-15
Add secret watchers on Peering Acceptor/Dialer Controllers

thisisnotashwin:
Changes proposed in this PR:
- Add watches to both the Peering Acceptor and Dialer controller. These watch secrets with a specific label and reconcile the associated acceptor/dialer when the secret that matches their status/spec secret is updated. Additionally, when the acceptor creates the token secret, it now adds that label to the created secret.

How I've tested this PR:
- Unit tests.
- Created a peering acceptor and dialer in different clusters. After the secret was created and copied over, deleted the contents of the secret in the secret on the acceptor side. This generated a new token. Copied this secret over again and observed the reconcile happen on the dialer end as well.

- 👀 
- attempt to perform the above as well.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/connect-inject/peering_acceptor_controller.go
+++ b/control-plane/connect-inject/peering_acceptor_controller.go
@@ -14,8 +14,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // PeeringAcceptorController reconciles a PeeringAcceptor object.
@@ -318,7 +323,11 @@ func (r *PeeringAcceptorController) deleteK8sSecret(ctx context.Context, accepto
 func (r *PeeringAcceptorController) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&consulv1alpha1.PeeringAcceptor{}).
-		Complete(r)
+		Watches(
+			&source.Kind{Type: &corev1.Secret{}},
+			handler.EnqueueRequestsFromMapFunc(r.requestsForPeeringTokens),
+			builder.WithPredicates(predicate.NewPredicateFuncs(r.filterPeeringAcceptors)),
+		).Complete(r)
 }
 
 // generateToken is a helper function that calls the Consul api to generate a token for the peer.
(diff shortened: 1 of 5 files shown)


control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -344,12 +353,55 @@ func (r *PeeringAcceptorController) deletePeering(ctx context.Context, peerName
 	return nil
 }
 
+// requestsForPeeringTokens creates a slice of requests for the peering acceptor controller.
+// It enqueues a request for each acceptor that needs to be reconciled. It iterates through
+// the list of acceptors and creates a request for the acceptor that has the same secret is it's
```
kschoche, line 358:
```suggestion
// the list of acceptors and creates a request for the acceptor that has the same secret as it's
```

control-plane/connect-inject/peering_acceptor_controller_test.go
```diff
@@ -943,3 +946,218 @@ func TestAcceptorUpdateStatusError(t *testing.T) {
...
+				},
+			},
+			result: []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Namespace: "test",
+						Name:      "peering-1",
+					},
+				},
+			},
```
kschoche, line 1145: Does it make sense to add a "does not match any acceptor" test that has a kubernetes backend?
I think the logic is tested through the other tests I suppose. Thoughts?

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -257,3 +266,43 @@ func (r *PeeringDialerController) deletePeering(ctx context.Context, peerName st
...
+// the list of dialers and creates a request for the dialer that has the same secret as it's
+// secret and that of the updated secret that is being watched.
+// We compare it to the secret in the spec as the resource is dependent on the secret.
+func (r *PeeringDialerController) requestsForPeeringTokens(object client.Object) []reconcile.Request {
+	r.Log.Info("received update for Peering Token Secret", "name", object.GetName())
+
+	// Get the list of all dialers.
+	var dialerList consulv1alpha1.PeeringDialerList
+	if err := r.Client.List(r.Context, &dialerList); err != nil {
+		r.Log.Error(err, "failed to list endpoints")
```
kschoche, line 281: is `failed to list endpoints` the correct message here (and also in the acceptor controller) ?

kschoche (approved):
Looks great!
I left a few questions/comments but nothing blocking.

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -257,3 +266,43 @@ func (r *PeeringDialerController) deletePeering(ctx context.Context, peerName st
...
+func (r *PeeringDialerController) filterPeeringDialers(object client.Object) bool {
+	secretLabels := object.GetLabels()
+	isPeeringToken, ok := secretLabels[labelPeeringToken]
+	if !ok {
+		return false
+	}
+	if isPeeringToken == "true" {
+		return true
+	}
+	return false
```
ndhanushkodi, line 307:
```suggestion
	return isPeeringToken == "true"
```

control-plane/connect-inject/peering_dialer_controller.go
```diff
@@ -257,3 +266,43 @@ func (r *PeeringDialerController) deletePeering(ctx context.Context, peerName st
...
 	return nil
 }
+
+// requestsForPeeringTokens creates a slice of requests for the peering dialer controller.
+// It enqueues a request for each dialer that needs to be reconciled. It iterates through
+// the list of dialers and creates a request for the dialer that has the same secret as it's
+// secret and that of the updated secret that is being watched.
+// We compare it to the secret in the spec as the resource is dependent on the secret.
+func (r *PeeringDialerController) requestsForPeeringTokens(object client.Object) []reconcile.Request {
+	r.Log.Info("received update for Peering Token Secret", "name", object.GetName())
```
ndhanushkodi, line 276: same nit from acceptor, could we add ns? 🙏

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -344,12 +353,55 @@ func (r *PeeringAcceptorController) deletePeering(ctx context.Context, peerName
...
+	var acceptorList consulv1alpha1.PeeringAcceptorList
+	if err := r.Client.List(r.Context, &acceptorList); err != nil {
+		r.Log.Error(err, "failed to list Peering Acceptors")
+		return []ctrl.Request{}
+	}
+	for _, acceptor := range acceptorList.Items {
+		if acceptor.SecretRef().Backend == "kubernetes" {
+			if acceptor.SecretRef().Name == object.GetName() && acceptor.Namespace == object.GetNamespace() {
+				return []ctrl.Request{{NamespacedName: types.NamespacedName{Namespace: acceptor.Namespace, Name: acceptor.Name}}}
+			}
```
ndhanushkodi, line 374: We currently don't do any validation on if there were multiple acceptors that had the same secret name/namespace, and in theory someone could get into that case, so maybe the logic here can stay the same, but maybe we'd want to add a duplicate check? 

It may not be in scope for this PR--

I'm imagining a label on the secret that describes the namespaced name of the acceptor that generated it, and then in the reconcile loop, erroring out if there's an existing secret with the same name and namespace in the acceptor spec, that references a different acceptor resource than the one being reconciled.

ndhanushkodi (approved):
This looks amazing!! I was wondering what your thoughts are on a duplicate check, but it doesn't need to be in scope of this PR.

control-plane/connect-inject/peering_acceptor_controller.go
thisisnotashwin: hmm.. that is a good point. I think it makes sense to move that into a validation? ill add that as a follow-up task on Asana. This can potentially be true for dialers as well, where multiple dialers have the same spec secret.

PR #1291, opened 2022-06-17
Updating retry from 10 to 60 seconds for cloud clean up because of AKS failure.

jmurret:
Changes proposed in this PR:
- Updating retry from 10 to 60 seconds for cloud clean up because of AKS failure.

If this does not solve things, we might even consider just doing like 5, 10, or 15 minutes since we are just trying to clean up and if its not ready, having a low number is not helpful when your signal gets corrupted with noise.

How I've tested this PR:
👀 
👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/framework/consul/helm_cluster.go
+++ b/acceptance/framework/consul/helm_cluster.go
@@ -201,9 +201,9 @@ func (h *HelmCluster) Destroy(t *testing.T) {
 			}
 		}
 	}
-	// Retry a few times because sometimes certain resources (like PVC) take time to delete
+	// Retry because sometimes certain resources (like PVC) take time to delete
 	// in cloud providers.
-	retry.RunWith(&retry.Counter{Wait: 1 * time.Second, Count: 10}, t, func(r *retry.R) {
+	retry.RunWith(&retry.Counter{Wait: 1 * time.Second, Count: 60}, t, func(r *retry.R) {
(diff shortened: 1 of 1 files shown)


curtbushko (approved):
It is worth a try!

acceptance/framework/consul/helm_cluster.go
```diff
@@ -201,9 +201,9 @@ func (h *HelmCluster) Destroy(t *testing.T) {
 			}
 		}
 	}
-	// Retry a few times because sometimes certain resources (like PVC) take time to delete
+	// Retry because sometimes certain resources (like PVC) take time to delete
 	// in cloud providers.
-	retry.RunWith(&retry.Counter{Wait: 1 * time.Second, Count: 10}, t, func(r *retry.R) {
+	retry.RunWith(&retry.Counter{Wait: 1 * time.Second, Count: 60}, t, func(r *retry.R) {
```
kschoche, line 206: I was wondering after merging this if 10 seconds was enough, and now I recall that its more like 10 minutes sometimes for these things to sort themselves :-/ 
Should we retry 600 times or just retry every 10 seconds, 60 times?
```suggestion
	retry.RunWith(&retry.Counter{Wait: 1 * time.Second, Count: 600}, t, func(r *retry.R) {
```

kschoche (approved):

PR #1303, opened 2022-06-23
Add annotations for enabling TLS on Envoy Prometheus endpoint

kyhavlov:
Changes proposed in this PR:
- Adds annotations for pointing to CA/cert/key files for Envoy to use to enable TLS on the Prometheus metrics endpoint. Corresponding Consul PR here: https://github.com/hashicorp/consul/pull/13481

How I've tested this PR:
- Updated a container_init test to make sure the CLI flags get passed through correctly to the Envoy bootstrap command.
- Testing this out manually with a CSI volume to make sure Envoy watches the cert paths correctly when they're updated on disk.

- The surface area of the change is pretty small; it should be covered by the updated unit test.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ FEATURES:
 
 IMPROVEMENTS:
 * Control Plane
-  * Added annotations for configuring TLS on Prometheus metrics endpoint for Envoy sidecars. [[GH-1303](https://github.com/hashicorp/consul-k8s/pull/1303)]
+  * Added annotations `consul.hashicorp.com/prometheus-ca-file`, `consul.hashicorp.com/prometheus-ca-path`, `consul.hashicorp.com/prometheus-cert-file`, and `consul.hashicorp.com/prometheus-key-file` for configuring TLS scraping on Prometheus metrics endpoints for Envoy sidecars. To enable, set the cert and key file annotations along with one of the ca file/path annotations. [[GH-1303](https://github.com/hashicorp/consul-k8s/pull/1303)]
 
 ## 0.45.0 (June 17, 2022)
 FEATURES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ FEATURES:
   * Add support for secret watchers on the Peering Acceptor and Peering Dialer controllers. [[GH-1284](https://github.com/hashicorp/consul-k8s/pull/1284)]
   * Add support for version annotation on the Peering Acceptor and Peering Dialer controllers. [[GH-1302](https://github.com/hashicorp/consul-k8s/pull/1302)]
 
+IMPROVEMENTS:
+* Control Plane
+  * Added annotations for configuring TLS on Prometheus metrics endpoint for Envoy sidecars. [[GH-1303](https://github.com/hashicorp/consul-k8s/pull/1303)]
+
 ## 0.45.0 (June 17, 2022)
 FEATURES:
 * [Experimental] Cluster Peering: Support Consul cluster peering, which allows service connectivity between two independent clusters.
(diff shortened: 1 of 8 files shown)


charts/consul/templates/client-daemonset.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.global.imageK8s }}{{ fail "global.imageK8s is not a valid key, use global.imageK8S (note the capital 'S')" }}{{ end -}}
 {{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
-{{- if (and (and .Values.global.tls.enabled .Values.global.tls.httpsOnly) (and .Values.global.metrics.enabled .Values.global.metrics.enableAgentMetrics))}}{{ fail "global.metrics.enableAgentMetrics cannot be enabled if TLS (HTTPS only) is enabled" }}{{ end -}}
+# {{- if (and (and .Values.global.tls.enabled .Values.global.tls.httpsOnly) (and .Values.global.metrics.enabled .Values.global.metrics.enableAgentMetrics))}}{{ fail "global.metrics.enableAgentMetrics cannot be enabled if TLS (HTTPS only) is enabled" }}{{ end -}}
```
kyhavlov, line 3: Does this check (and the one in server-statefulset.yml) need to be modified at all, or can it just be removed? Further down in this file we set the http port to 8501 if tls.enabled is `true`, so it seems like this can probably be deleted because now we'll support scraping both agent and Envoy prometheus metrics over TLS.

control-plane/connect-inject/container_init.go
```diff
@@ -214,6 +219,19 @@ func (w *MeshWebhook) containerInit(namespace corev1.Namespace, pod corev1.Pod,
 		data.PrometheusScrapePath = prometheusScrapePath
 		data.PrometheusBackendPort = mergedMetricsPort
 	}
+	// Pull the TLS config from the relevant annotations.
```
kschoche, line 222: Are all of these annotations required or just one?
If they are all required, or required in sets we should enforce that with some validation and errors, in which case we should also update the unit tests accordingly.

kyhavlov: Cert, key, and one of CA file or path is required - would that validation go in the client/server daemonset templates, or somewhere else?

david-yu:
Could we get some reviews on this from @thisisnotashwin and @ndhanushkodi. It looks like the feature is now merged on the Core side in main.

thisisnotashwin (approved):
Thanks so much for doing this Kyle!

control-plane/connect-inject/container_init.go
kschoche: Since it can be overridden by annotations I think we need it on both ends:
* in the template logic for the defaults
* here for the overrides

kyhavlov: I added the check/tests here - these won't be settable in the template/defaults though so this should be the only place the validation is needed.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ FEATURES:
   * Add support for secret watchers on the Peering Acceptor and Peering Dialer controllers. [[GH-1284](https://github.com/hashicorp/consul-k8s/pull/1284)]
   * Add support for version annotation on the Peering Acceptor and Peering Dialer controllers. [[GH-1302](https://github.com/hashicorp/consul-k8s/pull/1302)]
 
+IMPROVEMENTS:
+* Control Plane
+  * Added annotations for configuring TLS on Prometheus metrics endpoint for Envoy sidecars. [[GH-1303](https://github.com/hashicorp/consul-k8s/pull/1303)]
```
kschoche, line 10:
```suggestion
* Control Plane
  * Added annotations `consul.hashicorp.com/prometheus-ca-file`, `consul.hashicorp.com/prometheus-ca-path`, `consul.hashicorp.com/prometheus-cert-file`, and `consul.hashicorp.com/prometheus-key-file` for configuring TLS scraping on Prometheus metrics endpoints for Envoy sidecars. To enable, set the cert and key file annotations along with one of the ca file/path annotations. [[GH-1303](https://github.com/hashicorp/consul-k8s/pull/1303)]
```

kschoche (approved):
This looks great, thanks for doing this! I left just a couple comments which should help clear up the changelog so users can pick it up and run with it!

CHANGELOG.md
kschoche: @kyhavlov Should we also note that support for this requires `consul-x.x.x` ?

PR #1308, opened 2022-06-25
don't set securitycontext on openshift/okd

eb4x:
If running on OpenShift, these settings cause problems because the
user and group ids are below the range OpenShift expects. If not
specified, they're automatically handled by OpenShift.


kschoche (approved):
🥇 
Awesome work on this thank you!

nathancoleman (approved):

PR #1313, opened 2022-06-29
Tests

eb4x:
Various fixes and improvements to tests.

### DIFF
--- a/charts/consul/test/unit/client-daemonset.bats
+++ b/charts/consul/test/unit/client-daemonset.bats
@@ -1299,7 +1299,7 @@ local actual=$(echo $object |
       --set 'global.acls.manageSystemACLs=true' \
       --set 'global.tls.enabled=false' \
       . | yq '.spec.template.spec.initContainers[0].volumeMounts[] | select(.name=="consul-ca-cert")' | tee /dev/stderr)
-  [ "${actual}" == "" ]
+  [ "${object}" == "" ]
 }
 
 @test "client/DaemonSet: fail when externalServers is enabled but the externalServers.hosts is not provided" {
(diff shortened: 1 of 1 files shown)


jmurret:
PR for this is here: https://github.com/hashicorp/consul-k8s/pull/1314

eb4x:
> @eb4x for the changes that you are making to the capitalization of the release-name in the tests, which version of helm are you using? (just run `helm version`). The helm version causes this mismatch. I am running 3.8.2. I think our docs in pre-requisites might be outdated.

I was at 3.6.3, updated to 3.8.2 and those tests pass, so I'll just remove that commit. Thanks for the tip :+1:

eb4x:
Ok, I think I'm done looking at these tests. Leave a review :smiley_cat:

jmurret (changes requested):
We use this checksums to highlight when this configuration changes for us.  If we just that it is not null, this will always be true and it won't be apparent in PRs when this changes.  Can you please change this to take the new checksum that is in the failing test and update these test with the new checksum?

With this approach you will have to add these tests to the other PR.

eb4x:
> We use this checksums to highlight when this configuration changes for us. If we just that it is not null, this will always be true and it won't be apparent in PRs when this changes. Can you please change this to take the new checksum that is in the failing test and update these test with the new checksum?

Yep, it was probably acting up because of the helm version I was using. I've dropped it and moved the two others to the other PR.

So now this PR is just unrelated fixes/improvements that stood out.

eb4x (commented):
Dropped checksum changes

jmurret (approved):
Thank you for your contribution @eb4x! I'll merge these and then the other PR.

kschoche (approved):
Thanks for this PR!

PR #1315, opened 2022-06-30
Add annotations to support specifying userVolumes and userVolumeMounts for the envoy sidecar

kschoche:
Changes proposed in this PR:
- Adds 2 new annotations `consul.hashicorp.com/consul-sidecar-user-volume` `consul.hashicorp.com/consul-sidecar-user-volume-mount` that accept JSON objects that define Volumes and VolumeMounts which will be attached to the envoy sidecar.
- This will allow the user to pass in an arbitrary mount that can be used to for example mount the TLS certs/keys via a Kube secret, CSI volume, etc, for prometheus scraping.
- This should work in the same manner that istio's `sidecar.istio.io/userVolumeMount` and `sidecar.istio.io/userVolume`
annotations work.

How I've tested this PR:
* unit tests
* built a test image under `kyleschochenmaier/consul-k8s-prom-tls` and deployed consul-k8s with a demo application that used the annotation `consul.hashicorp.com/consul-sidecar-user-volume-mount: "[{\"name\": \"secrets-store-inline\", \"mountPath\": \"/mnt/secrets-store\"}]"` and confirmed the mount and secrets contents are attached to the envoy-sidecar.

unit tests

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/control-plane/connect-inject/annotations.go
+++ b/control-plane/connect-inject/annotations.go
@@ -86,6 +86,10 @@ const (
 	annotationConsulSidecarMemoryLimit   = "consul.hashicorp.com/consul-sidecar-memory-limit"
 	annotationConsulSidecarMemoryRequest = "consul.hashicorp.com/consul-sidecar-memory-request"
 
+	// annotations for sidecar volumes.
+	annotationConsulSidecarUserVolume      = "consul.hashicorp.com/consul-sidecar-user-volume"
+	annotationConsulSidecarUserVolumeMount = "consul.hashicorp.com/consul-sidecar-user-volume-mount"
+
 	// annotations for sidecar concurrency.
 	annotationEnvoyProxyConcurrency = "consul.hashicorp.com/consul-envoy-proxy-concurrency"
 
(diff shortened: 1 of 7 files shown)


control-plane/connect-inject/mesh_webhook_ent_test.go
```diff
@@ -30,7 +30,7 @@ func TestHandler_MutateWithNamespaces(t *testing.T) {
 
 	basicSpec := corev1.PodSpec{
 		Containers: []corev1.Container{
-			corev1.Container{
```
kschoche, line 33: The linter was forcing this.

control-plane/connect-inject/annotations.go
```diff
@@ -86,6 +86,10 @@ const (
 	annotationConsulSidecarMemoryLimit   = "consul.hashicorp.com/consul-sidecar-memory-limit"
 	annotationConsulSidecarMemoryRequest = "consul.hashicorp.com/consul-sidecar-memory-request"
 
+	// annotations for sidecar volumes.
+	annotationConsulSidecarUserVolume      = "consul.hashicorp.com/consul-sidecar-user-volume"
```
ndhanushkodi, line 90: clarifying q for my understanding: these will reference a volume that exists on the pod? and then mount that to the sidecar container as well?

ndhanushkodi (approved):

control-plane/connect-inject/annotations.go
kschoche: Correct!
The user will need to either:
* specify both annotations, which would add a volume to the pod + mount to the sidecar.
* specify only the volumeMount annotation and already have the volume defined in the pod's spec.

PR #1317, opened 2022-07-01
CNI Helm Charts

curtbushko:
This PR contains the base helm charts for installing CNI.

Changes proposed in this PR:

 - A daemonset for the cni installer
 - Service account for the cni plugin
 - Clusterrole and clusterrolebinding for the service account
 - Bats tests for all of the above.

Note: Helm changes coming in follow on PRs

- Openshift support (SCC)
- Disabling redirect-traffic in connect-inject

How I've tested this PR:

- bats tests
- manual installing in my kind cluster

👀

Checklist:

[x] Tests added
[ ] CHANGELOG entry added
    HashiCorp engineers only, community PRs should not add a changelog entry.
    Entries should use present tense (e.g. Add support for...)

### DIFF
--- a/charts/consul/templates/cni-clusterrole.yaml
+++ b/charts/consul/templates/cni-clusterrole.yaml
@@ -1,14 +1,24 @@
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: consul-cni
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Release.Namespace }}
   labels:
-    app: consul-cni
-    release: cni-poc
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: consul-cni 
 rules:
 - apiGroups: [""]
-  resources: ["pods"]
-  verbs: ["get", "list", "watch", "patch", "update"]
-- apiGroups: [""]
-  resources: ["nodes"]
-  verbs: ["get", "list", "watch", "patch", "update"]
+  resources: 
+  - pods
+  verbs: 
+  - get
+  - list
+  - watch
+  - patch
+  - update
+{{- end }}
--- a/charts/consul/templates/cni-clusterrole.yaml
+++ b/charts/consul/templates/cni-clusterrole.yaml
@@ -10,7 +10,7 @@ metadata:
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-    component: consul-cni 
+    component: cni 
 rules:
 - apiGroups: [""]
   resources: 
--- a/charts/consul/templates/cni-clusterrole.yaml
+++ b/charts/consul/templates/cni-clusterrole.yaml
@@ -1,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: consul-cni
+  labels:
+    app: consul-cni
+    release: cni-poc
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch", "patch", "update"]
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["get", "list", "watch", "patch", "update"]
--- a/charts/consul/templates/cni-clusterrole.yaml
+++ b/charts/consul/templates/cni-clusterrole.yaml
@@ -1,5 +1,4 @@
-{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
-{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.enabled) }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
--- a/charts/consul/templates/cni-clusterrole.yaml
+++ b/charts/consul/templates/cni-clusterrole.yaml
@@ -1,4 +1,4 @@
-{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.enabled) }}
+{{- if .Values.connectInject.cni.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
--- a/charts/consul/templates/cni-clusterrolebinding.yaml
+++ b/charts/consul/templates/cni-clusterrolebinding.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: consul-cni
+  labels:
+    app: consul-cni
+    release: cni-poc
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: consul-cni
+subjects:
+- kind: ServiceAccount
+  name: consul-cni
+  namespace: consul
--- a/charts/consul/templates/cni-clusterrolebinding.yaml
+++ b/charts/consul/templates/cni-clusterrolebinding.yaml
@@ -1,5 +1,4 @@
-{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
-{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.enabled) }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
--- a/charts/consul/templates/cni-clusterrolebinding.yaml
+++ b/charts/consul/templates/cni-clusterrolebinding.yaml
@@ -17,5 +17,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: {{ template "consul.fullname" . }}-cni
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.connectInject.cni.namespace }}
 {{- end }}
--- a/charts/consul/templates/cni-clusterrolebinding.yaml
+++ b/charts/consul/templates/cni-clusterrolebinding.yaml
@@ -9,7 +9,7 @@ metadata:
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-    component: auth-method
+    component: consul-cni
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
--- a/charts/consul/templates/cni-clusterrolebinding.yaml
+++ b/charts/consul/templates/cni-clusterrolebinding.yaml
@@ -9,7 +9,7 @@ metadata:
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-    component: consul-cni
+    component: cni
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
--- a/charts/consul/templates/cni-clusterrolebinding.yaml
+++ b/charts/consul/templates/cni-clusterrolebinding.yaml
@@ -16,5 +16,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: {{ template "consul.fullname" . }}-cni
-  namespace: {{ .Values.connectInject.cni.namespace }}
+  namespace: {{ .Release.Namespace }}
 {{- end }}
--- a/charts/consul/templates/cni-clusterrolebinding.yaml
+++ b/charts/consul/templates/cni-clusterrolebinding.yaml
@@ -1,15 +1,21 @@
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: consul-cni
+  name: {{ template "consul.fullname" . }}-cni
   labels:
-    app: consul-cni
-    release: cni-poc
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: auth-method
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: consul-cni
+  name: {{ template "consul.fullname" . }}-cni
 subjects:
 - kind: ServiceAccount
-  name: consul-cni
-  namespace: consul
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Release.Namespace }}
+{{- end }}
--- a/charts/consul/templates/cni-clusterrolebinding.yaml
+++ b/charts/consul/templates/cni-clusterrolebinding.yaml
@@ -1,4 +1,4 @@
-{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.enabled) }}
+{{- if .Values.connectInject.cni.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
--- a/charts/consul/templates/cni-daemonset.yaml
+++ b/charts/consul/templates/cni-daemonset.yaml
@@ -72,9 +72,9 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
           volumeMounts:
-            - mountPath: /host{{ .Values.connectInject.cni.cniBinDir }}
+            - mountPath: {{ .Values.connectInject.cni.cniBinDir }}
               name: cni-bin-dir
-            - mountPath: /host{{ .Values.connectInject.cni.cniNetDir }}
+            - mountPath: {{ .Values.connectInject.cni.cniNetDir }}
               name: cni-net-dir
       volumes:
         # Used to install CNI.
--- a/charts/consul/templates/cni-daemonset.yaml
+++ b/charts/consul/templates/cni-daemonset.yaml
@@ -1,5 +1,5 @@
-{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
-{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.enabled)) }}{{ fail "connectInject.enabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.enabled) }}
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
--- a/charts/consul/templates/cni-daemonset.yaml
+++ b/charts/consul/templates/cni-daemonset.yaml
@@ -1,0 +1,76 @@
+{{- if .Values.global.imageK8s }}{{ fail "global.imageK8s is not a valid key, use global.imageK8S (note the capital 'S')" }}{{ end -}}
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled=true" }}{{ end -}}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: consul-cni-node
+  namespace: consul
+  labels:
+    k8s-app: consul-cni-node
+    release: cni-poc
+spec:
+  selector:
+    matchLabels:
+      k8s-app: consul-cni-node
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
+  template:
+    metadata:
+      labels:
+        k8s-app: consul-cni-node
+        sidecar.consul.io/inject: "false"
+      annotations:
+        sidecar.consul.io/inject: "false"
+        # Add Prometheus Scrape annotations
+        prometheus.io/scrape: 'true'
+        prometheus.io/port: "15014"
+        prometheus.io/path: '/metrics'
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
+      tolerations:
+        # Make sure consul-cni-node gets scheduled on all nodes.
+        - effect: NoSchedule
+          operator: Exists
+        # Mark the pod as a critical add-on for rescheduling.
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - effect: NoExecute
+          operator: Exists
+      priorityClassName: system-node-critical
+      serviceAccountName: consul-cni
+      # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
+      # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
+      terminationGracePeriodSeconds: 5
+      containers:
+        # This container installs the consul CNI binaries
+        # and CNI network config file on each node.
+        - name: install-cni
+          image: "curtbushko/cni-poc:0.2"
+          # readinessProbe:
+          #   httpGet:
+          #     path: /readyz
+          #     port: 8000
+          securityContext:
+            runAsGroup: 0
+            runAsUser: 0
+            runAsNonRoot: false
+            privileged: true 
+          command: ["/bin/sh", "-ec", "/bin/consul-k8s install-cni", "-cni-net-dir=/etc/cni/net.d/multus.d", "-multus=true"]
+          volumeMounts:
+            - mountPath: /host/opt/cni/bin
+              name: cni-bin-dir
+            - mountPath: /host/etc/cni/net.d
+              name: cni-net-dir
+      volumes:
+        # Used to install CNI.
+        - name: cni-bin-dir
+          hostPath:
+            path: "/opt/cni/bin"
+        - name: cni-net-dir
+          hostPath:
+            path: "/etc/cni/net.d"
+{{- end }}
--- a/charts/consul/templates/cni-daemonset.yaml
+++ b/charts/consul/templates/cni-daemonset.yaml
@@ -5,7 +5,7 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: {{ template "consul.fullname" . }}-cni
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.connectInject.cni.namespace }}
   labels:
     app: {{ template "consul.name" . }}
     chart: {{ template "consul.chart" . }}
--- a/charts/consul/templates/cni-daemonset.yaml
+++ b/charts/consul/templates/cni-daemonset.yaml
@@ -11,7 +11,7 @@ metadata:
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-    component: consul-cni 
+    component: cni 
 spec:
   {{- if .Values.connectInject.cni.updateStrategy }}
   updateStrategy:
@@ -22,17 +22,18 @@ spec:
       app: {{ template "consul.name" . }}
       chart: {{ template "consul.chart" . }}
       release: {{ .Release.Name }}
-      component: consul-cni
+      component: cni
   template:
     metadata:
       labels:
         app: {{ template "consul.name" . }}
         chart: {{ template "consul.chart" . }}
         release: {{ .Release.Name }}
-        component: consul-cni
+        component: cni
       annotations:
         consul.hashicorp.com/connect-inject: "false"
     spec:
+      # consul-cni only runs on linux operating systems
       nodeSelector:
         kubernetes.io/os: linux
       tolerations:
--- a/charts/consul/templates/cni-daemonset.yaml
+++ b/charts/consul/templates/cni-daemonset.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.global.imageK8s }}{{ fail "global.imageK8s is not a valid key, use global.imageK8S (note the capital 'S')" }}{{ end -}}
 {{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
 {{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
 apiVersion: apps/v1
@@ -37,9 +36,6 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
       tolerations:
-        # Make sure consul-cni-node gets scheduled on all nodes.
-        - effect: NoSchedule
-          operator: Exists
         # Mark the pod as a critical add-on for rescheduling.
         - key: CriticalAddonsOnly
           operator: Exists
--- a/charts/consul/templates/cni-daemonset.yaml
+++ b/charts/consul/templates/cni-daemonset.yaml
@@ -4,7 +4,7 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: {{ template "consul.fullname" . }}-cni
-  namespace: {{ .Values.connectInject.cni.namespace }}
+  namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "consul.name" . }}
     chart: {{ template "consul.chart" . }}
@@ -56,7 +56,7 @@ spec:
         - name: install-cni
           image: {{ .Values.global.imageK8S }}
           securityContext:
-            privileged: {{ .Values.connectInject.cni.privileged }}
+            privileged: true
           command:
           - "/bin/sh"
           - "-ec"
--- a/charts/consul/templates/cni-daemonset.yaml
+++ b/charts/consul/templates/cni-daemonset.yaml
@@ -1,33 +1,37 @@
 {{- if .Values.global.imageK8s }}{{ fail "global.imageK8s is not a valid key, use global.imageK8S (note the capital 'S')" }}{{ end -}}
-{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled=true" }}{{ end -}}
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
 {{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
-  name: consul-cni-node
-  namespace: consul
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Release.Namespace }}
   labels:
-    k8s-app: consul-cni-node
-    release: cni-poc
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: consul-cni 
 spec:
+  {{- if .Values.connectInject.cni.updateStrategy }}
+  updateStrategy:
+    {{ tpl .Values.connectInject.cni.updateStrategy . | nindent 4 | trim }}
+  {{- end }}
   selector:
     matchLabels:
-      k8s-app: consul-cni-node
-  updateStrategy:
-    type: RollingUpdate
-    rollingUpdate:
-      maxUnavailable: 1
+      app: {{ template "consul.name" . }}
+      chart: {{ template "consul.chart" . }}
+      release: {{ .Release.Name }}
+      component: consul-cni
   template:
     metadata:
       labels:
-        k8s-app: consul-cni-node
-        sidecar.consul.io/inject: "false"
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: consul-cni
       annotations:
-        sidecar.consul.io/inject: "false"
-        # Add Prometheus Scrape annotations
-        prometheus.io/scrape: 'true'
-        prometheus.io/port: "15014"
-        prometheus.io/path: '/metrics'
+        consul.hashicorp.com/connect-inject: false
     spec:
       nodeSelector:
         kubernetes.io/os: linux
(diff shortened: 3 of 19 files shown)


charts/consul/templates/cni-clusterrole.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
```
kschoche, line 2: I *think* this second line is redundant? On line 1 we fail if `cni is enabled && tproxy is disabled`

kschoche: nevermind! this would otherwise render always! :)

curtbushko: No, I don't thinks so.  

Throw an error if cni.enabled=true and transparentProxy.defaultEnabled=false

If both cni.enabled=true and transparentProxy.defaultEnabled=true continue on.

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,89 @@
...
+        {{- toYaml .Values.connectInject.cni.securityContext | nindent 8 -}}
+      {{- end }}
+      # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
+      # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
+      terminationGracePeriodSeconds: 5
+      containers:
+        # This container installs the consul CNI binaries and CNI network config file on each node
+        - name: install-cni
+          image: {{ .Values.global.imageK8S }}
+          securityContext:
```
kschoche, line 61: Do we need to make any changes to this securityContext block based on openshift?

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,89 @@
...
+              -cni-net-dir={{ .Values.connectInject.cni.cniNetDir }}
+          {{- with .Values.connectInject.cni.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            - mountPath: /host{{ .Values.connectInject.cni.cniBinDir }}
+              name: cni-bin-dir
+            - mountPath: /host{{ .Values.connectInject.cni.cniNetDir }}
+              name: cni-net-dir
```
kschoche, line 80: hostPath volumeMounts should always be readOnly whenever possible, are we able to add that to the volumeMount?  (please double check my indentation)
https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

```suggestion
          volumeMounts:
            - mountPath: /host{{ .Values.connectInject.cni.cniBinDir }}
              name: cni-bin-dir
              readOnly: true              
            - mountPath: /host{{ .Values.connectInject.cni.cniNetDir }}
              name: cni-net-dir
              readOnly: true              
```
```suggestion
          volumeMounts:
            - mountPath: /host{{ .Values.connectInject.cni.cniBinDir }}
              name: cni-bin-dir
            - mountPath: /host{{ .Values.connectInject.cni.cniNetDir }}
              name: cni-net-dir
```

kschoche: Also I have no idea how that second suggestion appeared, I didnt type it!!

curtbushko: They cannot be because we are writing to the file system.

charts/consul/templates/cni-daemonset.yaml
kschoche: Cool!

charts/consul/templates/cni-resourcequota.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.global.imageK8s }}{{ fail "global.imageK8s is not a valid key, use global.imageK8S (note the capital 'S')" }}{{ end -}}
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: consul-cni-resource-quota
```
kschoche, line 7: Just double-checking the name:
```suggestion
  name: {{ template "consul.fullname" . }}-cni
```

charts/consul/templates/cni-clusterrole.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: ClusterRole
+metadata:
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: consul-cni
```
jmurret, line 13: we don't use the consul prefix on our other components.

```suggestion
    component: cni 
```

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,89 @@
...
+kind: DaemonSet
+metadata:
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Values.connectInject.cni.namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: consul-cni
```
jmurret, line 14: same as above

```suggestion
    component: cni 
```

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,89 @@
...
+  {{- if .Values.connectInject.cni.updateStrategy }}
+  updateStrategy:
+    {{ tpl .Values.connectInject.cni.updateStrategy . | nindent 4 | trim }}
+  {{- end }}
+  selector:
+    matchLabels:
+      app: {{ template "consul.name" . }}
+      chart: {{ template "consul.chart" . }}
+      release: {{ .Release.Name }}
+      component: consul-cni
```
jmurret, line 25:
```suggestion
      component: cni
```

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,89 @@
...
+  template:
+    metadata:
+      labels:
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: consul-cni
```
jmurret, line 32:
```suggestion
        component: cni
```

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,89 @@
...
+              -multus={{ .Values.connectInject.cni.multus }} \
+              -log-level={{ default .Values.global.logLevel .Values.connectInject.cni.logLevel }} \
+              -cni-bin-dir={{ .Values.connectInject.cni.cniBinDir }} \
+              -cni-net-dir={{ .Values.connectInject.cni.cniNetDir }}
+          {{- with .Values.connectInject.cni.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            - mountPath: /host{{ .Values.connectInject.cni.cniBinDir }}
```
jmurret, line 77: [nit/future idea]I see you have a values comment that this needs to start with a `/`.  Is it worth it to add a template function that prepends a `/` if it does not start with one?

charts/consul/templates/cni-serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: ServiceAccount
+metadata:
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Values.connectInject.cni.namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: consul-cni
```
jmurret, line 13:
```suggestion
    component:cni 
```

charts/consul/templates/cni-resourcequota.yaml
kschoche: One more thing, does it make sense to add the `metadata.Labels` field like the other resources?

```
  labels:
    release: {{ .Release.Name }}
```

curtbushko: It sure does!  Thanks for catching that!

charts/consul/values.yaml
```diff
@@ -1,4 +1,4 @@
-# Available parameters and their default values for the Consul chart.
+# Available parameters and tienteir default values for the Consul chart.
```
jmurret, line 1: is `tienteir` a thing?

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,89 @@
...
+      labels:
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: consul-cni
+      annotations:
+        consul.hashicorp.com/connect-inject: "false"
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
```
kschoche, line 37: QQ: is this a hard requirement due to CNI spec, or other? Maybe we can put a comment in place?

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,86 @@ connectInject:
...
+    # The security context for the cni installer daemonset. This should be a YAML map corresponding to a
+    # Kubernetes [SecurityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) object.
+    # By default, servers will run as non-root, with user ID `100` and group ID `1000`,
+    # which correspond to the consul user and group created by the Consul docker image.
+    # Note: if running on OpenShift, this setting is ignored because the user and group are set automatically
+    # by the OpenShift platform.
+    # @type: map
+    # @recurse: false
+    securityContext:
+      runAsNonRoot: false
```
jmurret, line 1921: These values do not match what you have in the comments.

jmurret (commented):
This looks really good.  I left a few possible small changes and questions, but nothing big.

charts/consul/templates/cni-daemonset.yaml
curtbushko: I think it is required because of the paths and iptables command...

charts/consul/templates/cni-clusterrole.yaml
curtbushko: This is a good catch. I will update to just be 'cni'

charts/consul/templates/cni-daemonset.yaml
curtbushko: I added a comment to the section about this only working on linux.

charts/consul/templates/cni-daemonset.yaml
kschoche: +1 to this, I found it a bit wonky to read and was hoping we could have a cleaner looking option.

curtbushko: The `/etc/cni/net.d` is the absolute path and I would find it weird to see people do `etc/cni/net.d` instead. Technically, I guess we could add the `/` but I am not sure if that is great. I would rather point them to some docs where they could see the CNI compatibility and what they need to setup to run CNI on different cloud platforms. For example, GKE requires that the cniBinDir is different...

There will be validation in the installer itself also...

charts/consul/test/unit/cni-daemonset.bats
```diff
@@ -0,0 +1,254 @@
...
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" ]]
+}
+
+@test "cni/DaemonSet: image defaults to global.imageK8S" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/cni-daemonset.yaml  \
+      --set 'connectInject.cni.enabled=true' \
+      --set 'global.imageK8S=foo' \
```
kschoche, line 48: I wonder if something isn't plumbed correctly here because most of these tests do not have `tproxy.DefaultEnabled` set and if I understand correctly from the daemonset templating the daemonset should NOT be rendered at all without:
`{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}`
?

i.e. we need `      --set 'connectInject.transparentProxy.defaultEnabled=true' \`
for all of the below tests?

kschoche: Ah, okay that makes sense.
Usually we are explicit about the flags being passed, but I think this is fine!

charts/consul/values.yaml
curtbushko: It is a Canadian word!

charts/consul/templates/cni-daemonset.yaml
jmurret: 👍 if we find a lot of people confused by this, we could always add a Helm fail.  Resolving the convo.

charts/consul/templates/cni-clusterrole.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
```
ishustava, line 1: Why do we require tproxy to be default enabled? I don't think it needs to be since CNI could be installed with tproxy default disabled and read tproxy annotations on the pod or the namespace.

ishustava: We should gate it on or fail if connectInject is disabled.

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,91 @@
+{{- if .Values.global.imageK8s }}{{ fail "global.imageK8s is not a valid key, use global.imageK8S (note the capital 'S')" }}{{ end -}}
```
ishustava, line 1: why do we need the image check here?

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,91 @@
...
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: cni
+      annotations:
+        consul.hashicorp.com/connect-inject: "false"
+    spec:
+      # consul-cni only runs on linux operating systems
+      nodeSelector:
+        kubernetes.io/os: linux
```
ishustava, line 38: I think all of our components run on linux only so maybe it's redundant?

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,91 @@
...
+      tolerations:
+        # Make sure consul-cni-node gets scheduled on all nodes.
+        - effect: NoSchedule
+          operator: Exists
```
ishustava, line 42: Why do we need this? Isn't daemonset responsible for scheduling on all nodes?

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,91 @@
...
+        # Mark the pod as a critical add-on for rescheduling.
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - effect: NoExecute
+          operator: Exists
```
ishustava, line 47: why do we need these?

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,91 @@
...
+        {{- toYaml .Values.connectInject.cni.securityContext | nindent 8 -}}
+      {{- end }}
+      # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
+      # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
+      terminationGracePeriodSeconds: 5
+      containers:
+        # This container installs the consul CNI binaries and CNI network config file on each node
+        - name: install-cni
+          image: {{ .Values.global.imageK8S }}
+          securityContext:
```
ishustava, line 62: Why are we applying two separate security contexts for the pod and container? Should we consolidate them?

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,91 @@
...
+      # Tell kubernetes that this daemonset is critical so that it will be scheduled on a new node before other pods
+      priorityClassName: system-node-critical
+      serviceAccountName: {{ template "consul.fullname" . }}-cni
+      {{- if not .Values.global.openshift.enabled }}
+      securityContext:
+        {{- toYaml .Values.connectInject.cni.securityContext | nindent 8 -}}
+      {{- end }}
+      # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
+      # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
+      terminationGracePeriodSeconds: 5
```
ishustava, line 57: I understand that we'd want to minimize downtime, but force deletion doesn't sound good. Why do we want to do a force deletion?

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,91 @@
...
+              -log-level={{ default .Values.global.logLevel .Values.connectInject.cni.logLevel }} \
+              -cni-bin-dir={{ .Values.connectInject.cni.cniBinDir }} \
+              -cni-net-dir={{ .Values.connectInject.cni.cniNetDir }} \
+              -dns-prefix={{ template "consul.fullname" . }}
+          {{- with .Values.connectInject.cni.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            - mountPath: /host{{ .Values.connectInject.cni.cniBinDir }}
```
ishustava, line 79: why do we need a `/host` prefix?

charts/consul/templates/cni-resourcequota.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if .Values.global.imageK8s }}{{ fail "global.imageK8s is not a valid key, use global.imageK8S (note the capital 'S')" }}{{ end -}}
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.transparentProxy.defaultEnabled) }}
+apiVersion: v1
+kind: ResourceQuota
```
ishustava, line 5: We don't define ResourceQuota for any other components. Why do we need this for the CNI?

charts/consul/templates/cni-serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: ServiceAccount
+metadata:
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Values.connectInject.cni.namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: cni
```
ishustava, line 13: need to add support for `imagePullSecrets`: https://github.com/hashicorp/consul-k8s/blob/023007f6b830a08df876557a2cd858120eb46975/charts/consul/templates/connect-inject-serviceaccount.yaml#L17-L21

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,85 @@ connectInject:
     # Note: This value has no effect if transparent proxy is disabled on the pod.
     defaultOverwriteProbes: true
 
+  # Configures consul-cni plugin for Consul Service mesh services
+  # Using this feature requires Consul 1.13.0-beta+
```
ishustava, line 1863: Why does it require consul 1.13?

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,85 @@ connectInject:
...
+  cni:
+    # If true, then all traffic redirection setup with use the consul-cni plugin
```
ishustava, line 1865:
```suggestion
    # If true, then all traffic redirection setup will use the consul-cni plugin
```

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,85 @@ connectInject:
...
+    # @type: boolean
+    enabled: false
+
+    # Namespace to install cni plugin.
+    # @type: string
+    namespace: kube-system
```
ishustava, line 1871: Why is it kube-system and not the same as the helm chart?

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,85 @@ connectInject:
...
+ 
+    # If multus plugin is enabled or not. If true, the cni installer will generate a consul cni configuration
```
ishustava, line 1873: nit: We should use proper capitalization everywhere since these are used to generate docs.
```suggestion
    # If multus plugin is enabled or not. If true, the CNI installer will generate a consul CNI configuration
```

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,85 @@ connectInject:
...
+    # in a location for the multus program to use. Also, the helm chart will create a custom resource for
+    # NetworkAttachmentDefintion that multis. cniBinDir and cniNetDir must be changed to work with multus.
```
ishustava, line 1875: I think this needs to be rephrased?

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,85 @@ connectInject:
...
+    # Please refer to your cloud platform for more informaton on using multus.
```
ishustava, line 1876: Are there any official docs for multus we can link to?

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,85 @@ connectInject:
...
+    # Location on the kubernetes node where the CNI plugin is installed. Shoud be the absolute path and start with a '/'
+    # @type: string
+    cniBinDir: "/opt/cni/bin"
+
+    # Location on the kubernetes node of all CNI configuration. Should be the absolute path and start with a '/'
+    # @type: string
+    cniNetDir: "/etc/cni/net.d"
+
+    # The daemonset should in privileged mode on all kubernetes platforms as it needs to interact with the
+    # kubernetes host file system
```
ishustava, line 1893: If it needs to always be privileged, then why do we allow to set `privileged` to `false`?

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -135,6 +135,11 @@ spec:
                 {{- else }}
                 -default-enable-transparent-proxy=false \
                 {{- end }}
+                {{- if .Values.connectInject.cni.enabled }}
+                -enable-cni=true \
+                {{- else }}
+                --enable-cni=false \
+                {{- end }}
```
kschoche, line 142: If you default the flag to `false` in `command.go` you could get away with just passing `-enable-cni=true` when it is enabled.
In either case you've got a typo on the `-enable-cni=false` line (2 hyphens)

curtbushko: I do have the command set to false by default but I was following the `if .Values.connectInject.transparentProxy.defaultEnabled ` pattern above (and just copying it).

charts/consul/values.yaml
curtbushko: I was following Istios pattern of allowing for a toggle. I am assuming that someone somewhere would want to toggle it off. Maybe that is premature optimization...

charts/consul/values.yaml
curtbushko: There is very little that I can find. There is the official "how to install multus on your kubernetes cluster" docs and at least 1 for Istio.  I think that we will need a learn guide for using Multus (on OpenShift most likely) and I want to write that post CNI. We have at least 1 customer that could use a doc for it.

charts/consul/values.yaml
curtbushko: I cleaned it up a bit.

charts/consul/values.yaml
curtbushko: Thank you!  I have updated them to be capitalized unless they were mentioning a specific setting  (ie cniBinDir) or a binary (ie consul-cni)

charts/consul/values.yaml
curtbushko: I updated the comments.

charts/consul/templates/cni-clusterrole.yaml
curtbushko: Ok, this was my miss-understanding then since CNI is a continuation of tproxy. I will update this and a bunch of bats tests.

charts/consul/templates/cni-daemonset.yaml
curtbushko: I was just copying the pattern from another template. I guess that one will catch it and this is redundant. I shall remove it in the couple places that I had it.

charts/consul/templates/cni-clusterrolebinding.yaml
```diff
@@ -0,0 +1,21 @@
...
+    release: {{ .Release.Name }}
+    component: cni
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "consul.fullname" . }}-cni
+subjects:
+- kind: ServiceAccount
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Values.connectInject.cni.namespace }}
```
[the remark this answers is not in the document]
curtbushko: I am following the pattern of other CNI plugins. Calico, cilium, istio, and multus all put it in kube-system. Linkerd is the only one that I could see that put it in the release namespace.

charts/consul/templates/cni-daemonset.yaml
curtbushko: Might be redundant but it is following the pattern that other CNI helm charts use. You make the call and I will remove it as I am on the fence. 😄

charts/consul/templates/cni-daemonset.yaml
curtbushko: I think you are right. I will remove it.

charts/consul/templates/cni-daemonset.yaml
curtbushko: This is so that kubernetes will make an attempt to schedule the daemonset before all other pods on the server. From what I have read, Kubernetes hasn't decided yet how it wants to handle critical pod scheduling so both `CriticalAddonsOnly` and `priorityClassName: system-node-critical` are two separate half-baked attempts to set some sort of priority. Both of those settings came about while I was researching bugs that other plugins had encountered...

charts/consul/templates/cni-daemonset.yaml
curtbushko: Mainly for ease of testing as I do not know how to test for the absence of an entry when there are subvalues there...

If it was:
```
      securityContext:
        {{- toYaml .Values.connectInject.cni.securityContext | nindent 8 -}}
      {{- end }}
        privileged: {{ .Values.connectInject.cni.privileged }}
```

I cannot test for the empty part if openshift.enabled is false.

The bats test:
```
@test "cni/DaemonSet: securityContext is not set when global.openshift.enabled=true" {
  cd `chart_dir`
  local actual=$(helm template \
      -s templates/cni-daemonset.yaml  \
      --set 'connectInject.cni.enabled=true' \
      --set 'global.openshift.enabled=true' \
      . | tee /dev/stderr |
      yq -r '.spec.template.spec.securityContext' | tee /dev/stderr)
  [ "${actual}" = "null" ]
}

```

charts/consul/templates/cni-daemonset.yaml
curtbushko: My worry, maybe unfounded, is that I do not want any chance that 2 daemonset pods could write to the filesystem at the same time. Maybe it is too aggressive and would have the negative effect of the config cleanup not running. 

Note: Istio uses 5 seconds, Calico is 0 seconds, our client daemonset uses 10. Cilium & Linkerd do not even set it.

charts/consul/templates/cni-daemonset.yaml
curtbushko: I thought that hostPath and mountPath had to be different. I just looked it up and they can be the same. Maybe I will switch it as it might make the code a little easier to comprehend also...

charts/consul/templates/cni-resourcequota.yaml
curtbushko: I haven't found an answer for all Kubernetes versions but I do know that GKE wouldn't let me run the daemonset in the kube-system namespace unless I set a ResourceQuota for it.

charts/consul/templates/cni-serviceaccount.yaml
curtbushko: Thanks!  (and done)

charts/consul/values.yaml
curtbushko: Thanks for catching that!

charts/consul/values.yaml
curtbushko: Explanations are above.

charts/consul/templates/cni-clusterrole.yaml
curtbushko: I have updated all of the templates to check for connectInject.cni.enabled + connectInject.enabled, updated all the bats tests and removed that imageK8s check.

Thanks for looking at this and catching these things!

charts/consul/templates/cni-clusterrolebinding.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.transparentProxy.defaultEnabled)) }}{{ fail "connectInject.transparentProxy.defaultEnabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
```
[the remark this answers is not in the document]
curtbushko: The check is now only in the cni-daemonset.

charts/consul/templates/cni-clusterrolebinding.yaml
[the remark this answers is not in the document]
ishustava: Ah interesting! If there's no known reason to be in kube-system, then I think it'd be more consistent if we put it in the release namespace.

charts/consul/templates/cni-daemonset.yaml
ishustava: >My worry, maybe unfounded, is that I do not want any chance that 2 daemonset pods could write to the filesystem at the same time
Ah I see! The downside is that we don't give the process enough time to terminate. I think we should maybe test it and think about how much time will this process need (doesn't have to be for this PR), especially that there is no consistent value between the different CNIs.  

>Cilium & Linkerd do not even set it.

If you don't set it, it defaults to 30sec.

charts/consul/values.yaml
ishustava: Ah ok, got it!

charts/consul/values.yaml
curtbushko: I will remove the toggle.

charts/consul/test/unit/cni-daemonset.bats
```diff
@@ -0,0 +1,250 @@
...
+  local actual=$(helm template \
+      -s templates/cni-daemonset.yaml  \
+      --set 'connectInject.cni.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'connectInject.cni.cniNetDir=bar' \
+      . | tee /dev/stderr |
+      yq -r -c '.spec.template.spec.containers[0].volumeMounts[] | select(.name == "cni-net-dir")' | tee /dev/stderr)
+      [ "${actual}" = '{"mountPath":"/hostbar","name":"cni-net-dir"}' ]
+}
+
```
ishustava, line 250: need tests for command flags

charts/consul/test/unit/cni-serviceaccount.bats
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "cni/daemonset: disabled by default" {
```
ishustava, line 5: same as above: these should be for ServiceAccount

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,76 @@ connectInject:
...
 
+  # Configures consul-cni plugin for Consul Service mesh services
+  cni:
+    # If true, then all traffic redirection setup with use the consul-cni plugin
+    # @type: boolean
+    enabled: false
+ 
+    # If multus plugin is enabled or not. If true, the CNI installer will generate a consul CNI configuration
+    # in a location for the multus program to use. Also, the helm chart will create a custom resource for
+    # NetworkAttachmentDefintion that multis will consume. The values for cniBinDir and cniNetDir must also be
```
ishustava, line 1870:
```suggestion
    # NetworkAttachmentDefintion that multus will consume. The values for cniBinDir and cniNetDir must also be 
```

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,76 @@ connectInject:
...
+  # Configures consul-cni plugin for Consul Service mesh services
+  cni:
+    # If true, then all traffic redirection setup with use the consul-cni plugin
+    # @type: boolean
+    enabled: false
+ 
+    # If multus plugin is enabled or not. If true, the CNI installer will generate a consul CNI configuration
+    # in a location for the multus program to use. Also, the helm chart will create a custom resource for
+    # NetworkAttachmentDefintion that multis will consume. The values for cniBinDir and cniNetDir must also be 
+    # changed for multus to work.
```
ishustava, line 1871: Could you describe how they need to be changed?

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,76 @@ connectInject:
...
+      requests:
+        memory: "50Mi"
+        cpu: "50m"
+      limits:
+        memory: "50Mi"
+        cpu: "50m"
+
+    # Resource quotas for running the daemonset as system critical pods
+    resourceQuota:
+      pods: 5000
```
ishustava, line 1901: do we still need this? I thought you said we only needed it if running in `kube-system` namespace

charts/consul/values.yaml
ishustava: The comment is still the same.

charts/consul/values.yaml
```diff
@@ -1859,6 +1859,76 @@ connectInject:
...
+    # Kubernetes [SecurityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) object.
+    # By default, servers will run as root, with user ID `0` and group ID `0`.
+    # Note: if running on OpenShift, this setting is ignored because the user and group are set automatically
+    # by the OpenShift platform.
+    # @type: map
+    # @recurse: false
+    securityContext:
+      runAsNonRoot: false
+      runAsGroup: 0
+      runAsUser: 0
```
ishustava, line 1913: Why does installer need to be root by default. Also why do we ignore it for openshift. If it needs to be root and openshift will set it to some non-default user, will things break?

charts/consul/templates/cni-daemonset.yaml
curtbushko: I think these two settings are doing 2 different things but are very similar.

- CriticalAddonsOnly: will make sure that a pod gets scheduled before other pods. Think of a scenario where 3 pods were restarted. consul-cni will get scheduled before the other pods.
- system-node-critical: will make sure that the consul-cni daemonset is started on the node immediately on startup.

I think both cases may be handled by different schedulers/controllers.  

I don't think I can test this unless I create a big cluster with a lot of pods running and just watch which pods startup first.

charts/consul/templates/cni-daemonset.yaml
curtbushko: In my testing, the cleanup is almost instantaneous. To be on the safe side, I have set it to 10.

charts/consul/templates/cni-daemonset.yaml
curtbushko: I have now changed this and it works. It has made the code a bit simpler too!  Thanks for finding it!

charts/consul/templates/cni-resourcequota.yaml
curtbushko: kube-system was a red herring. `system-node-critical` the setting that require resource quotas.

charts/consul/values.yaml
curtbushko: kube-system has been removed and the section is not required.

charts/consul/values.yaml
curtbushko: The toggle was removed.

charts/consul/templates/cni-daemonset.yaml
```diff
@@ -0,0 +1,87 @@
+{{- if (and (.Values.connectInject.cni.enabled) (not .Values.connectInject.enabled)) }}{{ fail "connectInject.enabled must be true if connectInject.cni.enabled is true" }}{{ end -}}
+{{- if (and .Values.connectInject.cni.enabled .Values.connectInject.enabled) }}
```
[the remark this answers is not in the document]
curtbushko: Done.

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -135,6 +135,11 @@ spec:
                 {{- else }}
                 -default-enable-transparent-proxy=false \
                 {{- end }}
+                {{- if .Values.connectInject.cni.enabled }}
+                -enable-cni=true \
+                {{- else }}
+                -enable-cni=false \
+                {{- end }}
```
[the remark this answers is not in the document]
curtbushko: I changed only this part to clean it up.

charts/consul/test/unit/cni-clusterrole.bats
```diff
@@ -0,0 +1,65 @@
...
+      --set 'connectInject.cni.enabled=false' \
+      --set 'connectInject.enabled=true' \
+      -s templates/cni-daemonset.yaml  \
+      .
+}
+
+#--------------------------------------------------------------------
+# rules
+
+@test "cni/clusterrole: sets get, list, watch, patch, update to pods in all api groups" {
```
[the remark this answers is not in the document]
curtbushko: These have been removed.

charts/consul/test/unit/cni-clusterrole.bats
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "cni/daemonset: disabled by default" {
```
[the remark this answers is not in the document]
curtbushko: cut and paste error.  😄

charts/consul/test/unit/cni-daemonset.bats
```diff
@@ -0,0 +1,250 @@
...
+load _helpers
+
+@test "cni/daemonset: disabled by default" {
+  cd `chart_dir`
+  assert_empty helm template \
+      -s templates/cni-daemonset.yaml  \
+      .
+}
+
+@test "cni/daemonset: enabled with connectInject.cni.enabled=true and connectInject.enabled=true" {
```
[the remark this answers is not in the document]
curtbushko: I think I got all of these...

charts/consul/test/unit/cni-clusterrolebinding.bats
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "cni/daemonset: disabled by default" {
```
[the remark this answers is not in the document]
curtbushko: I think I found all of these too

charts/consul/test/unit/cni-daemonset.bats
```diff
@@ -0,0 +1,250 @@
...
+      --set 'connectInject.enabled=true' \
+      --set 'connectInject.cni.securityContext.runAsNonRoot=true' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.securityContext' | tee /dev/stderr)
+
+  local actual=$(echo $security_context | jq -r .runAsNonRoot)
+  [ "${actual}" = "true" ]
+}
+
+@test "cni/DaemonSet: sets default privileged security context settings" {
```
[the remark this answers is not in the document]
curtbushko: Removed.

charts/consul/values.yaml
curtbushko: I added an example for GKE to the comment.

charts/consul/values.yaml
curtbushko: Yes we do.

charts/consul/values.yaml
curtbushko: I think you need to be root so that you can write to cniBinDir and cniNetDir. I believe privileged in openshift gets around this.  This setting was done while I was researching a bunch of bugs that other CNI plugins encounter.

charts/consul/templates/cni-daemonset.yaml
ishustava: ah got it! thanks for the clarification!

charts/consul/templates/cni-podsecuritypolicy.yaml
```diff
@@ -0,0 +1,31 @@
...
+    component: cni 
+spec:
+  privileged: true
+  # Required to prevent escalations to root.
+  allowPrivilegeEscalation: false
+  volumes:
+  - hostPath
+  - secret
+  - emptyDir
+  hostNetwork: true
```
ishustava, line 21: do we need hostNetwork?

charts/consul/templates/cni-resourcequota.yaml
ishustava: ah got it!

charts/consul/values.yaml
```diff
@@ -1876,6 +1875,83 @@ connectInject:
...
+    # Log level for the installer and plugin. Overrides global.logLevel
+    # @type: string
+    logLevel: null 
+    
+    # Location on the kubernetes node where the CNI plugin is installed. Shoud be the absolute path and start with a '/'
+    # Example on GKE:
+    #
+    # ```yaml
+    # cniBinDir: "/home/kubernetes/bin"
+    # ```
```
ishustava, line 1902: Hmm I'm not sure if this will included in the helm reference docs by the generator. Are these docs meant to be for the `cniBinDir`?

charts/consul/values.yaml
```diff
@@ -1876,6 +1875,83 @@ connectInject:
...
+  cni:
+    # If true, then all traffic redirection setup will use the consul-cni plugin. 
+    # Requires connectInject.enabled to also be true.
+    # @type: boolean
+    enabled: false
+ 
+    # If multus plugin is enabled or not. If true, the CNI installer will generate a consul CNI configuration
+    # in a location for the multus program to use. Also, the helm chart will create a custom resource for
+    # NetworkAttachmentDefintion that multus will consume. The values for cniBinDir and cniNetDir must also be 
+    # changed for multus to work. There is a GKE example for cniBinDir below.
```
ishustava, line 1888: I think one thing that's still confusing from the example is whether the example is for multus or for GKE or maybe both?

charts/consul/values.yaml
ishustava: got it, thanks!

ishustava (approved):
Great work on this Curt! This PR was very easy to review. Thanks so much for answering all my questions and doing the research!!

I left a few more comments, but they are not blocking.

charts/consul/values.yaml
curtbushko: You are correct. I removed the GKE reference as it doesn't make sense. Configuration on other platforms and for plugins will need it's own docs as they will be too big to fit in a values file.

jmurret (approved):

Issue #1309, opened 2022-06-28
The Consul k8s component auth method name in UI is not the same with the one in the 'consul-k8s-control-plane acl-init' command when specify the  global.federation.primaryDatacenter value


mntforever:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

The consul controller and mesh gateway are crashed when running helm with the global.federation.enabled = true and the global.federation.primaryDatacenter = "dc1".
- The component auth name in consul is: **dc1-consul-k8s-component-auth-method**
- The component auth name controller/mesh gateway needed is: **dc1-consul-k8s-component-auth-method-dc1**

In the helm chart template, we can see that:
[consul-k8s](https://github.com/hashicorp/consul-k8s/tree/v0.45.0)/[charts](https://github.com/hashicorp/consul-k8s/tree/v0.45.0/charts)/[consul](https://github.com/hashicorp/consul-k8s/tree/v0.45.0/charts/consul)/[templates](https://github.com/hashicorp/consul-k8s/tree/v0.45.0/charts/consul/templates)/controller-deployment.yaml
<img width="1026" alt="Screen Shot 2022-06-28 at 13 51 52" src="https://user-images.githubusercontent.com/31396681/176112803-ef8e6839-72d1-4630-90a1-1e9d59db0a32.png">

[consul-k8s](https://github.com/hashicorp/consul-k8s/tree/v0.45.0)/[charts](https://github.com/hashicorp/consul-k8s/tree/v0.45.0/charts)/[consul](https://github.com/hashicorp/consul-k8s/tree/v0.45.0/charts/consul)/[templates](https://github.com/hashicorp/consul-k8s/tree/v0.45.0/charts/consul/templates)/mesh-gateway-deployment.yaml
<img width="1169" alt="Screen Shot 2022-06-28 at 13 53 12" src="https://user-images.githubusercontent.com/31396681/176113143-ebf71d6a-7373-42c5-b7ce-7a5e086c3cfa.png">

[...]

PR #1318, opened 2022-07-01
Updating values docs to only set  global.federation.primaryDatacenter outside of the primary datacenter.

jmurret:
Related to issues found in https://github.com/hashicorp/consul-k8s/issues/1309 when setting this in the primary datacenter.

Changes proposed in this PR:
- Updating values docs to only set  global.federation.primaryDatacenter outside of the primary datacenter.

How I've tested this PR:
N/A

👀


curtbushko (approved):
Looks good!

PR #1323, opened 2022-07-05
Enable connect-injector to talk to external servers.

ishustava:
With agentless, connect-injector now needs to be able to talk to servers directly even when they live outside the cluster. Previously, connect-injector always needed to talk to clients and so this deployment did not need to worry about external servers. 

Changes proposed in this PR:
- Enable connect-injector talk to external servers. This involves providing both acl-init and inject-connect commands with external server configuration
- The `inject-connect` command has been changed to accept external server configuration similar to other commands (e.g. `server-acl-init`) that already support external servers

How I've tested this PR:
acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -96,13 +96,15 @@ spec:
                   name: {{ .Values.connectInject.aclInjectToken.secretName }}
                   key: {{ .Values.connectInject.aclInjectToken.secretKey }}
             {{- end }}
+            {{- if not .Values.externalServers.enabled }}
             - name: CONSUL_HTTP_ADDR
               {{- if .Values.global.tls.enabled }}
               value: https://{{ template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc:8501
               {{- else }}
               value: http://{{ template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc:8500
             {{- end }}
-            {{- if (and .Values.global.tls.enabled (not .Values.externalServers.useSystemRoots)) }}
+            {{- end }}
+            {{- if (and .Values.global.tls.enabled (not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots))) }}
             - name: CONSUL_CACERT
               {{- if .Values.global.secretsBackend.vault.enabled }}
               value: "/vault/secrets/serverca.crt"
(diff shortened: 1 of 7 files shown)


control-plane/subcommand/inject-connect/command.go
```diff
@@ -238,9 +251,21 @@ func (c *Command) Run(args []string) int {
...
+	hclogger, err := common.Logger(c.flagLogLevel, c.flagLogJSON)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Unable to create logger: %s", err))
+		return 1
+	}
+	serverAddresses, err := common.GetResolvedServerAddresses(c.flagServerAddresses, c.providers, hclogger)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Unable to discover any Consul addresses from %q: %s", c.flagServerAddresses[0], err))
+		return 1
+	}
```
ishustava, line 265: This code is not unit-tested because the testing is involved and it will be switched out for the go-netaddr library before merging to `main`. I've added the cloud auto-join code here for consistency because currently `externalServers.hosts` value supports it.

thisisnotashwin (approved):
This is excellent!! i love the acceptance test using the same cluster for server while treating it as an external server!! it's brilliant!!

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -96,13 +96,15 @@ spec:
                   name: {{ .Values.connectInject.aclInjectToken.secretName }}
                   key: {{ .Values.connectInject.aclInjectToken.secretKey }}
             {{- end }}
+            {{- if not .Values.externalServers.enabled }}
             - name: CONSUL_HTTP_ADDR
               {{- if .Values.global.tls.enabled }}
               value: https://{{ template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc:8501
               {{- else }}
               value: http://{{ template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc:8500
             {{- end }}
```
kschoche, line 105: super-nit: but I think the alignment is off on this one? maybe its my screen.
```suggestion
              {{- end }}
```

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -364,6 +381,18 @@ spec:
...
               -partition={{ .Values.global.adminPartitions.name }} \
               {{- end }}
+              {{- if .Values.externalServers.enabled }}
+              {{- if .Values.global.tls.enabled }}
+              -use-https \
+              {{- end }}
+              {{- range .Values.externalServers.hosts }}
+              -server-address={{ quote . }} \
+              {{- end }}
+              -server-port={{ .Values.externalServers.httpsPort }} \
```
kschoche, line 391: Same question here, is server-port always the httpsPort in this configuration?

kschoche (approved):
This looks great! I love the acceptance test too!
I left a couple questions but otherwise this looks good!

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -125,6 +127,19 @@ spec:
...
                 -listen=:8080 \
+                {{- if and .Values.externalServers.enabled (not .Values.externalServers.hosts) }}{{ fail "externalServers.hosts must be set if externalServers.enabled is true" }}{{ end -}}
+                {{- if .Values.externalServers.enabled }}
+                {{- if .Values.global.tls.enabled }}
+                -use-https \
+                {{- end }}
+                {{- range .Values.externalServers.hosts }}
+                -server-address={{ quote . }} \
+                {{- end }}
+                -server-port={{ .Values.externalServers.httpsPort }} \
```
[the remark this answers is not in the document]
ishustava: Yeah good question! I'll need to consolidate this logic in the future to make it more consistent and also so we don't have to keep adding flags to each command. I'll add it as a todo in the future but for now will keep it as is because if you set this flag when TLS is disabled, it had no effect, so user should see no difference.

charts/consul/templates/connect-inject-deployment.yaml
ishustava: oh sorry; missed this one! I'll fix it in my next PR

PR #1326, opened 2022-07-07
Enable CRD controller to talk to Consul servers

ishustava:
Changes proposed in this PR:
- Change CRD controller to talk to servers instead of clients so that it can work with agentless

How I've tested this PR:
acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -344,7 +344,7 @@ spec:
         - mountPath: /consul/login
           name: consul-data
           readOnly: false
-        {{- if .Values.global.tls.enabled }}
+        {{- if and .Values.global.tls.enabled (not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots))}}
         - name: consul-ca-cert
           mountPath: /consul/tls/ca
           readOnly: true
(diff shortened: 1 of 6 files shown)


thisisnotashwin (approved):
This is excellent!! This was super easy to review as well!! :)

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -364,6 +364,18 @@ spec:
               {{- if .Values.global.adminPartitions.enabled }}
               -partition={{ .Values.global.adminPartitions.name }} \
               {{- end }}
+              {{- if .Values.externalServers.enabled }}
+              {{- if .Values.global.tls.enabled }}
```
jmurret, line 368: [non-blocking] would would you think about putting this in a template since it seems to be repeated quite a bit?

jmurret (approved):
🎉 This looks great.  I had one suggestion that is non-blocking.  Great work.

charts/consul/templates/connect-inject-deployment.yaml
ishustava: yeah great question! I think I'll actually be able to get rid of this init container from all components altogether once agentless is done. My plan is to move login/logout logic into controllers to do at startup and shutdown. Other components that currently use this will switch to consul-dataplane instead doing the login for them.

PR #1328, opened 2022-07-08
Fixes for AWS acceptance tests cleanup

ishustava:
* Fix a bug with ELB deletion where we were comparing pointer values for VPC ID instead of strings
* Delete VPCs and their resources even if there's no associated EKS cluster. This so that if EKS cleanup fails in the middle and we
  re-run it, it will continue deleting resources it couldn't on a previous run.
* Fix a bug with deleting security group rules where we tried to delete them if there are no rules (AWS API will return an error in that case)

How I've tested this PR:
ran it locally

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


thisisnotashwin (approved):
Thanks so much for doing this Iryna!

kschoche (approved):
Looks great, thanks!

PR #1334, opened 2022-07-10
Parameterize client agent hostPort to prevent conflicts with existing daemonset services.

dvpatel:
Changes proposed in this PR:
- Option to parameterize hostPort client agent values for http, https, grpc and serflan
- Introduced client.agentPorts parameter in helm
- Use default hostPort values if client.agentPorts is not provided

How I've tested this PR:
Tested as follows:
-  Created client agent with non-default hostPorts using client.agentPorts parameter
-  Ran consul member command to ensure datacenter is coming up with servers and clients
-  Ran consul kv command to ensure key and values can be added using client agent URL
-  Ran negative tests by removing leader and follower pods to ensure datacenter resiliency

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


lkysow:
Hi @dvpatel there's a lot more required here because all our consul-k8s components assume certain ports. Since this is such a big change I think it makes more sense to create a feature issue and let us prioritize it.

dvpatel:
> Hi @dvpatel there's a lot more required here because all our consul-k8s components assume certain ports. Since this is such a big change I think it makes more sense to create a feature issue and let us prioritize it.

Hi @lkysow , I appreciate you getting back.  Who would create the feature issue?  Also, how long does it typically take to prioritize and enable this feature?  The reason I ask is if it takes "too long", then I'm perfectly fine taking the fork as an interim solution.  

From my observation and testing, the communication between the client and servers within the datacenter looks to be fine even though it's all within the same k8s cluster.  Also there is no server to server communications between multiple datacenters.  Not part of my use case.  Given this, are there other technical pitfalls I should be aware of with using a non-default hostPort?

Thanks for your time.
-Dipesh

lkysow:
> Who would create the feature issue?

If you have time then you can create it, otherwise I can.

> The reason I ask is if it takes "too long", then I'm perfectly fine taking the fork as an interim solution.

If you need it right away then I'd suggest going with a fork. We actually have plans this year to remove the need for the Consul clients which means there will no longer be a need for a host port, so that's why it doesn't make sense on us working on this right now.

dvpatel:
Hi Luke.  I was following up from our conversation last year.  I see multiple updates and wanted to know if Consul was refactored to make the client either optional or removed all together.

PR #1336, opened 2022-07-11
Envoy Debugging End-to-End Tests

t-eckert:
Changes proposed in this PR:
- Allow for passing in any output buffer for the UI
- Allow passing in a buffer for terminal output
- Test end-to-end for list command
- Test end-to-end for read command

How I've tested this PR:

- Added two end-to-end unit tests

- Run unit tests

Checklist:
- [x] Tests added

### DIFF
--- a/cli/cmd/version/version.go
+++ b/cli/cmd/version/version.go
@@ -1,8 +1,6 @@
 package version
 
 import (
-	"sync"
-
 	"github.com/hashicorp/consul-k8s/cli/common"
 	"github.com/hashicorp/consul-k8s/cli/common/terminal"
 )
@@ -12,8 +10,6 @@ type Command struct {
 
 	// Version is the Consul on Kubernetes CLI version.
 	Version string
-
-	once sync.Once
 }
 
 // Run prints the version of the Consul on Kubernetes CLI.


cli/cmd/proxy/list/command.go
```diff
@@ -192,7 +192,7 @@ func (c *ListCommand) fetchPods() ([]v1.Pod, error) {
 func (c *ListCommand) output(pods []v1.Pod) {
 	if c.flagAllNamespaces {
 		c.UI.Output("Namespace: All Namespaces\n")
-	} else {
+	} else if c.namespace != "" {
```
t-eckert, line 195: This just avoids the problem of printing "Namespace :" if there is no namespace set.

cli/common/terminal/ui.go
```diff
@@ -66,31 +65,6 @@ type Input struct {
 	Secret bool
 }
 
-// Interpret decomposes the msg and arguments into the message, style, and writer.
-func Interpret(msg string, raw ...interface{}) (string, string, io.Writer) {
```
t-eckert, line 70: I had to move this to be a method because it needs access to the buffer it's going to write to as a default (but that buffer may be overwritten in this function). It reduces reusability, but I feel it's unlikely we have more that one output implementation for this project. (e.g. the CLI will always output to a command line/terminal)

cli/cmd/uninstall/uninstall_test.go
```diff
@@ -356,6 +357,7 @@ func getInitializedCommand(t *testing.T) *Command {
 
 	baseCommand := &common.BaseCommand{
 		Log: log,
+		UI:  terminal.NewBasicUI(context.TODO()),
```
t-eckert, line 360: Added this because this assignment was deferred before I changed the UI to allow writing into any buffer. This just creates a UI that writes to `stdout`.

cli/cmd/proxy/list/command_test.go
```diff
@@ -13,23 +16,25 @@ import (
...
 
 	c := &ListCommand{
 		BaseCommand: baseCommand,
 	}
 	c.init()
-	return c
+	return c, buffer
 }
 
 func TestArgumentParsing(t *testing.T) {
```
curtbushko, line 40: A nit. In other commands, ie control-plane/subcommand/*, these are called flags instead of arguments. Are they called arguments all through the CLI?

cli/cmd/proxy/list/command_test.go
```diff
@@ -234,3 +239,86 @@ func TestFetchPods(t *testing.T) {
...
+
+	c, buf := getInitializedCommand(t)
+	c.kubernetes = client
+
+	out := c.Run([]string{"-A"})
+	require.Equal(t, 0, out)
+
+	actual := buf.String()
+
+	for _, expression := range expected {
```
curtbushko, line 321: I *really* like this. I am definitely going to use it in the future!

cli/cmd/proxy/read/command_test.go
```diff
@@ -35,30 +38,96 @@ func TestArgumentParsing(t *testing.T) {
...
+		"server.*server\\.default\\.dc1\\.internal\\.bc3815c2-1a0f-f3ff-a2e9-20d791f08d00\\.consul/.*2022-05-24T17:41:59\\.078Z",
+
+		"==> Secrets \\(2\\)",
+		"Name.*Type.*Last Updated",
+		"default.*Dynamic Active.*2022-05-24T17:41:59.078Z",
+		"ROOTCA.*Dynamic Warming.*2022-03-15T05:14:22.868Z",
+	}
+
+	// A fetchConfig function that just returns the test Envoy config.
+	mockFetchConfig := func(context.Context, common.PortForwarder) (*EnvoyConfig, error) {
```
curtbushko, line 95: What is the difference between a mock and a fake?

curtbushko (commented):
Looks pretty good to me. I have have some little questions/suggestions but nothing major. 

I really like the use of regex in places.

cli/cmd/proxy/list/command_test.go
t-eckert: Makes sense. I'll swap it for flag.

cli/cmd/proxy/list/command_test.go
t-eckert: Thanks! I wanted a solution that wasn't too brittle.

cli/cmd/proxy/list/command_test.go
```diff
@@ -13,23 +16,25 @@ import (
 )
 
 // getInitializedCommand sets up a command struct for tests.
-func getInitializedCommand(t *testing.T) *ListCommand {
+func getInitializedCommand(t *testing.T) (*ListCommand, *bytes.Buffer) {
 	t.Helper()
```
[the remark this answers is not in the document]
t-eckert: Good thinking! Moved.

cli/cmd/proxy/list/command_test.go
```diff
@@ -13,23 +16,25 @@ import (
...
 	log := hclog.New(&hclog.LoggerOptions{
 		Name:   "test",
 		Level:  hclog.Debug,
 		Output: os.Stdout,
 	})
 
+	buffer := new(bytes.Buffer)
 	baseCommand := &common.BaseCommand{
 		Log: log,
+		UI:  terminal.NewUI(context.TODO(), buffer),
```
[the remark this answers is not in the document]
t-eckert: Thanks! I changed to Background.

cli/cmd/proxy/read/command_test.go
t-eckert: Good question. I actually think this is more a "fake" than a "mock" after doing some research. 

There doesn't seem to be any formal definition, but there is some consensus that:
- a mock represents an external _something_ like an API or DB that is hit by the code and tracks that it is reached. 
- a fake is a stand in for some part of the code that obviates some external call, long running process, or something you just don't want running in the test.

I think this is actually a fake.

cli/cmd/proxy/list/command_test.go
```diff
@@ -234,3 +217,108 @@ func TestFetchPods(t *testing.T) {
...
+
+	actual := buf.String()
+
+	for _, expression := range expected {
+		require.Regexp(t, expression, actual)
+	}
+}
+
+// getInitializedCommand sets up a command struct for tests.
+func getInitializedCommand(t *testing.T) (*ListCommand, *bytes.Buffer) {
```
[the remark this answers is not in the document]
thisisnotashwin: Additionally, if the init() method is a noOp, we shouldn't invoke it here as it is confusing. We should also rename this method as no initialization is taking place. At this point, I feel like this behavior can just be inlined into each test to simplify how the tests read.

cli/common/base.go
```diff
@@ -39,6 +39,7 @@ func (c *BaseCommand) Close() error {
 
 // Init should be called FIRST within the Run function implementation.
 func (c *BaseCommand) Init() {
-	ui := terminal.NewBasicUI(c.Ctx)
-	c.UI = ui
+	// TODO: with the new UI setup, this is a no-op and should be removed in a
+	// later cleanup.
```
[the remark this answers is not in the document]
thisisnotashwin: Thanks for the comment!! But if it's mostly removal of code, I think it is ok to merge it into a single PR.

cli/common/terminal/basic.go
```diff
@@ -15,15 +15,22 @@ import (
...
+	bufOut io.Writer
 }
 
+// NewBasicUI creates a new instance of the basicUI struct for the terminal
+// with color output.
 func NewBasicUI(ctx context.Context) *basicUI {
-	return &basicUI{
-		ctx: ctx,
-	}
+	return &basicUI{ctx: ctx, bufOut: color.Output}
```
thisisnotashwin, line 27: this can be resue NewUI(ctx, color.Output)

cli/common/terminal/basic.go
```diff
@@ -147,3 +157,31 @@ func (ui *basicUI) NamedValues(rows []NamedValue, opts ...Option) {
 func (ui *basicUI) OutputWriters() (io.Writer, io.Writer, error) {
 	return os.Stdout, os.Stderr, nil
 }
+
+// interpret decomposes the msg and arguments into the message, style, and writer.
+func (ui *basicUI) interpret(msg string, raw ...interface{}) (string, string, io.Writer) {
```
thisisnotashwin, line 162: `parse` is usually the verbiage utilized for similar behavior. `interpret` seemed a little confusing.

cli/common/terminal/basic.go
```diff
@@ -147,3 +157,31 @@ func (ui *basicUI) NamedValues(rows []NamedValue, opts ...Option) {
...
+	// Build our args and options
+	var args []interface{}
+	var opts []Option
+	for _, r := range raw {
+		if opt, ok := r.(Option); ok {
+			opts = append(opts, opt)
+		} else {
+			args = append(args, r)
+		}
```
thisisnotashwin, line 171: nicely done!

thisisnotashwin (commented):
This looks good! I feel like there are a couple of changes that could be made to improve the readability of the tests a little bit that I have commented on.

cli/common/base.go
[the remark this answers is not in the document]
t-eckert: Cool, did the removal.

cli/cmd/proxy/read/command_test.go
```diff
@@ -35,30 +38,96 @@ func TestArgumentParsing(t *testing.T) {
...
-			c := getInitializedCommand(t)
+			c, _ := getInitializedCommand(t)
 			c.kubernetes = fake.NewSimpleClientset()
 			out := c.Run(tc.args)
 			require.Equal(t, tc.out, out)
 		})
 	}
 }
 
+func TestEndToEnd(t *testing.T) {
```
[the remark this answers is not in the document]
t-eckert: Cool, I made the change to `TestReadCommandOutput`. I see sometimes we use the `_` in test names, but is there a consistent time when we apply that styling?

curtbushko (approved):
Looks good to me!

PR #1337, opened 2022-07-11
T-Proxy for peering tests

thisisnotashwin:
Changes proposed in this PR:
- Support t-proxy in acceptance tests for peers

How I've tested this PR:
- Ran acceptance tests locally

- 👀 

Checklist:
- [x] Tests added

### DIFF
--- a/acceptance/tests/peering/peering_connect_namespaces_test.go
+++ b/acceptance/tests/peering/peering_connect_namespaces_test.go
@@ -2,7 +2,7 @@ package peering
 
 import (
 	"context"
-	"github.com/hashicorp/go-version"
+	"fmt"
 	"strconv"
 	"testing"
 
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
 	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-version"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -87,8 +88,8 @@ func TestPeering_ConnectNamespaces(t *testing.T) {
 
 				"global.acls.manageSystemACLs": strconv.FormatBool(c.ACLsAndAutoEncryptEnabled),
 
-				"connectInject.enabled":                         "true",
-				"connectInject.transparentProxy.defaultEnabled": "false",
+				"connectInject.enabled": "true",
+
 				// When mirroringK8S is set, this setting is ignored.
 				"connectInject.consulNamespaces.consulDestinationNamespace": c.destinationNamespace,
 				"connectInject.consulNamespaces.mirroringK8S":               strconv.FormatBool(c.mirrorK8S),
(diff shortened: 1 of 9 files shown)


acceptance/tests/peering/peering_connect_namespaces_test.go
```diff
@@ -79,16 +76,18 @@ func TestPeering_ConnectNamespaces(t *testing.T) {
 				"global.peering.enabled":        "true",
 				"global.enableConsulNamespaces": "true",
 
-				"global.image": "hashicorp/consul-enterprise:1.13.0-alpha2-ent",
+				// "global.image": "hashicorp/consul-enterprise:1.13.0-alpha2-ent",
+				"global.image":    "thisisnotashwin/consul@sha256:0733380a1a177d269c53fff62464e3a4840ea0c3ca24c6164282f8a822ec8825",
+				"global.imageK8S": "thisisnotashwin/consul-k8s@sha256:6fe1ec532876073813c824f27b2c972c03a41376e0729a502f6f3302dc352379",
```
ndhanushkodi, line 81: Just confirming these are the images we want, and if so we can remove L79!

acceptance/tests/peering/peering_connect_test.go
```diff
@@ -52,21 +49,25 @@ func TestPeering_Connect(t *testing.T) {
 			commonHelmValues := map[string]string{
 				"global.peering.enabled": "true",
 
-				"global.image": "hashicorp/consul:1.13.0-alpha2",
+				// "global.image": "hashicorp/consul:1.13.0-alpha2",
+				"global.image":    "thisisnotashwin/consul@sha256:0733380a1a177d269c53fff62464e3a4840ea0c3ca24c6164282f8a822ec8825",
+				"global.imageK8S": "thisisnotashwin/consul-k8s@sha256:6fe1ec532876073813c824f27b2c972c03a41376e0729a502f6f3302dc352379",
```
ndhanushkodi, line 54: We can remove L52

ndhanushkodi (approved):
Looks good Ashwin! Thanks for all the debugging!

Should we re-enable tproxy on the nightly tests with this PR or on a follow on PR since we had removed tproxy due to peering not yet supporting tproxy?

thisisnotashwin:
> Should we re-enable tproxy on the nightly tests with this PR or on a follow on PR since we had removed tproxy due to peering not yet supporting tproxy?

Great catch! I just enabled t-proxy on the nightly tests.

charts/consul/templates/server-statefulset.yaml
```diff
@@ -296,6 +296,7 @@ spec:
                 -config-dir=/consul/userconfig/{{ .name }} \
                 {{- end }}
                 {{- end }}
+                -hcl='ports { grpc = 8503 }' \
```
ishustava, line 299: could we put it in server configmap?

charts/consul/templates/server-service.yaml
```diff
@@ -39,6 +39,9 @@ spec:
       port: 8501
       targetPort: 8501
     {{- end }}
+    - name: grpc
+      port: 8503
+      targetPort: 8503
```
ishustava, line 44: since it's not a host port, could it be 8502?

thisisnotashwin: you mean the port is 8502 but the targetPort remains 8503?

charts/consul/templates/server-statefulset.yaml
thisisnotashwin: Good catch!

charts/consul/templates/server-service.yaml
thisisnotashwin: Because it will need to be a hostPort on kind. Then it will clash with the client GRPC port.

ishustava: you're totally right, I looked at the service instead of the stateful set!

ishustava (approved):

PR #1340, opened 2022-07-13
Set output with `-output` flag.

t-eckert:
After clarifying with @david-yu on the best way forward for the JSON flag, we decided to have an `output` flag which takes `table`, `json`, or `raw` to set the output.

This matches with David's original intent in the PRD.

Changes proposed in this PR:
- Add GetSetFlags method which allows access to groups of flags.
- Add `-output` (`-o`) flag which takes `table`, `json`, or `raw` for configuring output.
- Set config values to empty slices by default.

How I've tested this PR:

- Added a unit test for the new flag.

- Unit tests or manually.

Checklist:
- [x] Tests added

### DIFF
--- a/cli/cmd/proxy/read/command.go
+++ b/cli/cmd/proxy/read/command.go
@@ -240,21 +240,21 @@ func (c *ReadCommand) initKubernetes() (err error) {
 	return nil
 }
 
-func (c *ReadCommand) outputConfig(config *EnvoyConfig) {
+func (c *ReadCommand) outputConfig(config *EnvoyConfig) error {
 	if c.flagRawConfig {
 		c.UI.Output(string(config.rawCfg))
-		return
+		return nil
 	}
 
 	if c.flagJSON {
-		c.outputAsJSON(config)
-		return
+		return c.outputAsJSON(config)
 	}
 
 	c.outputAsTables(config)
+	return nil
 }
 
-func (c *ReadCommand) outputAsJSON(config *EnvoyConfig) {
+func (c *ReadCommand) outputAsJSON(config *EnvoyConfig) error {
 	cfg := make(map[string]interface{})
 	if !c.filtersPassed() || c.flagClusters {
 		cfg["clusters"] = config.Clusters
@@ -271,11 +271,14 @@ func (c *ReadCommand) outputAsJSON(config *EnvoyConfig) {
 	if !c.filtersPassed() || c.flagSecrets {
 		cfg["secrets"] = config.Secrets
 	}
+
 	out, err := json.MarshalIndent(cfg, "", "\t")
 	if err != nil {
-		panic(err)
+		return err
 	}
+
 	c.UI.Output(string(out))
+	return nil
 }
 
 func (c *ReadCommand) outputAsTables(config *EnvoyConfig) {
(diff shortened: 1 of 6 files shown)


cli/cmd/proxy/read/command.go
```diff
@@ -189,9 +201,19 @@ func (c *ReadCommand) validateFlags() error {
 	if errs := validation.ValidateNamespaceName(c.flagNamespace, false); c.flagNamespace != "" && len(errs) > 0 {
 		return fmt.Errorf("invalid namespace name passed for -namespace/-n: %v", strings.Join(errs, "; "))
 	}
+	if c.flagRawConfig && c.filtersPassed() {
+		return fmt.Errorf("-raw-config cannot be combined with filtering flags: %s.",
+			strings.Join(c.set.GetSetFlags("Output Filtering Options"), ", "))
+	}
```
t-eckert, line 207: Need to add a test for this.

cli/common/flag/set.go
```diff
@@ -56,6 +56,22 @@ func (f *Sets) NewSet(name string) *Set {
 	return flagSet
 }
 
+// GetSetFlags returns a slice of flags for a given set.
+// If the requested set does not exist, this will return an empty slice.
+func (f *Sets) GetSetFlags(setName string) []string {
```
t-eckert, line 61: There was previously no way to get "all the flags in the 'foo' group of flags" where 'foo' could be "Global Options", "Command Options", etc. I added a method for that.

cli/cmd/proxy/read/command.go
```diff
@@ -222,90 +244,144 @@ func (c *ReadCommand) initKubernetes() (err error) {
...
-		c.UI.Output(fmt.Sprintf("Clusters (%d)", len(config.Clusters)), terminal.WithHeaderStyle())
-		clusters := terminal.NewTable("Name", "FQDN", "Endpoints", "Type", "Last Updated")
-		for _, cluster := range config.Clusters {
-			clusters.AddRow([]string{cluster.Name, cluster.FullyQualifiedDomainName, strings.Join(cluster.Endpoints, ", "),
-				cluster.Type, cluster.LastUpdated}, []string{})
-		}
-		c.UI.Table(clusters)
-		c.UI.Output("")
+	c.UI.Output(fmt.Sprintf("Clusters (%d)", len(clusters)), terminal.WithHeaderStyle())
+	table := terminal.NewTable("Name", "FQDN", "Endpoints", "Type", "Last Updated")
```
jmurret, line 302: I don't think this is blocking, but since we can't test the output, we should consider separating the building table logic from the output logic, so that we can write assertions against the build table code.  Basically, it would be extracting lines 302-306 into another function and writing a test for that with some mocked structs.   
Same for similar functions in this file.
If we have plans to be able to assert against terminal output, that works too.

jmurret (approved):
Nice work.  It looks really great.  Thank you for putting so much thought into this.  I left a non-blocking comment about testing.

cli/cmd/proxy/read/command.go
t-eckert: We do! Actually in [this PR](https://github.com/hashicorp/consul-k8s/pull/1336) modified the command structs to write to a buffer instead of stdout. In that PR, I added tests that assert the correctness of the output tables.

jmurret: oh-gnarly

ndhanushkodi (approved):
Looks amazing Thomas!!

PR #1341, opened 2022-07-13
Peering non flat

ndhanushkodi:
Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/server-statefulset.yaml
+++ b/charts/consul/templates/server-statefulset.yaml
@@ -180,6 +180,27 @@ spec:
       {{- if .Values.server.priorityClassName }}
       priorityClassName: {{ .Values.server.priorityClassName | quote }}
       {{- end }}
+      initContainers:
+        - name: consul-init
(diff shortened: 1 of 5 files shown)


charts/consul/templates/server-statefulset.yaml
```diff
@@ -266,9 +284,12 @@ spec:
               {{- end }}
 
               {{ template "consul.extraconfig" }}
+              
+              WAN_ADDR="$(cat /consul/data/address.txt)"
 
               exec /usr/local/bin/docker-entrypoint.sh consul agent \
                 -advertise="${ADVERTISE_IP}" \
+                -hcl="external_addresses = [\"${WAN_ADDR}\"]" \
```
ndhanushkodi, line 292: This would be updated to include addresses from the helm value global.peering.service.serverExternalAddrs

PR #1345, opened 2022-07-15
CNI File watcher and pre applying rules setup

curtbushko:
Changes proposed in this PR:
- increase limits for CNI plugin so that it runs on GKE
- add annotations for transparent proxy status (enabled, waiting)
- Initial setup (CNI_ARGS) for getting information to the CNI plugin
- file watcher for config file changes and for when the config file does not exists
- added wait for annotation to be used before applying ipconfig traffic redirection

How I've tested this PR:

- unit tests
- manual testing

- You can run this in kind by disabling kinds default CNI, installing calico first and installing consul with cni turn on.

Or you can clone my [cni-demo repo](https://github.com/curtbushko/cni-demo/tree/main/consul-cni-on-kind) and run:

`cd consul-cni-on-kind; make create-cluster deploy-calico deploy-consul hashicups`

Note: the consul-cni will crashloop as I need to add the file watcher to it.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


Makefile
```diff
@@ -15,17 +15,18 @@ bats-tests: ## Run Helm chart bats tests.
 # ===========> Control Plane Targets
 
 control-plane-dev: ## Build consul-k8s-control-plane binary.
-	@$(SHELL) $(CURDIR)/control-plane/build-support/scripts/build-local.sh -o $(GOOS) -a $(GOARCH)
+	@$(SHELL) $(CURDIR)/control-plane/build-support/scripts/build-local.sh -o linux -a amd64
```
t-eckert, line 18: Is this temporary? I figure you made the change to make debugging easier?

curtbushko: It is temporary as there are changes in main that will need to come over here for the Makefile/Dockerfile.

curtbushko, line 18: Ignore this file for now. I will merge in Iryna's Makefile/Dockerfile changes when they get merged into main (and clean this up)

control-plane/cni/main.go
```diff
@@ -135,56 +162,119 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	if !ok {
+		return iptables.Config{}, fmt.Errorf("could not find %s annotation for %s pod", annotation, pod.Name)
+	}
+	cfg := iptables.Config{}
+	err := json.Unmarshal([]byte(anno), &cfg)
+	if err != nil {
+		return iptables.Config{}, fmt.Errorf("could not unmarshal %s annotation for %s pod", annotation, pod.Name)
+	}
+	return cfg, nil
+}
```
t-eckert, line 262: I feel this function is doing two separate things that should be divided up. The first is read the annotation from the Pod and the second is convert the value from the annotation to a `iptables.Config`. I think it might be cleaner and easier to test if you split these two actions into different functions.

control-plane/cni/main.go
```diff
@@ -135,56 +162,119 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+// searchDNSIPFromEnvironment gets the consul server DNS IP from the pods environment variables. The prefix makes searching easier
+// return an empty string.
+func searchDNSIPFromEnvironment(pod corev1.Pod, prefix string) string {
+	var result string
+	upcaseResourcePrefix := strings.ToUpper(prefix)
+	upcaseResourcePrefixWithUnderscores := strings.ReplaceAll(upcaseResourcePrefix, "-", "_")
+	dnsName := strings.Join([]string{upcaseResourcePrefixWithUnderscores, dnsServiceHostEnvSuffix}, "_")
+
+	// Environment variables are buried in the pod spec
+	vars := pod.Spec.Containers[0].Env
```
t-eckert, line 273: This is a Pod we control, right? I'm not 100% up to speed on CNI, but because this is a Pod that we have full control over, we are safe making the assumption that this container is always first in the spec?

control-plane/cni/main.go
```diff
@@ -135,56 +162,119 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	for k := range vars {
+		if vars[k].Name == dnsName {
+			result = vars[k].Value
```
t-eckert, line 276: You can add a break here once the case is matched.

t-eckert (commented):
First pass with some initial comments! More for me to read and do!

control-plane/cni/config/config.go
```diff
@@ -10,10 +10,13 @@ type CNIConfig struct {
...
-	Multus bool `json:"multus"      mapstructure:"multus"`
+	// DNSPrefix is used to determine the Consul Server DNS IP. The IP is set as an environment variable and the prefix allows us
+	// to search for it.
+	DNSPrefix string `json:"dns_prefix" mapstructure:"dns_prefix"`
 	// Kubeconfig file name. Can be set as a cli flag.
 	Kubeconfig string `json:"kubeconfig"  mapstructure:"kubeconfig"`
 	// LogLevl is the logging level. Can be set as a cli flag.
 	LogLevel string `json:"log_level"   mapstructure:"log_level"`
+	// Multus is if the plugin is a multus plugin. Can be set as a cli flag.
+	Multus bool `json:"multus"      mapstructure:"multus"`
```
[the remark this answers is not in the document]
curtbushko: I didn't really consider it because I worked from the other direction from the config file to the struct. I would like them all to match up.

control-plane/cni/main.go
```diff
@@ -22,36 +26,58 @@ import (
 )
 
 const (
-	keyCNIStatus    = "consul.hashicorp.com/cni-status"
-	keyInjectStatus = "consul.hashicorp.com/connect-inject-status"
-	injected        = "injected"
+	// These annotations are duplicated from control-plane/connect-inject/annotations.go in
+	// order to prevent pulling in dependencies.
```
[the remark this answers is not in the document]
curtbushko: You meant control-plane/config?  I am not sure if you have the same problem of pulling in all the dependencies or not.

control-plane/cni/main.go
```diff
@@ -22,36 +26,58 @@ import (
...
+	// injection is explicitly enabled or disabled for a pod.
+	annotationInject = "consul.hashicorp.com/connect-inject"
+	// annotationCNIProxyConfig stores iptables.Config information so that the CNI plugin can use it to apply iptables rules.
+	annotationCNIProxyConfig = "consul.hashicorp.com/cni-proxy-config"
+	// retries is the number of backoff retries to attempt while waiting for an cni-proxy-config annotation to poplulate.
+	retries = 10
+	// dnsServiceHostEnvSuffix is the suffix that is used to get the DNS host IP. The DNS IP is saved as an environment variable
+	// with prefix + suffix. The prefix is passed in as cli flag to the endpoints controller which is then passed on in the
+	// cni-proxy-config annotation so that the CNI plugin can use it. The DNS environment variable usually looks like:
+	// CONSUL_CONSUL_DNS_SERVICE_HOST but the prefix can change depending on the helm install.
```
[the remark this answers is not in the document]
curtbushko: Yah, it is all over the place. I tend to try and stick around 120. I have seen some go farther than that in control-plane.

control-plane/cni/main.go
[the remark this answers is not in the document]
t-eckert: By pulling in dependencies, do you mean importing the package?

I was thinking if there was a common package with all of the annotations, we could import it and use `annotation.NameOfAnnotation`. That way we prevent duplication that could lead to bugs.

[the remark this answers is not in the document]
curtbushko: Yes, by pulling in the package and adding a bunch of package to go.mod/sum. The annotations package would need to sit outside of control-plane to prevent that.  I've seen Istio do a similar thing to what you are proposing and it is pretty nice... `annotation.Foo` is nice to read IMHO.

I originally had the annotations down in control-plane/connect-inject/annotations.go but the cni plugin is an independent binary even though it is under control-plane/cni.  It pulled in A LOT of packages that it just doesn't need (like consul sdk/api). Hence, the duplication.

control-plane/cni/main.go
curtbushko: It is our pod yes but environment variables are for an entire pod. It wouldn't have mattered if I used Containers[1] or Containers[2].

control-plane/cni/main.go
curtbushko: Nice catch! I have updated it to break.

Makefile
ishustava: Sorry! I'll merge this today!

control-plane/cni/main.go
```diff
@@ -22,36 +26,58 @@ import (
...
+	// Type of plugin (consul-cni).
 	Type string `json:"type"`
-	// CNIBinDir is the location of the cni config files on the node. Can bet as a cli flag.
+	// CNIBinDir is the location of the cni config files on the node. Can be set as a cli flag.
 	CNIBinDir string `json:"cni_bin_dir"`
 	// CNINetDir is the locaion of the cni plugin on the node. Can be set as a cli flag.
 	CNINetDir string `json:"cni_net_dir"`
+	// DNSPrefix is used to determine the Consul Server DNS IP. The IP is set as an environment variable and the prefix allows us
+	// to search for it. The DNS IP is determined using the prefix and the dnsServiceHostEnvSuffix constant.
+	DNSPrefix string `json:"dns_prefix"`
```
ishustava, line 80: Why do we have two config structs with a lot of duplicated information? Do you think we could somehow consolidate them?

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
-			keyCNIStatus: "true",
-		}
-		pod.SetAnnotations(annotations)
-		_, err = client.CoreV1().Pods(podNamespace).Update(context.TODO(), pod, metav1.UpdateOptions{})
-		if err != nil {
-			return err
-		}
-	} else {
-		logger.Debug("skipping traffic redirect on un-injected pod")
+	// skip injecting pod that has consul.hashicorp.com/connect-inject: false
```
ishustava, line 173:
```suggestion
	// Skip pod that has consul.hashicorp.com/connect-inject: false
```

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	if skipConnectInject(*pod) {
+		logger.Debug("skipping traffic redirect on un-injected pod: %s", pod.Name)
+		return types.PrintResult(result, cfg.CNIVersion)
+	}
+
+	// check to see if the cni-proxy-config annotation exists and if not, wait, retry and backoff.
```
ishustava, line 179: nit: all comments should start capitalized (so that it's a sentence)

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	exists := waitForAnnotation(*pod, annotationCNIProxyConfig, uint64(retries))
+	if !exists {
+		logger.Error("could not retrieve annotation: %s on pod: %s", annotationCNIProxyConfig, pod.Name)
+		return types.PrintResult(result, cfg.CNIVersion)
```
ishustava, line 183: Should we return an error in this case?

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	}
+
+	// parse the cni-proxy-config annotation into an iptables.Config object.
+	iptablesCfg, err := parseAnnotation(*pod, annotationCNIProxyConfig)
+	if err != nil {
+		logger.Error("could not parse annotation: %s, error: %v", annotationCNIProxyConfig, err)
+		return types.PrintResult(result, cfg.CNIVersion)
```
ishustava, line 190: same question as above: why don't we return the error here?

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	}
+
+	// get the DNS IP from the environment.
+	dnsIP := searchDNSIPFromEnvironment(*pod, cfg.DNSPrefix)
+	if dnsIP != "" {
+		logger.Error("assigned consul dns ip to %s", dnsIP)
```
ishustava, line 196:
```suggestion
		logger.Info("assigned consul DNS IP to %s", dnsIP)
```

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 
-func hasBeenInjected(pod corev1.Pod) bool {
-	if anno, ok := pod.Annotations[keyInjectStatus]; ok && anno == injected {
+// skipConnectInject determines if the connect-inject annotation is false.
+func skipConnectInject(pod corev1.Pod) bool {
+	if anno, ok := pod.Annotations[annotationInject]; ok && anno == "false" {
 		return true
 	}
 	return false
 }
```
ishustava, line 234: What about the behavior with the `connectInject.default` Helm value? If you have it set to false, then you wouldn't need to set connect inject annotation to false (because it's redundant). This function will return false, but it should return true. 

We should probably instead look at connect inject status like we do in the `shouldInject` function on the webhook which also checks for other rules.

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	err := json.Unmarshal([]byte(anno), &cfg)
+	if err != nil {
+		return iptables.Config{}, fmt.Errorf("could not unmarshal %s annotation for %s pod", annotation, pod.Name)
+	}
+	return cfg, nil
+}
+
+// searchDNSIPFromEnvironment gets the consul server DNS IP from the pods environment variables. The prefix makes searching easier
+// return an empty string.
+func searchDNSIPFromEnvironment(pod corev1.Pod, prefix string) string {
```
ishustava, line 266: I think we need a different way to do this. The env var will only be available if the DNS service and the application are running in the same namespace (https://kubernetes.io/docs/concepts/containers/container-environment/)

ishustava (commented):
Leaving my comments so far. I didn't get to review the tests yet! Will continue reviewing later today.

control-plane/cni/main.go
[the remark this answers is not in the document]
t-eckert: I understand now! I didn't realize it was an independent binary.

control-plane/cni/main.go
t-eckert: That makes sense. I hadn't considered that of course the environment variables are the same across all of the containers in a Pod, but it's obvious when you say it.

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -736,17 +736,17 @@ type gatewayRulesGenerator func(name, namespace string) (string, error)
...
-	//ConsulDC is the name of the DC where the gateways will be registered
+	// ConsulDC is the name of the DC where the gateways will be registered
 	ConsulDC string
-	//PrimaryDC is the name of the Primary Data Center
+	// PrimaryDC is the name of the Primary Data Center
 	PrimaryDC string
-	//Primary specifies whether the ConsulDC is the Primary Data Center
+	// Primary specifies whether the ConsulDC is the Primary Data Center
 	Primary bool
 }
```
t-eckert, line 751: Some of these comments might get flagged by the linter for not ending in a period.

control-plane/cni/main_test.go
```diff
@@ -0,0 +1,146 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestWaitForAnnotation(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name       string
+		annotation string
+		pod        func(*corev1.Pod) *corev1.Pod
```
t-eckert, line 19: This might be clearer as `annotatePod`.

```suggestion
		annotatePod        func(*corev1.Pod) *corev1.Pod
```

control-plane/cni/main_test.go
```diff
@@ -0,0 +1,146 @@
...
+		})
+	}
+}
+
+func TestParseAnnotation(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name       string
+		annotation string
+		pod        func(*corev1.Pod) *corev1.Pod
```
t-eckert, line 56: This might be clearer as `configurePod`.

```suggestion
		configurePod        func(*corev1.Pod) *corev1.Pod
```

control-plane/cni/main_test.go
```diff
@@ -0,0 +1,146 @@
...
+			pod: func(pod *corev1.Pod) *corev1.Pod {
+				pod.Spec.Containers[0].Env = []corev1.EnvVar{
+					{
+						Name:  "CONSUL_DNS_SERVICE_HOST",
+						Value: "127.0.0.1",
+					},
+				}
+				return pod
+			},
+		},
```
t-eckert, line 118: I would add a case where the env vars are not set.
Also, similarly to my comments above, I think `pod` might be clearer as `configurePod`.

control-plane/cni/main_test.go
```diff
@@ -0,0 +1,146 @@
...
+				{
+					Name: "web",
+				},
+				{
+					Name: "web-side",
+				},
+			},
+		},
+	}
+}
```
t-eckert, line 146: I like this and I plan on stealing it.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -158,7 +159,13 @@ func (r *EndpointsController) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	if isLabeledIgnore(serviceEndpoints.Labels) {
 		// We always deregister the service to handle the case where a user has registered the service, then added the label later.
-		r.Log.Info("Ignoring endpoint labeled with `consul.hashicorp.com/service-ignore: \"true\"`", "name", req.Name, "namespace", req.Namespace)
+		r.Log.Info(
+			"Ignoring endpoint labeled with `consul.hashicorp.com/service-ignore: \"true\"`",
+			"name",
+			req.Name,
+			"namespace",
+			req.Namespace,
+		)
```
t-eckert, line 168: I'm glad you opened these up a bit. Much easier to read.z

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -302,6 +357,17 @@ func getServiceCheck(client *api.Client, healthCheckID string) (*api.AgentCheck,
...
 
+// getServiceNameChecks will return all of the health checks for a service.
+func getServiceNameChecks(client *api.Client, serviceName string) (map[string]*api.AgentCheck, error) {
+	filter := fmt.Sprintf("ServiceName == %q", serviceName)
+	checks, err := client.Agent().ChecksWithFilter(filter)
+	if err != nil {
+		return nil, err
+	}
+	// This will be nil (does not exist) or an actual check.
+	return checks, nil
```
t-eckert, line 368: Only if you are feeling spicy.

```suggestion
	return client.Agent().ChecksWithFilter(filter)
```

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -380,7 +458,8 @@ func (r *EndpointsController) upsertHealthCheck(pod corev1.Pod, client *api.Clie
 func getServiceName(pod corev1.Pod, serviceEndpoints corev1.Endpoints) string {
 	serviceName := serviceEndpoints.Name
 	// If the annotation has a comma, it is a multi port Pod. In that case we always use the name of the endpoint.
-	if serviceNameFromAnnotation, ok := pod.Annotations[annotationService]; ok && serviceNameFromAnnotation != "" && !strings.Contains(serviceNameFromAnnotation, ",") {
+	if serviceNameFromAnnotation, ok := pod.Annotations[annotationService]; ok && serviceNameFromAnnotation != "" &&
+		!strings.Contains(serviceNameFromAnnotation, ",") {
```
t-eckert, line 462: I think it's more clear which part is an if statement if you format it this way:

```suggestion
	if serviceNameFromAnnotation, ok := pod.Annotations[annotationService]; 
		ok && serviceNameFromAnnotation != "" && !strings.Contains(serviceNameFromAnnotation, ",") {
```

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -635,11 +731,49 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
...
 		}
-	}
 
+	}
 	return service, proxyService, nil
 }
 
+// redirectTrafficAnnotation creates an annotation that contains iptables configuration.
+func (r *EndpointsController) redirectTrafficAnnotation(pod corev1.Pod, cfg iptables.Config) error {
+	ctx := context.Background()
```
t-eckert, line 741: `EnpointsController` has a context you can use.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -635,11 +731,49 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
...
 			}
 		}
-	}
 
+	}
 	return service, proxyService, nil
 }
 
+// redirectTrafficAnnotation creates an annotation that contains iptables configuration.
+func (r *EndpointsController) redirectTrafficAnnotation(pod corev1.Pod, cfg iptables.Config) error {
```
t-eckert, line 740: I think `addTrafficRedirectAnnotation` is clearer.

control-plane/cni/main.go
curtbushko: @ishustava I was following how connect-inject does it in [container_init.go](https://github.com/hashicorp/consul-k8s/blob/main/control-plane/connect-inject/container_init.go#L148url)

I am assuming there that it is running in the context of a container so the environment variable is set already...?

control-plane/cni/main_test.go
curtbushko: I stole it too! I really like using the function in the test case to setup pods.

control-plane/cni/main.go
ishustava: The code you linked to is executed in the webhook, meaning the webhook will do the searching of the dns service env variable, not at the time when the container-init container runs. I think we have two options:

1. Figure out the dns ip in the cni installer and pass it down to cni. The problem with this is that it won't work if IP changes (ie. kube service gets recreated)
2. Do dns lookup of the consul dns service in this function and pass the dns service name from the installer.

control-plane/cni/main.go
```diff
@@ -95,7 +121,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	podNamespace := string(cniArgs.K8S_POD_NAMESPACE)
 	podName := string(cniArgs.K8S_POD_NAME)
 
-	// We only run in a pod
+	// We should never encounter this unless there has been an error in the kubelet. A good safeguard.
 	if podNamespace == "" && podName == "" {
```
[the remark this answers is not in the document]
curtbushko: Good call! The error should never happen but if any of those are empty then bad things are happening. Changed to ||.

control-plane/cni/main.go
curtbushko: Thanks! I will try find all of these.

control-plane/cni/main.go
curtbushko: You are correct on all of these, we should throw errors for them. Updated to return correctly.

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	// apply the iptables rules.
+	err = iptables.Setup(iptablesCfg)
+	if err != nil {
+		logger.Error("could not apply iptables setup: %v", err)
+		return types.PrintResult(result, cfg.CNIVersion)
+	}
 
-	// Pass through the result for the next plugin
+	logger.Debug("traffic redirect rules applied to pod: %s", pod.Name)
+	// Pass through the result for the next plugin even though we are the final plugin in the chain.
```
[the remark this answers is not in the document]
curtbushko: Yes, because I make it so in the file watcher. It covers the following scenarios:

- If a config file does not exist yet (ie our installer daemonset if the first started), then we will wait for the config file to show up.
- If a config file exists, append our config.
- If the config file changes, validate it to make sure our config is the last in the chain and fix if need be.

control-plane/subcommand/server-acl-init/command.go
curtbushko: Thanks!  I was failing at linting.

control-plane/cni/main.go
```diff
@@ -135,56 +162,120 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+func waitForAnnotation(pod corev1.Pod, annotation string, retries uint64) bool {
+	var err error
+	err = backoff.Retry(func() error {
+		var ok bool
+		_, ok = pod.Annotations[annotation]
+		if !ok {
+			return fmt.Errorf("annotation %s does not exist yet", annotation)
+		}
+		return err
+	}, backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), retries))
```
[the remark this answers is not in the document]
curtbushko: You were correct. The backoff is constant and would last 10 seconds total. I have set this to 30 retries (ie 30 seconds). We could do an ExponentialBackoff (it is a cool package) to slow down the retries but I think being responsive is better.

control-plane/cni/main.go
curtbushko: Still working on this piece.

control-plane/cni/main.go
curtbushko: Will look at this at a later date, maybe when multus is done.

control-plane/cni/main.go
curtbushko: This has been updated in several places. The helm chart passes the `consul-fullname`.`Release.Namespace` to the installers command line as a `-dns-service` argument. That ends up in the config, the plugin grabs it and gets the IP address from Kubernetes on every run.

control-plane/cni/main_test.go
curtbushko: Done.

control-plane/connect-inject/endpoints_controller.go
curtbushko: Nice find! Done!

control-plane/connect-inject/endpoints_controller.go
curtbushko: Done.

t-eckert (approved):
👍🏻

control-plane/subcommand/install-cni/cniconfig.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// Get the 'plugins' map embedded inside of the exisingMap.
+	plugins, err := pluginsFromMap(cfgMap)
+	if err != nil {
+		return err
+	}
+
+	// Find the 'consul-cni' plugin and remove it.
+	for i, p := range plugins {
+		plugin, ok := p.(map[string]interface{})
```
ishustava, line 193: could you add a comment here as to why we're not unmarshalling into `map[string]map[string]interface{}`

control-plane/subcommand/install-cni/command_test.go
```diff
@@ -1,81 +1,129 @@
...
-		Kubeconfig: defaultKubeconfig,
-		LogLevel:   defaultLogLevel,
+	// Event 3: consul config was removed from the config file. Should detect and fix.
+	err = replaceFile(baseConfigFile, filepath.Join(tempDir, configFile))
+	require.NoError(t, err)
+	time.Sleep(50 * time.Millisecond)
+	// Filewatcher should have detected change, fixed and appended to the config file. Make sure
+	// files match.
+	require.Equal(t, string(expected), string(actual))
+	cancel()
```
ishustava, line 94: I think this would need to be in t.Cleanup in the beginning of the test somewhere
```suggestion
       t.Cleanup(cancel)
```

control-plane/subcommand/install-cni/command_test.go
```diff
@@ -1,81 +1,129 @@
...
-		Multus:     defaultMultus,
-		Kubeconfig: defaultKubeconfig,
-		LogLevel:   defaultLogLevel,
+	// Event 3: consul config was removed from the config file. Should detect and fix.
+	err = replaceFile(baseConfigFile, filepath.Join(tempDir, configFile))
+	require.NoError(t, err)
+	time.Sleep(50 * time.Millisecond)
+	// Filewatcher should have detected change, fixed and appended to the config file. Make sure
+	// files match.
+	require.Equal(t, string(expected), string(actual))
```
ishustava, line 93:
```suggestion
	// Filewatcher should have detected change, fixed and appended to the config file. Make sure
	// files match.
	retry.Run(t, func(r *retry.R) {
		 require.Equal(r, string(expected), string(actual))
	})
```

ishustava (commented):
Amazing work Curt!! You did such a great job on testing! I had a few suggestions in-line

control-plane/cni/main.go
curtbushko: This is addressed in a further PR when we do traffic redirect and the webhook work

control-plane/cni/main.go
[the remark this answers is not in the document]
curtbushko: Removed retries.

control-plane/cni/main.go
curtbushko: This has been removed and will be addressed by getting the environment variable in the webhook.

control-plane/cni/main.go
```diff
@@ -22,45 +25,68 @@ import (
...
 	keyInjectStatus = "consul.hashicorp.com/connect-inject-status"
-	injected        = "injected"
+
+	// keyTransparentProxyStatus is the key of the annotation that is added to
+	// a pod when transparent proxy is done.
+	keyTransparentProxyStatus = "consul.hashicorp.com/transparent-proxy-status"
+
+	// annotationCNIProxyConfig stores iptables.Config information so that the CNI plugin can use it to apply
+	// iptables rules.
+	annotationCNIProxyConfig = "consul.hashicorp.com/cni-proxy-config"
```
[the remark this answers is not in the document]
curtbushko: Renamed it to `annotationTrafficRedirection = "consul.hashicorp.com/traffic-redirection-config`.

control-plane/cni/main.go
```diff
@@ -135,56 +162,93 @@ func cmdAdd(args *skel.CmdArgs) error {
...
-		}
-	} else {
-		logger.Debug("skipping traffic redirect on un-injected pod")
+	// Skip traffic redirection the correct annotations are not on the pod.
+	if skipTrafficRedirection(*pod, retries) {
+		logger.Debug("skipping traffic redirect on un-injected pod: %s", pod.Name)
+		return types.PrintResult(result, cfg.CNIVersion)
+	}
+
+	pod.Annotations[keyTransparentProxyStatus] = waiting
```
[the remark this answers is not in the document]
curtbushko: This is a simple way to show some sort of status that the CNI plugin is working. I have added a simple `updateTransparentProxyStatusAnnotation` function to clean it up a bit so that we can have a `waiting` and `complete` annotation.

control-plane/cni/main.go
```diff
@@ -135,56 +162,93 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+		return fmt.Errorf("error adding waiting annotation to pod: %s", err)
 	}
 
-	// TODO: Get transparent proxy annotations and merge with proxy config
-	// TODO: Redirect traffic :)
+	// Check to see if the cni-proxy-config annotation exists and if not, wait, retry and backoff.
+	exists := waitForAnnotation(*pod, annotationCNIProxyConfig, uint64(retries))
+	if !exists {
+		return fmt.Errorf("could not retrieve annotation")
+	}
```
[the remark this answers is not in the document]
curtbushko: Removed.

control-plane/cni/main.go
```diff
@@ -135,56 +162,93 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 	skel.PluginMain(cmdAdd, cmdCheck, cmdDel, version.All, bv.BuildString("consul-cni"))
 }
 
-func hasBeenInjected(pod corev1.Pod) bool {
-	if anno, ok := pod.Annotations[keyInjectStatus]; ok && anno == injected {
-		return true
+// skipTrafficRedirection looks for annotations on the pod and determines if it should skip traffic redirection.
+// The absence of the annotations is the equivalent of "disabled" because it means that the connect inject mutating
+// webhook did not run against the pod.
+func skipTrafficRedirection(pod corev1.Pod, retries uint64) bool {
```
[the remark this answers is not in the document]
curtbushko: Removed all usages of retries.

control-plane/cni/main_test.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			actual := skipTrafficRedirection(*c.annotatedPod(minimalPod()), c.retries)
+			require.Equal(t, c.expectedSkip, actual)
+		})
+	}
+}
+
+func TestWaitForAnnotation(t *testing.T) {
```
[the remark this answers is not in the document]
curtbushko: Removed.

control-plane/subcommand/install-cni/binary.go
```diff
@@ -0,0 +1,56 @@
...
+	filename := filepath.Base(srcFile)
+	// If the destDir does not exist then the incorrect command line argument was used or
+	// the CNI settings for the kubelet are not correct.
+	info, err := os.Stat(destDir)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("destination directory %s does not exist: %v", destDir, err)
+	}
+
+	if !info.IsDir() {
+		return fmt.Errorf("destination directory %s is not a directory: %v", destDir, err)
```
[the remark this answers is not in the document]
curtbushko: I think I have changed all of these in all the /install-cni and /cni directories.

control-plane/subcommand/install-cni/binary.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	// Check if the user bit is enabled in file permission.
+	if info.Mode().Perm()&(1<<(uint(7))) == 0 {
+		return fmt.Errorf("cannot write to destination directory %s: %v", destDir, err)
+	}
```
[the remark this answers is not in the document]
curtbushko: I think `os.WriteFile` would return a generic `permission denied` error. I kind of like doing the check so that I fail fast before opening the source file.

control-plane/subcommand/install-cni/binary.go
```diff
@@ -0,0 +1,56 @@
...
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		// Nothing to delete.
+		return nil
+	}
+
+	if err := os.Remove(path); err != nil {
+		return fmt.Errorf("error removing file %s: %v", path, err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
curtbushko: This one was fun! I added several tests for copyFile and removeFile.

control-plane/subcommand/install-cni/cniconfig.go
```diff
@@ -0,0 +1,272 @@
...
+// Adapted from kubelet: https://github.com/kubernetes/kubernetes/blob/954996e231074dc7429f7be1256a579bedd8344c/pkg/kubelet/dockershim/network/cni/cni.go#L134.
+func defaultCNIConfigFile(dir string) (string, error) {
+	files, err := libcni.ConfFiles(dir, []string{".conf", ".conflist", ".json"})
+	switch {
+	case err != nil:
+		// A real error has been found.
+		return "", fmt.Errorf("error while trying to find files in %s: %v", dir, err)
+	case len(files) == 0:
+		// No config files have shown up yet and it is ok to run this function again.
+		return "", nil
```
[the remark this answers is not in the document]
curtbushko: It has been simplified.

control-plane/subcommand/install-cni/cniconfig.go
```diff
@@ -0,0 +1,272 @@
...
+		return err
+	}
+
+	// Find the 'consul-cni' plugin and remove it.
+	for i, p := range plugins {
+		plugin, ok := p.(map[string]interface{})
+		if !ok {
+			return fmt.Errorf("error reading plugin from plugin list")
+		}
+		if plugin["type"] == "consul-cni" {
```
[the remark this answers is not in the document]
curtbushko: Good call, I had a constant that was being used called `consulCNIBinName`. I changed it to `consulCNIName` and use it in both locations now since it is the same.

control-plane/subcommand/install-cni/cniconfig.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+	}
+
+	if !found {
+		return fmt.Errorf("consul-cni config missing from config file")
+	}
+
+	// Compare the config that is passed to the installer to what is in the config file. There could be a
+	// difference if the config was corrupted or during a helm update or upgrade.
+	equal := reflect.DeepEqual(existingCfg, cfg)
```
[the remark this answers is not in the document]
curtbushko: Done!

control-plane/subcommand/install-cni/cniconfig.go
curtbushko: Done.

```
// We do not unmarshall this into a map[string]map[string]interface{} because there is no structure to
// the plugins. They are unstructured json and can be in any format that the plugin provider wishes.
```

control-plane/subcommand/install-cni/command.go
```diff
@@ -140,231 +113,148 @@ func (c *Command) Run(args []string) int {
...
 
-	// Create the install Config for working with files
-	install := &installConfig{
-		MountedCNIBinDir: "/host" + c.flagCNIBinDir,
-		MountedCNINetDir: "/host" + c.flagCNINetDir,
-		CNIBinSourceDir:  c.flagCNIBinSourceDir,
-	}
+	ctx, cancel := context.WithCancel(context.Background())
+	go func(sigChan chan os.Signal, cancel context.CancelFunc) {
+		<-sigChan
```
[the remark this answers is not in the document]
curtbushko: I got rid of the go routine and pulled this into directory watcher. I manually tests the SIGTERM and it still works and cleans up the files I put down when I delete a pod.

control-plane/subcommand/install-cni/command_test.go
curtbushko: Done.

control-plane/subcommand/install-cni/command_test.go
curtbushko: I added this to a few places that I thought were relevant.

control-plane/cni/main_test.go
```diff
@@ -0,0 +1,168 @@
...
+			exists:  false,
+		},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			actual := waitForAnnotation(*c.annotatedPod(minimalPod()), c.annotation, c.retries)
+			require.Equal(t, c.exists, actual)
+		})
+	}
+}
```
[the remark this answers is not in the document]
curtbushko: I added a bunch of tests to cmdAdd but more will come when traffic redirect shows up.

control-plane/cni/main.go
```diff
@@ -106,85 +134,122 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	pod, err := c.client.CoreV1().Pods(podNamespace).Get(ctx, podName, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("Error initializing Kubernetes client: %s", err)
+		return fmt.Errorf("error retrieving pod: %s", err)
 	}
 
-	pod, err := client.CoreV1().Pods(podNamespace).Get(ctx, podName, metav1.GetOptions{})
-	if err != nil {
-		return err
+	// Skip traffic redirection the correct annotations are not on the pod.
```
ishustava, line 178:
```suggestion
	// Skip traffic redirection if the correct annotations are not on the pod.
```

ishustava (approved):
Looks good!! Just one minor suggestion. 

Approving, assuming unit and acceptance tests are fixed before merge. For acceptance, you might need to rebase to get latest peering fixes.

PR #1350, opened 2022-07-19
Enable ubi images in the build workflow

ishustava:
Changes proposed in this PR:
- Enable ubi images in the build workflow
- Make `PKG_NAME` consistent with project name in `ci.hcl` and the artifactory directory
- Fix Dockerfile args

How I've tested this PR:
I didn't test it as we'll need to run them in the pipeline to do so

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ on:
       - "release/**"
 
 env:
-  PKG_NAME: "consul-k8s-control-plane"
+  PKG_NAME: "consul-k8s"
 
 jobs:
   get-go-version:
@@ -189,7 +189,7 @@ jobs:
           name: ${{ env.DEB_PACKAGE }}
           path: out/${{ env.DEB_PACKAGE }}
 
-  build-docker-default:
+  build-docker:
     name: Docker ${{ matrix.arch }} default release build
     needs: [get-product-version, build]
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -221,44 +221,40 @@ jobs:
             docker.io/hashicorp/${{ env.repo }}-control-plane:${{ env.version }}
             986891699432.dkr.ecr.us-east-1.amazonaws.com/hashicorp/${{ env.repo }}-control-plane:${{ env.version }}
           dev_tags: |
-            hashicorppreview/${{ env.repo }}-control-plane:${{ env.version }}-dev
+            hashicorppreview/${{ env.repo }}-control-plane:${{ env.version }}
```
ishustava, line 224: The dev version already has the `-dev` suffix and so our image before ended up having `-dev-dev` suffix

t-eckert (approved):
Looks good! Does the package name in artifactory end up being used by operators? I just worry about having two things called `consul-k8s`, but I understand that the control plane originally was labeled as `consul-k8s` and it might be a PITA to change it.

curtbushko (approved):
👍

ishustava:
>Does the package name in artifactory end up being used by operators?

I'm not sure what you mean by operators. It's for internal use only (staging and dev artifacts). Production artifacts end up in the same places as before (releases.hashicorp and dockerhub)

t-eckert:
> > Does the package name in artifactory end up being used by operators?
> 
> I'm not sure what you mean by operators. It's for internal use only (staging and dev artifacts). Production artifacts end up in the same places as before (releases.hashicorp and dockerhub)

That answers the question. I was just wondering if an end-user would see a download for `consul-k8s` that would be the control plane and not the CLI.

PR #1354, opened 2022-07-19
Envoy debugging multiport support

t-eckert:
📹 Demo

https://user-images.githubusercontent.com/29112081/180005754-694d5aa4-7c81-44fb-b07a-a7d63d2b4995.mov

Changes proposed in this PR:

- Get admin port for all services
- Add UI to all commands
- Output config for each service
- Separate out formatting of tables
- Add fake pod to fix command_test
- Add tests for formatting

How I've tested this PR:

- New unit tests, and I ran this on a multiport pod.

- Unit tests or multiport pod manual tests.

Checklist:
- [x] Tests added

### DIFF
--- a/cli/cmd/proxy/read/command.go
+++ b/cli/cmd/proxy/read/command.go
@@ -156,7 +156,7 @@ func (c *ReadCommand) Run(args []string) int {
 		return 1
 	}
 
-	c.outputConfigs(configs)
+	err = c.outputConfigs(configs)
 	if err != nil {
 		c.UI.Output(err.Error(), terminal.WithErrorStyle())
 		return 1
--- a/cli/cmd/proxy/read/command.go
+++ b/cli/cmd/proxy/read/command.go
@@ -12,13 +12,14 @@ import (
 	"github.com/hashicorp/consul-k8s/cli/common/terminal"
 	helmCLI "helm.sh/helm/v3/pkg/cli"
 	"k8s.io/apimachinery/pkg/api/validation"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/utils/strings/slices"
 )
 
-// adminPort is the port where the Envoy admin API is exposed.
-const adminPort int = 19000
+// defaultAdminPort is the port where the Envoy admin API is exposed.
+const defaultAdminPort int = 19000
 
 const (
 	Table = "table"
@@ -243,16 +244,26 @@ func (c *ReadCommand) initKubernetes() (err error) {
 }
 
 func (c *ReadCommand) fetchAdminPorts() (map[string]int, error) {
-	// Grab the annotations on the Pod.
-	// Check if it's multiport
-	// If it is, return the multiple ports
-	// If it isn't, return default.
+	adminPorts := make(map[string]int, 0)
 
-	defaultPort := map[string]int{
-		c.flagPodName: 19000,
+	pod, err := c.kubernetes.CoreV1().Pods(c.flagNamespace).Get(c.Ctx, c.flagPodName, metav1.GetOptions{})
+	if err != nil {
+		return adminPorts, err
 	}
 
-	return defaultPort, nil
+	connectService, isMultiport := pod.Annotations["consul.hashicorp.com/connect-service"]
+
+	if !isMultiport {
+		// Return the default port configuration.
+		adminPorts[c.flagPodName] = defaultAdminPort
+		return adminPorts, nil
+	}
+
+	for index, service := range strings.Split(connectService, ",") {
+		adminPorts[service] = defaultAdminPort + index
+	}
+
+	return adminPorts, nil
 }
 
 func (c *ReadCommand) outputConfigs(configs map[string]*EnvoyConfig) error {
--- a/cli/cmd/proxy/read/command.go
+++ b/cli/cmd/proxy/read/command.go
@@ -207,7 +207,7 @@ func (c *ReadCommand) validateFlags() error {
 	return nil
 }
 
-// filtersPassed returns true if the user has passed a flag which filters th
+// filtersPassed returns true if the user has passed a flag which filters the
 // output.
 func (c *ReadCommand) filtersPassed() bool {
 	return c.flagClusters || c.flagEndpoints || c.flagListeners || c.flagRoutes || c.flagSecrets
--- a/cli/cmd/proxy/read/command.go
+++ b/cli/cmd/proxy/read/command.go
@@ -143,25 +143,24 @@ func (c *ReadCommand) Run(args []string) int {
 		return 1
 	}
 
-	pf := common.PortForward{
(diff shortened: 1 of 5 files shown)


cli/cmd/proxy/read/command.go
```diff
@@ -143,25 +144,24 @@ func (c *ReadCommand) Run(args []string) int {
 		return 1
 	}
 
-	pf := common.PortForward{
-		Namespace:  c.flagNamespace,
-		PodName:    c.flagPodName,
-		RemotePort: adminPort,
-		KubeClient: c.kubernetes,
-		RestConfig: c.restConfig,
+	adminPorts, err := c.fetchAdminPorts()
```
t-eckert, line 147: Instead of pinning to the default admin port, we need to fetch the number of services and all of the ports they serve the admin interface on.

cli/cmd/proxy/read/format_test.go
```diff
@@ -0,0 +1,220 @@
...
+			Status:  "HEALTHY",
+		},
+	}
+
+	expectedHeaders := []string{"Address:Port", "Cluster", "Weight", "Status"}
+
+	actual := formatEndpoints(given)
+
+	require.Equal(t, expectedHeaders, actual.Headers)
+	require.Equal(t, len(given), len(actual.Rows))
```
thisisnotashwin, line 119: Given we are actually testing the formatting here, I think a strong assertion would be where we could actually compare the contents of actual vs what we think the output should be as a string. That would make the tests more robust, and also, we could actually test that it is formatted as expected.

thisisnotashwin (commented):
Great work Thomas. Left some comments inline.

cli/cmd/proxy/read/format_test.go
t-eckert: I like that. I'll need to run the render method and have it output into a buffer. I'll give that a quick shot. I think it's valuable and not too dissimilar from what I did for commands.

cli/cmd/proxy/read/command.go
```diff
@@ -243,53 +243,132 @@ func (c *ReadCommand) initKubernetes() (err error) {
...
+	for name, adminPort := range adminPorts {
+		pf := common.PortForward{
+			Namespace:  c.flagNamespace,
+			PodName:    c.flagPodName,
+			RemotePort: adminPort,
+			KubeClient: c.kubernetes,
+			RestConfig: c.restConfig,
+		}
+
+		config, err := c.fetchConfig(c.Ctx, &pf)
```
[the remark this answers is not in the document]
thisisnotashwin: 🤔 yeah that makes sense given how it is configured in this context. It definitely is a smell though. Maybe we can look into this after the release to see if there is a cleaner way of expressing this behavior.

cli/cmd/proxy/read/format_test.go
t-eckert: Added!

cli/cmd/proxy/read/command.go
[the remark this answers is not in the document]
t-eckert: Definitely. I am super open to looking into other ways of doing this.

cli/cmd/proxy/read/command.go
```diff
@@ -363,12 +404,7 @@ func (c *ReadCommand) outputRoutesTable(routes []Route) {
 	}
 
 	c.UI.Output(fmt.Sprintf("Routes (%d)", len(routes)), terminal.WithHeaderStyle())
-	table := terminal.NewTable("Name", "Destination Cluster", "Last Updated")
-	for _, route := range routes {
-		table.AddRow([]string{route.Name, route.DestinationCluster, route.LastUpdated}, []string{})
-	}
-	c.UI.Table(table)
-	c.UI.Output("")
+	c.UI.Table(formatRoutes(routes))
```
thisisnotashwin, line 407: It feels like the entire `outputRoutesTable` and `formatRoutes` can be one large method. There isn't behavior that is in here except the printing of headers which I think can be moved into the other file and maintained under a single method. This extends to the secrets, listeners endpoints and clusters as well. I think we should aim to have lesser indirection.

cli/cmd/proxy/read/format_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+import (
+	"bytes"
+	"context"
+	"testing"
+
+	"github.com/hashicorp/consul-k8s/cli/common/terminal"
+	"github.com/stretchr/testify/require"
+)
+
```
thisisnotashwin, line 11: These tests are excellent!

thisisnotashwin (commented):
These look good! Have a comment about consolidating the output*Table and format* methods to simplify reading the code.

cli/cmd/proxy/read/command_test.go
```diff
@@ -3,6 +3,7 @@ package read
 import (
 	"bytes"
```
ndhanushkodi, line 4: Should we have tests for some of the helper functions in command.go? such as fetchAdminPorts, fetchConfigs?

cli/cmd/proxy/read/command_test.go
```diff
@@ -53,9 +56,11 @@ func TestFlagParsing(t *testing.T) {
 }
 
 func TestReadCommandOutput(t *testing.T) {
```
ndhanushkodi, line 58: Should we add happy path cases for Raw and JSON output cases? And maybe we could consider adding a happy path multiport case, maybe just with one of the types of output? Then this file could kind of cover some end to end happy path cases, and format_test.go is covering the table formatting cases.

ndhanushkodi (commented):
Amazing work Thomas, the demo looks awesome and love the format tests! Had a few questions about some additional test cases we could add, let me know what you think about adding those!

cli/cmd/proxy/read/format_test.go
t-eckert: Thank you!

cli/cmd/proxy/read/command.go
t-eckert: I like to avoid indirection too. In this case, however, I like the testability of the format functions for now, especially because they avoid having to set up the command for every test. I also want to see how this looks in the context of the filtering functionality which is in another open PR.

Is this a blocker for you, or would you be ok with me keeping this as is and re-evaluating once the added context of filtering is involved? I definitely agree with your point of view here on indirection, I'm just weighing it against how this might feel once all of the pieces are in place.

cli/cmd/proxy/read/command_test.go
t-eckert: Yes, I'll make a note of that. I feel they are tested enough to go in, at least through other PRs and whatnot. Are these tests a blocker for merging?

cli/cmd/proxy/read/command_test.go
t-eckert: Yes, are you ok if I make a note of it and come back around to add the tests?

t-eckert:
I spoke with @ndhanushkodi and she said she was ok for me to delay adding the tests she suggested. I will merge this PR in good faith with her blessing.

PR #1359, opened 2022-07-20
Read created secrets from the API object

thisisnotashwin:
Changes proposed in this PR:
- Avoid reading objects from the cache immediately after creating them as they are not guaranteed to exist. This leads to a bug caused by the race condition.
- Add update RBAC to the secrets for the connect inject controller.
- Assert that the created secret has a resource version upon creating.

How I've tested this PR:
- acceptance tests

- 👀 and faith

Checklist:
- [x] Tests added

### DIFF
--- a/acceptance/tests/peering/peering_connect_test.go
+++ b/acceptance/tests/peering/peering_connect_test.go
@@ -2,7 +2,6 @@ package peering
 
 import (
 	"context"
-	"github.com/hashicorp/go-version"
 	"strconv"
 	"testing"
 
@@ -13,6 +12,7 @@ import (
 	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
 	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-version"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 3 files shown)


acceptance/tests/peering/peering_connect_test.go
```diff
@@ -111,6 +111,9 @@ func TestPeering_Connect(t *testing.T) {
 			helpers.Cleanup(t, cfg.NoCleanupOnFailure, func() {
 				k8s.KubectlDelete(t, staticClientPeerClusterContext.KubectlOptions(t), "../fixtures/bases/peering/peering-acceptor.yaml")
 			})
+			acceptorSecretResourceVersion, err := k8s.RunKubectlAndGetOutputE(t, staticClientPeerClusterContext.KubectlOptions(t), "get", "peering-acceptor", "server", "-o", "jsonpath={.status.secret.resourceVersion}")
```
ishustava, line 114:
```suggestion
			acceptorSecretResourceVersion, err := k8s.RunKubectlAndGetOutputE(t, staticClientPeerClusterContext.KubectlOptions(t), "get", "peeringacceptor", "server", "-o", "jsonpath={.status.secret.resourceVersion}")
```

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -314,13 +314,7 @@ func (r *PeeringAcceptorController) createOrUpdateK8sSecret(ctx context.Context,
...
 		}
 	}
-	// The newly created or updated secret should exist at this point, so we can get it and return the resourceVersion.
-	newSecret := &corev1.Secret{}
-	if err := r.Client.Get(ctx, types.NamespacedName{Name: secretName, Namespace: secretNamespace}, newSecret); err != nil {
-		return "", err
-	}
-
-	return newSecret.ResourceVersion, nil
+	return secret.ResourceVersion, nil
```
ishustava, line 317: do we need the same change in peering dialer?

ishustava: I don't think so because we don't create a secret there

ndhanushkodi (approved):
Nice find!! Faith!!!

This might potentially be the cause of https://app.asana.com/0/1202437710094276/1202600819414278/f and may fix that issue, but we'll see!

PR #1364, opened 2022-07-21
Remove ECR tags for docker images

ishustava:
Changes proposed in this PR:
- We don't need to publish to ECR

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -225,7 +225,6 @@ jobs:
           workdir: control-plane
           tags: |
             docker.io/hashicorp/${{ env.repo }}-control-plane:${{ env.version }}
-            986891699432.dkr.ecr.us-east-1.amazonaws.com/hashicorp/${{ env.repo }}-control-plane:${{ env.version }}
           dev_tags: |
             hashicorppreview/${{ env.repo }}-control-plane:${{ env.version }}
             docker.io/hashicorppreview/${{ env.repo }}-control-plane:${{ env.version }}-${{ github.sha }}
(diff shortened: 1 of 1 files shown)


thisisnotashwin (approved):
🙏

curtbushko (approved):
Sorry about that. I must have added that when I was using another repo (or their example) as a reference!

ishustava:
> Sorry about that. I must have added that when I was using another repo (or their example) as a reference!

all good! I saw it too and didn't question it 😄

ilpianista:
@ishustava `consul-k8s-control-plane` images are missing on ECR, while `consul` and `consul-dataplane` are present. Maybe you need to revert this?

lkysow:
@ilpianista we only publish to ECR to enable our ECS integration since in some cases ECS can't pull from non-ECR repositories. For Kubernetes this isn't a requirement (at least last we tried it) so that's why we don't publish k8s images to ECR. Is that assumption incorrect for your use-case? Thanks for your help!

ilpianista:
@lkysow your assumption is still right. Initially I thought this was a bug because several hashicorp Docker images are on ECR (even `vault` or the `vault-csi-provider`) while only `consul-k8s-control-plane` is missing.

david-yu:
@ilpianista I think this would be a feature request to consider in the future, could you file an issue so we could start to track this request?

ilpianista:
Done, #3091.

Thanks!

PR #1366, opened 2022-07-21
Enable TLS on Consul Peering

thisisnotashwin:
Changes proposed in this PR:
- Allow TLS to be enabled with peering tests.
- Use updated TLS config on Servers and Clients when peering is enabled.

How I've tested this PR:
- Acceptance tests & updates BATS tests.
The updated TLS config was work that had already been done that needed to be reverted in order to continue to support Consul 1.11.

- 👀 

Checklist:
- [x] Tests added

### DIFF
--- a/charts/consul/test/unit/client-daemonset.bats
+++ b/charts/consul/test/unit/client-daemonset.bats
@@ -899,6 +899,48 @@ load _helpers
   [ "${actual}" = "true" ]
 }
 
+@test "client/DaemonSet: sets verify_* flags to true by default when global.tls.enabled and global.peering.enabled" {
+  cd `chart_dir`
+  local command=$(helm template \
+      -s templates/client-daemonset.yaml  \
+      --set 'global.tls.enabled=true' \
+      --set 'global.peering.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | join(" ")' | tee /dev/stderr)
+
+  local actual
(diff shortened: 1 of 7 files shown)


charts/consul/test/unit/client-daemonset.bats
```diff
@@ -899,6 +899,27 @@ load _helpers
...
+  [ "${actual}" = "true" ]
+
+  actual=$(echo $command | jq -r '. | contains("tls { defaults { verify_outgoing = true }}")' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  actual=$(echo $command | jq -r '. | contains("tls { internal_rpc { verify_server_hostname = true }}")' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
 @test "client/DaemonSet: doesn't set the verify_* flags when global.tls.enabled is true and global.tls.verify is false" {
```
ndhanushkodi, line 923: do we need bats tests for these lines in client-daemonset?

```
-hcl='tls { defaults ... \
```

basically any of the lines that starts with "tls { defaults" that has new config, we should make sure if peering is enabled, that those lines have the new config

ndhanushkodi: You have it in the server config test already so I think we just need similar tests here!

ndhanushkodi (commented):
Looks great Ashwin, just thought there were a few helm tests we could add for the client daemonset for the new config.

charts/consul/test/unit/client-daemonset.bats
thisisnotashwin: good catch!! done!!

ishustava (approved):

ndhanushkodi (approved):

PR #1380, opened 2022-07-28
Add extraLabels to connectInject and serverACLInit pods

s-andresen:
Changes proposed in this PR:
Added extraLabels to connectInject and serverACLInit in values.yaml so if declared in the values file then these additional labels will be added to the pods

How I've tested this PR:
Tested the helm chart on my kubernetes cluster with and without declaring the extraLabels to ensure everything works as expected.

Same as how I tested

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -1935,6 +1935,19 @@ connectInject:
   # @type: string
   annotations: null
   
+  # Extra labels to attach to the connect injector pods. This should be a YAML map.
+  #
+  # Example:
+  #
+  # ```yaml
+  # extraLabels:
+  #   labelKey: label-value
+  #   anotherLabelKey: another-label-value
+  # ```
+  #
+  # @type: map
+  extraLabels: null
+  
   # The Docker image for Consul to use when performing Connect injection.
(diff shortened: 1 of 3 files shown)


david-yu:
@s-andresen Could you tell us whether you expect to apply the same labels for both components? We had a similar PR previously but opted for an approach to create a label for all components for a global label instead.

charts/consul/values.yaml
```diff
@@ -2934,3 +2947,17 @@ prometheus:
 # is only useful when running helm template.
 tests:
   enabled: true
+
+serverACLInit:
```
lkysow, line 2951: Let's put this under `global.acls.serverACLInit`?

lkysow (changes requested):
Looking good, but can you please add bats tests? You should be able to follow the pattern for other `extraLabels` tests.

david-yu:
Closing in favor of https://github.com/hashicorp/consul-k8s/pull/1678, could you open a separate PR That enables ServerACLInit with bats test included?

PR #1384, opened 2022-07-29
peering: support setting externalServers hosts in peering token for non-default partitions

ndhanushkodi:
Changes proposed in this PR:
- Support setting externalServers hosts in peering token for non-default partitions

How I've tested this PR:
- Set up 4 clusters in 2 datacenters and tested peering across non-default partitions across datacenters with this [reference config and steps](https://github.com/ndhanushkodi/cluster-peering-demo/tree/main/non-default-non-flat)

- 👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/test/unit/connect-inject-deployment.bats
+++ b/charts/consul/test/unit/connect-inject-deployment.bats
@@ -1883,17 +1883,62 @@ EOF
   [ "${actual}" = "false" ]
 }
 
-@test "connectInject/Deployment: -read-server-expose-service is not set when global.peering.tokenGeneration.serverAddresses.source is not equal to empty string" {
+@test "connectInject/Deployment: -read-server-expose-service and -server-address is not set when global.peering.tokenGeneration.serverAddresses.source is not equal to empty string" {
   cd `chart_dir`
-  local actual=$(helm template \
+  local command=$(helm template \
       -s templates/connect-inject-deployment.yaml  \
       --set 'connectInject.enabled=true' \
       --set 'global.peering.enabled=true' \
       --set 'global.peering.tokenGeneration.serverAddresses.source="notempty"' \
       . | tee /dev/stderr |
-      yq '.spec.template.spec.containers[0].command | any(contains("-read-server-expose-service=true"))' | tee /dev/stderr)
+      yq '.spec.template.spec.containers[0].command')
 
+  local actual=$(echo $command | jq -r ' . | any(contains("-read-server-expose-service=true"))' | tee /dev/stderr)
   [ "${actual}" = "false" ]
+
+  local actual=$(echo $command | jq -r ' . | any(contains("-server-address"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "connectInject/Deployment: when servers are not enabled and externalServers.enabled=true, passes in -server-address flags with hosts" {
+  cd `chart_dir`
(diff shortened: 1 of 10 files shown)


ndhanushkodi:
The unit tests need a different consul binary but they pass locally with the newer binary.

charts/consul/test/unit/connect-inject-deployment.bats
```diff
@@ -1883,19 +1883,64 @@ EOF
...
 
-@test "connectInject/Deployment: -read-server-expose-service is not set when global.peering.tokenGeneration.serverAddresses.source is not equal to empty string" {
+@test "connectInject/Deployment: -read-server-expose-service and -server-address is not set when global.peering.tokenGeneration.serverAddresses.source is not equal to empty string" {
   cd `chart_dir`
-  local actual=$(helm template \
+  local command=$(helm template \
       -s templates/connect-inject-deployment.yaml  \
       --set 'connectInject.enabled=true' \
       --set 'global.peering.enabled=true' \
       --set 'global.peering.tokenGeneration.serverAddresses.source="notempty"' \
```
ishustava, line 1892: not a blocker, but we should probably fail if you provide a value that we don't support

control-plane/connect-inject/peering_acceptor_controller_test.go
```diff
@@ -475,6 +598,11 @@ func TestReconcile_CreateUpdatePeeringAcceptor(t *testing.T) {
 			require.Contains(t, string(decodedTokenData), "\"CA\":null")
 			require.Contains(t, string(decodedTokenData), "\"ServerAddresses\"")
 			require.Contains(t, string(decodedTokenData), "\"ServerName\":\"server.dc1.consul\"")
+			if len(tt.expectedTokenAddresses) > 0 {
+				for _, addr := range tt.externalAddresses {
+					require.Contains(t, string(decodedTokenData), addr)
+				}
+			}
```
ishustava, line 605: ❤️

control-plane/subcommand/inject-connect/command.go
```diff
@@ -194,6 +195,8 @@ func (c *Command) init() {
 		"Enable or disable JSON output format for logging.")
 	c.flagSet.BoolVar(&c.flagReadServerExposeService, "read-server-expose-service", false,
 		"Enables polling the Consul servers' external service for its IP(s).")
+	c.flagSet.Var((*flags.AppendSliceValue)(&c.flagServerAddresses), "server-address",
+		"An address the Consul server(s), formatted host:port, where host may be an IP or DNS name. May be specified multiple times for multiple addresses.")
```
ishustava, line 199:
```suggestion
		"An address of the Consul server(s), formatted host:port, where host may be an IP or DNS name and port must be a gRPC port. May be specified multiple times for multiple addresses.")
```

ishustava (approved):
Looks great ❤️ I left a couple of comments, but they are not blocking.

charts/consul/test/unit/connect-inject-deployment.bats
ndhanushkodi: Will add this in the last stacked PR, so we can fail on values that are not "static" or "consul" or ""

ndhanushkodi: added to the 3rd of the stacked PRs!

PR #1389, opened 2022-08-01
Acceptance test for Envoy Debugging

t-eckert:
Changes proposed in this PR:

- Add ability to run CLI commands without re-compiling each time
- Test CLI Envoy debugging with a happy-path scenario.

How I've tested this PR:

Running the new acceptance test!

Running the new acceptance test.

Checklist:
- [x] Tests added


acceptance/framework/cli/helpers.go
```diff
@@ -0,0 +1,25 @@
+package cli
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TranslateListOutput takes the raw output from the proxy list command and
+// translates the table into a map.
+func TranslateListOutput(raw []byte) map[string]string {
```
thisisnotashwin, line 10: I think this can live in the same file as the connect test for now as it is the only place that it is used. Also, given it is a test fixture, it should be fine to not test this.

thisisnotashwin (commented):
This looks great!! thanks for making the change. I left a comment about not needing to extract behavior to helpers prematurely.. everything else looks good

thisisnotashwin (approved):
EXCELLENT!!

PR #1390, opened 2022-08-01
peering: set peering server config only when peering Helm value is true

ishustava:
Changes proposed in this PR:
- set peering server config only when peering Helm value is true
- This is because Consul now defaults to false: https://github.com/hashicorp/consul/pull/13963

How I've tested this PR:

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/server-config-configmap.yaml
+++ b/charts/consul/templates/server-config-configmap.yaml
@@ -32,6 +32,11 @@ data:
       },
       "recursors": {{ .Values.global.recursors | toJson }},
       "retry_join": ["{{template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc:{{ .Values.server.ports.serflan.port }}"],
+      {{- if .Values.global.peering.enabled }}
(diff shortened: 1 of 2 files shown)


charts/consul/templates/server-config-configmap.yaml
```diff
@@ -32,6 +32,11 @@ data:
       },
       "recursors": {{ .Values.global.recursors | toJson }},
       "retry_join": ["{{template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc:{{ .Values.server.ports.serflan.port }}"],
+      {{- if .Values.global.peering.enabled }}
+      "peering": {
+        "enabled": true
+      },
+      {{- end }}
```
ishustava, line 39: we only want to provide `peering` configuration if the helm value is true and omit it otherwise. This is so that we're compatible with older consul versions.

ndhanushkodi (approved):
Looks great Iryna!!

PR #1392, opened 2022-08-01
peering: support static addresses for exposing servers

ndhanushkodi:
Changes proposed in this PR:
- only requires helm changes because the logic to support non-default partitions also involves just passing in addresses to the token, so this only requires that we set the flags using the appropriate address source

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -36,13 +36,18 @@ global:
     enabled: false
     tokenGeneration:
       serverAddresses:
-        # Source can be set to "" or "consul".
+        # Source can be set to "","consul" or "static".
         #
-        # "" is the default source. If servers are enabled, it will check if server.exposeService is enabled, and read
+        # "" is the default source. If servers are enabled, it will check if `server.exposeService` is enabled, and read
         # the addresses from that service to use as the peering token server addresses.
         #
         # "consul" will use the Consul advertise addresses in the peering token.
+        #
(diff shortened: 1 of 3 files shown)


charts/consul/values.yaml
```diff
@@ -36,13 +36,18 @@ global:
...
         #
-        # "" is the default source. If servers are enabled, it will check if server.exposeService is enabled, and read
+        # "" is the default source. If servers are enabled, it will check if `server.exposeService` is enabled, and read
         # the addresses from that service to use as the peering token server addresses.
         #
         # "consul" will use the Consul advertise addresses in the peering token.
+        #
+        # "static" will use the addresses specified in `global.peering.tokenGeneration.serverAddresses.static`.
         source: ""
+        # Static addresses must be formatted "hostname|ip:port".
```
ishustava, line 48: should we say that the port should be a gRPC port of the consul server?

ishustava (approved):
Looks good!

charts/consul/values.yaml
ndhanushkodi: Good idea!

ndhanushkodi: added!

ndhanushkodi: I also added the failure case from the previous pr here

PR #1399, opened 2022-08-08
peering: generate new token only on user-triggered events 

ishustava:
Changes proposed in this PR:
Previously, we will call generate token endpoint when secret contents change,
i.e. when the resource version in the status of the Acceptor CR doesn't match
the resouce version in the existing secret version. However, this results in a race
condition when multiple reconciles run concurrently. When that happens, one of them ends
up with an outdated CR object and as a result decides that it needs to re-generate the peering
token because those resource versions don't match. This is also undesirable because now
peering tokens have an establishment secret that can only be used one time. If we re-generate
tokens without user knowing, this could result in errors because the dialer will use an outdated secret.

Also, make sure we always update the latest version of the acceptor object by re-fetching
it before updating.

How I've tested this PR:
acceptance tests: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/6913/workflows/523e8352-6cb6-448c-948e-efa60ed46e82

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,7 @@ commands:
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
+                          -consul-image=docker.mirror.hashicorp.services/hashicorppreview/consul-enterprise:1.13-dev \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
                       echo "Tests in ${pkg} failed, aborting early"
@@ -132,6 +133,7 @@ commands:
                       -enable-multi-cluster \
                       ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                       -debug-directory="$TEST_RESULTS/debug" \
+                      -consul-image=docker.mirror.hashicorp.services/hashicorppreview/consul-enterprise:1.13-dev \
                       -consul-k8s-image=<< parameters.consul-k8s-image >>
 
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1000,13 +1000,52 @@ workflows:
             - build-distros-linux
       # Run acceptance tests using the docker image built for the control plane
       - acceptance:
+          name: acceptance1
           context: consul-ci
           requires:
             - dev-upload-docker
       - acceptance-tproxy:
+          name: acceptance2
           context: consul-ci
           requires:
             - dev-upload-docker
+      - acceptance:
+          name: acceptance3
+          context: consul-ci
+          requires:
+            - dev-upload-docker
+      - acceptance-tproxy:
+          name: acceptance4
+          context: consul-ci
+          requires:
+            - dev-upload-docker
+      - acceptance:
+          name: acceptance5
+          context: consul-ci
+          requires:
+            - dev-upload-docker
+      - acceptance-tproxy:
+          name: acceptance6
+          context: consul-ci
+          requires:
+            - dev-upload-docker
+      - acceptance:
+          name: acceptance7
+          context: consul-ci
+          requires:
+            - dev-upload-docker
+      - acceptance-tproxy:
+          name: acceptance8
+          context: consul-ci
+          requires:
+            - dev-upload-docker
+      - acceptance:
+          name: acceptance9
+          context: consul-ci
+          requires:
+            - dev-upload-docker
+      - acceptance-tproxy:
+          name: acceptance10
   nightly-acceptance-tests:
     triggers:
       - schedule:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,7 +99,6 @@ commands:
                     if ! gotestsum --no-summary=all --jsonfile=jsonfile-${pkg////-} -- $pkg -p 1 -timeout 2h -failfast \
                           << parameters.additional-flags >> \
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
-                          -run TestPeering* \
                           -enable-multi-cluster \
                           -debug-directory="$TEST_RESULTS/debug" \
                           -consul-image=docker.mirror.hashicorp.services/hashicorppreview/consul-enterprise:1.13-dev \
@@ -1000,52 +999,13 @@ workflows:
             - build-distros-linux
       # Run acceptance tests using the docker image built for the control plane
       - acceptance:
-          name: acceptance1
           context: consul-ci
           requires:
             - dev-upload-docker
       - acceptance-tproxy:
-          name: acceptance2
           context: consul-ci
           requires:
             - dev-upload-docker
-      - acceptance:
-          name: acceptance3
-          context: consul-ci
-          requires:
-            - dev-upload-docker
-      - acceptance-tproxy:
-          name: acceptance4
-          context: consul-ci
-          requires:
-            - dev-upload-docker
-      - acceptance:
-          name: acceptance5
-          context: consul-ci
-          requires:
-            - dev-upload-docker
-      - acceptance-tproxy:
-          name: acceptance6
-          context: consul-ci
-          requires:
-            - dev-upload-docker
-      - acceptance:
-          name: acceptance7
-          context: consul-ci
-          requires:
-            - dev-upload-docker
-      - acceptance-tproxy:
-          name: acceptance8
-          context: consul-ci
-          requires:
-            - dev-upload-docker
-      - acceptance:
-          name: acceptance9
-          context: consul-ci
-          requires:
-            - dev-upload-docker
-      - acceptance-tproxy:
-          name: acceptance10
   nightly-acceptance-tests:
     triggers:
       - schedule:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,7 @@ commands:
                     if ! gotestsum --no-summary=all --jsonfile=jsonfile-${pkg////-} -- $pkg -p 1 -timeout 2h -failfast \
                           << parameters.additional-flags >> \
                           ${ENABLE_ENTERPRISE:+-enable-enterprise} \
+                          -run TestPeering* \
                           -enable-multi-cluster \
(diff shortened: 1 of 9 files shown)


acceptance/framework/k8s/kubectl.go
```diff
@@ -94,15 +94,15 @@ func KubectlApplyK(t *testing.T, options *k8s.KubectlOptions, kustomizeDir strin
 // deletes it from the cluster by running 'kubectl delete -f'.
 // If there's an error deleting the file, fail the test.
 func KubectlDelete(t *testing.T, options *k8s.KubectlOptions, configPath string) {
-	_, err := RunKubectlAndGetOutputE(t, options, "delete", "-f", configPath)
+	_, err := RunKubectlAndGetOutputE(t, options, "delete", "--timeout=60s", "-f", configPath)
```
ishustava, line 97: Adding these so that we don't hang forever in case CRD deletion hands

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -142,14 +141,12 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
...
-				if err != nil {
-					r.updateStatusError(ctx, acceptor, KubernetesError, err)
-					return ctrl.Result{}, err
-				}
+		if acceptor.SecretRef() != nil {
+			r.Log.Info("stale secret in status; deleting stale secret", "name", acceptor.Name, "secret-name", acceptor.SecretRef().Name)
+			err = r.deleteK8sSecret(ctx, acceptor.SecretRef().Name, acceptor.Namespace)
+			if err != nil {
+				r.updateStatusError(ctx, acceptor, KubernetesError, err)
+				return ctrl.Result{}, err
```
ishustava, line 149: This diff and some below simplify some of the conditional logic. It also changes everything to only look at the spec secret except for the case when we need to delete a stale secret from status

thisisnotashwin (approved):
This is excellent!! I have left a couple comments neither of which are blocking but are nice to haves. This was an excellent find and love how much simpler this is now!

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -223,48 +211,47 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
...
+	if existingSecret == nil {
 		return true, false, nil
 	}
+
 	return false, false, nil
 }
 
 // updateStatus updates the peeringAcceptor's secret in the status.
-func (r *PeeringAcceptorController) updateStatus(ctx context.Context, acceptor *consulv1alpha1.PeeringAcceptor, secretResourceVersion string) error {
+func (r *PeeringAcceptorController) updateStatus(ctx context.Context, acceptorObjKey types.NamespacedName) error {
```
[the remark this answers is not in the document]
ishustava: yes! I'm sorry i forgot to do it in the dialer!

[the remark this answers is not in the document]
thisisnotashwin: no worries 😄

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -159,62 +156,53 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
 			return ctrl.Result{}, err
 		}
 		if acceptor.Secret().Backend == "kubernetes" {
-			secretResourceVersion, err = r.createOrUpdateK8sSecret(ctx, acceptor, resp)
+			err = r.createOrUpdateK8sSecret(ctx, acceptor, resp)
 			if err != nil {
```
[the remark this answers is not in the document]
ishustava: yes absolutely! this isn't even a nit 😄

control-plane/connect-inject/peering_acceptor_controller.go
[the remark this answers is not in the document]
thisisnotashwin: thank you! this is just excellence in the form of code!

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -223,48 +207,47 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
...
 
 // updateStatus updates the peeringAcceptor's secret in the status.
-func (r *PeeringAcceptorController) updateStatus(ctx context.Context, acceptor *consulv1alpha1.PeeringAcceptor, secretResourceVersion string) error {
+func (r *PeeringAcceptorController) updateStatus(ctx context.Context, acceptorObjKey types.NamespacedName) error {
+	// Get the latest resource before we update it.
+	acceptor := &consulv1alpha1.PeeringAcceptor{}
+	err := r.Client.Get(ctx, acceptorObjKey, acceptor)
+	if err != nil {
+		return fmt.Errorf("error fetching acceptor resource before status update: %w", err)
+	}
```
thisisnotashwin, line 248: nit (in case you have time): can be one-lined. not a blocker.

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -278,7 +261,7 @@ func (r *PeeringAcceptorController) updateStatus(ctx context.Context, acceptor *
 			acceptor.Status.LatestPeeringVersion = pointerToUint64(peeringVersion)
 		}
 	}
-	err := r.Status().Update(ctx, acceptor)
+	err = r.Status().Update(ctx, acceptor)
 	if err != nil {
 		r.Log.Error(err, "failed to update PeeringAcceptor status", "name", acceptor.Name, "namespace", acceptor.Namespace)
 	}
```
thisisnotashwin, line 267: nit (in case you have time): can be one-lined. not a blocker.

thisisnotashwin (approved):
Found a few more nits for you in case you were interested 😉

control-plane/connect-inject/peering_acceptor_controller.go
ishustava: this one we can't update to one-liner because it'll change the scope of `err` and so we won't have the right error when we return on the below line (I already looked 😄 )

thisisnotashwin: ohno!! good point!! thanks so much for checking Iryna!!

control-plane/connect-inject/peering_acceptor_controller.go
```diff
@@ -223,48 +207,47 @@ func (r *PeeringAcceptorController) Reconcile(ctx context.Context, req ctrl.Requ
...
+			if err != nil {
+				return false, false, err
+			}
+			if acceptor.Status.LatestPeeringVersion == nil || *acceptor.Status.LatestPeeringVersion < peeringVersion {
+				return true, false, nil
+			}
+		}
+	}
 
-	} else {
```
ndhanushkodi, line 257: if the acceptor.SecretRef() == nil, we used to set shouldGenerate to true. right now this function would return false, but if we've somehow managed to get an acceptor that has a secret in the spec but the status doesn't reference that secret AND for some reason that peering still is leftover in Consul (maybe because its not fully cleaned up), then something in our controller has out of date state, and I think in that case we'd want to generate a new token. It may happen if someone has deleted the acceptor, and maybe the peering didn't fully get deleted in Consul because of the deferred deletion, and then creates a new acceptor.

control-plane/connect-inject/peering_acceptor_controller_test.go
```diff
@@ -503,9 +493,53 @@ func TestReconcile_CreateUpdatePeeringAcceptor(t *testing.T) {
 				}
 				return []*corev1.Secret{secret}
 			},
-			expectDeletedK8sSecret: &types.NamespacedName{
-				Name:      "some-old-secret",
-				Namespace: "default",
+			initialConsulPeerName: "acceptor-created",
+		},
+		{
+			name: "Peering exists in Consul, but secret doesn't and status is not set",
```
ndhanushkodi, line 499: My comment above is for this case-- and I think it'd mean adding an expShouldGenerate = true here

ndhanushkodi (approved):
This looks great, thank you for making this fix!!
I think we need to add just one more case where shouldGenerate should be true, but let me know what you think, and then I think this looks great to merge!

.circleci/config.yml
```diff
@@ -997,13 +1000,52 @@ workflows:
             - build-distros-linux
       # Run acceptance tests using the docker image built for the control plane
       - acceptance:
+          name: acceptance1
```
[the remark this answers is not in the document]
ndhanushkodi: deleting, you've already updated this!

control-plane/connect-inject/peering_acceptor_controller.go
ndhanushkodi: I think this case is similar to name/key/backend being different between spec and status, it would just handle if status is empty but spec is not when a peering exists

ishustava: >if the acceptor.SecretRef() == nil, we used to set shouldGenerate to true. right now this function would return false

If secret ref is nil (i.e. no secret in status), then we will check if existing secret is nil, and if it is we will return true for `shouldGenerate`, otherwise we will return false.

ishustava: Sorry 🤦‍♀️ if existing secret is not nil, then yeah we'll return false and we won't regenerate because we want to instead update the status in the CR to avoid extra calls to generate token

ndhanushkodi: Yup this makes sense, we'd want to keep the existing secret if its there!

control-plane/connect-inject/peering_acceptor_controller_test.go
ishustava: I think `expShouldGenerate` one would be for the tests for `shouldGenerate` function. The behavior of that function actually didn't change (i just re-ordered some if conditions), so I didn't need to change the tests there. this test already asserts that the secret is created

PR #1401, opened 2022-08-09
peering: update unit tests to not reuse peering token

ndhanushkodi:
Changes proposed in this PR:
- Removing some code that edits the peering token after generating to add server addresses(we don't need to do this anymore because the token has the correct server addresses in it)
- Doesn't reuse the same peering token twice-- if a peering token has already been used, we need to generate a new one if we want to update the peering during a reconcile loop.

How I've tested this PR:
- Ran unit tests with 1.13

- 👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


.github/workflows/test.yml
```diff
@@ -146,7 +146,7 @@ jobs:
         working-directory: control-plane
         run: |
           mkdir -p $HOME/bin
-          wget https://github.com/ndhanushkodi/binaries/releases/download/v4.1oss/consul -O consulbin && \
+          wget https://github.com/ndhanushkodi/binaries/releases/download/v6oss/consul -O consulbin && \
```
ndhanushkodi, line 149: This is based of of 1.13.x branch of consul

thisisnotashwin (approved):

ishustava (approved):

.github/workflows/test.yml
ishustava: Should we wait for core release before merging?

ndhanushkodi: ya that works or I can make an explicit revert change to switch back to releases.hashicorp.com right after core release-- i think if it takes them past the early afternoon, then I'll merge this for today and make the explicit revert change! that way we can have our unit tests running against 1.13 on main as we're merging other k8s PRs too

ndhanushkodi: theyve released! So i'll update this PR

PR #1411, opened 2022-08-11
feat: Add support for webhook-cert-manager container resources configuration

jwitko:
Changes proposed in this PR:
- Add support for webhook-cert-manager container resources configuration via values file

How I've tested this PR:
- I have vendored this chart, with these changes, locally and use it for my production deployments

- There should be no change in testing process for this MR as compared to existing tests
- A valid test would be to template the helm chart without modification and ensure there is the expected default values
- A valid test would be to template the helm chart after modifying the default values and ensure updated values appear
- A valid test would be to deploy consul using the standard base chart into a kubernetes cluster then switch to this branch and run a diff or an additional deployment and ensure no changes happened.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/webhook-cert-manager-deployment.yaml
+++ b/charts/consul/templates/webhook-cert-manager-deployment.yaml
@@ -47,11 +47,11 @@ spec:
         name: webhook-cert-manager
         resources:
           limits:
-            cpu: 100m
-            memory: 50Mi
+            cpu: {{ .Values.webhookCertManager.resources.limits.cpu }}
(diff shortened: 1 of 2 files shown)


jmurret:
Thank you @jwitko for the contribution!  This looks good.  We are in the midst of a release and will take a look at this next week and get back to you.

kjbflood (approved):
Looks good to me and solves an issue that I face currently.

jwitko:
@jmurret Poke :)

PR #1412, opened 2022-08-11
Fix some smaller Envoy Debugging bugs. :bug:

t-eckert:
This PR begins to address some of the smaller bugs found in the Envoy Debugging work.

Changes proposed in this PR:
- Clusters are now displayed by their names (e.g. the first part of the fully qualified domain name (FQDN) up to `.`).
- Clusters with FQDNs that are the same as their names, therefore not "domain names", are not displayed as FQDNs. For instance, `local_app` is no longer shown as an FQDN.
- If the `proxy list` command returns no pods, a message is printed saying none were found. This is similar to the behavior of kubectl.
- If the user passes a field filter and table filter where the field doesn't appear in the filtered tables, a warning is printed saying that the field filter does not apply. For example, if a user were to pass `-fqdn default -listeners`, they would get a warning as FQDN is not a property of the Listeners table.

How I've tested this PR:

- I added unit tests!

- Running the unit tests or messing around with the CLI manually.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


cli/cmd/proxy/list/command.go
```diff
@@ -150,23 +148,27 @@ func (c *ListCommand) initKubernetes() error {
...
 	} else if c.flagNamespace != "" {
-		c.namespace = c.flagNamespace
+		return c.flagNamespace
 	} else {
-		c.namespace = settings.Namespace()
+		return settings.Namespace()
 	}
-
-	return err
 }
```
t-eckert, line 164: I pulled `namespace` out into its own method because when it was part of `initKubernetes`, which only runs when `c.kubernetes` is `nil`, the `namespace` being passed in was ignored during tests as they patch in a fake Kubernetes client to `c.kubernetes`, bypassing the method call.

thisisnotashwin (approved):
This looks good. I have one comment, but once that is addressed, I'm comfortable not needing to review the PR another time. Nice work.

cli/cmd/proxy/list/command.go
```diff
@@ -150,23 +148,27 @@ func (c *ListCommand) initKubernetes() error {
 		return fmt.Errorf("error creating Kubernetes client %v", err)
 	}
 
+	return err
```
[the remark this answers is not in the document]
t-eckert: Great catch! I'll make the change through a suggestion.

```suggestion
	return nil
```

cli/cmd/proxy/read/command_test.go
```diff
@@ -146,6 +140,41 @@ func TestReadCommandOutput(t *testing.T) {
...
 
+// TestFilterWarnings ensures that a warning is printed if the user applies a
+// field filter (e.g. -fqdn default) and a table filter (e.g. -secrets) where
+// the former does not affect the output of the latter.
+func TestFilterWarnings(t *testing.T) {
+	podName := "fakePod"
+	cases := map[string][]string{
+		"fully qualified domain name doesn't apply to secrets":                {"-fqdn", "default", "-secrets"},
+		"port doesn't apply to secrets or routes":                             {"-port", "8080", "-secrets", "-routes"},
+		"fully qualified domain name doesn't apply to endpoints or listeners": {"-fqdn", "default", "-endpoints", "-listeners"},
```
ishustava, line 151: I think we need test cases for:
- `-address` flag warning
- all possible combindations when `-fqdn` flag would produce a warning
- all possible combinations when `-port` flag would produce a warning
- a case when mutliple warnings are returned
- happy case scenarios when no warnings are returned (if that's not tested elsewhere).

cli/cmd/proxy/read/config.go
```diff
@@ -219,9 +219,15 @@ func parseClusters(rawCfg map[string]interface{}, clusterMapping map[string][]st
 			}
 		}
 
+		// Don't display "non-domain name" FQDNs (e.g. local_app)
+		fqdn := cluster.Cluster.FQDN
```
ishustava, line 223: Looking at this a bit more, I think FQDNs are not required to have `.`. For example, `localhost` is a valid FQDN. So I think we should skip this change altogether because there won't be a straightforward way to determine whether a string is an FQDN or not

ishustava (commented):
Great work so far, Thomas! Leaving some comments in-line

cli/cmd/proxy/read/config.go
t-eckert: I think that makes sense. Envoy refers to `local_app` as an FQDN so I think we can get away with doing the same.

cli/cmd/proxy/read/command_test.go
```diff
@@ -146,6 +140,97 @@ func TestReadCommandOutput(t *testing.T) {
...
+			c.fetchConfig = func(context.Context, common.PortForwarder) (*EnvoyConfig, error) {
+				return testEnvoyConfig, nil
+			}
+
+			exitCode := c.Run(append([]string{podName}, tc.input...))
+			require.Equal(t, 0, exitCode) // This shouldn't error out, just warn the user.
+
+			if tc.shouldWarn {
+				require.Regexp(t, "The filter.*does not apply to the tables displayed.", buf.String())
+			} else {
```
ishustava, line 227: how do we check that multiple warnings are displayed? Also, could we check for the specific warning string that's applicable to each case?

t-eckert: I improved it to check for a list of warnings so we can confirm multiple warnings are produced. :)

ishustava (approved):

PR #1414, opened 2022-08-12
Various acceptance tests fixes

ishustava:
Changes proposed in this PR:
- parallelize nightly tests (they now take ~5h which is the max time circleci will allow us to run w/o output)
- re-establish port forwarding if we lose connection. Sometimes, we lose port-forwarding connection and never re-establish it, which leads to any calls to consul to fail. Instead, we now monitor the local port, and if we can't connect to it, we re-establish port-forwarding
- Add retries to aws cleanup script

How I've tested this PR:
ran them in ci: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/7036/workflows/0f11f813-0c69-4090-a9ae-63ba844e101e

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/framework/consul/helm_cluster.go
+++ b/acceptance/framework/consul/helm_cluster.go
@@ -311,6 +311,7 @@ func (h *HelmCluster) CreatePortForwardTunnelToResourcePort(t *testing.T, resour
 
 func (h *HelmCluster) monitorPortForwardedServer(t *testing.T, port int, tunnel *terratestk8s.Tunnel, doneChan chan bool, resourceName string, remotePort int) {
 	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
 
 	for {
 		select {
@@ -320,9 +321,8 @@ func (h *HelmCluster) monitorPortForwardedServer(t *testing.T, port int, tunnel 
 			return
 		case <-ticker.C:
 			conn, err := net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", port))
-			logger.Log(t, "lost connection to port-forwarded server", "port", port)
 			if err != nil {
-				logger.Log(t, "restarting port-forwarding", "port", port)
+				logger.Log(t, "lost connection to port-forwarded server; restarting port-forwarding", "port", port)
 				tunnel.Close()
 				tunnel = terratestk8s.NewTunnelWithLogger(
 					h.helmOptions.KubectlOptions,
--- a/acceptance/framework/consul/helm_cluster.go
+++ b/acceptance/framework/consul/helm_cluster.go
@@ -3,6 +3,7 @@ package consul
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


acceptance/framework/consul/helm_cluster.go
```diff
@@ -297,12 +298,51 @@ func (h *HelmCluster) CreatePortForwardTunnelToResourcePort(t *testing.T, resour
...
+		close(doneChan)
 	})
 
+	go h.monitorPortForwardedServer(t, localPort, tunnel, doneChan, resourceName, remotePort)
+
 	return fmt.Sprintf("127.0.0.1:%d", localPort)
+}
 
+func (h *HelmCluster) monitorPortForwardedServer(t *testing.T, port int, tunnel *terratestk8s.Tunnel, doneChan chan bool, resourceName string, remotePort int) {
+	ticker := time.NewTicker(1 * time.Second)
```
curtbushko, line 313: Do you need a defer ticker.Stop() somewhere?

acceptance/framework/consul/helm_cluster.go
```diff
@@ -297,12 +298,51 @@ func (h *HelmCluster) CreatePortForwardTunnelToResourcePort(t *testing.T, resour
...
+
+	for {
+		select {
+		case <-doneChan:
+			logger.Log(t, "stopping monitor of the port-forwarded server")
+			tunnel.Close()
+			return
+		case <-ticker.C:
+			conn, err := net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", port))
+			logger.Log(t, "lost connection to port-forwarded server", "port", port)
```
curtbushko, line 323: Is this log message true 100% of the time?  (Or am I miss-understanding this) The ticker is running every 1 second, you establish a connection, check if there was an error, if there was an error re-establish tunnel.

hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -379,11 +377,17 @@ func realMain(ctx context.Context) error {
 		})
 		for _, igw := range igws.InternetGateways {
 			fmt.Printf("Internet gateway: Detaching from VPC... [id=%s]\n", *igw.InternetGatewayId)
-			_, err := ec2Client.DetachInternetGatewayWithContext(ctx, &ec2.DetachInternetGatewayInput{
-				InternetGatewayId: igw.InternetGatewayId,
-				VpcId:             vpcID,
-			})
-			if err != nil {
+			if err := destroyBackoff(ctx, "Internet Gateway", *igw.InternetGatewayId, func() error {
```
curtbushko, line 380: Not to be address here but are there several more of these resources that need to be cleaned up with a backoff also?  How about on GKE?

curtbushko (commented):
Looks pretty good. I just had some questions for my learnings.

t-eckert (approved):
Curt covered anything I thought might need coverage. Overall looks good! Happy to have parallelized tests!

acceptance/framework/consul/helm_cluster.go
ishustava: great catch! we absolutely do!

acceptance/framework/consul/helm_cluster.go
ishustava: Great question. No it's not right. It should only be printed if there's an error. Great catch and thank you!!

hack/aws-acceptance-test-cleanup/main.go
ishustava: On GKE we only need to cleanup k8s clusters because we don't create a separate VPC or any vpc-related resources I think.

curtbushko (approved):
Come on successful nightly tests!

PR #1426, opened 2022-08-16
Rename flag for peering token server addresses.

thisisnotashwin:
- The flag was initally named `server-address` but the `server-address` flag has been used across the project to imply ONLY the address of the external server when they are enabled. This meaning is overloaded in the context of the server address used for generating the peering token. This leads to errors (specifically in the agentless context), where the deployment needs to explicitly know the list of external servers.

How I've tested this PR:
- bats
- acceptance tests

- 👀 
- comments on the name of the flag.

### DIFF
--- a/control-plane/go.mod
+++ b/control-plane/go.mod
@@ -26,6 +26,7 @@ require (
 	k8s.io/apimachinery v0.22.2
 	k8s.io/client-go v0.22.2
 	k8s.io/klog/v2 v2.9.0
+	k8s.io/utils v0.0.0-20220812165043-ad590609e2e5
 	sigs.k8s.io/controller-runtime v0.10.2
 )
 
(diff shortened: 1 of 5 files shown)


control-plane/go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	k8s.io/apimachinery v0.22.2
 	k8s.io/client-go v0.22.2
 	k8s.io/klog/v2 v2.9.0
+	k8s.io/utils v0.0.0-20220812165043-ad590609e2e5
```
thisisnotashwin, line 29: This seems related to changes not made in the PR but caused automatically by a `go mod tidy`.

ishustava (approved):

control-plane/go.mod
kschoche: Hrm, this was a change I made yesterday for the `pointer` changes.. I wonder how that worked..

thisisnotashwin: yeah..not quiet sure myself

kschoche (approved):

PR #1431, opened 2022-08-16
Kubernetes 1.24 support

### LINKED ISSUE #1145: Kubernetes 1.21+: Token-handling code assumes auto-created service account token secrets
### Community Note

---

### Overview of the Issue

Sweeping token-scraping of auto-generated Kubernetes token secrets in preparation for Kubernetes 1.24 showed the following code locations assume auto-generated tokens will exist:

https://github.com/hashicorp/consul-k8s/blob/7ae81732981bb971243b52e4b25f08e80f5ccfd7/control-plane/connect-inject/handler.go#L279-L283

https://github.com/hashicorp/consul-k8s/blob/7ae81732981bb971243b52e4b25f08e80f5ccfd7/acceptance/framework/vault/vault_cluster.go#L173-L180

That assumption is not universally correct.

In 1.21+, secret-based tokens are no longer used for mounting into pods (ephemeral time-limited tokens are), and the token controller can be turned off.

In 1.24+, secret-based tokens are no longer auto-created by default for new service accounts.

### Reproduction Steps

Create a 1.24 Kubernetes cluster or a 1.21+ Kubernetes cluster with the token controller disabled.

### Additional Context

Using ephemeral time-bound tokens is preferred in 1.21+ (see the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API) if possible.

If a secret-based token is still desired, one can be [created manually](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token), but will not be referenced from the service account's .secrets list.

kschoche:
Kubernetes 1.24 no longer automatically deploys a Secret alongside a ServiceAccount, so pods no longer have a long term JWT token that can be used. The recommendation from Kubernetes is to either use the temporary JWT token associated with the ServiceAccount or to deploy your own Kubernetes Secret which references the ServiceAccount via annotation.
Fixes #1145

This PR introduces/uses both models:
* ServiceAccounts which require a long-term JWT token will now be deployed with a Kube secret. This is consumed by things like Consul auth methods where we need a stable JWT token that Consul can validate against the TokenReview API. 
* ServiceMesh applications do not need a long-term JWT token since they only use the Consul auth method at startup to fetch an ACL token for Envoy. We will use the temporary JWT token that comes with the ServiceAccount from Kubernetes. As a result of this no UX changes are expected for end-users.
* ServiceMesh applications which use the multiport workaround ***must*** deploy a Kube Secret which references its respective ServiceAccount and has the same name as the ServiceAccount in the same namespace.
This is a **breaking change** and will be included in a follow-on docs update.

Example:
```
apiVersion: v1
kind: Secret
metadata:
  name: multiport-svc1
  annotations:
    kubernetes.io/service-account.name: multiport-svc1
type: kubernetes.io/service-account-token
---
apiVersion: v1
kind: Secret
metadata:
  name: multiport-svc2
  annotations:
    kubernetes.io/service-account.name: multiport-svc2
type: kubernetes.io/service-account-token
```

Changes proposed in this PR:
- Adds a deployed Kube Secret for the component auth method.
- Updates `server-acl-init` to check for the existence of the deployed secret IF it's ServiceAccount does not have SecretRefs. This should preserve backward compatibility with < Kubernetes 1.24
- Updates the `mesh_webhook` to check for and mount Secrets associated with multiport applications. Previously we would iterate through the ServiceAccounts referenced by the multiport service and grab the JWT tokens from their secrets and add them as volumeMounts to the pod. Now we also check for the user's deployed secrets and grab those as volumeMounts. The assumption is that these Secrets share the same name as the ServiceAccounts.
- Updates circleci to use Kube-1.24 for acceptance tests.
- Update Kind version in acceptance tests to latest.
- Update some acceptance tests to also deploy secrets for Vault Kube Auth Methods and multiport.

UX Changes for users:
* Users of multiport workarounds must deploy a secret alongside their serviceaccount for each service.
* Users of the Vault Secrets Backend must also deploy a Secret for each Kube Auth Method that is configured (1 / datacenter). This is a requirement for Vault's own support of Kube-1.24+
* A [docs PR](https://github.com/hashicorp/consul/pull/14339) has been raised to update based on the above.

How I've tested this PR:
Acceptance + unit tests pass 🍏 
Manual test with Kind running Kubernetes 1.24 🍏 
While all acceptance tests will now run against Kind 1.24, the various nightly tests will still run differing versions of Kubernetes for some additional coverage. I propose we possibly extend our nightly tests to also run various Kube versions as a follow-on PR.

Update Kind to the latest version and deploy a cluster by default. This should bring in Kubernetes 1.24 (or 1.25 starting next week), and deploy Consul or run the acceptance tests using a consul-k8s image based on this branch.
👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/acceptance/framework/k8s/kubectl.go
+++ b/acceptance/framework/k8s/kubectl.go
@@ -102,8 +102,9 @@ func KubectlDelete(t *testing.T, options *k8s.KubectlOptions, configPath string)
 // deletes it from the cluster by running 'kubectl delete -k'.
 // If there's an error deleting the file, fail the test.
 func KubectlDeleteK(t *testing.T, options *k8s.KubectlOptions, kustomizeDir string) {
-	_, err := RunKubectlAndGetOutputE(t, options, "delete", "--timeout=60s", "-k", kustomizeDir)
-	require.NoError(t, err)
+	_, _ = RunKubectlAndGetOutputE(t, options, "delete", "--timeout=60s", "-k", kustomizeDir)
+	// TODO: this is to prevent test failures when Kubernetes automatically cleans up the kube secrets, fix me.
+	//	require.NoError(t, err)
 }
 
 // RunKubectl runs an arbitrary kubectl command provided via args and ignores the output.
--- a/acceptance/framework/k8s/kubectl.go
+++ b/acceptance/framework/k8s/kubectl.go
@@ -102,9 +102,10 @@ func KubectlDelete(t *testing.T, options *k8s.KubectlOptions, configPath string)
 // deletes it from the cluster by running 'kubectl delete -k'.
 // If there's an error deleting the file, fail the test.
 func KubectlDeleteK(t *testing.T, options *k8s.KubectlOptions, kustomizeDir string) {
-	_, _ = RunKubectlAndGetOutputE(t, options, "delete", "--timeout=60s", "-k", kustomizeDir)
-	// TODO: this is to prevent test failures when Kubernetes automatically cleans up the kube secrets, fix me.
-	//	require.NoError(t, err)
+	// Ignore not found errors because Kubernetes automatically cleans up the kube secrets that we deployed
+	// referencing the ServiceAccount when it is deleted.
+	_, err := RunKubectlAndGetOutputE(t, options, "delete", "--timeout=60s", "--ignore-not-found", "-k", kustomizeDir)
+	require.NoError(t, err)
 }
 
 // RunKubectl runs an arbitrary kubectl command provided via args and ignores the output.
--- a/acceptance/framework/vault/vault_cluster.go
+++ b/acceptance/framework/vault/vault_cluster.go
@@ -166,15 +166,27 @@ func (v *VaultCluster) ConfigureAuthMethod(t *testing.T, vaultClient *vapi.Clien
 	// service account token.
 	// The JWT token and CA cert is what Vault server will use to validate service account token
 	// with the Kubernetes API.
+	secretName := saName
 	var sa *corev1.ServiceAccount
 	retry.Run(t, func(r *retry.R) {
 		sa, err = v.kubernetesClient.CoreV1().ServiceAccounts(saNS).Get(context.Background(), saName, metav1.GetOptions{})
 		require.NoError(r, err)
-		require.Len(r, sa.Secrets, 1)
+		// In Kube 1.21+ the serviceAccount does not have a secret automatically generated.
+		if len(sa.Secrets) == 0 {
+			_, err = v.kubernetesClient.CoreV1().Secrets(saNS).Create(context.Background(), &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        secretName,
+					Annotations: map[string]string{"kubernetes.io/service-account.name": saName},
+				},
+			}, metav1.CreateOptions{})
+			require.NoError(t, err)
+		} else {
+			secretName = sa.Secrets[0].Name
+		}
 	})
 
 	v.logger.Logf(t, "updating vault kubernetes auth config for %s auth path", authPath)
-	tokenSecret, err := v.kubernetesClient.CoreV1().Secrets(saNS).Get(context.Background(), sa.Secrets[0].Name, metav1.GetOptions{})
+	tokenSecret, err := v.kubernetesClient.CoreV1().Secrets(saNS).Get(context.Background(), secretName, metav1.GetOptions{})
 	require.NoError(t, err)
 	_, err = vaultClient.Logical().Write(fmt.Sprintf("auth/%s/config", authPath), map[string]interface{}{
 		"token_reviewer_jwt": string(tokenSecret.Data["token"]),
--- a/acceptance/framework/vault/vault_cluster.go
+++ b/acceptance/framework/vault/vault_cluster.go
@@ -153,9 +153,8 @@ func (v *VaultCluster) bootstrap(t *testing.T, vaultNamespace string) {
 	// In Kube 1.24+ the serviceAccount does not have a secret automatically generated so we will create one prior to
 	// needing to use it in ConfigureAuthMethod. Vault expects the user to create the secret manually for any
 	// Kube AuthMethod use.
-	var sa *corev1.ServiceAccount
 	retry.Run(t, func(r *retry.R) {
-		sa, err = v.kubernetesClient.CoreV1().ServiceAccounts(namespace).Get(context.Background(), vaultServerServiceAccountName, metav1.GetOptions{})
+		sa, err := v.kubernetesClient.CoreV1().ServiceAccounts(namespace).Get(context.Background(), vaultServerServiceAccountName, metav1.GetOptions{})
 		require.NoError(r, err)
 		if len(sa.Secrets) == 0 {
 			_, err = v.kubernetesClient.CoreV1().Secrets(namespace).Create(context.Background(), &corev1.Secret{
--- a/acceptance/framework/vault/vault_cluster.go
+++ b/acceptance/framework/vault/vault_cluster.go
@@ -149,6 +149,23 @@ func (v *VaultCluster) bootstrap(t *testing.T, vaultNamespace string) {
 
 	namespace := v.helmOptions.KubectlOptions.Namespace
 	vaultServerServiceAccountName := fmt.Sprintf("%s-vault", v.releaseName)
+
+	// In Kube 1.24+ the serviceAccount does not have a secret automatically generated so we will create one prior to
+	// needing to use it in ConfigureAuthMethod. Vault expects the user to create the secret manually for any
+	// Kube AuthMethod use.
+	sa, err := v.kubernetesClient.CoreV1().ServiceAccounts(vaultServerServiceAccountName).Get(context.Background(), vaultServerServiceAccountName, metav1.GetOptions{})
+	require.NoError(t, err)
+	if len(sa.Secrets) == 0 {
+		_, err = v.kubernetesClient.CoreV1().Secrets(namespace).Create(context.Background(), &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        vaultServerServiceAccountName,
+				Annotations: map[string]string{"kubernetes.io/service-account.name": vaultServerServiceAccountName},
+			},
+			Type: "kubernetes.io/service-account-token",
+		}, metav1.CreateOptions{})
+		require.NoError(t, err)
+	}
+
 	v.ConfigureAuthMethod(t, v.vaultClient, "kubernetes", "https://kubernetes.default.svc", vaultServerServiceAccountName, namespace)
 }
 
@@ -163,7 +180,8 @@ func (v *VaultCluster) ConfigureAuthMethod(t *testing.T, vaultClient *vapi.Clien
 	require.NoError(t, err)
 
 	// To configure the auth method, we need to read the token and the CA cert from the auth method's
-	// service account token.
+	// service account token. In Kube-1.24 and above the secret is not automatically generated so we
+	// rely on it being created prior to calling ConfigureAuthMethod.
 	// The JWT token and CA cert is what Vault server will use to validate service account token
 	// with the Kubernetes API.
 	secretName := saName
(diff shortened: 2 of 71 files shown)


acceptance/framework/vault/vault_cluster.go
```diff
@@ -149,6 +149,26 @@ func (v *VaultCluster) bootstrap(t *testing.T, vaultNamespace string) {
...
 	vaultServerServiceAccountName := fmt.Sprintf("%s-vault", v.releaseName)
+
+	// In Kube 1.24+ the serviceAccount does not have a secret automatically generated so we will create one prior to
+	// needing to use it in ConfigureAuthMethod. Vault expects the user to create the secret manually for any
+	// Kube AuthMethod use.
+	var sa *corev1.ServiceAccount
+	retry.Run(t, func(r *retry.R) {
+		sa, err = v.kubernetesClient.CoreV1().ServiceAccounts(namespace).Get(context.Background(), vaultServerServiceAccountName, metav1.GetOptions{})
+		require.NoError(r, err)
+		if len(sa.Secrets) == 0 {
```
kschoche, line 160:
```suggestion
	retry.Run(t, func(r *retry.R) {
		sa, err := v.kubernetesClient.CoreV1().ServiceAccounts(namespace).Get(context.Background(), vaultServerServiceAccountName, metav1.GetOptions{})
		require.NoError(r, err)
		if len(sa.Secrets) == 0 {
```

acceptance/tests/vault/vault_test.go
```diff
@@ -203,6 +203,7 @@ func TestVault(t *testing.T) {
 	vaultCASecret := vault.CASecretName(vaultReleaseName)
 
 	consulHelmValues := map[string]string{
+		"global.imageK8S":             "kyleschochenmaier/consul-k8s-dev",
```
kschoche, line 206:
```suggestion
```

control-plane/connect-inject/mesh_webhook.go
```diff
@@ -292,22 +292,31 @@ func (w *MeshWebhook) Handle(ctx context.Context, req admission.Request) admissi
...
 					sa, err := w.Clientset.CoreV1().ServiceAccounts(req.Namespace).Get(ctx, svc, metav1.GetOptions{})
 					if err != nil {
 						w.Log.Error(err, "couldn't get service accounts")
 						return admission.Errored(http.StatusInternalServerError, err)
 					}
 					if len(sa.Secrets) == 0 {
+						// check to see if there is a secret with the same name as the serviceaccount for Kube1.21+
 						w.Log.Info(fmt.Sprintf("service account %s has zero secrets exp at least 1", svc))
-						return admission.Errored(http.StatusInternalServerError, fmt.Errorf("service account %s has zero secrets, expected at least one", svc))
+						sec, err := w.Clientset.CoreV1().Secrets(req.Namespace).Get(ctx, svc, metav1.GetOptions{})
```
kschoche, line 304: TODO: Need to add unit tests for this section.

control-plane/connect-inject/mesh_webhook.go
```diff
@@ -292,22 +292,31 @@ func (w *MeshWebhook) Handle(ctx context.Context, req admission.Request) admissi
...
 			if w.AuthMethod != "" {
 				if svc != "" && pod.Spec.ServiceAccountName != svc {
+					secretName := ""
 					sa, err := w.Clientset.CoreV1().ServiceAccounts(req.Namespace).Get(ctx, svc, metav1.GetOptions{})
 					if err != nil {
 						w.Log.Error(err, "couldn't get service accounts")
 						return admission.Errored(http.StatusInternalServerError, err)
 					}
 					if len(sa.Secrets) == 0 {
+						// check to see if there is a secret with the same name as the serviceaccount for Kube1.21+
```
kschoche, line 302:
```suggestion
						// check to see if there is a secret with the same name as the serviceaccount for Kube-1.24+
```

control-plane/connect-inject/mesh_webhook.go
```diff
@@ -292,22 +292,31 @@ func (w *MeshWebhook) Handle(ctx context.Context, req admission.Request) admissi
...
 			if w.AuthMethod != "" {
 				if svc != "" && pod.Spec.ServiceAccountName != svc {
+					secretName := ""
 					sa, err := w.Clientset.CoreV1().ServiceAccounts(req.Namespace).Get(ctx, svc, metav1.GetOptions{})
 					if err != nil {
 						w.Log.Error(err, "couldn't get service accounts")
 						return admission.Errored(http.StatusInternalServerError, err)
 					}
 					if len(sa.Secrets) == 0 {
+						// check to see if there is a secret with the same name as the serviceaccount for Kube-1.24+
```
kschoche, line 302:
```suggestion
						// Check to see if there is a secret with the same name as the serviceaccount for Kube-1.24+.
```

control-plane/subcommand/server-acl-init/connect_inject.go
```diff
@@ -94,17 +94,28 @@ func (c *Command) createAuthMethodTmpl(authMethodName string, useNS bool) (api.A
 		return api.ACLAuthMethod{}, err
 	}
 
-	// ServiceAccounts always have a secret name. The secret
-	// contains the JWT token.
+	var saSecret *apiv1.Secret
+	var secretNames []string
+	if len(authMethodServiceAccount.Secrets) == 0 {
+		// In Kube 1.21+ there is no automatically generated long term JWT token for a ServiceAccount.
```
kschoche, line 100:
```suggestion
		// In Kube 1.24+ there is no automatically generated long term JWT token for a ServiceAccount.
```

control-plane/subcommand/server-acl-init/connect_inject.go
```diff
@@ -94,17 +94,28 @@ func (c *Command) createAuthMethodTmpl(authMethodName string, useNS bool) (api.A
...
+		// Furthermore, there is no reference to a Secret in the ServiceAccount. Instead we have deployed
+		// a Secret in Helm which references the ServiceAccount and contains a permanent JWT token.
+		secretNames = append(secretNames, c.withPrefix("auth-method"))
+	} else {
+		// ServiceAccounts always have a SecretRef in k8s < 1.21. The Secret contains the JWT token.
```
kschoche, line 105:
```suggestion
		// ServiceAccounts always have a SecretRef in Kubernetes < 1.24. The Secret contains the JWT token.
```

control-plane/subcommand/server-acl-init/connect_inject.go
```diff
@@ -94,17 +94,28 @@ func (c *Command) createAuthMethodTmpl(authMethodName string, useNS bool) (api.A
...
+		for _, secretRef := range authMethodServiceAccount.Secrets {
+			secretNames = append(secretNames, secretRef.Name)
+		}
+	}
 	// Because there could be multiple secrets attached to the service account,
 	// we need pick the first one of type "kubernetes.io/service-account-token".
-	var saSecret *apiv1.Secret
-	for _, secretRef := range authMethodServiceAccount.Secrets {
+	// This is valid regardless of whether we created the Secret or Kubernetes did automatically.
```
kschoche, line 112:
```suggestion
	// We will fetch the Secrets regardless of whether we created the Secret or Kubernetes did automatically.
```

acceptance/framework/k8s/kubectl.go
```diff
@@ -102,7 +102,9 @@ func KubectlDelete(t *testing.T, options *k8s.KubectlOptions, configPath string)
 // deletes it from the cluster by running 'kubectl delete -k'.
 // If there's an error deleting the file, fail the test.
 func KubectlDeleteK(t *testing.T, options *k8s.KubectlOptions, kustomizeDir string) {
-	_, err := RunKubectlAndGetOutputE(t, options, "delete", "--timeout=60s", "-k", kustomizeDir)
+	// Ignore not found errors because Kubernetes automatically cleans up the kube secrets that we deployed
+	// referencing the ServiceAccount when it is deleted.
```
kschoche, line 106: This is just for the multiport test because I've added the Secrets to deploy alongside the ServiceAccounts, I suppose an alternative approach would be to create the secrets through the go-client instead of using Kustomize. Thoughts?

curtbushko (commented):
Looks great!  I only have a few questions/suggestions but nothing blocking...

acceptance/tests/fixtures/bases/multiport-app/serviceaccount.yaml
```diff
@@ -7,3 +7,19 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: multiport-admin
+---
+apiVersion: v1
+kind: Secret
+metadata:
```
curtbushko, line 13: Another question. Is there no way that we could generate these secrets for the user, like say in the webhook, when it is multi-port?

kschoche: I think we could, it wouldnt be a substantial amount of work, but I think it *kinda* goes in the face of the spirit of the change that Kubernetes was making here. Secondly, and more importantly, if we did that we'd need to update the RBAC for our `clusterrole` for the controller to have `write` on `secrets` resources.
Thoughts?

curtbushko: I think it is probably not worth it right now. If we end up with more multi-port users in the future then maybe do it then. It sounds like a low priority backlog item to me.

acceptance/framework/vault/vault_cluster.go
```diff
@@ -149,6 +149,25 @@ func (v *VaultCluster) bootstrap(t *testing.T, vaultNamespace string) {
...
+	retry.Run(t, func(r *retry.R) {
+		sa, err := v.kubernetesClient.CoreV1().ServiceAccounts(namespace).Get(context.Background(), vaultServerServiceAccountName, metav1.GetOptions{})
+		require.NoError(r, err)
+		if len(sa.Secrets) == 0 {
+			_, err = v.kubernetesClient.CoreV1().Secrets(namespace).Create(context.Background(), &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        vaultServerServiceAccountName,
+					Annotations: map[string]string{"kubernetes.io/service-account.name": vaultServerServiceAccountName},
+				},
+				Type: "kubernetes.io/service-account-token",
```
thisisnotashwin, line 165: There is a constant for the Type that we could use across all the secrets to make them look nicer!!
```suggestion
				Type: corev1.SecretTypeServiceAccountToken,
```

control-plane/connect-inject/mesh_webhook.go
```diff
@@ -292,22 +292,31 @@ func (w *MeshWebhook) Handle(ctx context.Context, req admission.Request) admissi
...
 						w.Log.Error(err, "couldn't get service accounts")
 						return admission.Errored(http.StatusInternalServerError, err)
 					}
 					if len(sa.Secrets) == 0 {
+						// Check to see if there is a secret with the same name as the serviceaccount for Kube-1.24+.
 						w.Log.Info(fmt.Sprintf("service account %s has zero secrets exp at least 1", svc))
-						return admission.Errored(http.StatusInternalServerError, fmt.Errorf("service account %s has zero secrets, expected at least one", svc))
+						sec, err := w.Clientset.CoreV1().Secrets(req.Namespace).Get(ctx, svc, metav1.GetOptions{})
+						if err != nil {
+							w.Log.Error(err, "couldn't get secret associated with service accounts")
```
thisisnotashwin, line 306:
```suggestion
							w.Log.Error(err, "couldn't get Secret associated with Service Account")
```

thisisnotashwin (approved):
This is great!! left a few minor suggestions!! Thanks for taking care of this bud!

thisisnotashwin:
Is the acceptance test failure related?

curtbushko (approved):
Fantastic!

Bernix01:
go go go

vaibhavgulati:
So now kubernetes 1.21+ supports consul client?

Bernix01:
You have to wait for next release but it sounds that it will! Thanks to the efforts given here and other related prs!

PR #1437, opened 2022-08-19
fix: missing MaxInboundConnections in service-defaults CRD

### LINKED ISSUE #1400: crd-servicedefaults misses MaxInboundConnections
#### Feature Description

Consul service-defaults config entry supports setting the max number of [inbound connection](https://github.com/hashicorp/consul/blob/main/.changelog/13143.txt), which limits the number of downstream connection to the proxy service. However, this field is missing in the service defaults CRD.

#### Use Case(s)

User needs to limit the inbound connection to a service by setting the MaxInboundConnections in the proxy.

huikang:
Changes proposed in this PR:
- Add `maxInboundConnections` to service-defaults CRD
- fix #1400 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


control-plane/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/go-logr/logr v0.4.0
 	github.com/google/go-cmp v0.5.7
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
-	github.com/hashicorp/consul/api v1.14.0
+	github.com/hashicorp/consul/api v1.10.1-0.20220822180451-60c82757ea35
```
david-yu, line 9: I think this should point to the 1.14 API package?

huikang: @david-yu , thanks for the feedback. This PR depends on the change in consul after [api/v1.14.0](https://github.com/hashicorp/consul/releases/tag/api%2Fv1.14.0). So shall I create a new tag for consul/api?

control-plane/api/v1alpha1/servicedefaults_types.go
```diff
@@ -244,17 +247,22 @@ func (in *ServiceDefaults) SyncedConditionStatus() corev1.ConditionStatus {
 
 // ToConsul converts the entry into it's Consul equivalent struct.
 func (in *ServiceDefaults) ToConsul(datacenter string) capi.ConfigEntry {
+	maxInboundConnections := 0
+	if in.Spec.MaxInboundConnections != nil {
+		maxInboundConnections = *in.Spec.MaxInboundConnections
+	}
```
thisisnotashwin, line 253: If Consul can accept an `int`, and it does default to `0` which implies unlimited, I think we can set the type as just `int` instead of `*int`. Kube values that tend to use `*int` is because they do care that 0 isn't set accidently. This should simplify the logic here as it can be a direct assignment!!

control-plane/api/v1alpha1/servicedefaults_types.go
```diff
@@ -280,6 +288,11 @@ func (in *ServiceDefaults) Validate(consulMeta common.ConsulMeta) error {
 	if err := in.Spec.Destination.validate(path.Child("destination")); err != nil {
 		allErrs = append(allErrs, err...)
 	}
+
+	if in.Spec.MaxInboundConnections != nil && *in.Spec.MaxInboundConnections < 0 {
+		allErrs = append(allErrs, field.Invalid(path.Child("maxinboundconnections"), in.Spec.MaxInboundConnections, "MaxInboundConnections must > 0"))
+	}
```
thisisnotashwin, line 294: This can just check if `in.Spec.MaxInboundConnections < 0` if we use `int`.

control-plane/go.mod
thisisnotashwin: I think it is ok not to create a tag! We do use the above format to include unreleased API changes. It is strange only because golang interprets it as some `1.10.1-*` version even though it contains all the 1.14 changes but that has been true for sometime now.

thisisnotashwin (commented):
This is excellent work!! thank you so much for doing this!! have a suggestion to change the types of the field a little but this is super helpful!

thisisnotashwin:
@huikang can you check if you have signed into circle CI? Im surprised at the lack of checks being run against this PR 🤔

david-yu:
@thisisnotashwin I believe this is because he is running on his own fork as opposed to a branch on this repo. Should the change be made directly as branch to this repo? If so, we may need to change our CONTRIBUTING guide.

thisisnotashwin:
> @thisisnotashwin I believe this is because he is running on his own fork as opposed to a branch on this repo. Should the change be made directly as branch to this repo? If so, we may need to change our CONTRIBUTING guide.

i think you're right! we can merge this! it is safe.

thisisnotashwin (approved):
TYSM!!

t-eckert (approved):
Looks great!

PR #1442, opened 2022-08-22
Fix Listener Filter bugs and cover more Envoy Listener Filter Types

t-eckert:
Changes proposed in this PR:
- Extend the timeout for port-forward (I was seeing some unneeded failures when connecting to AWS).
- Expand the set of [Envoy Listener Filters](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto) that may be parsed and output to the Listeners table.
- Fix [Proxy read: terminating gateway listeners show up weird](https://hashicorp.atlassian.net/browse/NET-707?atlOrigin=eyJpIjoiNGM4OGMzYzhhZDgxNGEzNTlkZDVhNzdjMzllZWYwODAiLCJwIjoiaiJ9)
- Fix [Proxy read: blank filters being displayed](https://hashicorp.atlassian.net/browse/NET-709?atlOrigin=eyJpIjoiMWZiZGY2NTM2ZmZkNDkzNzljN2U0MzZjOTVlNTBiZDciLCJwIjoiaiJ9)
- Change `to` back to `->` in filter formatting output.
- Unescape `>` in JSON output so it doesn't show up weird.

How I've tested this PR:
- Added at least one unit test for each of the filter extensions.
- Tested the code against a terminating gateway to ensure it behaves properly.

- Unit tests
- Testing against a terminating gateway if you like!

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,10 +37,14 @@ IMPROVEMENTS:
   * Display clusters by their short names rather than FQDNs for the `proxy read` command. [[GH-1412](https://github.com/hashicorp/consul-k8s/pull/1412)]
   * Display a message when `proxy list` returns no results. [[GH-1412](https://github.com/hashicorp/consul-k8s/pull/1412)]
   * Display a warning when a user passes a field and table filter combination to `proxy read` where the given field is not present in any of the output tables. [[GH-1412](https://github.com/hashicorp/consul-k8s/pull/1412)]
+  * Extend the timeout for `consul-k8s proxy read` to establish a connection from 5s to 10s.
+  * Expand the set of Envoy Listener Filters that may be parsed and output to the Listeners table.
 
 BUG FIXES:
 * Helm
   * API Gateway: Configure ACL auth for controller correctly when deployed in secondary datacenter with federation enabled [[GH-1462](https://github.com/hashicorp/consul-k8s/pull/1462)]
+* CLI
+  * Fix issue where SNI filters for Terminating Gateways showed up as blank lines.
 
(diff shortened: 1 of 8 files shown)


cli/cmd/proxy/read/config.go
```diff
@@ -396,45 +396,154 @@ func parseSecrets(rawCfg map[string]interface{}) ([]Secret, error) {
 	return secrets, nil
 }
 
-func formatFilters(filterChain filterChain) (filters []string) {
+func formatFilters(filters []filter) (formatted []string) {
```
thisisnotashwin, line 399: We avoid using named return parameters as it could be confusing. In this case, given we are explicitly returning `formatted` there is no need to keep this.
```suggestion
func formatFilters(filters []filter) []string {
```

thisisnotashwin (approved):
Nicely done!

CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ IMPROVEMENTS:
   * Display a message when `proxy list` returns no results. [[GH-1412](https://github.com/hashicorp/consul-k8s/pull/1412)]
   * Display a warning when a user passes a field and table filter combination to `proxy read` where the given field is not present in any of the output tables. [[GH-1412](https://github.com/hashicorp/consul-k8s/pull/1412)]
 
+IMPROVEMENTS:
```
jmurret, line 12: Can you merge this into the above section?

jmurret (approved):
Looks good.  Had one request to merge some redundant sections of the readme.

PR #1445, opened 2022-08-24
CNI Accptance test on Kind

curtbushko:
Background: In order to run acceptance tests on kind there needs to be some extra setup. Kind needs the Calico CNI plugin to be installed first and Kind needs its network to match what Calico expects.

Some tests have been disabled as running them with CNI is redundant to other acceptance tests.

There is no magic here. Helm install the CNI plugin and the tests are run. Everything should work the same as normal.

Changes proposed in this PR:
- Added a new make target for installing Calico in Kind
- Added a helper make target for setting up local Kind with Calico just in case anyone wants to run it.
- Added a kind.config for setting up the Kind cluster
- Added a new -enable-cni flag to the acceptance test config so that it can be passed through from CircleCI
- Added a new nightly circleci job for running the CNI kind tests
- I had missed a bats test for the connect-inject template that I merged in a previous PR

Note to Iryna, I added connect-inject-init user uid to the redirect PR last night so not to muddle this PR.

How I've tested this PR:

- "nightly" circle CI and local runs

How I expect reviewers to test this PR:

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #1446, opened 2022-08-25
CNI Acceptance test on GKE

curtbushko:
This PR builds on top of the [CNI acceptance tests on Kind](https://github.com/hashicorp/consul-k8s/pull/1445) PR to get things running on GKE.

GKE is the only cloud platform that has a different requirement for `connectInject.cni.cniBinDir` and does not use the default. The consul-cni binary needs to be installed to `/home/kubernetes/bin`.

Changes proposed in this PR:
- Added a `-use-gke` flag to the acceptance test command so that the flag can be passed through to the tests
- If both, `-enable-cni` and `-use-gke` are passed through then the cniBinDir gets set for helm values.
- GKE would not install the CNI plugin unless the `use` security policy was set.
- GKE was throwing an error saying that `allowPrivilegeEscalation: true` needed to be set if `privileged: true`. Once set, tests passed.

How I've tested this PR:

- ran the acceptance tests in pipelines

:eyes:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -922,7 +922,7 @@ jobs:
           failure_message: "Acceptance tests against Kind with Kubernetes v1.23 failed. Check the logs at: ${CIRCLE_BUILD_URL}"
 
   acceptance-kind-cni-1-23:
-    parallelism: 6 
+    parallelism: 6
     environment:
       - TEST_RESULTS: /tmp/test-results
     machine:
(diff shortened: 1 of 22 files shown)


.circleci/config.yml
```diff
@@ -1089,6 +1158,11 @@ workflows:
       #     requires:
       #       - cleanup-gcp-resources
```
curtbushko, line 1159: Will clean this all up before merging

thisisnotashwin (approved):
:chef-kiss:

.circleci/config.yml
```diff
@@ -684,6 +684,75 @@ jobs:
...
+
+      # Restore go module cache if there is one
+      - restore_cache:
+          keys:
+            - consul-helm-acceptance-modcache-v2-{{ checksum "acceptance/go.mod" }}
+
+      - run: mkdir -p $TEST_RESULTS
+
+      - run-acceptance-tests:
+          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig" -enable-pod-security-policies -enable-transparent-proxy -enable-cni -enable-gke
```
ishustava, line 736: nit: could this flag be called `-use-gke` similar to `-use-kind`?

ishustava (approved):
Looks good! Just one tiny comment

.circleci/config.yml
curtbushko: Nit addressed. 😺

PR #1453, opened 2022-08-25
Allow providing custom http port and using system roots to the connect-init container

ishustava:
Previously, if you have configured `externalServers.useSystemRoots` or provided a different https port for external servers, we would ignore because we were hard-coding http port and were determining if TLS is enabled only if CA cert is set. That made sense when we were talking to clients, but now that we're talking to servers we need to account for those values so that the init container is correctly configured to talk to the servers

How I've tested this PR:
- acceptance tests
- manual tests with HCP 

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


thisisnotashwin (approved):
Excellence! I really like the cleanup to how the address is parsed!!

t-eckert (approved):
Nice!

PR #1454, opened 2022-08-26
CNI CRT Builds

curtbushko:
This PR is an attempt to get the CNI added to the release process using CRT.

Changes proposed in this PR:
- Use the current build matrix for control-plane to also build the CNI plugin
- The build adds the CNI plugin to /dist and it is zipped up so that it can be used in another build step (docker)
- The docker build takes the zip file, extracts it and copies it into the docker image
- Note, at this time, the above way is the only way to build the binaries and docker images because the docker build action does not support multiple binaries. The Release Team would need to change the docker build action to handle multiple binaries.

All of the above works, except...

The security scanner is looking at the zip file and expects 1 binary per zip and thus fails the security scan. I have contacted the Release Team about this to see if there is a possible workaround. 

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


curtbushko:
Closed in favour of a cleaner https://github.com/hashicorp/consul-k8s/pull/1458

PR #1456, opened 2022-08-26
CNI Plugin for Consul-k8s

curtbushko:
This PR adds a [CNI](https://github.com/containernetworking/cni) networking plugin for Consul on Kubernetes. The Consul Cni plugin is an improvement on [Transparent Proxy](https://www.consul.io/docs/connect/transparent-proxy) that added the ability to transparently intercept and redirect traffic using sidecar proxies which allowed applications to be deployed without modification. 

The major benefit of the Consul CNI plugin is that the connect-inject init container no longer requires CAP_NET_ADMIN privileges in order to generate iptables rules that redirect TCP traffic to the sidecar proxy. Removing CAP_NET_ADMIN permissions increases Kubernetes security and lowers friction to Consul Service Mesh adoption. 

Changes proposed in this PR:

- Added a CNI installer daemonset to the helm install. On every node, the installer daemonset:
   - Appends the consul-cni configuration to the main CNI configuration as the final plugin in the CNI chain
   - Installs the consul-cni binary to the node
   - Creates a kubeconfig file so that the the consul-cni plugin can communicate with pods.
   - Repairs the consul-cni configuration if another CNI plugin corrupts the configuration

- Added a consul-cni CNI plugin that is executed by the kubelet when a pod is created.
  - The consul-cni plugin is a chained plugin and must be installed alongside another CNI plugin (eg. calico)
  - In order to install and use the plugin you must ensure that connect inject (connectInject.enabled = true) and CNI (connectInject.cni.enabled = true) are enabled in the helm chart.
  - Note: if using consul-cni on GKE, connectInject.cni.cniBinDir must be set to /home/kubernetes/bin
  
- Unit, acceptance, and bats (helm) tests to cover all of the above changes/additions.

How I've tested this PR:

- Unit, acceptance tests and manual testing.

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


ishustava (approved):

thisisnotashwin (approved):
🥳

PR #1469, opened 2022-08-31
Register Terminating Gateways with Endpoints Controller

t-eckert:
Changes proposed in this PR:
- Added a `terminating-gateway-service.yaml` template. In order to register the deployment with the Endpoints Controller, a "dummy" service must exist so that a Kubernetes endpoint object is created, triggering the Endpoints Controller reconcile loop.
- Modified the `terminating-gateway-deployment.yaml` template:
  - Added a "managed by" label.
  - Added annotations to configure the behavior of the Endpoints Controller during registration.
  - Removed the `copy-consul-bin` init container.
  - Changed the `terminating-gateway-init` init container to use `connect-init` instead of `acl-init` as the client `acl-init` talked to will no longer exist in agentless.
  - Removed the `preStop` hook from the `terminating-gateway` container.
  - Removed the `consul-sidecar` container in preparation for the blessings of the Consul Dataplane. 
- Removed BATS tests which covered features no longer present in agentless.
- Added BATS tests for `terminating-gateway-service` and for Consul namespace handling in `terminating-gateway-deployment`.
- Added annotation `consul.hashicorp.com/terminating-gateway-namespace` to configure which Consul namespace a terminating gateway will be placed in when using Consul Enterprise.
- Modified the Endpoints Controller to register terminating gateways
  - Added a `switch` to the `createGatewayRegistrations` method which handles configuration of the endpoint particular to the terminating gateway.
  - Added unit tests which verify the registrations created by `createGatewayRegistrations` is correct.
  - Added a unit test to verify the the `Reconcile` method does register the endpoint.

How I've tested this PR:
- Added tests to the Endpoints Controller
- Added (mostly removed) BATS tests for the Terminating Gateway Deployment template
- Added BATS tests for the Terminating Gateway Service

- Running unit tests.
- The new terminating gateway configuration can be deployed and will be registered, but until integration with the Consul Dataplane is complete in future work, it will not work as intended.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/terminating-gateways-service.yaml
```diff
@@ -0,0 +1,32 @@
...
+    release: {{ $root.Release.Name }}
+    component: terminating-gateways
+spec:
+  selector:
+    app: {{ template "consul.name" $root }}
+    release: "{{ $root.Release.Name }}"
+    component: terminating-gateways
+  ports:
+    - name: gateway
+      port: 80 # TODO what should this be set to?
```
t-eckert, line 27: Not sure what to set this to.

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -202,50 +203,21 @@ spec:
...
+                consul-k8s-control-plane connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
+                  -consul-api-timeout={{ .Values.global.consulAPITimeout }} \
+                  -gateway -gateway-name="terminating-gateway" \
+                  -consul-node-name=${POD_NAMESPACE} \
+                  {{- if and .Values.global.federation.enabled .Values.global.federation.primaryDatacenter }}
+                  -acl-auth-method={{ template "consul.fullname" . }}-k8s-component-auth-method-{{ .Values.global.datacenter }} \
+                  -primary-datacenter={{ .Values.global.federation.primaryDatacenter }} \
+                  {{- else }}
+                  -acl-auth-method={{ template "consul.fullname" . }}-k8s-component-auth-method \
                   {{- end }}
```
t-eckert, line 215: Don't need global tokens

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -202,50 +203,21 @@ spec:
...
             - |
                 {{- if $root.Values.global.acls.manageSystemACLs }}
-                consul-k8s-control-plane acl-init \
-                  -component-name=terminating-gateway/{{ template "consul.fullname" $root }}-{{ .name }} \
-                  -acl-auth-method={{ template "consul.fullname" $root }}-k8s-component-auth-method \
-                  {{- if $root.Values.global.adminPartitions.enabled }}
-                  -partition={{ $root.Values.global.adminPartitions.name }} \
+                consul-k8s-control-plane connect-init -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
+                  -consul-api-timeout={{ .Values.global.consulAPITimeout }} \
+                  -gateway -gateway-name="terminating-gateway" \
```
t-eckert, line 208: gateway name should be service name

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -202,50 +203,21 @@ spec:
...
+                  -consul-node-name=${POD_NAMESPACE} \
```
t-eckert, line 209: Node name should not be namespace.

charts/consul/templates/terminating-gateways-service.yaml
t-eckert: Omit ports, we only need the service to create an endpoint.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -653,6 +665,96 @@ func (r *EndpointsController) createGatewayRegistrations(pod corev1.Pod, service
...
+	}
+	for k, v := range pod.Annotations {
+		if strings.HasPrefix(k, annotationMeta) && strings.TrimPrefix(k, annotationMeta) != "" {
+			if v == "$POD_NAME" {
+				meta[strings.TrimPrefix(k, annotationMeta)] = pod.Name
+			} else {
+				meta[strings.TrimPrefix(k, annotationMeta)] = v
+			}
+		}
+	}
```
t-eckert, line 686: Cut

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -653,6 +665,96 @@ func (r *EndpointsController) createGatewayRegistrations(pod corev1.Pod, service
...
+
+	consulNS := r.consulNamespace(pod.Namespace)
+	service := &api.AgentService{
+		Kind:      api.ServiceKindTerminatingGateway,
+		ID:        svcID,
+		Service:   "terminating-gateway",
+		Port:      8443,
+		Address:   pod.Status.PodIP,
+		Meta:      meta,
+		Namespace: consulNS,
```
t-eckert, line 726: This is not right.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -653,6 +665,96 @@ func (r *EndpointsController) createGatewayRegistrations(pod corev1.Pod, service
...
+			   namespace = "{{ (default $defaults.consulNamespace .consulNamespace) }}"
+			   {{- end }}
+			   {{- if $root.Values.global.adminPartitions.enabled }}
+			   partition = "{{ $root.Values.global.adminPartitions.name }}"
+			   {{- end }}
+		✅	   address = "${POD_IP}"
+		✅	   port = 8443
+			   {{- if (and $root.Values.global.metrics.enabled $root.Values.global.metrics.enableGatewayMetrics) }}
+			   proxy { config { envoy_prometheus_bind_addr = "${POD_IP}:20200" } }
+			   {{- end }}
```
t-eckert, line 705: Need to thread through enableGatewayMetrics.
Get that value either at install or from an annotation.

t-eckert: Get the enable gateway metrics into the metrics struct.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -653,6 +665,96 @@ func (r *EndpointsController) createGatewayRegistrations(pod corev1.Pod, service
...
+		Checks: api.HealthChecks{
+			&api.HealthCheck{
+				Name: "Terminating Gateway Listening",
+				Definition: api.HealthCheckDefinition{
+					TCP:                                    fmt.Sprintf("%s:8443", pod.Status.PodIP),
+					IntervalDuration:                       time.Duration(10 * time.Second),
+					DeregisterCriticalServiceAfterDuration: time.Duration(6 * time.Hour),
+				},
+			},
+		},
```
t-eckert, line 751: Remove this for k8s health checks.

charts/consul/templates/terminating-gateways-deployment.yaml
t-eckert: From RFC, Node name is the Consul Synthetic Node which in this case is "k8s-service-mesh".

control-plane/connect-inject/endpoints_controller.go
t-eckert: Should this just be "k8s-service-mesh"?

charts/consul/templates/terminating-gateways-service.yaml
```diff
@@ -0,0 +1,31 @@
...
+    component: terminating-gateways
+spec:
+  selector:
+    app: {{ template "consul.name" $root }}
+    release: "{{ $root.Release.Name }}"
+    component: terminating-gateways
+  type: ClusterIP
+  ports:
+    - port: 80
+      targetPort: 8443
```
t-eckert, line 28: This port will not get used under normal operation. It is needed however in order to meet the spec for a service. Otherwise Helm/K8s will throw a fit.

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -202,50 +186,17 @@ spec:
...
-                  {{- if $root.Values.global.adminPartitions.enabled }}
-                  -partition={{ $root.Values.global.adminPartitions.name }} \
-                  {{- end }}
-                  -token-sink-file=/consul/service/acl-token \
+                consul-k8s-control-plane connect-init \
+                  -pod-name=${POD_NAME} -pod-namespace=${POD_NAMESPACE} \
                   -consul-api-timeout={{ $root.Values.global.consulAPITimeout }} \
+                  -gateway -gateway-kind="terminating-gateway" \
+                  -consul-node-name="k8s-service-mesh" \
+                  -acl-auth-method={{ template "consul.fullname" $root }}-k8s-component-auth-method \
```
ishustava, line 194: This should only be passed if ACLs are enabled

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -202,50 +186,17 @@ spec:
...
+                  -service-name={{ .name }} \
+                  -acl-token-sink=/consul/service/acl-token \
```
ishustava, line 196: Same here: only needed if ACLs are enabled

charts/consul/test/unit/terminating-gateways-deployment.bats
```diff
@@ -1175,161 +865,18 @@ key2: value2' \
   [ "${actual}" = "value2" ]
 }
 
-#--------------------------------------------------------------------
-# terminating-gateway-init init container command
-
-@test "terminatingGateways/Deployment: terminating-gateway-init init container defaults" {
```
ishustava, line 1181: These should not be deleted because the init container still exists but there is now different logic.

charts/consul/templates/terminating-gateways-service.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if and .Values.terminatingGateways.enabled }}
```
ishustava, line 1: probably don't need this `and` or did you mean to `and` it with something?

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -609,36 +604,61 @@ func (r *EndpointsController) createGatewayRegistrations(pod corev1.Pod, service
...
+	case "terminating":
+		service.Kind = api.ServiceKindTerminatingGateway
+		service.Service = "terminating-gateway"
+		service.Port = 8443
+
+		if ns, ok := pod.Annotations[annotationTerminatingGatewayNamespace]; ok {
+			consulNS = ns
+		}
 	}
+	service.Namespace = consulNS
```
ishustava, line 652: nit: I think this would be a bit simpler if you set that in the terminating gw case because then you don't need the extra variable.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -5035,6 +5099,99 @@ func TestMapAddresses(t *testing.T) {
...
+			if c.metricsConfig.DefaultEnableMetrics && c.metricsConfig.EnableGatewayMetrics {
+				require.Equal(t, "1.2.3.4:20200", registration.Service.Proxy.Config["envoy_prometheus_bind_addr"])
+			}
+		})
+	}
+}
+
 func createPod(name, ip string, inject bool, managedByEndpointsController bool) *corev1.Pod {
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
```
ishustava, line 5197: Could we also test the deregistration code path and check that the ACL token gets deleted?

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -1630,6 +1629,71 @@ func TestReconcileCreateEndpoint(t *testing.T) {
...
+						MetaKeyPodName:         "terminating-gateway",
+						MetaKeyKubeServiceName: "service-created",
+						MetaKeyKubeNS:          "default",
+						MetaKeyManagedBy:       managedByValue,
+					},
+					ServiceTags: []string{},
+				},
+			},
+			expectedProxySvcInstances: []*api.CatalogService{},
+			expectedHealthChecks:      []*api.HealthCheck{},
```
ishustava, line 1695: Could we add expected health checks to check that we register them correctly?

ishustava: Also need an enterprise test to check that we register gateway in the correct consul namespace

ishustava (commented):
Good work so far! I'm leaving my comments in-line

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -66,7 +65,13 @@ spec:
         release: {{ $root.Release.Name }}
         component: terminating-gateway
         terminating-gateway-name: {{ template "consul.fullname" $root }}-{{ .name }}
+        consul.hashicorp.com/connect-inject-managed-by: consul-k8s-endpoints-controller
       annotations:
+        "consul.hashicorp.com/connect-inject": "false"
```
ishustava, line 70: btw don't know if you saw it, but this is a duplicate annotation. It is already there after all the vault stuff

t-eckert: Oh geez, good catch. I didn't see that. Thank you!

charts/consul/test/unit/terminating-gateways-deployment.bats
t-eckert: Gotcha, I replicated all of the coverage in with new tests that are a little different. The gateway namespace ones are moved to a section called `consul namespaces` which starts on line 965 because they now look for annotations rather than having the value passed in to connect-init.

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -324,17 +269,7 @@ spec:
...
-            - envoy
-            - -gateway=terminating
-            - -proxy-id=$(POD_NAME)
-            {{- if $root.Values.global.enableConsulNamespaces }}
-            - -namespace={{ default $defaults.consulNamespace .consulNamespace }}
-            {{- end }}
-            {{- if $root.Values.global.adminPartitions.enabled }}
-            - -partition={{ $root.Values.global.adminPartitions.name }}
-            {{- end }}
+            - sleep infinity
```
t-eckert, line 272: This will be replaced with the startup for consul dataplane in a future PR into this branch.

control-plane/connect-inject/endpoints_controller_ent_test.go
```diff
@@ -333,12 +375,8 @@ func TestReconcileCreateEndpointWithNamespaces(t *testing.T) {
...
 			for _, expectedCheck := range setup.expectedHealthChecks {
 				var checks api.HealthChecks
-				filter := fmt.Sprintf("CheckID == `%s`", expectedCheck.CheckID)
-				checks, _, err := consulClient.Health().Checks(expectedCheck.ServiceName, &api.QueryOptions{Filter: filter})
-				if expectedCheck.ServiceName == "mesh-gateway" {
-					checks, _, err = consulClient.Health().Checks("mesh-gateway", &api.QueryOptions{Namespace: "default"})
-					require.NoError(t, err)
-				}
+				filter := fmt.Sprintf("ServiceID == `%s`", expectedCheck.ServiceID)
+				checks, _, err = consulClientDefaultNS.Health().Checks(expectedCheck.ServiceName, &api.QueryOptions{Filter: filter, Namespace: expectedCheck.Namespace})
```
thisisnotashwin, line 375: NICE!

control-plane/subcommand/connect-init/command.go
```diff
@@ -77,7 +77,7 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagProxyIDFile, "proxy-id-file", defaultProxyIDFile, "File name where proxy's Consul service ID should be saved.")
 	c.flagSet.BoolVar(&c.flagMultiPort, "multiport", false, "If the pod is a multi port pod.")
 	c.flagSet.BoolVar(&c.flagGateway, "gateway", false, "If the pod is a Consul gateway pod.")
-	c.flagSet.StringVar(&c.flagGatewayKind, "gateway-kind", "", "Name of the gateway that is being registered.")
+	c.flagSet.StringVar(&c.flagGatewayKind, "gateway-kind", "", "Kind of gateway that is being registered: ingress, terminating, or mesh.")
```
thisisnotashwin, line 80: 🙏

control-plane/connect-inject/endpoints_controller_ent_test.go
```diff
@@ -299,12 +332,17 @@ func TestReconcileCreateEndpointWithNamespaces(t *testing.T) {
...
 			require.NoError(t, err)
-			service, _, err := consulClient.Catalog().Service("mesh-gateway", "", &api.QueryOptions{Namespace: "default"})
+			meshGatewayService, _, err := consulClient.Catalog().Service("mesh-gateway", "", &api.QueryOptions{Namespace: "default"})
+			require.NoError(t, err)
+			serviceInstances = append(serviceInstances, meshGatewayService...)
+			terminatingGatewayService, _, err := consulClient.Catalog().Service("terminating-gateway", "", &api.QueryOptions{Namespace: testCase.ExpConsulNS})
 			require.NoError(t, err)
-			serviceInstances = append(serviceInstances, service...)
+			serviceInstances = append(serviceInstances, terminatingGatewayService...)
+
```
thisisnotashwin, line 345: I wonder if there is a way this could be a touch more elegant. I don't like that each gateway has to be explicit :\

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -5685,6 +5864,98 @@ func Test_GetWANData(t *testing.T) {
 	}
 }
 
+func TestCreateTerminatingGatewayRegistrations(t *testing.T) {
+	cases := map[string]struct {
+		podName          string
+		serviceName      string
+		serviceNamespace string
```
thisisnotashwin, line 5871: It looks like these fields are constant across every test. id just move them into the fixture.

thisisnotashwin (commented):
Have some more reviewing to do but this looks pretty good so far. here are some thoughts so far!

control-plane/connect-inject/endpoints_controller_test.go
t-eckert: Cool, I originally parameterized this with the thought that you would have similar tests for mesh gateways that I would combine with this set.

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -66,7 +65,13 @@ spec:
         release: {{ $root.Release.Name }}
         component: terminating-gateway
         terminating-gateway-name: {{ template "consul.fullname" $root }}-{{ .name }}
+        consul.hashicorp.com/connect-inject-managed-by: consul-k8s-endpoints-controller
       annotations:
+        "consul.hashicorp.com/connect-inject": "false"
+        "consul.hashicorp.com/gateway-kind": "terminating"
+        {{- if $root.Values.global.enableConsulNamespaces }}
+        "consul.hashicorp.com/gateway-namespace": {{ (default $defaults.consulNamespace .consulNamespace) }}
```
t-eckert, line 73: Originally, I called this `terminating-gateway-namespace`. But I've changed it to `gateway-namespace` so that it may also be used for Ingress gateways when we get to them. My goal here was to reduce the overall number of different annotations by... one. :shrug:

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -548,6 +591,148 @@ func (r *EndpointsController) createServiceRegistrations(pod corev1.Pod, service
...
+	meta := map[string]string{
+		MetaKeyPodName:         pod.Name,
+		MetaKeyKubeServiceName: serviceEndpoints.Name,
+		MetaKeyKubeNS:          serviceEndpoints.Namespace,
+		MetaKeyManagedBy:       managedByValue,
+	}
+
+	if r.EnableWANFederation {
+		meta[MetaKeyConsulWANFederation] = "1"
+	}
```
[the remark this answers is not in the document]
t-eckert: No, I tucked it away in the mesh only code now.

control-plane/connect-inject/endpoints_controller_ent_test.go
t-eckert: The rebase mayhem unmoored this comment, but I did find a more scalable and elegant way to do this.

charts/consul/test/unit/terminating-gateways-deployment.bats
```diff
@@ -761,135 +686,16 @@ load _helpers
...
-
-  local actual=$(echo $object | yq -r '.requests.cpu' | tee /dev/stderr)
-  [ "${actual}" = "gwcpu" ]
-
-  local actual=$(echo $object | yq -r '.limits.memory' | tee /dev/stderr)
-  [ "${actual}" = "gwmemory2" ]
-
-  local actual=$(echo $object | yq -r '.limits.cpu' | tee /dev/stderr)
-  [ "${actual}" = "gwcpu2" ]
-}
```
thisisnotashwin, line 831: These tests should still be present.

thisisnotashwin (approved):
This looks great! Im approving it because I don't think I need to review this again, but we should reintroduce the bats tests for init container resources before we merge this in.

thisisnotashwin:
Looks like there are some test failures.

charts/consul/test/unit/terminating-gateways-deployment.bats
thisisnotashwin: Actually, these can stay deleted. My B.

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -639,15 +658,28 @@ func (r *EndpointsController) createGatewayRegistrations(pod corev1.Pod, service
 				Address: wanAddr,
 				Port:    wanPort,
 			},
-		},
+		}
+	case TerminatingGateway:
+		service.Kind = api.ServiceKindTerminatingGateway
+		service.Service = TerminatingGateway
```
thisisnotashwin, line 664: Sorry I just spotted this. This has to be the name of the service that maps to this endpoint. It will need to be different for each terminating gateway.

control-plane/connect-inject/endpoints_controller_ent_test.go
```diff
@@ -1534,6 +1673,140 @@ func TestReconcileDeleteEndpointWithNamespaces(t *testing.T) {
...
+			ConsulNS: "default",
+		},
+		"other Consul namespace": {
+			ConsulNS: "other",
+		},
+	}
+	for name, ts := range cases {
+		cases := []struct {
+			name              string
+			consulSvcName     string
```
ishustava, line 1806: It looks like `consulSvcName` is always the same between test cases so we can remove it.

control-plane/connect-inject/endpoints_controller_ent_test.go
```diff
@@ -1639,61 +1954,56 @@ func TestReconcileDeleteEndpointWithNamespaces(t *testing.T) {
 					// Create a token for it if ACLs are enabled.
 					if tt.enableACLs {
```
ishustava, line 1955: What do you think about simplifying it the inside of this `if` condition like so:

```
var writeOpts api.WriteOptions
if svc.Kind == api.ServiceKindMeshGateway {
	writeOpts.Namespace = "default"
} else {
	writeOpts.Namespace = ts.ConsulNS
}
writeOpts.Namespace = "default" // Mesh Gateways must always be registered in the "default" namespace.
test.SetupK8sAuthMethod(t, consulClient, svc.Service, svc.Meta[MetaKeyKubeNS])
token, _, err = consulClient.ACL().Login(&api.ACLLoginParams{
	AuthMethod:  test.AuthMethod,
	BearerToken: test.ServiceAccountJWTToken,
	Meta: map[string]string{
		TokenMetaPodNameKey: fmt.Sprintf("%s/%s", svc.Meta[MetaKeyKubeNS], svc.Meta[MetaKeyPodName]),
		"component":         svc.ID,
	},
}, &writeOpts)

require.NoError(t, err)
```

It looks like the namespace in `WriteOpts` is the only thing different between the cases.

control-plane/connect-inject/endpoints_controller_ent_test.go
```diff
@@ -1712,10 +2022,10 @@ func TestReconcileDeleteEndpointWithNamespaces(t *testing.T) {
...
 					require.Empty(t, gatewayInstances)
 				} else {
-					serviceInstances, _, err := consulClient.Catalog().Service(tt.consulSvcName, "", &api.QueryOptions{Namespace: ts.ExpConsulNS})
+					serviceInstances, _, err := consulClient.Catalog().Service(tt.consulSvcName, "", &api.QueryOptions{Namespace: ts.ConsulNS})
 					require.NoError(t, err)
 					require.Empty(t, serviceInstances)
-					proxyServiceInstances, _, err := consulClient.Catalog().Service(fmt.Sprintf("%s-sidecar-proxy", tt.consulSvcName), "", &api.QueryOptions{Namespace: ts.ExpConsulNS})
+					proxyServiceInstances, _, err := consulClient.Catalog().Service(fmt.Sprintf("%s-sidecar-proxy", tt.consulSvcName), "", &api.QueryOptions{Namespace: ts.ConsulNS})
 					require.NoError(t, err)
 					require.Empty(t, proxyServiceInstances)
```
ishustava, line 2030: Gateways don't have sidecar proxies, so these assertions can be removed.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -3428,7 +3633,7 @@ func TestReconcileDeleteEndpoint(t *testing.T) {
 							},
 						}, nil)
 						require.NoError(t, err)
-					case api.ServiceKindTypical:
+					default:
```
ishustava, line 3636: 🤔 It looks like the terminating gateway will fall into the `default` case which means it doesn't matter if the `component` meta data is set on the token. Maybe we then don't need the `switch` case since the behavior is the same regardless of whether the token is for a gateway or not.

control-plane/connect-inject/endpoints_controller_test.go
```diff
@@ -5804,6 +6009,91 @@ func Test_GetWANData(t *testing.T) {
 	}
 }
 
+func TestCreateTerminatingGatewayRegistrations(t *testing.T) {
```
ishustava, line 6012: Do we still need this test?

ishustava (commented):
It's pretty close!! I left some more comments in-line, nothing super major - they're all mostly about code cleanup and a few testing suggestions. Other than that, it looks good!

control-plane/connect-inject/endpoints_controller_test.go
t-eckert: No, it's being tested by the "wider scoped" tests.

t-eckert: Removed

charts/consul/templates/terminating-gateways-deployment.yaml
```diff
@@ -270,29 +212,23 @@ spec:
       containers:
         - name: terminating-gateway
           image: {{ $root.Values.global.imageEnvoy | quote }}
+          volumeMounts:
+            - name: consul-service
+              mountPath: /consul/service
+              readOnly: true
+            {{- if $root.Values.global.tls.enabled }}
```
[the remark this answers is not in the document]
t-eckert: Handled! Here and in the `terminating-gateway-init` spec too. I added the relevant bats test. I tried to have it test the presence/absence of `volumeMounts` but got mired in some JQ issues so I deferred that for now. If you think it's vital, maybe we can look at it together.

charts/consul/test/unit/terminating-gateways-deployment.bats
```diff
@@ -200,69 +199,13 @@ load _helpers
...
-      --set 'terminatingGateways.enabled=true' \
-      --set 'connectInject.enabled=true' \
-      --set 'global.tls.enabled=true' \
-      --set 'global.tls.enableAutoEncrypt=true' \
-      . | tee /dev/stderr |
-      yq -s '.[0].spec.template.spec.initContainers[] | select(.name == "get-auto-encrypt-client-ca") | length > 0' | tee /dev/stderr)
-  [ "${actual}" = "true" ]
-}
-
-@test "terminatingGateways/Deployment: consul-ca-cert volume is not added if externalServers.enabled=true and externalServers.useSystemRoots=true" {
```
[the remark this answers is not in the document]
t-eckert: Cool, I re-added it.

control-plane/connect-inject/endpoints_controller_ent_test.go
```diff
@@ -336,10 +290,195 @@ func TestReconcileCreateEndpointWithNamespaces(t *testing.T) {
...
+	}
+	for name, testCase := range cases {
+		setup := struct {
+			k8sObjects                 func() []runtime.Object
+			expectedConsulSvcInstances []*api.CatalogService
+			expectedProxySvcInstances  []*api.CatalogService
+			expectedHealthChecks       []*api.HealthCheck
+		}{
+			k8sObjects: func() []runtime.Object {
+				meshGateway := createGatewayWithNamespace("mesh-gateway", "default", "3.3.3.3", map[string]string{
```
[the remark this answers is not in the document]
t-eckert: I think it makes sense to leave it so that it is similar to the existing `createPodWithNamespace` function and just in case we need to modify the namespace in future tests. I'll give you veto power on that decision. Just let me know.

control-plane/connect-inject/endpoints_controller_ent_test.go
t-eckert: Great call!

control-plane/connect-inject/endpoints_controller_test.go
t-eckert: Cleaned this up

PR #1473, opened 2022-09-01
release 0.48.0

kschoche:
Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ BREAKING CHANGES:
   type: kubernetes.io/service-account-token
   ```
 
+* Control Plane
+  * Rename flag `server-address` to `token-server-address` in the `inject-connect` subcommand to avoid overloading the context of the `server-address` flag. [[GH-1426](https://github.com/hashicorp/consul-k8s/pull/1426)]
+
 IMPROVEMENTS:
 * CLI:
   * Display clusters by their short names rather than FQDNs for the `proxy read` command. [[GH-1412](https://github.com/hashicorp/consul-k8s/pull/1412)]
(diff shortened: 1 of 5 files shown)


charts/consul/Chart.yaml
```diff
@@ -15,7 +15,7 @@ annotations:
     - name: consul
       image: hashicorp/consul:1.13.1
     - name: consul-k8s-control-plane
-      image: hashicorp/consul-k8s-control-plane:0.47.1
+      image: hashicorp/consul-k8s-control-plane:0.48.0
     - name: envoy
       image: envoyproxy/envoy:v1.22.4
```
ishustava, line 20: could we update envoy image to 1.23.x? We downgraded because of this issue but looks like it was fixed in 1.23.1: https://github.com/envoyproxy/envoy/issues/22261

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## UNRELEASED
+## 0.48.0 (September 01, 2022)
 
 FEATURES:
```
ishustava, line 3: could we add to breaking changes that support for k8s 1.19 and 1.20 is dropped

ishustava: Also the breaking change for `Kubernetes 1.24 Support` says "Kubernetes 1.24+" but we don't support anything above 1.24 yet.

kschoche: Good call, thanks! How does that look now?

ndhanushkodi (approved):

CHANGELOG.md
```diff
@@ -33,6 +34,9 @@ BREAKING CHANGES:
   type: kubernetes.io/service-account-token
   ```
 
+* Control Plane
+  * Rename flag `server-address` to `token-server-address` in the `inject-connect` subcommand to avoid overloading the context of the `server-address` flag. [[GH-1426](https://github.com/hashicorp/consul-k8s/pull/1426)]
+
 IMPROVEMENTS:
```
ishustava, line 40: One last thing: could you add envoy version bump to changelog as an improvement?

ishustava (approved):

Issue #1479, opened 2022-09-05
API Gateway controller ACL init is broken in v0.48.0


manobi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

v0.48.0 uses k8s-auth when in secondary datacenters [[GH-1462](https://github.com/hashicorp/consul-k8s/pull/1462)](by @nathancoleman), but after this upgrade API Gateway controller acl-init never finishes.

As mentioned in [original issue](https://github.com/hashicorp/consul-k8s/issues/1344#issuecomment-1235661387), the consul-api-gateway-controller service account does not seems to have enough permission to perform authentication:

```
2022-09-02T15:51:21.019Z [ERROR] unable to login: error="Unexpected response code: 403 (rpc error making call: rpc error making call: rpc error making call: Permission denied)"
```
I've managed to run the following command in `controller-acl-init` but not in `api-gateway-controller-acl-init` container:

```
consul-k8s-control-plane acl-init \	
            -component-name=api-gateway-controller \	
            -acl-auth-method=consul-consul-k8s-component-auth-method-REDACTED \	
            -primary-datacenter=REDACTED \	
            -consul-api-timeout=1m \	
            -log-level=info \	
            -log-json=false
```
I have also been able to complete the `initContainer` using the "consul-controller" service account instead of "consul-api-gateway-controller".

But right now the Helm chart is broken and I have to keep api gateway disabled to keep using it.

### Expected behavior

Consul api-gateway-controller service account is expected to have authorization to run api gateway acl init.

### Environment details

- Kubernetes version: v1.21.x
- AWS EKS

nathancoleman:
Hi @manobi , looking into this

manobi:
@nathancoleman if there is something I can do by editing the Helm release, just tell me and I can try to help you debug.
Thank you.

nathancoleman:
[...]

PR #1462, opened 2022-08-29
api-gateway: configure ACL auth appropriately in secondary dc

nathancoleman:
Fixes #1344 

**Changes proposed in this PR:**
When federation is enabled and the API Gateway controller is deployed in a secondary datacenter:
- use correct `acl-auth-method`
- configure the `primary-datacenter`

This matches what we do today for [mesh gateway](https://github.com/hashicorp/consul-k8s/blob/4b630fc687c82860a5fece2a148f82db6f495ab9/charts/consul/templates/mesh-gateway-deployment.yaml#L157-L162) and [consul controller](https://github.com/hashicorp/consul-k8s/blob/a16c4ee0a02320c812f78bc2f33d192f409f62be/charts/consul/templates/controller-deployment.yaml#L110-L115) deployments.

Shout out to @t-eckert for pointing me in the right direction. Since he sent me this patch, I've included him as a co-author on the change commit.

**How I've tested this PR:**

**How I expect reviewers to test this PR:**

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

Issue #1344, opened 2022-07-15
API Gateway Controller in secondary datacenter has insufficient permissions


krarey:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

When deploying a federated secondary Consul datacenter via Helm chart, the API Gateway Controller deployment is configured to retrieve a token at launch time via Kubernetes auth method. This token has the `local` flag set, and the associated policy is further scoped only to the secondary datacenter.

Because this token is used to create config-entry resources, which are globally created in the primary datacenter and replicated back to the secondaries, attachment of new HTTPRoute and TCPRoute resources within the secondary cluster fails to complete as the attached token is invalid in the primary DC. This prevents creation of the underlying Consul `*-gateway`, `service-defaults`, and `service-intentions` resources managed by the API Gateway Controller.

### Reproduction Steps
1. Apply Kubernetes Gateway SIG and Consul API Gateway CRDs to a cluster
2. Use Helm chart to deploy a Consul secondary datacenter with ACLs enabled
3. Deploy a Gateway resource in the Kubernetes secondary DC. e.g.:
```
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: Gateway
metadata:
  name: ns-api-gateway
  namespace: consul-infra
spec:
  gatewayClassName: consul-api-gateway
  listeners:
  - protocol: HTTP
    port: 8080
    name: http
    allowedRoutes:
      namespaces:
        from: All
```
5. Associate an `HTTPRoute` or `TCPRoute` with the deployed Gateway, that references a running and connect-injected upstream service. e.g.:
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: dashboard
  namespace: webapp
---
apiVersion: v1
kind: Service
metadata:
  name: dashboard
  namespace: webapp
spec:
  selector:
    app: dashboard
  ports:
  - port: 9002
    targetPort: 9002
    name: dashboard
---
apiVersion: apps/v1
kind: Deployment
metadata:
[...]

PR #1481, opened 2022-09-06
server-acl-init: Create global ACL auth method for API Gateway in secondary dc

nathancoleman:
This is a followup for #1462 

API Gateway needs to discover services in other datacenters (see #1344 and #1479), so
- Use global ACL policy for `api-gateway-controller` instead of local
- Use correct auth method name for binding when deployed in secondary datacenter

**How I've tested this PR:**
Install primary and secondary Consul clusters as described in [Federation Between Kubernetes Clusters](https://www.consul.io/docs/k8s/deployment-configurations/multi-cluster/kubernetes), making sure to add to the Consul configuration to enable API Gateway and use the control plane image from this PR:
```yaml
global:
  imageK8S: "hashicorpdev/consul-k8s-control-plane:f1a9304a"
...

apiGateway:
  enabled: true
  image: "hashicorp/consul-api-gateway:0.4.0"
```

The difference that we're looking for is that the API Gateway controller will not complete its init process w/ the current `main` branch but will with these changes pinned for `global.imageK8S`. This is because the global auth method exists with the right policies and bindings now due to the changes in the init job.

> **Note**
> The API Gateway controller still will not be able to create functional Gateways in the secondary datacenter. This is due to the issue described in https://github.com/hashicorp/consul-api-gateway/issues/361; however, the changes in this PR fix the acl-init which is half of the problem.

See above

**Checklist:**
- [x] Tests added
- [x] CHANGELOG entry added 


manobi:
I've tried with `hashicorpdev/consul-k8s-control-plane:f1a9304a`. 
It fixes the `api-gateway-controller-acl-init` issue, but `api-gateway-controller` keeps restarting with the following log message:

```
2022-09-07T14:09:34.911Z [ERROR] consul/certmanager.go:253: consul-api-gateway-server.cert-manager: error grabbing leaf certificate: error="Unexpected response code: 403 (rpc error making call: rpc error making call: Permission denied: token with AccessorID 'REDACTED' lacks permission 'service:write' on \"consul-api-gateway-controller\")"
```

nathancoleman:
@manobi that matches what I was seeing while testing yesterday. Have been discussing internally what all it will take to get the setup with a controller per cluster working, and this is the first of 2-3 I believe. The error that the controller is logging I believe will require a followup PR over in the consul-api-gateway repo. I will pursue merge of this change and then spend some time today working on that followup PR.

manobi:
@nathancoleman Thank you for all the effort so far.
I'm happy to confirm that this PR fixes the original issue [1479](https://github.com/hashicorp/consul-k8s/issues/1479).

If you need help to test something in consul-api-gateway repo just ping me.

nathancoleman:
I've identified the next couple of issues preventing Consul API Gateway from working in federated setups with a controller per datacenter. Still working to determine scope and priority.

In the meantime, we've added a section to our docs describing the current limitations of API Gateway in federated setups [here](https://www.consul.io/docs/api-gateway/tech-specs#limited-support-of-some-consul-features)

andrewstucki (approved):
This looks fine as far as I can tell, but I'd want someone who is a bit more familiar with the acl-init code to sign off before merging.

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -592,7 +592,15 @@ func (c *Command) Run(args []string) int {
 			return 1
 		}
 		serviceAccountName := c.withPrefix("api-gateway-controller")
-		if err := c.createACLPolicyRoleAndBindingRule("api-gateway-controller", rules, consulDC, primaryDC, localPolicy, primary, localComponentAuthMethodName, serviceAccountName, consulClient); err != nil {
+
+		// API gateways require a global policy/token because they must
+		// discover services in other datacenters.
```
mikemorris, line 597:
```suggestion
		// create config-entry resources in the primary, even when deployed to a secondary datacenter
```
Isn't ^ the actual reason why this is necessary? We aren't enabling routing to services in other datacenters yet.

nathancoleman: You're right, good catch!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+BUG FIXES:
+* Control plane
+  * Create global ACL auth method for API Gateway in secondary datacenter [[GH-1481](https://github.com/hashicorp/consul-k8s/pull/1481)]
```
kschoche, line 5:
```suggestion
  * Use global ACL auth method to provision ACL tokens for API Gateway in secondary datacenter [[GH-1481](https://github.com/hashicorp/consul-k8s/pull/1481)]
```

kschoche (approved):
Looks great!!!
I left one suggestion in the changelog to be a bit more accurate in the description of the issue.

Issue #536, opened 2021-06-16
Graceful shutdown with injected envoy-sidecar


svobol13:
Envoy proxy sidecar receives SIGTERM at the exact same moment as my main container. In oppossite to my main container (which shuts down in like 15-30 seconds) envoy sidecar shuts down immediately (0.5 - 3s). This means that my main lost upstream connections and cannot gracefully shutdown - even rolling update means lost requests/data.

There should be some kind of mechanism so the upstream listeners exits as last. My main container is Consul Connect enabled and communicate with upstreams through Connect but the service itself is not being accessed through Connect but through Consul DNS instead.

Is there a workaround/hack (some kind of prestop sleep) or do I have get rid of Consul Connect?

Similar issues:
- https://github.com/istio/istio/issues/7136
- https://github.com/linkerd/linkerd2/issues/3747

pedrohdz:
Have you tried adding `terminationGracePeriodSeconds` to your pod?  You might end up with https://github.com/hashicorp/consul-k8s/issues/540 instead then.

ishustava:
Thanks for the issue @svobol13! This is the problem that is a larger issue in [Kubernetes itself](https://docs.google.com/document/d/1Drw9C_Ljpcr4X9UPLvms1fn8uMRnTfJLb-xipgX4C1M/edit#heading=h.a79rvo5izd4l) in that there is no lifecycle hooks that can help us control container shutdown. We'd need to investigate how to work around this until there's a proper solution in k8s.

kevin-lindsay-1:
Posting a reference to a comment that I made after going down this rabbit hole, to hopefully save others some time:
https://github.com/istio/istio/pull/18333#issuecomment-895522403

Joxit:
@pedrohdz I did some tests and `terminationGracePeriodSeconds` is not enough. As you can see in this image, the grace period starts after the SIGTERM signal, at this moment it's already too late.
![image](https://user-images.githubusercontent.com/5153882/147744551-5127e4f1-df3d-43e2-8dc0-ba7430ed2916.png)

The problem really comes from Envoy which should not stop without having completed the last received requests on SIGTERM.

My workaround was add a `preStop: sleep 30s` on my container + `terminationGracePeriodSeconds` and to overload the envoy docker image to ignore SIGTERM.

This work fine because consul remove the service from its catalog when the SIGTERM is triggered, so I have 30 second to finish the work.

[...]

PR #1482, opened 2022-09-06
Adding support for lifecycle hooks and health probe for sidecars

narendrapatel:
Changes proposed in this PR:
- Adding support to start main app containers after sidecar is started. Done by adding a postStart lifecycle hook to the sidecar container. 
(via boolean annotation: consul.hashicorp.com/sidecar-hold-app-until-proxy-starts)
With reference from: https://github.com/istio/istio/issues/11130#issuecomment-645424556

- Adding support to stop sidecar gracefully after main app container stops. Done by adding a preStop lifecycle hook to the sidecar container. 
(via boolean annotation consul.hashicorp.com/sidecar-proxy-graceful-shutdown)
With reference from: https://github.com/istio/istio/issues/7136

- Support to add health probes. Done by adding Liveness check to the sidecar container.
(via boolean annotation consul.hashicorp.com/sidecar-configure-probes)

One impact due to this:
On adding `sidecar-hold-app-until-proxy-starts`, the sidecar becomes the default container when using `kubectl exec`

These features are added as opt in via annotations

How I've tested this PR:
- Added the annotations to a k8s deployment.
- Post adding could see postStart, preStop and liveness gets added to the deployment manifest
- Tested holding app container startup by running continuous calls to sidecar on startup from app. Working as expected. Without the annotation the call to sidecar fails as the app container starts up first.
- Tested graceful shutdown by adding a preStop to the app container. Works as expected. Without the annotation the envoy sidecar gets terminated immediately while the app is still completing its preStop and all calls to the sidecar start failing. This is running in our environment perfectly from quite some time. https://github.com/hashicorp/consul-k8s/issues/536#issuecomment-1211722794
- On adding annotation for liveness it comes up correctly in the pod manifest. This is useful for mesh deployments that do not yet use transparent proxies
- The annotation expects only boolean value. Any other value would skip adding the annotation functionality.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


david-yu:
Hi @narendrapatel have you been able to test and see that this works on Consul K8s?

narendrapatel:
> Hi @narendrapatel have you been able to test and see that this works on Consul K8s?

@david-yu yes I have tested this functionality in my local K8s cluster.

david-yu:
Hi @narendrapatel thanks for the PR, I initially thought this was going to be a PR to address health probes for sidecars but I am as able to see the whole picture here a bit better now that the whole PR is put up. It looks like the use case is to address a few different proxy lifecycle management use cases. 

The Consul K8s team is intending to support this use case after our 1.14 release, however we can't merge this PR as is since we will be making significant architecture changes in how the proxy is deployed and configured as part of that release. I'll leave this PR open for the time being so others can use this for their own purposes as well.

junjie-landing:
@david-yu thanks for the sharing. 

Correct me I'm wrong, it seems we just release 1.0.1 not long ago: https://github.com/hashicorp/consul-k8s/tags, Does it mean 1.14 won't be coming any time soon? May I ask what is the expected time to release 1.14?

junjie-landing:
@david-yu thanks for the sharing. 



I don't it was mentioned in the 1.14 release doc. Maybe I am overlooking something? 

https://github.com/hashicorp/consul/releases/tag/v1.14.0

narendrapatel:
I've updated the PR to be compatible with consul version v1.14 and consul control plane 1.0.0. But with the consul dataplane now using distroless base container, we'll need to add couple of binaries like `sh`, `wget`, `sleep`, `netstat`, `grep` and `wc`. The docker file for consul dataplane is available [here](https://github.com/hashicorp/consul-dataplane).
Changes required for consul version < 1.14 and consul k8s control plane < 1.0.0 is [here](https://github.com/narendrapatel/consul-k8s/pull/3/files)

oliver-buckley-salmon-db:
Hi,
Is this ever going to get merged? This functionality is required by many consumers, due to the non-graceful shutdown of services in the mesh.
Thanks

shanilempel:
Any news on that? We'd could highly benefit from such functionality.

oliver-buckley-salmon-db (approved):

david-yu:
Closing as this is now resolved by https://github.com/hashicorp/consul-k8s/pull/2233 and will be released in upcoming patch releases for 1.0.x, 1.1.x and 1.2.x.

oliver-buckley-salmon-db:
Thanks every one, this was a critical fix for us.

psypuff:
@david-yu it seems #2233 doesn't address the startup issue fixed by this PR (Envoy proxy sidecar goes up only after the app container). Without a fix Consul Connect is pretty much unusable for us, we wished a fix would make its way to the just-released 1.2 version. What's the reason to not fix both lifecycle issues at once? Since this PR is now closed is there any other place to track the startup issue progress?

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -140,6 +154,54 @@ func (w *MeshWebhook) consulDataplaneSidecar(namespace corev1.Namespace, pod cor
...
+
+	if err != nil || !hold {
+		return nil, err
+	}
+	postStart := &corev1.Handler{
+		Exec: &corev1.ExecAction{
+			Command: []string{
+				"/bin/sh",
+				"-c",
+				`total_time=0; until wget --spider localhost:19000;` +
```
ilpianista, line 192: The `/ready` endpoint should be more appropriate for this check IMHO:

```
#!/bin/sh

while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' localhost:19000/ready)" != "200" ]]; do
  sleep 1
done
```

komapa: We do `while [[ "$(curl -sf http://127.0.0.2:19000/ready)" != "LIVE" ]]; do sleep 1; done` with great success :)

PR #1484, opened 2022-09-06
update service config crds, update passive health check

malizz:
Changes proposed in this PR:
- update service default config's passive health check to add EnforcingConsecutive5xx parameter. 
-

How I've tested this PR:

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


acceptance/tests/fixtures/bases/crds-oss/servicedefaults.yaml
```diff
@@ -18,6 +18,7 @@ spec:
       passiveHealthCheck:
         interval: 1s
         maxFailures: 10
+      enforcing_consecutive_5xx: 60
```
DanStough, line 21: I think the indent is off here, right?

control-plane/api/v1alpha1/servicedefaults_types_test.go
```diff
@@ -188,8 +191,9 @@ func TestServiceDefaults_ToConsul(t *testing.T) {
 							MaxConcurrentRequests: intPointer(10),
 						},
 						PassiveHealthCheck: &capi.PassiveHealthCheck{
-							Interval:    2 * time.Second,
-							MaxFailures: uint32(20),
+							Interval:                2 * time.Second,
+							MaxFailures:             uint32(20),
+							EnforcingConsecutive5xx: uint32(100),
```
DanStough, line 196: Shouldn't this be a pointer?

DanStough (changes requested):
Awesome work @malizz 👏 ! Some minor comments and one big question. I was wondering why the tests use the `uint32` instead of the pointer, and I think there is a mismatch between the struct in Consul [config](https://github.com/hashicorp/consul/blob/f23cdac8e4c886bba4ddd3a8f78e9cf69b6895f5/api/config_entry.go#L203) and [api](https://github.com/hashicorp/consul/blob/f23cdac8e4c886bba4ddd3a8f78e9cf69b6895f5/agent/structs/config_entry.go#L969). I think they need to be the same. WDYT?

Also feel free to add the kubernetes team for review since there needs to be 2 approvers.

control-plane/api/v1alpha1/servicedefaults_types_test.go
```diff
@@ -193,7 +193,7 @@ func TestServiceDefaults_ToConsul(t *testing.T) {
 						PassiveHealthCheck: &capi.PassiveHealthCheck{
 							Interval:                2 * time.Second,
 							MaxFailures:             uint32(20),
-							EnforcingConsecutive5xx: uint32(100),
+							EnforcingConsecutive5xx: uint32Pointer(100),
```
kschoche, line 196: We recently went through and moved everything in the `consul-k8s` repo to use the pointer `k8s.io/utils/pointer` library for these, could you swap them out for it please? 
```suggestion
							EnforcingConsecutive5xx: pointer.Uint32(100),
```

malizz: Great to know, thanks!

control-plane/config/crd/bases/consul.hashicorp.com_servicedefaults.yaml
```diff
@@ -234,6 +234,13 @@ spec:
                           upstream proxy instances will be monitored for removal from
                           the load balancing pool.
                         properties:
+                          enforcing_consecutive_5xx:
+                            description: EnforcingConsecutive5xx is the % chance that
+                              a host will be actually ejected when an outlier status
+                              is detected through consecutive 5xx. This setting can
+                              be used to disable ejection or to ramp it up slowly.
+                            format: int32
+                            type: integer
```
kschoche, line 243: One more quick question, and it's more that I don't know the answer than anything, but we're defining the `enforcing_consecutive_5xx` property as `int32` here (and everywhere) but have  as a `uint32` in `servicedefaults_typos.go`: 
```
EnforcingConsecutive5xx *uint32 `json:"enforcing_consecutive_5xx,omitempty"`
```

I don't think we'll ever have an overflow(underflow) here but it would make sense to have them both be the same type?

kschoche (approved):
This looks great!
I left one semi-blocking question about a type mismatch but once that's cleared/clarified it looks great!

DanStough (approved):
🚀

control-plane/config/crd/bases/consul.hashicorp.com_servicedefaults.yaml
malizz: Good point, I believe those yaml files were generated automatically via `make ctrl-manifests`. I can change them to `uint32` manually but I'm concerned about next time we run the script.

kschoche: I did a little bit of research and it looks like we're already using this model everywhere else since `int32` is used by Kubernetes for CRDs everywhere. So no changes needed!

malizz: great, thanks!

david-yu:
@malizz Are you able to rebase? Perhaps we should get this backported to 0.49.x using the `backport/0.49.x` label

david-yu:
Thanks cc @curtbushko @wilkermichael for the release tomorrow.

PR #1492, opened 2022-09-09
Rolling back helm upgrade in CLI back to 3.6.1

jmurret:
- Reverts hashicorp/consul-k8s#1465
- CI is failing for TestConnectInject when the CLI runs `proxy list` and `proxy read` commands and has this error: `Error initializing Kubernetes client:
        exec plugin: invalid apiVersion "client.authentication.k8s.io/v1alpha1"`


t-eckert:
I'll investigate the root cause of this more today. I would like to avoid rolling back the update as it fixes a CVE. The error seems to be with the install process which more cleanly maps to the issue with the Helm dependency. However, I was not able to replicate the error with a new EKS cluster.

t-eckert (approved):
Let's do this with a note to undo it before the next release.

t-eckert:
Opened an issue on the Helm repo requesting a backport of the CVE fix to `3.6.x`: https://github.com/helm/helm/issues/11342

kschoche (approved):

PR #1493, opened 2022-09-12
Update docs storageclass

im2nguyen:
Changes proposed in this PR:
- Update storageclass docs to link to the reference arch page

> Most customers are using default values for storageclass for Consul helm chart which is not suitable for running enterprise workloads and since Consul is write limited by disk I/O  and as per reference architecture it needs disk with minimum of 3000+ IOPS ( even for Dev clusters) this can result in performance issues.

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


kschoche (approved):
Excellent work, thank you!!

david-yu (approved):

PR #1527, opened 2022-09-23
CNI: Add support for OpenShift and Multus

curtbushko:
Changes proposed in this PR:

This PR adds support for OpenShift and Multus in CNI. OpenShift uses Multus as the base CNI plugin.

- Added SecurityContextConstraints CRD so that CNI has the permissions to run on OpenShift and gave the consul-cni service account permissions to reach OpenShifts SCC CRD.
- Added a --multus flag to the CNI installer so that the flag can be used by the CNI plugin.
- When --multus flag is set, create a default Network Attachment Definition CRD for Multus to use. Note, Pods require an annotation in order for Multus to run the consul-cni plugin. Please see [Multus documentation](https://github.com/k8snetworkplumbingwg/multus-cni) for more information about how to use Multus.
- Fix: Fixed installer to rename .conf file .conflist file when consul-cni is added to a base plugin.
- Improvement: The plugin is now more consistent at updating annotations. 

How I've tested this PR:

- Manual testing of Multus on OpenShift
- Manual testing of Multus on Kind
- Acceptance and unit tests.

:eyes:

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/cni-securitycontextconstraints.yaml
+++ b/charts/consul/templates/cni-securitycontextconstraints.yaml
@@ -1,0 +1,50 @@
+{{- if (and (.Values.connectInject.cni.enabled) (.Values.global.openshift.enabled)) }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: {{ template "consul.fullname" . }}-cni
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: cni
+  annotations:
+    kubernetes.io/description: {{ template "consul.fullname" . }}-cni are the security context constraints required
+      to run consul-cni.
+allowHostDirVolumePlugin: true
+allowHostIPC: false
+allowHostNetwork: true
+allowHostPID: false
+allowHostPorts: true
+allowPrivilegeEscalation: true
+allowPrivilegedContainer: true
+allowedCapabilities: null
+defaultAddCapabilities: null
+fsGroup:
+  type: MustRunAs
+groups: []
+priority: null
+readOnlyRootFilesystem: false
+requiredDropCapabilities:
+- KILL
+- MKNOD
+- SETUID
+- SETGID
+runAsUser:
+  type: MustRunAsRange
+seLinuxContext:
+  type: MustRunAs
+supplementalGroups:
+  type: MustRunAs
+users: []
+volumes:
+- configMap
+- downwardAPI
+- emptyDir
+- persistentVolumeClaim
+- projected
+- secret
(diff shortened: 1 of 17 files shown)


charts/consul/templates/cni-securitycontextconstraints.yaml
```diff
@@ -0,0 +1,50 @@
...
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: cni
+  annotations:
+    kubernetes.io/description: {{ template "consul.fullname" . }}-cni are the security context constraints required
+      to run consul-cni.
+allowHostDirVolumePlugin: true
+allowHostIPC: false
+allowHostNetwork: true
```
ishustava, line 18: Double checking: Do we need host network?

charts/consul/templates/cni-securitycontextconstraints.yaml
```diff
@@ -0,0 +1,50 @@
...
+allowHostPID: false
+allowHostPorts: true
```
ishustava, line 20: do we need host ports?

charts/consul/values.yaml
```diff
@@ -1960,7 +1960,20 @@ connectInject:
...
+    # CNI plugin. Instead, a NetworkAttachementDefinition CustomResourceDefinition (CRD) will be created in the helm
+    # release namespace. Following multus plugin standards, an annotation is required in order for the consul-cni plugin
+    # to be executed and for your service to be added to the Consul Service Mesh. 
+    #
+    # Add the annotation `'k8s.v1.cni.cncf.io/networks': '[{ "name":"consul-cni","namespace": "consul" }]'` to your pod
+    # to use the default installed NetworkAttachementDefinition CRD.
+    #
+    # Please refer to the [Multus Quickstart Guide](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/quickstart.md)
+    # for more information about using multus. 
+    # @type: string
```
ishustava, line 1974: Beautiful!

ishustava (approved):
It's perfect!!! Just a couple of clarifying comments.

charts/consul/templates/cni-securitycontextconstraints.yaml
curtbushko: Nice catch!

I just sanity tested and we do not.  They have been removed.

charts/consul/templates/cni-securitycontextconstraints.yaml
curtbushko: I just sanity tested and we do not.  They have been removed.

david-yu:
Thanks @curtbushko would also be possible to create a PR for how to use the CNI Plugin on Openshift here: https://www.consul.io/docs/k8s/installation/install#enable-the-consul-cni-plugin

wilkermichael (approved):

PR #1533, opened 2022-09-26
agentless: enable admin partitions with agentless

ishustava:
Changes proposed in this PR:
* integrate partition-init job with consul-server-connection-manager
* integrate controller with consul-server-connection-manager
* modify partition token to be be able read all services so that it works with connection manager

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/controller-deployment.yaml
```diff
@@ -1,5 +1,6 @@
 {{- if .Values.controller.enabled }}
 {{- if and .Values.global.adminPartitions.enabled (not .Values.global.enableConsulNamespaces) }}{{ fail "global.enableConsulNamespaces must be true if global.adminPartitions.enabled=true" }}{{ end }}
+{{- if and .Values.externalServers.enabled (not .Values.externalServers.hosts) }}{{ fail "externalServers.hosts must be set if externalServers.enabled is true" }}{{ end -}}
```
ishustava, line 3: I'm not going to update bats for controller in this PR because we'll combine this deployment with connect-inject deployment in a future PR before we go GA (NET-994)

PR #1537, opened 2022-09-27
Ingress Gateways integration with Agentless and Consul Dataplane

thisisnotashwin:
Changes proposed in this PR:
- Register Ingress Gateways using the endpoints controller.
- Use Consul Dataplane instead of Envoy

How I've tested this PR:
- Acceptance tests
- Unit and BATS tests.

- Code Review

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/ingress-gateways-deployment.yaml
```diff
@@ -107,374 +146,165 @@ spec:
...
+          requests:
+            memory: "50Mi"
+            cpu: "50m"
+          limits:
+            memory: "50Mi"
+            cpu: "50m"
+      containers:
+      - name: ingress-gateway
+        image: {{ $root.Values.global.imageConsulDataplane | quote }}
+        {{- if (default $defaults.resources .resources) }}
```
t-eckert, line 217:
```suggestion
        {{- if (default $defaults.resources $root.resources) }}
```
Does the non-default value need to be pinned to `$root` here and below?

t-eckert (approved):
Looks great!

charts/consul/templates/ingress-gateways-deployment.yaml
thisisnotashwin: Its only those that start with a `.`

thisisnotashwin:
Closing in favor of https://github.com/hashicorp/consul-k8s/pull/1544 to avoid rebase hell.

PR #1544, opened 2022-09-27
Ingress Gateways integration with Agentless and Consul Dataplane

thisisnotashwin:
Changes proposed in this PR:
- Register Ingress Gateways using the endpoints controller.
- Use Consul Dataplane instead of Envoy

How I've tested this PR:
- Acceptance tests
- Unit and BATS tests.

- Code Review

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
+++ b/acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
@@ -31,10 +31,10 @@ func TestIngressGatewaySingleNamespace(t *testing.T) {
 		secure bool
 	}{
 		{
-			false,
+			secure: false,
 		},
 		{
-			true,
+			secure: true,
 		},
 	}
 	for _, c := range cases {
@@ -42,8 +42,7 @@ func TestIngressGatewaySingleNamespace(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			ctx := suite.Environment().DefaultContext(t)
 
-			// Install the Helm chart without the ingress gateway first
-			// so that we can create the namespace for it.
+			igName := "ingress-gateway"
 			helmValues := map[string]string{
 				"connectInject.enabled": "true",
 				"connectInject.consulNamespaces.consulDestinationNamespace": testNamespace,
@@ -51,6 +50,11 @@ func TestIngressGatewaySingleNamespace(t *testing.T) {
 				"global.enableConsulNamespaces": "true",
 				"global.acls.manageSystemACLs":  strconv.FormatBool(c.secure),
 				"global.tls.enabled":            strconv.FormatBool(c.secure),
+
+				"ingressGateways.enabled":                     "true",
+				"ingressGateways.gateways[0].name":            igName,
+				"ingressGateways.gateways[0].replicas":        "1",
+				"ingressGateways.gateways[0].consulNamespace": testNamespace,
 			}
 
 			releaseName := helpers.RandomName()
@@ -59,25 +63,6 @@ func TestIngressGatewaySingleNamespace(t *testing.T) {
 			consulCluster.Create(t)
 
 			consulClient, _ := consulCluster.SetupConsulClient(t, c.secure)
-
-			// Create the destination namespace in the non-secure case.
-			// In the secure installation, this namespace is created by the server-acl-init job.
-			if !c.secure {
-				logger.Logf(t, "creating the %s namespace in Consul", testNamespace)
-				_, _, err := consulClient.Namespaces().Create(&api.Namespace{
-					Name: testNamespace,
-				}, nil)
-				require.NoError(t, err)
-			}
-
-			igName := "ingress-gateway"
-			logger.Log(t, "upgrading with ingress gateways enabled")
-			consulCluster.Upgrade(t, map[string]string{
-				"ingressGateways.enabled":                     "true",
-				"ingressGateways.gateways[0].name":            igName,
-				"ingressGateways.gateways[0].replicas":        "1",
-				"ingressGateways.gateways[0].consulNamespace": testNamespace,
-			})
 
 			logger.Logf(t, "creating Kubernetes namespace %s", testNamespace)
 			k8s.RunKubectl(t, ctx.KubectlOptions(t), "create", "ns", testNamespace)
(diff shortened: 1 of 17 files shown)


acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
```diff
@@ -44,13 +44,19 @@ func TestIngressGatewaySingleNamespace(t *testing.T) {
 
 			// Install the Helm chart without the ingress gateway first
 			// so that we can create the namespace for it.
+			igName := "ingress-gateway"
```
t-eckert, line 47: Does this change need to be undone? Are you going to run into issues installing without the Consul namespace?

acceptance/tests/ingress-gateway/ingress_gateway_test.go
```diff
@@ -18,24 +18,17 @@ const StaticClientName = "static-client"
...
-			false,
-		},
-		{
-			true,
-			false,
 		},
 		{
 			true,
-			true,
 		},
```
t-eckert, line 28: IDC either way, but in your other PR, I remember you changed the struct values to be named (e.g. `{ secure: true }`).

acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
thisisnotashwin: This test only runs when namespaces are enabled

acceptance/tests/ingress-gateway/ingress_gateway_test.go
thisisnotashwin: that was the other file. but i can make it in this file as well.

acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
```diff
@@ -31,10 +31,10 @@ func TestIngressGatewaySingleNamespace(t *testing.T) {
 		secure bool
 	}{
 		{
-			false,
+			secure: false,
```
t-eckert, line 34: You could probably cut the last three lines of the comment on this test.

acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
t-eckert: Right, namespaces are enabled, but does the particular namespace the ingress gateway is being registered in exist? I thought that's why the previous code had you install, create the particular namespace, then upgrade.

thisisnotashwin: aaha! i ignored that comment entirely lol. i have updated the endpoints controller to now create the gateway namespace if it does not already exist

t-eckert: Sweet! Then it's good to go.

t-eckert: Oh, actually, there is still test code that creates a Consul namespace in the secure case. I believe it's a no-op and can be removed.

thisisnotashwin: done..good catch

acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
```diff
@@ -60,25 +64,6 @@ func TestIngressGatewaySingleNamespace(t *testing.T) {
...
 
-			// Create the destination namespace in the non-secure case.
-			// In the secure installation, this namespace is created by the server-acl-init job.
-			if !c.secure {
-				logger.Logf(t, "creating the %s namespace in Consul", testNamespace)
-				_, _, err := consulClient.Namespaces().Create(&api.Namespace{
-					Name: testNamespace,
-				}, nil)
-				require.NoError(t, err)
-			}
```
ishustava, line 71: Oh nice! because endpoints controller will create the ns now, right?

charts/consul/templates/ingress-gateways-deployment.yaml
```diff
@@ -107,374 +146,172 @@ spec:
...
-          {{- $initContainer := .initCopyConsulContainer }}
-          {{- if (or $initContainer $defaults.initCopyConsulContainer) }}
-          {{- if (default $defaults.initCopyConsulContainer.resources $initContainer.resources) }}
-          resources: {{ toYaml (default $defaults.initCopyConsulContainer.resources $initContainer.resources) | nindent 12 }}
-          {{- end }}
-          {{- end }}
-        {{- if (and $root.Values.global.tls.enabled $root.Values.global.tls.enableAutoEncrypt) }}
-        {{- include "consul.getAutoEncryptClientCA" $root | nindent 8 }}
+        volumeMounts:
+        {{- if $root.Values.global.tls.enabled }}
```
ishustava, line 225: this doesn't need to be mounted if external servers with use system roots is set

charts/consul/templates/ingress-gateways-deployment.yaml
```diff
@@ -107,374 +146,172 @@ spec:
...
+          -service-name={{ template "consul.fullname" $root }}-{{ .name }} \
+          -log-level={{ default $root.Values.global.logLevel }} \
+          -log-json={{ $root.Values.global.logJSON }}
+        volumeMounts:
         - name: consul-service
-          emptyDir:
-            medium: "Memory"
+          mountPath: /consul/service
         {{- if $root.Values.global.tls.enabled }}
-        {{- if not (and $root.Values.externalServers.enabled $root.Values.externalServers.useSystemRoots) }}
```
ishustava, line 116: I think we still need the this `if` condition

charts/consul/templates/ingress-gateways-deployment.yaml
```diff
@@ -107,374 +146,172 @@ spec:
...
+              fieldPath: metadata.namespace
+        command:
+        - /bin/sh
+        - -ec
+        - |
+          consul-dataplane \
+            -envoy-ready-bind-address=$POD_IP \
+            -envoy-ready-bind-port=21000 \
+            {{- if $root.Values.externalServers.enabled }}
+            -addresses={{ $root.Values.externalServers.hosts | first }} \
```
ishustava, line 251: we need to quote it so that command like `-addresses="exec=..."` works
```suggestion
            -addresses={{ $root.Values.externalServers.hosts | first | quote }} \
```

ishustava (approved):
Beautiful!! I had some comments, but won't block assuming they're addressed before merging! Amazing work!

acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
thisisnotashwin: exactly!!

PR #1554, opened 2022-09-28
agentless: support providing envoy extra args and multi-port workaround

ishustava:
Changes proposed in this PR:
- Change to use the latest consul-dataplane image. This resulted in renames of some `-login*` flags.
- Allow passing extra args to envoy:
  - user-provided via annotation
  - `--base-id` for multi-port
- Allow setting envoy concurrency via consul-dataplane
- Remove `service-address` command as it's no longer used
- Don't set TLS server name to something like `server.dc1.consul` unless it's set in the `externalServers` helm values. This should now be inferred in consul-dataplane 

How I've tested this PR:
acceptance tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


thisisnotashwin (approved):
HANG IT IN THE LOUVRE!! :chefkiss:

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -242,8 +242,7 @@ const multiportAdmin = "multiport-admin"
 // two ports. This tests inbound connections to each port of the multiport app, and outbound connections from the
 // multiport app to static-server.
 func TestConnectInject_MultiportServices(t *testing.T) {
-	t.Skipf("skipping until multi-port workaround is supported")
-	for _, secure := range []bool{false, true} {
+	for _, secure := range []bool{true} {
```
t-eckert, line 245: If we aren't testing the `secure=false` case anymore, we could just hardcode it and remove the loop.

t-eckert (approved):
LGTM, just the one comment non-blocking.

PR #1555, opened 2022-09-28
Update changelog for agentless beta

ishustava:


thisisnotashwin (approved):
I love this! I only have 1 non-blocking question. Do we want to maintain some mapping between dataplane version and the version of envoy that it runs under the hood? Do users care to have that information, and if so, is the changelog an appropriate place for it? (it could be in the dataplane changelog)

t-eckert (approved):

david-yu (approved):

PR #1557, opened 2022-09-29
Helm: Support minAvailable on connect injector PDB

hamishforbes:
It looks like the PDB for the connect injector has copied the logic from the Server PDB.
In that it is extremely conservative and attempts to maintain n+1 pods, which makes complete sense for Consul Servers as you do want to maintain a quorum at all times.

The injector deployment uses leader election and only 1 pod is actually active at any one time, requiring n/2-1 pods to be available is unnecessary.

Furthermore, with the default 2 replicas `maxUnavailable` is set at 0, which requires manual human operator intervention to, for example, drain a node running an injector pod.
See [Single-instance Stateful Application](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#think-about-how-your-application-reacts-to-disruptions) 

For the connect injector deployment a `minAvailable` configuration makes much more sense.

These changes do not change the default behaviour, although I think the default behaviour probably *should* be changed.
This is opt-in only, users must configure the new value explicitly.

Changes proposed in this PR:
- Add `connectInject.disruptionBudget.minAvailable` to values
- `minAvailable` is null by default
- `minAvailable` takes precedence over the `maxUnavailable` configuration

How I've tested this PR:
* Added tests

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/connect-injector-disruptionbudget.yaml
+++ b/charts/consul/templates/connect-injector-disruptionbudget.yaml
@@ -17,7 +17,11 @@ metadata:
     release: {{ .Release.Name }}
     component: connect-injector
 spec:
+  {{- if .Values.connectInject.disruptionBudget.minAvailable }}
+  minAvailable: {{ .Values.connectInject.disruptionBudget.minAvailable }}
+  {{- else }}
   maxUnavailable: {{ template "consul.pdb.connectInject.maxUnavailable" . }}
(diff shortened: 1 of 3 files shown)


kschoche (approved):
Thanks for this PR!

ishustava (approved):

PR #1567, opened 2022-09-29
Add node meta to represent synthetic node

thisisnotashwin:
- Service and Gateway registration now have Node Meta that specify that they are registered on a synthetic node.


t-eckert (approved):

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -32,6 +32,7 @@ const (
 	MetaKeyKubeServiceName     = "k8s-service-name"
 	MetaKeyKubeNS              = "k8s-namespace"
 	MetaKeyManagedBy           = "managed-by"
+	MetaKeySyntheticNode       = "synthetic-node"
```
ishustava, line 35: It should be prefixed with `consul-`
```suggestion
	MetaKeySyntheticNode       = "consul-synthetic-node"
```

ishustava (approved):
Looks good, just one tiny comment!

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -628,6 +636,7 @@ func (r *EndpointsController) createGatewayRegistrations(pod corev1.Pod, service
 		MetaKeyKubeServiceName: serviceEndpoints.Name,
 		MetaKeyKubeNS:          serviceEndpoints.Namespace,
 		MetaKeyManagedBy:       managedByValue,
+		MetaKeySyntheticNode:   "true",
```
wilkermichael, line 639: Would it be useful to set all the `MetaKeySyntheticNode` in this file and the test file using a const? In case we need to change the state in the future?

wilkermichael (approved):
One minor comment, great work!

control-plane/connect-inject/endpoints_controller.go
ishustava: Ashwin and I synced up, and he said that we can't use because it's reserved. Resolving this!

PR #1570, opened 2022-09-30
Configure the Go workspace for better multi-module editing

mkeeler:
Changes proposed in this PR:
- Addition of Go 1.18 workspace files. This allows gopls to actually work when the whole repo is opened in VS Code (or other IDEs) at once.

How I've tested this PR:
- Editing code in VS Code with gopls in use.

- I don't

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


david-yu:
Thank you! For future reference you can assign 'hashicorp/consul-kubernetes` in the reviewers and it will auto assign two folks from our team to review

go.work.sum
```diff
@@ -0,0 +1,255 @@
+github.com/Azure/azure-sdk-for-go v46.0.0+incompatible h1:4qlEOCDcDQZTGczYGzbGYCdJfVpZLIs8AEo5+MoXBPw=
```
curtbushko, line 1: I am trying to Google the answer but my only question is what keeps this `go.work.sum` file updated if say something is changed under `./control-plane` ?

curtbushko (commented):
I only have 1 question about adding this but nothing blocking.

go.work.sum
mkeeler: I think you call `go work sync` to do this.

t-eckert: Could you add this to the contributing doc, please?

t-eckert:
> Is there a reason why `hashicorp/consul` also shouldn't have this as well?

I don't know if they would really see a benefit as this tool is really for projects where it is useful to have multiple Go modules open at once and Consul takes the approach of having a single module at the root of the repo.

go.work.sum
mkeeler: @t-eckert Is there any particular place in those docs you think would be good. I went looking for where module tidying might be mentioned to add it nearby but that isn't documented?

mkeeler:
@david-yu & @t-eckert hashicorp/consul should get it as well. I happened to be perusing some code locally for consul-k8s and was running into issues with vscode not being able to follow symbol definitions etc so I PR'ed it here.

Consul does have two nested modules, one at the root and then a few others with ./api, ./sdk being the notable two.

go.work.sum
t-eckert: Totally! I think it could go in the [CONTRIBUTING.md](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#rebasing-contributions-against-main) file in a section below "Rebasing contributions against Main".

mkeeler:
After further experimentation, this seems to work great for getting the IDE/gopls to do the right thing. But then will cause some issues with other Go commands (go test being one of them that attempts to be workspace aware by doesn't seem to do it well)

So I think its probably best to close this and wait for the workspaces feature to mature a bit.

For, anyone just wanting their IDE to work they can run `go work init && go work use -r .` locally to see the benefits without these having to be committed to the repo.

t-eckert:
> After further experimentation, this seems to work great for getting the IDE/gopls to do the right thing. But then will cause some issues with other Go commands (go test being one of them that attempts to be workspace aware by doesn't seem to do it well)
> 
> So I think its probably best to close this and wait for the workspaces feature to mature a bit.
> 
> For, anyone just wanting their IDE to work they can run `go work init && go work use -r .` locally to see the benefits without these having to be committed to the repo.

And if it helps, I've only seen this issue with VS Code. GoLand and Neovim work fine without `go work`.

PR #1573, opened 2022-09-30
Remove check for clients in status command

t-eckert:
Changes proposed in this PR:
- Remove the check for Consul clients in the `consul-k8s status` command. Clients don't exist in agentless (RIP).

How I've tested this PR:
- Unit tests
- Manual install -> status check -> uninstall

- Unit tests
- Manual install -> status check -> uninstall
- :eyes:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/cli/cmd/status/status.go
+++ b/cli/cmd/status/status.go
@@ -69,9 +69,7 @@ func (c *Command) Run(args []string) int {
 		c.helmActionsRunner = &helm.ActionRunner{}
 	}
 
-	// The logger is initialized in main with the name cli. Here, we reset the name to status so log lines would be prefixed with status.
 	c.Log.ResetNamed("status")
-
 	defer common.CloseWithError(c.BaseCommand)
 
 	if err := c.set.Parse(args); err != nil {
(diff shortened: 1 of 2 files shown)


cli/cmd/status/status.go
```diff
@@ -128,13 +126,6 @@ func (c *Command) Run(args []string) int {
 		c.UI.Output(s, terminal.WithSuccessStyle())
 	}
 
-	if s, err := c.checkConsulClients(namespace); err != nil {
-		c.UI.Output(err.Error(), terminal.WithErrorStyle())
-		return 1
-	} else {
-		c.UI.Output(s, terminal.WithSuccessStyle())
-	}
-
```
ishustava, line 137: Should we instead check if clients are enabled and only check status then? It's also a bit weird that we check servers all the time even for clusters that might not have servers installed.

t-eckert: Yeah, that's possible. We can have it read the deployment configuration and determine what to check then. That would be a separate set of work. Should I close this PR and take that approach instead?

ishustava: Ok, sounds good. I'll approve this one - I think it's good to get at least that merged in.

ishustava (approved):

cli/cmd/status/status.go
t-eckert: https://hashicorp.atlassian.net/browse/NET-1037

wilkermichael (approved):

PR #1581, opened 2022-10-03
Add `tolerations` and `nodeSelector` to Server ACL init jobs and `nodeSelector` to Webhook cert manager

t-eckert:
This solution is in response to an ask from a customer who wants to set `nodeSelector` for all Pods in their cluster. This adds that capability to `server-acl-init` jobs and `webhook-cert-manager`.

Changes proposed in this PR:
- Add a Helm stanza, `global.acls.job` for configuring `server-acl-init` and `server-acl-init-cleanup` jobs.
- Add the Helm value `global.acls.job.tolerations` for setting tolerations on `server-acl-init` and `server-acl-init-cleanup` jobs.
- Add the Helm value `global.acls.job.nodeSelector` for setting the nodeSelector on `server-acl-init` and `server-acl-init-cleanup` jobs.
- Add the Helm value `webhookCertManager.nodeSelector` for setting the nodeSelector on `webhook-cert-manager-deployment`.
- Add BATS tests for all of the above changes.

How I've tested this PR:
- I deployed the following Helm values into an AKS account to ensure that Consul runs normally when `server-acl-init` and `webhook-cert-manager` are deployed on a particular node. There is no _anticipated_ reason why they shouldn't.

- :eyes:
- Run a Helm chart that sets `nodeSelector`.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/server-acl-init-cleanup-job.yaml
+++ b/charts/consul/templates/server-acl-init-cleanup-job.yaml
@@ -62,6 +62,14 @@ spec:
             limits:
               memory: "50Mi"
               cpu: "50m"
+          {{- if .Values.global.acls.job.tolerations }}
+          tolerations:
+            {{ tpl .Values.global.acls.job.tolerations . | indent 12 | trim }}
+          {{- end }}
+          {{- if .Values.global.acls.job.nodeSelector }}
+          nodeSelector:
+            {{ tpl .Values.global.acls.job.nodeSelector . | indent 12 | trim }}
+          {{- end }}
   {{- end }}
   {{- end }}
   {{- end }}
--- a/charts/consul/templates/server-acl-init-cleanup-job.yaml
+++ b/charts/consul/templates/server-acl-init-cleanup-job.yaml
@@ -62,14 +62,14 @@ spec:
             limits:
               memory: "50Mi"
               cpu: "50m"
-          {{- if .Values.global.acls.job.tolerations }}
-          tolerations:
-            {{ tpl .Values.global.acls.job.tolerations . | indent 12 | trim }}
-          {{- end }}
-          {{- if .Values.global.acls.job.nodeSelector }}
-          nodeSelector:
-            {{ tpl .Values.global.acls.job.nodeSelector . | indent 12 | trim }}
-          {{- end }}
+      {{- if .Values.global.acls.job.tolerations }}
+      tolerations:
+        {{ tpl .Values.global.acls.job.tolerations . | indent 12 | trim }}
+      {{- end }}
+      {{- if .Values.global.acls.job.nodeSelector }}
+      nodeSelector:
+        {{ tpl .Values.global.acls.job.nodeSelector . | indent 12 | trim }}
+      {{- end }}
   {{- end }}
   {{- end }}
   {{- end }}
--- a/charts/consul/templates/server-acl-init-cleanup-job.yaml
+++ b/charts/consul/templates/server-acl-init-cleanup-job.yaml
@@ -62,13 +62,13 @@ spec:
             limits:
               memory: "50Mi"
               cpu: "50m"
-      {{- if .Values.global.acls.job.tolerations }}
+      {{- if .Values.global.acls.tolerations }}
       tolerations:
-        {{ tpl .Values.global.acls.job.tolerations . | indent 12 | trim }}
+        {{ tpl .Values.global.acls.tolerations . | indent 12 | trim }}
       {{- end }}
-      {{- if .Values.global.acls.job.nodeSelector }}
+      {{- if .Values.global.acls.nodeSelector }}
       nodeSelector:
-        {{ tpl .Values.global.acls.job.nodeSelector . | indent 12 | trim }}
+        {{ tpl .Values.global.acls.nodeSelector . | indent 12 | trim }}
       {{- end }}
   {{- end }}
   {{- end }}
--- a/charts/consul/templates/server-acl-init-cleanup-job.yaml
+++ b/charts/consul/templates/server-acl-init-cleanup-job.yaml
@@ -64,11 +64,11 @@ spec:
               cpu: "50m"
       {{- if .Values.global.acls.tolerations }}
       tolerations:
-        {{ tpl .Values.global.acls.tolerations . | indent 12 | trim }}
+        {{ tpl .Values.global.acls.tolerations . | indent 8 | trim }}
       {{- end }}
       {{- if .Values.global.acls.nodeSelector }}
       nodeSelector:
-        {{ tpl .Values.global.acls.nodeSelector . | indent 12 | trim }}
+        {{ tpl .Values.global.acls.nodeSelector . | indent 8 | trim }}
       {{- end }}
   {{- end }}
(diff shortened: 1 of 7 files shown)


charts/consul/values.yaml
```diff
@@ -515,6 +515,25 @@ global:
       # @type: string
       secretKey: null
 
+    # job configures settings for the server-acl-init and server-acl-cleanup jobs
+    job:
```
t-eckert, line 519: This seemed like the least funky place to add this functionality, but I definitely recognize the funkiness and am happy to change it.

ishustava: Could we remove the `job` nesting to make it consistent with other components? e.g. `server.nodeSelector` or `connectInject.nodeSelector`

ishustava (commented):
Looks good, just one comment about helm values.

charts/consul/test/unit/server-acl-init-cleanup-job.bats
```diff
@@ -70,3 +70,48 @@ load _helpers
...
+  local actual=$(helm template \
+      -s templates/server-acl-init-job.yaml  \
+      --set 'global.acls.manageSystemACLs=true' \
+      --set 'global.acls.tolerations=- key: value' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.tolerations[0].key' | tee /dev/stderr)
+  [ "${actual}" = "value" ]
+}
+
+@test "serverACLInit/Job: nodeSelector not set by default" {
```
ishustava, line 98:
```suggestion
@test "serverACLInitCleanup/Job: nodeSelector not set by default" {
```

charts/consul/test/unit/server-acl-init-cleanup-job.bats
```diff
@@ -70,3 +70,48 @@ load _helpers
...
   [ "${actual}" = "true" ]
 }
+
+#--------------------------------------------------------------------
+# global.acls.tolerations and global.acls.nodeSelector
+
+@test "serverACLInitCleanup/Job: tolerations not set by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-acl-init-job.yaml  \
```
ishustava, line 80: These should be testing `server-acl-init-cleanup` template files

charts/consul/test/unit/server-acl-init-cleanup-job.bats
```diff
@@ -70,3 +70,48 @@ load _helpers
...
+      --set 'global.acls.manageSystemACLs=true' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.nodeSelector' | tee /dev/stderr)
+  [ "${actual}" = "null" ]
+}
+
+@test "serverACLInit/Job: nodeSelector can be set" {
```
ishustava, line 108: Also needs to test a different template file
```suggestion
@test "serverACLInitCleanup/Job: nodeSelector can be set" {
```

charts/consul/templates/server-acl-init-job.yaml
```diff
@@ -336,6 +336,14 @@ spec:
             limits:
               memory: "50Mi"
               cpu: "50m"
+      {{- if .Values.global.acls.tolerations }}
+      tolerations:
+        {{ tpl .Values.global.acls.tolerations . | indent 12 | trim }}
```
ishustava, line 341: It doesn't look like there's 12 spaces here 👀 ? Are we indenting correctly?

charts/consul/values.yaml
```diff
@@ -515,6 +515,23 @@ global:
       # @type: string
       secretKey: null
 
+    # tolerations configures the taints and tolerations for the server-acl-init
+    # and server-acl-cleanup jobs. This should be a multi-line string matching the
```
ishustava, line 519:
```suggestion
    # and server-acl-init-cleanup jobs. This should be a multi-line string matching the
```

charts/consul/values.yaml
```diff
@@ -515,6 +515,23 @@ global:
...
+    # Tolerations (https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) array in a Pod spec.
+    tolerations: ""
+
+    # This value defines `nodeSelector` (https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector)
+    # labels for the server-acl-init and server-acl-cleanup jobs pod assignment, formatted as a multi-line string.
```
ishustava, line 524:
```suggestion
    # labels for the server-acl-init and server-acl-init-cleanup jobs pod assignment, formatted as a multi-line string.
```

charts/consul/values.yaml
```diff
@@ -3043,6 +3060,19 @@ webhookCertManager:
   # @type: string
   tolerations: null
 
+  # This value defines `nodeSelector` (https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector)
+  # labels for webhook cert manager pod assignment, formatted as a multi-line string.
```
ishustava, line 3064:
```suggestion
  # labels for the webhook-cert-manager pod assignment, formatted as a multi-line string.
```

ishustava (commented):
Sorry! A few more comments. I didn't notice these in my first review

charts/consul/test/unit/server-acl-init-cleanup-job.bats
t-eckert: Oh geez, that's embarrassing. Thank you for catching the copypasta issue.

charts/consul/templates/server-acl-init-job.yaml
t-eckert: I changed it to 8. Thank you for catching that.

charts/consul/templates/server-acl-init-cleanup-job.yaml
```diff
@@ -62,6 +62,14 @@ spec:
             limits:
               memory: "50Mi"
               cpu: "50m"
+      {{- if .Values.global.acls.tolerations }}
+      tolerations:
+        {{ tpl .Values.global.acls.tolerations . | indent 12 | trim }}
```
ishustava, line 67: These also don't look like 12 spaces. Why aren't our tests failing? Maybe this is not necessary?

ishustava (approved):

charts/consul/templates/server-acl-init-cleanup-job.yaml
t-eckert: We do need it, I will change it to 8. The reason our tests don't fail is that YAML only cares about the level of indent exceeding the current indent.

I tested this by changing the value to 2 and watching it fail because the next indent level was less than the "tolerations:" indent level.

YAML doesn't see the difference between these two:
```yaml
pizzas:
  - one: pepperoni
  - two: cheese
```

```yaml
pizzas:
                   - one: pepperoni
                   - two: cheese
```

That's why the 12 works here. But is is better as 8.

ishustava: Ah that's interesting! Thanks for looking into it!

PR #1583, opened 2022-10-04
set consul-k8s-image directly through consul-k8s-image parameter for consul nightly tests instead of setting it through additional flags.

jmurret:
Background:
Ci for nightly-acceptance-tests-consul are failing because it is using a current control-plane that no longer has consul-api-timeout specified for inject-connector-deployment.

This fixes, version mismatches as you can see here the failed tests go from 12 to 6 (there are still 3more failed tests/subtests that have individual reasons for failing that need to be investigated individually.)

Changes proposed in this PR:
- set consul-k8s-image directly through consul-k8s-image parameter for consul nightly tests instead of setting it through additional flags.
-

How I've tested this PR:
CI

👀

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1279,13 +1279,13 @@ workflows:
 #            - acceptance-aks-1-22
 
   nightly-acceptance-tests-consul:
-    # triggers:
-    #   - schedule:
-    #       cron: "0 0 * * *"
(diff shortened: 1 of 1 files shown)


kschoche (approved):

curtbushko (approved):
LGTM!

PR #1589, opened 2022-10-05
fix(sync-catalog): modify syncCatalog deployment to work with agentless & externalServers

pree:
Changes proposed in this PR:
- syncCatalog wasn't able to be deployed when using externalServers & useSystemRoots

Error:
```
Error: INSTALLATION FAILED: Deployment.apps "consul-sync-catalog" is invalid: [spec.template.spec.containers[0].volumeMounts[1].name: Not found: "consul-ca-cert", spec.template.spec.initContainers[0].volumeMounts[1].name: Not found: "consul-auto-encrypt-ca-cert"]
```

How I've tested this PR:
Deployed using HCP and my fork on EKS.

Deploy syncCatalog on a agentless setup using externalServers (HCP for example)

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


ishustava:
Hey @pree 

Thanks so much for the PR! We'll need to do more work to have sync-catalog properly support agentless and external servers, and that is coming in the GA release. It will need to be integrated with the `consul-server-connection-manager` library, and the changes you made as is won't quite work because sync-catalog deployment doesn't know how to read those environment variables yet. 

We'll be making these changes soon and will use your work so far as the base, but for now I'll close this PR because it needs more work that is currently in flight.

PR #1591, opened 2022-10-06
configure -tls-server-name when global.cloud.enabled=true so that it matches the server certificate minted from HCP

jmurret:
Changes proposed in this PR:
- configure -tls-server-name for consul-k8s components that use connection manager when global.cloud.enabled = true so that it matches the server certificate minted from HCP 
-

How I've tested this PR:
- bats tests
- manual integration testing with HCP

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/mesh-gateway-deployment.yaml
+++ b/charts/consul/templates/mesh-gateway-deployment.yaml
@@ -213,6 +213,8 @@ spec:
             {{- end }}
             {{- if and .Values.externalServers.enabled .Values.externalServers.tlsServerName }}
             -tls-server-name={{.Values.externalServers.tlsServerName }} \
+            {{- else if .Values.global.cloud.enabled }}
+            -consul-tls-server-name=server.{{ .Values.global.datacenter}}.{{ .Values.global.domain}} \
             {{- end }}
             {{- else }}
             -tls-disabled \
--- a/charts/consul/templates/mesh-gateway-deployment.yaml
+++ b/charts/consul/templates/mesh-gateway-deployment.yaml
@@ -214,7 +214,7 @@ spec:
             {{- if and .Values.externalServers.enabled .Values.externalServers.tlsServerName }}
             -tls-server-name={{.Values.externalServers.tlsServerName }} \
             {{- else if .Values.global.cloud.enabled }}
-            -consul-tls-server-name=server.{{ .Values.global.datacenter}}.{{ .Values.global.domain}} \
+            -tls-server-name=server.{{ .Values.global.datacenter}}.{{ .Values.global.domain}} \
             {{- end }}
             {{- else }}
             -tls-disabled \
--- a/cli/version/version.go
+++ b/cli/version/version.go
@@ -19,7 +19,7 @@ var (
 	// A pre-release marker for the version. If this is "" (empty string)
 	// then it means that it is a final release. Otherwise, this is a pre-release
 	// such as "dev" (in development), "beta", "rc1", etc.
-	VersionPrerelease = "dev"
+	VersionPrerelease = "beta1"
 )
 
 // GetHumanVersion composes the parts of the version in a way that's suitable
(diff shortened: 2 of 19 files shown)


ndhanushkodi (approved):
Maybe at some point we could standardize the flag name to either -tls-server-name or -consul-tls-server-name across all the commands but that's definitely not blocking for this fix!

curtbushko (approved):

charts/consul/templates/mesh-gateway-deployment.yaml
```diff
@@ -213,6 +213,8 @@ spec:
             {{- end }}
             {{- if and .Values.externalServers.enabled .Values.externalServers.tlsServerName }}
             -tls-server-name={{.Values.externalServers.tlsServerName }} \
```
thisisnotashwin, line 215: Should this be `consul-tls-server-name` as well?

thisisnotashwin (approved):
As Nitya suggested, it is a little unclear where we use `consul-tls-server-name` and `tls-server-name` and we should put something in the backlog to clear that up.

charts/consul/templates/mesh-gateway-deployment.yaml
kschoche: I think this should be `-tls-server-name`  for `consul-dataplane` and `-consul-tls-server-name` for other commands.
We should probably clear this up though.

charts/consul/templates/mesh-gateway-deployment.yaml
```diff
@@ -213,6 +213,8 @@ spec:
...
+            {{- else if .Values.global.cloud.enabled }}
+            -consul-tls-server-name=server.{{ .Values.global.datacenter}}.{{ .Values.global.domain}} \
```
kschoche, line 217:
```suggestion
            -tls-server-name=server.{{ .Values.global.datacenter}}.{{ .Values.global.domain}} \
```
I believe this is `-tls-server-name` ?

kschoche (changes requested):
I'm not sure the flags are all being passed correctly, happy to pair and take a look with you, but feel free to ignore if I'm actually wrong :)

One other thing, non-blocking, since this is repeated verbatim in so many places, should we move it into a template in `_helpers.tpl`?

```
                {{- if and $root.Values.externalServers.enabled $root.Values.externalServers.tlsServerName }}
                -tls-server-name={{$root.Values.externalServers.tlsServerName }} \
                {{- else if $root.Values.global.cloud.enabled }}
                -consul-tls-server-name=server.{{ $root.Values.global.datacenter}}.{{ $root.Values.global.domain}} \
```

charts/consul/templates/mesh-gateway-deployment.yaml
jmurret: I just updated this to use tls-server-name everywhere.  I think differentiating for dataplane vs other components is more of an implementation detail, where this setting in every case is meant to set the TLSConfig.

charts/consul/templates/mesh-gateway-deployment.yaml
jmurret: this has been fixed.

control-plane/version/version.go
```diff
@@ -19,7 +19,7 @@ var (
 	// A pre-release marker for the version. If this is "" (empty string)
 	// then it means that it is a final release. Otherwise, this is a pre-release
 	// such as "dev" (in development), "beta", "rc1", etc.
-	VersionPrerelease = "dev"
+	VersionPrerelease = "beta1"
```
thisisnotashwin, line 22: Is this correct? I think this should still be `dev`?

cli/version/version.go
```diff
@@ -19,7 +19,7 @@ var (
 	// A pre-release marker for the version. If this is "" (empty string)
 	// then it means that it is a final release. Otherwise, this is a pre-release
 	// such as "dev" (in development), "beta", "rc1", etc.
-	VersionPrerelease = "dev"
+	VersionPrerelease = "beta1"
```
thisisnotashwin, line 22: this should be `dev` too

PR #1593, opened 2022-10-06
consul-k8s CLI - hcp self-managed clusters - fix the required environment variables that should just default to production settings

jmurret:
Changes proposed in this PR:
The following envvars can be passed to the consul-k8s CLI so they get used in the cluster creation and bootstrapping:

HCP_CLIENT_ID (required)

HCP_CLIENT_SECRET (required)

HCP_AUTH_URL

HCP_SCADA_ADDRESS

HCP_API_HOST

All of the ones that are not marked as required were exposed by the hcp-sdk so that we could test in other environments besides production.  When they are not provided, hcp sdk will default them to prod settings, and they are truly optional.

There is currently a bug in consul-k8s CLI where when these are not provided to consul-k8s CLI, they still get set on the server pod as empty strings.  The hcp sdk then recognizes that they are there and says they cannot be empty strings.

How I've tested this PR:
- unit tests
- building the CLI locally and using this to install an hcp self managed cluster without providing the optional envvars.

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/cli/preset/preset.go
+++ b/cli/preset/preset.go
@@ -2,12 +2,19 @@ package preset
 
 import (
 	"fmt"
+	"os"
 )
 
 const (
 	PresetSecure     = "secure"
 	PresetQuickstart = "quickstart"
 	PresetCloud      = "cloud"
+
+	EnvHCPClientID     = "HCP_CLIENT_ID"
+	EnvHCPClientSecret = "HCP_CLIENT_SECRET"
+	EnvHCPAuthURL      = "HCP_AUTH_URL"
+	EnvHCPAPIHost      = "HCP_API_HOST"
+	EnvHCPScadaAddress = "HCP_SCADA_ADDRESS"
 )
 
 // Presets is a list of all the available presets for use with CLI's install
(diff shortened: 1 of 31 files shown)


cli/preset/preset.go
```diff
@@ -42,3 +49,36 @@ func GetPreset(config *GetPresetConfig) (Preset, error) {
...
+	if clientID, ok := os.LookupEnv(EnvHCPClientID); ok {
+		hcpConfig.ClientID = clientID
+	}
+
+	// Read clientSecret from environment
+	if clientSecret, ok := os.LookupEnv(EnvHCPClientSecret); ok {
+		hcpConfig.ClientSecret = clientSecret
+	}
+
+	// Read clientID from environment
```
kschoche, line 68:
```suggestion
	// Read authURL from environment
```

kschoche (approved):
Looks great! I left a couple non-blocking comments!

t-eckert (approved):

PR #1595, opened 2022-10-06
partitions: Deprecate partition service in favor of expose server service

ndhanushkodi:
Changes proposed in this PR:
- Breaking change for 1.14 GA to remove the partition service and instead use the expose servers service for setting up partitions rather than using 2 separate services. The expose servers service was added for peering, but would be the recommended way to expose any ports on the Consul servers over a K8s service moving forward. (The partition service was originally used for server discovery across partitions).

Step 4 in [these docs](https://www.consul.io/docs/enterprise/admin-partitions#install-the-consul-server-cluster) are the relevant steps for setting up admin partitions using the partition service. However, that step doesn't need to change, since the expose-servers service will be found via the same labels described in step 4.

How I've tested this PR:
- Acceptance tests (the partitions tests still pass)

- 👀 

Checklist:
- [x] CHANGELOG entry added 


thisisnotashwin:
I think we also need to make changes to our documentation and _maybe_ our learn guides to ensure people use the correct service to read the loadbalancer IP from.

thisisnotashwin (approved):
nice work!!

jmurret (approved):
🎉 :shipit:

PR #1596, opened 2022-10-06
peering: update exported services CRD and bump api module

ndhanushkodi:
Changes proposed in this PR:
- fix the peering acceptance test by updating the exported services CRD to use Peer instead of PeerName to match the config entry
- update the consul api module
- makes a unit test fix with the api module bump-- duplicate server names are no longer allowed across 2 peers

How I've tested this PR:
- unit and acceptance tests

- 👀 

Checklist:
- [ ] CHANGELOG entry added 


control-plane/connect-inject/peering_dialer_controller_test.go
```diff
@@ -251,7 +251,12 @@ func TestReconcile_CreateUpdatePeeringDialer(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 
 			// Create test consul server.
-			acceptorPeerServer, err := testutil.NewTestServerConfigT(t, nil)
```
ndhanushkodi, line 254: This change was due to a failure after updating the api module.

thisisnotashwin: I think there are a few more tests that need a similar update made to their fixtures.

jmurret (approved):
👍 thank you for finding and fixing this!

thisisnotashwin (approved):
Excellent!!!

t-eckert (approved):
Great find!

PR #1600, opened 2022-10-07
Add flag to disable dns tests on AKS

curtbushko:
The consul dns tests have been flaky ever since we upgraded AKS from 1.21.14 to 1.22.11 on August 24th. I have been able to reproduce the problem on a new AKS cluster and one that has sat around for a while thus eliminating the chance that this was caused by something kube not starting.

Sometimes the tests pass, sometimes they fail and I do not know why.

Research:

- The problem still happens on newer versions of AKS (1.23.12)
- I could not find anything in the Kubernetes or AKS release notes/CHANGELOGs
- I have not been able to find any issues in Github for both Kubernetes and AKS related to the problem.
- Added `sh -c` to the dig command so that the TTY wasn't swallowed but it didn't make a difference. Reverted because it is probably not a good idea to output networking things in tests.

Changes proposed in this PR:
- Add a flag to the acceptance framework for `use-aks`
- Disable the dns test when config.UseAKS is set
- Add the flag to the acceptance tests in CircleCI

How I've tested this PR:

- Ran the tests locally against an AKS cluster with the flag set or not

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -829,7 +829,7 @@ jobs:
       - run: mkdir -p $TEST_RESULTS
 
       - run-acceptance-tests:
-          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig" -consul-image=hashicorppreview/consul-enterprise:1.14-dev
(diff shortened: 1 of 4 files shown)


thisisnotashwin (approved):
Thanks for doing this!! 🙏

kschoche (approved):

PR #1603, opened 2022-10-07
Show server statuses only when they are deployed into the Kubernetes cluster

t-eckert:
Previously, if a user had servers deployed outside of a given K8s cluster, the `status` command would fail when it attempted to show the health status of servers. We recently removed a similar feature for clients which had the same problem when we moved to the "agentless" architecture. 

This PR changes the status command to only print the health status of clients and servers if they are expected to be present in the K8s cluster. If the user is running with the agentless architecture or has servers deployed outside of the cluster (e.g. HCP). The `status` command will simply not show the health status instead of failing.

Changes proposed in this PR:
- Only show health status for servers if they are expected to be present in the K8s cluster.
- Move some constants and vars up to the top of a file where I think it makes more sense for them to be. 
- Fix any tests broken as a result of this change.

<details>
  <summary>Screenshots</summary>
  <strong>Agentless status with unhealthy servers</strong>
  no-clients-yes-servers-unhealthy

  <strong>Agentless status with healthy servers</strong>
  no-clients-yes-servers-healthy

  <strong>0.48.0 unhealthy clients and servers</strong>
  clients-and-servers-unhealthy

  <strong>0.48.0 unhealthy clients and healthy servers</strong>
  clients-and-servers-semi-healthy

  <strong>0.48.0 healthy clients a
  clients-and-servers-healthy
nd servers</strong>

</details>

How I've tested this PR:
- Unit tests for the CLI cover this change.
- Deployed Consul to a Kind cluster and ran the status command both when the clients/servers were not ready and as they came online.

- Running unit tests to check the behavior is expected.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/cli/cmd/status/status.go
+++ b/cli/cmd/status/status.go
@@ -120,7 +120,7 @@ func (c *Command) Run(args []string) int {
 	}
 
 	if err := c.checkConsulAgents(namespace); err != nil {
-		c.UI.Output("Unable to check Kubernetes cluster for Consul clients: %v", err)
+		c.UI.Output("Unable to check Kubernetes cluster for Consul agents: %v", err)
 		return 1
 	}
 
--- a/cli/cmd/status/status.go
+++ b/cli/cmd/status/status.go
@@ -119,11 +119,9 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 
-	if s, err := c.checkConsulServers(namespace); err != nil {
-		c.UI.Output(err.Error(), terminal.WithErrorStyle())
+	if err := c.checkConsulAgents(namespace); err != nil {
+		c.UI.Output("Unable to check Kubernetes cluster for Consul clients: %v", err)
 		return 1
-	} else {
-		c.UI.Output(s, terminal.WithSuccessStyle())
 	}
 
 	return 0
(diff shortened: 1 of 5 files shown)


cli/cmd/status/status.go
```diff
@@ -216,24 +214,39 @@ func validEvent(events []release.HookEvent) bool {
...
+	clients, err := c.kubernetes.AppsV1().DaemonSets(namespace).List(c.Ctx, metav1.ListOptions{LabelSelector: "app=consul,chart=consul-helm"})
 	if err != nil {
-		return "", err
-	} else if len(servers.Items) == 0 {
-		return "", errors.New("no server stateful set found")
-	} else if len(servers.Items) > 1 {
-		return "", errors.New("found multiple server stateful sets")
+		return err
+	}
+	if len(clients.Items) != 0 {
```
jmurret, line 226: minor nit.  you could make this a shared func for this and below and pass in:
- agentType string (servers vs clients.  also used for string replacement in the output)
- desiredAgents int
- readyAgents int

cli/cmd/status/status_test.go
```diff
@@ -28,39 +28,89 @@ import (
...
+			// Before deployment, we shouldn't see the checks or errors.
+			err := c.checkConsulAgents(namespace)
+			require.NoError(t, err)
+
+			actual := buf.String()
+			require.NotContains(t, actual, "Consul Clients Healthy")
+			require.NotContains(t, actual, "Consul Servers Healthy")
+			buf.Reset()
+
+			if tc.clients != 0 {
```
jmurret, line 59: minor nit - might be something to think about for the next time.

this test would probably be easier to read if these were just part of the test cases.  I find it easier to read a state test that sets up and asserts one state, so the description and set up show it.  This rather shows the initial state, sets it up, asserts it, changes it, asserts it again.

So instead of these descriptions:
```
		"No clients, no agents",
		"3 clients, no agents",
		"3 clients, 3 agents"},
		"No clients, 3 agents",
```
you would have all the permutations of the states shown in this 2 state template:
```
		"No clients, no agents",
		"3 clients - 1 healthy, no agents",
		"3 clients - all healthy, no agents",
		"3 clients - 1 healthy, 3 agents - 1 healthy",
		"3 clients - all healthy, 3 agents - all healthy",
		"No clients, 3 agents - 1 healthy",
		"No clients, 3 agents - all healthy",

**Review by jmurret — APPROVED** [review]
Looks good.  Nice work!  One suggestion on some text and then minor observations.

File: cli/cmd/status/status.go (hunk 1 above)
Comment (t-eckert) [reply cli/cmd/status/status.go:226]: I appreciate this feedback! I did think about doing that to "DRY" out this function. My thinking with the current way it is written is that one day we will nix the clients check and I don't want an extra function floating around. I think it's nice to just delete the specific lines and not have to change anything else. Once those lines are gone, there won't be any more deduping.

What do you think of that? Am I overthinking it?

File: cli/cmd/status/status_test.go (hunk 2 above)
Comment (t-eckert) [reply cli/cmd/status/status_test.go]: I love that feedback. I implemented it and it cleans up a lot. Thank you!

File: cli/cmd/status/status_test.go [hunk 3]
```diff
@@ -28,39 +28,51 @@ import (
...
+		desiredServers int
+		healthyServers int
+	}{
+		"No clients, no agents":                       {0, 0, 0, 0},
+		"3 clients - 1 healthy, no agents":            {3, 1, 0, 0},
+		"3 clients - 3 healthy, no agents":            {3, 3, 0, 0},
+		"3 clients - 1 healthy, 3 agents - 1 healthy": {3, 1, 3, 1},
+		"3 clients - 3 healthy, 3 agents - 3 healthy": {3, 3, 3, 3},
+		"No clients, 3 agents - 1 healthy":            {0, 0, 3, 1},
+		"No clients, 3 agents - 3 healthy":            {0, 0, 3, 3},
```
kschoche, line 45: I think we're using confusing terminology here, it should be `clients` and `servers` since `agents` means both :)

kschoche (approved):
Looks great thanks for this.

PR #1608, opened 2022-10-10
set up individual secrets for what used to be HCPConfig secret

jmurret:
*NOTE: At first glance this looks huge, but its really because the same sets of bats tests and helm failure logic are in multiple files to check that for each secret, if one of the `secretName` or the `secretKey` is set, then both are set.  Besides this, this actually a very small change.*

Motivation:
- the question of whether or not each of these environment variables needs to be individually configured in the helm chart was first posed by Iryna in my original PR
- there are 6 six environment variables that canbe set for the hcpsdk.  3 are required and 3 are optional.  in order to make the setting of these environment variables optional in server-statefulset, they need to be individual helm values so that the system can check whether they are nil or not.
- this change replaces what https://github.com/hashicorp/consul-k8s/pull/1593 tried to do bu did notwork (described below).
- the current method of trying to allow the  environment variable to read a null key from a secret without any conditional logic such as below errors on the pod during startup and therefor still causes the user to set all 6 environment variables when using the CLI.
```         - name: HCP_SCADA_ADDRESS
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.global.cloud.secretName }}
                  key: scada-address 
```

Changes proposed in this PR:
- make all 6 of the environment variable their own configurable helm chart value
- conditionally set the 3 optional ones in server-statefulset based on the presence of a value
- conditionally set them in the CLI preset values for cloud preset based on the presence of a value when the user runs the CLI
- add helm check for the 3 required values when `global.cloud.enabled=true`
- add helm check for all 6 six secrets so that when either the secretName or secretValue is provided, both are provided

How I've tested this PR:
- unit / acceptance tests
- manually 

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


charts/consul/templates/_helpers.tpl
```diff
@@ -371,13 +371,50 @@ Consul server environment variables for consul-k8s commands.
 {{- end -}}
 
 {{/*
-Fails global.cloud.enabled is true and global.cloud.secretName is nil or tempty.
+Fails global.cloud.enabled is true and one of the following secrets is nil or empty.
+- global.cloud.resourceId.secretName
+- global.cloud.clientId.secretName
+- global.cloud.clientSecret.secretName
 
-Usage: {{ template "consul.validateCloudConfiguration" . }}
```
jmurret, line 376: changing the name since a 6 secrets used to be stored as keys in on "HCPConfig" secret.

charts/consul/templates/_helpers.tpl
```diff
@@ -371,13 +371,50 @@ Consul server environment variables for consul-k8s commands.
...
+Usage: {{ template "consul.validaterequiredCloudSecretsExist" . }}
```
jmurret, line 379: changed name.  this used to be one secret with 6 six keys (3 required and 3 optional).  Now there are 6 secrets for more flexibility and the ability to allow optional values.

charts/consul/templates/_helpers.tpl
```diff
@@ -371,13 +371,50 @@ Consul server environment variables for consul-k8s commands.
...
-{{ end }}
+{{- define "consul.validaterequiredCloudSecretsExist" -}}
+{{- if (and .Values.global.cloud.enabled (or (not .Values.global.cloud.resourceId.secretName) (not .Values.global.cloud.clientId.secretName) (not .Values.global.cloud.clientSecret.secretName))) }}
+{{fail "When global.cloud.enabled is true, global.cloud.resourceId.secretName, global.cloud.clientId.secretName, and global.cloud.clientSecret.secretName must also be set."}}
+{{- end }}
+{{- end -}}
+
+{{/*
+Fails global.cloud.enabled is true and one of the following secrets has either an empty secretName or secretKey.
+- global.cloud.resourceId.secretName / secretKey
```
jmurret, line 390: This simply checks all 6 secrets so that if either `secretName` or `secretKey` are provided, then both are provided.

charts/consul/templates/server-statefulset.yaml
```diff
@@ -254,42 +255,54 @@ spec:
                   name: {{ .Values.global.acls.replicationToken.secretName | quote }}
                   key: {{ .Values.global.acls.replicationToken.secretKey | quote }}
             {{- end }}
-            {{- if and .Values.global.cloud.enabled .Values.global.cloud.secretName }}
+            {{- if and .Values.global.cloud.enabled}}
```
jmurret, line 258: changed so the check for the presence of a value is on each individual secret below.

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -2,7 +2,8 @@
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true for api gateway" }}{{ end }}
```
jmurret, line 2: this will be repeated in all below helm files below, with `server-statefulset.yaml` also having a true logic change.

charts/consul/test/unit/api-gateway-controller-deployment.bats
```diff
@@ -908,7 +908,7 @@ load _helpers
 #--------------------------------------------------------------------
 # global.cloud
 
-@test "apiGateway/Deployment: fails when global.cloud.enabled is set and global.cloud.secretName is not set" {
+@test "apiGateway/Deployment: fails when global.cloud.enabled is set and global.cloud.clientId.secretName is not set but global.cloud.clientSecret.secretName and global.cloud.resourceId.secretName is set" {
```
jmurret, line 911: these `fail` tests will be repeated in all below bats files below, with `server-statefulset.bats` also having a true logic change.

cli/preset/cloud_preset.go
```diff
@@ -179,7 +200,18 @@ global:
...
+    resourceId:
+      secretName: %s
+      secretKey: %s
+    clientId:
+      secretName: %s
+      secretKey: %s
+    clientSecret:
+      secretName: %s
+      secretKey: %s
+    %s
```
jmurret, line 212: here are the placeholders for the optional configs.  when they are not set, the call to `getOptionalSecretFromHCPConfig` will return an empty string.

cli/preset/cloud_preset.go
```diff
@@ -188,57 +220,139 @@ connectInject:
...
+	}
+	if err := i.saveSecret(secretNameHCPClientSecret, data, corev1.SecretTypeOpaque); err != nil {
 		return err
 	}
-	i.UI.Output(fmt.Sprintf("HCP config saved in '%s' secret in namespace '%s'.",
-		secretNameHCPConfig, i.KubernetesNamespace), terminal.WithSuccessStyle())
+	i.UI.Output(fmt.Sprintf("HCP client secret saved in '%s' secret in namespace '%s'.",
+		secretKeyHCPClientSecret, i.KubernetesNamespace), terminal.WithSuccessStyle())
 
-	if err := i.saveBootstrapTokenSecret(config); err != nil {
```
jmurret, line 218: got rid of the small helper functions like `saveBootstrapTokenSecret()` because they really did not do much and I moved everything in line so its quicker to read and know that they are all behaving the same way.

charts/consul/templates/_helpers.tpl
t-eckert, line 379: Nit: 
```suggestion
Usage: {{ template "consul.validateRequiredCloudSecretsExist" . }}
```

charts/consul/templates/_helpers.tpl
t-eckert: Neat! TIL this syntax.

charts/consul/templates/server-statefulset.yaml
t-eckert: Can we cut the `and` then?

charts/consul/test/unit/api-gateway-controller-deployment.bats
t-eckert: Just a suggestion to clarify `true`/`false` vs `set`/`not set`. I feel the former is easier to read in a long line like this.

```suggestion
@test "apiGateway/Deployment: fails when global.cloud.enabled is true and global.cloud.clientId.secretName is false but global.cloud.clientSecret.secretName and global.cloud.resourceId.secretName are both true" {
```

t-eckert:
```suggestion
@test "apiGateway/Deployment: fails when global.cloud.enabled is true and global.cloud.clientId.secretName is not set but global.cloud.clientSecret.secretName and global.cloud.resourceId.secretName is set" {
```

charts/consul/test/unit/api-gateway-controller-deployment.bats
```diff
@@ -919,8 +919,226 @@ load _helpers
...
+      --set 'global.cloud.clientId.secretKey=client-id-key' \
+      --set 'global.cloud.clientSecret.secretName=client-secret-id-name' \
+      --set 'global.cloud.clientSecret.secretKey=client-secret-id-key' \
+      --set 'global.cloud.resourceId.secretName=resource-id-name' \
+      --set 'global.cloud.resourceId.secretKey=resource-id-key' \
+      --set 'global.cloud.authUrl.secretName=auth-url-name' \
+      .
+
+  [ "$status" -eq 1 ]
+  echo "$output"
```
t-eckert, line 1017: I think you can remove these `echo "$output"` lines. Were they just here for debugging?

jmurret: 🙏  this has been corrected.

charts/consul/test/unit/api-gateway-controller-deployment.bats
jmurret: 🙏  this has been corrected.

charts/consul/test/unit/api-gateway-controller-deployment.bats
t-eckert: Thanks!

t-eckert (approved):
Great work!

cli/preset/cloud_preset.go
```diff
@@ -21,19 +21,24 @@ import (
...
-	secretNameServerCert     = "consul-server-cert"
-	secretKeyHCPClientID     = "client-id"
-	secretKeyHCPClientSecret = "client-secret"
-	secretKeyHCPResourceID   = "resource-id"
-	secretKeyHCPAuthURL      = "auth-url"
-	secretKeyHCPAPIHostname  = "api-hostname"
-	secretKeyHCPScadaAddress = "scada-address"
-	secretKeyGossipKey       = "key"
-	secretKeyBootstrapToken  = "token"
+	secretNameHCPClientID     = "consul-hcp-client-id"
```
curtbushko, line 24: Nit: I would have organized these by name/key pairs so that it is easier to tell if any of them are missing.

charts/consul/values.yaml
```diff
@@ -654,11 +654,59 @@ global:
     # self-managed cluster.
     enabled: false
     
-    # The name of the Kubernetes secret that holds the HCP cloud configuration.
-    # It contains the HCP service principal client_id and client_secret as well
-    # as the HCP resource_id.
-    # @type: string
-    secretName: null
+    # The name of the Kubernetes secret that holds the HCP resource id.
```
curtbushko, line 657: Is there a way to indicate which of these are required and which are optional?

cli/preset/cloud_preset.go
```diff
@@ -188,57 +220,139 @@ connectInject:
...
 		return err
 	}
 
-	i.UI.Output(fmt.Sprintf("Saving HCP configuration as secrets in %s namespace", i.KubernetesNamespace), terminal.WithHeaderStyle())
-	if err := i.saveServerHCPConfigSecret(config); err != nil {
+	// HCP resource id
+	data := map[string][]byte{
+		secretKeyHCPResourceID: []byte(config.HCPConfig.ResourceID),
+	}
+	if err := i.saveSecret(secretNameHCPResourceID, data, corev1.SecretTypeOpaque); err != nil {
```
curtbushko, line 258: Nit: I've seen in a lot of our code that we use a lower case letter to indicate the interface. For example, `i *CloudPreset` becomes `c *CloudPreset`

cli/preset/cloud_preset.go
```diff
@@ -188,57 +220,139 @@ connectInject:
...
 func (i *CloudPreset) saveSecretsFromBootstrapConfig(config *CloudBootstrapConfig) error {
+	// create namespace
 	if err := i.createNamespaceIfNotExists(); err != nil {
 		return err
 	}
 
-	i.UI.Output(fmt.Sprintf("Saving HCP configuration as secrets in %s namespace", i.KubernetesNamespace), terminal.WithHeaderStyle())
-	if err := i.saveServerHCPConfigSecret(config); err != nil {
+	// HCP resource id
+	data := map[string][]byte{
```
curtbushko, line 255: Even though these names/keys are not optional, should you also check to make sure they are not empty before trying to save them? (non-blocking)

charts/consul/templates/_helpers.tpl
```diff
@@ -371,13 +371,50 @@ Consul server environment variables for consul-k8s commands.
...
 
-Usage: {{ template "consul.validateCloudConfiguration" . }}
+Usage: {{ template "consul.validateRequiredCloudSecretsExist" . }}
 
 */}}
-{{- define "consul.validateCloudConfiguration" -}}
-{{- if and .Values.global.cloud.enabled (not .Values.global.cloud.secretName) }}
-{{fail "When global.cloud.enabled is true, global.cloud.secretName must also be set."}}
-{{ end }}
+{{- define "consul.validateRequiredCloudSecretsExist" -}}
```
curtbushko, line 382: I like that this is a template and isn't repeated several times!

charts/consul/templates/api-gateway-controller-deployment.yaml
curtbushko: I don't know if you need the check and bats tests for every file. 

If the settings are incorrect, helm will catch it right away in `server-statefulset.yaml` which makes `connect-inject-deployment.yaml`, etc redundant. 

Maybe keep it in the gateways as those can be deployed separately?

curtbushko (changes requested):
Looks pretty good! I have a few questions/suggestions but nothing in particular that I would block on.

cli/preset/cloud_preset.go
jmurret: done.

charts/consul/test/unit/server-acl-init-job.bats
```diff
@@ -1890,21 +1890,199 @@ load _helpers
...
+      --set 'global.cloud.clientId.secretKey=client-id-key' \
+      --set 'global.cloud.clientSecret.secretName=client-secret-id-name' \
+      --set 'global.cloud.clientSecret.secretKey=client-secret-id-key' \
+      --set 'global.cloud.resourceId.secretName=resource-id-name' \
+      --set 'global.cloud.resourceId.secretKey=resource-id-key' \
+      --set 'global.cloud.authUrl.secretKey=auth-url-key' \
+      .
+
+  [ "$status" -eq 1 ]
+
```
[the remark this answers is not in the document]
jmurret: thank you.   removed.

charts/consul/values.yaml
jmurret: done.  thank you for the suggestion!

charts/consul/templates/api-gateway-controller-deployment.yaml
jmurret: i spoke about this repetition thing several months ago.  its hard to get around it because of the conditional installation of each file.  For example, if a kubernetes cluster is installed with only clients/components and server.enabled is false, then it will not pick up these settings.

curtbushko: Sounds good!

curtbushko (approved):

PR #1610, opened 2022-10-11
peering: support peering establishment over mesh gateways, remove ability to pass external addresses to token generation endpoint

ndhanushkodi:
Changes proposed in this PR:

In Consul 1.13, the way to establish a peering across 2 different networks was to use the expose-servers service to make the Consul servers reachable for peering establishment, and pass the externally reachable addresses of that service into the generate-token endpoint. This way, the peering token would contain these externally reachable server addresses rather than the Consul server pod ips. 

Now, with support for peering establishment over mesh gateways, we can use the mesh gateways to reach the servers instead. So this PR removes support entirely for using the expose-servers service for peering. However, the service itself should remain since that will be used for admin partitions. 

Now, the peering test uses mesh gateways for peering establishment by configuring the mesh config entry. 

How I've tested this PR:

Ran the test locally, and verified via logs that peerings are being established both through the local and remote mesh gateways. 

Ran the acceptance tests against all the clouds, passing build: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/8588/workflows/8fdab6f0-0fc6-48d4-abbc-5d86535ce55b

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


charts/consul/templates/mesh-gateway-deployment.yaml
```diff
@@ -237,7 +237,7 @@ spec:
             {{- if .Values.global.adminPartitions.enabled }}
             -service-partition={{ .Values.global.adminPartitions.name }} \
             {{- end }}
-            -log-level={{ default .Values.global.logLevel }} \
+            -log-level="trace" \
```
ndhanushkodi, line 240: will remove

acceptance/framework/k8s/deploy.go
```diff
@@ -96,7 +96,7 @@ func CheckStaticServerConnectionMultipleFailureMessages(t *testing.T, options *k
 		expectedOutput = expectedSuccessOutput
 	}
 
-	retrier := &retry.Timer{Timeout: 160 * time.Second, Wait: 2 * time.Second}
+	retrier := &retry.Timer{Timeout: 320 * time.Second, Wait: 2 * time.Second}
```
ndhanushkodi, line 99: This was because I noticed that by the time the peering is established and the exported service is actually replicated to the other side, this curl was timing out. It is taking a bit longer for the peering to get set up-- things that cause this are dialing a follower node and needing to retry the peering, a delta xds bug that causes the mesh gateways to take extra time to get their config.

acceptance/tests/peering/peering_connect_test.go
```diff
@@ -53,10 +53,14 @@ func TestPeering_Connect(t *testing.T) {
 
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
+			fmt.Println("starting test")
```
kschoche, line 56: replace with t.Log?

acceptance/tests/peering/peering_connect_test.go
```diff
@@ -90,14 +95,13 @@ func TestPeering_Connect(t *testing.T) {
 				staticServerPeerHelmValues["server.exposeGossipAndRPCPorts"] = "true"
 				staticServerPeerHelmValues["meshGateway.service.type"] = "NodePort"
 				staticServerPeerHelmValues["meshGateway.service.nodePort"] = "30100"
-				staticServerPeerHelmValues["server.exposeService.type"] = "NodePort"
```
kschoche, line 93: QQ: why are we removing the grpc port configs for each of these?

acceptance/tests/peering/peering_connect_test.go
```diff
@@ -138,9 +157,13 @@ func TestPeering_Connect(t *testing.T) {
 				require.NotEmpty(r, acceptorSecretName)
 			})
 
+			fmt.Println("copying secret from client peer to server peer, printing contents:")
 			// Copy secret from client peer to server peer.
 			k8s.CopySecret(t, staticClientPeerClusterContext, staticServerPeerClusterContext, "api-token")
+			resp, _ := k8s.RunKubectlAndGetOutputE(t, staticClientPeerClusterContext.KubectlOptions(t), "secret", "api-token", "-o", "yaml")
+			fmt.Println(resp)
```
kschoche, line 164: Might want to check the syntax on this one.. since we're not using `resp` anywhere but the println?
```suggestion
			k8s.RunKubectl(t, staticClientPeerClusterContext.KubectlOptions(t), "secret", "api-token", "-o", "yaml")
```

charts/consul/templates/mesh-gateway-deployment.yaml
```diff
@@ -237,7 +237,7 @@ spec:
             {{- if .Values.global.adminPartitions.enabled }}
             -service-partition={{ .Values.global.adminPartitions.name }} \
             {{- end }}
-            -log-level={{ default .Values.global.logLevel }} \
```
kschoche, line 240: was this intentional?

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 ## UNRELEASED
 
 FEATURES:
-* Peering: Add support for `PeerThroughMeshGateways` in Mesh CRD. [[GH-1478](https://github.com/hashicorp/consul-k8s/pull/1478)]
+* Peering:
+  * Add support for `PeerThroughMeshGateways` in Mesh CRD. [[GH-1478](https://github.com/hashicorp/consul-k8s/pull/1478)]
+  * move support for customizing the server addresses in peering token generation. Instead, mesh gateways should be used
```
curtbushko, line 6: I am not sure but should the token changes from the helm chart be listed as a breaking change?

curtbushko (changes requested):
Looks fantastic! I just have a few comments but nothing breaking/blocking!

charts/consul/templates/server-config-configmap.yaml
```diff
@@ -27,7 +27,12 @@ data:
       "data_dir": "/consul/data",
       "domain": "{{ .Values.global.domain }}",
       "ports": {
+        {{- if not .Values.global.tls.enabled }}
         "grpc": 8502,
+        {{- end }}
+        {{- if .Values.global.tls.enabled }}
+        "grpc_tls": 8502,
+        {{- end }}
```
kschoche, line 35: I found this the other day and thought I'd share, no need to change it! :D 
```suggestion
        {{ ternary "\"grpc_tls\":" "\"grpc\":" .Values.global.tls.enabled }} 8502,
```

charts/consul/templates/mesh-gateway-deployment.yaml
curtbushko: Nitya mentioned above that she will remove it before merge. I have not idea why it didn't show up in the diffs for review. 🤷

kschoche: weird!

kschoche (approved):
This looks great! I have not had a chance to test it yet but assuming we're 🍏 everywhere I don't have any blocking comments beyond seconding Curt's comment about probably marking this as a breaking change in the changelog!

curtbushko (approved):

acceptance/tests/peering/peering_connect_test.go
ndhanushkodi: We no longer need the exposeService at all for peering connections, so this removes any configuration of it.

charts/consul/templates/server-config-configmap.yaml
ndhanushkodi: oh interesting, thats a slick one! I'll skip that change for now since I personally find the existing format a little easier to read!

acceptance/tests/peering/peering_connect_test.go
ndhanushkodi: Removed these lines instead, they actually weren't working

acceptance/tests/peering/peering_connect_test.go
```diff
@@ -53,10 +53,14 @@ func TestPeering_Connect(t *testing.T) {
 
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
+			fmt.Println("starting test")
 			staticServerPeerClusterContext := env.DefaultContext(t)
 			staticClientPeerClusterContext := env.Context(t, environment.SecondaryContextName)
 
 			commonHelmValues := map[string]string{
+				"global.imageK8S": "ndhanushkodi/consul-k8s-dev:pmgw1",
```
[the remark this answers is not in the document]
ndhanushkodi: Thanks!

CHANGELOG.md
ndhanushkodi: Yup added it to breaking instead!

acceptance/tests/peering/peering_connect_test.go
ndhanushkodi: whoops thanks i forgot to remove those, I was just using those locally so i could see what was going on while running the test!

PR #1619, opened 2022-10-13
remove deprecated annotations in server-service template

kschoche:
Changes proposed in this PR:
- We previously relied on setting an annotation ` service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"` which has been deprecated for 5 years and replaced with `spec.publicNotReadyAddresses` which we also already use.
- Remove it and continue to use the correct spec field on the server-service.yaml.

https://github.com/kubernetes/kubernetes/pull/63742
https://github.com/kubernetes/kubernetes/issues/63741

How I've tested this PR:
CI and Unit tests should pass.
Bats test updated.

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+* Helm:
+  * Remove use of deprecated annotation `service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"` in server-service template. [[GH-1619](https://github.com/hashicorp/consul-k8s/pull/1619)]
+
 ## 1.0.0-beta3 (October 12, 2022)
 
 FEATURES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 IMPROVEMENTS:
 * Helm:
-  * Remove use of deprecated annotation `service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"` in server-service template. [[GH-1619](https://github.com/hashicorp/consul-k8s/pull/1619)]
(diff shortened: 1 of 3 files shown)


charts/consul/test/unit/server-service.bats
```diff
@@ -103,15 +98,6 @@ load _helpers
...
 # annotations
 
-@test "server/Service: one annotation by default" {
-  cd `chart_dir`
-  local actual=$(helm template \
-      -s templates/server-service.yaml  \
-      . | tee /dev/stderr |
-      yq -r '.metadata.annotations | length' | tee /dev/stderr)
-  [ "${actual}" = "1" ]
-}
```
ishustava, line 113: Could this be changed to `no annotation by default` case?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+* Helm:
+  * Remove use of deprecated annotation `service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"` in server-service template. [[GH-1619](https://github.com/hashicorp/consul-k8s/pull/1619)]
```
ishustava, line 5:
```suggestion
* Helm:
  * Remove deprecated annotation `service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"` in the `server-service` template. [[GH-1619](https://github.com/hashicorp/consul-k8s/pull/1619)]
```

ishustava (approved):
Looks good! Just a couple of minor comments

thisisnotashwin (approved):
NICE!

charts/consul/test/unit/server-service.bats
kschoche: Totes!

PR #1623, opened 2022-10-17
Delete Consul custom resources with `consul-k8s uninstall`

t-eckert:
## Changes proposed in this PR
- All custom resources managed by Consul will be fetched before Helm uninstall is run.
- These custom resources will be sent a deletion request with a timeout of 1 second and 5 retries if the custom resources are not deleted. 
- If custom resources remain after these 5 attempts, their finalizers will be patched to be an empty list, freeing them up to be deleted by Kubernetes.
- Changed timeout to be a duration var instead of a string that gets parsed to a duration.
- Fixed the "header" when installing.

## How I've tested this PR
- Unit tests to cover CR fetching, deletion, and patching.
- Manually tested that CRs are successfully deleted in the following scenarios:
  - Consul deployed with controller.
  - Consul deployed with controller, all Consul servers deleted (making CR deletion impossible) and falling back to patching finalizers to get rid of CRs
  - Consul deployed without controller (but CRs exist from an earlier installation). Deletion of CRs fails without a controller and uninstall falls back to patching finalizers.

There is a case that this does not cover, but I feel it is too much of an edge case to go after. That is, the user deploys Consul with a controller, the user manually deletes the controller deployment. The user does not delete the `mutatingwebhookconfiguration` for the controller. This will cause the uninstall request to fail. If the user deletes `mutatingwebhookconfiguration` for the controller, uninstall will succeed. 

A good, simple run manual test is

Deploy Consul with this configuration:
```yaml
  global:
    name: consul
    logLevel: debug
  controller:
    enabled: true
  server:
    replicas: 1
  ```

Then create a resource that will create a custom resource:
```yaml
# server.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: server
spec:
  replicas: 3
  selector:
    matchLabels:
      app: server
  template:
    metadata:
      name: server
      labels:
        app: server
      annotations:
    spec:
      containers:
        - name: server
          image: docker.mirror.hashicorp.services/hashicorp/http-echo:latest
          args:
            - -text="hello world"
            - -listen=:8080
          ports:
            - containerPort: 8080
              name: http
          resources:
            requests:
              memory: "64Mi"
              cpu: "250m"
            limits:
              memory: "128Mi"
              cpu: "500m"
      serviceAccountName: server
      terminationGracePeriodSeconds: 0
---
apiVersion: v1
kind: Service
metadata:
  name: server
spec:
  selector:
    app: server
  ports:
    - name: http
      port: 80
      targetPort: 8080
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: server
---
kind: ServiceDefaults
metadata:
  name: server
spec:
  protocol: http
```

Then perform `consul-k8s uninstall` using this branch's CLI.

## Checklist
- [x] Tests added
- [x] CHANGELOG entry added 


cli/cmd/uninstall/uninstall.go
```diff
@@ -117,125 +123,112 @@ func (c *Command) init() {
...
 	if c.helmActionsRunner == nil {
 		c.helmActionsRunner = &helm.ActionRunner{}
 	}
 
 	if err := c.set.Parse(args); err != nil {
-		c.UI.Output(err.Error(), terminal.WithErrorStyle())
-		return 1
-	}
-	if len(c.set.Args()) > 0 {
-		c.UI.Output("Should have no non-flag arguments.", terminal.WithErrorStyle())
```
jmurret, line 140: if you are changing this here, you will also need to know that the other commands have similar output messages.  so we either need to change them as well or remove this from the PR.  I think any UX messages we probably want to be more explicit and forward thinking where a designer / product person is helping dictate these changes so that they are consistent and done in isolation.  Otherwise, each of us engineers will just refactor messages here and there and make our already inconsistent UX more inconsistent.  IMO, I would remove most of the discretionary changes you have in here related to output messages that are not related to the deleting of CRs/CRDs and make a separate PR.

cli/cmd/uninstall/uninstall.go
```diff
@@ -277,81 +283,190 @@ func (c *Command) Run(args []string) int {
...
-		c.UI.Output(err.Error(), terminal.WithErrorStyle())
-		return 1
+	if c.flagKubeContext != "" {
+		settings.KubeContext = c.flagKubeContext
 	}
 
-	if err := c.deleteClusterRoles(foundReleaseName); err != nil {
-		c.UI.Output(err.Error(), terminal.WithErrorStyle())
-		return 1
+	restConfig, err := settings.RESTClientGetter().ToRESTConfig()
```
jmurret, line 357: nit - could move this into `if c.kubernetes == nil` block below

cli/cmd/uninstall/uninstall.go
```diff
@@ -264,7 +257,20 @@ func (c *Command) Run(args []string) int {
...
-			Prompt: fmt.Sprintf("WARNING: Proceed with deleting PVCs, Secrets, Service Accounts, Roles, Role Bindings, Jobs, Cluster Roles, and Cluster Role Bindings for the following installation? \n\n   Name: %s \n   Namespace: %s \n\n   Only approve if all data from this installation can be deleted. (y/N)", foundReleaseName, foundReleaseNamespace),
+			Prompt: fmt.Sprintf(`WARNING: Proceed with deleting 
+	- Persistent Volume Claims
+	- Secrets 
+	- Service Accounts 
+	- Roles 
+	- Role Bindings
+	- Jobs
+	- Cluster Roles 
+	- Cluster Role Bindings
```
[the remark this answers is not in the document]
t-eckert: No, we decided that CRs will always be deleted, not as part of `-wipe-data`. I also removed this improvement for a later PR.

cli/cmd/uninstall/uninstall.go
```diff
@@ -45,7 +51,10 @@ type Command struct {
 
 	helmActionsRunner helm.HelmActionsRunner
 
+	// Configuration for interacting with Kubernetes.
```
t-eckert, line 54: I needed to add the dynamic client for interacting with custom resources as unstructured objects and the apiextensions client for interacting with custom resource definitions.

cli/cmd/uninstall/uninstall.go
```diff
@@ -160,20 +169,8 @@ func (c *Command) Run(args []string) int {
...
-		if err != nil {
-			c.UI.Output("retrieving Kubernetes auth: %v", err, terminal.WithErrorStyle())
-			return 1
-		}
-		c.kubernetes, err = kubernetes.NewForConfig(restConfig)
-		if err != nil {
-			c.UI.Output("initializing Kubernetes client: %v", err, terminal.WithErrorStyle())
-			return 1
-		}
+	if err := c.initKubernetes(settings); err != nil {
```
t-eckert, line 172: With the additional clients, I wanted to centralize kubernetes initialization work to a method as has been done in other commands.

cli/cmd/uninstall/uninstall.go
```diff
@@ -361,10 +427,106 @@ func (c *Command) uninstallHelmRelease(releaseName, namespace, releaseType strin
...
+			Group:    group,
+			Version:  version,
+			Resource: strings.ToLower(cr.GetKind()),
+		}
+
+		err := c.dynamic.
+			Resource(target).
+			Namespace(cr.GetNamespace()).
+			Delete(c.Ctx, cr.GetName(), metav1.DeleteOptions{})
+		if common.IgnoreNotFoundError(err) != nil {
```
t-eckert, line 488: We ignore the not found error here to handle the case where a previous deletion request has deleted the CR, but the CR was still present when the last `fetchCustomResources` function call was made. As an example:

```
Initial CRs [A, B, C]
CR Deletion Attempt 1 {
    Fetch 1 returns [A, B, C]
    Deletion request sent to [A, B, C]
    A successfully deleted.
    B successfully deleted.
    Fetch 2 returns [C]
}
CR Deletion Attempt 2 {
    Fetch 1 returns [C]
    C successfully deleted (from deletion request in attempt 1).
    Deletion request sent to [C].
    C cannot be found (this is ignored).
    Fetch 2 returns []
}
Deletion attempts complete as Fetch 2 returned [].
```

cli/cmd/uninstall/uninstall.go
```diff
@@ -361,10 +427,106 @@ func (c *Command) uninstallHelmRelease(releaseName, namespace, releaseType strin
...
+			Group:    group,
+			Version:  version,
+			Resource: strings.ToLower(cr.GetKind()),
+		}
+
+		_, err := c.dynamic.
+			Resource(target).
+			Namespace(cr.GetNamespace()).
+			Patch(c.Ctx, cr.GetName(), types.JSONPatchType, finalizerPatch, metav1.PatchOptions{})
+		if common.IgnoreNotFoundError(err) != nil {
```
t-eckert, line 522: Similar to the other ignore case, we will get here after 5 deletion attempts have failed to remove all custom resources. This can mean that there is a deadlock that makes the deletion impossible, or that the deletion is just taking a long time to complete. In the latter case, the deletion request may succeed before the finalizer patch request is sent, but after this function starts executing. So we may be trying to patch a custom resource that no longer exists and don't want to fail the uninstall task.

cli/cmd/install/install.go
```diff
@@ -401,7 +401,7 @@ func (c *Command) installConsul(valuesYaml []byte, vals map[string]interface{},
 	if len(vals) == 0 {
 		c.UI.Output("\nNo overrides provided, using the default Helm values.", terminal.WithInfoStyle())
 	} else {
-		c.UI.Output("\nHelm value overrides\n-------------------\n"+string(valuesYaml), terminal.WithInfoStyle())
+		c.UI.Output("\nHelm value overrides\n--------------------\n"+string(valuesYaml), terminal.WithInfoStyle())
```
t-eckert, line 404: This fixes a little annoyance. Before it would print:
```
Helm value overrides
-------------------
```

cli/cmd/uninstall/uninstall.go
```diff
@@ -345,6 +374,53 @@ func (c *Command) uninstallHelmRelease(releaseName, namespace, releaseType strin
...
 
+	// Delete any custom resources managed by Consul. If they cannot be deleted,
+	// patch the finalizers to be empty on each one.
+	if releaseType == common.ReleaseTypeConsul {
+		uiLogger("Deleting custom resources managed by Consul")
+		crds, err := c.fetchCustomResourceDefinitions()
+		if err != nil {
+			return fmt.Errorf("unable to fetch Custom Resource Definitions for Consul deployment: %v", err)
+		}
+		kindToResource := mapCRKindToResourceName(crds)
```
t-eckert, line 385: CR patching and deletion searches for the custom resources by their resource name. But the CR itself only knows its Kind. So I make a map of Kind to resource name using the CRDs which know both and pass it to the patching and deletion functions so they can find the resources.

t-eckert:
@ishustava and @jmurret, please ignore the GH actions acceptance test failures. They are due to an issue where the pipeline is using the wrong version of Go (1.17) to build the CLI binary. The acceptance tests running in CircleCI are passing: https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/8835/workflows/6d5f1da4-55dd-47f0-8f7d-8aadbecfc33f

cli/cmd/uninstall/uninstall_test.go
```diff
@@ -414,8 +434,165 @@ func TestTaskCreateCommand_AutocompleteArgs(t *testing.T) {
...
+	require.NoError(t, err)
+
+	crds, err := c.fetchCustomResourceDefinitions()
+	require.NoError(t, err)
+
+	actual, err := c.fetchCustomResources(crds)
+	require.NoError(t, err)
+	require.Len(t, actual, 1)
+	require.Contains(t, actual, cr)
+	require.NotContains(t, actual, nonConsulCR)
```
jmurret, line 474: 👏

cli/cmd/uninstall/uninstall_test.go
```diff
@@ -432,7 +609,8 @@ func TestUninstall(t *testing.T) {
 			messages: []string{
 				"\n==> Checking if Consul demo application can be uninstalled\n    No existing Consul demo application installation found.\n",
 				"\n==> Checking if Consul can be uninstalled\n ✓ Existing Consul installation found.\n",
-				"\n==> Consul Uninstall Summary\n    Name: consul\n    Namespace: consul\n ✓ Successfully uninstalled Consul Helm release.\n ✓ Skipping deleting PVCs, secrets, and service accounts.\n",
+				"\n==> Consul Uninstall Summary\n    Name: consul\n    Namespace: consul",
```
jmurret, line 612: is there a reason to split this?  these were intentionally grouped based on `\n==>` so that formatting was ensured.

jmurret: also effects all of the other places where this change was made.

jmurret (commented):
looks good.  just a question on the changing of the tests that no longer ensures that test appears within sections and formatted the same.

cli/cmd/uninstall/uninstall_test.go
t-eckert: Great question. I split these up for tests the output now prints that the CRs were deleted. It will print that between the uninstall summary and the success messages. For example, for this test, it prints:

```

==> Checking if Consul demo application can be uninstalled
    No existing Consul demo application installation found.
    No existing Consul demo application installation found.

==> Checking if Consul can be uninstalled
  ✓ Existing Consul installation found.

==> Consul Uninstall Summary
    Name: consul
    Namespace: consul
--> Deleting custom resources managed by Consul
  ✓ Successfully uninstalled Consul Helm release.
  ✓ Skipping deleting PVCs, secrets, and service accounts.
```

jmurret: So why are taking a test that matched exact output of a section and making it less precise where output that is not visible in the test could be introduced?

jmurret: also, why is `Deleting custom resources managed by Consul` not included in this output test?  Can we change this so that:
- the test still validates the entire section in one string
- the new sentence is a part of that string

this would be more precise and follow the current pattern that was put in place because the output was inconsistent between different subcommands and mostly due to the fact that it was not under test where you could explicitly see what the expected sections and their contents would be.

t-eckert: I see what you mean. I added a check for that output, thanks!

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ FEATURES:
 IMPROVEMENTS:
 * CLI
   * Update minimum go version for project to 1.19 [[GH-1633](https://github.com/hashicorp/consul-k8s/pull/1633)]
+  * `consul-k8s uninstall` now deletes custom resources when uninstalling Consul. [[GH-1623](https://github.com/hashicorp/consul-k8s/pull/1623)]
```
jmurret, line 25: [nit] I think we have a style guide on this someone where, but I think these things are supposed to start with `Enable .....` or `Add support ....`.  I would take a look at our other entries and possibly revise this to feel like the others.

jmurret (approved):
nice work.  I have some nits and just the mention of those two remaining uninstall tests that should validate the new output.  approving and will let you figure that all out.  🎉

cli/cmd/uninstall/uninstall.go
```diff
@@ -345,13 +367,60 @@ func (c *Command) uninstallHelmRelease(releaseName, namespace, releaseType strin
...
+				return err
+			}
+
+			crs, err = c.fetchCustomResources(crds)
+			if err != nil {
+				return err
+			}
+			if len(crs) != 0 {
+				return common.NewDeletionError(fmt.Sprintf("%d custom resources remain after deletion request. Retrying deletion", len(crs)))
+			}
```
ishustava, line 398: 🤔 I think this is the only thing that needs to be in the retry block, right? We don't need to retry deletion, we just need to give some time to check that they are deleted if I remember correctly. I suppose it would work either way, but just wondering if there's some reason why we needed to do that.

cli/cmd/uninstall/uninstall.go
```diff
@@ -345,13 +367,60 @@ func (c *Command) uninstallHelmRelease(releaseName, namespace, releaseType strin
...
+		if common.IsDeletionError(err) {
+			uiLogger("Patching finalizers on custom resources managed by Consul")
+			crs, err := c.fetchCustomResources(crds)
+			if err != nil {
+				return err
+			}
+
+			if err = c.patchCustomResources(crs, kindToResource); err != nil {
+				return err
+			}
```
ishustava, line 411: After we patch CRs, we should have another retry block to check that they are deleted properly for completeness

cli/cmd/uninstall/uninstall.go
```diff
@@ -361,10 +430,105 @@ func (c *Command) uninstallHelmRelease(releaseName, namespace, releaseType strin
...
+
+// deleteCustomResources takes a list of unstructured custom resources and
+// sends a request to each one to be deleted.
+func (c *Command) deleteCustomResources(crs []unstructured.Unstructured, kindToResource map[string]string) error {
+	for _, cr := range crs {
+		apiVersion := strings.Split(cr.GetAPIVersion(), "/")
+		group, version := apiVersion[0], apiVersion[1]
+		if group == "" || version == "" {
+			return fmt.Errorf("malformed api version: %s", apiVersion)
+		}
```
ishustava, line 478: I think you can use `schema.ParseGroupVersion` from k8s `apimachinery` package

cli/cmd/uninstall/uninstall.go
ishustava: You could use `k8serrors.IsNotFound` instead

cli/cmd/uninstall/uninstall_test.go
ishustava: +1 great job on test! It'd be nice to add another CR in a different k8s namespace but not a blocker.

cli/cmd/uninstall/uninstall.go
```diff
@@ -361,10 +430,105 @@ func (c *Command) uninstallHelmRelease(releaseName, namespace, releaseType strin
...
+	}
+
+	return crs, nil
+}
+
+// deleteCustomResources takes a list of unstructured custom resources and
+// sends a request to each one to be deleted.
+func (c *Command) deleteCustomResources(crs []unstructured.Unstructured, kindToResource map[string]string) error {
+	for _, cr := range crs {
```
ishustava, line 473: I don't think I saw UI logging for deletion. It'd be nice to tell a user that we're deleting the CRs unless I missed these logs somewhere

cli/common/utils.go
```diff
@@ -81,3 +82,12 @@ func IsValidLabel(label string) bool {
 
 	return true
 }
+
+// IgnoreNotFoundError is a convenience function which takes an error and will
+// return it unless it represents a NotFound, 404 error.
+func IgnoreNotFoundError(err error) error {
+	if statusError, ok := err.(*errors.StatusError); ok && statusError.ErrStatus.Code == 404 {
+		return nil
+	}
```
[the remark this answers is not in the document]
ishustava: +1 I think we should use the `k8serrors` pkg instead

ishustava (commented):
Great work on this, Thomas! I'm leaving some comments in-line, but I don't think they are too significant. Good work!

cli/cmd/uninstall/uninstall.go
t-eckert: Oh nice! I didn't know about that!

cli/cmd/uninstall/uninstall.go
t-eckert: The logging could be easy to miss. It's [on line 373](https://github.com/hashicorp/consul-k8s/blob/1f36db0ed79545d2ea8fcb1dfb972d492d61b8b2/cli/cmd/uninstall/uninstall.go#L373). I used the logger that gets passed to Helm because the "Deleting custom resources managed by Consul" info seems to mesh nicely with the other logs directly from Helm (e.g. "Deleting PVCs for..."). Because I wanted them to look the same and be grouped together, I used that logger so that in the future, if we change how that looks, the "Deleting custom resources managed by Consul" logging changes with it.

ishustava: I meant that we should also log the individual resources that get deleted. I thought we do it for other resources like PVCs etc

cli/cmd/uninstall/uninstall_test.go
t-eckert: Thank you! I added the CR in another namespace

cli/cmd/uninstall/uninstall.go
t-eckert: I'm now using the `k8serrors.IsNotFound` instead and deleted the solution I wrote.

CHANGELOG.md
t-eckert: Changed to "Enable `consul-k8s uninstall` to delete custom resources when uninstalling Consul. This is done by default. [[GH-1623](https://github.com/hashicorp/consul-k8s/pull/1623)]"

cli/cmd/uninstall/uninstall.go
t-eckert: Yes!

cli/cmd/uninstall/uninstall.go
t-eckert: You are totally right. I had this idea that maybe if the CRs change after the initial deletion request, this would help with that, but this request moves so quickly, I don't think it's a case we will reasonably move into. It's also not really how people tend to use CRs unless they have an operator that is creating CRs for them.

cli/cmd/uninstall/uninstall.go
t-eckert: Added! It will print out an error if there is a situation where the CRs cannot be deleted.

cli/cmd/uninstall/uninstall.go
ishustava: Ah perfect!

cli/cmd/uninstall/uninstall.go
```diff
@@ -361,10 +390,164 @@ func (c *Command) uninstallHelmRelease(releaseName, namespace, releaseType strin
...
+		return err
+	}
+
+	crs, err = c.fetchCustomResources(crds)
+	if err != nil {
+		return err
+	}
+	if len(crs) != 0 {
+		return fmt.Errorf("unable to remove all custom resources managed by Consul. %d custom resources remain and will need to be removed manually", len(crs))
+	}
```
ishustava, line 452: This should be in a retry block as well in case it doesn't happen right away.

ishustava (approved):
Great work!

PR #1625, opened 2022-10-17
agentless: enable transparent proxy when running within a single k8s cluster

ishustava:
Changes proposed in this PR:
- Make `connect-init` command apply traffic redirection rules instead of calling consul CLI. This allows us to get rid of the init container that copies the consul binary.
  - The `connect-init` command now accepts `iptables.Config` as JSON string in  a flag or env variable, parses it and calls `iptables.Setup` with that config
  - It also applies additional rules based on proxy configuration (partially adopted from consul's `redirect-traffic` command implementation)
- The webhook now provides the iptables configuration as json to the `connect-init` command at injection times. The rules are based on the work we've done as part of the Consul CNI plugin.

How I've tested this PR:
acceptance and unit tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,7 +123,6 @@ commands:
                     if ! gotestsum --no-summary=all --jsonfile=jsonfile-${pkg////-} -- $pkg -p 1 -timeout 2h -failfast \
                           << parameters.additional-flags >> \
                           -enable-multi-cluster \
-                          ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -debug-directory="$TEST_RESULTS/debug" \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,6 +123,7 @@ commands:
                     if ! gotestsum --no-summary=all --jsonfile=jsonfile-${pkg////-} -- $pkg -p 1 -timeout 2h -failfast \
                           << parameters.additional-flags >> \
                           -enable-multi-cluster \
+                          ${ENABLE_ENTERPRISE:+-enable-enterprise} \
                           -debug-directory="$TEST_RESULTS/debug" \
                           -consul-k8s-image=<< parameters.consul-k8s-image >>
                     then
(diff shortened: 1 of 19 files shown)


acceptance/tests/basic/basic_test.go
```diff
@@ -50,6 +50,7 @@ func TestBasicInstallation(t *testing.T) {
 				"global.tls.enabled":                   strconv.FormatBool(c.secure),
 				"global.gossipEncryption.autoGenerate": strconv.FormatBool(c.secure),
 				"global.tls.enableAutoEncrypt":         strconv.FormatBool(c.autoEncrypt),
+				"client.enabled":                       "true",
```
ishustava, line 53: Clients are now disabled by default but still supported. We can use this test to check that they still work.

control-plane/connect-inject/container_init_test.go
```diff
@@ -340,156 +248,43 @@ func TestHandlerContainerInit_transparentProxy(t *testing.T) {
...
-			container, err := w.containerInit(ns, *pod, multiPortInfo{})
-			require.NoError(t, err)
-			actualCmd := strings.Join(container.Command, " ")
-
-			require.Contains(t, actualCmd, c.expectedContainsCmd)
-		})
-	}
-}
-
-func TestHandler_constructDNSServiceHostName(t *testing.T) {
```
ishustava, line 470: this test has been moved to the `redirect_traffic_test.go` because that's where this logic now lives

.circleci/config.yml
```diff
@@ -521,7 +521,7 @@ jobs:
       - run: mkdir -p $TEST_RESULTS
       - run-acceptance-tests:
           failfast: true
-          additional-flags: -use-kind -kubecontext="kind-dc1" -secondary-kubecontext="kind-dc2" -consul-image=hashicorppreview/consul-enterprise:1.14-dev
+          additional-flags: -use-kind -kubecontext="kind-dc1" -secondary-kubecontext="kind-dc2" -consul-image=ishustava/consul:latest@sha256:3985a25200ef3f593c75ffbbbadc9fcc3a1d07da4aafb2222ab65e9e893d468a
```
ishustava, line 524: I'll remove this change once hashicorp/consul#15025 is merged

PR #1629, opened 2022-10-18
[wip] Pod Security Standard support

kschoche:
Changes proposed in this PR:
- Adds support for specifying `global.podSecurityStandards.securityContext` which applies Pod Security Standards compliant `securityContexts` to all Consul container resources, including the sidecars that are attached by the webhook. 

There is limited compatibility for `tproxy + psa` in that the CNI plugin is required if the `PSA` policy restricts containers from having `root` which we require for iptables commands when CNI is not used.

How I've tested this PR:
* Manually by deploying a Kind Kubernetes-1.25 cluster with PSA enabled.
* Acceptance tests should pass without PSA enabled
* Unit tests added/updated

```
demo $ cat values.yaml
global:
  enabled: false
  podSecurityStandards:
    securityContext:
      runAsNonRoot: true
      runAsUser: 100
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      seccompProfile:
        type: RuntimeDefault
server:
  enabled: true
  replicas: 1
```
 Enable `restricted:1.24` PSA:
 ```shell
 $ kubectl label --overwrite ns default \
 ```
Deploy using the above helm values.

NOTES: 
* It looks like to enable Openshift 1.25 support will require us to not use the `global.openshift.enabled` field.
* Transparent Proxy support without CNI is unlikely to work with PSA enabled due to most PSS policies enforcing non-root users, we will need to document this somehow.

👀 and check out the UX.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/server-statefulset.yaml
```diff
@@ -124,7 +124,9 @@ spec:
       serviceAccountName: {{ template "consul.fullname" . }}-server
       {{- if not .Values.global.openshift.enabled }}
       securityContext:
-        {{- toYaml .Values.server.securityContext | nindent 8 }}
+        {{- $tmp := dict -}}
```
kschoche, line 127: `merge` will fill `$tmp` with the merged results, if we don't setup a tmp variable it will overwrite the original data (`.Values.global.podSecurityStandards.securityContext`) which gets used elsewhere.

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -33,13 +33,23 @@ func TestConnectInject(t *testing.T) {
 			cfg := suite.Config()
 			ctx := suite.Environment().DefaultContext(t)
 
+			helmvals := map[string]string{
```
kschoche, line 36: The Acceptance test changes will be backed out before merging.

PR #1643, opened 2022-10-21
PR GitHub Actions Acceptance Tests

wilkermichael:
Changes proposed in this PR:
- This PR adds Github actions acceptance tests for new PRs
- I've added more detailed changes to each commit message

How I've tested this PR:
- Ran pipelines

👀
**Please go through this PR commit by commit, it'll be much easier than looking at all the files changed at once**

Checklist:
- [x] Tests added
- [N/A] CHANGELOG entry added 

TODO: all instances of local reusable workflows (i.e. ./.github/workflows/reusable-acceptance.yml) need to be changed back to read from main only before this is merged in

### DIFF
--- a/.github/workflows/reusable-acceptance.yml
+++ b/.github/workflows/reusable-acceptance.yml
@@ -11,9 +11,8 @@ on:
         type: string
         default: ""
       consul-k8s-image:
-        required: false 
+        required: false
         type: string
-        default: docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:latest
       directory:
         required: true
         type: string
@@ -26,7 +25,11 @@ on:
       kind-version:
         required: false
         type: string
-        default: "v1.22.4"
+        default: "v1.24.6"
+      checkout-ref:
+        required: false
+        type: string
+        default: ${{ github.sha }}
     secrets:
         CONSUL_ENT_LICENSE:
           required: true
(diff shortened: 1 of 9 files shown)


.github/workflows/reusable-acceptance.yml
```diff
@@ -26,40 +25,50 @@ on:
...
-          - {runner: "2", test-packages: "mesh-gateway metrics"}
-          - {runner: "3", test-packages: "partitions sync terminating-gateway"}
-          - {runner: "4", test-packages: "vault"}
+        include:
+          - {runner: "0", test-packages: "basic"}
+          - {runner: "0", test-packages: "basic consul-dns"}
+          - {runner: "1", test-packages: "connect"}
+          - {runner: "2", test-packages: "controller example"}
+          - {runner: "3", test-packages: "ingress-gateway"}
+          - {runner: "4", test-packages: "metrics partitions"}
```
ishustava, line 54: `partitions` also take a while to run so prob better to not group it with anything else. `metrics could run with `basic` I think

.github/workflows/reusable-acceptance.yml
```diff
@@ -26,40 +25,50 @@ on:
...
       matrix:
-        include: # I am really sorry for this but I could not find a way to automatically split our tests into several runners. For now, split manually.
-          - {runner: "0", test-packages: "basic connect consul-dns"}
-          - {runner: "1", test-packages: "controller example ingress-gateway"}
-          - {runner: "2", test-packages: "mesh-gateway metrics"}
-          - {runner: "3", test-packages: "partitions sync terminating-gateway"}
-          - {runner: "4", test-packages: "vault"}
+        include:
+          - {runner: "0", test-packages: "basic"}
+          - {runner: "0", test-packages: "basic consul-dns"}
```
ishustava, line 50: Is this  a duplicate?

.github/workflows/reusable-acceptance.yml
```diff
@@ -26,40 +25,50 @@ on:
...
+          - {runner: "1", test-packages: "connect"}
+          - {runner: "2", test-packages: "controller example"}
+          - {runner: "3", test-packages: "ingress-gateway"}
+          - {runner: "4", test-packages: "metrics partitions"}
+          - {runner: "5", test-packages: "peering"}
+          - {runner: "6", test-packages: "snapshot-agent sync terminating-gateway vault wan-federation"}
 
-      fail-fast: true
+      fail-fast: false
```
ishustava, line 58: I think we'd want to fail-fast, right?

.github/workflows/test.yml
```diff
@@ -207,120 +219,167 @@ jobs:
...
+    uses: ./.github/workflows/reusable-acceptance.yml
+    with:
+      name: acceptance
+      directory: acceptance/tests
+      go-version: ${{ needs.get-go-version.outputs.go-version }}
+      additional-flags: "-use-kind -kubecontext=kind-dc1 -secondary-kubecontext=kind-dc2 -consul-image=hashicorppreview/consul-enterprise:1.14-dev"
+      gotestsum-version: 1.8.2
+      consul-k8s-image: docker.io/hashicorppreview/${{ github.event.repository.name }}-control-plane:${{ needs.get-product-version.outputs.product-version }}-pr-${{ github.sha }}
+    secrets:
+      CONSUL_ENT_LICENSE: ${{ secrets.CONSUL_ENT_LICENSE }}
```
ishustava, line 372: In circleci today we also set a directory for debug logs in case of a failure (https://github.com/hashicorp/consul-k8s/blob/5668d4c877b4abd18aa8cc86d1ccfec29e70b55d/.circleci/config.yml#L501-L502). Where would the logs be in GHA?

ishustava (commented):
Great work Michael!! This looks great. I left a few comments and questions, but once they are resolved, this is good to go!

.github/workflows/reusable-acceptance.yml
wilkermichael: oh weird, good catch

.github/workflows/reusable-acceptance.yml
wilkermichael: So if we don't fail fast, it lets all the other acceptance tests run through to completion. Which is great if the rest of your tests pass. Then you can `rerun failed` and only re-run the failed cases (i.e. rerun just the connect tests)

ishustava: Ahhh ok! That's a good point.

.github/workflows/reusable-acceptance.yml
wilkermichael: sounds good to me

.github/workflows/test.yml
wilkermichael: oh okay, good question. I'll take a look

ishustava (approved):
❤️

t-eckert (approved):
Great work, Michael!

PR #1651, opened 2022-10-26
Use HashiCorp Actions Go Build for Consul K8s CLI

t-eckert:
There is a bug in the GH Actions for running acceptance tests. The Go binary that gets set up is not the same version of Go that gets used in subsequent calls to `go`. I have a feeling this is because the downloaded one is lower on the PATH. You can see this issue occurring here: https://github.com/hashicorp/consul-k8s/actions/runs/3331265834/jobs/5510872576

Changes proposed in this PR:
- Use the HashiCorp Go Build Action to build the Consul K8s CLI

How I've tested this PR:
- I will watch it run in GH actions. I had to make a modification to tell it to not run from `main`.

- You can watch it with me.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -344,7 +344,7 @@ jobs:
 
   acceptance:
     needs: [ get-product-version, dev-upload-docker, get-go-version ]
-    uses: hashicorp/consul-k8s/.github/workflows/reusable-acceptance.yml@main
+    uses: hashicorp/consul-k8s/.github/workflows/reusable-acceptance.yml@te/use-hc-actions-build
     with:
       name: acceptance
       directory: acceptance/tests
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -344,7 +344,7 @@ jobs:
 
   acceptance:
     needs: [ get-product-version, dev-upload-docker, get-go-version ]
-    uses: hashicorp/consul-k8s/.github/workflows/reusable-acceptance.yml@main
+    uses: hashicorp/consul-k8s/.github/workflows/reusable-acceptance.yml@te/use-hc-actions-build
```
t-eckert, line 347: This will change before I merge.

t-eckert:
@wilkermichael will we not be running acceptance tests for PRs in GH actions for now? If so, I'll close this.

wilkermichael:
We'll still need this, this reusable acceptance test will be used in the nightlies as well

.github/workflows/test.yml
```diff
@@ -344,7 +344,7 @@ jobs:
 
   acceptance:
     needs: [ get-product-version, dev-upload-docker, get-go-version ]
-    uses: hashicorp/consul-k8s/.github/workflows/reusable-acceptance.yml@main
+    uses: hashicorp/consul-k8s/.github/workflows/reusable-acceptance.yml@536f034d6647fdadca8892036d0b06e28d1c9a57
```
t-eckert, line 347: Change before merging.

t-eckert:
Looks like we will not be doing acceptance tests on PRs with GH actions. Closing this PR.

PR #1662, opened 2022-10-31
Mw/circle ci stable release testing

wilkermichael:
# Changes proposed in this PR:
- This PR serves to implement the new testing strategy required by our support of stable releases
- As always, you can go through my PR commit by commit for more context
- After this PR we should support the following testing strategy
   - Nightly cloud/kind on main
   - Nightly cloud/kind on release branches (once backported)
   - PR kind tests off of every branch
   - Nightly Consul compatibility tests off of main

## What's New:
- New separate cleanup nightly, this is so that the cleanup step does not conflict with multiple running cloud steps
- New scheduled nightly that will run on release branches
- Separate scheduled nightly that will run on main branch, this is so the release branch and main branch nightlies can run at a separate cadence (maybe we'll want to run release nightlies one night a week? a "weekly"?). For now both main and release will run nightly
- Nightlies now also run the kind-tproxy acceptance tests 
- For now, all nightlies will run at 5 pm PST (12 am UTC) and cleanup will run at 5 am PST (12 pm UTC). We can re-evaluate later what the testing cadence should be.

How I've tested this PR:
- Ran nightlies at an accelerate schedule. Some tests in the clouds fail, but talked with Iryna these will be fixed in a later PR
- 👀
👀

Checklist:
- [x] Tests added
- [N/A] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,6 @@ cli-path: &cli-path cli
 acceptance-mod-path: &acceptance-mod-path acceptance
 acceptance-test-path: &acceptance-test-path acceptance/tests
 acceptance-framework-path: &acceptance-framework-path acceptance/framework
-charts-consul-path: &charts-consul-path charts/consul
 helm-gen-path: &helm-gen-path hack/helm-reference-gen
 gke-terraform-path: &gke-terraform-path charts/consul/test/terraform/gke
 eks-terraform-path: &eks-terraform-path charts/consul/test/terraform/eks
@@ -25,6 +24,10 @@ openshift-terraform-path: &openshift-terraform-path charts/consul/test/terraform
 # This image is built from test/docker/Test.dockerfile
 consul-helm-test-image: &consul-helm-test-image docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.15.0
 
+########################
+# COMMANDS
+########################
+# Commands define a sequence of steps as a map to be executed and reused in jobs
 commands:
   install-prereqs:
     steps:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -16,7 +16,6 @@ cli-path: &cli-path cli
 acceptance-mod-path: &acceptance-mod-path acceptance
 acceptance-test-path: &acceptance-test-path acceptance/tests
 acceptance-framework-path: &acceptance-framework-path acceptance/framework
-charts-consul-path: &charts-consul-path charts/consul
```
wilkermichael, line 19: this was not being used anywhere

.circleci/config.yml
```diff
@@ -1252,7 +1273,70 @@ workflows:
           context: consul-ci
           requires:
             - dev-upload-docker
-  nightly-acceptance-tests:
+
+  nightly-cleanup:
+    triggers:
+      - schedule:
+          cron: "0 12 * * *"
```
ndhanushkodi, line 1280: Could you add a comment for this cron schedule?

.circleci/config.yml
```diff
@@ -1252,7 +1273,70 @@ workflows:
...
+
+
+  nightly-acceptance-tests-release:
+    description: |
+      Tests which run on a release branch nightly. These exist separate from the main
+      acceptance tests so that they can run at their own cadence, but 
+      contains the same sequence of jobs.
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
```
ndhanushkodi, line 1298: Could you add a comment for this one?

.circleci/config.yml
```diff
@@ -1268,36 +1352,35 @@ workflows:
...
           requires:
             - acceptance-eks-1-21
       - acceptance-aks-1-22:
           requires:
-            - cleanup-azure-resources
             - dev-upload-docker
       - acceptance-aks-cni-1-22:
           requires:
             - acceptance-aks-1-22
+      - acceptance-tproxy:
```
ndhanushkodi, line 1375: how come we're adding acceptance-tproxy here? I think the cloud tests run with tproxy enabled right? Plus PRs run this step, so maybe we don't need it in the nightlies?

ndhanushkodi (approved):
Amazing work on this Michael!

.circleci/config.yml
wilkermichael: Sure, if we feel they're not needed I'm totally okay with that. Was just going by what we decided on for the testing strategy in the swim lane diagram

ndhanushkodi: Actually since its running overnight, it might be nice to have a kind test case at night too, might be a nice comparison point against cloud tests also. (if only clouds fail but kind passes on the same commit)

kschoche (approved):

PR #1656, opened 2022-10-27
Add PodSecurityPolicy for Consul API Gateway controller

nathancoleman:
**Changes proposed in this PR:**
- Add a `PodSecurityPolicy` for the API gateway controller when both
  - API Gateway is enabled in `values.yaml`
  - PodSecurityPolicies are enabled in `values.yaml`
- Policy mimics the one already in place for the controller at `controller-podsecuritypolicy.yaml`
- Slight adjustment to disable write access to the local file system since the API gateway controller does not require this

> **Note** `PodSecurityPolicy` was removed with Kubernetes 1.25, so this change must be tested on <= 1.24. I'm planning to work with @kschoche on supporting Pod Security Admission, the 1.25-friendly replacement, in the coming weeks.

> **Note** This adds a `PodSecurityPolicy` for the API gateway controller but not for the `Deployments` created by the controller for every `Gateway`. That will be done in future PR(s).

**How I've tested this PR:**
Create a Kubernetes cluster that requires `PodSecurityPolicies`. For GKE:
```shell
# Create the Kubernetes cluster
$ gcloud container clusters create cluster-1

# Require PodSecurityPolicy - this will take a while, approx. 30 minutes in my case
$ gcloud beta container clusters update cluster-1 --enable-pod-security-policy
```

Install Consul using consul-k8s `0.49.0` with Consul API Gateway enabled. Notice that the `consul-api-gateway-controller` Deployment is unable to admit any pods due to the missing `PodSecurityPolicy`:
```shell
$ kubectl get deployment consul-api-gateway-controller --output yaml
...
status:
  conditions:
    ...
    - lastTransitionTime: "2022-10-27T22:30:00Z"
      lastUpdateTime: "2022-10-27T22:30:00Z"
      message: 'pods "consul-api-gateway-controller-5bccc6b66d-" is forbidden: PodSecurityPolicy: unable to admit pod: []'
      reason: FailedCreate
      status: "True"
      type: ReplicaFailure
```

Now, install Consul into the same cluster using this version of the Helm chart with Consul API Gateway and PodSecurityPolicies enabled. Then create a `Gateway`, some services, and one or more `HTTPRoutes`. I pretty much follow [this guide](https://developer.hashicorp.com/consul/tutorials/kubernetes/kubernetes-api-gateway) but use this branch of consul-k8s.

[...]

PR #1672, opened 2022-11-02
api-gateway: allow controller to bind PodSecurityPolicy to ServiceAccounts that it creates

nathancoleman:
> **Note** This builds on #1656 

When a user has both `global.enablePodSecurityPolicies=true` and `apiGateway.enabled=true`:
- Create a `PodSecurityPolicy` for use by all `Deployments` that the API gateway controller creates to back a `Gateway`
- Allow the API gateway controller to use `PodSecurityPolicies`

When `global.acls.manageSystemACLs=true` is also set:
- Provide the name of the `PodSecurityPolicy` to the `GatewayClassConfig` so that the API gateway controller can bind the policy to the `ServiceAccount` created for each `Gateway`. It does this by creating a `Gateway`-specific `Role` and `RoleBinding`.

The API gateway controller could create the `PodSecurityPolicy` itself; however, this isn't nearly as flexible as any user requiring a slight change in the `PSP` would need a code release from us. This way, they can modify the `PSP` created by the Helm chart however they wish, or they can create their own `PSP` and reference it in the `GatewayClassConfig`.

**How I've tested this PR:**

Create a K8s cluster that requires `PodSecurityPolicies`. On GKE, for example:
```shell
$ gcloud beta container clusters create cluster-1 --enable-pod-security-policy --region us-east1
```

Then install this version of the Helm chart, verifying that the `PodSecurityPolicy` is created and the `ClusterRole` for the API gateway controller has the modified set of permissions.
```yaml
global:
  enablePodSecurityPolicies: true
  acls:
    manageSystemACLs: true
...
apiGateway:
  enabled: true
  ...
```

👀 verify that the permissions added here are as restricted as they can possibly be while meeting requirements

**Checklist:**
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/api-gateway-controller-clusterrole.yaml
+++ b/charts/consul/templates/api-gateway-controller-clusterrole.yaml
@@ -11,6 +11,24 @@ metadata:
     release: {{ .Release.Name }}
     component: api-gateway-controller
 rules:
+{{- if .Values.global.enablePodSecurityPolicies }}
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  verbs:
+  - use
+{{- end}}
 - apiGroups:
   - api-gateway.consul.hashicorp.com
   resources:
--- a/charts/consul/templates/api-gateway-controller-clusterrole.yaml
+++ b/charts/consul/templates/api-gateway-controller-clusterrole.yaml
@@ -11,24 +11,6 @@ metadata:
     release: {{ .Release.Name }}
     component: api-gateway-controller
 rules:
-{{- if .Values.global.enablePodSecurityPolicies }}
-- apiGroups:
-  - rbac.authorization.k8s.io
-  resources:
-  - roles
-  - rolebindings
-  verbs:
-  - create
-  - get
-  - list
-  - watch
-- apiGroups:
-  - policy
-  resources:
-  - podsecuritypolicies
-  verbs:
-  - use
-{{- end}}
 - apiGroups:
   - api-gateway.consul.hashicorp.com
   resources:
@@ -263,11 +245,21 @@ rules:
   - patch
   - update
 {{- if .Values.global.enablePodSecurityPolicies }}
-- apiGroups: ["policy"]
-  resources: ["podsecuritypolicies"]
-  resourceNames:
-    - {{ template "consul.fullname" . }}-api-gateway-controller
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
   verbs:
-    - use
+  - use
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
+  verbs:
+  - create
+  - get
+  - list
+  - watch
 {{- end }}
 {{- end }}
--- a/charts/consul/templates/api-gateway-gatewayclassconfig.yaml
+++ b/charts/consul/templates/api-gateway-gatewayclassconfig.yaml
@@ -15,6 +15,9 @@ spec:
       {{- if .Values.global.acls.manageSystemACLs }}
       managed: true
       method: {{ template "consul.fullname" . }}-k8s-auth-method
+      {{- if .Values.global.enablePodSecurityPolicies }}
+      podSecurityPolicy: {{ template "consul.fullname" . }}-api-gateway
+      {{- end }}
       {{- end }}
     {{- if .Values.global.tls.enabled }}
     scheme: https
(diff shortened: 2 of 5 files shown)


charts/consul/templates/api-gateway-controller-clusterrole.yaml
```diff
@@ -245,11 +245,21 @@ rules:
...
-- apiGroups: ["policy"]
-  resources: ["podsecuritypolicies"]
-  resourceNames:
-    - {{ template "consul.fullname" . }}-api-gateway-controller
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  verbs:
+  - use
```
nathancoleman, line 253: This was previously narrowed to the specific `PodSecurityPolicy` created for the API Gateway controller; however, the controller needs to be able to create `Roles/RoleBindings` that use whatever `PSP` is named in the `GatewayClassConfig`. Since you can't grant permissions that you don't have, the controller needs to be able to use more than just the `PSP` that applies to itself.

charts/consul/templates/api-gateway-controller-clusterrole.yaml
```diff
@@ -245,11 +245,21 @@ rules:
...
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
   verbs:
-    - use
+  - create
+  - get
+  - list
+  - watch
```
nathancoleman, line 263: I'm not a huge fan of allowing the API gateway controller to create `Roles` + `RoleBindings`; however, I don't see any other way to accomplish attaching a `PodSecurityPolicy` to each and every `Deployment` that the controller creates. Due to this, I've narrowed down the permissions as much as possible - note that it can't `update` existing resources - and only granting these permissions when `global.enablePodSecurityPolicies`.

charts/consul/templates/api-gateway-gatewayclassconfig.yaml
```diff
@@ -15,6 +15,9 @@ spec:
       {{- if .Values.global.acls.manageSystemACLs }}
       managed: true
       method: {{ template "consul.fullname" . }}-k8s-auth-method
+      {{- if .Values.global.enablePodSecurityPolicies }}
+      podSecurityPolicy: {{ template "consul.fullname" . }}-api-gateway
```
nathancoleman, line 19: This specifies the name of the `podSecurityPolicy` that should apply to any `Gateway` using this `GatewayClassConfig`. The API gateway controller will then create a `Role` and `RoleBinding` to bind this PSP to the `ServiceAccount` that it creates.

This is intentionally pretty flexible so that a user doesn't need a code release from us if they, for example, wanted to create their own `PodSecurityPolicy` and have the API gateway controller use that. By allowing this, we let users tailor to their needs without requiring a code release from us.

charts/consul/templates/api-gateway-podsecuritypolicy.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if and .Values.apiGateway.enabled .Values.global.enablePodSecurityPolicies }}
```
nathancoleman, line 1: This is the PSP that will be applied to all `Gateway` deployments by default when `global.enablePodSecurityPolicies=true`. As I mentioned above, customers needing something more or less restrictive could create their own `PodSecurityPolicy` or modify this one and have the API gateway controller use it instead.

This is based on the PSP that we use for the API gateway controller with some slight modifications. Please review this for sensibility.

nathancoleman (commented):
Personal review

charts/consul/templates/api-gateway-controller-clusterrole.yaml
thisisnotashwin: Thanks so much for the detail in the comment. I really appreciate it.

thisisnotashwin (approved):
Loved the comment and the detail! This looks great!

t-eckert (approved):
Great work! Thanks for the detail on how to test the PR.

charts/consul/test/unit/api-gateway-controller-clusterrole.bats
```diff
@@ -20,14 +20,26 @@ load _helpers
...
+
+@test "apiGateway/ClusterRole: can create roles and rolebindings with apiGateway.enabled=true and global.enablePodSecurityPolicies=true" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/api-gateway-controller-clusterrole.yaml \
+      --set 'global.enablePodSecurityPolicies=true' \
+      --set 'apiGateway.enabled=true' \
+      --set 'apiGateway.image=foo' \
+      . | tee /dev/stderr |
+      yq '.rules[] | select((.resources[0] == "roles") and (.resources[1] == "rolebindings") and (.verbs[0] == "create")) | length > 0' | tee /dev/stderr)
```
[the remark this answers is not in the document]
nathancoleman: No reason not to, I'll add that in a few minutes 💯

kschoche (approved):
🔥 
Great work on this!

PR #1673, opened 2022-11-02
Default install and upgrade to "Y" when no answer is given

t-eckert:
Changes proposed in this PR:
- When a user is promped to install Consul, if they give no answer (e.g. just hit enter) the install will occur.

How I've tested this PR:
- Manually by running an install.

- You may also run an install.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/cli/helm/install.go
+++ b/cli/helm/install.go
@@ -99,6 +99,7 @@ func InstallHelmRelease(options *InstallOptions) error {
 		if err != nil {
 			return err
 		}
+		// The install will proceed if the user presses enter or responds with "y"/"yes" (case-insensitive).
 		if confirmation != "" && common.Abort(confirmation) {
 			options.UI.Output("Install aborted. Use the command `consul-k8s install -help` to learn how to customize your installation.",
 				terminal.WithInfoStyle())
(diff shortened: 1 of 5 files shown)


cli/helm/install.go
```diff
@@ -91,15 +91,15 @@ func InstallHelmRelease(options *InstallOptions) error {
...
+			Prompt: "Proceed with installation? (Y/n)",
 			Style:  terminal.InfoStyle,
 			Secret: false,
 		})
 
 		if err != nil {
 			return err
 		}
-		if common.Abort(confirmation) {
+		if confirmation != "" && common.Abort(confirmation) {
```
wilkermichael, line 102: I just suggest adding a comment on the intention here, that we intend to be approving on default

wilkermichael (approved):
Looks good! very minor comment

jmurret (commented):
Should uninstall and upgrade follow the same paradigm of "you typed this command so I am going to default to you wanted to do this" or are we leaving those with N as the default because there is more risk to those?

t-eckert:
> Should uninstall and upgrade follow the same paradigm of "you typed this command so I am going to default to you wanted to do this" or are we leaving those with N as the default because there is more risk to those?

I could see doing this for upgrade, but the convention is to only auto-yes for "additive" commands. I wouldn't do it for uninstall because it could be a footgun.

david-yu:
Still waiting on @jmurret to review.

jmurret (approved):
looks good!  🙏

PR #1674, opened 2022-11-02
agentless: update acl-init to use server-connection-manager-library

thisisnotashwin:
Changes proposed in this PR:
- Update ACL init to use Consul Server Connection Manager

How I've tested this PR:
- Unit tests
- Acceptance test (basic)

- 👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/client-daemonset.yaml
+++ b/charts/consul/templates/client-daemonset.yaml
@@ -493,38 +493,25 @@ spec:
         {{- if (and .Values.global.tls.enabled (not .Values.externalServers.useSystemRoots)) }}
         - name: CONSUL_CACERT
           {{- if .Values.global.secretsBackend.vault.enabled }}
-          value: "/vault/secrets/serverca.crt" 
+          value: "/vault/secrets/serverca.crt"
           {{- else }}
           value: "/consul/tls/ca/tls.crt"
           {{- end }}
+        {{- end }}
+        {{- include "consul.consulK8sConsulServerEnvVars" . | nindent 8 }}
+        {{- if .Values.global.acls.manageSystemACLs }}
+        - name: CONSUL_LOGIN_AUTH_METHOD
+          value: {{ template "consul.fullname" . }}-k8s-component-auth-method
+        - name: CONSUL_LOGIN_META
+          value: "component=client"
         {{- end }}
         command:
           - "/bin/sh"
           - "-ec"
           - |
             consul-k8s-control-plane acl-init \
-              -component-name=client \
-              -acl-auth-method="{{ template "consul.fullname" . }}-k8s-component-auth-method" \
-              {{- if .Values.global.adminPartitions.enabled }}
-              -partition={{ .Values.global.adminPartitions.name }} \
-              {{- end }}
               -log-level={{ default .Values.global.logLevel .Values.client.logLevel }} \
               -log-json={{ .Values.global.logJSON }} \
-              {{- if .Values.externalServers.enabled }}
-              {{- if .Values.global.tls.enabled }}
-              -use-https \
-              {{- end }}
-              {{- range .Values.externalServers.hosts }}
-              -server-address={{ quote . }} \
-              {{- end }}
-              -server-port={{ .Values.externalServers.httpsPort }} \
-              {{- if .Values.externalServers.tlsServerName }}
-              -tls-server-name={{ .Values.externalServers.tlsServerName }} \
-              {{- end }}
-              {{- else if .Values.global.cloud.enabled }}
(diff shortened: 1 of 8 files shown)


control-plane/helper/test/test_util.go
```diff
@@ -151,7 +151,10 @@ func SetupK8sComponentAuthMethod(t *testing.T, consulClient *api.Client, service
 	_, _, err := consulClient.ACL().AuthMethodCreate(&authMethodTmpl, nil)
 	require.NoError(t, err)
 
-	rules := `acl = "write"`
+	rules := `acl = "write"
```
thisisnotashwin, line 154: This is required for the unit tests to run correctly.

control-plane/subcommand/acl-init/command.go
```diff
@@ -162,47 +137,31 @@ func (c *Command) Run(args []string) int {
...
+			c.logger.Error("Failed to create Consul client", "error", err)
 		}
-		secret, err = common.ConsulLogin(c.consulClient, loginParams, c.logger)
+		state, err := c.watcher.State()
 		if err != nil {
-			c.logger.Error("Consul login failed", "error", err)
-			return 1
+			c.logger.Error("Failed to get Consul watcher state", "error", err)
 		}
+		secret = state.Token
```
thisisnotashwin, line 164: I'd love a suggestion here with regards to finding a way to get the token without having to read the `State` again. I could use the variant of `NewClientFromConnManager` that accepts that watcher state but I think we want to move away from using that method directly.

charts/consul/templates/client-daemonset.yaml
```diff
@@ -493,38 +493,25 @@ spec:
...
-          value: "/vault/secrets/serverca.crt" 
+          value: "/vault/secrets/serverca.crt"
           {{- else }}
           value: "/consul/tls/ca/tls.crt"
           {{- end }}
         {{- end }}
+        {{- include "consul.consulK8sConsulServerEnvVars" . | nindent 8 }}
+        {{- if .Values.global.acls.manageSystemACLs }}
+        - name: CONSUL_LOGIN_AUTH_METHOD
+          value: {{ template "consul.fullname" . }}-k8s-component-auth-method
```
curtbushko, line 504: Do you need datacenter also?
```
               - name: CONSUL_LOGIN_DATACENTER
              {{- if and .Values.global.federation.enabled .Values.global.federation.primaryDatacenter .Values.global.enableConsulNamespaces }}
              value: {{ .Values.global.federation.primaryDatacenter }}
              {{- else }}
              value: {{ .Values.global.datacenter }}
              {{- end }}
```

control-plane/subcommand/acl-init/command.go
curtbushko: I believe that was pretty much how Iryna had to do it too in the sync-catalog PR @ https://github.com/hashicorp/consul-k8s/blob/main/control-plane/subcommand/sync-catalog/command.go#L232

curtbushko (approved):
Looks great!  Just one question but I wouldn't block you on it!

charts/consul/test/unit/client-daemonset.bats
```diff
@@ -1241,24 +1241,36 @@ load _helpers
...
   local actual=$(echo $object |
-      yq -r '.command | any(contains("component-name=client"))' | tee /dev/stderr)
-  [ "${actual}" = "true" ]  
+      yq -r '.env[1].value | contains("release-name-consul-server.default.svc")' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
 
   local actual=$(echo $object |
-      yq -r '.command | any(contains("init-type=\"client\""))' | tee /dev/stderr)
-  [ "${actual}" = "true" ]  
+      yq -r '.env[6].name | contains("CONSUL_LOGIN_AUTH_METHOD")' | tee /dev/stderr)
```
curtbushko, line 1257: To make these environment variable tests less flaky, could you use $env instead of $object to get the variables?

See: https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/test/unit/sync-catalog-deployment.bats#L73

thisisnotashwin:
@curtbushko @ishustava Im going to close this PR. While debugging the failing acceptance tests for Clients, I realized that the watcher performs a Consul logout when it's process is terminated, which means the token it writes is no longer valid when the client spins up and attempts to use it. I don't think using the watcher will work and I will need to use a solution that uses the netaddr library. That will looks very different from the contents of this PR. Ill put a new one with them changes. Sorry for the time taken to review this :(

PR #1680, opened 2022-11-04
Set 5min timeout on HTTP Client

thisisnotashwin:
Changes proposed in this PR:

- The call to the bootstrap endpoint of the server can take some time to respond when running on cloud platforms. This commit re-introduces the 5min timeout to the HTTP client from the default 5 seconds.

How I've tested this PR:
- acceptance tests against the clouds

- 👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1272,7 +1272,7 @@ workflows:
       - acceptance-tproxy:
           context: consul-ci
           requires:
-            - dev-upload-docker
+          - dev-upload-docker
 
   nightly-cleanup:
     triggers:
--- a/control-plane/subcommand/server-acl-init/servers.go
+++ b/control-plane/subcommand/server-acl-init/servers.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/consul/api"
 	apiv1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


thisisnotashwin:
While some of the AKS tests are falling, looking into the logs, the failures due to server-acl-init are no longer something we are running into.

.circleci/config.yml
```diff
@@ -1269,10 +1269,34 @@ workflows:
           context: consul-ci
           requires:
             - dev-upload-docker
+      - acceptance-gke-1-23:
```
thisisnotashwin, line 1272: I will revert these changes before merging.

curtbushko, line 1272: Will remove before merge?

curtbushko (approved):

.circleci/config.yml
thisisnotashwin: haha! yup! i had that comment written down but forgot to hit publish 😂

control-plane/subcommand/server-acl-init/servers.go
```diff
@@ -48,8 +50,8 @@ func (c *Command) bootstrapServers(serverAddresses []net.IPAddr, bootstrapToken,
 func (c *Command) bootstrapACLs(firstServerAddr, bootTokenSecretName string) (string, error) {
 	config := c.consulFlags.ConsulClientConfig().APIClientConfig
 	config.Address = firstServerAddr
-	consulClient, err := consul.NewClient(config,
-		c.consulFlags.APITimeout)
+	config.HttpClient = &http.Client{Timeout: 5 * time.Minute}
+	consulClient, err := consul.NewClient(config, c.consulFlags.APITimeout)
```
ishustava, line 54: Could you also keep the comment from before? I think it's useful.

ishustava (approved):

control-plane/subcommand/server-acl-init/servers.go
thisisnotashwin: Yup!! Will do!

PR #1681, opened 2022-11-04
Use netaddrs instead of go-discover in acl-init

thisisnotashwin:
Changes proposed in this PR:
- Migrate `acl-init` job to use the `netaddrs` library to resolve the address passed in instead of go-discover in order to align it with changes being made in the project.

How I've tested this PR:
- Unit tests are still 🟢 
- Acceptance tests for the basic case as that running clients with ACLs enabled.

- 👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


thisisnotashwin:
The failing tproxy test is because it ran into the timeout issue. That is fixed is a separate PR but isn't rebased on this one yet and hence the failure

curtbushko (approved):
Approving because it looks alright to me but I think Iryna gets the final say.  :)

thisisnotashwin:
Closing in favor of #1689

PR #1683, opened 2022-11-07
peering: remove unnecessary expose servers service, add error cases, fix flaky test

ndhanushkodi:
Changes proposed in this PR:
- Expose servers service is only needed when partitions are enabled, since peering now requires mesh gateways. So this PR makes it so the helm chart doesn't deploy the expose servers service when only peering (and not partitions) is enabled. 
- Fail fast in the helm chart if tls or mesh gateways are not enabled since they are required for peering
- In acceptance tests- fix flakiness due to mesh configuration not being recognized before the applying the acceptor resource (which generates the token), to ensure that the token is generated with mesh gateway addresses rather than server pod ips. 

How I've tested this PR:
acceptance tests, unit tests

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ BREAKING_CHANGES:
 * Peering:
   * Remove support for customizing the server addresses in peering token generation. Instead, mesh gateways should be used
     to establish peering connections if the server pods are not directly reachable. [[GH-1610](https://github.com/hashicorp/consul-k8s/pull/1610)]
-  * Enabling peering requires `tls.enabled`. [[GH-1610](https://github.com/hashicorp/consul-k8s/pull/1610)]
+  * Require `global.tls.enabled` when peering is enabled. [[GH-1610](https://github.com/hashicorp/consul-k8s/pull/1610)]
+  * Require `meshGateway.enabled` when peering is enabled. [[GH-1683](https://github.com/hashicorp/consul-k8s/pull/1683)]
 
 FEATURES:
 * Consul-dataplane:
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ BREAKING_CHANGES:
 * Peering:
   * Remove support for customizing the server addresses in peering token generation. Instead, mesh gateways should be used
     to establish peering connections if the server pods are not directly reachable. [[GH-1610](https://github.com/hashicorp/consul-k8s/pull/1610)]
-  * Enabling peering requires `tls.enabled`. [[GH-1610](https://github.com/hashicorp/consul-k8s/pull/1610)]
+  * Enabling peering requires `global.tls.enabled`. [[GH-1610](https://github.com/hashicorp/consul-k8s/pull/1610)]
+  * Enabling peering requires `meshGateway.enabled`. [[GH-1683](https://github.com/hashicorp/consul-k8s/pull/1683)]
```
ishustava, line 16: nit: could we rephrase to be "verb ..." so "Require `global.tls.enabled` when peering is enabled" 🙏

CHANGELOG.md
```diff
@@ -34,6 +35,7 @@ IMPROVEMENTS:
   * API Gateway: Add `tolerations` to `apiGateway.managedGatewayClass` and `apiGateway.controller` [[GH-1650](https://github.com/hashicorp/consul-k8s/pull/1650)]
   * API Gateway: Create PodSecurityPolicy for controller when `global.enablePodSecurityPolicies=true`. [[GH-1656](https://github.com/hashicorp/consul-k8s/pull/1656)]
   * API Gateway: Create PodSecurityPolicy and allow controller to bind it to ServiceAccounts that it creates for Gateway Deployments when `global.enablePodSecurityPolicies=true`. [[GH-1672](https://github.com/hashicorp/consul-k8s/pull/1672)]
+  * Expose servers service is only deployed when Admin Partitions(ENT) is enabled. [[GH-1683](https://github.com/hashicorp/consul-k8s/pull/1683)]
```
ishustava, line 38: Same here:
```suggestion
  * Deploy `expose-servers` service only deployed when Admin Partitions(ENT) is enabled. [[GH-1683](https://github.com/hashicorp/consul-k8s/pull/1683)]
```

ishustava (approved):
Looks great!

jmurret (approved):
👍 Looks good!  Nice work!

CHANGELOG.md
jmurret: The suggestion above has the word `deploy` twice.

PR #1686, opened 2022-11-08
Add externalServers.skipServerWatch to helm for working with load balancers

curtbushko:
Changes proposed in this PR:
- Add helm value of `externalServers.skipServerWatch`
- Add the environment variable CONSUL_SKIP_SERVER_WATCH to helm _helpers.tpl so that the environment variable will get automatically picked up by connection manager when talking to consul servers
- Even though not directly needed there, add the check that `externalServers.enabled` is also set when `externalServers.skipServerWatch=true` in connect-inject-deployment.yaml
- Bats tests for connect-inject-deployment to make sure that it fails the condition above and that the CONSUL_SKIP_SERVER_WATCH is set in the environment.
- Read the env variable in flags/consul.go so that it can be passed to connection manager and down into discover.Config (used by partition-init, catalog-sync, server-acl-init, inject-connect, etc)
- Specifically, set the flag in the webhook so that it is passed onto consul-dataplane as a `-server-watch-disabled=true` flag

How I've tested this PR:
- bats tests
- unit tests

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -355,6 +355,10 @@ Consul server environment variables for consul-k8s commands.
   value: {{ .Values.externalServers.tlsServerName }}
 {{- end }}
 {{- end }}
+{{- if and .Values.externalServers.enabled .Values.externalServers.skipServerWatch }}
+- name: CONSUL_SKIP_SERVER_WATCH
+  value: "true" 
+{{- end }}
 {{- end -}}
 
 {{/*
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -8,6 +8,7 @@
 {{- $serverEnabled := (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) -}}
 {{- $serverExposeServiceEnabled := (or (and (ne (.Values.server.exposeService.enabled | toString) "-") .Values.server.exposeService.enabled) (and (eq (.Values.server.exposeService.enabled | toString) "-") .Values.global.adminPartitions.enabled)) -}}
 {{- if and .Values.externalServers.enabled (not .Values.externalServers.hosts) }}{{ fail "externalServers.hosts must be set if externalServers.enabled is true" }}{{ end -}}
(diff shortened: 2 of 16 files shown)


charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -8,6 +8,7 @@
 {{- $serverEnabled := (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) -}}
 {{- $serverExposeServiceEnabled := (or (and (ne (.Values.server.exposeService.enabled | toString) "-") .Values.server.exposeService.enabled) (and (eq (.Values.server.exposeService.enabled | toString) "-") .Values.global.adminPartitions.enabled)) -}}
 {{- if and .Values.externalServers.enabled (not .Values.externalServers.hosts) }}{{ fail "externalServers.hosts must be set if externalServers.enabled is true" }}{{ end -}}
+{{- if and .Values.externalServers.skipServerWatch (not .Values.externalServers.enabled) }}{{ fail "externalServers.enabled must be set if externalServers.skipServerWatch is true" }}{{ end -}}
```
curtbushko, line 11: Putting the fail condition here so that it is obvious that it pairs with the checks in connect-inject-deployment.bats. 

A similar thing is done for externalServers.hosts and _helpers.tpl. The checks are here in connect-inject-deployment.yaml.

charts/consul/templates/_helpers.tpl
```diff
@@ -355,6 +355,10 @@ Consul server environment variables for consul-k8s commands.
   value: {{ .Values.externalServers.tlsServerName }}
 {{- end }}
 {{- end }}
+{{- if and .Values.externalServers.enabled .Values.externalServers.skipServerWatch }}
+- name: CONSUL_SKIP_SERVER_WATCH
+  value: "true"
```
thisisnotashwin, line 360: it's a nit but can we unqoute the "true"

thisisnotashwin (approved):
This is really nicely done and so concise!!

charts/consul/values.yaml
```diff
@@ -1183,6 +1183,10 @@ externalServers:
   # @type: string
   k8sAuthMethodHost: null
 
+  # If true, setting this prevents the consul-dataplane from watching the consul server upstream for changes. This is
```
ishustava, line 1186:
```suggestion
  # If true, setting this prevents the consul-dataplane and consul-k8s components from watching the Consul servers for changes. This is
```

ishustava (approved):
Looks great!!

PR #1688, opened 2022-11-09
helm: default dns values inherit from connectInject.transparentProxy.defaultEnabled

ndhanushkodi:
Changes proposed in this PR:
- Previously, `dns.enabled` defaulted to inherit from `global.enabled`, and `dns.enableRedirection` defaulted to `false`. Now, both of these dns values will default to inherit from `connectInject.transparentProxy.defaultEnabled`.
- This is a breaking change

How I've tested this PR:
- bats

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


ndhanushkodi:
The failing acceptance tests are due to an agentless change for distroless image, so this is ready for review!

t-eckert (approved):
This looks great!

analogue (approved):

PR #1689, opened 2022-11-09
Use netaddrs instead of go-discover in acl-init

thisisnotashwin:
Changes proposed in this PR:
- Migrate `acl-init` job to use the `netaddrs` library to resolve the address passed in instead of go-discover in order to align it with changes being made in the project.
- This migrates acl-init to use the new consul flags but uses a traditional Consul login instead of the watcher as the watcher cleans up the token when the process stops. Using the Consul login allows the client to use the correct token when they come up and their `preStop` hook logouts and cleans up the token.

How I've tested this PR:
- Unit tests are still 🟢 
- Acceptance tests for the basic case as that running clients with ACLs enabled.

- 👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/client-daemonset.yaml
+++ b/charts/consul/templates/client-daemonset.yaml
@@ -484,47 +484,28 @@ spec:
       - name: client-acl-init
         image: {{ .Values.global.imageK8S }}
         env:
-        - name: CONSUL_HTTP_ADDR
-          {{- if .Values.global.tls.enabled }}
-          value: https://{{ template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc:8501
-          {{- else }}
-          value: http://{{ template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc:8500
-          {{- end }}
-        {{- if (and .Values.global.tls.enabled (not .Values.externalServers.useSystemRoots)) }}
-        - name: CONSUL_CACERT
-          {{- if .Values.global.secretsBackend.vault.enabled }}
-          value: "/vault/secrets/serverca.crt" 
-          {{- else }}
-          value: "/consul/tls/ca/tls.crt"
-          {{- end }}
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: POD_NAME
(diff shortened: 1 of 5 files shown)


charts/consul/templates/client-daemonset.yaml
```diff
@@ -493,38 +493,25 @@ spec:
         {{- if (and .Values.global.tls.enabled (not .Values.externalServers.useSystemRoots)) }}
         - name: CONSUL_CACERT
```
ishustava, line 494: This one and `CONSUL_HTTP_ADDR` should be removed now, right?

charts/consul/templates/client-daemonset.yaml
```diff
@@ -493,38 +493,25 @@ spec:
...
           {{- else }}
           value: "/consul/tls/ca/tls.crt"
           {{- end }}
         {{- end }}
+        {{- include "consul.consulK8sConsulServerEnvVars" . | nindent 8 }}
+        {{- if .Values.global.acls.manageSystemACLs }}
+        - name: CONSUL_LOGIN_AUTH_METHOD
+          value: {{ template "consul.fullname" . }}-k8s-component-auth-method
+        - name: CONSUL_LOGIN_META
+          value: "component=client"
```
ishustava, line 506: What about pod name?

control-plane/subcommand/server-acl-init/rules.go
```diff
@@ -90,7 +90,7 @@ partition "{{ .PartitionName }}" {
   namespace_prefix "" {
 {{- end }}
     service_prefix "" {
-      policy = "read"
+      policy = "write"
```
ishustava, line 93: Why do these rules need to change?

ishustava (commented):
Looking good! Just a few questions and comments

control-plane/subcommand/server-acl-init/rules.go
thisisnotashwin: dont need it anymore!! its been done!!

ishustava (approved):
Looks great!

curtbushko (approved):
Looks fantastic!  Thanks!

PR #1690, opened 2022-11-09
agentless: support updating health checks on consul clients during an upgrade to agentless

ishustava:
Changes proposed in this PR:
We will need to handle the case when a user has many applications on the service mesh and needs to take time to re-inject all those applications. If that is the case, there may be situations where the application changes its health status, but because it still hasn’t been re-started (re-injected), its health status needs to be synced into Consul clients rather than servers the way we used to do before. To do this, we will add a new annotation to all pods pointing to the current `consul-k8s version: consul.hashicorp.com/consul-k8s-version: 1.0.0`. If this annotation is not there or if the version is less than the version that supports “agentless”, we will continue to sync health checks to Consul clients for that particular pod

How I've tested this PR:
manual upgrade testing in multiple setups (not-secure, tls, tls +acls, tls+auto-encrypt+acls)

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -236,24 +236,12 @@ spec:
                 {{- end }}
                 {{- end }}
 
-                {{- if .Values.global.consulSidecarContainer }}
```
ishustava, line 239: These flags are no longer used

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -283,38 +271,38 @@ spec:
...
           {{- else }}
-          secretName: {{ template "consul.fullname" . }}-ca-cert
+            secretName: {{ template "consul.fullname" . }}-ca-cert
           {{- end }}
-          items:
-          - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
-            path: tls.crt
+            items:
+              - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
+                path: tls.crt
```
ishustava, line 305: My IDE decided to format these

charts/consul/templates/connect-inject-deployment.yaml
curtbushko: The removal of these are actually in the PR I am working on right now as I missed them last week.

curtbushko (approved):
Nice job!

charts/consul/templates/connect-inject-deployment.yaml
thisisnotashwin: Hopefully the rebase of main onto your PR should automatically clear some of that from your commits.

thisisnotashwin (approved):
Excellent work, as usual 😄

PR #1691, opened 2022-11-09
Add support for maxConnections, maxConcurrentRequests, and maxPendingRequests to IngressGateway CRD

analogue:
Changes proposed in this PR (NET-1139):
- Builds on https://github.com/hashicorp/consul/pull/14749 by allowing users to configure consul-k8s ingress-gateway

How I've tested this PR:
- Unit tests
- Created an ingress-gateway CR using the new fields and verified the fields were present on the resource without error.

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


ndhanushkodi (approved):

control-plane/api/v1alpha1/ingressgateway_types.go
```diff
@@ -144,6 +161,18 @@ type IngressService struct {
...
+	// will be allowed to establish against the given upstream. Use this to limit
+	// HTTP/1.1 traffic, since HTTP/1.1 has a request per connection.
+	MaxConnections *uint32 `json:"maxConnections,omitempty"`
+	// The maximum number of requests that will be queued
+	// while waiting for a connection to be established.
+	MaxPendingRequests *uint32 `json:"maxPendingRequests,omitempty"`
+	// The maximum number of concurrent requests that
+	// will be allowed at a single point in time. Use this to limit HTTP/2 traffic,
+	// since HTTP/2 has many requests per connection.
+	MaxConcurrentRequests *uint32 `json:"maxConcurrentRequests,omitempty"`
```
thisisnotashwin, line 175: qq: can we inline the IngressServiceConfig struct here since it has the exact same fields?

thisisnotashwin: That should potentially allow us to reuse it's validate behavior as well.

thisisnotashwin (commented):
Hey Semir!! This looks excellent!! had a quick comment but looks good otherwise

control-plane/api/v1alpha1/ingressgateway_types.go
analogue: That is good suggestion. I was torn between doing that and maintaining consistency with the structure of the same fields in consul's config entries.

https://github.com/hashicorp/consul/blob/52c2dd6d831475216fe8326300ceb6e62c18bee0/agent/structs/config_entry_gateways.go#L102

Happy to make the change if we've historically not valued or maintained such consistency in similar situations.

thisisnotashwin: I think that is fine. i dont think we have maintained such consistency in our case honestly. i suspect it was oversight in the case of consul as well!

analogue: Digging into this a little more and considering the practioner UX, I think making this change would introduce subtle differences in consistency that may be surprising or worse yet, nuanced confusion.

Let's use https://developer.hashicorp.com/consul/docs/connect/config-entries/ingress-gateway#wildcard-http-listener as an example:

Practionioner experience:
- Look at `HCL` tab for the configuration of an ingress gateway. Notice the general overall structure and nesting.
- Switch to the `Kubernetes YAML`.  Notice the general overall structure.
- Observation: these two (HCL vs K8s yaml) are structured in pretty much identical ways.
- Takeaway: configuration structure is consistent between both so if I know how to do it in one format, the other will be obvious and predictable.
- Scroll down to https://developer.hashicorp.com/consul/docs/connect/config-entries/ingress-gateway#available-fields
- Read `HCL` tab and then compare with `Kubernetes YAML` tab
- Observation: reinforces that structure is the same between the two.

However, we're proposing to break the consistency of configuration structure (regardless of whether the initially implemented Consul way is ideal or not) and introducing a subtle difference.

Specifically regarding the config options in this PR:

Consul
```
Kind = "ingress-gateway"
Name = "my-ingress-gateway"

Defaults {
    MaxConnections = 100
    MaxPendingRequests = 200
    MaxConcurrentRequests = 300
}
Listeners = [
 {
   Port = 8082
   Protocol = "http"
   Services = [
     {
        Name = "counting"
        Hosts = ["*"]
        MaxConnections = 101
        MaxPendingRequests = 201
        MaxConcurrentRequests = 301
     },
   ]
 }
]
```

Kubernetes YAML
```
---
kind: IngressGateway
metadata:
  name: my-ingress-gateway
spec:
  defaults:
    maxConnections: 100
    maxPendingPendingRequests: 200
    maxConcurrentRequests: 300
  listeners:
    - port: 8082
      protocol: http
      services:
        - name: counting
          hosts: ["*"]
          maxConnections: 5120
          serviceConfig:            # <-- serviceConfig not present in HCL 
            maxConnections: 101
            maxPendingRequests: 201
            maxConcurrentRequets: 301
``` 

I want to make sure we're OK breaking this consistency even if the `serviceConfig` structure is arguably better.

thisisnotashwin (approved):

control-plane/api/v1alpha1/ingressgateway_types.go
analogue: TIL json inline :)

PR #1692, opened 2022-11-09
helm: pass partition and datacenter to consul-api-gateway server command

mikemorris:
#### Changes proposed in this PR:
- Plumb configured Consul datacenter and partition values from Helm config through to Consul API Gateway command as CLI args.

#### How I expect reviewers to test this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -92,6 +92,10 @@ spec:
             {{- end }}
             {{- end }}
             {{- end }}
+            {{- if .Values.global.adminPartitions.enabled }}
+            -partition {{ .Values.global.adminPartitions.name }} \
+            {{- end }}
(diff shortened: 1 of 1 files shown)


andrewstucki:
Hey Mike, sorry I missed the morning sync today, but want to point out that any additional flags passed in as CLI arguments result in backwards incompatible changes to our binary. IF we decide we want to keep backwards compat in our next set of releases we'll need to pass this in as something more "optional" like an environment variable. If you take a look what happens now when you pass undefined flags:

```
➜  consul-api-gateway git:(main) ./consul-api-gateway server -foo
flag provided but not defined: -foo
...
```

We probably want to lean towards using `CONSUL_PARTITION` for partition passing, and may want to do something custom for datacenter.

t-eckert:
@mikemorris, I handled this using env vars in #1721 so I will close this PR.

PR #1694, opened 2022-11-10
Update Helm Chart for running API Gateways in Agentless config and with external servers

t-eckert:
Changes proposed in this PR:
- Add a key to the Helm chart `externalServers.httpPort`.
- Set `CONSUL_HTTP_ADDR` to the first external server host when `externalServers` is enabled.
- Set `CONSUL_HTTP_ADDR` using Kube DNS when running with local servers.
- Change addresses to not have `http(s)://` and instead set this using the [`CONSUL_HTTP_SSL`](https://developer.hashicorp.com/consul/commands#consul_http_ssl) environment variable.
- Add environment variable to API Gateway controller deployment for `CONSUL_TLS_SERVER_NAME` to support SNI with external servers.
- Set the `consul.address` for the GatewayClassConfig whether using external or local servers.
- Set ports for the three protocols when running with external servers.
- Add the BATS tests for these changes.

How I've tested this PR:
- [x] BATS!
- [x] Deployed API Gateways `0.4.0` with Consul-k8s `1.0.0-beta5` (Agentful) and local servers
- [x] Deployed API Gateways `0.5.0` with Consul-k8s `1.0.0-beta5` (Agentful) and local servers
- [x] Deployed API Gateways `0.4.0` with Consul-k8s `1.0.0-beta5` (Agentless) and local servers
- [x] Deployed API Gateways `0.5.0` with Consul-k8s `1.0.0-beta5` (Agentless) and local servers
- [x] Deployed API Gateways `0.4.0` with Consul-k8s `1.0.0-beta5` (Agentful) and external servers
- [x] Deployed API Gateways `0.5.0` with Consul-k8s `1.0.0-beta5` (Agentful) and external servers
- [x] Deployed API Gateways `0.5.0` with Consul-k8s `1.0.0-beta5` (Agentless) and external servers

Also verified that "Deployed API Gateways `0.4.0` with Consul-k8s `1.0.0-beta5` (Agentless) and external servers" does not work as is expected.

- BATS!

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -82,6 +82,12 @@ spec:
           value: {{ template "consul.fullname" . }}-server:8500
           {{- end }}
           {{- end }}
+        - name: CONSUL_HTTP_SSL
+          value: {{ .Values.global.tls.enabled }}
+        {{- if and .Values.externalServers.enabled .Values.externalServers.tlsServerName }}
+        - name: CONSUL_TLS_SERVER_NAME
+          value: {{ .Values.externalServers.tlsServerName }}
+        {{- end }}
         command:
         - "/bin/sh"
         - "-ec"
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -69,10 +69,18 @@ spec:
           value: "/consul/login/acl-token"
         {{- end }}
         - name: CONSUL_HTTP_ADDR
+          {{- if .Values.externalServers.enabled }}
           {{- if .Values.global.tls.enabled }}
-          value: https://$(HOST_IP):8501
+          value: {{ first .Values.externalServers.hosts }}:{{ .Values.externalServers.httpsPort }}
           {{- else }}
-          value: http://$(HOST_IP):8500
+          value: {{ first .Values.externalServers.hosts }}:{{ .Values.externalServers.httpPort }}
+          {{- end }}
+          {{- else }}
+          {{- if .Values.global.tls.enabled }}
+          value: {{ template "consul.fullname" . }}-server:8501
+          {{- else }}
+          value: {{ template "consul.fullname" . }}-server:8500
+          {{- end }}
           {{- end }}
         command:
         - "/bin/sh"
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -71,12 +71,11 @@ spec:
         - name: CONSUL_HTTP_ADDR
           {{- if .Values.client.enabled }}
           {{/*
-            We use client agent nodes if we have them to support backwards compatibility in <=0.4 releases which
-            require connectivity between the registered Consul agent node and a deployment for health checking
-            (originating from the Consul node). Always leveraging the agents in the case that they're explicitly
-            opted into allows us to support users with agent node + "externalServers" configuration upgrading a
-            helm chart without upgrading api gateways. Otherwise, using "externalServers" when provided
-            without local agents will break gateways <=0.4.
+            We use client agent nodes if we have them to support backwards compatibility for Consul API Gateway
+            v0.4 and older, which requires connectivity between the registered Consul agent node and a
+            deployment for health checking (originating from the Consul node). Always leveraging the agents in
+            the case that they're explicitly opted into allows us to support users with agent node +
+            "externalServers" configuration upgrading a Helm chart without upgrading API gateways.
           */}}
           {{- if .Values.global.tls.enabled }}
           value: $(HOST_IP):8501
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -83,7 +83,7 @@ spec:
           {{- end }}
           {{- end }}
         - name: CONSUL_HTTP_SSL
-          value: {{ .Values.global.tls.enabled }}
+          value: "{{ .Values.global.tls.enabled }}"
         {{- if and .Values.externalServers.enabled .Values.externalServers.tlsServerName }}
         - name: CONSUL_TLS_SERVER_NAME
           value: {{ .Values.externalServers.tlsServerName }}
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -94,7 +94,7 @@ spec:
           {{- else }}
           {{/*
             We have local network connectivity between deployments and the internal cluster, this
-            should be supported in all versions of api-gateway
+            should be supported in all versions of Consul API Gateway
           */}}
           {{- if .Values.global.tls.enabled }}
           value: {{ template "consul.fullname" . }}-server:8501
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -84,7 +84,8 @@ spec:
           {{- end }}
           {{- else if .Values.externalServers.enabled }}
           {{/*
-            "externalServers" specified and running in "agentless" mode, this will only work 0.5+
+            "externalServers" specified and running in "agentless" mode, this will only work with
+            Consul API Gateway v0.5 or newer
           */}}
           {{- if .Values.global.tls.enabled }}
           value: {{ first .Values.externalServers.hosts }}:{{ .Values.externalServers.httpsPort }}
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -69,13 +69,19 @@ spec:
           value: "/consul/login/acl-token"
         {{- end }}
         - name: CONSUL_HTTP_ADDR
-          {{- if .Values.externalServers.enabled }}
+          {{- if and .Values.externalServers.enabled (not .Values.client.enabled) }}
           {{- if .Values.global.tls.enabled }}
           value: {{ first .Values.externalServers.hosts }}:{{ .Values.externalServers.httpsPort }}
           {{- else }}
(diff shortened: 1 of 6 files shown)


charts/consul/templates/api-gateway-gatewayclassconfig.yaml
```diff
@@ -11,6 +11,11 @@ metadata:
     component: api-gateway
 spec:
   consul:
+    {{- if .Values.externalServers.enabled }}
+    address: {{ first .Values.externalServers.hosts }}
```
andrewstucki, line 15: Actually, thinking this through a bit more, I don't think we should fully reduce what the previous PR had.

Specifically, our last release (0.4) supported agent-based "externalServers" configurations because it still just talked to the local agent which could directly health-check the deployments. This change means that that setup would potentially no longer work because of the lack of network connectivity between Consul and the deployment in a setup where the deployment is in some sort of restricted network segment.

I think what this means is that we should _only_ be leveraging `externalServers` if both `externalServers` is specified and you're running Agentless -- which would still give us the ability to support 0.4 with `externalServers` but still just say, "You want to run agentless? You'll need to upgrade your gateways."

andrewstucki: You should be able to verify that this behavior would break here by deploying `hashicorp/consul-api-gateway:0.4` with 0.49 and 1.0.0-beta5 (i.e. this change) with node-based agents + external servers configurations.

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -69,11 +69,25 @@ spec:
...
+          {{- end }}
+          {{- else }}
+          {{- if .Values.global.tls.enabled }}
+          value: {{ template "consul.fullname" . }}-server:8501
+          {{- else }}
+          value: {{ template "consul.fullname" . }}-server:8500
           {{- end }}
+          {{- end }}
+        - name: CONSUL_HTTP_SSL
+          value: {{ .Values.global.tls.enabled }}
```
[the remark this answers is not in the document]
t-eckert: I added quotes to ensure that it does get formatted as a string. I also added BATS for it.

charts/consul/templates/api-gateway-gatewayclassconfig.yaml
t-eckert: Sounds good! I added the check for clients. I didn't add the checks that you had for `client.enabled` being set to `"-"` because the expected value is a boolean. I figure if the user is setting the value to `"-"`, they are holding the Helm chart wrong. But I am willing to add the check for `"-"` equivalency if you think that is needed.

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -69,11 +69,31 @@ spec:
...
+          {{- else }}
+          value: $(HOST_IP):8500
           {{- end }}
+          {{- else }} # Internal servers, no clients
+          {{- if .Values.global.tls.enabled }}
+          value: {{ template "consul.fullname" . }}-server:8501
+          {{- else }}
+          value: {{ template "consul.fullname" . }}-server:8500
+          {{- end }}
+          {{- end }}
```
andrewstucki, line 90: Thinking that this logic could probably be simplified to something like this for clarity (along with comments):

```
{{- if .Values.client.enabled }}
{{/*
We use client agent nodes if we have them to support backwards compatibility in <=0.4 releases which
require connectivity between the registered Consul agent node and a deployment for health checking.
Always leveraging the agents in the case that they're explicitly opted into allows us to support users
with agent node + "externalServers" configuration upgrading a helm chart without upgrading api
gateways. Otherwise, always specifying "externalServers" when provided will break any gateways <=0.4.
*/}}
value: $(HOST_IP):8501
{{- else }}
value: $(HOST_IP):8500
{{- end }}
{{- else if .Values.externalServers.enabled }}
{{/*
"externalServers" specified and running in "agentless" mode, this will only work 0.5+
*/}}
value: {{ first .Values.externalServers.hosts }}:{{ .Values.externalServers.httpsPort }}
{{- else }}
value: {{ first .Values.externalServers.hosts }}:{{ .Values.externalServers.httpPort }}
{{- end }}
{{- else }}
{{/*
We have local network connectivity between deployments and the internal cluster, this
should be supported in all versions of api-gateway
*/}}
value: {{ template "consul.fullname" . }}-server:8501
{{- else }}
value: {{ template "consul.fullname" . }}-server:8500
{{- end }}
{{- end }}
```

charts/consul/templates/api-gateway-gatewayclassconfig.yaml
andrewstucki: Cool, sounds good. I added a quick comment about simplifying the logic to avoid some of the compound booleans for clarity + comments, but generally this looks like you're capturing the backwards-compat case we were talking about.

charts/consul/templates/api-gateway-controller-deployment.yaml
t-eckert: Nice! I really like this reworking.

charts/consul/test/unit/api-gateway-gatewayclassconfig.bats
```diff
@@ -66,3 +66,92 @@ load _helpers
...
+  local actual=$(helm template \
+      -s templates/api-gateway-gatewayclassconfig.yaml \
+      --set 'apiGateway.enabled=true' \
+      --set 'apiGateway.image=foo' \
+      . | tee /dev/stderr |
+      yq '.spec.consul.address == "release-name-consul-server"' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "apiGateway/GatewayClassConfig: Consul server address set with local servers and clients." {
```
andrewstucki, line 98: Sorry, just since we have pretty clear expectations around which modes we work in for all of this, can we add a test for external servers + node clients (the mode we want to support for being backwards compatible with 0.4)? I think that's actually the test case that we discovered the bug in when we were verifying all of this manually.

charts/consul/templates/api-gateway-gatewayclassconfig.yaml
```diff
@@ -11,6 +11,13 @@ metadata:
     component: api-gateway
 spec:
   consul:
+    {{- if .Values.client.enabled }}
```
andrewstucki, line 14: Think it's worth adding in the comments from the deployment file here too? Either way is fine with me, but maybe just for consistency?

andrewstucki (approved):
One last change to add a test and then I'm good on feedback at this point, though you should probably wait for someone from the k8s team to sign off too.

charts/consul/templates/api-gateway-gatewayclassconfig.yaml
t-eckert: Sure! More clarity is better.

charts/consul/test/unit/api-gateway-gatewayclassconfig.bats
t-eckert: Added!

charts/consul/test/unit/api-gateway-gatewayclassconfig.bats
```diff
@@ -66,3 +66,108 @@ load _helpers
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/api-gateway-gatewayclassconfig.yaml \
+      --set 'apiGateway.enabled=true' \
+      --set 'apiGateway.image=foo' \
+      --set 'externalServers.enabled=true' \
+      --set 'externalServers.hosts[0]=external-consul.host' \
+      --set 'server.enabled=false' \
+      . | tee /dev/stderr |
+      yq '.spec.consul.address == "external-consul.host"' | tee /dev/stderr)
```
andrewstucki, line 98:
```suggestion
      --set 'server.enabled=false' \
      --set 'client.enabled=true' \
      . | tee /dev/stderr |
      yq '.spec.consul.address == "$(HOST_IP)"' | tee /dev/stderr)
```

andrewstucki (changes requested):

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -69,11 +69,46 @@ spec:
...
           {{- if .Values.global.tls.enabled }}
-          value: https://$(HOST_IP):8501
+          value: $(HOST_IP):8501
           {{- else }}
-          value: http://$(HOST_IP):8500
+          value: $(HOST_IP):8500
+          {{- end }}
+          {{- else if .Values.externalServers.enabled }}
+          {{/*
+            "externalServers" specified and running in "agentless" mode, this will only work 0.5+
```
mikemorris, line 88:
```suggestion
            "externalServers" specified and running in "agentless" mode, this will only work with
            Consul API Gateway v0.5 or newer
```

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -69,11 +69,46 @@ spec:
...
+          */}}
+          {{- if .Values.global.tls.enabled }}
+          value: {{ first .Values.externalServers.hosts }}:{{ .Values.externalServers.httpsPort }}
+          {{- else }}
+          value: {{ first .Values.externalServers.hosts }}:{{ .Values.externalServers.httpPort }}
           {{- end }}
+          {{- else }}
+          {{/*
+            We have local network connectivity between deployments and the internal cluster, this
+            should be supported in all versions of api-gateway
```
mikemorris, line 98:
```suggestion
            should be supported in all versions of Consul API Gateway
```

charts/consul/values.yaml
```diff
@@ -2876,6 +2879,8 @@ apiGateway:
   enabled: false
 
   # Image to use for the api-gateway-controller pods and gateway instances
+  #
+  # ~> **Note:** Using API Gateway <= 0.4 with external servers requires setting `client.enabled: true`.
```
mikemorris, line 2883: 👍

mikemorris (approved):
Minor suggested tweaks to the comments for clarity, overall LGTM and feels like we'll have a good backwards compatibility/upgrade path story with this implemented.

andrewstucki (approved):

thisisnotashwin (approved):

PR #1699, opened 2022-11-10
Explicitly disable grpc_tls when not using encryption.

hashi-derek:
In Consul 1.14, the grpc_tls port will be enabled by default on servers. This explicitly disables the port if encryption is not configured so that the defaults do not affect installations.


thisisnotashwin (approved):

curtbushko (approved):
Looks good to me!

PR #1719, opened 2022-11-15
Replace httpPort with httpsPort to avoid confusion.

thisisnotashwin:
Changes proposed in this PR:
- Unify usage of httpsPort across the project to avoid confusion as `httpPort` is used solely for the API gateway and setting this value would lead to confusion amongst end users who would expect all consul-k8s components to use that value and they wouldn't.

How I've tested this PR:
- BATS

- 👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


t-eckert (approved):
This looks good. I understand the desire to avoid confusion here.

curtbushko (approved):

PR #1727, opened 2022-11-16
CHANGELOG updates for 1.0

curtbushko:
Changes proposed in this PR:
- I went back through all the PRs until the last release in October when the last check was done.

How I've tested this PR:

:eyes:

Checklist:
- [ ] Tests added
- [X] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ BREAKING CHANGES:
     Previously, `dns.enabled` defaulted to the value of `global.enabled` and `dns.enableRedirection` defaulted to the
     value to `false`. [[GH-1688](https://github.com/hashicorp/consul-k8s/pull/1688)]
   * Remove `global.imageEnvoy` and replace with `global.imageConsulDataplane` for running the sidecar proxy.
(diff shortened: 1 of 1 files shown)


david-yu:
Thanks, could you pull in my latest changelog addition from the UBI image update? I think I also had a formatting error on the link to the PR. 

Also, I assume this is what the 1.0 contents would look like? Do we intend to remove the Beta1-5 changelings? That is typically what Core does and @ishustava seemed to suggest we do the same.

curtbushko:
@david-yu I pulled in your UBI change.

I am not too sure about cleaning up the changelog and removing Beta1-5. This pass was to make sure that we caught all the content from PRs in the last while.

david-yu:
@curtbushko I see. Perhaps we can do the cleanup in this PR too. Lets wait until @ishustava comes back for air to see if there is alignment on cleaning up the changelog.

david-yu (approved):

david-yu:
Will approve for now, we can do a separate PR to move/remove beta release notes if needed.

thisisnotashwin (approved):

PR #1731, opened 2022-11-16
default to dataplane being supported when checking version annotation

ndhanushkodi:
Changes proposed in this PR:
- Default to consul-dataplane being supported even when the version annotation is malformed. If the version annotation exists, it already means we're running consul-k8s 1.0.0+.

Eric ran into this when using an image built from our pipeline that builds images from main. It has a version formatted like "v1.0.0-dev (<GITSHA>)" which is not a recognized version by the client health checks code and causes endpoints controller to think that consul-dataplane is not supported and attempt to register client health checks which won't work. 

We would have run into this with our released image as well since they are built the same way as the images from main. We did NOT run into this in acceptance tests because the images built for acceptance tests are built differently and do not have the (<GITSHA>) part.

In the future we will fix our pipelines to include versions that can be parsed by the go-version library, but talking to @ishustava, we don't want to do that so close to the release. 

Checklist:
- [x] Tests added
~- [ ] CHANGELOG entry added~


curtbushko (approved):
Nice!

ishustava (approved):

PR #1732, opened 2022-11-17
Add opt-in server discovery for API Gateway Controller

andrewstucki:
Changes proposed in this PR:

For backwards compatibility with old helm charts where we only ever talk to the agent nodes which don't have gRPC APIs for Consul dataplane, we need to use our old login/logout flow and the static agent node address rather than leveraging the consul server connection manager. This PR coincides with the changes introduced in https://github.com/hashicorp/consul-api-gateway/pull/454.

How I've tested this PR:

Manually verified in explicitly agent-based and agentless installation modes and added unit tests.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -119,6 +119,10 @@ spec:
           value: {{ .Values.global.adminPartitions.name }}
         {{- end }}
         {{- end }}
+        {{- if not .Values.client.enabled }}
+        - name: CONSUL_DYNAMIC_SERVER_DISCOVERY
+          value: "true"
+        {{- end }}
         command:
         - "/bin/sh"
         - "-ec"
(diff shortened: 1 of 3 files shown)


charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -119,6 +119,10 @@ spec:
           value: {{ .Values.global.adminPartitions.name }}
         {{- end }}
         {{- end }}
+        {{- if not .Values.client.enabled }}
```
andrewstucki, line 122: If clients are enabled we still leverage `$(HOST_IP)` -- so in any other case (i.e. cluster-based or external servers) we know we should use the server gRPC discovery components. This is also passed explicitly as an env variable to support older versions of api-gateway rather than having the entrypoints fail with an unknown flag.

ndhanushkodi (approved):

mikemorris (approved):
To be clear this isn't really "opt-in" from the end-user UX, as it just passes a flag to the binary based on other existing settings in the chart (which will default to clients disabled in the upcoming v1.0 release), and therefore enable the gRPC server discovery functionality on supporting versions of Consul API Gateway (v0.5 and newer).

PR #1733, opened 2022-11-17
fix acls when peering and partitions are both enabled

ndhanushkodi:
Changes proposed in this PR:
- On a non-default partition, this rule errors with this log on server-acl-init
```
Failure: creating mesh-gateway-policy policy: err="Unexpected response code: 500 (rpc error making call: Invalid policy rules: partitioned policy cannot use partition_prefix rules)"
```

This only happens on a non-default partition when peering and partitions are both enabled, so I found this when testing non-default partitions with ACLs. So this adds a condition to the partition_prefix rule

How I've tested this PR:
manually, on non-default partitions, acceptance tests for other cases

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1273,9 +1273,6 @@ workflows:
           context: consul-ci
           requires:
           - dev-upload-docker
-      - acceptance-gke-1-23:
-          requires:
-            - dev-upload-docker
 
   nightly-cleanup:
     triggers:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1273,6 +1273,9 @@ workflows:
           context: consul-ci
           requires:
           - dev-upload-docker
+      - acceptance-gke-1-23:
+          requires:
+            - dev-upload-docker
```
ndhanushkodi, line 1278: Will remove before merging.

thisisnotashwin (approved):

PR #1740, opened 2022-11-18
remove controller webhook cert roles from vault test and the values.yaml

kschoche:
The CRD controller deployment no longer exists as it was merged into connect-inject.

Changes proposed in this PR:
- Removes the webhook cert roles for the controller from the vault test
- updates values.yaml to not reference this field

How I've tested this PR:
👀 

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


thisisnotashwin (approved):
god's work and all that 🥹

wilkermichael (approved):

Issue #1256, opened 2022-06-02
Configure Consul agents, Envoy Proxy, and snapshotAgent log levels using global.Loglevel configuration


Ranjandas:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Feature Description

Currently, the `global.Loglevel` in the values file only seems to set the log level for Consul-K8S Control-Plane components. This is a feature request to use the same key to set the log level for 

- Consul Servers - currently no logLevel value exists and logLevel is passed through extraConfig
- Consul Clients - currently no logLevel value exists and logLevel is passed through extraConfig
- Gateways (Mesh, Ingress, Terminating). As an example the logLevel could be set here for Mesh Gateways [link](https://github.com/hashicorp/consul-k8s/blob/405e36f2199dc99e7b2ada6fb1e48c5daa5399a2/charts/consul/templates/mesh-gateway-deployment.yaml#L326)
- Sidecars - currently a [logLevel](https://www.consul.io/docs/k8s/helm#v-connectinject-loglevel) value is set explicitly instead of globally but could default to this global value. 
- Snapshot Agent - currently logLevel is set by a Kubernetes secret: https://www.consul.io/docs/k8s/helm#v-client-snapshotagent 

Log level for servers, clients, gateways, sidecars, and the snapshotAgent should be set by this global value and also overridable via a Helm value for each of these components to allow for a way to customize log levels. Documentation for the `global.logLevel` value should also describe this new behavior.

Log levels for `global.logLevel` should only support the following values, otherwise an error must be presented upon `helm install` or `helm upgrade`: trace, debug, info, warn, and error 

#### Log Level Reference 

Consul K8s Control Plane:  trace, debug, info, warn, and error (see [acl-init](https://github.com/hashicorp/consul-k8s/blob/b9452e659c488b5c86bff02e3d58155df93e4579/control-plane/subcommand/acl-init/command.go#L98))
[...]

PR #1741, opened 2022-11-18
Add trace to logLevel

im2nguyen:
Changes proposed in this PR:
- Update Helm chart docs to include trace logLevel

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -11,7 +11,7 @@ global:
 
   # The default log level to apply to all components which do not otherwise override this setting.
   # It is recommended to generally not set this below "info" unless actively debugging due to logging verbosity.
-  # One of "debug", "info", "warn", or "error".
+  # One of "trace", "debug", "info", "warn", or "error".
   # @type: string
   logLevel: "info"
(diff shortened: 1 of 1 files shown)


kschoche:
Hi @im2nguyen! 
Thanks for filing this PR!
The reason that `trace` does not exist in our docs here is that these log levels are applied to two different logging systems: `hclog` (most things) and `zap` (for controllers that use controller-runtime, eg connect-injector), and Zap does not support `trace`:
https://github.com/uber-go/zap/blob/ed5598a9e26d75aae01f4846ce31b3fe785c57c7/level.go#L29-L49.

We might be able to get trace into some of these fields now that we've consolidated some of the components, but definitely not `global.LogLevel`

im2nguyen:
Thanks! For the time being I'll close this PR then 😄

david-yu:
This is the issue that we need to address to get this working as intended to align with Consul log levels: https://github.com/hashicorp/consul-k8s/issues/1256

PR #1743, opened 2022-11-18
Add fix for api-gateway when using system-wide trusted CAs for external servers

andrewstucki:
Changes proposed in this PR:

This adds some logic that was forgotten around leveraging `externalServers.useSystemRoots`. Without it, attempting to use something like HCP as an external server without agents causes the gateway controller to fail validating the Consul server connection since it's using the wrong certificate authority.

How I've tested this PR:

Validated against HCP with the api-gateway changes in https://github.com/hashicorp/consul-api-gateway/pull/459

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+BUG FIXES:
+* Helm:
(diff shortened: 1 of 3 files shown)


curtbushko (approved):
Looks good!

t-eckert (approved):

PR #1753, opened 2022-11-28
Fix TLS Cert when using `enableAutoEncrypt`

t-eckert:
Changes proposed in this PR:
- The generated cert will always mount to `consul-ca-cert`
- The controller will use the cert at `consul-ca-cert` to talk to servers if running without clients. It will use `consul-auto-encrypt-ca-cert` to talk to clients only if running _with_ clients.

How I've tested this PR:
- Updated BATS
- Deployed successfully with the following configurations, verifying that the controller received the correct cert and was able to talk to the server or client depending on the configuration.
  - `tls.enableAutoEncrypt=false` `client.enabled=false` `apiGateway.image=...0.5.1`
  - `tls.enableAutoEncrypt=true` `client.enabled=false` `apiGateway.image=...0.5.1`
  - `tls.enableAutoEncrypt=false` `client.enabled=true` `apiGateway.image=...0.5.1`
  - `tls.enableAutoEncrypt=true` `client.enabled=true` `apiGateway.image=...0.5.1`
  - `tls.enableAutoEncrypt=true` `client.enabled=true` `apiGateway.image=...0.4.0`

- BATS
- Testing that the configuration that originally caught this bug works correctly:
```yaml
global:
  logLevel: "trace"
  acls:
    manageSystemACLs: true
  image: hashicorp/consul:1.14.1
  imageK8S: hashicorp/consul-k8s-control-plane:1.0.1
  tls:
    enabled: true
    enableAutoEncrypt: true
apiGateway:
  enabled: true
  image: hashicorp/consul-api-gateway:0.5.1
  logLevel: debug
  managedGatewayClass:
    enabled: true
    serviceType: LoadBalancer
connectInject:
  enabled: true
controller:
  enabled: true
server:
  replicas: 1
client:
  enabled: true
```

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -243,11 +243,7 @@ spec:
           name: consul-data
```
ishustava, line 243: nit: you probably don't need `HOST_IP` env variable on this container anymore because it's not used anymore

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -243,11 +243,7 @@ spec:
...
           readOnly: false
         {{- if .Values.global.tls.enabled }}
```
ishustava, line 245: I think this needs similar logic for not mounting the ca cert if external servers with `useSystemRoots` are enabled (like on line 180)

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -243,11 +243,7 @@ spec:
...
-        {{- if .Values.global.tls.enableAutoEncrypt }}
-        - name: consul-auto-encrypt-ca-cert
-        {{- else }}
         - name: consul-ca-cert
-        {{- end }}
           mountPath: /consul/tls/ca
           readOnly: true
         {{- end }}
```
ishustava, line 249: nit: `-datacenter`, `-partition` and `-api-timeout` flag values below are duplicated because they're already provided as env vars `{{- include "consul.consulK8sConsulServerEnvVars" . | nindent 8 }}` I believe. It'd be nice if we removed them to avoid any future weirdness.

ishustava (approved):

charts/consul/templates/api-gateway-controller-deployment.yaml
andrewstucki: 👍 without the mount skip, this seems to cause problems when `useSystemRoots` are enabled due to the cert secrets referenced here not existing. I'm assuming it means we don't bootstrap the CA/self-signed cert when that flag is enabled since we know we aren't going to use it?

charts/consul/templates/api-gateway-controller-deployment.yaml
t-eckert: Thanks for catching that!

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -150,7 +150,7 @@ spec:
           mountPath: /consul-bin
         {{- end }}
         {{- if .Values.global.tls.enabled }}
-        {{- if .Values.global.tls.enableAutoEncrypt }}
+        {{- if and .Values.client.enabled .Values.global.tls.enableAutoEncrypt }}
         - name: consul-auto-encrypt-ca-cert
         {{- else }}
         - name: consul-ca-cert
```
andrewstucki, line 156: So, like mentioned below (in the init portion), if we _do_ want to fix the behavior with `useSystemRoots` in this PR (or I think we can save it for a subsequent PR), I believe we'll have to drop this from being referenced if and only if we have: 1) `externalServers.useSystemRoots` set to `true` and 2) `client.enabled` set to `false`. In that case, I don't believe we'll actually ever create the `consul-ca-cert` secret.

Without those changes to the conditional mount we get the following issue on install (surfaced in a recent HCP configuration when we were trying to update the learn guide and dropping what we thought were unused options):

> │ Error: Deployment.apps "consul-api-gateway-controller" is invalid: [spec.template.spec.containers[0].volumeMounts[1].name: Not found: "consul-ca-cert", spec.template.spec.initContainers[1].volumeMounts[1].name: Not found: "consul-ca-cert"]
│

t-eckert: I fixed it for the init container and I need to fix it for the _container_ container.

charts/consul/templates/api-gateway-controller-deployment.yaml
ishustava: We only bootstrap the cert if servers are enabled (https://github.com/hashicorp/consul-k8s/blob/ef421f968770e96ee2cbbb471423169b579aa1de/charts/consul/templates/tls-init-job.yaml#L1). In this case, servers are external so we wouldn't bootstrap it either way. We'd either expect to be provided via a secret or ignore it if `useSystemRoots` is `true`

charts/consul/templates/api-gateway-controller-deployment.yaml
t-eckert: Fixed in both places. I'll do more testing to confirm in the morning.

charts/consul/templates/api-gateway-controller-deployment.yaml
t-eckert: Great! I did this apart from the `-datacenter` flag. I noticed that `acl-init` is using Consul datacenter instead of Consul login datacenter and if this is removed, the acl init will use the local dc. I will follow up on whether we need to change `acl-init` under the hood to use Consul login datacenter instead.

curtbushko (approved):
Looks good!

PR #1759, opened 2022-11-30
Mount certs when using clients even with external servers

t-eckert:
Changes proposed in this PR:
- Mount autoencrypt certs when using clients
- Fix changelog

How I've tested this PR:
- BATS

- BATS

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -149,7 +149,7 @@ spec:
         - name: consul-bin
           mountPath: /consul-bin
         {{- end }}
-        {{- if not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots) }}
+        {{- if or (not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots)) .Values.client.enabled }}
```
t-eckert, line 152: This fix is repeated in 3 spots. It allows the use of certs to talk to local clients when connected to an external server like HCP.

curtbushko (approved):
Approving so that we can get it in the release for tomorrow

t-eckert:
@andrewstucki I will be watching this to see if there are any issues. If you don't see any, feel free to do the merge.

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -240,7 +240,7 @@ spec:
         - mountPath: /consul/login
           name: consul-data
           readOnly: false
-        {{- if not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots) }}
+        {{- if or (not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots)) .Values.client.enabled }}
```
ishustava, line 243: This init container should only talk to servers now, so we probably don't need this additional `or` here.

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -179,7 +179,7 @@ spec:
         emptyDir: { }
       {{- end }}
       {{- if .Values.global.tls.enabled }}
-      {{- if not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots) }}
+      {{- if or (not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots)) .Values.client.enabled }}
```
ishustava, line 182: If I'm reading this logic correctly, this will add the `consul-ca-cert` volume when clients are enabled regardless of whether  `useSystemRoots` is set. I don't think we want to do that. When `useSystemRoots` is true, you probably haven't provided the CA cert secret that this volume needs, and your deployment will fail.

andrewstucki: So just to make sure of the behavior then @ishustava -- when you have `client.enabled` and are using TLS, but also using system roots with external servers, is it possible to configure helm to use anything other than auto-encrypt?

For example, if someone was using something like HCP could they bring their own root that they want the client node certificates signed with? Because that would seem to be the case that this would cover -- making sure that whatever the root used in configuring the client nodes was able to be mounted to establish connections to clients. If not, it sounds like we only ever support auto-encrypt for client nodes when using publicly trusted certs for external servers?

ishustava: > For example, if someone was using something like HCP could they bring their own root that they want the client node certificates signed with?

This is not something we support with HCP. HCP only supports auto-encrypt. Plus, we provide a separate CA cert for RPC comms that client should be using to talk to servers. And so in the case of HCP, you would provide the secret and the volume will be mounted (although it won't be used for the api gw deployment because there's nothing here that uses RPC APIs). 

I'm thinking of a case when you created a cert for the servers that is signed by a well-known CA. In that case, you probably don't have access to the CA key, and so you can only use auto-encrypt for client certs. You would then configure your helm chart to use external servers with `useSystemRoots` and auto-encrypt enabled. I can't think of a use case when someone would need to use `useSystemRoots` without auto-encrypt otherwise.

andrewstucki: Got it, that makes sense here then -- we'll still need some variation of the guard above for actually mounting the auto-encrypt cert regardless of `useSystemRoots` when clients are enabled in the actual controller container, but given that this case: `useSystemRoots` + clients + no auto-encrypt isn't actually something we support, I'm thinking that as a follow-up to this we should explicitly disallow that configuration.

I'll try and push the recommended changes here to drop the "bring your own cert" case.

t-eckert: Thank you, @andrewstucki. Feel free to take control of this branch.

charts/consul/templates/api-gateway-controller-deployment.yaml
andrewstucki: Update: this is added only here in order to have an auto-encrypt-based cert mounted to establish proper client-node communication if clients are explicitly enabled (in which case our controller still leverages the agent node) -- otherwise we'll try and use the system roots to verify a connection to the client node with a cert provisioned through the auto-encrypt process and the TLS handshake will fail.

ishustava:
@andrewstucki looks like this one is a flake. It should be safe to re-run, and if it passes, we're good.

ishustava (approved):

Issue #759, opened 2021-10-04
Custom labels for connectInject, acl init, and mesh gateway pods via Helm


david-yu:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Feature Description

This tracks a feature request to provide the ability to apply custom K8s labels on component that are deployed by the Consul K8s helm chart (some of which are already available today). The feature request describes the ability to set pod labels globally using a single config i.e. `global.extraLabels` for the Pods that are deployed by the Helm chart. 

Currently the pods with missing labels include: 
- connect inject
- ~~catalog sync~~
- acl init
- mesh gw

#### Use Case(s)

This allows practitioners to clearly organize which components of a Helm chart are deployed via the Consul K8s helm chart for either compliance reasons or for tracking purposes.

jeanmorais:
@david-yu I've opened a PR to solve this issue. I ended up not including the catalog sync as the `Deployment` already has custom labels ([see here](https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/sync-catalog-deployment.yaml#L29)).

david-yu:
Thank you @jeanmorais! I'll ask someone to review later this sprint. You are correct about catalog sync!

david-yu:
Closing as mentioned in the PR, a better approach would be to do provide global labels instead.

PR #1678, opened 2022-11-04
Allow to apply additional labels to the connect inject pods

ilpianista:
Changes proposed in this PR:
- Allow to apply additional labels to the connect inject pods

How I've tested this PR:
We run this in our production.

How I expect reviewers to test this PR:
I've added a unit tests.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #1771, opened 2022-12-07
Add global.extraLabels values.yaml setting

brucec5:
Currently, there isn't a way to add a set of labels to every pod (see #759 for example). This PR adds a global values.yaml setting that lets you specify labels to apply to every pod/deployment/daemonset/statefulset/job created by the helm chart. This is in addition to the existing `extraLabels` config for things like clients and servers.

Changes proposed in this PR:
- Add in `global.extraLabels` values.yaml setting
- Use that setting to apply extra labels to all pods/deployments/daemonsets/statefulsets/jobs created by the consul-k8s helm chart
- bats tests, I guess

How I've tested this PR:

Added bats tests to cover testing extra labels for all applicable resources

Run the tests, maybe deploy the helm chart locally just to make sure it works

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/sync-catalog-deployment.yaml
+++ b/charts/consul/templates/sync-catalog-deployment.yaml
@@ -32,6 +32,9 @@ spec:
         {{- if .Values.syncCatalog.extraLabels }}
           {{- toYaml .Values.syncCatalog.extraLabels | nindent 8 }}
         {{- end }}
+        {{- if .Values.global.extraLabels }}
+          {{- toYaml .Values.global.extraLabels | nindent 8 }}
+        {{- end }}
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
         {{- if .Values.syncCatalog.annotations }}
--- a/charts/consul/templates/api-gateway-controller-deployment.yaml
+++ b/charts/consul/templates/api-gateway-controller-deployment.yaml
@@ -46,6 +46,9 @@ spec:
         chart: {{ template "consul.chart" . }}
         release: {{ .Release.Name }}
         component: api-gateway-controller
+        {{- if .Values.global.extraLabels }}
+          {{- toYaml .Values.global.extraLabels | nindent 8 }}
+        {{- end }}
     spec:
       serviceAccountName: {{ template "consul.fullname" . }}-api-gateway-controller
       containers:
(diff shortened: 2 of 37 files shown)


david-yu:
Thanks @brucec5 ! We'll take a look!

david-yu:
Out of curiosity are only global labels needed, as in the label should be the same values across all components?

@ilpianista submitted a PR for adding some labels before. @ilpianista would using `global.extralLabels` work instead of setting specific component labels like you proposed here? https://github.com/hashicorp/consul-k8s/pull/1678

brucec5:
Basically, we will eventually have a Gatekeeper/OPA rule set up that mandates all pods and deployments have a certain set of labels applied. I guess I could see exposing this more granularly such that every component has `extraLabels` rather than having a single global set of extra labels.

david-yu:
Ok I just wanted to check. I don't think we need the extraneous `extraLabels` since my experience tells me you probably want to only set the same labels across all components for tracking, so I would think removing the extra `extraLabels` that existed before should be fine.

ilpianista:
> @ilpianista submitted a PR for adding some labels before. @ilpianista would using `global.extralLabels` work instead of setting specific component labels like you proposed here? #1678

`global.extraLabels` would suit our use case. Yes.

charts/consul/templates/sync-catalog-deployment.yaml
```diff
@@ -32,6 +32,9 @@ spec:
         {{- if .Values.syncCatalog.extraLabels }}
           {{- toYaml .Values.syncCatalog.extraLabels | nindent 8 }}
         {{- end }}
+        {{- if .Values.global.extraLabels }}
```
thisisnotashwin, line 35: In this file, we need to add extraLabels to `.metadata.Labels` as well.

curtbushko (approved):
Thanks for doing this! This looks great!  I went over all of the yaml files and it seems that there are corresponding bats tests.

david-yu:
Thanks @brucec5 as an aside, we can take care of the removal of component level extraLabels ourselves and will likely remove them in Consul K8s 1.1.x, as this would be a breaking change.

charts/consul/templates/sync-catalog-deployment.yaml
brucec5: Ah, good catch! I wrote tests for that bit too but my bats filtering accidentally excluded that one from running...

All tests for affected files pass on my end now, at least

PR #1786, opened 2022-12-13
JIRA: Use env when creating with custom fields

david-yu:
Changes proposed in this PR:

- Adding this in towards end of create JIRA action

  ```
  env:
      JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
      JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
      JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
  -
  ```

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/jira.yaml
+++ b/.github/workflows/jira.yaml
@@ -24,16 +24,37 @@ jobs:
       - name: Set ticket type
         id: set-ticket-type
         run: |
-          echo "::set-output name=type::GH Issue"
+          if [[ "${{ github.event_name == 'pull_request' && github.event.action == 'opened' ]]; then
+            echo "::set-output name=type::PR"
+          else
+            echo "::set-output name=type::GH Issue"
+          fi
+          
       - name: Set ticket labels
         if: github.event.action == 'opened'
         id: set-ticket-labels
         run: |
           LABELS="[consul-k8s]"
           echo "::set-output name=labels::${LABELS}"
+          
+      - name: Check if team member PR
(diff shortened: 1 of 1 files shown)


.github/workflows/jira.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
 
       - name: Create ticket
         if: ( github.event.action == 'opened' && steps.set-ticket-type.outputs.type != 'Task' ) || ( github.event.action == 'opened' && steps.set-ticket-type.outputs.type == 'Task' )
-        uses: tomhjp/gh-action-jira-create@v0.1.3
+        uses: tomhjp/gh-action-jira-create@v0.2.0
```
curtbushko, line 37: Going back to 2.0?

david-yu: Yea I don't think the 0.1.3 made any difference, the issue was Auth was not working when using the action when custom fields were used in JIRA

curtbushko (approved):

.github/workflows/jira.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
     name: Jira sync
     steps:    
       - name: Login
-        uses: atlassian/gajira-login@v3
+        uses: atlassian/gajira-login@v2.0.0
```
david-yu, line 18: Seems to be some kind auth issue so going back to v2.0.0 just in case

PR #1804, opened 2022-12-19
Ignore partition/namespace on SourceIntention list to match top-level compare logic

kyhavlov:
This PR updates the logic in `MatchesConsul` for intentions to add partition/namespace to the list of ignored fields for the `SourceIntention` list to match the ignored fields on the top-level config entry struct. This is to fix an issue where upgrading from Consul pre-1.12 to 1.12+ can trigger an update of all config entries due to the partition/namespace fields in the `SourceIntention`s from Consul being filled in as `"default"` instead of `""`. 

Here's what the entry coming from consul 1.12+ looks like:
```
{
  "Kind": "service-intentions",
  "Name": "*",
  "Partition": "default",
  "Namespace": "default",
  "Sources": [
    {
      "Name": "foo",
      "Partition": "default",
      "Namespace": "default",
      "Action": "allow",
      ...
    }
  ],
  ...
}
```
And this is what consul-k8s generates via the ToConsul function:
```
{
  "Kind": "service-intentions",
  "Name": "*",
  "Sources": [
    {
      "Name": "foo",
      "Action": "allow",
      ...
    }
  ],
  ...
}
```

How I've tested this PR:
Created a consul 1.11.1+ent cluster alongside consul-k8s with a couple basic service intention config entries, then upgraded to 1.12.3+ent. Without this change there's an update to each intention because of the diff/MatchesConsul returns false.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added


ndhanushkodi (approved):
I think we just need a changelog, and are we planning to backport this? If so we'll need backport labels. 0.49.x of consul-k8s is supposed to work with 1.13.x and 1.12.x of Consul. I don't think we test with Consul 1.11.x anymore but in theory that should have worked with 0.49.x at some point. 

Thanks for fixing, it looks great to me!

david-yu:
Thanks I've added a backport label for 0.49.x and 1.0.x

wilkermichael (approved):

PR #1805, opened 2022-12-19
values.yaml - helm chart docs edits for 0.49.x

david-yu:
Changes proposed in this PR:
- Missing changes from https://github.com/hashicorp/consul/pull/15452 that were not back ported to values.yaml also see https://github.com/hashicorp/consul/pull/15849

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -299,7 +299,7 @@ global:
           {}
 
       controller:
-        # Configuration to the Vault Secret that Kubernetes will use on 
+        # Configuration to the Vault Secret that Kubernetes uses on 
         # Kubernetes CRD creation, deletion, and update, to get TLS certificates
         # used issued from vault to send webhooks to the controller.
         tlsCert:
@@ -308,7 +308,7 @@ global:
           # @type: string
           secretName: null
           
-        # Configuration to the Vault Secret that Kubernetes will use on 
+        # Configuration to the Vault Secret that Kubernetes uses on 
         # Kubernetes CRD creation, deletion, and update, to get CA certificates
         # used issued from vault to send webhooks to the controller.
         caCert:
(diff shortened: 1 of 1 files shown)


charts/consul/values.yaml
```diff
@@ -776,7 +776,7 @@ server:
   # storage classes, the PersistentVolumeClaims would need to be manually created.
   # A `null` value will use the Kubernetes cluster's default StorageClass. If a default
   # StorageClass does not exist, you will need to create one.
-  # Refer to the [Read/Write Tuning](https://www.consul.io/docs/install/performance#read-write-tuning) 
+  # Refer to the [Read/Write Tuning](/docs/install/performance#read-write-tuning)
```
nathancoleman, line 779: Will this link work?

david-yu: It should generate something like 

```
 Refer to the [Read/Write Tuning](/consul/docs/install/performance#read-write-tuning)
     section of the Server Performance Requirements documentation for considerations
```

which does work from mdx and is the convention we use in our docs.

nathancoleman: that makes sense, but I was thinking people aren't looking at a `values.yaml` file through the docs site which is why it probably had the hostname included

david-yu: What's interesting is that this pattern seems to be already be used in https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/values.yaml#L2321 and https://github.com/hashicorp/consul-k8s/blob/64e8011be5c1eaf7b32ff6d697d6f9fc71cb3699/charts/consul/values.yaml#L303. 

I need to defer to the Consul docs team as to what might be the best recommendation here.

nathancoleman: FWIW this isn't something I have a strong opinion on. Was asking more out of curiosity.

nathancoleman (approved):

charts/consul/values.yaml
im2nguyen: can we use the absolute links? this updates it to the latest absolute link

```suggestion
  # Refer to the [Read/Write Tuning](https://developer.hashicorp.com/consul/docs/install/performance#read-write-tuning) 
```

im2nguyen (approved):

PR #1821, opened 2023-01-06
Add health checks for services that are synced

thisisnotashwin:
Changes proposed in this PR:
- When the type of the service is ClusterIP, a health check will be added to the catalog registration with the health info of the service based on the state of the readiness probe of the pod associated with the service.
- Replace `apiv1` with `corev1` to be consistent across the project.
- Run `go mod tidy`.

How I've tested this PR:
- Unit tests
- Acceptance tests

- 👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/sync/sync_catalog_test.go
+++ b/acceptance/tests/sync/sync_catalog_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
+	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/stretchr/testify/require"
 )
@@ -67,6 +68,11 @@ func TestSyncCatalog(t *testing.T) {
 			require.NoError(t, err)
 			require.Len(t, service, 1)
 			require.Equal(t, []string{"k8s"}, service[0].ServiceTags)
+			filter := fmt.Sprintf("ServiceID == %q", service[0].ServiceID)
+			healthChecks, _, err := consulClient.Health().Checks(syncedServiceName, &api.QueryOptions{Filter: filter})
+			require.NoError(t, err)
+			require.Len(t, healthChecks, 1)
+			require.Equal(t, api.HealthPassing, healthChecks[0].Status)
 		})
 	}
 }
--- a/acceptance/tests/sync/sync_catalog_test.go
+++ b/acceptance/tests/sync/sync_catalog_test.go
@@ -65,7 +65,7 @@ func TestSyncCatalog(t *testing.T) {
 
 			service, _, err := consulClient.Catalog().Service(syncedServiceName, "", nil)
 			require.NoError(t, err)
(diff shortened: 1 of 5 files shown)


control-plane/catalog/to-consul/resource.go
```diff
@@ -674,6 +680,16 @@ func (t *ServiceResource) registerServiceInstance(
...
+			r.Check = &consulapi.AgentCheck{
+				CheckID:   consulHealthCheckID(endpoints.Namespace, serviceID(r.Service.Service, addr)),
+				Name:      consulKubernetesCheckName,
+				Namespace: baseService.Namespace,
+				Type:      consulKubernetesCheckType,
+				Status:    consulapi.HealthPassing,
+				ServiceID: serviceID(r.Service.Service, addr),
+				Output:    kubernetesSuccessReasonMsg,
+			}
+
```
thisisnotashwin, line 692: This is the crux of the PR

control-plane/go.sum
```diff
@@ -346,12 +346,8 @@ github.com/hashicorp/consul-k8s/control-plane/cni v0.0.0-20220831174802-b8af6526
 github.com/hashicorp/consul-server-connection-manager v0.1.0 h1:XCweGvMHzra88rYv2zxwwuUOjBUdcQmNKVrnQmt/muo=
```
thisisnotashwin, line 346: This was because of a `go mod tidy`

curtbushko (approved):
Looks good!

acceptance/tests/sync/sync_catalog_test.go
```diff
@@ -65,7 +65,7 @@ func TestSyncCatalog(t *testing.T) {
 
 			service, _, err := consulClient.Catalog().Service(syncedServiceName, "", nil)
 			require.NoError(t, err)
-			require.Equal(t, 1, len(service))
+			require.Len(t, service, 1)
```
ndhanushkodi, line 68: Would it make sense to add assertions here on the service's healthcheck, maybe just that its healthy to confirm it got synced?

ndhanushkodi (approved):
Just had a comment on potentially adding some assertions to the acceptance test, but not blocking on it! Looks great!

acceptance/tests/sync/sync_catalog_test.go
thisisnotashwin: I tried to add an assertion here but a Catalog Service did not contain health check info 😞

thisisnotashwin: Like a catalog service has HealthChecks but not AgentChecks which is what we configure which is a part of the catalog registration

thisisnotashwin: i added an assertion based on how endpoints controller tests healthchecks. 🤞

rvgud:
@thisisnotashwin just wondering, why not for nodeport services?

ishustava:
Hey @rvgud 

ClusterIP services are synced with their pod IPs and unless we sync readiness status from kube, there's no way for consul to know whether a specific pod IP is not ready.

For load balancer and NodePort services, we'll register the load balancer address or the node address respectively. When using those addresses, kube-proxy will only route to healthy pods when you dial the registered address, and so consul doesn't need to know about the health status of those.

PR #1838, opened 2023-01-17
Add health checks for services that are synced (#1821)

thisisnotashwin:
* Add health checks for services that are synced
- When the type of the service is ClusterIP, a health check will be added to the catalog registration with the health info of the service based on the state of the readiness probe of the pod associated with the service.
- Replace `apiv1` with `corev1` to be consistent across the project.
- Run `go mod tidy`.


curtbushko (approved):
Looks good!

PR #1845, opened 2023-01-20
Enable Automatic Consul Injection into Windows Workloads

joselo85:
Changes proposed in this PR:
- Created function to detect pod's OS
- Modified container_init.go and consul_dataplane_sidecar.go to assign values depending on pod's OS.
- Added Windows images fields to MeshWebhook struct in mesh_webhook.go
- Added new flags to inject-connect/command.go to set the Windows images values
- Added the default Windows image values to values.yaml and added the new config flags in connect-injector-deployment.yaml template.
- Created WindowsSupport.md documenting this process.

How I've tested this PR:
Built and pushed the required Windows images see this [guide](https://github.com/hashicorp-education/learn-consul-k8s-windows/tree/main/k8s-v1.0.x/dockerfiles#readme)   
Using `helm package` command, packaged the modified Consul Helm chart. 
Created custom values to install the packaged Helm chart.
Deployed HashiCorp's counting service in Linux node and dashboard service in Windows node. Both deployments with the annotation to inject Consul.

Follow this [guide](https://github.com/hashicorp-education/learn-consul-k8s-windows), along with WindowsSupport.md 

Checklist:
- [ x] Tests added
- [ ] CHANGELOG entry added 


control-plane/connect-inject/webhook/container_init.go
```diff
@@ -255,6 +272,14 @@ func (w *MeshWebhook) containerInit(namespace corev1.Namespace, pod corev1.Pod,
...
 			}
+		} else if !w.EnableCNI && isWindows(pod) {
+			// To disable ip tables on Windows, we just need to pass an empty string as a value for this envVar.
+			redirectTrafficConfig := ""
+			container.Env = append(container.Env,
+				corev1.EnvVar{
+					Name:  "CONSUL_REDIRECT_TRAFFIC_CONFIG",
+					Value: redirectTrafficConfig,
+				})
 		}
```
ishustava, line 283: Similar to my above comment, I think we should fail here instead and give an error to the user if iptables is not supported on Windows. It should be sufficient to fail just when tproxy is enabled because CNI will only run if tproxy is true.

control-plane/connect-inject/webhook/mesh_webhook.go
```diff
@@ -209,6 +215,12 @@ type multiPortInfo struct {
 	serviceName  string
 }
 
+func isWindows(pod corev1.Pod) bool {
+	podOS := pod.Spec.NodeSelector["kubernetes.io/os"]
```
ishustava, line 219: Because it's an optional value, we should check if it's set:
```suggestion
       podOS, ok := pod.Spec.NodeSelector["kubernetes.io/os"]
	if !ok {
		return false
	}
```

control-plane/connect-inject/webhook/mesh_webhook.go
```diff
@@ -209,6 +215,12 @@ type multiPortInfo struct {
...
+
+	return podOS == "windows"
+}
```
ishustava, line 222: Could we add tests for this function?

control-plane/subcommand/inject-connect/command.go
```diff
@@ -146,10 +149,16 @@ func (c *Command) init() {
 		"Directory with PEM-encoded TLS certificate and key to serve.")
 	c.flagSet.StringVar(&c.flagConsulImage, "consul-image", "",
 		"Docker image for Consul.")
+	c.flagSet.StringVar(&c.flagConsulImageWindows, "consul-image-windows", "",
+		"Windows Docker image for Consul.")
```
ishustava, line 153: Do we need this flag? We don't use the consul image for injection I think.

control-plane/subcommand/inject-connect/command.go
```diff
@@ -146,10 +149,16 @@ func (c *Command) init() {
...
 	c.flagSet.StringVar(&c.flagConsulDataplaneImage, "consul-dataplane-image", "",
 		"Docker image for Consul Dataplane.")
+	c.flagSet.StringVar(&c.flagConsulDataplaneImageWindows, "consul-dataplane-image-windows", "",
+		"Windows Docker image for Consul Dataplane.")
 	c.flagSet.StringVar(&c.flagConsulK8sImage, "consul-k8s-image", "",
 		"Docker image for consul-k8s. Used for the connect sidecar.")
+	c.flagSet.StringVar(&c.flagConsulK8sImageWindows, "consul-k8s-image-windows", "",
+		"Windows Docker image for consul-k8s. Used for the connect sidecar.")
```
ishustava, line 161:
```suggestion
		"Windows Docker image for consul-k8s-control-plane.")
```

control-plane/subcommand/inject-connect/command.go
```diff
@@ -738,12 +750,21 @@ func (c *Command) validateFlags() error {
 	if c.flagConsulK8sImage == "" {
 		return errors.New("-consul-k8s-image must be set")
 	}
+	if c.flagConsulK8sImageWindows == "" {
+		return errors.New("-consul-k8s-image-windows must be set")
+	}
```
ishustava, line 755: I don't think we should require these images to be set. For linux only deployments, I'd expect these to be empty.

ishustava (commented):
Great work so far, Jose!! Apologies for the late review. Thank you so much for writing it in the docs and adding tests. I think we will probably need to find a better place for these docs. We can leave them here for now, but I'd probably want to rewrite them and put them on consul.io.

I had some comments around tproxy and DNS names for consul server that I left inline.

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -159,21 +171,37 @@ func (w *MeshWebhook) consulDataplaneSidecar(namespace corev1.Namespace, pod cor
 				return corev1.Container{}, fmt.Errorf("container %q has runAsUser set to the same UID \"%d\" as consul-dataplane which is not allowed", c.Name, sidecarUserAndGroupID)
 			}
 		}
-		container.SecurityContext = &corev1.SecurityContext{
-			RunAsUser:              pointer.Int64(sidecarUserAndGroupID),
-			RunAsGroup:             pointer.Int64(sidecarUserAndGroupID),
-			RunAsNonRoot:           pointer.Bool(true),
-			ReadOnlyRootFilesystem: pointer.Bool(true),
+		// Security Context should not be set when using Windows.
```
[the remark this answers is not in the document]
joselo85: Windows pods does not support any of the security context settings. For example, trying to set `securityContext.privileged: true` will crash the pod when initializing. You can read more [here](https://kubernetes.io/docs/concepts/windows/intro/#compatibility-v1-pod-spec-containers-securitycontext)

control-plane/subcommand/inject-connect/command.go
joselo85: We don't need this flag. I added it just for consistency. If you agree, I can remove it.

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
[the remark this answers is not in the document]
ishustava: Ah thank you.
```suggestion
		// Security Context is not supported on Windows. See https://kubernetes.io/docs/concepts/windows/intro/#compatibility-v1-pod-spec-containers-securitycontext.
```

control-plane/subcommand/inject-connect/command.go
ishustava: Yep, let's remove it!

komapa:
Is this PR planned on being finished? Thank you!

PR #1849, opened 2023-01-23
Set envoy log level per pod

jm96441n:
Changes proposed in this PR:
- Allow users to set the log level on the envoy proxy in a specified pod, they can either set the level for all loggers by calling `consul-k8s proxy log POD_NAME -l warning`, or on specific loggers by `consul-k8s proxy log POD_NAME -l grpc:warning` or `consul-k8s proxy log POD_NAME -l grpc:warning,http:info`
- Small refactor to pull out an `envoy` package to encapsulate interactions with envoy

How I've tested this PR:
* Unit tests
* ran consul on k8's locally using kind https://gist.github.com/jm96441n/a56cd2ebd53f4484c3b40e725e305372
     * create a cluster using kind: `kind create cluster`
     * install consul on that cluster using the `consul-values.yml` file in the above gist: `consul-k8s install -config-file=consul-values.yaml -set global.image=hashicorp/consul:1.14.3`
     * once that is up apply the `server.yaml` to your cluster `kubectl apply server.yaml`
     * once that is setup get one of the server pod names using `kubectl get pods`
     * in another terminal window set up port forwarding for that pod `kubectl port-forward <POD_NAME> 19000:19000` (where `19000` is the default admin port for the envoy proxy)
     * in the original terminal run `consul-k8s proxy log <POD_NAME> -l warning` and compare the output to the envoy admin interface which would be at `localhost:19000/logging` while port forwarding is enabled (rinse and repeat for other combinations of changing levels)

code review
potentially running above steps (but probably not necessary)

**NOTE:** Didn't add a CHANGELOG yet, planning on adding the CHANGELOG entry once the work on the consul-k8s proxy log command is complete, though can add one for the current work if that's preferred (the next PR should be the last one to add reset functionality to put logs back to the default level for envoy)

Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added 


cli/cmd/proxy/envoy/http.go
```diff
@@ -12,11 +14,13 @@ import (
...
 )
 
+var ErrNoLoggersReturned = errors.New("No loggers were returned from Envoy")
+
 // EnvoyConfig represents the configuration retrieved from a config dump at the
 // admin endpoint. It wraps the Envoy ConfigDump struct to give us convenient
 // access to the different sections of the config.
 type EnvoyConfig struct {
-	rawCfg    []byte
+	RawCfg    []byte
```
jm96441n, line 23: so this change and changing the tests to utilize the `testdata` directory rather than embedding are the only substantial changes here other than the move

cli/cmd/proxy/envoy/http.go
```diff
@@ -1,8 +1,10 @@
-package read
+package envoy
```
t-eckert, line 1: I think this makes more sense being placed in `cli/common/`. The pattern for the `cmd` directory so far has been to have each directory correspond to a command directly.

For instance `cmd/proxy/read/` corresponds to the CLI command `consul-k8s proxy read`. Having `cmd/proxy/envoy/` implies to me that `consul-k8s proxy envoy` is a command and its code is contained here.

I like that you are taking the opportunity to refactor these things out while you are adding new functionality, but I think it makes sense to store that in `common/` where "shared functionality" has been placed.

jm96441n: that makes sense, I kept it here initially because it's really only shared between the proxy commands but let me move this into the `common` package

cli/cmd/proxy/envoy/http.go
```diff
@@ -69,6 +73,72 @@ type Secret struct {
 	LastUpdated string
 }
 
+// CallLoggingEndpoint requests the logging endpoint from Envoy Admin Interface for a given port
+// more can be read about that endpoint https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--logging
+func CallLoggingEndpoint(ctx context.Context, portForward common.PortForwarder, params string) (map[string]string, error) {
+	params = parseParams(params)
```
t-eckert, line 79: For separation of concerns, I think it would make sense to have a struct passed in for params instead of a string. I'm imagining something like this:

```go
type LoggingConfig struct {
    ...
}

func ParseLoggingParams(params string) LoggingConfig {
    ...
    return LoggingConfig {
        ...
    }
}

func UpdateLogLevel(ctx context.Context, portForward common.PortForwarder, cfg LoggingConfig) (map[string]string, error) {
    ...
}
````

File: cli/cmd/proxy/envoy/http.go (hunk 2 above)
Comment (t-eckert) [reply cli/cmd/proxy/envoy/http.go:1]: Now that I'm looking at this more, because it's not common across multiple commands, I think it works better as a separate file in the `cli/cmd/logLevel/` directory.

Comment (jm96441n) [reply cli/cmd/proxy/envoy/http.go:1]: so we use this here and in the `read` command

File: cli/cmd/proxy/envoy/http.go (hunk 3 above)
Comment (jm96441n) [reply cli/cmd/proxy/envoy/http.go:79]: we could do that, but we'd have to have the `log` package call this function from the `envoy` package to parse the params into the struct, and then pass that struct into the CallLoggingEndpoint (I'm not 100% on this name tbh) function to transform it back into a string for the http call

I was kinda on the fence about where to put this parsing command, on the one hand it's specific to envoy, but thinking more about it having the parse know about what format the command takes seems like it knows a bit too much about it's consumers

File: cli/cmd/proxy/envoy/http.go (hunk 2 above)
Comment (jm96441n) [reply cli/cmd/proxy/envoy/http.go]: so I ended up moving this into the `common` package since it's used in both the `read` and `log` commands

File: cli/cmd/proxy/envoy/http.go (hunk 3 above)
Comment (jm96441n) [reply cli/cmd/proxy/envoy/http.go]: so for this I left it as a string and moved parsing up to the `loglevel` package, I'm gonna noodle a bit on making it a struct, the params for this call are a bit wonky so I wanna think of something that makes sense, for reference it's:
`?level={level}` to set the level on all loggers
`?{someLogger}:{level}` to set the level on one logger
`?paths={someLogger}:{level},{anotherLogger}:{maybeDifferentLevel}` to set the level on multiple loggers

I feel like _maybe_ the following could work:

```
type LoggerConfig struct {
      GlobalLevel string
      IndividualLevels map[string]string
}
```
and have the param parsing in the envoy package check if `GlobalLevel` is set and if so then just use that and ignore the `IndividualLevels` map, what do you think?

File: cli/cmd/proxy/loglevel/command.go [hunk 4]
```diff
@@ -222,7 +227,7 @@ func (l *LogLevelCommand) fetchAdminPorts() (map[string]int, error) {
 	return adminPorts, nil
 }
 
-func (l *LogLevelCommand) fetchLogLevels(adminPorts map[string]int) (map[string]LoggerConfig, error) {
+func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
```
jm96441n, line 230: so I also consolidated the calls to envoy into this function, because they were both the exact same code, I'd prefer to keep them separate so it reads better, but having the envoy API is kinda forcing my hand to keep this code from being completely duplicated

cli/common/envoy/http.go
```diff
@@ -69,6 +73,53 @@ type Secret struct {
 	LastUpdated string
 }
 
+// CallLoggingEndpoint requests the logging endpoint from Envoy Admin Interface for a given port
+// more can be read about that endpoint https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--logging
+func CallLoggingEndpoint(ctx context.Context, portForward common.PortForwarder, params string) (map[string]string, error) {
```
jm96441n, line 78: part of me thinks we should probably be validating the params when they come in, as of right now this is only called from the `cmd/proxy` package, but it could be called from elsewhere in the future and it would be nice to give feedback about bad formatting earlier in the process before we hit envoy. OTOH this could potentially be mitigated by having a struct that holds the params and parsing that into a string here

cli/cmd/proxy/envoy/http.go
jm96441n: ended up implementing this in `common/envoy/logger_params.go`

cli/common/envoy/http.go
jm96441n: ended up handling the validation in setting the logger params

cli/common/envoy/http.go
```diff
@@ -69,6 +73,53 @@ type Secret struct {
 	LastUpdated string
 }
 
+// CallLoggingEndpoint requests the logging endpoint from Envoy Admin Interface for a given port
```
andrewstucki, line 76: Can you add some description of the fact that we both retrieve and update loggers with this call?

andrewstucki (commented):
So, cursory pass and the code looks generally fine to me, but I guess I have a question/observation about the usability and/or flag naming conventions. It seems a bit confusing to me that what appears to be the "listing" log-level command also acts as double-duty for setting said log-levels.

If we're fine going with that approach, what do you think about making the flag name a bit more specific? Saying `-level` or `-l` means we don't really have any user-indication that they're actually doing some sort of "set" command. What about something like `-update-level` or `-u`?

cli/common/envoy/http.go
jm96441n: can do

jm96441n:
I like the idea of using `-u`/`-update-level` as opposed to just level, let me make that change

andrewstucki (approved):
LGTM

PR #1852, opened 2023-01-26
Bump Kubernetes versions to existing +2 for clouds

thisisnotashwin:
Changes proposed in this PR:
- Bump the version of K8s on the pipelines

How I've tested this PR:
- https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/10356/workflows/21b91c9b-2fd4-4a6e-aeef-bc267bb62b3b for AKS and GKE
- https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/10362/workflows/0c656b2b-44aa-4d13-b346-77181663aec3 for EKS
- https://app.circleci.com/pipelines/github/hashicorp/consul-k8s/10367/workflows/905bd862-0255-488b-bca0-f5d1688f884e latest kind

look at them pipelines and how pretty and green they are and then hit approve 😛

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -573,7 +573,7 @@ jobs:
       - checkout
       - install-prereqs
       - create-kind-clusters:
-          version: "v1.25.3"
+          version: "v1.26.0"
       - restore_cache:
           keys:
             - consul-helm-modcache-v2-{{ checksum "acceptance/go.mod" }}
@@ -606,7 +606,7 @@ jobs:
       - checkout
       - install-prereqs
       - create-kind-clusters:
-          version: "v1.25.3"
+          version: "v1.26.0"
       - restore_cache:
           keys:
             - consul-helm-modcache-v2-{{ checksum "acceptance/go.mod" }}
@@ -1261,12 +1261,18 @@ workflows:
           context: consul-ci
           requires:
             - build-distros-linux
-      - acceptance-eks-1-23:
+      - acceptance:
+          context: consul-ci
           requires:
           - dev-upload-docker
-      - acceptance-eks-cni-1-23:
+      - acceptance-tproxy-cni:
+          context: consul-ci
           requires:
-          - acceptance-eks-1-23
+          - dev-upload-docker
+      - acceptance-tproxy:
+          context: consul-ci
+          requires:
+          - dev-upload-docker
 
 
   nightly-cleanup:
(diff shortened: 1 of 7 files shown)


david-yu:
I assume kind will be also be bumped to 1.26? Is that in a separate PR perhaps?

david-yu:
Ahh I just saw this https://twitter.com/BenTheElder/status/1601280095184117760, it looks like its possible to use 1.26.x but not sure its been used before this way for us.

thisisnotashwin:
Updated kind as well to 1.26.

I have disabled pod security policies for now. we need some further investigations into what it's future looks like given it isn't currently supported. that can be done in the future as another independent PR.

.circleci/config.yml
```diff
@@ -773,7 +773,7 @@ jobs:
       - run: mkdir -p $TEST_RESULTS
 
       - run-acceptance-tests:
-          additional-flags: -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig" -enable-pod-security-policies -enable-transparent-proxy -consul-image=hashicorppreview/consul-enterprise:1.15-dev-23aaa4f83845d0e2eced9ea69f731d7eedf840d1
```
curtbushko, line 776: Does -pod-security-policies break things on the new versions?

curtbushko (approved):
Look at that fancy terraform! 😍

.circleci/config.yml
thisisnotashwin: Kubernetes no longer supports pod security policies starting 1.25. https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

thisisnotashwin:
Should we backport this to 0.49.x? (it doesn't matter as much because those only run against GKE)

DanStough (approved):
Thank you for fixing 🚀

charts/consul/test/terraform/eks/main.tf
```diff
@@ -80,6 +81,47 @@ module "eks" {
...
+resource "aws_iam_role_policy_attachment" "csi" {
+  count      = var.cluster_count
+  role       = aws_iam_role.csi-driver-role[count.index].name
+  policy_arn = data.aws_iam_policy.csi-driver-policy.arn
+}
+
+resource "aws_eks_addon" "csi-driver" {
+  count                    = var.cluster_count
+  cluster_name             = module.eks[count.index].cluster_id
+  addon_name               = "aws-ebs-csi-driver"
```
ishustava, line 119: Does the addon name need to be unique per cluster?

charts/consul/test/terraform/gke/main.tf
```diff
@@ -29,7 +29,7 @@ resource "google_container_cluster" "cluster" {
     machine_type = "e2-standard-4"
   }
   pod_security_policy_config {
-    enabled = true
+    enabled = false
   }
```
ishustava, line 33: nit: we can just remove this config altogether since it's not enabled by default. Another thing here is that we can switch the google provider regular instead of beta.

ishustava (approved):
Great work, Ashwin! Thank you for fixing it. I left a few comments but they are not blocking.

ishustava:
@thisisnotashwin we'll also need to update readme and add it to Changelog.

charts/consul/test/terraform/eks/main.tf
thisisnotashwin: that is a good point. i can add a static name field.

charts/consul/test/terraform/gke/main.tf
thisisnotashwin: can do!

charts/consul/test/terraform/eks/main.tf
ishustava: Ah ok! no need then if it doesn't matter.

Issue #1867, opened 2023-02-01
Persistent volume fails to provision when using using OpenEBS Local PV Hostpath


billgrant:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

When using the `openebs-hostpath` storageclass in Kubernetes. Consul server pods get stuck in a pending state while attempting to create a consul cluster using the Consul Helm chart. The consul-connect-injector pod with get stuck in a restart loop. 

Here is are troubleshoot steps

```bash
kubectl get pods -n consul

NAME                                           READY   STATUS    RESTARTS      AGE
consul-connect-injector-749df9cb86-nk7lj       0/1     Running   2 (21s ago)   2m28s
consul-server-0                                0/1     Pending   0             2m28s
consul-server-1                                0/1     Pending   0             2m28s
consul-server-2                                0/1     Pending   0             2m27s
consul-server-acl-init-cleanup-jntc8           1/1     Running   0             2m28s
consul-server-acl-init-d62rc                   1/1     Running   0             2m28s
consul-webhook-cert-manager-7687cd6dd7-lflvh   1/1     Running   0             2m28s
```

```bash
kubectl logs consul-connect-injector-749df9cb86-nk7lj -n consul

2023-02-01T18:30:11.064Z [INFO]  consul-server-connection-manager: trying to connect to a Consul server
2023-02-01T18:30:11.079Z [ERROR] consul-server-connection-manager: connection error: error="failed to discover Consul server addresses: failed to resolve DNS name: consul-server.consul.svc: lookup consul-server.consul.svc on 10.96.0.10:53: no such host"
```

```bash
kubectl get pvc -n consul

NAME                          STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS       AGE
data-consul-consul-server-0   Pending                                      openebs-hostpath   4m18s
data-consul-consul-server-1   Pending                                      openebs-hostpath   4m18s
[...]

PR #1869, opened 2023-02-01
Exclude openebs namespace from injection.

lkysow:
OpenEBS is a Kubernetes storage solution. When you spin up a PVC, under the hood OpenEBS creates a pod to handle the necessary storage operations. If the openebs namespace is not excluded from injection, that pod can't start because our mutatingwebhook config requires all pod scheduling requests make it to our webhook and our webhook isn't running yet because the consul servers aren't running.

This is a breaking change but I think it's worth it because it's very unlikely anyone is using the openebs namespace for anything other than openebs.

Fixes https://github.com/hashicorp/consul-k8s/issues/1867

Checklist:
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 ## UNRELEASED
 
+BREAKING CHANGES:
+* Helm:
+  * Change defaults to exclude the `openebs` namespace from sidecar injection. If you previously had pods in that namespace
+    that you wanted to be injected, you must now set `namespaceSelector` as follows:
+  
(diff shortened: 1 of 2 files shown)


lkysow:
Only downside with this is that do we keep adding namespaces here? But in this case I think it's in line with why we added the kind namespace.

david-yu:
Assigning reviewers.

hashi-derek (approved):

malizz (approved):
LGTM

PR #1874, opened 2023-02-02
Exclude openebs namespace from injection. (#1869)

thisisnotashwin:
* Exclude openebs namespace from injection.

OpenEBS is a Kubernetes storage solution. When you spin up a PVC, under the hood OpenEBS creates a pod to handle the necessary storage operations. If the openebs namespace is not excluded from injection, that pod can't start because our mutatingwebhook config requires all pod scheduling requests make it to our webhook and our webhook isn't running yet because the consul servers aren't running.

This is a breaking change but I think it's worth it because it's very unlikely anyone is using the openebs namespace for anything other than openebs.

* Changelog

Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 ## UNRELEASED
 
+BREAKING CHANGES:
+* Helm:
+  * Change defaults to exclude the `openebs` namespace from sidecar injection. If you previously had pods in that namespace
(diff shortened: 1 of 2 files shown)


thisisnotashwin:
Closes #1872

PR #1877, opened 2023-02-03
Refine server TLS Vault PKI role config in acceptance tests

jkirschner-hashicorp:
The generate_lease=true configuration is unnecessary and generates a note about performance implications in Vault logs. Remove this configuration from the acceptance tests so that the default value of generate_lease=false is used instead.

A follow on PR will update the public Consul docs to reflect this change in recommendation.

How I've tested this PR:
- I'm relying on the existing k8s acceptance tests to ensure the integration still works

Checklist:
- [ ] Tests added: No - relying on existing tests
- [ ] CHANGELOG entry added: Seems unnecessary since this only affects the acceptance tests, not consul k8s itself

### DIFF
--- a/acceptance/framework/vault/helpers.go
+++ b/acceptance/framework/vault/helpers.go
@@ -40,7 +40,6 @@ func ConfigurePKICerts(t *testing.T,
 		"allow_bare_domains": "true",
 		"allow_localhost":    "true",
 		"allow_subdomains":   "true",
-		"generate_lease":     "true",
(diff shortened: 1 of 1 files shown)


ishustava (approved):
Thanks for fixing it! Since the tests all passed, we should be good to merge. I don't this value has any impact on the integration itself.

david-yu:
Adding a backport to 1.0.x and 0.49.x and merging.

PR #1881, opened 2023-02-06
NET-1750 Fixes ACL init command Consul Login Datacenter mixup

missylbytes:
Changes proposed in this PR:
- [JIRA NET-1750](https://hashicorp.atlassian.net/browse/NET-1750)

> In the Consul on Kubernetes Control Plane, the ACL init command uses flag/env var pairing -datacenter/CONSUL_DATACENTER to perform a login. This can cause an issue for resources running in a federated context that need to talk to a primary datacenter to perform a login. This should be changed so that the login uses the values configured by -consul-login-datacenter/CONSUL_LOGIN_DATACENTER.
> 
> To fix this, the value passed for Datacenter [here](https://github.com/hashicorp/consul-k8s/pull/1881). Should be changed to c.consul.ConsulLogin.Datacenter.
> 
> Any resources where the -datacenter flag was used to configure the login datacenter should be evaluated to see if they should be changed to use the CONSUL_LOGIN_DATACENTER environment variable. This is the case when a resource is in the non-primary datacenter and requires contacting the primary datacenter to perform a login.
> 
> There is at-least one known instance where this change needs to be made in the api-gateway-controller-deployment.yaml template [here](https://github.com/hashicorp/consul-k8s/blob/5e901000ec321be1d609e9678ca6d86572cae482/charts/consul/templates/api-gateway-controller-deployment.yaml#L257).

How I've tested this PR:

- Followed [this tutorial](https://developer.hashicorp.com/consul/docs/k8s/deployment-configurations/multi-cluster/kubernetes) to create two clusters. However, the consul deployment for DC2 was done from this branch's helm chart.
- Detailed steps can be found [in this gist](https://gist.github.com/missylbytes/10ee3802b3c492497df1eb2cc748e96b).

- If you would like to spin up your own GCP temporary account, you should be able to follow the steps in the gist above to test.

Checklist:
- [x] Tests added


control-plane/subcommand/acl-init/command.go
```diff
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/hashicorp/go-netaddrs"
```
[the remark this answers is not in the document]
missylbytes: I'll fix this.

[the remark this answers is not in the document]
missylbytes: fixed

ndhanushkodi (approved):

ndhanushkodi:
Will the changelog come in a separate PR?

Thanks for making this fix!

missylbytes:
> Will the changelog come in a separate PR?
> 
> Thanks for making this fix!

@ ndhanushkodi I didn't know if sure if there should be a changelog, as I don't think it will affect the user. 
It seems like the two places that the values were mixed up (and were fixed) were cancelling each other out. 
i.e., the acl-init command *should* be using the consul login datacenter, and the two places where federation.enabled were edited to ensure the login datacenter would be set to the correct datacenter.

david-yu:
Ahh ok will remove 0.49.x then. Thanks @ishustava

charts/consul/templates/client-daemonset.yaml
```diff
@@ -509,6 +509,12 @@ spec:
         - name: CONSUL_LOGIN_META
           value: "component=client,pod=$(NAMESPACE)/$(POD_NAME)"
         {{- end }}
+        - name: CONSUL_LOGIN_DATACENTER
```
curtbushko, line 512: I think you might need bats tests in client-daemonset.bats to check the output of the toggle.

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -240,6 +241,12 @@ spec:
               fieldPath: metadata.name
         - name: CONSUL_LOGIN_META  
           value: "component=api-gateway-controller,pod=$(NAMESPACE)/$(POD_NAME)"
+        - name: CONSUL_LOGIN_DATACENTER
```
curtbushko, line 244: I think you might need bats tests in api-gateway-controller-deployment.bats to test the toggle.

curtbushko (changes requested):
Looks good. Just a few changes that I *think* are needed.

charts/consul/templates/client-daemonset.yaml
missylbytes: Yeah I thought this was covered by https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/test/unit/api-gateway-controller-deployment.bats#L497
However, I think I misunderstood that test, it is for the command and I also need to add one for the env vars. Does that seem correct?

missylbytes: Added

charts/consul/templates/api-gateway-controller-deployment.yaml
missylbytes: Added

curtbushko (approved):

PR #1882, opened 2023-02-06
Dockerfile: Remove gnupg from Consul k8s container

david-yu:
Changes proposed in this PR:
- `gnupg` is likely an artifact from pre-CRT days when we used to validate Consul K8s releases ourselves for publishing with Docker. This is is likely not needed so will attempt removing it. 

How I've tested this PR:
- Acceptance tests

Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 


curtbushko (approved):
LGTM!

PR #1770, opened 2022-12-05
Update server-acl-init to always check for the deployed serviceAccountToken secret

kschoche:
Changes proposed in this PR:
- OpenShift injects `SecretRefs` into `ServiceAccounts` (one for service account and the other for docker registry credentials), even in Kubernetes 1.24+. In the current logic of `server-acl-init` we expect to use the Secret deployed by the helm chart, but only in the case where the ServiceAccount does not contain SecretRefs. Due to OpenShift injecting these we never look for the deployed Secret for the consul-auth-method. 
- Since the helm chart and consul-k8s versions are sync'd and we do not have backward compatibility issues we should be able to rely on the consul auth method Secret always existing and we should use that.
- I believe this should resolve #1768.

How I've tested this PR:
Unit + acceptance tests should pass.

How I expect reviewers to test this PR:
👀 
Looking for some review comments about the approach as well :) 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #1907, opened 2023-02-14
server-acl-init to always check for the deployed serviceAccountToken secret (#1770)

thisisnotashwin:
backport of #1770


kschoche (approved):
👯

PR #1914, opened 2023-02-15
[0.49.x] remove livenessProbe from pods with preStop lifecycle hooks that delete ACL tokens

rboyer:
Currently we use an init container to create a consul ACL token via an auth method. We configure a preStop lifecycle hook to delete this token to avoid leaking it but this executes when containers get restarted, not just when the pod is destroyed.

This means that if a container livenessProbe fails and triggers a container to be restarted in place the token it was using will be destroyed but not recreated.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
   golangci-lint-helm-gen:
    needs:
     - get-go-version
-   uses: ./.github/workflows/golangci-lint
+   uses: ./.github/workflows/reusable-golangci-lint.yml
    with:
       directory: hack/helm-reference-gen
       go-version: ${{ needs.get-go-version.outputs.go-version }}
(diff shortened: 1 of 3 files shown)


thisisnotashwin (commented):
LGTM! Thanks for the `@main` removal!

thisisnotashwin (approved):

david-yu:
Out of curiosity, is this problem only apply in 0.49.x or is there a separate solution for 1.0.x?

rboyer:
> Out of curiosity, is this problem only apply in 0.49.x or is there a separate solution for 1.0.x?
The `preStop` hooks in question are no longer present for `>=1.x.x`
```
(1.0.x) $ git grep preStop charts | grep -v unit
charts/consul/templates/api-gateway-controller-deployment.yaml:          preStop:
charts/consul/templates/client-daemonset.yaml:            preStop:
$
```

PR #1921, opened 2023-02-17
Add automountServiceAccountToken to server serviceAccount

mehta-ankit:
Changes proposed in this PR:
- Allow one to set `automountServiceAccountToken` on the serviceAccount. By default, the server pod mounts the API credentials token
```
kube-api-access-xxxxx:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           xxxxx
    ConfigMapOptional:       xxxxx
    DownwardAPI:             true
```
I am just making it explicit on the serviceAccount resource in this helm chart. We have a requirement to set this key instead of getting the default behavior without knowing about it.
I could have created my own `serviceAccount` but this chart does not allow me to stop the creation of the serviceAccount and even if I create my own I cannot provide it as a value to the server Statefulet (https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/server-statefulset.yaml#L138). At some point i would want to make those changes too, so the chart is a little more customizable.

How I've tested this PR:
- templated the chart and it has the value set correctly: 
image

- 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


mehta-ankit:
Can this be reviewed and merged please 🙏🏼

mehta-ankit:
Any update on this ? Is there a reason this cannot be merged ? 🤔

PR #1948, opened 2023-02-24
Prepare 1.1.0 release

curtbushko:
Changes proposed in this PR:
- Update the versions of troubleshoot, envoyextentions and api to match what is tagged in Consul 1.15.
- There was a bug with consul-k8s troubleshoot because we were pointing to an older SHA for the troubleshoot package. It is minor and only affected the output.
- Made sure all the changes from 1.0.0 + patches are in the Changelog
- Update consul image to be 1.15
- Update envoy to be the latest 1.25.1

How I've tested this PR:

👀 

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


thisisnotashwin (approved):

charts/consul/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: consul
-version: 1.1.0-dev
-appVersion: 1.14.4
+version: 1.1.0
+appVersion: 1.15.0
 kubeVersion: ">=1.21.0-0"
```
david-yu, line 5:
```suggestion
kubeVersion: ">=1.22.0-0"
```

We should bump this since most K8s distros (i.e. EKS) now support 1.22.x and above.

curtbushko: I you think so, I will bump it.

david-yu: Yes please.

david-yu (approved):

Issue #1911, opened 2023-02-14
server-acl-init generates incomplete API Gateway Controller policy


krarey:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

When `global.adminPartitions.enabled = true` is defined in Helm values, the policy generated for the API Gateway Controller is missing `operator = "write"` permissions, rendering the gateway controller unable to create new namespaces.

The following conditional misses this permission in the event admin partitions have been enabled: https://github.com/hashicorp/consul-k8s/blob/767f242a291a550cd4d00801f3fe1f0ac0a28fb7/control-plane/subcommand/server-acl-init/rules.go#L146-L153

### Reproduction Steps

* Deploy Consul Enterprise on Kubernetes with the following features enabled:
```yaml
global:
  image: "hashicorp/consul-enterprise:1.14.3-ent"
  enableConsulNamespaces: true
  adminPartitions:
    enabled: true
    name: "default"
  acls:
    manageSystemACLs: true
connectInject:
  enabled: true
  consulNamespaces:
    mirroringK8S: true
apiGateway:
  enabled: true
  image: hashicorp/consul-api-gateway:0.5.1
  imageEnvoy: envoyproxy/envoy:v1.23.1
```

* Create a new API gateway in a Kubernetes namespace that does not yet exist in Consul
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: apigw
  namespace: apigateway
spec:
  gatewayClassName: consul-api-gateway
  listeners:
  - protocol: HTTP
    port: 8080
    name: http
    allowedRoutes:
      namespaces:
        from: All
```

* Observe the API Gateway Controller log output, it will enter a fail/retry loop encountering ACL denial errors attempting to create the new namespace.

### Logs

```
2023-02-14T21:33:43.453Z [WARN] gatewayclient/middleware.go:39: consul-api-gateway-server.k8s.Gateway: received non-Kubernetes error, delaying requeue:
[...]

PR #2029, opened 2023-03-24
Update ACLs, add namespace.write permission

sarahalsmiller:
Changes proposed in this PR:
- Resolve bug described in https://github.com/hashicorp/consul-k8s/issues/1911 where api-gateway controller is unable to create new namespaces

How I've tested this PR:
- Installed API gateway using a local version of this repository using the reproduction steps described in https://github.com/hashicorp/consul-k8s/issues/1911 and consul API image set to 1.5.2. (ACL permission is missing, but this change is also reliant on a bux fix with how partitions are handled in 1.5.2)

- CI passes

Checklist:
- [ X ] Tests added
- [ X ] CHANGELOG entry added 

### DIFF
--- a/control-plane/subcommand/server-acl-init/rules.go
+++ b/control-plane/subcommand/server-acl-init/rules.go
@@ -146,6 +146,7 @@ func (c *Command) apiGatewayControllerRules() (string, error) {
 	apiGatewayRulesTpl := `{{- if .EnablePartitions }}
 partition "{{ .PartitionName }}" {
   mesh = "write"
+  operator = "write"
   acl = "write"
 {{- else }}
 operator = "write"
--- a/control-plane/subcommand/server-acl-init/rules.go
+++ b/control-plane/subcommand/server-acl-init/rules.go
@@ -146,14 +146,16 @@ func (c *Command) apiGatewayControllerRules() (string, error) {
 	apiGatewayRulesTpl := `{{- if .EnablePartitions }}
 partition "{{ .PartitionName }}" {
   mesh = "write"
-  operator = "write"
   acl = "write"
+  operator = "write"
 {{- else }}
 operator = "write"
 acl = "write"
 {{- end }}
+
 {{- if .EnableNamespaces }}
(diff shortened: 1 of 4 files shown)


control-plane/subcommand/server-acl-init/rules.go
```diff
@@ -147,12 +147,15 @@ func (c *Command) apiGatewayControllerRules() (string, error) {
 partition "{{ .PartitionName }}" {
   mesh = "write"
   acl = "write"
+  operator = "write"
```
mikemorris, line 150: Is `operator = "write"` within the partition needed (this feels like more permission than we may want to grant our controller), or is 
```
namespace_prefix "" {
  policy = "write"
}
```
sufficient?

ndhanushkodi: +1 to Mike, it's worth investigating if we can avoid operator:write. And I saw this comment: https://github.com/hashicorp/consul-k8s/blob/main/control-plane/subcommand/server-acl-init/rules.go#L40-L41 higher up in the file as well, which indicates operator=write cannot be done within a partition anyways. Not sure if it'll cause issues when deployed (possibly not since you've tested this), but it sounds like if that's the case it should be possible to not include operator:write.

sarahalsmiller: After additional testing, I have removed the operator:write permission.

sarahalsmiller: Also worth nothing, I'm not sure what was wrong with my env on friday, but on this round of testing, the fix worked with both  api gateway 1.5.1 and 1.5.2

david-yu:
@sarahalsmiller Just an FYI, just linked #1911 to this PR.

control-plane/subcommand/server-acl-init/rules.go
david-yu: I assume you mean API Gateway 0.5.1 and 0.5.1?

PR #2030, opened 2023-03-24
add failover policy to service resolver and proxy default

malizz:
add failover policy to service resolver and proxy default
ticket: https://hashicorp.atlassian.net/browse/NET-2851

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


charts/consul/templates/crd-proxydefaults.yaml
```diff
@@ -143,6 +143,14 @@ spec:
                       type: object
                     type: array
                 type: object
+              failoverPolicy: 
+                description: FailoverPolicy specifies the exact mechanism used for failover.
+                properties:
+                  mode:
+                    description: Mode specifies the type of failover that will be performed. 
+                      Valid values are "default", "" (equivalent to "default") and "order-by-locality".
```
erichaberkorn, line 151: I'm currently changing the name of the "default" case to "sequential" in this PR https://github.com/hashicorp/consul-enterprise/pull/4734/files#diff-dadef19b7fe31571431416cb4db25634e4f83cad2b9547e543fa0277e8b4f221R1437

erichaberkorn (commented):
@wilkermichael will have more context on this because he is more familiar with the consul-k8s codebase.

charts/consul/templates/crd-proxydefaults.yaml
erichaberkorn: I just remembered that we are adding the regions field in that PR too

malizz: Thanks, updated the value, and will add the extra field when api package is updated.

control-plane/go.sum
wilkermichael, line 1: I had to run `go mod tidy` after pulling to get things working, so just try running that locally and committing the changes

wilkermichael (commented):
Great work Mali! Sorry I took so long to get to this

Some minor comments, also can you let me know if you tested things locally and update the PR under the `how I tested section` thanks!

charts/consul/templates/crd-proxydefaults.yaml
malizz: No worries. Thanks!

control-plane/connect-inject/controllers/endpoints/endpoints_controller_ent_test.go
```diff
@@ -2107,7 +2111,8 @@ func TestReconcileDeleteGatewayWithNamespaces(t *testing.T) {
 
 				if tt.enableACLs {
 					_, _, err = consulClient.ACL().TokenRead(token.AccessorID, nil)
-					require.EqualError(t, err, "Unexpected response code: 403 (ACL not found)")
+					require.EqualError(t, err,
```
wilkermichael, line 2114: I have a suggestion to potentially make these tests less brittle for the future. 

Can we switch all of the ACL error string matches to just check for `contains` like
`require.Contains(t, err, "ACL not found")` or something like that

wilkermichael (approved):
Great work Mali, thanks for fixing up all the tests and getting this running with 1.16-dev!

control-plane/connect-inject/controllers/endpoints/endpoints_controller_ent_test.go
malizz: Will do, the reason I didn't change them all to this assertion was to stay precise (if it actually matters to be able to distinguish between the two cases). Thanks!

PR #2036, opened 2023-04-04
Add GatewayClassConfig CRD

t-eckert:
Changes proposed in this PR:
- Updated dependencies that were needed to use API Gateway CRDs.
- Fixed some breaking changes from updating dependencies.
- Added our GatewayClassConfig with some values removed because they will now be set via Helm at installation time.
- Added tests for the functions associated with the CRDs.
- Generated the new CRDs and deep-copy code.
- Added a Make script for adding copyright headers.
- Leveraged that Make script in CRD generation.
- Removed old code in CRD deep-copy generation that is no longer needed.

How I've tested this PR:
- I have added tests for the CRD structs to ensure the functions that come along with them work correctly.

- Run the unit tests.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


Makefile
```diff
@@ -64,7 +64,7 @@ cni-plugin-lint:
 	cd control-plane/cni; golangci-lint run -c ../../.golangci.yml
 
 ctrl-generate: get-controller-gen ## Run CRD code generation.
-	cd control-plane; $(CONTROLLER_GEN) object:headerFile="build-support/controller/boilerplate.go.txt" paths="./..."
+	cd control-plane; $(CONTROLLER_GEN) object paths="./..."
```
t-eckert, line 67: This empty header file was originally needed to satisfy the Controller Gen CLI. We never had anything in it.

Makefile
```diff
@@ -139,6 +140,13 @@ else
 CONTROLLER_GEN=$(shell which controller-gen)
 endif
 
+add-copyright-header: ## Add copyright header to all files in the project
+ifeq (, $(shell which copywrite))
+	@echo "Installing copywrite"
+	@go install github.com/hashicorp/copywrite@latest
+endif
+	@copywrite headers --spdx "MPL-2.0"
```
t-eckert, line 148: This makes using `copywrite` the tool we use for adding copyright headers easier. It's a no-op for already copyrighted files.

One of the issues I was running into when generating the CRD YAML was that they lacked the necessary header. This script can now be executed after the generation to prevent that problem.

control-plane/api/common/configentry_webhook_test.go
```diff
@@ -115,7 +115,7 @@ func TestValidateConfigEntry(t *testing.T) {
 					},
 				},
 			},
-				logrtest.TestLogger{T: t},
+				logrtest.NewTestLogger(t),
```
t-eckert, line 118: An update to this library removed this direct access in favor of `NewTestLogger`. The behavior is unchanged.

control-plane/helper/controller/controller.go
```diff
@@ -62,7 +62,7 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 	// Add an event handler when data is received from the informer. The
 	// event handlers here will block the informer so we just offload them
 	// immediately into a workqueue.
-	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+	_, err := informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
t-eckert, line 65: This now returns an error and linting demands that we check it. I didn't want to change behavior so I just have it log the error.

david-yu:
Nice! This updates sig-controller runtime to 1.26.x. I assume this won't be back ported?

t-eckert:
> Nice! This updates sig-controller runtime to 1.26.x. I assume this won't be back ported?

Yep! This all goes in to Consul K8s 1.2. I'll verify that it works with 1.24 k8s and above.

david-yu:
> 

As far as Consul K8s 1.2, we probably also want to support K8s 1.27 which is slated to release in two weeks according to https://www.kubernetes.dev/resources/release/#summary

david-yu:
@t-eckert  I know this is not the scope of your work here, but how much work would it take to bump dependencies on Consul K8s 1.1.x for the following. I know we have some outstanding CVEs that folks have asked about fixing before. 


> Updated dependencies around Controller code.
> Fixed breaking changes from updating dependencies.

t-eckert:
> @t-eckert I know this is not the scope of your work here, but how much work would it take to bump dependencies on Consul K8s 1.1.x for the following. I know we have some outstanding CVEs that folks have asked about fixing before.
> 
> > Updated dependencies around Controller code.
> > Fixed breaking changes from updating dependencies.

I would be happy to!

Makefile
t-eckert: CC @lkysow because he had to add the empty file in the first place.

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Tolerations allow the scheduler to schedule nodes with matching taints.
+	// More Info: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// Annotation Information to copy to services or deployments
+	CopyAnnotations CopyAnnotationsSpec `json:"copyAnnotations,omitempty"`
+}
+
+// PortSpec defines the ports that the API Gateway will listen on.
```
andrewstucki, line 61: This seems to be unused, as it was formerly part of the Consul connection config which was stripped out (due to the fact that we should probably inline it to the deployment the controller creates to pass values from helm directly, similar to things like the connect injection hooks)

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -0,0 +1,260 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+
+	// Items is the list of Configs.
+	Items []GatewayClassConfig `json:"items"`
+}
+
+// RoleFor constructs a Kubernetes Role for the specified Gateway based
+// on the GatewayClassConfig.
+func (c *GatewayClassConfig) RoleFor(gw *gwv1beta1.Gateway) *rbac.Role {
```
andrewstucki, line 90: we'll probably need something like this eventually, but:

1. don't think we need it for now, and
2. it doesn't actually appear to use anything from the config anyway, so it likely doesn't belong here

So I think we can drop it

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -0,0 +1,260 @@
...
+			Resources: []string{"podsecuritypolicies"},
+			Verbs:     []string{"use"},
+		}},
+	}
+}
+
+// RoleBindingFor constructs a Kubernetes RoleBinding for the specified Gateway
+// based on the GatewayClassConfig. If the GatewayClassConfig is configured in
+// such a way that does not require a RoleBinding, nil is returned.
+func (c *GatewayClassConfig) RoleBindingFor(gw *gwv1beta1.Gateway) *rbac.RoleBinding {
```
andrewstucki, line 108: Same as above

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -0,0 +1,260 @@
...
+				Name:      serviceAccount.Name,
+				Namespace: serviceAccount.Namespace,
+			},
+		},
+	}
+}
+
+// ServiceAccountFor constructs a Kubernetes ServiceAccount for the specified
+// Gateway based on the GatewayClassConfig.
+func (c *GatewayClassConfig) ServiceAccountFor(gw *gwv1beta1.Gateway) *corev1.ServiceAccount {
```
andrewstucki, line 142: same as above

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -0,0 +1,260 @@
...
+	return &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      gw.Name,
+			Namespace: gw.Namespace,
+			Labels:    labelsForGateway(gw),
+		},
+	}
+}
+
+func MergeSecret(a, b *corev1.Secret) *corev1.Secret {
```
andrewstucki, line 152: actually pretty much for all of the helpers below -- I think we should add them as we find the need for them throughout the subsequent dev process. Think you can drop pretty much all of the helpers after the initial struct definitions.

control-plane/api/v1alpha1/api_gateway_types_test.go
```diff
@@ -0,0 +1,108 @@
...
+	require.Nil(t, nilConfigList.DeepCopyObject())
+	configList := &GatewayClassConfigList{
+		Items: []GatewayClassConfig{*config},
+	}
+	copyConfigList := configList.DeepCopy()
+	copyConfigListObject := configList.DeepCopyObject()
+	require.Equal(t, copyConfigList, copyConfigListObject)
+}
+
+func TestGatewayClassConfig_RoleFor(t *testing.T) {
```
andrewstucki, line 52: If you drop the helpers, can drop these last 3 tests too

andrewstucki (changes requested):
The scaffolding looks great to me in general, but I think we can drop some stuff that we aren't using yet and potentially add it back later if it makes sense to -- mainly just don't want to accumulate pieces of implementation that somehow wind up unused later.

.changelog/2036.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+api-gateway: Add GatewayClassConfig CRD to allow for configuration of GatewayClass.
```
sarahalsmiller, line 2: Obviously not part of this PR, but I imagine we'll also need to update our install docs to reflect this improvement.

charts/consul/templates/crd-gatewayclassconfigs.yaml
```diff
@@ -0,0 +1,120 @@
+{{- if .Values.connectInject.enabled }}
```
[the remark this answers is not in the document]
t-eckert: That's a good thought, but it's actually going to just be deployed if connectInject is enabled. That is, if a service mesh is being used.

.changelog/2036.txt
t-eckert: Definitely! We'll have a whole rewrite of install and upgrade docs.

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -0,0 +1,72 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	GatewayClassConfigKind = "GatewayClassConfig"
+	managedLabel           = "api-gateway.consul.hashicorp.com/managed"
+	nameLabel              = "api-gateway.consul.hashicorp.com/name"
+	namespaceLabel         = "api-gateway.consul.hashicorp.com/namespace"
+	createdAtLabel         = "api-gateway.consul.hashicorp.com/created"
```
andrewstucki, line 16: Can we also drop these, IIRC there are already equivalent labels within the [codebase](https://github.com/hashicorp/consul-k8s/blob/0e771fcd49be3af400caec8da0f46e52ccd29b88/control-plane/connect-inject/constants/annotations_and_labels.go) that we can leverage. So I don't think these are needed -- the `GatewayClassConfigKind` we can keep because we'll need it really soon for doing checks of parameterRef types for `GatewayClass` objects.

andrewstucki (approved):
One more quick set of changes I'd like to see before merging, but approving anyway so you can merge after the changes. Thanks!

PR #2037, opened 2023-04-04
Fix the indentation of the copyAnnotations example

t-eckert:
Changes proposed in this PR:
- Fixes the docs example where `copyAnnotations.service.annotation` has the wrong indentation. This caused some confusion for a customer recently. Just needs an extra two spaces.


nathancoleman (approved):
Confirmed this is what actually works when installing w/ API Gateway

lkysow (approved):

david-yu (approved):
Joining in on the party

PR #2038, opened 2023-04-04
Net 3484 add apigateway command

missylbytes:
- Adds dummy version of API Gateway Command

**How I've tested this PR:**
Tested locally with KIND

Will add instructions

```
kind create cluster --name=dc1
kubectl apply --kustomize "github.com/hashicorp/consul-api-gateway/config/crd?ref=v0.4.0"
```

Then create an override.yaml file with the following:

For using the global k8s image (this is my prebuilt image from this branch):
```
# override.yaml
global:
  imageK8S: melisaghashicorp/consul-k8s-control-plane-dev:net3484

apiGateway:
  enabled: true
```

For using the old APIGateway image:
```
# override.yaml
apiGateway:
  enabled: true
  image: hashicorp/consul-api-gateway
```
Then install consul with your override.yaml
```
helm install consul ./charts/consul -n consul <your-override>.yaml
```

If you are using the global consul-k8s image, when you describe the deployment it should return this:
```
  Containers:
   api-gateway-controller:
    Image:      melisaghashicorp/consul-k8s-control-plane-dev:net3484
    Port:       9090/TCP
    Host Port:  0/TCP
    Command:
      /bin/sh
      -ec
      consul-k8s-control-plane api-gateway \
        -sds-server-host consul-consul-api-gateway-controller.consul.svc \
        -k8s-namespace consul \
        -log-level info \
        -log-json=false
```
And the logs will say:
```
2023-04-04T20:57:06.448Z [INFO]  Welcome to API Gateway
```

If you are using the global consul-k8s image, when you describe the deployment it should return this:
```
  Containers:
   api-gateway-controller:
    Image:      hashicorp/consul-api-gateway
    Port:       9090/TCP
    Host Port:  0/TCP
    Command:
      /bin/sh
      -ec
      consul-api-gateway server \
        -sds-server-host consul-consul-api-gateway-controller.consul.svc \
        -k8s-namespace consul \
        -log-level info \
        -log-json=false
```
And the logs will start running all of the normal api-gateway logs.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added


control-plane/subcommand/api-gateway/command_test.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package apigateway
```
missylbytes, line 4: I feel like this test is probably completely unnecessary and will change as soon as the real command gets implemented. May remove it.

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -56,7 +59,7 @@ spec:
       serviceAccountName: {{ template "consul.fullname" . }}-api-gateway-controller
       containers:
       - name: api-gateway-controller
-        image: {{ .Values.apiGateway.image }}
+        image: {{ default .Values.global.imageK8S .Values.apiGateway.image}}
```
missylbytes, line 62: This *should* default the image to be the globalK8S image, and only insert the API Gateway image if it is defined.

charts/consul/templates/api-gateway-controller-deployment.yaml
```diff
@@ -1,6 +1,9 @@
 {{- if .Values.apiGateway.enabled }}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true for api gateway" }}{{ end }}
-{{- if not .Values.apiGateway.image}}{{ fail "apiGateway.image must be set to enable api gateway" }}{{ end }}
+{{- if not .Values.apiGateway.image}}
```
missylbytes, line 3: I didn't know a better way to offer a warning to users?

t-eckert:
We'll need to update this BATs test to match the change made to the deployment template. You can see that failure here:

```
not ok 8 apiGateway/Deployment: fails if no image is set
# (in test file test/unit/api-gateway-controller-deployment.bats, line [18](https://github.com/hashicorp/consul-k8s/actions/runs/4620343816/jobs/8170334969#step:4:19))
#   `[ "$status" -eq 1 ]' failed
```

The change needs to be made to the BATs test [here](https://github.com/hashicorp/consul-k8s/blob/d53f92e7044025a02fc669f96ae71681a68c8787/charts/consul/test/unit/api-gateway-controller-deployment.bats#L12). I would change this to test that the output uses the control-plane image if no `apiGateway.image` is set.

charts/consul/templates/api-gateway-controller-deployment.yaml
sarahalsmiller: Do we need to check the consul version here and fail if the version is too low for consul native api gateway?

missylbytes: Oh this might be a good idea, I do think we would have to wait to get everything merged in though first? Otherwise we wouldn't be able to use the chart until the new version went out.
Actually now I'm wondering how we would handle that anyway? It's kind of chicken and egg, right?

control-plane/subcommand/api-gateway/command.go
```diff
@@ -0,0 +1,136 @@
...
+	"flag"
+	"github.com/hashicorp/consul-k8s/control-plane/subcommand/common"
+	"github.com/hashicorp/consul-k8s/control-plane/subcommand/flags"
+	"github.com/mitchellh/cli"
+	"io"
+	"sync"
+)
+
+const (
+	defaultSDSServerHost = "consul-api-gateway-controller.default.svc.cluster.local"
```
missylbytes, line 17: Just a note, if you compare this file to the original some of the fields are missing. This was to please the linter gods, I imagine we can add them back when we implement the command.

control-plane/subcommand/api-gateway/command.go
```diff
@@ -0,0 +1,136 @@
...
+	return &Command{UI: ui, output: logOutput, ctx: ctx}
+}
+
+func (c *Command) init() {
+	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
+
+	c.flags.StringVar(&c.flagServerCAFile, "ca-file", "", "Path to CA for Consul server.")
+	c.flags.StringVar(&c.flagServerCASecret, "ca-secret", "", "CA Secret for Consul server.")
+	c.flags.StringVar(&c.flagServerCASecretNamespace, "ca-secret-namespace", "default", "CA Secret namespace for Consul server.")
+	c.flags.StringVar(&c.flagConsulAddress, "consul-address", "", "Consul Address.")
```
missylbytes, line 66: [Some other commands](https://github.com/hashicorp/consul-k8s/blob/main/control-plane/subcommand/create-federation-secret/command.go#L44) use special types for common flags, however these types have different actual names for the flags. i.e., consul-address just becomes `addresses`. I didn't want to start changing the flags that the user would pass in until we made sure the tradeoff on these changes were worth it, which I think lines up with the guidance from @t-eckert

control-plane/subcommand/api-gateway/command_test.go
missylbytes: I'm going to remove this, the bats tests are more useful at this point.

charts/consul/templates/api-gateway-controller-deployment.yaml
t-eckert: When it comes to checking versions, we've tried to validate the versions of images before and found it to be practically impossible. You can't truly rely on the image tag to verify the version. We have historically avoided it in this Helm chart because of the abundance of edge cases we'd have to cover.

t-eckert: So if something is deployed and the image is incorrect, we usually leave that error to appear at runtime when there is a verifiable way to tell what the image _is_.

control-plane/subcommand/api-gateway/command.go
t-eckert: I think while we have the ability to run either binary, it's best to keep these the same. When we get to version 1.3 of Consul K8s, we can start making changes because the old implementation will no longer be used.

.changelog/2038.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+api-gateway: Adds API Gateway command to CLI
```
t-eckert, line 2:
```suggestion
api-gateway: Add API Gateway subcommand to Control Plane.
```

Just to follow convention.

t-eckert (approved):
This is a great foundation!

control-plane/subcommand/api-gateway/command.go
missylbytes: Okay cool.

PR #2043, opened 2023-04-05
Customizing Vault Version for WanFed Test

20sr20:
Changes proposed in this PR:
- In the acceptance/tests, instead of hard coding Vault version, the vault version can be provided in the cli command.
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


acceptance/framework/config/config.go
```diff
@@ -52,6 +52,9 @@ type TestConfig struct {
 	ConsulVersion    *version.Version
 	EnvoyImage       string
 
+	VaultHelmChartVersion string
+	VaultServerVersion    *version.Version
```
kschoche, line 56: I *think* that `VaultServerVersion` can be treated as a string, iirc the reason `ConsulVersion` used type `version.Version` was so we could use a library to do the version comparisons for feature flagging: https://github.com/hashicorp/consul-k8s/blob/main/acceptance/tests/vault/vault_partitions_test.go#L34-L38

I don't think that the VaultServerVersion is going to be something that the acceptance tests would need to feature flag though, so it's probably okay to make this a string throughout!

acceptance/framework/vault/vault_cluster.go
```diff
@@ -29,6 +29,7 @@ import (
 const (
 	releaseLabel                 = "app.kubernetes.io/instance="
 	defaultVaultHelmChartVersion = "v0.21.0"
+	//defaultVaultHelmChartVersion = vaultVersion
```
eikenb, line 32: Please consider removing this.. it is generally better to not commit commented out code.

Ignore me if you're still working on this bit.

eikenb (changes requested):
After fixing the one type it will try to run the tests when I run it. I don't have kube installed locally, which seems to be required, so I can't test it further.

But this otherwise looks fine with the caveat that I couldn't actually run the test.

acceptance/framework/config/config.go
eikenb: I suggested switching to the type as the other code depended on it being that type (or at least not a string). You could also change those places to support a string if you'd rather go that way. Mainly the `nil` check and assignment below... (I'll comment there).

acceptance/framework/vault/vault_cluster.go
```diff
@@ -59,13 +59,20 @@ func NewVaultCluster(t *testing.T, ctx environment.TestContext, cfg *config.Test
 	if cfg.EnablePodSecurityPolicies {
 		values["global.psp.enable"] = "true"
 	}
+	if cfg.VaultServerVersion != nil {
+		values["server.image.tag"] = cfg.VaultServerVersion.String()
```
eikenb, line 63: If you change VaultServerVersion to a string, this would need to be updated to something like...
```go
if cfg.VaultServerVersion != "" {
    values["server.image.tag"] = cfg.VaultServerVersion
```

acceptance/framework/flags/flags.go
```diff
@@ -142,6 +146,7 @@ func (t *TestFlags) TestConfigFromFlags() *config.TestConfig {
 
 	// if the Version is empty consulVersion will be nil
 	consulVersion, _ := version.NewVersion(t.flagConsulVersion)
+	vaultserverVersion, _ := version.NewVersion(t.flagVaultServerVersion)
```
eikenb, line 149: You wouldn't need this anymore as this is what converts the string representation to a version.Version object.

eikenb (approved):
I'm not 100% on this code base but everything looks good and the tests run, so... :+1:

PR #2049, opened 2023-04-13
Net 3383/gatewayclassconfig controller

missylbytes:
Changes proposed in this PR:
- Adds a controller for the GatewayClassConfig

**How I've tested this PR:**
See below

Create a Kind cluster and install the local docker image:
```
make control-plane-dev-docker
docker tag consul-k8s-control-plane-dev consul-k8s-control-plane-dev:local
kind load docker-image consul-k8s-control-plane-dev:local
helm install consul ./charts/consul -n consul --set global.imageK8S=consul-k8s-control-plane-dev:local
```
Create a GatewayClassConfig by applying the file below with kubectl:
```
# GatewayClassConfig YAML File
kind: GatewayClassConfig
metadata:
  name: easy-to-see-gateway
```

You should see these logs in the connect-injector deployment:
```
 2023-04-14T15:06:42.638Z    INFO    controller.gateways    Reconcile the Gateway Class Config Controller    {"name": "easy-to-see-gateway", "namespace": ""}
2023-04-14T15:06:42.640Z    INFO    controller.gateways    Reconcile the Gateway Class Config Controller    {"name": "easy-to-see-gateway", "namespace": ""}
```

You should also be able to describe the GatewayClassConfig:
```
kubectl describe gatewayclassconfigs.consul.hashicorp.com
```
Checklist:
- [x] Tests added
- [x] CHANGELOG entry added


control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+const (
+	gatewayClassConfigFinalizer = "gateway-class-exists-finalizer.consul.hashicorp.com"
+)
+
+type Controller struct {
```
t-eckert, line 22: Can we rename this to `GatewayClassConfigController`? This will be in the same namespace as all the other controllers and they'll need to be named differently to avoid collisions.

Additionally, a short comment above the struct would be great to explain what this controller's responsibility is.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	client.Client
+
+	Log logr.Logger
+}
+
+// Reconcile reads the state of an Endpoints object for a Kubernetes Service and reconciles Consul services which
+// correspond to the Kubernetes Service. These events are driven by changes to the Pods backing the Kube service.
```
t-eckert, line 29: This is a vestige, my friend :P

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.Log.Info("Reconcile the Gateway Class Config Controller", "name", req.Name, "namespace", req.Namespace)
```
t-eckert, line 31: Probably better to log out "Reconciling" instead of "Reconcile".

control-plane/api-gateway/controllers/gateway_class_config_controller.go
missylbytes: Oh I cannot take credit for that, it is direct copy from old repo.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
t-eckert: Let's add a line before this defining a logger with the reconciler name and the namespaced name of the resource.

```go
log := r.Log.WithValues("gatewayClass", req.NamespacedName)
```

Then you can use `log` in the place of `r.Log` and it will always have the context printed along with it of what it was trying to reconcile when the info or error was logged.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	if err := r.Update(ctx, object); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// RemoveFinalizer ensures that the given finalizer is removed from the passed object
+// it returns a boolean saying whether a finalizer was removed, and any
+// potential errors.
+func (r *Controller) RemoveFinalizer(ctx context.Context, object client.Object, finalizer string) (bool, error) {
```
t-eckert, line 88: Great implementation, I was thinking this could be made private, but I'm now wondering if we can make this into a function (not associated with the controller) and use it across other controllers too!

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	paramaterRef := gc.Spec.ParametersRef
+	return paramaterRef != nil &&
+		string(paramaterRef.Group) == v1alpha1.ConsulHashicorpGroup &&
+		paramaterRef.Kind == v1alpha1.GatewayClassConfigKind &&
+		paramaterRef.Name == gcc.Name
+}
+
+// GatewayClassConfigInUse determines whether any GatewayClass in the cluster
+// references the provided GatewayClassConfig.
+func (r *Controller) GatewayClassConfigInUse(ctx context.Context, gcc *v1alpha1.GatewayClassConfig) (bool, error) {
```
t-eckert, line 120: I think this could also be private.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (r *Controller) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.GatewayClassConfig{}).
```
t-eckert, line 137: We also want to cover `v1beta2.GatewayClassConfig{}` and maybe `v1beta1.GatewayClassConfig{}`.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
andrewstucki: 😅

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if _, err := r.EnsureFinalizer(ctx, gcc, gatewayClassConfigFinalizer); err != nil {
+		r.Log.Error(err, "error adding gateway class config finalizer")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *Controller) EnsureFinalizer(ctx context.Context, object client.Object, finalizer string) (bool, error) {
```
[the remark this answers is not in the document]
andrewstucki: 👍 on making this private. In addition, can we actually just pass the kubernetes client as an argument to the method and make it a standalone method rather than being bound to the controller struct? We'll need something like this helper on our other gateway controllers as well, so might as well make this reusable from the get-go.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
andrewstucki: ah, just saw this comment, which is basically what I'm suggesting above, 👍

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+			return false, err
+		}
+	}
+	return found, nil
+}
+
+// gatewayClassUsesConfig determines whether a given GatewayClass references a
+// given GatewayClassConfig. Since these resources are scoped to the cluster,
+// namespace is not considered.
+func gatewayClassUsesConfig(gc gwv1beta1.GatewayClass, gcc *v1alpha1.GatewayClassConfig) bool {
```
andrewstucki, line 110: Just a thought, but we may be able to make this whole controller a little bit more terse and even less spammy by watching and using a custom index for `GatewayClass` objects that trigger upstream reconciliation of the `GatewayClassConfig` rather than just doing the 10 second retry loop. I'm thinking this is probably fine for now, but it might be worthwhile to modify or take another look in a second pass.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
andrewstucki: So, this is our internal CRD, not anything from the spec and all of our internal CRDs are `v1alpha1`, so this should be fine.

andrewstucki (changes requested):
A few changes, primarily on the log-level side of things just so we're not too spammy.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
t-eckert: Great minds or something

control-plane/api-gateway/controllers/gateway_class_config_controller.go
t-eckert: Oh gosh, that's right. I was just thinking of this in my own GatewayClass Controller code and got mixed up.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+		return false, err
+	}
+	return true, nil
+}
+
+// RemoveFinalizer ensures that the given finalizer is removed from the passed object
+// it returns a boolean saying whether a finalizer was removed, and any
+// potential errors.
+func (r *Controller) RemoveFinalizer(ctx context.Context, object client.Object, finalizer string) (bool, error) {
+	finalizers := []string{}
```
jm96441n, line 89: nit: since we know the length that this should be, what do you think about initializing this with the capacity up front:

```
existingFinalizers := object.GetFinalizers()
finalizers := make([]string, 0, len(existingFinalizers) -1)
```

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -6,12 +6,17 @@ package v1alpha1
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
 const (
 	GatewayClassConfigKind = "GatewayClassConfig"
 )
 
+func init() {
```
jm96441n, line 16: is there a way to do this without an init function? they can make the code more difficult to reason about (though I get if it's not easily doable)

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+	return found, nil
+}
+
+// gatewayClassUsesConfig determines whether a given GatewayClass references a
+// given GatewayClassConfig. Since these resources are scoped to the cluster,
+// namespace is not considered.
+func gatewayClassUsesConfig(gc gwv1beta1.GatewayClass, gcc *v1alpha1.GatewayClassConfig) bool {
+	paramaterRef := gc.Spec.ParametersRef
```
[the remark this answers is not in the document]
t-eckert: I'm such a rube, I was thinking of this guy:

control-plane/api-gateway/controllers/gateway_class_config_controller.go
missylbytes: So if I understand correctly, other people can put finalizers on our objects. So we wouldn't know for sure the capacity here.

jm96441n: ah gotcha, so `object.GetFinalizers()` could change in what it returns between calling it here and entering into the loop?

andrewstucki: `GetFinalizers` calls won't change between calls since we're acting on an object that we've already pulled, not re-reading from the API server.

RE the initial allocation, the reality is that this function is meant to remove a finalizer only if it exists, so doing the pre-capacity allocation could still technically trigger a resizing of the underlying slice. That said, go also does a multiplicative reallocation every time it needs to grow a slice capacity, doubling the size each time an reallocation occurs, so really we're:

1. Dealing with a fairly small number of elements,
2. Talking potentially doing a very small number of reallocations.
3. Dealing with resources that really shouldn't have reconciliation triggered that often

So... I really don't think that the pre-allocation matters or not here, stylistically feel free to do whichever you think makes the code scan easier.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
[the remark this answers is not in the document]
t-eckert: If you want to keep this as-is for now, I will add the EnsureFinalizer func and you can leverage it in a separate PR.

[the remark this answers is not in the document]
t-eckert: I need it for GatewayClass Controller anyways

control-plane/api-gateway/controllers/gateway_class_config_controller.go
t-eckert: I think we are ok skipping the allocation. Though I am in general a fan of sized allocation.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
missylbytes: I learned a new vocabulary word and also actually updated the comment to be from gateway repo

control-plane/api-gateway/controllers/gateway_class_config_controller.go
missylbytes: done

control-plane/api-gateway/controllers/gateway_class_config_controller.go
missylbytes: Just an FYI logr only has Info and Error. We would need to set the verbosity level of the log, so I am looking into what that means.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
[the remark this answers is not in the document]
missylbytes: fixed

control-plane/api-gateway/controllers/gateway_class_config_controller.go
missylbytes: done

control-plane/api/v1alpha1/api_gateway_types.go
missylbytes: I _think_ this is required by Kubebuilder to generate the CRDs @t-eckert might be able to answer this

t-eckert: It's unfortunately one of the ugly things about Kubebuilder 😢

control-plane/api-gateway/controllers/gateway_class_config_controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+func TestGatewayClassConfigReconcile(t *testing.T) {
+	t.Parallel()
+	deletionTimestamp := meta.Now()
+	cases := []struct {
+		name       string
+		k8sObjects func() []runtime.Object
+		expErr     string
+		reque      bool
+		requeAfter time.Duration
```
t-eckert, line 35: * Requeue

missylbytes: yeah but do we actually want to have to type that

control-plane/api-gateway/controllers/gateway_class_config_controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}{
+		{
+			name: "Happy Path",
```
t-eckert, line 38: Nit, could we call "Happy Path" "Successfully reconcile without any changes".

control-plane/api-gateway/controllers/gateway_class_config_controller_test.go
t-eckert: Lol, I believe you can do it

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Reconcile reads the state of an Endpoints object for a Kubernetes Service and reconciles Consul services which
+// correspond to the Kubernetes Service. These events are driven by changes to the Pods backing the Kube service.
+func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.Log.Info("Reconcile the Gateway Class Config Controller", "name", req.Name, "namespace", req.Namespace)
+
+	gcc := &v1alpha1.GatewayClassConfig{}
+	if err := r.Client.Get(ctx, req.NamespacedName, gcc); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.Log.Info("gateway class not found")
```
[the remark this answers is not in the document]
missylbytes: Logging is a larger discussion, will keep the same until https://github.com/hashicorp/consul-k8s/issues/1256 is resolved.

control-plane/api-gateway/controllers/gateway_class_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if !gcc.ObjectMeta.DeletionTimestamp.IsZero() {
+		// We have a deletion, ensure we're not in use.
+		used, err := r.GatewayClassConfigInUse(ctx, gcc)
+		if err != nil {
+			r.Log.Error(err, "failed to check if the gateway class config is still in use")
+			return ctrl.Result{}, err
+		}
+		if used {
+			r.Log.Info("gateway class config still in use")
```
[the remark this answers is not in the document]
missylbytes: Logging is a larger discussion, will keep the same until https://github.com/hashicorp/consul-k8s/issues/1256 is resolved.

control-plane/api-gateway/controllers/gateway_class_config_controller_test.go
missylbytes: Done

t-eckert (approved):
LGTM!

andrewstucki (approved):
Good enough for now! We might want to take a second pass at some of this in terms of optimizing lookups, triggering reconciliation rather than doing the 10s reschedule, etc. But all that can wait for follow-up.

PR #2055, opened 2023-04-14
Add GatewayClass Controller

t-eckert:
Changes proposed in this PR:
- Add `EnsureFinalizer` and `RemoveFinalizer` functions for adding and deleting finalizers from an object.
- Add a GatewayClass Controller which ensures that the correct finalizers are added and removed from the GatewayClass object.
- Give the connect-inject clusterrole the correct permissions for managing Gateway objects.

How I've tested this PR:
- With the new unit tests!
- Installed this version of Consul-k8s, created a GatewayClass and a Gateway. The controller would not let the GatewayClass get deleted until the Gateway was deleted.

You will need the Gateway API CRDs to make this work. Install them with:

``` bash
kubectl apply --kustomize http://github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.2
```

image

- Check with unit tests.

Checklist:
- [x] Tests added
~- [ ] CHANGELOG entry added~


charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -116,4 +116,232 @@ rules:
   verbs:
   - use
 {{- end }}
+- apiGroups:
+  - apps
+  resources:
+  - deployments
```
andrewstucki, line 122: Can we drop these resources until they're needed in the gateway controller to make the permissions a bit easier to review?

charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -116,4 +116,232 @@ rules:
...
+  verbs:
+  - update
+- apiGroups:
+  - gateway.networking.k8s.io
+  resources:
+  - gatewayclasses/status
+  verbs:
+  - get
+  - patch
+  - update
```
andrewstucki, line 250: I believe these are all we'll need for this particular PR.

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,101 @@
...
+// GatewayClassReconciler reconciles a GatewayClass object.
+// The GatewayClass is responsible for defining the behavior of API gateways
+// which reference the given class.
+type GatewayClassReconciler struct {
+	ControllerName string
+	Log            logr.Logger
+	client.Client
+}
+
+// +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gatewayclasses,verbs=get;list;watch;create;update;patch;delete
```
andrewstucki, line 27: Just wondering -- so this is usually used for generating RBAC permissions for a controller deployment using kubebuilder, but is our helm chart actually using these or is it manually updated? If it's manual, I think we can drop these annotations.

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Reconcile handles the reconciliation loop for GatewayClass objects.
+func (r *GatewayClassReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("gatewayClass", req.NamespacedName)
+
+	gc := &gwv1beta1.GatewayClass{}
+
+	err := r.Client.Get(ctx, req.NamespacedName, gc)
+	if err != nil {
+		log.Error(err, "unable to get GatewayClass")
```
andrewstucki, line 39: you'll want to do the `IsNotFound` check here or you're going to trigger infinite reconciliation if a gateway class has been fully GC'd.

control-plane/api-gateway/controllers/finalizer_test.go
```diff
@@ -0,0 +1,77 @@
...
+		shouldUpdate      bool
+	}{
+		"should update":     {[]string{}, finalizer, true},
+		"should not update": {[]string{finalizer}, finalizer, false},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			// It doesn't matter what the object is, as long as it implements client.Object.
+			// A Pod was as good as any other object here.
```
jm96441n, line 28: :+1: for this comment, definitely answered a question I had!

control-plane/api-gateway/controllers/finalizer_test.go
```diff
@@ -0,0 +1,77 @@
...
+					Finalizers: tc.initialFinalizers,
+				},
+			}
+
+			client := fake.NewClientBuilder().WithObjects(testObj).Build()
+
+			didUpdate, err := RemoveFinalizer(context.Background(), client, testObj, tc.finalizerToRemove)
+
+			require.NoError(t, err)
+			require.Equal(t, tc.shouldUpdate, didUpdate)
```
jm96441n, line 74: nit: what do you think of using something like `expectedUpdateValue` and `actualUpdateValue`? I got a little tripped up on the naming here

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,101 @@
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if string(gc.Spec.ControllerName) != r.ControllerName {
+		// This GatewayClass is not for this controller.
```
andrewstucki, line 44: In case the controller name changed, we should clean up any finalizers we've added here so that the object can actually be deleted. I know that the particular finalizer name is part of the spec, but this is another place where the upstream spec seems a bit odd.

Particularly, if a gateway class switches names from a valid controller name to an invalid controller name and it was previously in use (but now no longer is because it's invalid and can't be reconciled by a non-existent controller), then the class can never be deleted.

I'm thinking we can probably just write our own finalizer and then clean it up if the name switched just to have a clean separation between controller implementations and to make sure we're cleaning up the stuff we've added to the object -- though I think this is probably worth raising with upstream and I'm sure @mikemorris probably has some thoughts on it.

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,101 @@
...
+		// We have a deletion request. Ensure we are not in use.
+		used, err := r.isGatewayClassInUse(ctx, gc)
+		if err != nil {
+			log.Error(err, "unable to check if GatewayClass is in use")
+			return ctrl.Result{}, err
+		}
+		if used {
+			// Requeue after 10 seconds to check again.
+			log.Info("GatewayClass is in use, cannot delete")
+			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
andrewstucki, line 58: same with the implementation of the class config, I think we may want to move towards doing the edge triggering based off of a gateway update v. polling once something is soft deleted, but that can wait for follow-up.

control-plane/api-gateway/controllers/finalizer_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestEnsureFinalizer(t *testing.T) {
```
jm96441n, line 13: do we want to throw a `t.Parallel()` on this and the other test in this file?

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	didUpdate, err := EnsureFinalizer(ctx, r.Client, gc, GatewayClassFinalizer)
+	if err != nil {
+		log.Error(err, "unable to add finalizer")
+		return ctrl.Result{}, err
+	}
+	if didUpdate {
+		// We updated the GatewayClass, requeue to avoid another update.
+		return ctrl.Result{Requeue: true}, nil
+	}
+
```
andrewstucki, line 78: Here we should be validating the actual `parameterRef` object that the gatewayclass refers to. In particular:

1. any set `parametersRef` should reference only a `GatewayClassConfig` object, otherwise have a status set with `InvalidParameters` with a message saying we have the wrong parameter type
2. that `GatewayClassConfig` object should exist, otherwise also `InvalidParameters` with a message saying the parameter object doesn't exist
3. assuming that we are requiring `parametersRef` actually be set (which maybe we want to relax now that we're talking about stripping out all of the consul connection info from the config object?), then we'd also set an `InvalidParameters` condition if `parametersRef` is unspecified
4. otherwise, set `Accepted` == `True` with an `Accepted` reason.

All of this will need to be dirty-checked to make sure we don't infinitely trigger reconciliation.

andrewstucki (changes requested):

control-plane/api-gateway/controllers/finalizer.go
```diff
@@ -0,0 +1,41 @@
+package controllers
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// EnsureFinalizer ensures that the given object has the given finalizer.
+func EnsureFinalizer(ctx context.Context, client client.Client, object client.Object, finalizer string) (didUpdate bool, err error) {
```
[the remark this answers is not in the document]
t-eckert: It's just there for clarity. When I return a boolean and the function isn't in the form of `IsXYZ`, I like to name it so that it's clear what that boolean represents. I don't use the naked return though because that is danger.

control-plane/api-gateway/controllers/finalizer_test.go
t-eckert: Can't hurt!

control-plane/api-gateway/controllers/gatewayclass_controller.go
t-eckert: This is an absolute cargo cult scenario. I didn't really know what these did, but I included them in case I needed them.

control-plane/api-gateway/controllers/finalizer_test.go
t-eckert: I kind of like this naming because I thought it was clear. Can you help me understand what tripped you up about the naming? I wonder if there's something I'm not seeing.

control-plane/api-gateway/controllers/finalizer.go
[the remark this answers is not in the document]
jm96441n: oh yeah :-1:  on naked returns

control-plane/api-gateway/controllers/finalizer_test.go
jm96441n: I think it's mainly the `shouldUpdate`, like to me that reads as an assertion (like "should this update") rather than an expectation in conjunction with the names for the subtests being really close to this threw me off a little bit

t-eckert: I kind of split the difference

control-plane/api-gateway/controllers/gatewayclass_controller.go
t-eckert: So the logic here is that we would define our own finalizer -- say, `consul-gateway-finalizer` and if the controller name does not match, go remove the finalizer?

control-plane/api-gateway/controllers/gatewayclass_controller.go
t-eckert: Ah, I see. Instead of requeueing, just add Gateway as a triggering type?

control-plane/api-gateway/controllers/gatewayclass_controller.go
andrewstucki: Yep, that's my thought, I know the spec is prescriptive about the finalizer name, but frankly, it doesn't really make sense to me due to the reasons above.

control-plane/api-gateway/controllers/gatewayclass_controller.go
t-eckert: No worries! If it wasn't clear, I wouldn't be opposed to rewriting it to be more clear.

control-plane/api-gateway/controllers/gatewayclass_controller.go
andrewstucki: I think a replacement. I'd probably prefer the replacement in this implementation, and given that Mike is currently out, if we have some strong push-back in terms of a conformance to upstream, then we can go back and switch it.

control-plane/api-gateway/controllers/gatewayclass_controller.go
andrewstucki: Yep, but I think that can probably wait for a follow-up in which we handle both this and the gatewayclassconfig controller at the same time. Just moving towards strict edge-triggering for noise and performance v. polling every 10 seconds while something is in use.

control-plane/api-gateway/controllers/gatewayclass_controller_test.go
```diff
@@ -0,0 +1,176 @@
...
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "default",
+						Name:      "test-gateway",
+					},
+					Spec: gwv1beta1.GatewaySpec{
+						GatewayClassName: v1beta1.ObjectName(name),
+					},
+				},
+			},
+			expectedResult:     ctrl.Result{RequeueAfter: 10 * time.Second},
```
andrewstucki, line 137: these will need to be adjusted since we added the edge-triggering

control-plane/api-gateway/controllers/gatewayclass_controller_test.go
```diff
@@ -0,0 +1,176 @@
...
+			expectedError:      nil,
+			expectedFinalizers: []string{GatewayClassFinalizer},
+		},
```
andrewstucki, line 140: Can we add a couple tests for the `GatewayClassConfig` validation logic? Also for checking that the status is set appropriately?

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+			return ctrl.Result{}, err
+		}
+		if err != nil {
+			log.Error(err, "unable to fetch GatewayClassConfig")
+			return ctrl.Result{}, err
+		}
+	}
+
+	return ctrl.Result{}, nil
```
andrewstucki, line 118: One more change we'll need here, if we have no errors in validation, we should be setting the `GatewayClass` status to have an `Accepted == True` condition.

andrewstucki (changes requested):
Some minor refactoring suggestions and a few small changes, but think we're almost there.

charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -117,4 +117,30 @@ rules:
   verbs:
   - use
 {{- end }}
+- apiGroups:
```
missylbytes, line 120: missing a permission for list gateways

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+const (
+	GatewayClassFinalizer      = "gateway-exists-finalizer.consul.hashicorp.com"
+	GatewayClassControllerName = "hashicorp.com/consul-api-gateway-controller"
```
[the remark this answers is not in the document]
t-eckert: I made the finalizer private, but the controller name is used by the command.

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,269 @@
...
+				},
+			})
+		}
+
+		return requests
+	})
+}
+
+// gatewayFieldIndexEventHandler returns an EventHandler that will enqueue
+// reconcile.Requests for GatewayClass objects that reference the Gateway
```
missylbytes, line 244: nit: update comment here

t-eckert: Great catch!

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,268 @@
...
+		if err := r.Client.Status().Update(ctx, gc); err != nil {
+			return false, err
+		}
+
+		return true, nil
+	}
+
+	// Add a new condition to the GatewayClass.
+	gc.Status.Conditions = append(gc.Status.Conditions, condition)
+	gc.Status.Conditions[len(gc.Status.Conditions)-1].LastTransitionTime = metav1.Now()
```
andrewstucki, line 204: so I think a number of things here can be simplified or might need some tweaks:

1. you should be able to just do the `LastTransitionTime` setting on the `condition` as it's passed in here, or move it to some sort of constructor for conditions
2. you'll want to set the `ObservedGeneration` on each of these conditions you're creating
3. I think you can just do the above scan looking for the `Type` value, if it exists you can do the `equalConditions` check and return early, if not, then do the condition setting and set a boolean that says `updated` and break out of the loop
4. If `!updated` then append the condition
5. do the update in one place at the end of the function.

If you want you could even remove the call to `Update` and do it in one go after both the parameter validation and the accepted condition setting so that we don't have to do another whole reconciliation.

andrewstucki (changes requested):
So, we will want to set `ObservedGeneration` on each condition, and then some additional ideas for simplifying and/or avoiding unneeded reconciliation loops.

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,268 @@
...
+)
+
+const (
+	GatewayClassControllerName = "hashicorp.com/consul-api-gateway-controller"
+
+	gatewayClassFinalizer = "gateway-exists-finalizer.consul.hashicorp.com"
+
+	// GatewayClass status fields.
+	invalidParameters = "InvalidParameters"
+	accepted          = "Accepted"
```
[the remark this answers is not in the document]
t-eckert: Definitely! I thought of that too and wrote out an initial model using generics. I wanted to double back around to that when I got a better sense of what was going on in the different controllers and not over-optimize too early.

control-plane/api-gateway/controllers/gatewayclass_controller.go
t-eckert: I addressed these a bit differently from how you recommended, but I wanted to keep the condition setting simple and was able to make it simpler.

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,264 @@
...
+
+	// Set the condition if it already exists.
+	for i, c := range gc.Status.Conditions {
+		if c.Type == condition.Type {
+			// The condition already exists and is up to date.
+			if equalConditions(condition, c) {
+				return false
+			}
+
+			condition.ObservedGeneration = c.ObservedGeneration + 1
```
andrewstucki, line 196: The generation comes from the object that's being observed: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta.GetGeneration

What that means is that we should be doing `condition.ObservedGeneration = gc.GetGeneration()` and then using that as part of the equality check above (the generation value won't change on the object unless the spec has been updated, so it should be fine to do the status update and not worry about infinite loops even if the generation is part of the equality check)

andrewstucki (changes requested):
one last change on the usage of `ObservedGeneration`

control-plane/api-gateway/controllers/gatewayclass_controller.go
t-eckert: Great! Added!

control-plane/api-gateway/controllers/gatewayclass_controller.go
```diff
@@ -0,0 +1,264 @@
...
+	if didUpdate {
+		if err := r.Client.Update(ctx, gc); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, err
+	}
+	if err != nil {
+		log.Error(err, "unable to validate ParametersRef")
+		return ctrl.Result{}, err
+	}
```
andrewstucki, line 108: So, I think we might be able to write a test for this, but tracing through the logic since this is an idempotent call, assuming we're reconciling something that already has `InvalidParameters` set, then aren't all of these blocks going to be passed and we're going to set the status to `Accepted` as per below? If that's the case I think we'd have an infinite reconciliation loop because we'd be thrashing between `InvalidParameters` and `Accepted`.

Try running a `GatewayClass` with a finalizer already set and a non-existent `GatewayClassConfig` through reconciliation twice and I'm pretty sure you'll see it set `InvalidParameters` and then `Accepted`. If you do this manually I think you'd see reconciliation triggered without termination.

andrewstucki (changes requested):

andrewstucki (approved):

PR #2060, opened 2023-04-17
Add Hack for Generating CRDs from external sources

t-eckert:
For the Gateway API implementation to work, the target Kubernetes cluster must have the relevant CRDs installed. These CRDs are managed by the SIG Network group. We'll want to generate copies of these CRDs and store them in the `/crds` directory in our Helm chart so they can be installed for our users automatically.

This PR adds a `kustomization.yaml` file for managing external dependencies and a `generate-external-crds` command which will load in the dependencies and nicely format them in our `/crds` directory.

This will make it easy for us to keep the CRDs up to date and allows for extensibility if we ever need to add another set of external CRDs to the project in the future.

Changes proposed in this PR:
- Add `generate-external-crds`

How I've tested this PR:
- I ran the hack and it generated the correct CRDs which allowed them to be installed automatically when I installed Consul onto Kubernetes.

- You may also run the hack. 

Completes https://hashicorp.atlassian.net/browse/NET-3663 

Checklist:
~- [ ] Tests added~
~- [ ] CHANGELOG entry added~

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1219,7 +1219,7 @@ So that the documentation can look like:
 
 ## Managing External CRD Dependencies
 
-Some of the features of Consul on Kubernetes make use of custom resource definitions (CRDs) that we don't directly
+Some of the features of Consul on Kubernetes make use of CustomResourceDefinitions (CRDs) that we don't directly
 manage. One such example is the Gateway API CRDs which we use to configure API Gateways, but are managed by SIG
 Networking.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,7 @@
     1. [Writing Acceptance tests](#writing-acceptance-tests)
 1. [Using the Acceptance Test Framework to Debug](#using-acceptance-test-framework-to-debug)
 1. [Helm Reference Docs](#helm-reference-docs)
+1. [Managing External CRD Dependencies](#managing-external-crd-dependencies)
 1. [Adding a Changelog Entry](#adding-a-changelog-entry)
 
 ## Contributing 101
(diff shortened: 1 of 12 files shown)


hack/generate-external-crds/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.2
```
t-eckert, line 5: In the future, we can just change the values in this list and re-run the script to update CRDs.

CONTRIBUTING.md
```diff
@@ -24,6 +24,7 @@
     1. [Writing Acceptance tests](#writing-acceptance-tests)
 1. [Using the Acceptance Test Framework to Debug](#using-acceptance-test-framework-to-debug)
 1. [Helm Reference Docs](#helm-reference-docs)
+1. [Managing External CRD Dependencies](#managing-external-crd-dependencies)
```
thisisnotashwin, line 27: quick drive by question 🚗: Currently, these CRDs will always get installed with a helm install. We should add an `if` condition around them as we do with the other CRDs to ensure they only get installed on a cluster if the gateway is enabled. Was this change intentional?

thisisnotashwin: I guess it is ok for this PR since we arent merging it into `main` but it is definitely something we should consider before that happens. This looks really clever otherwise. Nicely done.

charts/consul/crds/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
lkysow, line 1: This shouldn't be in here I don't think

CONTRIBUTING.md
lkysow: This is using the Helm CRDs directory, not our normal CRDs method. That directory doesn't support templating.

charts/consul/crds/kustomization.yaml
lkysow: Oh I see this is being helm ignored. Can you add a comment at the top that this is helm ignored and only used for the make <whatever it is> command?

Makefile
```diff
@@ -123,6 +127,7 @@ lint: cni-plugin-lint ## Run linter in the control-plane, cli, and acceptance di
 ctrl-manifests: get-controller-gen ## Generate CRD manifests.
 	cd control-plane; $(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 	make copy-crds-to-chart
+	make generate-example-crds
```
lkysow, line 130:
```suggestion
	make generate-external-crds
```

lkysow (approved):

CONTRIBUTING.md
thisisnotashwin: Aaha! so this needs an explicit install! thanks for the clarification!

t-eckert: Good call out though! It's a pain and I tried to get around it, but there really wasn't a way to do it :shrug:

Makefile
t-eckert: 🤦🏻

CONTRIBUTING.md
```diff
@@ -1216,6 +1217,26 @@ So that the documentation can look like:
 - `ports` ((#v-ingressgateways-defaults-service-ports)) (`array<map>: [{port: 8080, port: 8443}]`) - Port docs
 ```
 
+## Managing External CRD Dependencies
+
+Some of the features of Consul on Kubernetes make use of custom resource definitions (CRDs) that we don't directly
```
nathancoleman, line 1222:
```suggestion
Some of the features of Consul on Kubernetes make use of Custom Resource Definitions (CRDs) that we don't directly
```

nathancoleman (approved):

andrewstucki (approved):

PR #2066, opened 2023-04-18
WIP: api-gateway resource conversion

jm96441n:
Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/controller/api-gateway/consul/config_entry_translation.go
+++ b/control-plane/controller/api-gateway/consul/config_entry_translation.go
@@ -8,6 +8,13 @@ import (
 	consulAPI "github.com/hashicorp/consul/api"
 )
 
+const (
+	metaKeyManagedBy       = "managed-by"
+	metaValueManagedBy     = "consul-k8s-gateway-controller"
+	metaKeyKubeNS          = "k8s-namespace"
+	metaKeyKubeServiceName = "k8s-service-name"
+)
+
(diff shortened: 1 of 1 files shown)


control-plane/controller/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,207 @@
...
+				LastTransitionTime: ptrTo(time.Now().UTC()),
+			}
+			conditions = append(conditions, listenerCondition)
+		}
+	}
+
+	return consulAPI.APIGatewayConfigEntry{
+		Kind:      k8sGW.Kind,
+		Name:      k8sGW.Name,
+		Meta:      map[string]string{}, // TODO: what should go in here?
```
jm96441n, line 85: this is kinda my main question: what meta do we copy over from the k8s type?

andrewstucki: So, here's an example of what gets used for our service registrations, we'll probably want something roughly similar to mark things as managed by us and, due to the way we transform names/namespaces to/from Kubernetes, something to mark what Kubernetes objects these actually correspond to:

https://github.com/hashicorp/consul-k8s/blob/11f842cbe94b5df544c03e2441a1820397f92b9b/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go#L404-L410

jm96441n:
going to re-open usng the ticket in the branch name

PR #2070, opened 2023-04-19
APIGW Resource Translation

jm96441n:
Changes proposed in this PR:
- Add function to translate k8s APIGW resource into a Consul Config Entry

How I've tested this PR:
Ran unit tests

Read code/run unit tests

Checklist:
- [ X] Tests added
- [ ] CHANGELOG entry added 


control-plane/controller/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,100 @@
...
+					SectionName: string(listener.Name),
+					Namespace:   k8sGW.Namespace,
+				},
+				LastTransitionTime: &condition.LastTransitionTime.Time,
+			}
+			conditions = append(conditions, listenerCondition)
+		}
+	}
+
+	return consulAPI.APIGatewayConfigEntry{
```
jm96441n, line 82: how do we support admin partitions here? is there something in the k8s definition that would hint to use what admin partition this should be in? also would we be worried about if a user was not mirroring consul namespaces/do we have a way to check that at runtime?

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,100 @@
...
+				Kind:      consulAPI.APIGateway,
+				Name:      k8sGW.Name,
+				Namespace: k8sGW.Namespace,
+			},
+			LastTransitionTime: &condition.LastTransitionTime.Time,
+		}
+		conditions = append(conditions, c)
+	}
+
+	for _, listener := range k8sGW.Status.Listeners {
```
andrewstucki, line 63: same here

andrewstucki (changes requested):

control-plane/controller/api-gateway/consul/config_entry_translation.go
andrewstucki: So, the admin partitions are something that a consul-k8s installation handles globally. Anything we synchronize over to Consul will wind up in a single partition since we set the Consul client in our deployment to use a single partition via env variable:

https://github.com/hashicorp/consul-k8s/blob/11f842cbe94b5df544c03e2441a1820397f92b9b/charts/consul/templates/_helpers.tpl#L309-L311

If you want to be explicit about writing the partition value in, you'll want to pull that out of the env variable, otherwise you might be able to rely on the defaulting normalization behavior in our Consul entries (though that may make some of the dirty-checking a bit more complex for determining whether we need to actually write back to Consul or not).

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+
+	return consulAPI.APIGatewayConfigEntry{
+		Kind: api.APIGateway,
+		Name: k8sGW.Name,
+		Meta: map[string]string{
+			metaKeyManagedBy:       metaValueManagedBy,
+			metaKeyKubeNS:          k8sGW.GetObjectMeta().GetNamespace(),
+			metaKeyKubeServiceName: k8sGW.GetObjectMeta().GetName(),
```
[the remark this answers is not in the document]
andrewstucki: oh, whoops, nvm, misread this, this is correct.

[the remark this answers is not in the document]
jm96441n: I'd assume we'd still want to do the namespace mapping right? especially if consul namespaces aren't enabled?

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,100 @@
...
+	conditions := make([]consulAPI.Condition, 0, len(k8sGW.Status.Conditions))
+	for _, listener := range k8sGW.Spec.Listeners {
+		certificates := make([]consulAPI.ResourceReference, 0, len(listener.TLS.CertificateRefs))
+		for _, certificate := range listener.TLS.CertificateRefs {
+			c := consulAPI.ResourceReference{
+				Kind:        consulAPI.InlineCertificate,
+				Name:        string(certificate.Name),
+				SectionName: "",
+				Partition:   "",
+				Namespace:   string(*certificate.Namespace),
```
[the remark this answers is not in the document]
jm96441n: and I'm assuming we'll need to do the same namespace mapping anywhere we use namespaces right?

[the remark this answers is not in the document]
jm96441n: and if we use the consulIdentifier map we're just pushing up the namespace resolution and annotation overrides to the caller right?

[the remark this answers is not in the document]
andrewstucki: Yep, we'd be pushing it up to the caller, if we wanted I could even see doing something like generating a map of all of the names/namespaces/types we need created prior to the actual resource field mapping code just to make doing the name/namespace resolution easier. It'd require two passes across all of the resources we want generated, but I think it'll make some of the ref calculation code easier since we won't have to do any sort of annotation/namespace lookups there.

RE this:
> and I'm assuming we'll need to do the same namespace mapping anywhere we use namespaces right?

yep.

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,94 @@
...
+	Namespace string
+	Name      string
+}
+
+type Translator struct {
+	EnableConsulNamespaces bool
+	ConsulDestNamespace    string
+	EnableK8sMirroring     bool
+	MirroringPrefix        string
+}
```
andrewstucki, line 39: 👍

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,94 @@
...
+type consulIdentifier struct {
+	name      string
+	namespace string
+	partition string
+}
+
+type NamespaceName struct {
+	Namespace string
+	Name      string
+}
```
andrewstucki, line 32: should be able to use `types.NamespacedName` from `"k8s.io/apimachinery/pkg/types"`

jm96441n: oh good call, I was looking for a `types` package in this codebase

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"os"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/types"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
+	"github.com/hashicorp/consul/api"
+	consulAPI "github.com/hashicorp/consul/api"
```
t-eckert, line 12: Generally, we will call this `capi` in the Consul on Kubernetes codebase.

t-eckert (approved):
This is fantastic!

control-plane/api-gateway/consul/config_entry_translation.go
jm96441n: updated!

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
+	"github.com/hashicorp/consul/api"
+	capi "github.com/hashicorp/consul/api"
+)
+
+const (
+	metaKeyManagedBy       = "managed-by"
+	metaValueManagedBy     = "consul-k8s-gateway-controller"
+	metaKeyKubeNS          = "k8s-namespace"
+	metaKeyKubeServiceName = "k8s-service-name"
```
andrewstucki, line 19: nit: I believe most of these are already defined elsewhere, so I'm pretty sure we could re-use them.

andrewstucki (approved):
Non-blocking comment, thanks for churning this out!

control-plane/api-gateway/consul/config_entry_translation.go
andrewstucki: I'm pretty sure that every place we use those they have the exact same value (`metaKeyKubeServiceName` appears to be defined in multiple packages)... so I'd definitely lean towards consolidation but I'd also be fine doing that later too.

jm96441n: I'm down to do that in a separate PR because that'll be a change that touches a lot of files

PR #2083, opened 2023-04-24
Adjust API gateway controller deployment appropriately when Vault configured as secrets backend

nathancoleman:
Consider whether Vault is being used as a secrets backend when determining whether to mount a TLS cert into the API gateway controller deployment.

Largely, this change is based off of how other deployments already handle Vault as a secrets backend. For example:

https://github.com/hashicorp/consul-k8s/blob/112ad18e12683559413c058d8e52f28d66597b07/charts/consul/templates/connect-inject-deployment.yaml#L293-L297

The only unique aspect here is that the api-gateway-controller communicates directly with agents when `client.enabled=true`, so we require some additional gating to make sure that we're using the correct cert in this case.

**How I've tested this PR:**
Install Consul w/ API gateway enabled in configurations both with and without Vault used as a secrets backend. In each case, verify that the API gateway controller is able to start up successfully and that deployed API gateways function correctly.

Install Consul w/ external server configuration (HCP, for example). Test with and without clients enabled.

**Checklist:**
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/2083.txt
+++ b/.changelog/2083.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api-gateway: fix issue where when Vault is configured as the secrets backend, the API Gateway controller is unable to start up successfully
+```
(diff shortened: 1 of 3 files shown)


david-yu:
We probably also need to backport to 1.1.x?

andrewstucki (commented):
Can we have a quick unit test for this?

nathancoleman:
@andrewstucki working on a test as we speak 😉

andrewstucki (approved):

PR #2088, opened 2023-04-26
APIGW4CONK8S: HTTP Route/TCPRoute/Secrets Translation

jm96441n:
Changes proposed in this PR:
- translate http routes in k8s to consul config entries
- translate tcp routes in k8s to consul config entries
- translate secrets in k8s to consul config entries

How I've tested this PR:
Unit Tests

Read the code/run unit tests

Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added 


control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -82,8 +90,198 @@ func (t Translator) GatewayToAPIGateway(k8sGW gwv1beta1.Gateway, certs map[types
...
+func translateHTTPRouteParentRefs(k8sParentRefs []gwv1beta1.ParentReference, parentRefs map[types.NamespacedName]consulIdentifier) []capi.ResourceReference {
+	parents := make([]capi.ResourceReference, 0, len(k8sParentRefs))
+	for _, k8sParentRef := range k8sParentRefs {
+		parentRef, ok := parentRefs[types.NamespacedName{Name: string(k8sParentRef.Name), Namespace: string(*k8sParentRef.Namespace)}]
+		if !ok {
+			// we drop any parent refs that consul does not know about
+			continue
+		}
+		ref := capi.ResourceReference{
+			Kind:        capi.APIGateway, // Will this ever not be a gateway? is that something we need to handle?
```
jm96441n, line 144: can we rely on a parent ref always being a gateway?

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -82,8 +90,198 @@ func (t Translator) GatewayToAPIGateway(k8sGW gwv1beta1.Gateway, certs map[types
...
+	consulHTTPRoute.Rules = t.translateHTTPRouteRules(k8sHTTPRoute.Spec.Rules)
+
+	return consulHTTPRoute
+}
+
+func translateHTTPRouteParentRefs(k8sParentRefs []gwv1beta1.ParentReference, parentRefs map[types.NamespacedName]consulIdentifier) []capi.ResourceReference {
+	parents := make([]capi.ResourceReference, 0, len(k8sParentRefs))
+	for _, k8sParentRef := range k8sParentRefs {
+		parentRef, ok := parentRefs[types.NamespacedName{Name: string(k8sParentRef.Name), Namespace: string(*k8sParentRef.Namespace)}]
+		if !ok {
```
jm96441n, line 139: is this a condition worth checking?

andrewstucki: I think we can keep it for now, we'll need to work through when we actually might have dropped parent refs though.

control-plane/api-gateway/consul/config_entry_translation.go
andrewstucki: no, we can't, we'll need to revalidate here and only translate refs that correspond to gateways since that's all we can handle (but the spec allows for refs of other implementation-defined types too), but I'm thinking that maybe we can just add this as a continuation clause a bit earlier? Something like this logic (which will also likely need to exist in the controller itself, so we'll likely want to eventually extract into something like ahelper somewhere):

```
if (k8sParentRef.Kind != nil && string(*k8sParentRef.Kind) != apiGatewaySpecKind) || 
   (k8sParentRef.Group != nil && string(*k8sParentRef.Group) != apiGatewaySpecGroup) {
  continue
} 
```

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -82,8 +90,198 @@ func (t Translator) GatewayToAPIGateway(k8sGW gwv1beta1.Gateway, certs map[types
...
+			// we drop any parent refs that consul does not know about
+			continue
+		}
+		ref := capi.ResourceReference{
+			Kind:        capi.APIGateway, // Will this ever not be a gateway? is that something we need to handle?
+			Name:        parentRef.name,
+			SectionName: string(*k8sParentRef.SectionName),
```
andrewstucki, line 146: this is not required, so this could result in a `nil`-pointer deref

andrewstucki (changes requested):
Initial pass

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -27,13 +33,15 @@ type consulIdentifier struct {
...
 
+// Translator handles translating K8s resources into Consul config entries.
 type Translator struct {
 	EnableConsulNamespaces bool
 	ConsulDestNamespace    string
 	EnableK8sMirroring     bool
 	MirroringPrefix        string
 }
 
+// GatewayToAPIGateway translates a k8s API gateway into a consul API gateway.
```
missylbytes, line 44: Nit: capital Consul?

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -18,7 +22,14 @@ const (
 	metaKeyKubeNS          = "k8s-namespace"
 	metaKeyKubeServiceName = "k8s-service-name"
 
+	// AnnotationGateway is the annotation used to override the gateway name.
 	AnnotationGateway = "consul.hashicorp.com/gateway"
+	// AnnotationHTTPRoute is the annotation used to override the http route name.
+	AnnotationHTTPRoute = "consul.hashicorp.com/http-route"
+	// AnnotationTCPRoute is the annotation used to override the http route name.
```
missylbytes, line 29: nit: override the tcp route name

jm96441n: whoooops, good catch

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -82,8 +100,279 @@ func (t Translator) GatewayToAPIGateway(k8sGW gwv1beta1.Gateway, certs map[types
...
 		Listeners: listeners,
 		Partition: consulPartition,
-		Namespace: namespaces.ConsulNamespace(k8sGW.GetObjectMeta().GetNamespace(), t.EnableK8sMirroring, t.ConsulDestNamespace, t.EnableK8sMirroring, t.MirroringPrefix),
+		Namespace: t.getConsulNamespace(k8sGW.GetObjectMeta().GetNamespace()),
+	}
+}
+
+// HTTPRouteToHTTPRoute translates a k8s HTTPRoute into a Consul HTTPRoute Config Entry.
+func (t Translator) HTTPRouteToHTTPRoute(k8sHTTPRoute gwv1beta1.HTTPRoute, parentRefs map[types.NamespacedName]consulIdentifier) capi.HTTPRouteConfigEntry {
+	consulPartition := os.Getenv("CONSUL_PARTITION")
```
t-eckert, line 109: Instead of getting this from the environment, I recommend you have it set as a field on the `Translator`. We will instantiate this in the `inject-connect` subcommand where it is passed in and then pass the `Translator` to the `GatewayController`.

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -82,8 +100,279 @@ func (t Translator) GatewayToAPIGateway(k8sGW gwv1beta1.Gateway, certs map[types
...
+	hostnames := make([]string, 0, len(k8sHTTPRoute.Spec.Hostnames))
+	for _, k8Hostname := range k8sHTTPRoute.Spec.Hostnames {
+		hostnames = append(hostnames, string(k8Hostname))
+	}
+	consulHTTPRoute.Hostnames = hostnames
+
+	// translate parent refs
+	consulHTTPRoute.Parents = translateRouteParentRefs(k8sHTTPRoute.Spec.CommonRouteSpec.ParentRefs, parentRefs)
+
+	// translate rules
```
t-eckert, line 139: You might get a linting error that these have to be "Sentences".

control-plane/api-gateway/consul/config_entry_translation.go
```diff
@@ -82,8 +90,206 @@ func (t Translator) GatewayToAPIGateway(k8sGW gwv1beta1.Gateway, certs map[types
...
+
+	return consulHTTPRoute
+}
+
+func translateHTTPRouteParentRefs(k8sParentRefs []gwv1beta1.ParentReference, parentRefs map[types.NamespacedName]consulIdentifier) []capi.ResourceReference {
+	parents := make([]capi.ResourceReference, 0, len(k8sParentRefs))
+	for _, k8sParentRef := range k8sParentRefs {
+		parentRef, ok := parentRefs[types.NamespacedName{Name: string(k8sParentRef.Name), Namespace: string(*k8sParentRef.Namespace)}]
+		if !(ok && isRefAPIGateway(k8sParentRef)) {
+			// we drop any parent refs that consul does not know about
```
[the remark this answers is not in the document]
t-eckert: I'm not certain, @andrewstucki?

t-eckert (approved):
Left some comments to fix, but approving optimistically. Great work on tackling this edgecase manifold.

control-plane/api-gateway/consul/config_entry_translation.go
jm96441n: I think that only applies to comments on methods rather than in method comments

PR #2091, opened 2023-04-27
Net 3508/gateway controller

missylbytes:
Changes proposed in this PR:
- Adds a stub-out of the Gateway Controller

How I've tested this PR:
With the steps below. Basically wanted to make sure the permissions and schemes were there for the controller to be able to react to changes on the object types.

```
kind create cluster
make control-plane-dev-docker
docker tag consul-k8s-control-plane-dev consul-k8s-control-plane-dev:local
kind load docker-image consul-k8s-control-plane-dev:local
helm install consul ./charts/consul -n consul --set global.imageK8S=consul-k8s-control-plane-dev:local
```
Then watch the logs for the connect-inject pod when you apply the following resource types, there shouldn't be any errors about access/nonexistent types.
`kubectl apply -f gateway-stuff.yml`
```
# gateway-stuff.yml
apiVersion: v1
kind: Namespace
metadata:
  name: gateway-namespace
---
kind: GatewayClassConfig
metadata:
  name: example-gateway-class-config
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: GatewayClass
metadata:
  name: example-gateway-class
spec:
  controllerName: 'hashicorp.com/consul-api-gateway-controller'
  parametersRef:
    group: api-gateway.consul.hashicorp.com
    kind: GatewayClassConfig
    name: example-gateway-class-config
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: example-gateway
  namespace: gateway-namespace
spec:
  gatewayClassName: example-gateway-class
  listeners:
    - protocol: HTTPS
      port: 443
      name: https
      allowedRoutes:
        namespaces:
          from: Same
      tls:
        certificateRefs:
          - name: cert
            namespace: secret-namespace
            group: ""
            kind: Secret
    - protocol: tcp
      port: 8888
      name: tcp
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: example-route
spec:
  parentRefs:
    - name: example-gateway
      namespace: gateway-namespace
  rules:
    - backendRefs:
        - kind: Service
          name: echo
          port: 8080

---
apiVersion: v1
kind: Namespace
metadata:
  name: secret-namespace
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: ReferenceGrant
metadata:
  name: reference-grant
  namespace: secret-namespace
spec:
  from:
    - group: gateway.networking.k8s.io
      kind: Gateway
      namespace: gateway-namespace
  to:
    - group: ""
      kind: Secret
      name: cert
---
apiVersion: v1
kind: Secret
metadata:
  name: cert
  namespace: secret-namespace
data:
  password: ZmFrZS1wYXNzd29yZA==/
```
Checklist:
- [x] Tests added (mainly just, does it build tests. More functionality to come)

### DIFF
--- a/charts/consul/templates/connect-inject-clusterrole.yaml
+++ b/charts/consul/templates/connect-inject-clusterrole.yaml
@@ -122,6 +122,8 @@ rules:
   resources:
   - gatewayclasses
   - gateways
+  - httproutes
+  - tcproutes
   verbs:
   - create
   - delete
@@ -144,4 +146,27 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+    - apps
+  resources:
+    - deployments
+  verbs:
+    - create
+    - get
+    - list
+    - update
+    - watch
+- apiGroups:
+    - core
+  resources:
+    - referencegrants
+    - services
+  verbs:
+    - watch
+- apiGroups: [ "" ]
+  resources: [ "secrets", "serviceaccounts" ]
+  verbs:
+    - "get"
+    - "list"
+    - "watch"
 {{- end }}
--- a/charts/consul/templates/connect-inject-clusterrole.yaml
+++ b/charts/consul/templates/connect-inject-clusterrole.yaml
@@ -164,7 +164,7 @@ rules:
   verbs:
     - watch
 - apiGroups: [ "" ]
-  resources: [ "secrets", "serviceaccounts" ]
+  resources: [ "secrets" ]
   verbs:
     - "get"
     - "list"
--- a/charts/consul/templates/connect-inject-clusterrole.yaml
+++ b/charts/consul/templates/connect-inject-clusterrole.yaml
@@ -124,6 +124,7 @@ rules:
   - gateways
   - httproutes
   - tcproutes
+  - referencegrants
   verbs:
   - create
   - delete
@@ -159,10 +160,10 @@ rules:
 - apiGroups:
     - core
   resources:
-    - referencegrants
     - services
   verbs:
     - watch
+    - list
 - apiGroups: [ "" ]
   resources: [ "secrets" ]
   verbs:
--- a/control-plane/api-gateway/controllers/gateway_controller.go
+++ b/control-plane/api-gateway/controllers/gateway_controller.go
@@ -90,7 +90,7 @@ func (r *GatewayController) SetupWithManager(ctx context.Context, mgr ctrl.Manag
 			source.NewKindWithCache(&gwv1alpha2.TCPRoute{}, mgr.GetCache()), r.tcpRouteFieldIndexEventHandler(ctx),
 		).
 		Watches(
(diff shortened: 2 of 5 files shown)


charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -144,4 +147,27 @@ rules:
...
+    - watch
+- apiGroups:
+    - core
+  resources:
+    - services
+  verbs:
+    - watch
+    - list
+- apiGroups: [ "" ]
+  resources: [ "secrets" ]
```
missylbytes, line 168: Should we put this behind a "if gateways.enabled"?

t-eckert: No, we decided to just have API Gateways gated behind "connectInject.enabled".

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	log := r.Log.WithValues("gateway", req.NamespacedName)
+	log.Info("Reconciling the Gateway: ", req.Name)
+
+	// If gateway does not exist, log an error.
+	gw := &gwv1beta1.Gateway{}
+	err := r.Client.Get(ctx, req.NamespacedName, gw)
+	if err != nil {
+		log.Error(err, "unable to get Gateway")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
```
t-eckert, line 45: This is the same thing I originally did and found that it isn't quite what we need because it will print a scary error under the normal operation of discovering a gateway does not exist. 

```suggestion
	gw := &gwv1beta1.Gateway{}
	err := r.Client.Get(ctx, req.NamespacedName, gw)
	if err != nil {
	    if k8serror.IsNotFound(err) {
	        return ctrl.Result{}, nil
	    }
		log.Error(err, "unable to get Gateway")
		return ctrl.Result{}, err
	}
```

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		if k8serrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		log.Error(err, "unable to get GatewayClass")
+		return ctrl.Result{}, err
+	}
+
+	if string(gwc.Spec.ControllerName) != GatewayClassControllerName || !gw.ObjectMeta.DeletionTimestamp.IsZero() {
+		// This Gateway is not for this controller or the gateway is being deleted.
+		// TODO: Cleanup Consul resources.
```
t-eckert, line 60: This is where we'll stick the `gatekeeper.DeleteGateway` call and also something else to inform Consul.

t-eckert (approved):
I think you can merge this as a first pass that we will build on. Thank you for creating this foundation!

PR #2093, opened 2023-04-27
Set Consul Server Locality From Node Label

erichaberkorn:
Changes proposed in this PR:
- Set Consul server locality From its topology region label.

How I've tested this PR:
- Unit tests

I have 4 K8s clusters that I'm using to test sameness and locality aware failover. I'll be automating this setup into an acceptance test over the next few weeks. Here are the steps I took:
1. I added labels to the nodes for each K8s cluster with:
  * `kubectl --context k3d-c1 label node k3d-c1-server-0 topology.kubernetes.io/region="us-east-1"`
  * `kubectl --context k3d-c3 label node k3d-c3-server-0 topology.kubernetes.io/region="us-east-2"`
  * `kubectl --context k3d-c4 label node k3d-c4-server-0 topology.kubernetes.io/region="us-west-2"`
2. I then peered the partition in `k3d-c1` to the the clusters in `k3d-c3` and `k3d-c4`
3. Run `curl -k 'https://localhost:8501/v1/peerings?token=???' | jq ".[] | [.Name, .Remote.Locality]"` and get the following:

```
[
  "cluster-03",
  {
    "Region": "us-east-2",
    "Zone": ""
  }
]
[
  "cluster-04",
  {
    "Region": "us-west-2",
    "Zone": ""
  }
]
```

This means that the Consul server locality was not just set on the Consul servers, but it also was successfully replicated over the peer stream.

👀 

Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/server-statefulset.yaml
+++ b/charts/consul/templates/server-statefulset.yaml
@@ -217,6 +217,22 @@ spec:
       {{- if .Values.server.priorityClassName }}
       priorityClassName: {{ .Values.server.priorityClassName | quote }}
       {{- end }}
+      initContainers:
+      - name: locality-init
+        image: {{ .Values.global.imageK8S }}
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            consul-k8s-control-plane fetch-server-region -node-name "$NODE_NAME" -output-file /consul/extra-config/locality.json
+        volumeMounts:
+          - name: extra-config
+            mountPath: /consul/extra-config
       containers:
         - name: consul
           image: "{{ default .Values.global.image .Values.server.image }}"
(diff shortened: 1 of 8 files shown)


control-plane/subcommand/fetch-server-region/command.go
```diff
@@ -0,0 +1,148 @@
...
+
+	config := c.fetchLocalityConfig()
+
+	jsonData, err := json.Marshal(config)
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	fmt.Println(string(jsonData))
```
thisisnotashwin, line 120: This subcommand could take a file path as a flag and write the contents of the config to that file too. It is a pattern we have used in other subcommands as well.

charts/consul/templates/client-daemonset.yaml
```diff
@@ -387,6 +387,7 @@ spec:
                 -recursor={{ quote $value }} \
                 {{- end }}
                 -config-file=/consul/extra-config/extra-from-values.json \
+                -config-file=/consul/extra-config/locality.json \
```
thisisnotashwin, line 390: How does the config file get configured here? we need the init container here too right? or am i missing something?

charts/consul/templates/server-statefulset.yaml
```diff
@@ -217,6 +217,22 @@ spec:
...
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            consul-k8s-control-plane fetch-server-region -node-name "$NODE_NAME" -output-file /consul/extra-config/locality.json
```
thisisnotashwin, line 232: ❤️

thisisnotashwin (commented):
I really like the approach! this does not look like someone writing a subcommand for the first time!! Outside of the change potentially required on the client daemonset, this looks really good!

thisisnotashwin (approved):
:chef-kiss:

PR #2105, opened 2023-05-04
chore(ci): fix changelog action for non-main base branches

DanStough:
Changes proposed in this PR:
- Fix the changelog checker action for when the base branch is not `main`. In this case the regex gets split because of the lack of quotes, causing an error. Also, the diff should be performed against the desired base branch in this case.

How I've tested this PR:
* This one is failing: https://github.com/hashicorp/consul-k8s/actions/runs/4887457091
* This one is passing with the changes: https://github.com/hashicorp/consul-k8s/actions/runs/4887398414

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -36,7 +36,7 @@ jobs:
             changelog_file_path=".changelog/[_0-9]*.txt"
           fi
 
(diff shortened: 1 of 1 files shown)


malizz (approved):
Thanks, I think we should do similar changes for backport (if branch is not main, don't backport).

PR #2118, opened 2023-05-10
APIGW4CONK8s: Add Consul Cache

jm96441n:
- add cache for consul resources
- additional translations of consul resources to k8s resources
- renames some translators to make them more descriptive
- modifies the SetupWithManager for the GatewayController to setup the cache and associated translators
- adds the setup to the inject command
How I've tested this PR:

Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added 


control-plane/api-gateway/translation/config_entry_translation.go
```diff
@@ -20,6 +20,7 @@ const (
 	metaValueManagedBy     = "consul-k8s-gateway-controller"
 	metaKeyKubeNS          = "k8s-namespace"
 	metaKeyKubeServiceName = "k8s-service-name"
+	metaKeyKubeName        = "k8s-name"
```
jm96441n, line 23: this came from the POC implementation https://github.com/hashicorp/consul-k8s/blob/408114a1fd7c82f2760839a2f22cc77c840208cb/control-plane/controller/apigateway_controller.go#L416, I'm not 100% on which we use as `metaKeyKubeService` name elsewhere in this repo (or do we use both keys in the meta?)

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -24,10 +30,20 @@ const (
 	kindGateway = "Gateway"
 )
 
+// GatewayControllerConfig holds the values necessary for configuring the GatewayController
+type GatewayControllerConfig struct {
+	ConsulClientConfig  *consul.Config
+	ConsulServerConnMgr consul.ServerConnectionManager
+	NamespacesEnabled   bool
+	Partition           string
+	Logger              logr.Logger
```
t-eckert, line 39: I think you could remove the Logger here and just use the Logger that the user has passed in on the `mgr` so both the Controller and the Cache are using the same logger.

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -90,14 +106,31 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-func (r *GatewayController) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
+// SetupWithGatewayControllerManager registers the controller with the given manager.
+func SetupGatewayControllerWithManager(ctx context.Context, mgr ctrl.Manager, config GatewayControllerConfig) (*cache.Cache, error) {
+	c := cache.New(cache.Config{
+		ConsulClientConfig:  config.ConsulClientConfig,
+		ConsulServerConnMgr: config.ConsulServerConnMgr,
+		NamespacesEnabled:   config.NamespacesEnabled,
+		Partition:           config.Partition,
+		Kinds:               []string{api.APIGateway, api.HTTPRoute, api.TCPRoute, api.InlineCertificate},
```
t-eckert, line 121: I think this would be worth pulling into a constant in the file where Cache config is defined. A user won't change this and we can keep the configuration as close to the cache as possible and in the code.

control-plane/api-gateway/translation/config_entry_translation.go
t-eckert: I think we can keep it around and if we get to the end of the implementation and find that we aren't using it, we can toss it.

control-plane/api-gateway/translation/k8s_cache_translation.go
```diff
@@ -0,0 +1,131 @@
...
+type TranslatorFn func(api.ConfigEntry) []types.NamespacedName
+
+type secretTransfomer func(context.Context) func(client.Object) []reconcile.Request
+
+type resourceGetter interface {
+	Get(api.ResourceReference) api.ConfigEntry
+}
+
+// ConsulToNSNTranslator handles translating consul config entries to k8s namespaced names
+type ConsulToNSNTranslator struct {
```
t-eckert, line 24:
```suggestion
type ConsulToNamespacedNameTranslator struct {
```
Takes longer to read, but saved me a few seconds of furrowing my brows to figure out what NSN was in this context. That will help keep my wrinkle-free for the years to come.

jm96441n: good call!

control-plane/api-gateway/translation/k8s_cache_translation.go
```diff
@@ -0,0 +1,131 @@
...
+func (c ConsulToNSNTranslator) TranslateConsulGateway(ctx context.Context) TranslatorFn {
+	return func(config api.ConfigEntry) []types.NamespacedName {
+		meta, ok := metaToK8sNamespacedName(config)
+		if !ok {
+			return nil
+		}
+
+		return []types.NamespacedName{meta}
+	}
+}
```
t-eckert, line 43: Does the comment/function name need to change here? It seems this function builds a translator function, rather than doing the translation.

control-plane/api-gateway/translation/k8s_cache_translation.go
```diff
@@ -0,0 +1,131 @@
...
+func (c ConsulToNSNTranslator) TranslateConsulTCPRoute(ctx context.Context) TranslatorFn {
+	return func(config api.ConfigEntry) []types.NamespacedName {
+		route, ok := config.(*api.TCPRouteConfigEntry)
+		if !ok {
+			return nil
+		}
+
+		return consulRefsToNSN(c.cache, route.Parents)
+	}
+}
```
t-eckert, line 67: Same here as above, these seems to build functions rather than executing the translation.

control-plane/api-gateway/controllers/gateway_controller.go
jm96441n: ohhh good call

control-plane/api-gateway/controllers/gateway_controller.go
jm96441n: I like that idea a lot

control-plane/api-gateway/translation/k8s_cache_translation.go
jm96441n: ah yeah, lemme change these so they're more indicative of what's being done

PR #2122, opened 2023-05-11
Net 3490/reference grants

missylbytes:
Changes proposed in this PR:
- Adds a validator for reference grants. The following scenarios are included
   - Gateway can reference a secret
   - HTTPRoute can reference a Gateway
   - HTTPRoute can reference a Backend (Services)  
   - TCPRoute can reference a Gateway
   - TCPRoute can reference a Backend (Services)  

How I've tested this PR:
Tests

Tests

Checklist:
- [x] Tests added

### DIFF
--- a/control-plane/api-gateway/controllers/reference_validator.go
+++ b/control-plane/api-gateway/controllers/reference_validator.go
@@ -25,49 +25,25 @@ func (rv ReferenceValidator) GatewayCanReferenceSecret(ctx context.Context, gate
 		Kind:  gateway.GroupVersionKind().Kind,
 	}
 
-	toName := string(secretRef.Name)
-	toNS := ""
-	if secretRef.Namespace != nil {
-		toNS = string(*secretRef.Namespace)
-	}
-
 	// Kind should default to Secret if not set
 	// https://github.com/kubernetes-sigs/gateway-api/blob/v0.6.2/apis/v1beta1/object_reference_types.go#LL59C21-L59C21
-	toGK := metav1.GroupKind{Kind: "Secret"}
-	if secretRef.Group != nil {
-		toGK.Group = string(*secretRef.Group)
-	}
-	if secretRef.Kind != nil {
-		toGK.Kind = string(*secretRef.Kind)
-	}
+	toNS, toGK := createValuesFromRef(secretRef.Namespace, secretRef.Group, secretRef.Kind, "Secret")
 
-	return referenceAllowed(ctx, fromGK, fromNS, toGK, toNS, toName, rv.Client)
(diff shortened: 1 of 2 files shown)


control-plane/api-gateway/controllers/reference_validator_test.go
```diff
@@ -0,0 +1,646 @@
+package controllers
+
```
missylbytes, line 2: There is probably some helper functions/DRYing to be done in these tests

control-plane/api-gateway/controllers/reference_validator.go
```diff
@@ -0,0 +1,174 @@
...
+	client.Client
+}
+
+func NewReferenceValidator(client client.Client) *ReferenceValidator {
+	return &ReferenceValidator{
+		client,
+	}
+}
+
+func (rv ReferenceValidator) GatewayCanReferenceSecret(ctx context.Context, gateway gwv1beta1.Gateway, secretRef gwv1beta1.SecretObjectReference) (bool, error) {
```
t-eckert, line 21: It will be slightly more efficient to make these pointer receivers to avoid copying the client on every function call.

control-plane/api-gateway/controllers/reference_validator.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (rv ReferenceValidator) HTTPRouteCanReferenceGateway(ctx context.Context, httproute gwv1beta1.HTTPRoute, parentRef gwv1beta1.ParentReference) (bool, error) {
+	fromNS := httproute.GetNamespace()
+	fromGK := metav1.GroupKind{
+		Group: httproute.GroupVersionKind().Group,
+		Kind:  httproute.GroupVersionKind().Kind,
+	}
+
+	// Kind should default to Gateway if not set
+	// https://github.com/kubernetes-sigs/gateway-api/blob/v0.6.2/apis/v1beta1/shared_types.go#L48
```
t-eckert, line 43: Love having these references to the spec in comments. #SheBroughtRecipts

t-eckert (approved):
I think this is great! Super clean and readable. Like we discussed earlier, there is likely a way to dry up the associated functions on the `ReferenceValidator`, but that is not high-priority/can be done after beta code freeze.

From looking at it, I think you could create an interface over `GetNamespace` and `GroupVersionKind` then have two associated functions. Both taking objects matching that interface as the "from" param and being differentiated by taking either `BackendRef` or `ParentRef` as the "to" param.

control-plane/api-gateway/controllers/reference_validator.go
missylbytes: I cannot take the credit here, it was done in consul-api-gateway. But still a good idea whoever did it at that point!

PR #2124, opened 2023-05-12
NET-2619 - save ClusterIPs to manual vips table

curtbushko:
Changes proposed in this PR:
1) Save the Kubernetes Service ClusterIP to Consuls virtual ip table when a service registered in endpoints controller
2) When a splitter or resolver CRD is created, get the corresponding service if possible and save the ClusterIP to Consuls virtual ip table.

3) Some edge cases:
  - Any errors that occur when doing the above are not fatal and the reconciler will continue 
  - If an older version of Consul is running, that does not have the VIP api endpoint, give an error message about upgrading and continue with the reconcile
  - If something is wrong with the service, such as it does not have a ClusterIP, give an error and continue on with the reconcile
  - If the CRD does not have a corresponding Kube Service, then carry on. When the service is registered in Consul the ClusterIP will be sent to Consul (#1 above)

How I've tested this PR:

- Unit tests & manual tests

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


control-plane/connect-inject/controllers/endpoints/endpoints_controller_test.go
```diff
@@ -6414,3 +6414,54 @@ func createGatewayPod(name, ip string, annotations map[string]string) *corev1.Po
 	}
 	return pod
 }
+
+func TestReconcileAssignServiceVirtualIP(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	cases := []struct {
```
thisisnotashwin, line 6421: Is there no way to check if `apiClient.Internal().AssignServiceVirtualIP(ctx, svc.Service, []string{IP}, &api.WriteOptions{})` executed correctly? as in, can we assert against the desired outcome?

control-plane/controllers/configentry_controller.go
```diff
@@ -381,6 +389,52 @@ func (r *ConfigEntryController) nonMatchingMigrationError(kubeEntry common.Confi
...
+		},
+	}
+
+	err := crdCtrl.Get(ctx, namespacedName, &service)
+	if err != nil {
+		// It is non-fatal if the service does not exist. The ClusterIP will get added when the service is registered in
+		// the endpoints controller
+		if strings.Contains(err.Error(), "not found") {
+			return nil
+		}
```
thisisnotashwin, line 418: I think we have syntactic sugar for similar errors in the `errors` package that we have used in the past:
`errors.IsNotFound(err)`

control-plane/controllers/configentry_controller.go
```diff
@@ -381,6 +389,52 @@ func (r *ConfigEntryController) nonMatchingMigrationError(kubeEntry common.Confi
...
+	}
+
+	service := corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      configEntry.KubernetesName(),
+			Namespace: namespacedName.Namespace,
+		},
+	}
+
+	err := crdCtrl.Get(ctx, namespacedName, &service)
```
thisisnotashwin, line 412: nit
```suggestion
	if err := crdCtrl.Get(ctx, namespacedName, &service); err != nil {
```
we try and use this pattern if the only param on the left is the error. We can do this for L424 as well

control-plane/controllers/configentry_controller_test.go
```diff
@@ -1891,3 +1892,211 @@ func TestConfigEntryController_Migration(t *testing.T) {
...
+			testClient.TestServer.WaitForLeader(t)
+			consulClient := testClient.APIClient
+
+			ctrl := c.reconciler(fakeClient, testClient.Cfg, testClient.Watcher, logrtest.TestLogger{T: t})
+			namespacedName := types.NamespacedName{
+				Namespace: kubeNS,
+				Name:      c.configEntryResource.KubernetesName(),
+			}
+
+			err := assignServiceVirtualIP(ctx, ctrl.Logger(namespacedName), consulClient, ctrl, namespacedName, c.configEntryResource)
```
thisisnotashwin, line 2094: similar to the other assertion, can we assert that the assignment worked as expected in Consul or is that not possible to test against?

control-plane/controllers/configentry_controller.go
```diff
@@ -381,6 +389,52 @@ func (r *ConfigEntryController) nonMatchingMigrationError(kubeEntry common.Confi
...
+// assignServiceVirtualIPs manually sends the ClusterIP for a matching service for ServiceRouter or ServiceSplitter
+// CRDs to Consul so that it can be added to the virtual IP table. The assignment is skipped if the matching service
+// does not exist or if an older version of Consul is being used. Endpoints Controller, on service registration, also
+// manually sends a ClusterIP when a service is created. This increases the chance of a real IP ending up in the
+// discovery chain.
+func assignServiceVirtualIP(ctx context.Context, logger logr.Logger, consulClient *capi.Client, crdCtrl Controller, namespacedName types.NamespacedName, configEntry common.ConfigEntryResource) error {
+
+	// Only assign the virtual IP for service routers and splitters. Every other type is skipped.
+	kubeKind := configEntry.KubeKind()
+	if kubeKind != common.ServiceRouter && kubeKind != common.ServiceSplitter {
```
thisisnotashwin, line 401: Does this not work for resolvers too?

control-plane/controllers/configentry_controller.go
```diff
@@ -267,6 +268,13 @@ func (r *ConfigEntryController) ReconcileEntry(ctx context.Context, crdCtrl Cont
 		return r.syncSuccessful(ctx, crdCtrl, configEntry)
 	}
 
+	// For resolvers and splitters, we need to set the ClusterIP of the matching service to Consul so that transparent
+	// proxy works correctly. Do not fail the reconcile if assigning the virtual IP returns an error.
+	err = assignServiceVirtualIP(ctx, logger, consulClient, crdCtrl, req.NamespacedName, configEntry)
```
thisisnotashwin, line 273: Can we check if the config entry is a splitter or a resolver before we assign the virtual IP? It makes it easier to understand why we invoke this method in that case.

thisisnotashwin (commented):
This looks good but i have a few comments!

control-plane/connect-inject/controllers/endpoints/endpoints_controller_test.go
curtbushko: I don't think so. We are operating blind here as there is no API endpoint and/or anything usable in the API/SDK. The VIP also ends up buried in the discovery chain...

control-plane/controllers/configentry_controller.go
curtbushko: Oof, you are correct!

control-plane/controllers/configentry_controller.go
curtbushko: Nice!  `k8serr.IsNotFound(err)`

control-plane/controllers/configentry_controller.go
curtbushko: Thank you. I knew there was a more idiomatic way but it was slipping my mind.

control-plane/controllers/configentry_controller_test.go
curtbushko: See above.

control-plane/controllers/configentry_controller.go
curtbushko: We could but having the check inside of assignServiceVirtualIP makes the tests a little bit nicer imho. Otherwise the only test I can do is 'does the service have a cluster IP'.

thisisnotashwin: I think that is ok? from a readability perspective, it makes sense to "assign service virtual IP" only when a particular config entry needs that assignment vs moving the config entry filtering inside the method. you could potentially create a method "needsVirtualIPAssignment" that returns true for certain config entries and false for other and test it's behavior therby covering test scenarios?

curtbushko: I like this idea!

curtbushko: @thisisnotashwin At your suggestion, I added a 'needs' method which does clean it up a little bit.

control-plane/controllers/configentry_controller_test.go
```diff
@@ -1891,3 +1892,266 @@ func TestConfigEntryController_Migration(t *testing.T) {
...
+					},
+				}
+			},
+			expectErr: false,
+		},
+		{
+			name:       "ServiceResolver no error and vip should be assigned",
+			kubeKind:   "ServiceResolver",
+			consulKind: capi.ServiceRouter,
+			consulPrereqs: []capi.ConfigEntry{
```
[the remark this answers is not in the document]
curtbushko: I removed the unneeded part. Thanks!

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -1258,6 +1265,25 @@ func (r *Controller) appendNodeMeta(registration *api.CatalogRegistration) {
 	}
 }
 
+// assignServiceVirtualIPs manually assigns the ClusterIP to the virtual IP table so that transparent proxy routing works.
+func assignServiceVirtualIP(ctx context.Context, apiClient *api.Client, svc *api.AgentService) error {
+	ip := svc.TaggedAddresses[clusterIPTaggedAddressName].Address
+	if ip == "" {
+		return nil
+	}
+	_, _, err := apiClient.Internal().AssignServiceVirtualIP(ctx, svc.Service, []string{ip}, &api.WriteOptions{})
```
hashi-derek, line 1274: Doesn't this need the namespace / partition info provided?

hashi-derek (approved):

PR #2134, opened 2023-05-15
Add telemetry collector deployment to consul-k8s

clly:
Changes proposed in this PR:
- Create a new component called the "consul-telemetry-collector" to act as an envoy MetricsServiceServer and receive mesh metrics. It will have a deployment, service account and service.
- Create a new component auth rule created by the server-acl-init job
- Include new values in values.yaml to manage the deployment 

How I've tested this PR:
* Bats tests
* Deploying into a `kind` cluster

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/test/unit/telemetry-collector-deployment.bats
+++ b/charts/consul/test/unit/telemetry-collector-deployment.bats
@@ -159,33 +159,6 @@ load _helpers
 }
 
 #--------------------------------------------------------------------
-# global.acls.manageSystemACLs
-
-@test "telemetryCollector/Deployment: consul login datacenter is set to primary when when federation enabled in non-primary datacenter" {
-  cd `chart_dir`
-  local object=$(helm template \
-      -s templates/telemetry-collector-deployment.yaml \
-      --set 'telemetryCollector.enabled=true' \
-      --set 'telemetryCollector.image=foo' \
-      --set 'meshGateway.enabled=true' \
-      --set 'global.acls.manageSystemACLs=true' \
-      --set 'global.datacenter=dc1' \
-      --set 'global.federation.enabled=true' \
-      --set 'global.federation.primaryDatacenter=dc2' \
-      --set 'global.tls.enabled=true' \
-      . | tee /dev/stderr |
-      yq '.spec.template.spec.initContainers[0]' | tee /dev/stderr)
-
-  local actual=$(echo $object |
-      yq '[.env[4].name] | any(contains("CONSUL_LOGIN_DATACENTER"))' | tee /dev/stderr)
-  [ "${actual}" = "true" ]
-
-  local actual=$(echo $object |
-      yq '[.env[4].value] | any(contains("dc2"))' | tee /dev/stderr)
-  [ "${actual}" = "true" ]
-}
-
-#--------------------------------------------------------------------
 # resources
 
 @test "telemetryCollector/Deployment: resources has default" {
@@ -590,7 +563,7 @@ load _helpers
       --set 'global.cloud.authUrl.secretName=auth-url-name' \
       .
   [ "$status" -eq 1 ]
-  
+
   [[ "$output" =~ "When either global.cloud.authUrl.secretName or global.cloud.authUrl.secretKey is defined, both must be set." ]]
 }
 
@@ -614,7 +587,7 @@ load _helpers
       --set 'global.cloud.authUrl.secretKey=auth-url-key' \
       .
   [ "$status" -eq 1 ]
-  
+
   [[ "$output" =~ "When either global.cloud.authUrl.secretName or global.cloud.authUrl.secretKey is defined, both must be set." ]]
 }
 
@@ -638,7 +611,7 @@ load _helpers
       --set 'global.cloud.apiHost.secretName=auth-url-name' \
       .
   [ "$status" -eq 1 ]
-  
+
   [[ "$output" =~ "When either global.cloud.apiHost.secretName or global.cloud.apiHost.secretKey is defined, both must be set." ]]
 }
 
@@ -662,7 +635,7 @@ load _helpers
       --set 'global.cloud.apiHost.secretKey=auth-url-key' \
       .
   [ "$status" -eq 1 ]
-  
+
   [[ "$output" =~ "When either global.cloud.apiHost.secretName or global.cloud.apiHost.secretKey is defined, both must be set." ]]
 }
 
@@ -686,7 +659,7 @@ load _helpers
       --set 'global.cloud.scadaAddress.secretName=scada-address-name' \
       .
   [ "$status" -eq 1 ]
-  
+
   [[ "$output" =~ "When either global.cloud.scadaAddress.secretName or global.cloud.scadaAddress.secretKey is defined, both must be set." ]]
 }
 
@@ -710,7 +683,7 @@ load _helpers
       --set 'global.cloud.scadaAddress.secretKey=scada-address-key' \
       .
   [ "$status" -eq 1 ]
-  
+
   [[ "$output" =~ "When either global.cloud.scadaAddress.secretName or global.cloud.scadaAddress.secretKey is defined, both must be set." ]]
 }
 


charts/consul/templates/telemetry-collector-deployment.yaml
```diff
@@ -0,0 +1,362 @@
...
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: consul-telemetry-collector
+        {{- if .Values.global.extraLabels }}
+          {{- toYaml .Values.global.extraLabels | nindent 8 }}
+        {{- end }}
+    spec:
+      serviceAccountName: {{ template "consul.fullname" . }}-telemetry-collector
+      initContainers:
+      - name: consul-connect-inject-init
```
curtbushko, line 74: I think I am missing something. Why are you doing this injection and not letting the webhook do it for you?

ishustava: @curtbushko This is because we don't want to create a dependency between this deployment and the webhook. When intalled in the same release, we cannot guarantee that the webhook will be running to the injection for this deployment. That's why we need to do manual injection

curtbushko: Sounds good! Thanks @ishustava

curtbushko:
> Thanks for the review @curtbushko. The only thing I was unsure about was whether or not I should include a pod security policy since I know that Kubernetes deprecated them at 1.21, and removed them completely in 1.25

@clly Good call!  I completely missed that.  Yes, you should probably add one as we are still supporting older versions of Kubernetes and we cannot remove the PSPs until we get passed 1.25. We will worry about that future work though and the removal won't fall on the Cloud team.

charts/consul/templates/_helpers.tpl
```diff
@@ -387,3 +387,39 @@ Usage: {{ template "consul.validateCloudSecretKeys" . }}
...
+- telemetryCollector.cloud.clientId.secretName
+- telemetryCollector.cloud.clientSecret.secretName
+- global.cloud.resourceId.secretName
+
+Usage: {{ template "consul.validateTelemetryCollectorCloud" . }}
+
+*/}}
+{{- define "consul.validateTelemetryCollectorCloud" -}}
+{{- if (and .Values.telemetryCollector.cloud.clientId.secretName (or (not .Values.global.cloud.resourceId.secretName) (not .Values.telemetryCollector.cloud.clientSecret.secretName))) }}
+{{fail "When telemetryCollector.cloud.clientId.secretName is set, global.cloud.resourceId.secretName, telemetryCollector.cloud.clientSecret.secretName must also be set."}}
```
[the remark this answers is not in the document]
curtbushko: I think you are good!

charts/consul/values.yaml
```diff
@@ -3069,3 +3074,86 @@ prometheus:
...
+
+telemetryCollector:
+  # Enables the consul-telemetry-collector deployment
+  # @type: boolean
+  enabled: false
+
+  # The name of the Docker image (including any tag) for the containers running
+  # the consul-telemetry-collector
+  # @type: string
+  image: "hashicorp/consul-telemetry-collector:0.0.1-dev"
```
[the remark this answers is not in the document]
clly: Since all of the scripts assume that the they're operating on _this_ git repository it sounds like it'll be simpler to include instructions in the release process to grab the latest docker tag.

david-yu:
@clly Does this need to go out with Consul 1.15.3? If so, this does need `backport/1.1.x`

charts/consul/templates/telemetry-collector-deployment.yaml
```diff
@@ -0,0 +1,362 @@
...
+          {{- if .Values.global.acls.manageSystemACLs }}
+          - -credential-type=login
+          - -login-bearer-token-path=/var/run/secrets/kubernetes.io/serviceaccount/token
+          {{- if and .Values.global.federation.enabled .Values.global.federation.primaryDatacenter }}
+          # fix this login method to use the component login method later
+          - -login-auth-method={{ template "consul.fullname" . }}-k8s-auth-method-{{ .Values.global.datacenter }}
+          - -login-datacenter={{ .Values.global.federation.primaryDatacenter }}
+          {{- else }}
+          - -login-auth-method={{ template "consul.fullname" . }}-k8s-auth-method
+          {{- end }}
```
ishustava, line 271: This should be the same auth method as in the init container.

charts/consul/templates/telemetry-collector-deployment.yaml
```diff
@@ -0,0 +1,362 @@
...
+          {{- if .Values.global.adminPartitions.enabled }}
+          - -service-partition={{ .Values.global.adminPartitions.name }}
+          {{- end }}
+          {{- if .Values.global.metrics.enabled }}
+          - -telemetry-prom-scrape-path=/metrics
+          {{- end }}
+          - -log-level={{ default .Values.global.logLevel }}
+          - -log-json={{ .Values.global.logJSON }}
+          - -envoy-concurrency=2
+          - -login-meta=pod=consul/
```
ishustava, line 286: I think it's already provided as env vars.

charts/consul/templates/telemetry-collector-deployment.yaml
```diff
@@ -0,0 +1,362 @@
...
+          readOnlyRootFilesystem: true
+          runAsGroup: 5995
+          runAsNonRoot: true
+          runAsUser: 5995
+        # dataplane volume mounts
+        volumeMounts:
+          - mountPath: /consul/connect-inject
+            name: consul-connect-inject-data
+            {{- if .Values.global.tls.enabled }}
+            {{- if not (or (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots) .Values.global.secretsBackend.vault.enabled) }}
```
ishustava, line 329: nit: this indentation is a bit off

ishustava (commented):
Great work so far. Leaving the comments I had about the implementation - will go back and look at tests once those are addressed

clly:
@david-yu ah, yes we want this to go out in 1.15.3. I'll replace the backport label

charts/consul/templates/telemetry-collector-deployment.yaml
```diff
@@ -0,0 +1,362 @@
...
+              fieldRef:
+                fieldPath: spec.nodeName
+          {{- if .Values.global.acls.manageSystemACLs }}
+            # The auth method should be k8s-component-auth-method but we would need to modify the acl-init job. We'll follow up with that
+          - name: CONSUL_LOGIN_AUTH_METHOD
+            {{- if and .Values.global.federation.enabled .Values.global.federation.primaryDatacenter }}
+            value: {{ template "consul.fullname" . }}-k8s-component-auth-method-{{ .Values.global.datacenter }}
+            {{- else }}
+            value: {{ template "consul.fullname" . }}-k8s-component-auth-method
+            {{- end }}
```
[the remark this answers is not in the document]
ishustava: Yeah if it needs perms beyond what service identity provides, it would need to use the component one. You'd need to add a role and binding rule for this to work though. So switching the method alone won't just work.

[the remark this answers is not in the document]
clly: There are changes to the `acl-init` job included in the PR to create the role and rolebinding with the anonymous rules (but not the anonymous policy ID).

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -663,6 +668,35 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
+	if c.flagTelemetryCollector {
+		// the telemetry collector doesn't currently need any special permissions and relies on the rest of the
+		// control plane to interact directly with Consul (e.g. service registration).
+		// We're creating our own RBAC here  so that we can login using the component auth method instead of the
+		// client application one
```
ishustava, line 675: Hmmm If it's on the mesh, it needs `service:write` which anonymous token doesn't have.

charts/consul/templates/server-acl-init-job.yaml
```diff
@@ -206,6 +206,10 @@ spec:
             -mesh-gateway=true \
             {{- end }}
 
+            {{- if .Values.telemetryCollector.enabled }}
+            -telemetry-collector=true \
```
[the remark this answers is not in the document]
clly: These changes got dropped from the PR

charts/consul/templates/telemetry-collector-deployment.yaml
clly: Fixed the indentation

control-plane/subcommand/server-acl-init/command.go
clly: We're switching back to the generic service auth-method

charts/consul/values.yaml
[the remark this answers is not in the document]
curtbushko: Sounds good!

curtbushko (approved):
Good from my perspective!

PR #2140, opened 2023-05-16
feat: update dataplane and consul images for 1.0.7 release

DanStough:
Changes proposed in this PR:
- Staging updates for the 1.0.7 release this week.
  - consul-dataplane 1.0.2 was released with security fixes
  - consul 1.14.7 is currently in the release pipeline

How I've tested this PR: CI

How I expect reviewers to test this PR: 👁️ 👁️ 👁️ 

Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/charts/consul/Chart.yaml
+++ b/charts/consul/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: consul
-version: 1.0.6
-appVersion: 1.14.5
+version: 1.0.7
+appVersion: 1.14.7
 kubeVersion: ">=1.21.0-0"
 description: Official HashiCorp Consul Chart
 home: https://www.consul.io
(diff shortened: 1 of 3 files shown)


david-yu (approved):

charts/consul/Chart.yaml
```diff
@@ -13,13 +13,13 @@ annotations:
...
+      image: hashicorp/consul:1.14.7
     - name: consul-k8s-control-plane
-      image: hashicorp/consul-k8s-control-plane:1.0.6
+      image: hashicorp/consul-k8s-control-plane:1.0.7
     - name: consul-dataplane
-      image: hashicorp/consul-dataplane:1.0.1
+      image: hashicorp/consul-dataplane:1.0.2
     - name: envoy
-      image: envoyproxy/envoy:v1.23.1
+      image: envoyproxy/envoy:v1.24.7
```
zalimeni, line 22: We have 1.25.x marked as supported in https://developer.hashicorp.com/consul/docs/k8s/compatibility#supported-consul-and-kubernetes-versions; should we be using that instead? (I see @david-yu already approved so this is probably more for my education)

david-yu: Yea we usually bump minor versions of Envoy for each minor version of consul-k8s and consul. This particular value is for the proxy image that is tied to Consul API Gateway, not sure why this wasn't bumped to 1.24.x earlier but that's probably the one we should use. cc @nathancoleman

david-yu: @DanStough It looks like this is still 1.23.1 here https://github.com/hashicorp/consul-k8s/blob/76a5e5b19e0d519a30ff0c50f4863d115e39d19b/charts/consul/values.yaml#L2915. So I'd look to @nathancoleman to see if that needs to be bumped as well. Good catch @zalimeni

DanStough: OK, I'll make the updates. Good catch all!

zalimeni (approved):

PR #2143, opened 2023-05-17
inject envoy_telemetry_bind_socket_dir proxy config when telemetry collector is enabled

nickethier:
Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


chapmanc (approved):
LGTM and tested with the acceptance tests.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -482,6 +487,10 @@ func (r *Controller) createServiceRegistrations(pod corev1.Pod, serviceEndpoints
 		proxyConfig.Config[envoyPrometheusBindAddr] = prometheusScrapeListener
 	}
 
+	if r.EnableTelemetryCollector {
+		proxyConfig.Config[envoyTelemetryCollectorBindSocketDir] = "/consul/connect-inject"
```
curtbushko, line 491: Not a blocker but I am thinking that maybe the directory should be `/consul/telemetry-collector` ?  Doesn't it create a <random string>.sock file?

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -761,6 +770,10 @@ func (r *Controller) createGatewayRegistrations(pod corev1.Pod, serviceEndpoints
 		}
 	}
 
+	if r.EnableTelemetryCollector {
+		service.Proxy.Config[envoyTelemetryCollectorBindSocketDir] = "/consul/service"
```
curtbushko, line 774: Similar to above. Maybe `/consul/telemetry-collector`?

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -119,6 +120,10 @@ type Controller struct {
 	// to Consul client agents.
 	EnableAutoEncrypt bool
 
+	// EnableTelemetryCollector controls whether the proxy service should be registered
+	// with config to enable telemetry forwarding.
+	EnableTelemetryCollector bool
```
curtbushko, line 125: Is this supposed to be toggleable on a per pod basis through annotations?  Most of our features allow you to turn a feature on globally through the helm chart or per pod using annotations. 

An example, is how we check if metrics is enabled on the annotation (https://github.com/hashicorp/consul-k8s/blob/main/control-plane/connect-inject/metrics/metrics_configuration.go#L72)

curtbushko (changes requested):
Looks pretty good to me!  

I have a few ideas/questions but nothing that I would block on

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
nickethier: It does If we want to create a dedicated emptyDir volume for it we can. I figured it was a bit heavy handed given we already have a dir for connect related coordination.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
nickethier: Thats an interesting idea. I guess if I was operating a cluster I wouldn't want consumers to be able to disable telemetry. We currently assume the feature is either enabled or disabled cluster wide.

curtbushko (approved):

PR #2158, opened 2023-05-19
Configure Gateway Controller with Helm values

t-eckert:
Changes proposed in this PR:
- Pass values to configure the behavior of API Gateways through the `connect-inject` subcommand and `GatewayClassConfig` CRD.
- Add the managed gateway class configuration.

This completes https://hashicorp.atlassian.net/browse/NET-3878 

How I've tested this PR:
- BATs and running an install in Kind

- Look for how the configuration gets into the right places for the controller to run

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


control-plane/subcommand/flags/gateway.go
```diff
@@ -0,0 +1,42 @@
...
+	// DeploymentMaxInstances is the maximum number of instances for the Gateway Deployment.
+	DeploymentMaxInstances int
+	// DeploymentMinInstances is the minimum number of instances for the Gateway Deployment.
+	DeploymentMinInstances int
+}
+
+func (f *GatewayFlags) Flags() *flag.FlagSet {
+	fs := flag.NewFlagSet("gateway", flag.ContinueOnError)
+
+	deploymentReplicas, _ := strconv.Atoi(os.Getenv(GatewayDeploymentReplicasEnvVar))
```
t-eckert, line 33: Is it clear enough here that the flags will fallback to the environment variables or should I add a comment?

control-plane/api-gateway/gatekeeper/deployment.go
```diff
@@ -69,13 +72,12 @@ func (g *Gatekeeper) deleteDeployment(ctx context.Context) error {
...
-			Namespace:   g.Gateway.Namespace,
-			Labels:      apigateway.LabelsForGateway(&g.Gateway),
-			Annotations: g.HelmConfig.CopyAnnotations,
+			Name:      g.Gateway.Name,
+			Namespace: g.Gateway.Namespace,
+			Labels:    apigateway.LabelsForGateway(&g.Gateway),
 		},
 		Spec: appsv1.DeploymentSpec{
-			Replicas: &g.HelmConfig.Replicas,
+			Replicas: g.GatewayClassConfig.Spec.DeploymentSpec.DefaultInstances,
```
andrewstucki, line 80: Just as a note, we'll want to use some of the pinning calculations so that we allow for this to range between min and max values passed in, and make it based off of the current number of replicas in the deployment. That doesn't need to go in this PR as I have some of that work in flight already, but just to describe the way it works, the logic is something like:

```go
if spec.MaxInstances != nil && currentReplicas > *spec.MaxInstances {
  replicas = *spec.MaxInstances
} else if spec.MinInstances != nil && currentReplicas < *spec.MinInstances {
  replicas = *spec.MinInstances
} else {
  replicas = firstNonNilOrZeroValueOf(currentReplicas, spec.DefaultInstances, 1)
}
```

control-plane/api-gateway/gatekeeper/deployment.go
```diff
@@ -69,13 +72,12 @@ func (g *Gatekeeper) deleteDeployment(ctx context.Context) error {
 func (g *Gatekeeper) deployment() *appsv1.Deployment {
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        g.Gateway.Name,
-			Namespace:   g.Gateway.Namespace,
-			Labels:      apigateway.LabelsForGateway(&g.Gateway),
-			Annotations: g.HelmConfig.CopyAnnotations,
```
andrewstucki, line 75: Can you just add a small TODO that we need to add this back in with the whitelisted values from the class config? We can handle it in a follow-up.

You're right that this shouldn't be coming from Helm, but I just want to make sure we don't accidentally drop it completely.

control-plane/api-gateway/gatekeeper/gatekeeper_test.go
```diff
@@ -863,6 +905,6 @@ func configureServiceAccount(name, namespace string, labels map[string]string, r
 	}
 }
 
-func ptrTo[T bool | string](t T) *T {
+func ptrTo[T bool | string | int32](t T) *T {
```
andrewstucki, line 908:
```suggestion
func ptrTo[T any](t T) *T {
```

andrewstucki (approved):
This seems fine to me once we move the CRD installation to happen via the CRDs directory like we mentioned.

Made some comments about changes we'll need, but I think those can pretty much all happen in a follow-up since there's still a good amount of in-flight changes that need to get merged in after this (i.e. the container sidecar definitions/arguments)

PR #2159, opened 2023-05-19
fix json tags for fields 1.0.x

thisisnotashwin:
Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


curtbushko (changes requested):
Blocking this one for a little bit while I fix some CI errors that should 🤞 make the tests go clear.

I will push through when complete.

curtbushko (approved):
Thanks for fixing the pipelines!

PR #2163, opened 2023-05-22
Run api gateway conformance tests in consul-k8s repository

sarahalsmiller:
Changes proposed in this PR:
- Run gateway-k8s conformance tests against consul-k8 native implementation of the gateway API as a github action by building a docker image of the control plane and loading it onto a kind cluster
- Tests currently run but fail due to changes that need to be made in the API Gateway code itself

How I've tested this PR:
- CI kicks off and runs as expected

- CI runs

Checklist:
- [ X ] Tests added
- [ ] CHANGELOG entry added 


andrewstucki:
Tagging @curtbushko to see if this should go here or in our workflows repo. In this case we're running OSS as the target Consul for the conformance tests, so wondering it we don't need the extra safety/configurability of the private runners and this can stay in-repo?

PR #2170, opened 2023-05-23
add helm chart values to configure global server side rate limiting

jmurret:
Changes proposed in this PR:
- add helm chart values for:
  - 
- add templating and bats tests to ensure that the the server-config-configmap.yml adds the configuration

How I've tested this PR:
- unit tests
- CI

👀 

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/2170.txt
+++ b/.changelog/2170.txt
@@ -1,0 +1,2 @@
+```release-note:feature
+Add support for configuring global level server rate limiting.
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -1131,6 +1131,41 @@ server:
     # @type: string
     caCert: null
 
+  # Settings for potentially limiting timeouts, rate limiting on clients as well
+  # as servers, and other settings to limit exposure too many requests, requests
+  # waiting for too long, and other runtime considerations.
+  limits:
+    # Settings for enabling or disabling request limiting across all requests
+    # on each Consul server.
+    requestLimits:   
+      # Setting for disabling or enabling rate limiting.  If not disabled, it
+      # enforces the action that will occur when RequestLimitsReadRate
+      # or RequestLimitsWriteRate is exceeded.  The default value of "disabled" will
+      # prevent any rate limiting from occuring.  A value of "enforce" will block
+      # the request from processings by returning an error.  A value of
+      # "permissive" will not block the request and will allow the request to 
+      # continue processing.
+      # @type: string
+      mode: "disabled"
+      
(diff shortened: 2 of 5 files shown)


charts/consul/values.yaml
```diff
@@ -1131,6 +1131,41 @@ server:
     # @type: string
     caCert: null
 
+  # Settings for potentially limiting timeouts, rate limiting on clients as well
```
curtbushko, line 1134: Nit. Could you add a link to the rate limits configuration?

https://developer.hashicorp.com/consul/docs/agent/config/config-files#request_limits

charts/consul/values.yaml
```diff
@@ -1131,6 +1131,41 @@ server:
...
+      mode: "disabled"
+      
+      # Setting that controls how frequently RPC, gRPC, and HTTP
+      # queries are allowed to happen. In any large enough time interval, rate
+      # limiter limits the rate to RequestLimitsReadRate tokens per second.
+      #
+      # See https://en.wikipedia.org/wiki/Token_bucket for more about token
+      # buckets.
+      # @type: integer
+      readRate: -1
```
curtbushko, line 1158: Why -1 instead of 100 which is the default?

curtbushko (commented):
Looks good.  Just a couple comments below but nothing I would block on.

charts/consul/values.yaml
jmurret: those are the defaults: https://github.com/hashicorp/consul/blob/main/agent/config/default.go#L106-L107

This documentation is incorrect.  I'll open a PR to correct it.

charts/consul/values.yaml
jmurret: Great call.  Added here: https://github.com/hashicorp/consul-k8s/pull/2170/commits/7ab9a14a946e9a6916f5aed01755555c55241101#diff-9d190488582b6dfdeb91fd565c91a267e5dc1c1b925a174174aade2a503f5373R1141

charts/consul/values.yaml
curtbushko: Awesome! thanks1

curtbushko (approved):
LGTM!

dhiaayachi (commented):
Thank you for putting this together so quickly @jmurret !

LGTM

.changelog/2170.txt
```diff
@@ -0,0 +1,2 @@
+```release-note:feature
```
david-yu, line 1: I think you forgot a closing ```

PR #2173, opened 2023-05-24
Net 4124/handle syncing consul lifecycle events

jm96441n:
Changes proposed in this PR:
- Add config entry deletion/upsert for changed consul resources
-

How I've tested this PR:
- Run tests
- Ran https://github.com/jm96441n/consul-experiments/blob/main/k8s/apigw4conk8s/deploy.sh and then modified and deleted the http route and checked that changes showed up in consul

Read code
Potentially run locally

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/api-gateway/controllers/gateway_controller.go
+++ b/control-plane/api-gateway/controllers/gateway_controller.go
@@ -402,12 +402,12 @@ func SetupGatewayControllerWithManager(ctx context.Context, mgr ctrl.Manager, co
 		).
 		Watches(
 			// Subscribe to changes from Consul for HTTPRoutes
-			&source.Channel{Source: c.Subscribe(ctx, api.APIGateway, translator.BuildConsulHTTPRouteTranslator(ctx)).Events()},
+			&source.Channel{Source: c.Subscribe(ctx, api.HTTPRoute, translator.BuildConsulHTTPRouteTranslator(ctx)).Events()},
 			&handler.EnqueueRequestForObject{},
 		).
 		Watches(
 			// Subscribe to changes from Consul for TCPRoutes
-			&source.Channel{Source: c.Subscribe(ctx, api.APIGateway, translator.BuildConsulTCPRouteTranslator(ctx)).Events()},
+			&source.Channel{Source: c.Subscribe(ctx, api.TCPRoute, translator.BuildConsulTCPRouteTranslator(ctx)).Events()},
 			&handler.EnqueueRequestForObject{},
 		).
 		Watches(
--- a/control-plane/api-gateway/controllers/gateway_controller.go
+++ b/control-plane/api-gateway/controllers/gateway_controller.go
@@ -64,29 +64,13 @@ func buildOpts(ref api.ConfigEntry) cmp.Option {
 	case *api.HTTPRouteConfigEntry:
 		return cmpopts.IgnoreFields(api.HTTPRouteConfigEntry{}, "Status", "ModifyIndex", "CreateIndex")
 	case *api.TCPRouteConfigEntry:
-		return cmpopts.IgnoreFields(api.TCPRouteConfigEntry{}, "Status", "ModifiyIndex", "CreateIndex")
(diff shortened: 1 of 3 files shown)


control-plane/cache/consul.go
```diff
@@ -584,7 +610,7 @@ func (c *Cache) ListServices() []api.CatalogService {
 	c.cacheMutex.Lock()
 	defer c.cacheMutex.Unlock()
 
-	entries := make([]api.CatalogService, len(c.serviceCache))
+	entries := make([]api.CatalogService, 0, len(c.serviceCache))
```
jm96441n, line 613: this and the above were causing issues because we later append to the slice which was leaving N null values in the slice at the beginning where N == len(c.serviceCache)

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -197,12 +231,22 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		// TODO: the actual delete
+		r.cache.Delete(deletion)
 	}
 
 	for _, update := range updates.Consul.Updates {
 		log.Info("updating in Consul", "kind", update.GetKind(), "namespace", update.GetNamespace(), "name", update.GetName())
-		// TODO: the actual update
+		ref := translation.EntryToReference(update)
+		old := r.cache.Get(ref)
+		if cmp.Equal(old, update, buildOpts(update)) {
```
jm96441n, line 241: I'm not super happy with cmp.Equal here, from reading the docs it can panic in weird ways, I'd prefer to write the comparison logic ourselves rather than rely on it

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -54,10 +57,41 @@ type GatewayController struct {
...
+// fmt.Println()
+// fmt.Println()
+// fmt.Println(empty)
+// fmt.Println()
+// fmt.Println()
+// fmt.Println()
+// fmt.Println()
+// fmt.Println()
+// return cmpopts.IgnoreFields(empty, "Status", "ModifiedIndex", "CreateIndex")
+// }
```
andrewstucki, line 89:
```suggestion
```

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -54,10 +57,41 @@ type GatewayController struct {
 	client.Client
 }
 
+func buildOpts(ref api.ConfigEntry) cmp.Option {
+	switch v := ref.(type) {
```
andrewstucki, line 61: I think all of these can technically just use a path-based comparison without doing the struct reflection that requires this switch, but I think it's fine to leave as-is for now.

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -54,10 +57,41 @@ type GatewayController struct {
...
+	case *api.APIGatewayConfigEntry:
+		return cmpopts.IgnoreFields(api.APIGatewayConfigEntry{}, "Status", "ModifyIndex", "CreateIndex")
+	case *api.HTTPRouteConfigEntry:
+		return cmpopts.IgnoreFields(api.HTTPRouteConfigEntry{}, "Status", "ModifyIndex", "CreateIndex")
+	case *api.TCPRouteConfigEntry:
+		return cmpopts.IgnoreFields(api.TCPRouteConfigEntry{}, "Status", "ModifiyIndex", "CreateIndex")
+	case *api.InlineCertificateConfigEntry:
+		return cmpopts.IgnoreFields(api.InlineCertificateConfigEntry{}, "Status", "ModifiyIndex", "CreateIndex")
```
andrewstucki, line 69:
```suggestion
		return cmpopts.IgnoreFields(api.InlineCertificateConfigEntry{}, "Status", "ModifyIndex", "CreateIndex")
```

andrewstucki (approved):
I think after correcting the field names, we can probably just merge this as-is with any additional changes coming in the subsequent end-to-end functionality PR that I'll try and open later today or tomorrow at latest.

PR #2169, opened 2023-05-23
Net 4125/hookup gatekeeper with snapshot

missylbytes:
Changes proposed in this PR:
Hooks up gatekeeper with new snapshot.

How I've tested this PR:
Manual testing, added tests.

How I expect reviewers to test this PR:
```
make control-plane-dev-docker
docker tag consul-k8s-control-plane-dev consul-k8s-control-plane-dev:local
kind load docker-image consul-k8s-control-plane-dev:local
kubectl create namespace consul
helm install consul ./charts/consul -n consul --set global.image=hashicorppreview/consul:1.16-dev --set global.imageK8S=consul-k8s-control-plane-dev:local
```
Make a gateway

Checklist:
- [ ] Tests added

PR #2175, opened 2023-05-24
[API Gateway] Hooking up API Gateways End-to-End

andrewstucki:
Changes proposed in this PR:
This is the minimal work needed to get a functional api gateway implementation working end-to-end. It builds off of a number of other PRs that are needed to be merged first including:

- https://github.com/hashicorp/consul-k8s/pull/2173
- https://github.com/hashicorp/consul-k8s/pull/2169

And fixes a number of issues in the process. I haven't fully rebased/merged in @t-eckert's helm configuration updates from https://github.com/hashicorp/consul-k8s/pull/2158 yet, and I imagine there will be some merge conflicts due to re-generating the CRD manifests and stuff, so I'll try and clear that up when those other two PRs are merged.

How I've tested this PR:

- Added and updated unit tests
- Added functionality to our main basic acceptance test
- Verified some stuff manually

If you want to see this working end-to-end, build and push a custom dev container to a kind environment and then change into the `acceptance/tests/api-gateway` directory and run `go test -consul-image=hashicorppreview/consul:1.16-dev -consul-k8s-image=consul-k8s-control-plane-dev:gauntlet`. The tests should pass. 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


acceptance/tests/fixtures/bases/api-gateway/apigateway.yaml
```diff
@@ -9,8 +9,8 @@ spec:
   gatewayClassName: gateway-class
   listeners:
   - protocol: HTTP
-    port: 80
+    port: 8080
```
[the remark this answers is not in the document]
t-eckert: Noted, I created a bug in our backlog for this.

charts/consul/crds/gatewayclasses.gateway.networking.k8s.io.yaml.yml
```diff
@@ -0,0 +1,323 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+apiVersion: apiextensions.k8s.io/v1
```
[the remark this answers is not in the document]
t-eckert: We need them checked into the repo as there is no way to programmatically add CRDs from an external source at installation time. They already existed in the `api-gateways` branch so I don't know why they are "new files" here. Maybe something to do with the generation process.

control-plane/api-gateway/gatekeeper/dataplane.go
```diff
@@ -0,0 +1,172 @@
...
+			RunAsGroup:             pointer.Int64(sidecarUserAndGroupID),
+			RunAsNonRoot:           pointer.Bool(true),
+			ReadOnlyRootFilesystem: pointer.Bool(true),
+		}
+	}
+
+	return container, nil
+}
+
+func getDataplaneArgs(namespace string, config apigateway.HelmConfig, bearerTokenFile string, name string) ([]string, error) {
```
t-eckert, line 116: This is definitely a fine approach. However, we can also set these values as environment variables in the container and that may lead to a more maintainable method of setting these. 

That can be a post-feature-freeze work item if we want to make the change.

t-eckert (approved):
This looks fantastic! Thank you for bringing everything together this way.

control-plane/api-gateway/binding/binder.go
```diff
@@ -219,9 +231,38 @@ func (b *Binder) Snapshot() Snapshot {
...
+
+		registrations := registrationsForPods(entry.Namespace, b.config.Gateway, registrationPods)
+		snapshot.Consul.Registrations = registrations
+
+		// deregister any not explicitly registered service
+		for _, service := range b.config.GatewayServices {
+			found := false
+			for _, registration := range registrations {
+				if service.ServiceID == registration.Service.ID {
+					found = true
```
jm96441n, line 250: should we break out of this loop here once we find the service?

control-plane/api-gateway/binding/result.go
```diff
@@ -373,14 +373,42 @@ var (
...
+}
+
+// programmedCondition returns a condition for the Programmed status type.
+func (l gatewayValidationResult) programmedCondition(generation int64) metav1.Condition {
+	now := metav1.Now()
+
+	switch l.programmedErr {
+	case errGatewayPending_Pods, errGatewayPending_Consul:
+		return metav1.Condition{
+			Type:               "Programmed",
```
jm96441n, line 395: do we have this type/reason in either the k8s deps or the consul api dep?

control-plane/api-gateway/binding/route_binding.go
```diff
@@ -419,7 +427,7 @@ func parentsForRoute(ref api.ResourceReference, existing []api.ResourceReference
 	parentSet := map[string]struct{}{}
 	for _, result := range results {
 		for _, r := range result.results {
-			if r.err != nil {
```
jm96441n, line 422: was this a previous bug?

control-plane/api-gateway/gatekeeper/dataplane.go
andrewstucki: 👍 I'm fine with either the args or env variable approach, just did this since I could base it off existing code and didn't have to put as much thought behind how to implement it at first 😅

control-plane/api-gateway/gatekeeper/gatekeeper.go
```diff
@@ -83,9 +83,9 @@ func (g Gatekeeper) namespacedName(gateway gwv1beta1.Gateway) types.NamespacedNa
 	}
 }
 
-func (g Gatekeeper) serviceAccountName() string {
-	authspecaccount := "" // TODO do I need to add this to GatewayClassConfig?
-	fmt.Println(authspecaccount)
-
-	return ""
+func (g Gatekeeper) serviceAccountName(gateway gwv1beta1.Gateway, config apigateway.HelmConfig) string {
```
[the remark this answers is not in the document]
andrewstucki: No prob, the reason for this is that our default auth method uses the kubernetes JWT token to associate a service with a [Consul service identity](https://developer.hashicorp.com/consul/docs/security/acl/acl-roles#service-identities). When dataplane then auths it gets a token associated with that service identity which gives it read/write on `"<service name>"` and `"<service name>-sidecar-proxy"` -- all of this requires aligning the service account in k8s with the name of the gateway service that we register in Consul.

control-plane/api-gateway/binding/binder.go
andrewstucki: Yeah, we could do the early break, just forgot to add it. Since this is so big though, think we can probably just add it in later as we identify bug fixes.

control-plane/api-gateway/binding/result.go
andrewstucki: Ah, you're right, I didn't realize that the spec now actually [defined these constants](https://github.com/kubernetes-sigs/gateway-api/blob/2dfd20ecb1d829183bb7f5137f7f991feebeb779/apis/v1beta1/gateway_types.go#L549), we should use them. You fine waiting for a refactor pass or need that to go in pre-merge?

andrewstucki: Fun enough though, due to the typed constants, using any of those values defined in the spec requires casting back to a string, just like we have to do the weird casting elsewhere while using the spec 😞

Issue #2144, opened 2023-05-17
consul-dataplane fails to bind to port 8600 in multiport connect-inject service mesh using local kind cluster


fredheidrich-aurizon:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

While following the multiport tutorial in Consul Service Mesh (https://developer.hashicorp.com/consul/docs/k8s/connect#kubernetes-pods-with-multiple-ports) on a local kind cluster the deployment fails due to consul-dataplane inject of the first pod "web" is already bound to port 8600.

### Reproduction Steps

1. deploy consul via helm with `values.yaml`:
```
global:
  domain: consul
  datacenter: dc1
server:
  replicas: 1
  bootstrapExpect: 1
connectInject:
  enabled: true
controller:
  enabled: true
```
2. deploy following `manifest.yaml` with `kubectl apply -f manifest.yaml`: 
```
---
# Two service accounts are needed
apiVersion: v1
kind: ServiceAccount
metadata:
  name: web
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: web-admin
---
# Two service objects are needed, these will be the name of the consul services
apiVersion: v1
kind: Service
metadata:
  name: web
spec:
  selector:
    app: web
  ports:
    - protocol: TCP
      port: 80
      targetPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: web-admin
spec:
  selector:
    app: web
  ports:
    - protocol: TCP
      port: 80
      targetPort: 9090
---
# Two secrets are needed that references the service accounts
# The secret must have the same name as the ServiceAccount
apiVersion: v1
kind: Secret
metadata:
  name: web
  annotations:
    kubernetes.io/service-account.name: web
type: kubernetes.io/service-account-token
---
apiVersion: v1
kind: Secret
metadata:
  name: web-admin
  annotations:
    kubernetes.io/service-account.name: web-admin
type: kubernetes.io/service-account-token
---
[...]

PR #2176, opened 2023-05-25
Disable DNS redirection when tproxy is disabled

lkysow:
Changes proposed in this PR:
- DNS redirection and the various settings that make that possible (like
the dataplane binding to a port for DNS) is only useful if tproxy is
enabled. Most of the code checked if tproxy was enabled but there
was one location where we didn't check. This resulted in a bug
with our multiport support where even though tproxy is disabled,
we tried to setup the dataplane to proxy DNS. This meant each dataplane
tried to bind to 8600 but because there are >1 dataplanes with
multiport, there was a port conflict.

  This PR fixes the location where we didn't check if tproxy was enabled
and as a result fixes the multiport issue.
- the reason this wasn't caught in our acceptance tests is because we don't run the multiport test when global tproxy is enabled and this bug only occurs when global tproxy is enabled but pod-level tproxy is disabled
- Fixes https://github.com/hashicorp/consul-k8s/issues/2144

How I've tested this PR:
- built out a dataplane and manually deployed it with multiport: `go test ./... -p 1 -timeout 20m -run TestConnectInject_MultiportServices -no-cleanup-on-failure -v -consul-image hashicorppreview/consul:1.16-dev -enable-transparent-proxy -consul-k8s-image ghcr.io/lkysow/consul-k8s-dev:mar25-23@sha256:fba790b616687308a2b1159da087eb52109d8acda239b1b69a1775bf12db9570`

- code

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
+++ b/control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
@@ -314,7 +314,11 @@ func (w *MeshWebhook) getContainerSidecarArgs(namespace corev1.Namespace, mpi mu
 
 	// If Consul DNS is enabled, we want to configure consul-dataplane to be the DNS proxy
 	// for Consul DNS in the pod.
-	if w.EnableConsulDNS {
+	dnsEnabled, err := consulDNSEnabled(namespace, pod, w.EnableConsulDNS, w.EnableTransparentProxy)
+	if err != nil {
+		return nil, err
+	}
+	if dnsEnabled {
 		args = append(args, "-consul-dns-bind-port="+strconv.Itoa(consulDataplaneDNSBindPort))
 	}
 
--- a/control-plane/connect-inject/webhook/container_init.go
+++ b/control-plane/connect-inject/webhook/container_init.go
@@ -267,7 +267,17 @@ func (w *MeshWebhook) containerInit(namespace corev1.Namespace, pod corev1.Pod, 
 // consulDNSEnabled returns true if Consul DNS should be enabled for this pod.
 // It returns an error when the annotation value cannot be parsed by strconv.ParseBool or if we are unable
 // to read the pod's namespace label when it exists.
-func consulDNSEnabled(namespace corev1.Namespace, pod corev1.Pod, globalEnabled bool) (bool, error) {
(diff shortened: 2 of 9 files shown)


control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -314,7 +314,11 @@ func (w *MeshWebhook) getContainerSidecarArgs(namespace corev1.Namespace, mpi mu
 
 	// If Consul DNS is enabled, we want to configure consul-dataplane to be the DNS proxy
 	// for Consul DNS in the pod.
-	if w.EnableConsulDNS {
+	dnsEnabled, err := consulDNSEnabled(namespace, pod, w.EnableConsulDNS, w.EnableTransparentProxy)
```
lkysow, line 317: Bugfix is here. Use `consulDNSEnabled` instead of just checking global DNS setting. @ishustava can you confirm this change is okay. Previously we only looked at the global setting but I see there is also pod and namespace settings so I made the change to look at those too (in addition to checking if tproxy is enabled)

control-plane/connect-inject/webhook/container_init.go
```diff
@@ -267,7 +267,17 @@ func (w *MeshWebhook) containerInit(namespace corev1.Namespace, pod corev1.Pod,
 // consulDNSEnabled returns true if Consul DNS should be enabled for this pod.
 // It returns an error when the annotation value cannot be parsed by strconv.ParseBool or if we are unable
 // to read the pod's namespace label when it exists.
-func consulDNSEnabled(namespace corev1.Namespace, pod corev1.Pod, globalEnabled bool) (bool, error) {
+func consulDNSEnabled(namespace corev1.Namespace, pod corev1.Pod, globalDNSEnabled bool, globalTProxyEnabled bool) (bool, error) {
+	// DNS is only possible when tproxy is also enabled because it relies
```
lkysow, line 271: Change to also check tproxy.

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
ishustava: Perfect, this is exactly what I was hoping we'd do.

ishustava (approved):
🎉

PR #2181, opened 2023-05-25
Fix tests

lkysow:
Fix deepcopy bug in tests (testNS.Labels is shared even when the struct is copied to ns)


curtbushko (approved):
🎉  Thanks!

PR #2183, opened 2023-05-25
Fix CVEs by updating controller-runtime

t-eckert:
Fixes https://hashicorp.atlassian.net/browse/NET-704

Changes proposed in this PR:
- Bump version of controller runtime
- Update all breaking changes for this fix

How I've tested this PR:
- Built and run unit tests

Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 


david-yu:
Thank you!

curtbushko:
@t-eckert It looks like there are some conflicts. We so many changes going on as of late, I bet a few files changed under you.

curtbushko (changes requested):
Thanks for this Thomas!  

This looks like a tedious amount of work so thank you for doing it.

I am not seeing anything wrong with it other than you checked in the control-plane binary 😉 

I suspect that the backports will fail miserably because of the go.mod/go.sum changes...

david-yu:
Perhaps this should target 1.1.x instead of main? Also we should backport to 1.0.x if you can.

curtbushko (approved):

PR #2184, opened 2023-05-25
NET-3908: allow configuration of SecurityContextConstraints when running on OpenShift

nathancoleman:
When deploying api-gateway into OpenShift, a `SecurityContextConstraints` (SCC) must exist, and the `Role` must have the ability to `use` that SCC.

Rather than deploy a duplicative SCC for each api-gateway, the Helm chart will default to using an SCC that already exists: `restricted-v2` and use it for each gateway made with the managed gateway class. The user can replace this value in the ManagedGatewayClass stanza of the helm chart.
The user can also use their own SCC if they desire by creating one and adding it to the gatewayClass that they create (or the managedGatewayClass in our chart).
 
```
# API Gateway ManagedGatewayClass stanza in the helm chart
   apiGateway:
    # Configuration settings for the GatewayClass installed by Consul on Kubernetes.
    managedGatewayClass:
      -------------
      # The name of the Openshift SecurityContextConstraint to use for Gateways
      # @type: string
      openshiftSccName: "restricted-v2"
```

**How I've tested this PR:**  
- Get access to consul-dev azure account.
- Follow this gist: https://gist.github.com/david-yu/9a636d909bc45efe072968c37dc8c615
When you get to the helm install step, use the local chart in this PR as well as the image from it: https://hub.docker.com/layers/hashicorppreview/consul-k8s-control-plane/1.3.0-dev-pr-23f0bf1a08cff2c49662332a6a4d5447fa57959b/images/sha256-e6347378502ec3c63f230d8afaa5872df17018258f47cecd75580d77a66eff1c?context=explore
^^ You will need to update the sha linked here if there have been new commits to the PR
- Apply an api-gateway and a backend for it.  

- Possibly repeat the steps above.
- Check out those tests!

**Checklist:**
- [x] Tests added
- [x] CHANGELOG entry added 


david-yu:
Just wanted to say that we will likely need to also test this for OpenShift 4.12 and likely change some of the SCC config to SCC v2 when we get to OpenShift 4.12 support. cc @eikenb

control-plane/api-gateway/gatekeeper/role.go
```diff
@@ -83,3 +82,62 @@ func (g *Gatekeeper) role(gateway gwv1beta1.Gateway, gcc v1alpha1.GatewayClassCo
 
 	return role
 }
+
+func (g *Gatekeeper) upsertOpenshiftRole(ctx context.Context, gateway gwv1beta1.Gateway, config common.HelmConfig) error {
```
missylbytes, line 86: We basically create a separate role and rolebinding for openshift. I think I can refactor these to just be in the "normal" role and role-binding, but didn't know if we wanted them to be separate?

nathancoleman: No need to have a separate role for openshift, we can just create and modify a single role

charts/consul/templates/api-gateway-securitycontextconstraints.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - SETUID
+  - SETGID
+runAsUser:
+  type: MustRunAsRange
+seLinuxContext:
+  type: MustRunAs
+supplementalGroups:
+  type: MustRunAs
+users: []
+volumes: []
```
nathancoleman, line 42:
```suggestion
allowHostDirVolumePlugin: false
allowHostIPC: false
allowHostNetwork: {{ .Values.client.hostNetwork }}
allowHostPID: false
allowHostPorts: true
allowPrivilegeEscalation: false
allowPrivilegedContainer: true
allowedCapabilities:
  - NET_BIND_SERVICE
defaultAddCapabilities: null
fsGroup:
  type: RunAsAny
readOnlyRootFilesystem: true
requiredDropCapabilities:
  - ALL
runAsUser:
  type: RunAsAny
seLinuxContext:
  type: RunAsAny
supplementalGroups:
  type: RunAsAny
```

control-plane/api-gateway/gatekeeper/role.go
missylbytes: removed

charts/consul/templates/api-gateway-securitycontextconstraints.yaml
```diff
@@ -0,0 +1,33 @@
+{{- if (and .Values.global.enabled .Values.global.openshift.enabled) }}
```
missylbytes, line 1: - [ ] Need to add bats test for this

charts/consul/templates/api-gateway-securitycontextconstraints.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: SecurityContextConstraints
+metadata:
+  name: {{ template "consul.fullname" . }}-api-gateway
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: api-gateway-controller  # TODO
```
missylbytes, line 12: @nathancoleman what was this TODO here?

charts/consul/templates/connect-inject-deployment.yaml
```diff
@@ -161,6 +161,11 @@ spec:
                 {{- if .Values.global.openshift.enabled }}
                 -enable-openshift \
                 {{- end }}
+                {{- if ne .Values.connectInject.apiGateway.openshiftSccName ""}}
+                -openshift-scc-name={{ .Values.connectInject.apiGateway.openshiftSccName }} \
+                {{- else }}
```
missylbytes, line 166: - [ ] this if else no longer needed, just make it the name of the scc

missylbytes:
- [ ] Need to rebase with main again

missylbytes:
- [ ] need to move config value for scc under the gatewayclassconfig

charts/consul/templates/crd-gatewayclassconfigs.yaml
```diff
@@ -138,6 +138,10 @@ spec:
                       type: string
                   type: object
                 type: array
+              openshiftSCCName:
+                description: The name of an existing Openshift SCC
```
missylbytes, line 142: This could be better description, did it quickly for testing

charts/consul/test/unit/connect-inject-clusterrole.bats
```diff
@@ -217,3 +217,31 @@ load _helpers
...
+  local actual=$(echo $object | yq -r '.resourceNames[| index("restricted-v2")' | tee /dev/stderr)
+  [ "${actual}" != null ]
+}
+
+@test "connectInject/ClusterRole: adds permission to securitycontextconstraints for Openshift with global.openshift.enabled=true and sets apiGateway Openshift SCC Name" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -s templates/connect-inject-clusterrole.yaml  \
+      --set 'global.openshift.enabled=true' \
+      --set '.Values.connectInject.apiGateway.openshiftSccName="fakescc"' \
```
missylbytes, line 241: - [ ] Need to update bats

missylbytes: These also aren't working at all, need to understand bats

missylbytes: fixed

charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -186,4 +186,14 @@ rules:
     - "get"
     - "list"
     - "watch"
+{{- if .Values.global.openshift.enabled }}
+- apiGroups:
+    - security.openshift.io
+  resources:
+    - securitycontextconstraints
+  resourceNames:
+    - {{ .Values.connectInject.apiGateway.managedGatewayClass.openshiftSccName }}
```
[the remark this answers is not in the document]
missylbytes: oh yeah I would need to also update it in the chart. I didn't think about abbreviations in helm

control-plane/api-gateway/gatekeeper/init.go
```diff
@@ -168,14 +168,17 @@ func initContainer(config common.HelmConfig, name, namespace string) (corev1.Con
...
+		container.SecurityContext = &corev1.SecurityContext{
+			RunAsUser:    pointer.Int64(initContainersUserAndGroupID),
+			RunAsGroup:   pointer.Int64(initContainersUserAndGroupID),
+			RunAsNonRoot: pointer.Bool(true),
+			Privileged:   pointer.Bool(false),
+			Capabilities: &corev1.Capabilities{
+				Drop: []corev1.Capability{"ALL"},
+			},
+		}
 	}
```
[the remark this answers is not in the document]
missylbytes: Should I instead gate it behind the type of SCC? I think if the user specifies another type it might cause issues.

charts/consul/templates/connect-inject-clusterrole.yaml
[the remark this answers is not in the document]
nathancoleman: I think it makes sense to make this change. It would be consistent with `manageSystemACLs` which already exists today

control-plane/api-gateway/gatekeeper/init.go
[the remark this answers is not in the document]
nathancoleman: Good point

charts/consul/templates/gateway-resources-job.yaml
```diff
@@ -99,6 +99,7 @@ spec:
             - {{- toYaml .Values.connectInject.apiGateway.managedGatewayClass.copyAnnotations.service.annotations | nindent 14 -}}
             {{- end }}
             - -service-type={{ .Values.connectInject.apiGateway.managedGatewayClass.serviceType }}
+            - -openshift-scc-name={{ .Values.connectInject.apiGateway.managedGatewayClass.openshiftSCCName }}
```
nathancoleman, line 102: I think it makes sense to gate this to only if `global.openshift.enabled = true`

charts/consul/values.yaml
```diff
@@ -2201,6 +2201,10 @@ connectInject:
         maxInstances: 1
         minInstances: 1
 
+      # The name of the Openshift SecurityContextConstraints resource to use for Gateways
+      # @type: string
+      openshiftSCCName: "restricted-v2"
```
nathancoleman, line 2206: I think it makes sense to include here that this value is only used if `global.openshift.enabled = true`

missylbytes:
Closing, looks like test failures are similar to this issue: https://github.com/hashicorp/consul-k8s/pull/2722
This PR does not affect partitions, CNI, Mesh-gateways, so is likely unrelated.

david-yu:
Out of curiosity would this be backported to 1.16.x or are we just targeting 1.17.x here?

nathancoleman:
This is intended to go out in consul-k8s 1.2.1, so I've added the `backport/1.2.x` label

PR #2222, opened 2023-05-31
Persist virtual-ips for intentions / service-defaults.

hashi-derek:
This PR expands upon some of the tproxy enhancements to better support sameness-groups. We now store the virtual IP if a service-defaults or intention targeting a specific destination service exists.


control-plane/controllers/configentry_controller.go
```diff
@@ -422,13 +439,14 @@ func assignServiceVirtualIP(ctx context.Context, logger logr.Logger, consulClien
 		return err
 	}
 
+	consulType := configEntry.ToConsul(datacenter)
```
curtbushko, line 442: I like this change. It was a spot that just didn't feel right to me when I did it...

PR #2223, opened 2023-05-31
Fix telemetry collector issue and fix for bat test

chapmanc:
Changes proposed in this PR:
- Check for nil on proxy config. It appears in Mesh Gateways , Terminating Gateways and Ingress Gateways if we do not enable metrics then we will not have a service proxy enabled

How I've tested this PR:
 - unit tested

Checklist:
- [X] Tests added
- [] CHANGELOG entry added 


curtbushko (approved):
LGTM!

(After a long debugging session we know that these changes are not causing problems)

PR #2227, opened 2023-05-31
Add the endpoint ignoring logic for triggering gateway reconciliation

andrewstucki:
Changes proposed in this PR:
This adds the same ignoring logic on endpoints that exists within our Kubernetes service --> Consul service mapping code so that we don't trigger excessive numbers of reconciliations when we'd just subsequently ignore the said service.

How I've tested this PR:
Added unit tests for it.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -363,10 +363,6 @@ func SetupGatewayControllerWithManager(ctx context.Context, mgr ctrl.Manager, co
 			source.NewKindWithCache(&corev1.Secret{}, mgr.GetCache()),
 			handler.EnqueueRequestsFromMapFunc(r.transformSecret(ctx)),
 		).
-		Watches(
-			source.NewKindWithCache(&gwv1beta1.ReferenceGrant{}, mgr.GetCache()),
-			handler.EnqueueRequestsFromMapFunc(r.transformReferenceGrant(ctx)),
-		).
```
andrewstucki, line 369: this was duplicated a few lines above, so just dropped it while I was adding this.

t-eckert (approved):

sarahalsmiller (approved):

PR #2232, opened 2023-06-01
Get consul-dataplane image from helm chart

curtbushko:
Changes proposed in this PR:
- This is similar to a make target I added last week for the consul-image
- We need this so that we can load the consul-dataplane image onto the Kind clusters and prevent dockerhub from throttling us.
- This will indirectly fix a bunch of Kind acceptance tests when I add the corresponding change to the workflows.

How I've tested this PR:

Ran 'make consul-dataplane-version'

:eyes:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


Makefile
```diff
@@ -1,5 +1,6 @@
 VERSION = $(shell ./control-plane/build-support/scripts/version.sh control-plane/version/version.go)
 CONSUL_IMAGE_VERSION = $(shell ./control-plane/build-support/scripts/consul-version.sh charts/consul/values.yaml)
+CONSUL_DATAPLANE_IMAGE_VERSION = $(shell ./control-plane/build-support/scripts/consul-dataplane-version.sh charts/consul/values.yaml)
```
DanStough, line 3: Just a thought, but you could have the script take the YAML path or global key and cut down on the number of files.

DanStough (approved):

PR #2237, opened 2023-06-01
Add acceptance test cleanup for API Gateway resources

andrewstucki:
Changes proposed in this PR:
This cleans up any orphaned resources created by our Helm chart for API Gateways in the case of a controller crash causing the in-chart cleanup logic to not work (the in-chart cleanup requires the controller be running in order to remove finalizers from the objects we manage).

Previously when a test was failing and causing the controller to crash, the actual error would be masked by the fact that any subsequent test would fail due to some managed resources getting orphaned due to finalizers being on them and failing to get deleted due to the controller being unable to run during the helm chart uninstallation. This forcibly cleans the resources we created so that we don't mask why tests are failing.

How I've tested this PR:
Commented out any cleanup logic so that we have referenced gateways and then ran the tests with and without this cleanup routine, after a successful pass, any subsequent test run fails with something like:

```
        	Error:      	Received unexpected error:
        	            	error while running command: exit status 1; Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: GatewayClass "consul-api-gateway" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "test-rgxxva": current value is "test-bnh1v4"
        	Test:       	TestPartitions_Connect/default_destination_namespace
```

With this code, no such errors occur.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


curtbushko (approved):
This is fantastic! 

If you want to be the hero I need, you could do another PR for a bunch of other resources ones that are left behind!

PR #2246, opened 2023-06-01
NET-4285 add check for pointer

chapmanc:
Changes proposed in this PR:
- Pointer wasn't checked in translation. Now it is :) 
-

How I've tested this PR:
With the unit of tests.
👀 

Checklist:
- [X] Tests added

### DIFF
--- a/control-plane/api/v1alpha1/servicedefaults_types.go
+++ b/control-plane/api/v1alpha1/servicedefaults_types.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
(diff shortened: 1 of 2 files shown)


malizz (commented):
Thanks for fixing this bug. 
You could use one variable `dur` of type duration to avoid defining two variables.

curtbushko (approved):

curtbushko:
> 


@malizz We were on a zoom when we fixed this and decided on 2 durations just so that we were guaranteed that the times would be different.

curtbushko:
I also reproduced this locally and tested the changes

PR #2248, opened 2023-06-02
API Gateway lifecycle acceptance tests

andrewstucki:
Changes proposed in this PR:
This adds an acceptance test for API gateways that checks the majority of the lifecycle management around cleaning up Kubernetes and Consul resources based on Gateway, Route, and Secret state.

How I've tested this PR:
Wrote an acceptance test and got it to pass on both OSS and Enterprise builds.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


control-plane/api-gateway/binding/binder_test.go
```diff
@@ -54,15 +54,16 @@ var (
...
-	consulHTTPRoutes []api.HTTPRouteConfigEntry
-	consulTCPRoutes  []api.TCPRouteConfigEntry
+	grants                   []gwv1beta1.ReferenceGrant
+	secrets                  []corev1.Secret
+	gateways                 []gwv1beta1.Gateway
+	httpRoutes               []gwv1beta1.HTTPRoute
+	tcpRoutes                []gwv1alpha2.TCPRoute
+	meshServices             []v1alpha1.MeshService
+	services                 []types.NamespacedName
+	consulInlineCertificates []api.InlineCertificateConfigEntry
```
andrewstucki, line 64: this was added since we were forgetting to reference count the inline certificates previously, which was leaving them orphaned in Consul when they needed to be cleaned up.

control-plane/api-gateway/binding/route_binding.go
```diff
@@ -20,6 +20,11 @@ func (r *Binder) bindRoute(route client.Object, boundCount map[gwv1beta1.Section
 	// on non-enterprise installations
 	routeConsulKey := r.config.Translator.NonNormalizedConfigEntryReference(entryKind(route), client.ObjectKeyFromObject(route))
 	filteredParents := filterParentRefs(r.key, route.GetNamespace(), getRouteParents(route))
+	filteredParentStatuses := filterParentRefs(r.key, route.GetNamespace(),
+		common.ConvertSliceFunc(getRouteParentsStatus(route), func(parentStatus gwv1beta1.RouteParentStatus) gwv1beta1.ParentReference {
+			return parentStatus.ParentRef
+		}),
+	)
```
andrewstucki, line 27: this was added due to the fact that we now include in the routes we pass in to be reconciled with the gateway the routes that just reference the gateway in their statuses (in order to purge them if necessary)

control-plane/api-gateway/binding/route_binding.go
```diff
@@ -80,6 +86,14 @@ func (r *Binder) bindRoute(route client.Object, boundCount map[gwv1beta1.Section
...
 		}
 	}
+	// if we're orphaned from this gateway we'll
+	// always need a status update.
+	if len(filteredParents) == 0 {
+		// we already checked that these refs existed, so no need to check
+		// the return value here.
+		_ = r.statusSetter.removeRouteReferences(route, filteredParentStatuses)
+		kubernetesNeedsStatusUpdate = true
+	}
```
andrewstucki, line 96: this goes along with the addition above --  basically if we no longer reference the gateway but do so in our statuses, clean up all of said statuses

control-plane/api-gateway/binding/route_binding.go
```diff
@@ -227,6 +241,11 @@ func (r *Binder) dropConsulRouteParent(snapshot *Snapshot, object client.Object,
 }
 
 func (r *Binder) mutateRouteWithBindingResults(snapshot *Snapshot, object client.Object, gatewayConsulKey api.ResourceReference, resources *common.ResourceMap, results parentBindResults) {
+	if results.boundSections().Cardinality() == 0 {
+		r.dropConsulRouteParent(snapshot, object, r.nonNormalizedConsulKey, r.config.Resources)
+		return
+	}
```
andrewstucki, line 247: this allows us to clean up the parents of a route when it has bound to nothing and also has no actual references to the gateway anymore.

control-plane/api-gateway/common/resources.go
```diff
@@ -400,37 +413,45 @@ func (s *ResourceMap) TranslateAndMutateHTTPRoute(key types.NamespacedName, onUp
...
 		mutated := mutateFn(&consulRoute.route, *translated)
+		if len(mutated.Parents) != 0 {
+			// if we don't have any parents set, we keep this around to allow the route
+			// to be GC'd.
+			delete(s.consulHTTPRoutes, consulKey)
+			s.consulMutations = append(s.consulMutations, &ConsulUpdateOperation{
+				Entry:    &mutated,
+				OnUpdate: onUpdate,
+			})
+		}
```
andrewstucki, line 425: these parent length checks basically mean that if we have no parents we've bound to we're not going to create a route, which allows a bit of a simplification around the cleanup logic -- basically we haven't bound a route to anything, so GC it

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -129,6 +129,15 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	}
 
+	// fetch our inline certificates from cache, this needs to happen
+	// here since the certificates need to be reference counted before
+	// the gateways.
+	r.fetchConsulInlineCertificates(resources)
+
+	// add our current gateway even if it's not controlled by us so we
+	// can garbage collect any resources for it.
+	resources.ReferenceCountGateway(gateway)
```
andrewstucki, line 139: this allows us to garbage collect resources when something happens like a gateway switches gateway classes and is no longer controlled by us since the fetch below this only fetches things that we control

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -153,31 +162,27 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		Gateway:                  gateway,
-		Pods:                     pods,
-		Service:                  service,
-		HTTPRoutes:               httpRoutes,
-		TCPRoutes:                tcpRoutes,
-		Resources:                resources,
-		ConsulGateway:            consulGateway,
-		ConsulHTTPRoutes:         consulHTTPRoutes,
-		ConsulTCPRoutes:          consulTCPRoutes,
-		ConsulInlineCertificates: consulInlineCertificates,
```
andrewstucki, line 179: dropped this and the two route fields above since they weren't used anywhere and a solely tracked by the resources map

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -426,7 +436,12 @@ func (r *GatewayController) transformHTTPRoute(ctx context.Context) func(o clien
 func (r *GatewayController) transformTCPRoute(ctx context.Context) func(o client.Object) []reconcile.Request {
 	return func(o client.Object) []reconcile.Request {
 		route := o.(*gwv1alpha2.TCPRoute)
-		return refsToRequests(common.ParentRefs(common.BetaGroup, common.KindGateway, route.Namespace, route.Spec.ParentRefs))
+
+		refs := refsToRequests(common.ParentRefs(common.BetaGroup, common.KindGateway, route.Namespace, route.Spec.ParentRefs))
+		statusRefs := refsToRequests(common.ParentRefs(common.BetaGroup, common.KindGateway, route.Namespace, common.ConvertSliceFunc(route.Status.Parents, func(parentStatus gwv1beta1.RouteParentStatus) gwv1beta1.ParentReference {
```
andrewstucki, line 441: added so we can trigger reconciliation for cleaning up gateway statuses from routes that have switched parent gateways

andrewstucki (commented):
self review

acceptance/tests/api-gateway/api_gateway_lifecycle_test.go
```diff
@@ -0,0 +1,444 @@
...
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+func TestAPIGateway_Lifecycle(t *testing.T) {
```
t-eckert, line 26: I think it would be worthwhile to add a comment here covering the _why_ of this test. Not so much every scenario because that could get out of date, but a simple "This test modifies Gateway resources to ensure that the controller handles changes in their configuration over time." Or something like that.

t-eckert (approved):

andrewstucki:
Going to merge, but will circle back and write some better doc strings for each of these tests to inform what they're testing from a high level

PR #1696, opened 2022-11-10
ISSUE-3411 as the comment from justas147 stated, the memory limit for…

TheGoderGuy:
… client agents is pretty low and we quickly faced problems we couldn't explain us. Only a describe at the right time showed us an OOM error. This behaviour exists for more than 5 years and could easily be fixed for standard installations with this quick fix for everyone trying out consul on a medium sized project.

Changes proposed in this PR:
- increased resource limit from 100m to 200m
-

How I've tested this PR:
Redeployed our consul with more than 50 routes and seeing no agent-client ooms

How I expect reviewers to test this PR:
Set up with 50 or more routes without the change, looking at various restarts of the client agent(s) and start again with doubled memory to see no restarts and lost service routes at all

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #2249, opened 2023-06-02
Update memory defaults for connect inject controller

andrewstucki:
Changes proposed in this PR:
We're [subscribing to a lot of resources](https://github.com/hashicorp/consul-k8s/blob/46055a324f8b1fcfb6e6debf95f5d059d7b50fe4/control-plane/api-gateway/controllers/gateway_controller.go#L341-L397) for API Gateways because of the complexity of what can trigger reconciliation due to the nature of distributed references within the spec. Running the controller on OpenShift got the controller OOM killed almost immediately. This gives the pod generally more headroom.

Marking as a draft for now to make sure that we verify that this is plenty of space for the controller.

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/2249.txt
+++ b/.changelog/2249.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+helm: Update the default amount of memory used by the connect-inject controller so that its less likely to get OOM killed.
+```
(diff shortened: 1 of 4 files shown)


david-yu:
Perhaps we should bump up Consul server memory limits as well? https://github.com/hashicorp/consul-k8s/pull/1696

david-yu (approved):
Makes sense to do this. Perhaps we should backport as well since it's a small change for consistency?

andrewstucki:
I'm fine with adding a backport, just waiting to mark this ready and merge until next week when we have an opportunity to try out the defaults a bit more.

david-yu:
Backporting to other branches for consistency. I believe this also needed across all other releases as well.

PR #2252, opened 2023-06-02
api-gateway: create RoleBinding attaching Role to ServiceAccount

nathancoleman:
We create a `Role` and `ServiceAccount` for the api-gateway `Deployment` in certain scenarios; however, we do not create a `RoleBinding` to bind them together.

This is causing issues when deploying to OpenShift where the `ServiceAccount` needs a `SecurityContextConstraints` resource attached, and that chain needs to go:
SecurityContextConstraints => Role => RoleBinding => ServiceAccount

**How I've tested this PR:**
Deploy into K8s with system-managed ACLs enabled and verify that the `RoleBinding` is now created alongside the `Role`, `ServiceAccount`, etc.

See above

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/templates/connect-inject-clusterrole.yaml
+++ b/charts/consul/templates/connect-inject-clusterrole.yaml
@@ -72,7 +72,6 @@ rules:
   - watch
   - delete
   - update
-# TODO Can we gate this so that we're only granting when necessary?
 - apiGroups: [ "rbac.authorization.k8s.io" ]
   resources: [ "roles", "rolebindings" ]
   verbs:
--- a/charts/consul/templates/connect-inject-clusterrole.yaml
+++ b/charts/consul/templates/connect-inject-clusterrole.yaml
@@ -72,8 +72,9 @@ rules:
   - watch
   - delete
   - update
+# TODO Can we gate this so that we're only granting when necessary?
 - apiGroups: [ "rbac.authorization.k8s.io" ]
-  resources: [ "roles" ]
+  resources: [ "roles", "rolebindings" ]
   verbs:
   - get
   - list
--- a/control-plane/api-gateway/gatekeeper/deployment.go
+++ b/control-plane/api-gateway/gatekeeper/deployment.go
@@ -75,8 +75,8 @@ func (g *Gatekeeper) upsertDeployment(ctx context.Context, gateway gwv1beta1.Gat
 	return nil
 }
 
-func (g *Gatekeeper) deleteDeployment(ctx context.Context, nsname types.NamespacedName) error {
-	err := g.Client.Delete(ctx, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: nsname.Name, Namespace: nsname.Namespace}})
+func (g *Gatekeeper) deleteDeployment(ctx context.Context, gwName types.NamespacedName) error {
+	err := g.Client.Delete(ctx, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: gwName.Name, Namespace: gwName.Namespace}})
 	if k8serrors.IsNotFound(err) {
 		return nil
 	}
--- a/control-plane/api-gateway/gatekeeper/gatekeeper.go
+++ b/control-plane/api-gateway/gatekeeper/gatekeeper.go
@@ -30,6 +30,7 @@ func New(log logr.Logger, client client.Client) *Gatekeeper {
 }
 
 // Upsert creates or updates the resources for handling routing of network traffic.
+// This is done in order based on dependencies between resources.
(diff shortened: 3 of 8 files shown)


control-plane/api-gateway/gatekeeper/role.go
nathancoleman, line 1: Will consider breaking the `RoleBinding` logic out from the existing `Role` logic before review + merge

charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -72,8 +72,9 @@ rules:
   - watch
   - delete
   - update
+# TODO Can we gate this so that we're only granting when necessary?
```
nathancoleman, line 75: Removing out-of-scope TODO
```suggestion
```

control-plane/api-gateway/gatekeeper/deployment.go
```diff
@@ -75,8 +75,8 @@ func (g *Gatekeeper) upsertDeployment(ctx context.Context, gateway gwv1beta1.Gat
 	return nil
 }
 
-func (g *Gatekeeper) deleteDeployment(ctx context.Context, nsname types.NamespacedName) error {
-	err := g.Client.Delete(ctx, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: nsname.Name, Namespace: nsname.Namespace}})
+func (g *Gatekeeper) deleteDeployment(ctx context.Context, gwName types.NamespacedName) error {
```
nathancoleman, line 78: These changes were just improving a variable name that was confusing to me coming into the project near the end

control-plane/api-gateway/gatekeeper/gatekeeper.go
```diff
@@ -53,20 +57,27 @@ func (g *Gatekeeper) Upsert(ctx context.Context, gateway gwv1beta1.Gateway, gcc
 }
 
 // Delete removes the resources for handling routing of network traffic.
-func (g *Gatekeeper) Delete(ctx context.Context, nsname types.NamespacedName) error {
-	if err := g.deleteRole(ctx, nsname); err != nil {
+// This is done in the reverse order of Upsert due to dependencies between resources.
```
nathancoleman, line 60: Calling this out as the reason for re-organizing the order of delete logic below

control-plane/api-gateway/gatekeeper/gatekeeper.go
```diff
@@ -76,14 +87,14 @@ func (g *Gatekeeper) Delete(ctx context.Context, nsname types.NamespacedName) er
 // resourceMutator is passed to create or update functions to mutate Kubernetes resources.
 type resourceMutator = func() error
 
-func (g Gatekeeper) namespacedName(gateway gwv1beta1.Gateway) types.NamespacedName {
+func (g *Gatekeeper) namespacedName(gateway gwv1beta1.Gateway) types.NamespacedName {
```
nathancoleman, line 90: We were using pointer receivers everywhere else except for this func and the one immediately below

nathancoleman (commented):
Personal review

control-plane/api-gateway/gatekeeper/rolebinding.go
```diff
@@ -0,0 +1,90 @@
...
+	roleBinding = g.roleBinding(gateway, gcc, config)
+	if err := ctrl.SetControllerReference(&gateway, roleBinding, g.Client.Scheme()); err != nil {
+		return err
+	}
+	if err := g.Client.Create(ctx, roleBinding); err != nil {
+		return err
+	}
+
+	return nil
+}
```
mikemorris, line 53: Not blocking, but this looks almost identical to the role upsert logic, would it be worth considering refactoring these into a generic upsert implementation?

mikemorris (approved):

control-plane/api-gateway/gatekeeper/rolebinding.go
nathancoleman: It's almost identical for every resource we're creating. I'd love to refactor but don't wanna blow up the scope of this PR. We can file some work in the current epic to look into this if you're okay with that.

control-plane/api-gateway/gatekeeper/role.go
```diff
@@ -24,27 +24,20 @@ func (g *Gatekeeper) upsertRole(ctx context.Context, gateway gwv1beta1.Gateway,
...
+	} else if !k8serrors.IsNotFound(err) {
 		// Ensure we own the Role.
 		for _, ref := range role.GetOwnerReferences() {
 			if ref.UID == gateway.GetUID() && ref.Name == gateway.GetName() {
 				// We found ourselves!
 				return nil
 			}
 		}
-		return errors.New("Role not owned by controller")
+		return errors.New("role not owned by controller")
```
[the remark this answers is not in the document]
nathancoleman: Yeah, the comment was a copy-pasta leftover from some previous work. No logic change 👍

PR #2260, opened 2023-06-02
Extend retry wait on lifecycle acceptance test

t-eckert:
Changes proposed in this PR:
- Adds a "wait" argument to the `retryCheck` function
- Uses this wait to avoid timeout failures

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


t-eckert:
We'll wait and ensure that the acceptance tests work.

andrewstucki:
Closing in favor of https://github.com/hashicorp/consul-k8s/pull/2261 -- I did some of the time extensions there, but the main thing was that we had a massive flake due to a bug in triggering reconciliation on secret deletion which is fixed there.

PR #2261, opened 2023-06-02
Fix Gateway trigger for when secret is modified

andrewstucki:
Changes proposed in this PR:
The original lifecycle test looks like it was flaky (and rarely passes) due to a bug with fetching the proper gateways to reconcile when a referenced secret is updated. This fixes the indexing error that was causing that, so the Lifecycle tests should pass regularly now.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/api-gateway/controllers/gateway_controller.go
+++ b/control-plane/api-gateway/controllers/gateway_controller.go
@@ -452,11 +452,12 @@ func (r *GatewayController) transformSecret(ctx context.Context) func(o client.O
 		secret := o.(*corev1.Secret)
 		gatewayList := &gwv1beta1.GatewayList{}
 		if err := r.Client.List(ctx, gatewayList, &client.ListOptions{
-			FieldSelector: fields.OneTermEqualSelector(Secret_GatewayIndex, secret.Name),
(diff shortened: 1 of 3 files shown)


control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -452,11 +452,12 @@ func (r *GatewayController) transformSecret(ctx context.Context) func(o client.O
...
 		gatewayList := &gwv1beta1.GatewayList{}
 		if err := r.Client.List(ctx, gatewayList, &client.ListOptions{
-			FieldSelector: fields.OneTermEqualSelector(Secret_GatewayIndex, secret.Name),
+			FieldSelector: fields.OneTermEqualSelector(Secret_GatewayIndex, client.ObjectKeyFromObject(secret).String()),
 		}); err != nil {
 			return nil
 		}
-		return common.ObjectsToReconcileRequests(pointersOf(gatewayList.Items))
+		requests := common.ObjectsToReconcileRequests(pointersOf(gatewayList.Items))
+		return requests
```
nathancoleman, line 460: not blocking, but this part of the change doesn't seem necessary

nathancoleman (approved):

t-eckert (approved):

PR #2265, opened 2023-06-05
Improvement- [NET-189] Added helm inputs for managing audit logs

Ganeshrockz:
Changes proposed in this PR:
- Adds the relevant inputs in values.yaml to configure audit logging in Consul enterprise.
- Behind the scenes, values get translated to the json configuration present [here](https://developer.hashicorp.com/consul/docs/agent/config/config-files#audit) and the whole configuration is added as a file in the Consul server's config map.

How I've tested this PR:
- Unit tests
- CI

- I've added the `auditLog` field under `server`. Should the same be added to `client`?

Checklist:
- [X] Tests added
- [X] CHANGELOG entry added 


charts/consul/templates/server-config-configmap.yaml
```diff
@@ -187,4 +187,29 @@ data:
       }
     }
   {{- end }}
+  {{- if .Values.server.auditLogs.enabled }}
+  audit-logging.json: |-
+    {
+      "audit": {
+        "enabled": "true",
+        "sink": {
```
Ganeshrockz, line 195: sink seems to be a map of `AuditSinks` according to https://github.com/hashicorp/consul/blob/88951bfafaa969f52235dc5b443cb64ba329660c/agent/config/config.go#L811 and so we need to iterate over the key value pair list

david-yu:
I think this one should be safe to backport to 1.0.x and 1.1.x, what do you think?

Ganeshrockz:
> I think this one should be safe to backport to 0.49.x, 1.0.x, and 1.1.x, what do you think?

Appears to be safe. I'll wait for someone from the approvers group to confirm this.

.changelog/2265.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+Add support to provide inputs via helm for audit log related configuration
```
david-yu, line 2:
```suggestion
(Consul Enterprise) Add support to provide inputs via helm for audit log related configuration
```

charts/consul/values.yaml
```diff
@@ -1144,6 +1144,58 @@ server:
     # @type: string
     caCert: null
 
+  # [Enterprise Only] Added in Consul 1.8, the audit object allow users to enable auditing 
+  # and configure a sink and filters for their audit logs
+  auditLogs:
```
david-yu, line 1149: Would it be better to use explicit values instead of an array? Probably easier for a user to provide config but then we'd have to think about defaults.

charts/consul/templates/server-config-configmap.yaml
absolutelightning: we can remove `countOuter` and `countInner` and make it more pragmatic by this 

```
 {
      "audit": {
        "enabled": "true",
        "sink": {
          {{- range $index, $element := .Values.server.auditLogs.sinks }}
          {{- if ne $index 0 }},{{end}}
          "{{ $element.name }}": {
            {{- $firstKeyValuePair := true }}
            {{- range $k, $v := . }}
              {{- if ne $k "name" }}
              {{- if ne $firstKeyValuePair true }},{{end}}
              {{- $firstKeyValuePair = false}}
              "{{ $k }}": "{{ $v }}"
              {{- end }}
            {{- end }}
          }
          {{- end }}
        }
      }
    }
```

absolutelightning (approved):
LGTM.

curtbushko (changes requested):
Thanks for putting this together Ganesh, this is very well done!  Welcome to the world of our helm charts and bats tests!

I only have a few questions & requests but overall this is very good.

A tip for the future:  

1) If you name your branch as `net-###/<anything>` then Jira will detect this and automatically add the branch to your ticket. This works across multiple repositories if you were working on something that affected several code bases like consul-dataplane + consul-k8s.

2) In general, from what I have seen, if we do not add net-### to the titles or description of our PRs. Since we are open source, it distracts from the contents of the PRs since our practitioners wouldn't be able to see what `net-###` is.

charts/consul/templates/server-config-configmap.yaml
```diff
@@ -187,4 +187,27 @@ data:
       }
     }
   {{- end }}
+  {{- if .Values.server.auditLogs.enabled }}
```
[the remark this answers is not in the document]
Ganeshrockz: Added a check here. Also added a condition to fail yaml generation if `.Values.server.auditLogs.enabled` is true but `.Values.global.acls.manageSystemACLs` is false.

Ganeshrockz:
Thanks @curtbushko for the review. Thanks a lot for highlighting the best practices as well ❤️

wilkermichael:
A reminder that the 1.2.x branch is currently under a code freeze. If this isn't a bug fix, unless it's a critical feature for the 1.2.x-rc can you please refrain from merging this until after the 1.2.0-rc1 release (Tuesday, June 13th)

curtbushko (approved):

david-yu:
Will go ahead merge now RC1 is now

david-yu:
@Ganeshrockz You will probably need to cherry pick any failed back ports and approve PRs to get the back ports to merge.

PR #2270, opened 2023-06-05
[API Gateway] Add external consul servers test

andrewstucki:
Changes proposed in this PR:
This adds a test to make sure that we're initializing deployments of API gateways properly when we have external Consul servers. It may not pass, or do so only periodically (be flaky) until https://github.com/hashicorp/consul/pull/17566 lands upstream and makes it into a test docker container -- since that bug is causing a testing issue/race condition with some of the ACL token patching code on deployed gateways.

How I've tested this PR:
Ran this test locally with a custom build of Consul incorporating https://github.com/hashicorp/consul/pull/17566 and saw it pass.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/framework/consul/helm_cluster.go
+++ b/acceptance/framework/consul/helm_cluster.go
@@ -362,13 +362,22 @@ func (h *HelmCluster) Upgrade(t *testing.T, helmValues map[string]string) {
 	k8s.WaitForAllPodsToBeReady(t, h.kubernetesClient, h.helmOptions.KubectlOptions.Namespace, fmt.Sprintf("release=%s", h.releaseName))
 }
 
-func (h *HelmCluster) CreatePortForwardTunnel(t *testing.T, remotePort int) string {
-	serverPod := fmt.Sprintf("%s-consul-server-0", h.releaseName)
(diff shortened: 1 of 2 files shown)


acceptance/framework/consul/cli_cluster.go
```diff
@@ -200,9 +200,14 @@ func (c *CLICluster) Destroy(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func (c *CLICluster) SetupConsulClient(t *testing.T, secure bool) (*api.Client, string) {
+func (c *CLICluster) SetupConsulClient(t *testing.T, secure bool, release ...string) (*api.Client, string) {
 	t.Helper()
 
+	releaseName := c.releaseName
```
jm96441n, line 206: what does release name signify here? I thought it would be something like a specific consul release, but I see we're using a random name in the tests you added

andrewstucki: It's the helm release (i.e. installation) name. The reason that this is here is because in this external servers test (that follows the other external server tests) we spin up two helm installations, one with only a Consul instance that acts as the "external" -- its ACL token that we need for talking to Consul gets persisted in a secret that doesn't match the expected name of the secret used when initializing the Consul client -- so this overrides that.

PR #2278, opened 2023-06-06
[API Gateway] Add partition test

andrewstucki:
Changes proposed in this PR:
This adds a test where a gateway targets a service in a separate partition using a combination of a MeshService CRD, a ServiceResolver, and an ExportedServices entry. It depends on https://github.com/hashicorp/consul/pull/17581 landing, as we had a bug in our core partition handling.

Additionally, I had to revamp the way that authentication was getting handled in deployed services with ACLs enabled.

How I've tested this PR:
Acceptance test passes.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/api-gateway/cache/consul.go
+++ b/control-plane/api-gateway/cache/consul.go
@@ -342,7 +342,7 @@ func (c *Cache) ensureRole(client *api.Client) (string, error) {
 
 	role := &api.ACLRole{
 		Name:        aclRoleName,
-		Description: fmt.Sprintf("ACL Role for Managed API Gateways"),
+		Description: "ACL Role for Managed API Gateways",
 		Policies:    []*api.ACLLink{{ID: policyID}},
 	}
 
--- a/control-plane/api-gateway/cache/consul.go
+++ b/control-plane/api-gateway/cache/consul.go
@@ -4,10 +4,12 @@
 package cache
 
 import (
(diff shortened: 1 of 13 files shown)


control-plane/api-gateway/cache/consul.go
```diff
@@ -278,6 +307,66 @@ func (c *Cache) Write(ctx context.Context, entry api.ConfigEntry) error {
...
+		Policies:    []*api.ACLLink{{ID: policyID}},
+	}
+
+	_, _, err = client.ACL().RoleCreate(role, &api.WriteOptions{})
+	return aclRoleName, err
+}
+
+func (c *Cache) gatewayPolicy() api.ACLPolicy {
+	var data bytes.Buffer
+	if err := gatewayTpl.Execute(&data, templateArgs{
```
jm96441n, line 355: this won't change during runtime right? is there a way we could build this once at start and just it use through the life of the program?

andrewstucki: We could wrap it in a `sync.Once` if we really want to, but part of it is that the namespace bit depends on the runtime flags you pass this. The template itself is compiled once during an init phase.

andrewstucki:
Just waiting for the next dev docker image to be cut for enterprise and then I'll re-run this.

PR #2281, opened 2023-06-06
[API Gateway] Add Consul status to routes and gateways

andrewstucki:
Changes proposed in this PR:
This adds the main status from a synchronized Consul ConfigEntry back into the Kubernetes object. This is helpful because there are some things that we just can't fully validate before sending off to Consul, for example, the protocol of the service that a route references that might result in a discovery chain error.

How I've tested this PR:
Added acceptance test criteria to validate the status gets set properly.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


t-eckert (approved):
I like this approach, I'll follow something similar with Gateway status updates

t-eckert (approved):
I like this approach, I'll follow something similar with Gateway status updates

PR #2282, opened 2023-06-06
Mw/fix pipeline 1 1 6

wilkermichael:
Changes proposed in this PR:
- Updated cloud acceptance test terraform files to fix issues with nightly tests
- updated `main.tf` for gke, eks and aks to use the following block as the old way of setting provider version was deprecated:
```hcl
terraform {
  required_providers {
    aws = {
      version = ">= 4.0.0"
    }
  }
}
```
- Updated aks to latest supported K8s version
- Updated eks cluster to latest supported K8s version
- gke is still using 1.25, which I think is fine for now

How I've tested this PR:
- Manually trigger nightly test: https://github.com/hashicorp/consul-k8s-workflows/actions/runs/5193186962/jobs/9378714025

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/test/terraform/aks/main.tf
+++ b/charts/consul/test/terraform/aks/main.tf
@@ -1,8 +1,15 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
+terraform {
+  required_providers {
+    azurerm = {
+      version = "3.40.0"
+    }
+  }
+}
+
 provider "azurerm" {
-  version = "3.40.0"
   features {}
 }
 
--- a/charts/consul/test/terraform/aks/main.tf
+++ b/charts/consul/test/terraform/aks/main.tf
@@ -48,7 +48,7 @@ resource "azurerm_kubernetes_cluster" "default" {
   location                          = azurerm_resource_group.default[count.index].location
   resource_group_name               = azurerm_resource_group.default[count.index].name
   dns_prefix                        = "consul-k8s-${random_id.suffix[count.index].dec}"
-  kubernetes_version                = "1.24.6"
+  kubernetes_version                = "1.26"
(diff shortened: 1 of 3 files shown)


charts/consul/test/terraform/aks/main.tf
```diff
@@ -48,7 +55,7 @@ resource "azurerm_kubernetes_cluster" "default" {
   location                          = azurerm_resource_group.default[count.index].location
   resource_group_name               = azurerm_resource_group.default[count.index].name
   dns_prefix                        = "consul-k8s-${random_id.suffix[count.index].dec}"
-  kubernetes_version                = "1.24.6"
+  kubernetes_version                = "1.26"
```
curtbushko, line 58: Should this be set to a minor version? Or are you thinking that if we keep at 1.26 we will be ok until the next upgrade?

charts/consul/test/terraform/eks/main.tf
```diff
@@ -61,7 +68,7 @@ module "eks" {
   kubeconfig_api_version = "client.authentication.k8s.io/v1beta1"
 
   cluster_name    = "consul-k8s-${random_id.suffix[count.index].dec}"
-  cluster_version = "1.23"
+  cluster_version = "1.26"
```
curtbushko, line 71: Same comment as above about the version.

wilkermichael:
Talked to @thisisnotashwin , and he thinks it would be a good idea to spread the tests over the supported kubernetes versions. So I will address that in a follow up PR.

charts/consul/test/terraform/eks/main.tf
wilkermichael: Talked to @thisisnotashwin I'll pin these to a patch version as well for better repro-ability

PR #2283, opened 2023-06-06
Remove check for reference grant for route to gateway

jm96441n:
Changes proposed in this PR:
- Reference grants are not necessary for allowing cross namespace talk from a gateway to a route, that is handled by the allowedRoutes section of a gateway:
https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.ReferenceGrant
> All cross-namespace references in Gateway API (with the exception of cross-namespace Gateway-route attachment) require a ReferenceGrant.

How I've tested this PR:
Unit tests

Read code/tests

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


mikemorris (approved):
Confirmed that `bindRoute` still checks `routeAllowedForListenerNamespaces` as required, LGTM.

PR #2293, opened 2023-06-07
Adding support for weighted k8s service

srahul3:
## Context

According to the [documentation](https://www.consul.io/docs/discovery/services#dns-srv-weights) Consul supports setting weights for services . Unfortunately currently there is no way of doing that when using consul-k8s, as there is no annotation that can control that + there is no easy way of doing it via Consul API and it looks like the changes will be anyway overwritten.

Related link: https://hashicorp.atlassian.net/browse/NET-4266

Changes proposed in this PR:
- While the k8s syncs to consul, the LB service annotation `consul.hashicorp.com/service-weight: <number>` is read and then the weight is divided among underlying ingress/ips. The calculated weight is persisted as passing weight in `consulapi.CatalogRegistration`

How I've tested this PR:
Added the Junit test for different scenarios

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


control-plane/catalog/to-consul/resource_test.go
```diff
@@ -56,6 +56,118 @@ func TestServiceResource_createDelete(t *testing.T) {
...
+		syncer.Lock()
+		defer syncer.Unlock()
+		actual := syncer.Registrations
+		require.Len(r, actual, 2)
+		require.Equal(r, "foo", actual[0].Service.Service)
+		require.Equal(r, "3.3.3.3", actual[0].Service.Address)
+		require.Equal(r, 0, actual[0].Service.Weights.Passing)
+		require.Equal(r, "foo", actual[1].Service.Service)
+		require.Equal(r, "4.4.4.4", actual[1].Service.Address)
+		require.Equal(r, 0, actual[1].Service.Weights.Passing)
```
absolutelightning, line 166: Default service weight is 1 for Passing. Should we set it to 1 in case weight is non int?

control-plane/catalog/to-consul/resource.go
```diff
@@ -641,6 +667,20 @@ func (t *ServiceResource) generateRegistrations(key string) {
...
 
+// Sets the passing service weight
+func (t *ServiceResource) setServiceWeight(weight string, appsCount int, r consulapi.CatalogRegistration) error {
+	// error validation if the input param is a number
+	weightI, err := strconv.Atoi(weight)
+	if err != nil {
+		return err
+	} else if weightI > 1 {
+		r.Service.Weights = consulapi.AgentWeights{
+			Passing: weightI / appsCount,
```
absolutelightning, line 678: Lets say weight is 10 and we have 3 service - 
Should we set weights 4, 3, 3 or 3, 3, 3?
Later one sums to 9 not 10.

control-plane/catalog/to-consul/resource.go
```diff
@@ -511,6 +511,18 @@ func (t *ServiceResource) generateRegistrations(key string) {
 			r.Service = &rs
 			r.Service.ID = serviceID(r.Service.Service, ip)
 			r.Service.Address = ip
+			// Adding information about service weight.
+			// Overrides the existing weight if present
+			if weight, ok := svc.Annotations[annotationServiceWeight]; ok || weight != "" {
+				err := t.setServiceWeight(weight, len(ips), r)
```
absolutelightning, line 517: Should this be `&r`?

absolutelightning:
Documentation has passing and warning two values for service weights. Should we also use both?

control-plane/catalog/to-consul/resource_test.go
srahul3: We set it or not, default is 1 😃

control-plane/catalog/to-consul/resource.go
srahul3: We can certainly try for 4,3,3 but 3,3,3 is also near accurate. In my experience, the traffic routing is not so precise.

srahul3:
> Documentation has passing and warning two values for service weights. Should we also use both?

Passing is the primary operation mode. According to last conversation with David, implementing passing is what is required for now.

control-plane/catalog/to-consul/resource.go
srahul3: What is the rational behind using `&r`? I came from java background so help me understand.

absolutelightning: Its about pass by reference. If tests are passing let it be.

control-plane/catalog/to-consul/resource.go
```diff
@@ -641,6 +667,24 @@ func (t *ServiceResource) generateRegistrations(key string) {
 	}
 }
 
+// Sets the passing service weight
+func (t *ServiceResource) setServiceWeight(weight string, appsCount int, r consulapi.CatalogRegistration) error {
+	// error validation if the input param is a number
+	weightI, err := strconv.Atoi(weight)
+	if err != nil {
+		return err
+	} else if weightI > 1 {
```
Ganeshrockz, line 676: It is always good to keep the happy path aligned towards the left which roughly translates to avoiding `else` and nested if blocks. This improves readability a lot. You could refactor this block something like

```go
weightI, err := strconv.Atoi(weight)
if err != nil {
     return err
}

if weightI <=1 {
   return nil
}

var perAppWeight = weightI / appsCount
if perAppWeight < 1 {
   perAppWeight = 1
}
r.Service.Weights = consulapi.AgentWeights{
	Passing: perAppWeight,
}

return nil
```

to make it more readable.

control-plane/catalog/to-consul/resource.go
Ganeshrockz: I guess it should be `&r` if you are mutating attributes of `r`. In the current state, the struct gets passed by value semantics

See https://go.dev/play/p/qX0awpC_yY9 and https://go.dev/play/p/OmmrCWLBhGh

control-plane/catalog/to-consul/resource.go
```diff
@@ -641,6 +667,24 @@ func (t *ServiceResource) generateRegistrations(key string) {
 	}
 }
 
+// Sets the passing service weight
+func (t *ServiceResource) setServiceWeight(weight string, appsCount int, r consulapi.CatalogRegistration) error {
```
Ganeshrockz, line 671: The caller need of be type `*ServiceResource` to use this function because it isn't getting used anywhere inside it. I would recommend to just have it as `func setServiceWeight` and move this function to the bottom of the file.

control-plane/catalog/to-consul/resource.go
```diff
@@ -999,3 +1025,26 @@ func (t *ServiceResource) isIngressService(key string) bool {
...
+	weightI, err := strconv.Atoi(weight)
+	if err != nil {
+		return err
+	}
+
+	if weightI <= 1 {
+		return nil
+	}
+
+	var perAppWeight = weightI / appsCount
```
absolutelightning, line 1041: I think we can set weight to whatever value we are getting.

absolutelightning: 1. One simple reason being - if a customer is setting weight - x for a service - and if they query the will get another value from APIs which will be divided.
2. Weight are proportional - if you have two services with weights 10 and 20 respectively, the service with weight 20 would receive twice as much traffic as the service with weight 10. If we divide, we will not get the correct traffic.

thisisnotashwin: yup! we should set the weight that is set on the annotation and we dont need the app count for the reasons @absolutelightning described!

control-plane/catalog/to-consul/resource.go
```diff
@@ -999,3 +1025,26 @@ func (t *ServiceResource) isIngressService(key string) bool {
 func consulHealthCheckID(k8sNS string, serviceID string) string {
 	return fmt.Sprintf("%s/%s", k8sNS, serviceID)
 }
+
+// Sets the passing service weight
+func setServiceWeight(weight string, appsCount int, r *consulapi.CatalogRegistration) error {
```
thisisnotashwin, line 1030: id change this method to return the service weight or an error, providing the value of the annotation as an input. we should avoid sending in the consul registration into it. it will make this is a cleaner interface that is significantly easier to test.

absolutelightning:
I think we need to add documentation of the new annotation in `website/content/docs/k8s/annotations-and-labels.mdx`

control-plane/catalog/to-consul/resource.go
```diff
@@ -547,6 +562,19 @@ func (t *ServiceResource) generateRegistrations(key string) {
 				r.Service.ID = serviceID(r.Service.Service, addr)
 				r.Service.Address = addr
 
+				// Adding information about service weight.
+				// Overrides the existing weight if present
+				if weight, ok := svc.Annotations[annotationServiceWeight]; ok && weight != "" {
```
absolutelightning, line 567: this line 567 - 577 and lines 518 - 528 are exactly the same. can we take it out in a function as well?

srahul3: The code at both place is constructing the CatalogRegistration object, there is logic which is duplicate. Read the comment https://github.com/hashicorp/consul-k8s/pull/2293#discussion_r1221683819

absolutelightning (approved):

control-plane/catalog/to-consul/resource.go
```diff
@@ -999,3 +1025,18 @@ func (t *ServiceResource) isIngressService(key string) bool {
...
+// Calculates the passing service weight
+func getServiceWeight(weight string) (int, error) {
+	// error validation if the input param is a number
+	weightI, err := strconv.Atoi(weight)
+	if err != nil {
+		return -1, err
+	}
+
+	if weightI <= 1 {
+		return -1, fmt.Errorf("expecting the service annotation %s value to be greater than 1", annotationServiceWeight)
```
thisisnotashwin, line 1038: Maybe we can add some additional checks here. Can the service weight be greater than 100? if it can, ignore this comment, but just wanted to ensure we have completeness here wrt failing for all invalid service weight!

control-plane/catalog/to-consul/resource_test.go
```diff
@@ -56,6 +56,153 @@ func TestServiceResource_createDelete(t *testing.T) {
...
+		require.Equal(r, 22, actual[0].Service.Weights.Passing)
+		require.Equal(r, "foo", actual[1].Service.Service)
+		require.Equal(r, "4.4.4.4", actual[1].Service.Address)
+		require.Equal(r, 22, actual[1].Service.Weights.Passing)
+		require.NotEqual(r, actual[0].Service.ID, actual[1].Service.ID)
+	})
+}
+
+// Test that Loadbalancer service weight is set from service annotation.
+func TestServiceWeight_invalidValue(t *testing.T) {
```
thisisnotashwin, line 138: IMO this test and the next can be simplified by adding table tests for the `getServiceWeight` method. it should make the setup simpler to read and test as well. what do you think?

control-plane/catalog/to-consul/resource.go
srahul3: Hey thanks for sharing the thoughts. I couldn't find any documentation which mentions the weight to be percentage. To start with all the service default weight is set to 1 hence it is 1:1 which can be considered as equal ratio diversion. I guess in that case it is allowed to go beyond 100.

I have seen some example on stackoverflow, users configuring it to 10000:1, not sure if it is valid scenario unless someone wants to do super safe canary test.

control-plane/catalog/to-consul/resource_test.go
srahul3: I will address this 🤝

srahul3: @thisisnotashwin I have refactored and collapses all the external IP test into one table test.

control-plane/catalog/to-consul/resource.go
thisisnotashwin: Cool! thanks for investigating this!

control-plane/catalog/to-consul/resource_test.go
```diff
@@ -56,6 +56,115 @@ func TestServiceResource_createDelete(t *testing.T) {
...
+			client := fake.NewSimpleClientset()
+			syncer := newTestSyncer()
+			serviceResource := defaultServiceResource(client, syncer)
+
+			// Start the controller
+			closer := controller.TestControllerRun(&serviceResource)
+			defer closer()
+
+			// Insert an LB service
+			svc := lbService("foo", metav1.NamespaceDefault, "1.2.3.4")
```
thisisnotashwin, line 143: we can simplify this test. it looks like all we want to do is test the method `getSeviceWeight`. We don't need to configure the syncer. i think a much simpler test is directly invoking the method with different string values and verifying if the returned value is `-1, err` or `weight, nil` if that makes sense. happy to discuss this on a call!!

thisisnotashwin (commented):
This is looking really great! I have one comment about simplifying the testing for `getServiceWeights` but after that, this looks merge ready!!

control-plane/catalog/to-consul/resource.go
srahul3: Implemented as suggested 😃

control-plane/catalog/to-consul/resource_test.go
srahul3: Implemented as suggested 😃

thisisnotashwin (approved):
excellent! can you check why the pipelines are failing? this is good once those are addressed

srahul3:
@david-yu can you test it on latest. If thing are working as expected, I will backport the changes.

david-yu:
Adding backports to 1.0.x, 1.1.x and 1.2.x. Thanks @srahul3 could you follow the backport PRs to see they are back ported correctly and passing tests?

PR #2295, opened 2023-06-07
[API Gateway] WAN Federation test and fixes

andrewstucki:
Changes proposed in this PR:
This adds a WAN federation acceptance test for API Gateways. A few fixes went into it:
1. We were disallowing ServiceResolver cross-DC redirects unless someone was explicitly in the default partition, but were only checking for "default" when an empty string means the same thing
2. We needed to normalize the auth method name that we use for our deployments so that we use the DC-local one that we attach our binding rules to.
3. Due to the way we watch and GC config entries, we needed to pass some additional metadata to ignore anything coming from a sync to a federated DC -- basically we were removing config entries if we didn't have the corresponding Kubernetes object backing them in our DC, when we shouldn't touch them because a corresponding Kubernetes object exists for them in another DC.

How I've tested this PR:
I tested this by running this locally on Kind with MetalLB installed so that the mesh gateways could find each other over WAN, but this doesn't typically work in a Kind setup, as the other acceptance test [mentions](https://github.com/hashicorp/consul-k8s/blob/644e02ea206fa4d28dcba34712fce00246ef655c/acceptance/tests/wan-federation/wan_federation_test.go#L47-L49), so it will be disabled for typical CI runs.

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/fixtures/cases/api-gateways/dc1-to-dc2-resolver/kustomization.yaml
+++ b/acceptance/tests/fixtures/cases/api-gateways/dc1-to-dc2-resolver/kustomization.yaml
@@ -1,0 +1,5 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+resources:
+  - serviceresolver.yaml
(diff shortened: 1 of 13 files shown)


wilkermichael:
A reminder that the 1.2.x branch is currently under a code freeze. If this isn't a bug fix/test introduction, and unless it's a critical feature for the 1.2.x-rc can you please refrain from merging this until after the 1.2.0-rc1 release (Tuesday, June 13th)

andrewstucki:
This is a bug fix/test introduction and is required for the subsequent GA release, so if possible I'd like to get it merged before then.

t-eckert (approved):
Great fixes and great test coverage with the acceptance test!

PR #2302, opened 2023-06-07
Added logLevel field for components 

Ganeshrockz:
Changes proposed in this PR:

This PR is aimed at fixing https://github.com/hashicorp/consul-k8s/issues/1256. I've added the `logLevel` attribute to certain low level components which could be used to override the `global.logLevel` flag. This gives users fine grained control over setting logLevels for individual components. For example, `ingressGateways.logLevel` is introduced to override `global.logLevel` within ingress gateway specific kube manifests.

To make it a non breaking change, I have assigned an empty string as the default value for the newly introduced logLevel fields. 

How I've tested this PR:

- Unit tests
- CI

Checklist:
- [X] Tests added
- [X] CHANGELOG entry added 


wilkermichael:
A reminder that the 1.2.x branch is currently under a code freeze. If this isn't a bug fix/test introduction, and unless it's a critical feature for the 1.2.x-rc can you please refrain from merging this until after the 1.2.0-rc1 release (Tuesday, June 13th)

Ganeshrockz:
@DanStough Can you take a look at this PR whenever possible?

absolutelightning:
I was thinking why not we keep `info` as default value rather than '' string. What do you think? @Ganeshrockz

charts/consul/templates/ingress-gateways-deployment.yaml
DanStough, line 1: This change is fine, but in addition, we need to pass the new log level to the main `ingress-gateway` container. 
https://github.com/hashicorp/consul-k8s/blob/8f474854c4a58e5b11290fb79b8ae434a430962e/charts/consul/templates/ingress-gateways-deployment.yaml#L322-L323

It appears that Envoy will inherit the log level that is passed to the [dataplane container](https://github.com/hashicorp/consul-dataplane/blob/33a162d66cc5c132080878b835ce97561da02cf6/pkg/envoy/proxy.go#L398-L427)

charts/consul/templates/mesh-gateway-deployment.yaml
DanStough, line 1: Same comment here about passing the loglevel to the dataplane container.

charts/consul/templates/telemetry-collector-deployment.yaml
DanStough, line 1: Same comment about passing into the consul-dataplane container.

charts/consul/templates/terminating-gateways-deployment.yaml
DanStough, line 1: Same comment about passing into the consul-dataplane container.

charts/consul/values.yaml
```diff
@@ -2503,6 +2514,10 @@ meshGateway:
   # Requirements: consul 1.6.0+ if using `global.acls.manageSystemACLs``.
   enabled: false
 
+  # Override global log verbosity level for mesh-gateway-deployment pods. One of "debug", "info", "warn", or "error".
+  # @type: string
+  logLevel: ""
+
```
DanStough, line 2520: For any `consul-dataplane` deployment, there are a larger set of options for Envoy's logLevels. Particularly `trace` is important for seeing traffic pass through Envoy. We need to add this to the doc, but we also need to find a way to make sure that we can set `trace` without breaking the control-plane image. I haven't looked into what would happen in this case.

This comment applies to all sections where I noted in the template file there is a dataplane instance (e.g. gateways)

charts/consul/test/unit/ingress-gateways-deployment.bats
```diff
@@ -1504,3 +1504,35 @@ key2: value2' \
...
+# logLevel
+
+@test "ingressGateways/Deployment: use global.logLevel by default" {
+  cd `chart_dir`
+  local cmd=$(helm template \
+      -s templates/ingress-gateways-deployment.yaml  \
+      --set 'ingressGateways.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.initContainers[0].command' | tee /dev/stderr)
```
DanStough, line 1518: You'll want to test both containers when you come up with a solution that let's you pass `trace` to dataplane but not the initContainer.

This applies to the wider set of deployments w/ dataplane containers.

DanStough (changes requested):
Thanks @Ganeshrockz. I like your attention to backward compatibility.  

My two biggest concerns:
1. Making sure we pass the log level to dataplane containers, where most of the interesting working will happen
2. Making sure that we can pass `trace` to dataplane without breaking any of the other deployments

Since I'm not as familiar with the k8s repo, it would also be good to get a review from @ndhanushkodi or maybe @curtbushko

david-yu:
Hi @Ganeshrockz any progress on this one?

Ganeshrockz:
> Hi @Ganeshrockz any progress on this one?

I couldn't focus on this over the past couple of weeks because of other stuff. I'll prioritise this early next week

.changelog/2302.txt
```diff
@@ -0,0 +1,10 @@
+```release-note:improvement
+Add support to provide the logLevel flag via helm for multiple low level components. Introduces the following fields
+1. server.logLevel
+2. global.federation.logLevel
+3. global.gossipEncryption.logLevel
+4. meshGateway.logLevel
+5. ingressGateways.logLevel
+6. terminatingGateways.logLevel
+7. telemetryCollector.logLevel
```
ndhanushkodi, line 9:
```suggestion
Add support to provide the logLevel flag via helm for multiple low level components. Introduces the following fields
1. `server.logLevel`
2. `global.federation.logLevel`
3. `global.gossipEncryption.logLevel`
4. `meshGateway.logLevel`
5. `ingressGateways.logLevel`
6. `terminatingGateways.logLevel`
7. `telemetryCollector.logLevel`
```

charts/consul/test/unit/ingress-gateways-deployment.bats
ndhanushkodi: +1

charts/consul/values.yaml
ndhanushkodi: We could consider:
Option 1:  separate fields with different log levels, where for consul-k8s components you can set debug/info/warn/error and for dataplane you can set debug/info/warn/error/trace
Option 2: allow consul-k8s components to accept "trace" log level, but internally just use debug for it instead.

ndhanushkodi (commented):
Just gave this a look since Dan's out this week! I mostly just agree with everything he's said so far, that we need to handle the dataplane container and handle the trace log case! I added a few options that I could think of for handling the trace log case.

charts/consul/values.yaml
Ganeshrockz: I like option 2 in this case. Option 1 seems to add a lot of fields to the configuration. @david-yu you have any thoughts here?

david-yu: Option 2 sounds good to me too.

Ganeshrockz: I dont think we need to handle anything special in case of `consul-k8s` components wrt log levels. Following are the consul-k8s manifests container the dataplane container

1. ingress-gateways-deployment
2. terminating-gateways-deployment
3. mesh-gateways-deployment
4. telemetry-collector-deployment

All the other files are k8s jobs which don't have a dependency with dataplane. It looks like connect-init already supports trace as a logLevel and I don't see differences between the two flags (in [connect-init](https://github.com/hashicorp/consul-k8s/blob/6624d34b9f1a4d0f12cef156c5f35152ee3e08b5/control-plane/subcommand/connect-init/command.go#L87) and [consul-dataplane](https://github.com/hashicorp/consul-dataplane/blob/2e4715d1939633ed16e4dc06784607f5d926b180/cmd/consul-dataplane/main.go#L41)). I will change the comments in the `values.yaml`

charts/consul/test/unit/ingress-gateways-deployment.bats
Ganeshrockz: Added additional tests

Ganeshrockz:
@ndhanushkodi Addressed all the comments. Please re-review this PR whenever you get time.

charts/consul/templates/server-acl-init-job.yaml
```diff
@@ -168,7 +168,7 @@ spec:
           CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
           consul-k8s-control-plane server-acl-init \
-            -log-level={{ .Values.global.logLevel }} \
+            -log-level={{ default .Values.global.logLevel .Values.server.logLevel}} \
```
ndhanushkodi, line 171: Does it make sense to have server-acl-init, server-acl-init-cleanup and tls-init use the `server.logLevel` value? The `server` values stanza is typically meant for configuring the consul servers themselves, and this value is now configuring a few consul-k8s jobs but right now it doesn't configure the consul servers themselves. Maybe for the server-acl-init and cleanup jobs, the log level could be under `.Values.acls`, and for tls-init the log level could be under `.Values.tls`.

ndhanushkodi (commented):
Looking great! 

One last comment about the `server.logLevel`-- I think it could be confusing to have that value configure only consul k8s helper jobs that configure acls/tls, and not the servers themselves.

charts/consul/values.yaml
ndhanushkodi: Ah yep, consul-k8s does seem to handle trace as a log level through the hclog logger! So it should be fine to pass trace level in.

charts/consul/templates/server-acl-init-job.yaml
Ganeshrockz: Makes sense. Removed `server.logLevel` and added ` global.acls.logLevel` and `global.tls.logLevel` instead

ndhanushkodi (approved):
Looks great!!

david-yu:
@ndhanushkodi How comfortable do you feel backporting this? Or should we only introduce in upcoming release of Consul K8s?

charts/consul/values.yaml
```diff
@@ -2636,6 +2651,10 @@ meshGateway:
   # Requirements: consul 1.6.0+ if using `global.acls.manageSystemACLs``.
   enabled: false
 
+  # Override global log verbosity level for mesh-gateway-deployment pods. One of "trace", "debug", "info", "warn", or "error".
+  # @type: string
+  logLevel: ""
```
ndhanushkodi, line 2656: Oh one last thing-- I think when we default to some higher level value, we typically use `"-"` rather than empty string as the default.

ndhanushkodi:
@david-yu It is backwards compatible, seems fine to backport!

david-yu:
> @david-yu It is backwards compatible, seems fine to backport!

Thanks added back ports for 1.0.x to 1.2.x!

charts/consul/values.yaml
Ganeshrockz: I see this pattern mainly for boolean based fields. Newly introduced string based fields always have an empty string. DO we still need to change this?

david-yu:
@Ganeshrockz Just to clarify snapshot agent logLevel was not added because the logLevel is buried within a K8s secret according to our docs? @ndhanushkodi Does it make sense to take this config out of the secret itself? https://developer.hashicorp.com/consul/docs/k8s/helm#v-server-snapshotagent-configsecret.

ndhanushkodi:
Snapshot agent is already something that [should be configured via the config secret](https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/values.yaml#L1170) because there are storage secrets in there, so you'd already always need to edit that file. If you already need to edit that secret to configure snapshot agent, I think it's probably ok that the log level is configured there also. 

This means the logging configuration (after this PR) looks like majority of consul-k8s components are configured via direct helm values, consul servers/client agents are configured via the [extraConfig value](https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/values.yaml#L937-L955), and snapshot agent logging is configured via a config secret.

To allow the consul servers/snapshot agent to be configured directly through helm values, it'd require a bit of extra work to template that in, especially with the snapshot agent (we'd need to test whether it can merge two pieces of configuration, from the secret and from the helm value). It could be worth doing if we want to fully streamline the logging config to helm values.

david-yu:
Ok I think what we have is ok but would it be possible to get at least the server and client logLevels controlled by helm instead of having to create extraConfig? I think we can live with the snapShot agent config for now.

david-yu:
I'm also ok with not backporting if that makes introducing the helm config for servers and clients easier.

ndhanushkodi:
@david-yu yeah I think that's doable-- I think it could be added here to this configmap: https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/server-config-configmap.yaml from the helm value for servers and similarly here https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/client-config-configmap.yaml for clients

david-yu:
@Ganeshrockz I assume this is ready for re-review?

.changelog/2302.txt
```diff
@@ -0,0 +1,11 @@
+```release-note:improvement
+Add support to provide the logLevel flag via helm for multiple low level components. Introduces the following fields
+1. `global.acls.logLevel`
+2. `global.tls.logLevel`
+3. `global.federation.logLevel`
+4. `global.gossipEncryption.logLevel`
```
david-yu, line 6: We should add server.logLevel and client.logLevel here.

Ganeshrockz:
> @Ganeshrockz I assume this is ready for re-review?

@david-yu Yes it is ready now. Fixed the changelog as well.

david-yu:
Thanks @ndhanushkodi could you review? We can go ahead and merge as well after tests run.

david-yu (approved):
Approving on my end. Thank you for making these changes.

ndhanushkodi (approved):
looks good to me!

Ganeshrockz:
> I'm also ok with not backporting if that makes introducing the helm config for servers and clients easier.



@david-yu Should this still be backported?

david-yu:
I would like to backport if it's possible as is, otherwise we can just merge to main. @ndhanushkodi what are your thoughts?

charts/consul/values.yaml
ndhanushkodi: Yup you're right!

ndhanushkodi:
It should be all additive changes, so it's fine to backport!

Ganeshrockz:
Thanks for all the help on this @david-yu  and @ndhanushkodi

Ganeshrockz:
Looking into the failing acceptance test. Don't think it's because of my change though

Ganeshrockz:
Will raise backports on Monday

PR #2304, opened 2023-06-07
Mw/net 3598 update kind for consul k8s acceptance tests with latest version of kind and k8s 1.27

wilkermichael:
Changes proposed in this PR:
- I used this branch to test 1.27 support in the workflow, the acceptance tests passing demonstrates that that works correctly
- I also updated the cloud tests to cover 1.24, 1.25 and 1.26. EKS seems to not support setting a patch version for kubernetes? In the past we haven't set it, so I've left it set to `1.26`. GKE and AKS, I have pinned to a patch number in order to help with repro-ability.

How I've tested this PR:
- I temporarily enabled cloud tests on this PR, here are the results (they didn't all pass, but I just wanted to verify that they setup with no issues with the version changes): https://github.com/hashicorp/consul-k8s-workflows/actions/runs/5216611633/jobs/9415586427

👀 

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.changelog/2304.txt
+++ b/.changelog/2304.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+helm: Kubernetes v1.27 is now supported. Minimum tested version of Kubernetes is now v1.24.
+```
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ by contacting us at [security@hashicorp.com](mailto:security@hashicorp.com).
 
 The following pre-requisites must be met before installing Consul on Kubernetes. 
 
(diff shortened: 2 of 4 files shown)


.changelog/2304.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+helm: Kubernetes v1.27 is now supported. Minimum tested version of Kubernetes is now v1.24.
```
wilkermichael, line 2: This didn't really require a code change to Consul-K8s, but I noticed we added a line like this for 1.26 last release... so did that here as well

curtbushko (approved):
LGTM!  

Thanks for testing through the cloud tests; I know it can take a while to run!

PR #2305, opened 2023-06-07
Ensure Reconciliation Stops

jm96441n:
Changes proposed in this PR:
- Test for ensuring reconciliation does not go on infinitely
-

How I've tested this PR:
Wrote the test/ran the test

read the code/run the test

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/api-gateway/controllers/gateway_controller_integration_test.go
+++ b/control-plane/api-gateway/controllers/gateway_controller_integration_test.go
@@ -68,7 +68,7 @@ func TestControllerDoesNotInfinitelyReconcile(t *testing.T) {
 			namespace:   "",
 			certFn:      createCert,
 			gwFn:        createFunkyCasingFieldsAPIGW,
-			httpRouteFn: createAllFieldsSetHTTPRoute,
+			httpRouteFn: createFunkyCasingFieldsHTTPRoute,
 			tcpRouteFn:  createFunkyCasingFieldsTCPRoute,
 		},
 	}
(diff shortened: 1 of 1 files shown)


control-plane/api-gateway/controllers/gateway_controller_integration_test.go
```diff
@@ -0,0 +1,671 @@
...
+		NamespacedName: types.NamespacedName{
+			Namespace: k8sGWObj.Namespace,
+			Name:      k8sGWObj.Name,
+		},
+	})
+	require.NoError(t, err)
+
+	wg := &sync.WaitGroup{}
+	wg.Add(3)
+	// wg.Add(4)
```
jm96441n, line 126: for some reason the inline cert never sends an event down the channel. It looks like it's something to do with the fact that when it's initially created the gateway doesn't reference it, but once the gateway does reference it there's no changes made to the cert so it never actually gets a notification from consul: 
So we never enter the loop here:
https://github.com/hashicorp/consul-k8s/blob/NET-4231-halting-problem/control-plane/api-gateway/cache/consul.go#L250

because the translator function called never sees that the gateway references the secret:
https://github.com/hashicorp/consul-k8s/blob/NET-4231-halting-problem/control-plane/api-gateway/controllers/gateway_controller.go#L556

jm96441n:
@wilkermichael thanks for the reminder! this is just a test introduction

control-plane/api-gateway/controllers/gateway_controller_integration_test.go
andrewstucki: talked about this on Slack, this is expected and due to the ordering of operations against Consul -- the cache notifier gets the cert notification and does the translation prior to writing the gateway back into the cache, so no reconciliation happens due to the cert creation, but the gateway update triggers it as soon as it references the cert

andrewstucki (approved):
Generally looks good to me, I'm sure we can add onto this as we run into any halting issues we may miss between the Consul/Kubernetes normalization. Feel free to merge when tests pass.

PR #2317, opened 2023-06-08
Rename GatewayClassController to prevent name collision

t-eckert:
Changes proposed in this PR:
- Rename the GatewayClass Controller to `consul.hashicorp.com/gatewayclass-controller`.

How I've tested this PR:
- Using acceptance tests

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


andrewstucki:
Adding a backport tag on this since it needs to go in or it will cause issues for anyone upgrading from the legacy gateway implementation.

acceptance/tests/api-gateway/api_gateway_test.go
```diff
@@ -25,7 +25,7 @@ import (
 
 const (
 	StaticClientName           = "static-client"
-	gatewayClassControllerName = "hashicorp.com/consul-api-gateway-controller"
+	gatewayClassControllerName = "consul.hashicorp.com/gatewayclass-controller"
```
andrewstucki, line 28: I know this is technically what's being targeted, but thinking it'd probably make more sense to use the shorter name since this is something users would need to add manually if creating their own `GatewayClass` objects:

```suggestion
	gatewayClassControllerName = "consul.hashicorp.com/gateway-controller"
```

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -755,7 +757,7 @@ func (c *GatewayController) getConfigForGatewayClass(ctx context.Context, gatewa
 	if ref := gatewayClassConfig.Spec.ParametersRef; ref != nil {
 		if string(ref.Group) != v1alpha1.GroupVersion.Group ||
 			ref.Kind != v1alpha1.GatewayClassConfigKind ||
-			gatewayClassConfig.Spec.ControllerName != GatewayClassControllerName {
+			string(gatewayClassConfig.Spec.ControllerName) != c.Name {
```
andrewstucki, line 760: since we're already using the `common.GatewayClassControllerName` above, I'd prefer to just stick with that here for minimal changes and consistency rather than passing in the name as a field one place and then using the constant in the `Reconcile` loop above.

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -782,7 +784,7 @@ func (c *GatewayController) fetchControlledGateways(ctx context.Context, resourc
 
 	list := gwv1beta1.GatewayClassList{}
 	if err := c.Client.List(ctx, &list, &client.ListOptions{
-		FieldSelector: fields.OneTermEqualSelector(GatewayClass_ControllerNameIndex, GatewayClassControllerName),
+		FieldSelector: fields.OneTermEqualSelector(GatewayClass_ControllerNameIndex, c.Name),
```
andrewstucki, line 787: same as above

control-plane/subcommand/inject-connect/command.go
```diff
@@ -476,7 +476,7 @@ func (c *Command) Run(args []string) int {
 	}
 
 	if err := (&gatewaycontrollers.GatewayClassController{
-		ControllerName: gatewaycontrollers.GatewayClassControllerName,
+		ControllerName: gatewaycommon.GatewayClassControllerName,
```
andrewstucki, line 479: ideally we'd probably just use the constant directly within the controller itself rather than doing this, for consistency like above, but this is fine to keep like this for now

andrewstucki (changes requested):
A few changes, but general idea looks good to me, not sure if it's the tests being flaky or what, but there are some acceptance test failures

acceptance/tests/api-gateway/api_gateway_test.go
t-eckert: Did this with a find/replace so I'll close your suggestion.

andrewstucki (approved):
Thanks for keeping this fix pretty tight -- going to approve -- feel free to merge when the tests pass.

PR #2321, opened 2023-06-09
[API Gateway] fix dangling service registrations

andrewstucki:
Changes proposed in this PR:
This bypasses the cache on gateway deletion to make sure we have dropped any registered services in Consul even if we haven't been informed about them via the cache yet. We had a race condition where if a pod was just registered with Consul we might not know about it if the gateway was deleted right away. I noticed this scenario when running conformance tests when we were leaving around services that were no longer there.

How I've tested this PR:
Ran conformance tests and made sure that quickly deleted gateways are cleaned up in Consul

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


t-eckert (approved):
Glad you caught this!

PR #2333, opened 2023-06-10
Add CRT docker changes for release workflow

skpratt:
build: https://github.com/hashicorp/consul-k8s/actions/runs/5227978283

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -271,6 +271,7 @@ jobs:
           unzip -j *.zip
       - name: Docker Build (Action)
         uses: hashicorp/actions-docker-build@v1
+        if: ${{ !matrix.fips }}
         with:
           smoke_test: |
             TEST_VERSION="$(docker run "${IMAGE_NAME}" consul-k8s-control-plane version  | awk '{print $2}')"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -380,4 +426,26 @@ jobs:
...
+            fi
+            echo "Test PASSED"
+          version: ${{ env.version }}
+          target: ubi-fips # duplicate target to distinguish FIPS builds in CRT machinery
+          arch: ${{ matrix.arch }}
+          pkg_name: consul-k8s-control-plane_${{ env.version }}
+          bin_name: consul-k8s-control-plane
+          workdir: control-plane
+          tags: |
+            docker.io/hashicorp/${{ env.repo }}-control-plane:${{ env.version }}-ubi
```
zalimeni, line 448: I noticed this tag had `-fips` in it: https://github.com/hashicorp/consul-k8s/pull/2333/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R313 

Should this line too?

zalimeni (approved):
One question and generally curious which FIPS bits need vs. don't need `fips` in the name, but approving to unblock once that's resolved.

zalimeni:
cc @wilkermichael

.github/workflows/build.yml
skpratt: yes! thank you for spotting!!

wilkermichael (approved):

PR #2346, opened 2023-06-12
Set Consul Service Instance Localities from K8s Node Labels

erichaberkorn:
Changes proposed in this PR:
- Set localities on service instances similarly to Consul servers https://github.com/hashicorp/consul-k8s/blob/main/control-plane/subcommand/fetch-server-region/command.go
- The connect injector already has permissions to read nodes https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/connect-inject-clusterrole.yaml#L68-L76

How I've tested this PR:

I manually tested that localities are properly set using the sameness demo.

👁️ 

Checklist:
- [X] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/.changelog/2346.txt
+++ b/.changelog/2346.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+Set locality on services registered with connect-inject.
+```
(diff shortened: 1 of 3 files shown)


thisisnotashwin:
@erichaberkorn sorry i should have spotted this but we should backport this to `1.2` now that we have cut a branch for it.

PR #2353, opened 2023-06-13
fix: use correct flag when translating namespaces

nathancoleman:
Fix use of incorrect config flag when translating namespaces

**How I've tested this PR:**
Run conformance tests defined in kubernetes-sigs/gateway-api. These were previously failing because every test involves deleting a Gateway - now they're passing.

CleanShot 2023-06-13 at 09 25 25@2x

See above

**Checklist:**
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/control-plane/api-gateway/cache/gateway.go
+++ b/control-plane/api-gateway/cache/gateway.go
@@ -18,7 +18,7 @@ import (
 )
 
 type GatewayCache struct {
-	config    *consul.Config
+	config    Config
 	serverMgr consul.ServerConnectionManager
 	logger    logr.Logger
 
@@ -35,7 +35,7 @@ type GatewayCache struct {
 
 func NewGatewayCache(ctx context.Context, config Config) *GatewayCache {
 	return &GatewayCache{
-		config:             config.ConsulClientConfig,
+		config:             config,
 		serverMgr:          config.ConsulServerConnMgr,
 		logger:             config.Logger,
 		events:             make(chan event.GenericEvent),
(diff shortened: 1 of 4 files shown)


control-plane/api-gateway/cache/gateway.go
```diff
@@ -53,13 +53,13 @@ func (r *GatewayCache) ServicesFor(ref api.ResourceReference) []api.CatalogServi
...
 func (r *GatewayCache) FetchServicesFor(ctx context.Context, ref api.ResourceReference) ([]api.CatalogService, error) {
-	client, err := consul.NewClientFromConnMgr(r.config, r.serverMgr)
+	client, err := consul.NewClientFromConnMgr(r.config.ConsulClientConfig, r.serverMgr)
 	if err != nil {
 		return nil, err
 	}
 
 	opts := &api.QueryOptions{}
-	if ref.Namespace != "" {
+	if r.config.NamespacesEnabled && ref.Namespace != "" {
```
nathancoleman, line 62: Just being extra careful when we go to set the namespace on the query

control-plane/api-gateway/common/translation.go
```diff
@@ -50,7 +50,7 @@ func (t ResourceTranslator) NormalizedResourceReference(kind, namespace string,
 }
 
 func (t ResourceTranslator) Namespace(namespace string) string {
-	return namespaces.ConsulNamespace(namespace, t.EnableK8sMirroring, t.ConsulDestNamespace, t.EnableK8sMirroring, t.MirroringPrefix)
+	return namespaces.ConsulNamespace(namespace, t.EnableConsulNamespaces, t.ConsulDestNamespace, t.EnableK8sMirroring, t.MirroringPrefix)
```
nathancoleman, line 53: This was just an oversight - compare to the function signature for `ConsulNamespace` and the added test coverage to verify correctness

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -212,7 +212,7 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		r.gatewayCache.RemoveSubscription(nonNormalizedConsulKey)
 		// make sure we have deregister all services even if they haven't
 		// hit cache yet
-		if err := r.deregisterAllServices(ctx, consulKey); err != nil {
+		if err := r.deregisterAllServices(ctx, nonNormalizedConsulKey); err != nil {
```
nathancoleman, line 215: We use the `nonNormalizedConsulKey` everywhere in this section of code because it doesn't convert the `""` namespace to `"default"` and result in us setting a namespace on the query. Between this and the additional guard above, we're doubly careful that we aren't setting the namespace on a query unless
1. namespaces are enabled
2. the key has a namespace present

nathancoleman (commented):
Personal review

sarahalsmiller (approved):
Thanks for jumping on this!

t-eckert (approved):
Excellent and well-tested 🏅

PR #2377, opened 2023-06-14
api-gateway: fix cache and service deletion issue

t-eckert:
Changes proposed in this PR:
- Fixes a bug where deleting a Pod would not delete the corresponding service for a Gateway.

🐞 *Bug*
When a user would manually delete a Pod running Consul Dataplane for a gateway, the corresponding service in Consul was not getting deleted. This was due to a caching issue where the controller was not updating its store of active Consul services to diff against existing Consul Dataplane pods.

How I've tested this PR:
- Ran the test locally in Kind, manually deleting Pod and ensuring that stale services don't hang around.
- Ran the tests against all acceptance tests.
- Ran the test against a new acceptance test that catches this error (SO to @sarahalsmiller)

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/api-gateway/api_gateway_gatewayclassconfig_test.go
+++ b/acceptance/tests/api-gateway/api_gateway_gatewayclassconfig_test.go
@@ -1,0 +1,208 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package apigateway
+
+import (
+	"context"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/consul"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
+	"github.com/hashicorp/consul-k8s/control-plane/api-gateway/common"
+	"github.com/hashicorp/consul-k8s/control-plane/api/v1alpha1"
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+	"testing"
+)
+
+// GatewayClassConfig tests the creation of a gatewayclassconfig object and makes sure that its configuration
+// is properly applied to any child gateway objects, namely that the number of gateway instances match the defined
+// minInstances,maxInstances and defaultInstances parameters, and that changing the parent gateway does not affect
+// the child gateways.
+func TestAPIGateway_GatewayClassConfig(t *testing.T) {
+	ctx := suite.Environment().DefaultContext(t)
+	cfg := suite.Config()
+	helmValues := map[string]string{
+		"global.logLevel":       "trace",
+		"connectInject.enabled": "true",
+	}
+	releaseName := helpers.RandomName()
+	consulCluster := consul.NewHelmCluster(t, helmValues, ctx, cfg, releaseName)
+	consulCluster.Create(t)
+	// Override the default proxy config settings for this test.
+	consulClient, _ := consulCluster.SetupConsulClient(t, false)
+	_, _, err := consulClient.ConfigEntries().Set(&api.ProxyConfigEntry{
+		Kind: api.ProxyDefaults,
+		Name: api.ProxyConfigGlobal,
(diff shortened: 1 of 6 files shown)


control-plane/api-gateway/binding/binder.go
```diff
@@ -110,6 +110,23 @@ func (b *Binder) Snapshot() *Snapshot {
 		}
 	}
```
nathancoleman, line 112:
```suggestion

// Loop through the list of Consul services and deregister any instances
// that have had their corresponding Pod deleted
```

control-plane/api-gateway/binding/binder.go
```diff
@@ -110,6 +110,25 @@ func (b *Binder) Snapshot() *Snapshot {
...
 
+// Loop through the list of Consul services and deregister any instances
+// that have had their corresponding Pod deleted
+	for _, service := range b.config.ConsulGatewayServices {
```
nathancoleman, line 115:
```suggestion
	// Loop through the list of Consul services and deregister any instances
	// that have had their corresponding Pod deleted
	for _, service := range b.config.ConsulGatewayServices {
```

acceptance/tests/api-gateway/api_gateway_gatewayclassconfig_test.go
```diff
@@ -0,0 +1,210 @@
...
+		var deployment appsv1.Deployment
+		err := client.Get(context.Background(), types.NamespacedName{Name: name, Namespace: namespace}, &deployment)
+		require.NoError(r, err)
+
+		deployment.Spec.Replicas = scaleTo
+		err = client.Update(context.Background(), &deployment)
+		require.NoError(r, err)
+
+		//time.Sleep(time.Minute * 100)
+	})
```
nathancoleman, line 186:
```suggestion
		require.NoError(r, err)
	})
```

acceptance/tests/api-gateway/api_gateway_gatewayclassconfig_test.go
```diff
@@ -0,0 +1,210 @@
...
+		//then check to make sure the number of gateway pods matches the replicas generated
+		podList := corev1.PodList{}
+		labels := common.LabelsForGateway(gateway)
+		err = k8client.List(context.Background(), &podList, client.InNamespace(namespace), client.MatchingLabels(labels))
+		require.NoError(r, err)
+		require.EqualValues(r, *wantNumber, len(podList.Items))
+
+		service, _, err := consulClient.Catalog().Service(name, "", nil)
+		require.NoError(r, err)
+		require.EqualValues(r, *wantNumber, len(service))
```
nathancoleman, line 208:
```suggestion
		services, _, err := consulClient.Catalog().Service(name, "", nil)
		require.NoError(r, err)
		require.EqualValues(r, *wantNumber, len(services))
```

sarahalsmiller (commented):
Assuming tests pass, looks good to me, but I did help pair on this. 👍

sarahalsmiller (approved):

control-plane/api-gateway/cache/gateway.go
```diff
@@ -14,16 +14,13 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/event"
 )
 
 type GatewayCache struct {
 	config    Config
 	serverMgr consul.ServerConnectionManager
 	logger    logr.Logger
 
-	events chan event.GenericEvent
```
nathancoleman, line 25: Confirmed that this channel isn't being used anywhere. It looks like it was a copy-paste leftover in a previous PR

control-plane/api-gateway/cache/gateway.go
```diff
@@ -140,18 +136,5 @@ func (r *GatewayCache) subscribeToGateway(ctx context.Context, ref api.ResourceR
...
-		}
-
-		select {
-		case <-ctx.Done():
-			r.dataMutex.Lock()
-			delete(r.data, ref)
-			r.dataMutex.Unlock()
-			return
-		case r.events <- event:
-		}
```
nathancoleman, line 155: Solidifying my own understanding by writing it out:

Since the `r.events` channel is not buffered and nothing was reading from the channel, the first run through this code would block on the write. The context also was still open, so `ctx.Done()` wasn't happening either.

As a result, we would never move past this `select` to the next iteration of our `for` loop to refresh our cache, so it would very quickly become stale. Removing this code allows our loop to continue executing just as intended so that our cache is updated regularly.

nathancoleman (approved):
Nice work team! You all killed it!

control-plane/api-gateway/cache/gateway.go
andrewstucki: Commenting here so it's less likely to get lost. Removing the channel writing code seems fine to me, I believe it was originally there to re-trigger reconciliation if a derived Consul service was changed (like if someone manually deregistered the service for some odd reason, in which case we'd just re-register it), but clearly I forgot to hook it into our `Watch` routines 😬 

It's probably "more correct" to do it with the Watch, but that seems to be a small enough edge-case to me where I think dropping this here is fine. If we need to add it back in later though, we can.

PR #2378, opened 2023-06-14
Acceptance test for permissive mTLS

pglass:

This adds a basic acceptance test for permissive mTLS that tests:

* An inbound non-mTLS connection is rejected in strict mode
* An inbound non-mTLS connection is accepted in permissive mode

**How I've tested this PR:**

* Ran it locally with kind

👀 

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/acceptance/tests/connect/permissive_mtls_test.go
+++ b/acceptance/tests/connect/permissive_mtls_test.go
@@ -1,0 +1,89 @@
+package connect
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/consul-k8s/acceptance/framework/config"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/connhelper"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/consul"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestConnectInject_PermissiveMTLS(t *testing.T) {
+	cfg := suite.Config()
+	if !cfg.EnableTransparentProxy {
+		t.Skipf("skipping this because -enable-transparent-proxy is not set")
+	}
+
+	ctx := suite.Environment().DefaultContext(t)
+
+	releaseName := helpers.RandomName()
+	connHelper := connhelper.ConnectHelper{
+		ClusterKind: consul.Helm,
+		Secure:      true,
+		ReleaseName: releaseName,
+		Ctx:         ctx,
+		Cfg:         cfg,
+	}
+	connHelper.Setup(t)
+	connHelper.Install(t)
+
+	deployNonMeshClient(t, connHelper)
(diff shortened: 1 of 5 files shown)


acceptance/tests/connect/permissive_mtls_test.go
```diff
@@ -0,0 +1,94 @@
...
+func writeCrd(t *testing.T, ch connhelper.ConnectHelper, path string) {
+	t.Helper()
+
+	t.Cleanup(func() {
+		_, _ = k8s.RunKubectlAndGetOutputE(t, ch.Ctx.KubectlOptions(t), "delete", "-f", path)
+	})
+
+	_, err := k8s.RunKubectlAndGetOutputE(t, ch.Ctx.KubectlOptions(t), "apply", "-f", path)
+	require.NoError(t, err)
+}
```
pglass, line 79: Most acceptance tests do not write config entries using CRDs, I notice. But I kind of wanted to exercise the CRDs.

acceptance/tests/connect/permissive_mtls_test.go
```diff
@@ -0,0 +1,94 @@
...
+	logger.Log(t, "Check that incoming mTLS connection is successful in MutualTLSMode = permissive")
+	k8s.CheckStaticServerConnectionSuccessful(t, kubectlOpts, "static-client", "http://static-server")
+}
+
+func deployNonMeshClient(t *testing.T, ch connhelper.ConnectHelper) {
+	t.Helper()
+
+	logger.Log(t, "Creating static-client deployment with connect-inject=false")
+	k8s.DeployKustomize(t, ch.Ctx.KubectlOptions(t), ch.Cfg.NoCleanupOnFailure, ch.Cfg.DebugDirectory, "../fixtures/bases/static-client")
+	requirePodContainers(t, ch, "app=static-client", 1)
```
pglass, line 59: `DeployKustomize` does a wait, so checking the number pod containers is probably not necessary. But this was a useful sanity check to make sure the static-client does not have a sidecar proxy container, and that the static-server does

``` 
command.go:100: 2023-06-14T17:34:21-05:00 Running command kubectl with args [--context kind-dc1 --namespace default wait --for=condition=available --timeout=5m deploy/static-server]
```

acceptance/tests/connect/permissive_mtls_test.go
```diff
@@ -0,0 +1,94 @@
...
+	logger.Log(t, "Creating static-server deployment with connect-inject=true")
+	k8s.DeployKustomize(t, ch.Ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-server-inject")
+	requirePodContainers(t, ch, "app=static-server", 2)
+}
+
+func writeCrd(t *testing.T, ch connhelper.ConnectHelper, path string) {
+	t.Helper()
+
+	t.Cleanup(func() {
+		_, _ = k8s.RunKubectlAndGetOutputE(t, ch.Ctx.KubectlOptions(t), "delete", "-f", path)
```
cthain, line 74: :thinking: Just thinking through this. So this is going to add each file to the cleanup stack and call delete for the `SeviceDefaults` on `static-server` multiple times, correct? This seems OK since this op is idempotent anyway.

acceptance/tests/connect/permissive_mtls_test.go
cthain: Nice, I think this is a great idea :+1:

cthain (approved):
This looks great. Nice work Paul :rocket:

PR #2392, opened 2023-06-19
always update acl policy if it exists

aahel:
Changes proposed in this PR:
- removed c.flagEnableNamespaces || c.flagSyncCatalog check to always allow  acl update

Context :- 
Between 0.49.0 and 1.0.x: we don't update ACL policies unless namespaces or sync catalog is enabled but between those versions, we merged controller and connect-inject. This meant we added required ACL rules to connect inject but those don't actually get updated and so now the controller can't do things like create intention CRDs.

How I've tested this PR:
  - deploy 0.49.x with connect inject enabled and acls enabled
  - Get the acl policy for connect injector and save it locally
  - build a control plane image and set imageK8s  with it.
  - upgrade to 1.0.x with your changes
  - Check the acl policy again and check that it has updated rules as opposed to the one from step 2
  

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/control-plane/subcommand/server-acl-init/create_or_update.go
+++ b/control-plane/subcommand/server-acl-init/create_or_update.go
@@ -315,42 +315,37 @@ func (c *Command) createOrUpdateACLPolicy(policy api.ACLPolicy, consulClient *ap
 	// Allowing the Consul node name to be configurable also requires any sync
 	// policy to be updated in case the node name has changed.
 	if isPolicyExistsErr(err, policy.Name) {
-		if c.flagEnableNamespaces || c.flagSyncCatalog {
-			c.log.Info(fmt.Sprintf("Policy %q already exists, updating", policy.Name))
+		c.log.Info(fmt.Sprintf("Policy %q already exists, updating", policy.Name))
 
-			// The policy ID is required in any PolicyUpdate call, so first we need to
-			// get the existing policy to extract its ID.
-			existingPolicies, _, err := consulClient.ACL().PolicyList(&api.QueryOptions{})
-			if err != nil {
-				return err
-			}
-
-			// Find the policy that matches our name and description
-			// and that's the ID we need
-			for _, existingPolicy := range existingPolicies {
-				if existingPolicy.Name == policy.Name && existingPolicy.Description == policy.Description {
-					policy.ID = existingPolicy.ID
-				}
-			}
-
-			// This shouldn't happen, because we're looking for a policy
-			// only after we've hit a `Policy already exists` error.
-			// The only time it might happen is if a user has manually created a policy
-			// with this name but used a different description. In this case,
-			// we don't want to overwrite the policy so we just error.
-			if policy.ID == "" {
-				return fmt.Errorf("policy found with name %q but not with expected description %q; "+
-					"if this policy was created manually it must be renamed to something else because this name is reserved by consul-k8s",
-					policy.Name, policy.Description)
-			}
-
-			// Update the policy now that we've found its ID
-			_, _, err = consulClient.ACL().PolicyUpdate(&policy, &api.WriteOptions{})
+		// The policy ID is required in any PolicyUpdate call, so first we need to
+		// get the existing policy to extract its ID.
+		existingPolicies, _, err := consulClient.ACL().PolicyList(&api.QueryOptions{})
+		if err != nil {
 			return err
-		} else {
-			c.log.Info(fmt.Sprintf("Policy %q already exists, skipping update", policy.Name))
-			return nil
 		}
+
+		// Find the policy that matches our name and description
+		// and that's the ID we need
+		for _, existingPolicy := range existingPolicies {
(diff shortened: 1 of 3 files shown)


david-yu:
Thanks we'll have someone also review from the service mesh team.

control-plane/subcommand/server-acl-init/create_or_update.go
```diff
@@ -315,42 +315,37 @@ func (c *Command) createOrUpdateACLPolicy(policy api.ACLPolicy, consulClient *ap
...
+			if existingPolicy.Name == policy.Name && existingPolicy.Description == policy.Description {
+				policy.ID = existingPolicy.ID
 			}
+		}
 
-			// Update the policy now that we've found its ID
-			_, _, err = consulClient.ACL().PolicyUpdate(&policy, &api.WriteOptions{})
-			return err
-		} else {
-			c.log.Info(fmt.Sprintf("Policy %q already exists, skipping update", policy.Name))
```
wilkermichael, line 351: Do we want to get rid of this log?

aahel: we are not skipping update now. Previously if c.flagEnableNamespaces || c.flagSyncCatalog were not enabled we would skip update.

aahel:
> Thanks for making this change, can you just verify that you tested the upgrade scenario as well? Add that to the "How I've tested this PR" section.
> 
> Thanks!

thnx for pointing this out. Can you suggest some way  testing this upgrade ? i am not exactly sure when the acls will be updated.

control-plane/subcommand/server-acl-init/create_or_update.go
wilkermichael: ah yep, I misread the brackets. All good then!

wilkermichael:
>

ishustava (changes requested):
This looks good but we need to add unit tests for this change.

aahel:
> This looks good but we need to add unit tests for this change.

Have added a couple of unit test cases. 
i haven't been able to manually verify the upgrade from 0.49.x to 1.0.x.

control-plane/subcommand/server-acl-init/create_or_update_test.go
```diff
@@ -70,3 +70,95 @@ func TestCreateOrUpdateACLPolicy_ErrorsIfDescriptionDoesNotMatch(t *testing.T) {
...
+				Rules:       connectInjectRule,
+			},
+		},
+		{
+			Name:              "update",
+			ID:                policy.ID,
+			PolicyDescription: policy.Description,
+			PolicyName:        policy.Name,
+			Rules:             aclReplRule,
+			Err:               nil,
```
ishustava, line 139: `err` is `nil` for both cases so we can omit it as a test parameter.

ishustava (commented):
This mostly looks good. I left a couple of comments about the test

control-plane/subcommand/server-acl-init/create_or_update_test.go
```diff
@@ -70,3 +70,95 @@ func TestCreateOrUpdateACLPolicy_ErrorsIfDescriptionDoesNotMatch(t *testing.T) {
...
+	}
+	for _, tt := range cases {
+		t.Run(tt.Name, func(t *testing.T) {
+			err = cmd.createOrUpdateACLPolicy(api.ACLPolicy{
+				Name:        tt.PolicyName,
+				Description: tt.PolicyDescription,
+				Rules:       tt.Rules,
+			}, consul)
+			require.Equal(tt.Err, err)
+			if tt.ID != "" {
```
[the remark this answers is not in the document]
aahel: createOrUpdateACLPolicy dosent return the id.  so i was just checking the error.
one way is to get all policies match them by name and description to find the created policy. should i do this ?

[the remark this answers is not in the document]
ishustava: You can use `client.ACL().PolicyReadByName(...)` for both cases

control-plane/subcommand/server-acl-init/create_or_update_test.go
aahel: had kept this as there we can later add additional cases for the error flows( ex:- if updatePolicy/createPolicy returns err).
but we will have to mock acl. which will require code changes. 
 should i remove it ?

ishustava: Yes please remove. We can add it when we're adding those test cases in the future.

ishustava (approved):

aahel:
> Thanks for making this change, can you just verify that you tested the upgrade scenario as well? Add that to the "How I've tested this PR" section.
> 
> Thanks!

done

wilkermichael:
Looks good to me, thank you for testing the upgrade @aahel !

wilkermichael (approved):

PR #2411, opened 2023-06-20
Mw/net 4622 release 0.49.x acceptance tests broken missing packages

wilkermichael:
Changes proposed in this PR:
- This PR adds a configurable acceptance test matrix to the repo that can be fetched via the following make targets ( depending on the test):
  - `make kind-test-packages`
  - `make aks-test-packages` 
  -  `make eks-test-packages`
  - `make gke-test-packages`
- Each branch can now have their own defined list of acceptance tests allowing branches to have different packages 

How I've tested this PR:
- Ran against this workflow: https://github.com/hashicorp/consul-k8s-workflows/pull/19
- Will point the pr.yaml file back at main after review
- I wasn't able to test the cloud tests because cloud test security does not allow for point tests to any other branch than main

👀 

Checklist:
- [x] Tests added
- [n/a] CHANGELOG entry added 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,17 @@ kind-node-image:
 kubectl-version:
 	@echo $(KUBECTL_VERSION)
 
+kind-test-packages:
+	@./control-plane/build-support/scripts/set_test_package_matrix.sh "acceptance/ci-inputs/kind_acceptance_test_packages.yaml"
+
+gke-test-packages:
+	@./control-plane/build-support/scripts/set_test_package_matrix.sh "acceptance/ci-inputs/gke_acceptance_test_packages.yaml"
+
+eks-test-packages:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -197,6 +197,17 @@ kind-node-image:
 kubectl-version:
 	@echo $(KUBECTL_VERSION)
 
+kind-test-packages:
+	@./control-plane/build-support/scripts/set_test_package_matrix.sh "acceptance/ci-inputs/kind_acceptance_test_packages.yaml"
```
wilkermichael, line 201: I had to go against putting this in a variable at the top of the file like the others because of the json output. When you put it in the variable at the top it does weird things with the `""` making it invalid JSON

thisisnotashwin (approved):

curtbushko (approved):
LGTM!

PR #2412, opened 2023-06-21
Adds back gateway controller integration test

missylbytes:
Changes proposed in this PR:
- Adds back tests from https://github.com/hashicorp/consul-k8s/pull/2305 that started failing after the merge and was reverted here (https://github.com/hashicorp/consul-k8s/pull/2341).

How I've tested this PR:
Ran the test.

Run the test!

Checklist:
- [x] Tests added

### DIFF
--- a/control-plane/api-gateway/binding/route_binding.go
+++ b/control-plane/api-gateway/binding/route_binding.go
@@ -457,7 +457,7 @@ func consulCondition(generation int64, status api.ConfigEntryStatus) *metav1.Con
 	for _, c := range status.Conditions {
 		// we only care about the top-level status that isn't in reference
 		// to a resource.
-		if c.Type == "Accepted" && c.Resource == nil { //c.Resource.Name == "" {
+		if c.Type == "Accepted" && c.Resource == nil || c.Resource.Name == "" {
 			return &metav1.Condition{
 				Type:               "ConsulAccepted",
 				Reason:             c.Reason,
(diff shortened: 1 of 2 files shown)


control-plane/api-gateway/binding/route_binding.go
```diff
@@ -457,7 +457,7 @@ func consulCondition(generation int64, status api.ConfigEntryStatus) *metav1.Con
 	for _, c := range status.Conditions {
 		// we only care about the top-level status that isn't in reference
 		// to a resource.
-		if c.Type == "Accepted" && c.Resource.Name == "" {
+		if c.Type == "Accepted" && c.Resource == nil { //c.Resource.Name == "" {
```
missylbytes, line 460: Need to review this

missylbytes: I think this accomplishes the same result, we are still checking the highest level status, and it looks like that would not have a resource (because the resource is the gateway)

missylbytes:
- [ ] Add back John in commit message

nathancoleman (approved):
LGTM, acknowledging that @ishustava said we're safe to ignore the acceptance-cni failures

nathancoleman:
It seems that the tests in this PR are failing on the `main` branch after merge ([example](https://github.com/hashicorp/consul-k8s-workflows/actions/runs/5403748799))

PR #2413, opened 2023-06-21
NET-4638: when multiple listeners have the same port, only add to K8s Service once

nathancoleman:
A valid `Gateway` can have multiple listeners with the same port. When this happens, we loop through the listeners and add each listener's port to the list of ports for the corresponding Kubernetes `Service`, resulting in an invalid resource that can't be created.

This change just tracks which ports have already been added so that we only add each port once.

**How I've tested this PR:**
🤖 tests pass

🤖 tests pass

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/control-plane/api-gateway/gatekeeper/gatekeeper_test.go
+++ b/control-plane/api-gateway/gatekeeper/gatekeeper_test.go
@@ -40,11 +40,18 @@ var (
 			Name:     "Listener 1",
 			Port:     8080,
 			Protocol: "TCP",
+			Hostname: common.PointerTo(gwv1beta1.Hostname("example.com")),
 		},
 		{
 			Name:     "Listener 2",
 			Port:     8081,
 			Protocol: "TCP",
+		},
+		{
+			Name:     "Listener 3",
+			Port:     8080,
+			Protocol: "TCP",
+			Hostname: common.PointerTo(gwv1beta1.Hostname("example.net")),
 		},
 	}
 )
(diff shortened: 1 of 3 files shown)


control-plane/api-gateway/gatekeeper/gatekeeper_test.go
```diff
@@ -40,12 +40,19 @@ var (
...
 			Name:     "Listener 2",
 			Port:     8081,
 			Protocol: "TCP",
 		},
+		{
+			Name:     "Listener 3",
+			Port:     8080,
+			Protocol: "TCP",
+			Hostname: common.PointerTo(gwv1beta1.Hostname("example.net")),
+		},
```
nathancoleman, line 55: We have multiple asserts on the `Service` being created. Adding a listener here with a duplicate port makes these tests more robust by ensuring our existing assertions pass.

control-plane/api-gateway/gatekeeper/service.go
```diff
@@ -65,14 +65,22 @@ func (g *Gatekeeper) deleteService(ctx context.Context, gwName types.NamespacedN
 }
 
 func (g *Gatekeeper) service(gateway gwv1beta1.Gateway, gcc v1alpha1.GatewayClassConfig) *corev1.Service {
+	seenPorts := map[gwv1beta1.PortNumber]struct{}{}
 	ports := []corev1.ServicePort{}
 	for _, listener := range gateway.Spec.Listeners {
+		if _, seen := seenPorts[listener.Port]; seen {
+			// We've already added this listener's port to the Service
```
sarahalsmiller, line 72: Should this be throwing some kind of error? It seems to me like its a configuration issue if multiple listeners are defined on the same port.

nathancoleman: The Gateway API spec makes a specific allowance for multiple listeners to use the same port ([source](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Listener)):

> Port is the network port. Multiple listeners may use the same port, subject to the Listener compatibility rules.

sarahalsmiller (approved):

PR #2415, opened 2023-06-21
Revert "added imagePullPolicy for images in values.yaml (#2310)"

thisisnotashwin:
This reverts commit 285096241e0d5c5b6d53dd8a37889ab3ea5a8af2.

Changes proposed in this PR:
- The above PR missed some places to add support for the imagePullPolicy leading to inconsistent behavior across the helm chart. This change will be reintroduced in a future PR once we have implemented it completely.


curtbushko (approved):
Thanks!

PR #2416, opened 2023-06-21
feat(helm): add configurable server-acl-init and cleanup resource limits

DanStough:
Changes proposed in this PR:
- Exposes resource configuration for `server-acl-init` and `sercer-acl-init-cleanup` Jobs as helm values with existing defaults. 

Checklist:
- [X] Tests added
- [X] CHANGELOG entry added 

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -436,6 +436,33 @@ global:
       # @type: string
       secretKey: null
 
+    # The resource requests (CPU, memory, etc.) for the server-acl-init and server-acl-init-cleanup pods. 
+    # This should be a YAML map corresponding to a Kubernetes
+    # [`ResourceRequirements``](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#resourcerequirements-v1-core)
+    # object.
+    #
+    # Example:
+    #
+    # ```yaml
+    # resources:
+    #   requests:
+    #     memory: '200Mi'
+    #     cpu: '100m'
+    #   limits:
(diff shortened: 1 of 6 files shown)


charts/consul/values.yaml
```diff
@@ -436,6 +436,33 @@ global:
...
+    #
+    # @recurse: false
+    # @type: map
+    resources:
+      requests:
+        memory: "50Mi"
+        cpu: "50m"
+      limits:
+        memory: "50Mi"
+        cpu: "50m"
```
thisisnotashwin, line 464: Are these what the values defaulted to in the absence of us providing them? Just wanted to know if we might be concerned this is not enough, but otherwise ignore this!!

thisisnotashwin (commented):
This looks great! found some copypasta issues and have another minor comment but otherwise, this is wonderful!

charts/consul/values.yaml
DanStough: Yes, these were the previous hard-coded values. As to whether they are enough, the practitioner reported that they get OOM because they have a large number (50+) of secondary clusters. Since it's just a job, we could double everything, and I don't know if most folks would mind, but this is also an extreme use case. I would defer to you if you have a strong opinion. WDYT @thisisnotashwin ?

thisisnotashwin: nah! this rationale works for me! thank you for all this information!! i really appreciate it!!

thisisnotashwin (approved):
magnifique!

PR #2420, opened 2023-06-21
NET-4482: set route condition appropriately when parent ref includes non-existent section

nathancoleman:
When a `HTTPRoute` declares a parent `Gateway`, it's allowed to specify a specific listener on that `Gateway` to attach to. If a listener doesn't exist matching that parent reference, then we need to set the `HTTPRoute` acceptance condition appropriately.

**How I've tested this PR:**
- Added unit tests pass
- Previously failing conformance test now passes
CleanShot 2023-06-21 at 18 07 15@2x

See above

**Checklist:**
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/control-plane/api-gateway/binding/result.go
+++ b/control-plane/api-gateway/binding/result.go
@@ -34,6 +34,7 @@ var (
 	errRouteNoMatchingListenerHostname      = errors.New("listener cannot bind route with a non-aligned hostname")
 	errRouteInvalidKind                     = errors.New("invalid backend kind")
 	errRouteBackendNotFound                 = errors.New("backend not found")
+	errRouteNoMatchingListenerName          = errors.New("no listener found with matching name")
 )
 
 // routeValidationResult holds the result of validating a route globally, in other
@@ -171,13 +172,16 @@ func (b bindResults) Condition() metav1.Condition {
 	// if we only have a single binding error, we can get more specific
 	if len(b) == 1 {
 		for _, result := range b {
-			// if we have a hostname mismatch error, then use the more specific reason
-			if result.err == errRouteNoMatchingListenerHostname {
+			switch result.err {
+			case errRouteNoMatchingListenerHostname:
+				// if we have a hostname mismatch error, then use the more specific reason
 				reason = "NoMatchingListenerHostname"
-			}
-			// or if we have a ref not permitted, then use that
-			if result.err == errRefNotPermitted {
+			case errRefNotPermitted:
+				// or if we have a ref not permitted, then use that
 				reason = "RefNotPermitted"
+			case errRouteNoMatchingListenerName:
+				// or if the route declares a parent with a section name that we can't find
+				reason = "NoMatchingParent"
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


control-plane/api-gateway/binding/result.go
```diff
@@ -128,13 +129,17 @@ type bindResult struct {
...
 func (b bindResults) Error() string {
 	messages := []string{}
 	for _, result := range b {
 		if result.err != nil {
-			messages = append(messages, fmt.Sprintf("%s: %s", result.section, result.err.Error()))
+			message := result.err.Error()
+			if result.section != "" {
+				message = fmt.Sprintf("%s: %s", result.section, result.err.Error())
+			}
+			messages = append(messages, message)
```
nathancoleman, line 142: It seems reasonable that some errors might not apply to a single listener or a listener name might not have been specified, in which case it was weird to include `: this is my error` for the message on the condition instead of `this is my error`

control-plane/api-gateway/binding/result.go
```diff
@@ -171,13 +176,16 @@ func (b bindResults) Condition() metav1.Condition {
...
 				reason = "NoMatchingListenerHostname"
-			}
-			// or if we have a ref not permitted, then use that
-			if result.err == errRefNotPermitted {
+			case errRefNotPermitted:
+				// or if we have a ref not permitted, then use that
 				reason = "RefNotPermitted"
+			case errRouteNoMatchingParent:
+				// or if the route declares a parent that we can't find
+				reason = "NoMatchingParent"
```
nathancoleman, line 188: This is part 2 of the core change where we consume the error that we detected while doing validation in order to construct the correct condition

control-plane/api-gateway/binding/route_binding.go
```diff
@@ -104,7 +104,22 @@ func (r *Binder) bindRoute(route client.Object, boundCount map[gwv1beta1.Section
...
+			var sectionName gwv1beta1.SectionName
+			if ref.SectionName != nil {
+				sectionName = *ref.SectionName
+			}
+
+			result = append(result, bindResult{
+				section: sectionName,
+				err:     errRouteNoMatchingParent,
+			})
+		}
```
nathancoleman, line 120: This is part 1 of the core change where we detect the specific case while doing validation in order to construct the correct condition later on

nathancoleman (commented):
Personal review

sarahalsmiller (approved):
Code logic looks fine to me, but I asked a semantics question in slack for my own understanding.

nathancoleman:
Since it may be useful for others:

> **Sarah Alsmiller**
> I'm a bit confused on some semantics, I thought that the routes parents were typically the gateway, but this is specifically checking for listeners.
> 
> 
> **Nathan Coleman**
> A route can bind to all listeners on a Gateway by omitting a sectionName  on the parent ref
>
> or it can bind to a specific listener by specifying a sectionName on the parent ref, where the section name should match one of the listeners on the named gateway
>
> so this is specifically handling the case where I try to attach to the listener named abc on a Gateway, but the Gateway does not have a listener with that name

PR #2470, opened 2023-06-28
ci: fix eks terraform quota error by cleaning up oidc providers

ndhanushkodi:
Changes proposed in this PR:
- cleans up oidc providers associated with our pipeline terraform

How I've tested this PR:
- ran it locally and it deleted the oidc providers

- 👀

### DIFF
--- a/hack/aws-acceptance-test-cleanup/main.go
+++ b/hack/aws-acceptance-test-cleanup/main.go
@@ -39,6 +39,11 @@ var (
 	errNotDestroyed = errors.New("not yet destroyed")
 )
 
+type oidc struct {
+	arn      string
+	buildUrl string
+}
+
 func main() {
 	flag.BoolVar(&flagAutoApprove, "auto-approve", false, "Skip interactive approval before destroying.")
 	flag.Parse()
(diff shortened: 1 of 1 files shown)


hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -68,6 +69,94 @@ func realMain(ctx context.Context) error {
 	eksClient := eks.New(clientSession, awsCfg)
 	ec2Client := ec2.New(clientSession, awsCfg)
 	elbClient := elb.New(clientSession, awsCfg)
+	iamClient := iam.New(clientSession, awsCfg)
+
+	// Find OIDC providers to delete.
+	oidcProvidersOutput, err := iamClient.ListOpenIDConnectProvidersWithContext(ctx, &iam.ListOpenIDConnectProvidersInput{})
```
wilkermichael, line 75: ListOpenIDConnectProvidersWithContext can return a nil value, so maybe we should check for that here since you access a member later

wilkermichael: also should check the error

wilkermichael (approved):

hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -68,6 +69,94 @@ func realMain(ctx context.Context) error {
...
+			text string
+			err  error
+		}
+		inputCh := make(chan input)
+
+		// Read input in a goroutine so we can also exit if we get a Ctrl-C
+		// (see select{} below).
+		go func() {
+			reader := bufio.NewReader(os.Stdin)
+			fmt.Println("\nDo you want to delete these OIDC Providers (y/n)?")
```
hashi-derek, line 127: It looks like this will be a manually run job instead of an automated one? Also, should we consider putting in a check and only deleting things older than 24 hours so we don't break tests that are currently in-flight?

wilkermichael: I'm not 100% on the history of why we have this "manual" looking step, but it's also in the other cleanup functions as well, so looks like @ndhanushkodi just used the same pattern that's established in the repo.

ndhanushkodi: The approval step is for running locally, so you can see all the things you're about to delete (and its tags) and approve the deletion. I added a separate approval step for this because IAM resources are separate from VPCs. The approval step is only prompted if you don't set --auto-approve, which our pipelines do when they run the script. 

Yeah the automated pipeline step should be safe to not interfere with the running job, but I also went ahead a timestamp check here anyways so if you run it manually you wouldn't delete the most recent oidc.

hashi-derek: Awesome, thank you.

hashi-derek (approved):

hack/aws-acceptance-test-cleanup/main.go
ndhanushkodi: nice catch! thanks!

PR #2478, opened 2023-06-28
APIGW: Validate length of RSA Keys

jm96441n:
Changes proposed in this PR:
- Ensure RSA keys are of the correct length when creating an inline cert

How I've tested this PR:
- unit tests

- read the code

Checklist:
- [X] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/2478.txt
+++ b/.changelog/2478.txt
@@ -1,0 +1,5 @@
+```release-note:bug
+api-gateway: fixes bug where envoy will silently reject RSA keys less than 2048 bits in length when not in FIPS mode, and
+will reject keys that are not of length 2048, 3072, or 4096 bits in length in FIPS mode. We now validate
+and reject invalid certs earlier.
+```
--- a/.changelog/2478.txt
+++ b/.changelog/2478.txt
@@ -1,5 +1,5 @@
 ```release-note:bug
 api-gateway: fixes bug where envoy will silently reject RSA keys less than 2048 bits in length when not in FIPS mode, and
-will reject keys that are not of length 2048, 3072, or 4096 bits in length in FIPS mode. We now validate
+will reject keys that are not 2048, 3072, or 4096 bits in length in FIPS mode. We now validate
 and reject invalid certs earlier.
 ```
--- a/control-plane/api-gateway/binding/validation.go
+++ b/control-plane/api-gateway/binding/validation.go
@@ -6,7 +6,6 @@ package binding
 import (
 	"crypto/x509"
 	"encoding/pem"
-	"errors"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
@@ -252,7 +251,7 @@ func validateKeyLength(privateKey string) error {
 func nonFipsLenCheck(keyLen int) error {
 	// ensure private key is of the correct length
 	if keyLen < MinKeyLength {
-		return errors.New("key length must be at least 2048 bits")
+		return errListenerInvalidCertificateRef_NonFIPSRSAKeyLen
 	}
 
 	return nil
@@ -260,7 +259,7 @@ func nonFipsLenCheck(keyLen int) error {
 
 func fipsLenCheck(keyLen int) error {
 	if keyLen != 2048 && keyLen != 3072 && keyLen != 4096 {
-		return errors.New("key length invalid: only RSA lengths of 2048, 3072, and 4096 are allowed in FIPS mode")
+		return errListenerInvalidCertificateRef_FIPSRSAKeyLen
 	}
 	return nil
 }
--- a/control-plane/api-gateway/binding/validation.go
+++ b/control-plane/api-gateway/binding/validation.go
@@ -4,11 +4,11 @@
 package binding
 
 import (
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
 	"strings"
 
-	"github.com/hashicorp/consul-k8s/control-plane/api-gateway/common"
-	"github.com/hashicorp/consul-k8s/control-plane/api/v1alpha1"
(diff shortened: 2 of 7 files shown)


.changelog/2478.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:bug
+api-gateway: fixes bug where envoy will silently reject RSA keys less than 2048 bits in length when not in FIPS mode, and
+will reject keys that are not of length 2048, 3072, or 4096 bits in length in FIPS mode. We now validate
```
nathancoleman, line 3:
```suggestion
will reject keys that are not 2048, 3072, or 4096 bits in length in FIPS mode. We now validate
```

.changelog/2478.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:bug
+api-gateway: fixes bug where envoy will silently reject RSA keys less than 2048 bits in length when not in FIPS mode, and
```
nathancoleman, line 2: I don't have a strong opinion but think it's worth asking: is this a bug, or is it an improvement?

control-plane/api-gateway/binding/validation.go
```diff
@@ -17,6 +16,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	gwv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/hashicorp/consul-k8s/control-plane/api-gateway/common"
+	"github.com/hashicorp/consul-k8s/control-plane/api/v1alpha1"
+	"github.com/hashicorp/consul-k8s/control-plane/version"
+	"github.com/hashicorp/consul/api"
```
nathancoleman, line 23: #nit doesn't follow the blocking pattern established in other files in this PR

I'm not sure if we have a consul-wide standard. I was wondering this myself the other day 🤔

control-plane/api-gateway/binding/validation.go
```diff
@@ -204,11 +208,59 @@ func validateTLS(gateway gwv1beta1.Gateway, tls *gwv1beta1.GatewayTLSConfig, res
...
 	if err != nil {
 		return errListenerInvalidCertificateRef_InvalidData
 	}
+
+	err = validateKeyLength(privateKey)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
nathancoleman, line 226: #nit

```suggestion
	return validateKeyLength(privateKey)
```

control-plane/api-gateway/binding/validation.go
```diff
@@ -204,11 +208,59 @@ func validateTLS(gateway gwv1beta1.Gateway, tls *gwv1beta1.GatewayTLSConfig, res
...
+}
+
+func validateKeyLength(privateKey string) error {
+	// we can assume this is non-nil as it would've caused an error from common.ParseCertificate if it was
+	privateKeyBlock, _ := pem.Decode([]byte(privateKey))
+
+	if privateKeyBlock.Type != "RSA PRIVATE KEY" {
```
nathancoleman, line 233: Is there a constant for this in the std library?

jm96441n: not that I've found 
https://pkg.go.dev/crypto/rsa#pkg-constants
https://pkg.go.dev/crypto/x509#pkg-constants

control-plane/api-gateway/binding/validation.go
jm96441n: can do, I usually like to have the error checking be explicit but this works for me

.changelog/2478.txt
jm96441n: I was on the fence here, I went with bug because to end users it would seem like a bug that they provided a certificate but it just wasn't working (unless they knew to dive into the envoy config and parse the output there).

control-plane/api-gateway/binding/validation.go
jm96441n: I can't find the actual docs now, but for some reason I added this to my goimports setup when I started :question: https://github.com/jm96441n/dotfiles/blob/master/.config/nvim/lua/plugins/config/formatter.lua#L16

nathancoleman (approved):

andrewstucki:
One QQ, have you validated that we don't attempt to create the certificate if we have the cert error on the listener? I believe we may have to also update some of the code here: https://github.com/hashicorp/consul-k8s/blob/920ee3233d41792134a02f7a0e9e1d2d5ed4482d/control-plane/api-gateway/binding/binder.go#L161-L164

Specifically as this routine also calls the parsing code: https://github.com/hashicorp/consul-k8s/blob/920ee3233d41792134a02f7a0e9e1d2d5ed4482d/control-plane/api-gateway/common/translation.go#L335-L355

Maybe we can DRY some of this up? If not, think you might just need to add in the same logic both in the translator and in the validator. Either is fine with me for the fix if you can't just lightly refactor to make this more DRY.

jm96441n:
@andrewstucki I haven't validated that portion yet, lemme run it and check on that. (also replying in a separate message here because github apparently doesn't allow responses to comments directly)

andrewstucki:
Sounds good, I'd basically just follow the above steps and then look for something like this in the error logs of the controller:

https://github.com/hashicorp/consul-k8s/blob/920ee3233d41792134a02f7a0e9e1d2d5ed4482d/control-plane/api-gateway/common/resources.go#L561

If you see that log then we should also have a stale `InlineCertificate` with the old Secret data still in Consul.

andrewstucki (approved):
LGTM, thanks for the changes!

PR #2481, opened 2023-06-28
Adding support for Enterprise and other improvement on the Customizing Vault Version for WanFed Test

20sr20:
Changes proposed in this PR:

This is the extension of the PR - 
https://github.com/hashicorp/consul-k8s/pull/2043

In this PR, the followings were addressed -

1. Now the vault enterprise version can be provided in the cli command.  The previous PR only addressed Vault OSS.
2. Two flags “-no-cleanup-wan-fed” and “test-duration” were introduced to not to cleanup the test environment after successful setup to give it time to do manual testing for features/to reproduce customer issues.  Default is 1 hour.

Update on July 14 on this - After the first review, the above flags are removed, instead a new flag "-no-cleanup" was introduced.

3. This was tested in Kind environment and it works fine.  The following was placed back to use the “use-kind” option for WanFed test.

    if cfg.UseKind {
      t.Skipf("Skipping this test because it's currently flaky on kind")
    }

How I've tested this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


acceptance/framework/config/config.go
```diff
@@ -59,6 +59,8 @@ type TestConfig struct {
 	VaultServerVersion    string
 
 	NoCleanupOnFailure bool
+	NoCleanupWanFed    bool
```
curtbushko, line 62: Could you change this from `NoCleanupWanFed` to `NoCleanup`? That way we can use it in other tests to always keep the environment around.

acceptance/framework/vault/vault_cluster.go
```diff
@@ -54,12 +57,32 @@ func NewVaultCluster(t *testing.T, ctx environment.TestContext, cfg *config.Test
...
+	vaultLicenseSecretName := fmt.Sprintf("%s-enterprise-license", vaultReleaseName)
+	vaultLicenseSecretKey := "license"
+
+	vaultEnterpriseLicense := os.Getenv("VAULT_LICENSE")
+
 	if cfg.VaultServerVersion != "" {
+
+		if strings.Contains(cfg.VaultServerVersion, entstr) {
+
+			//logger.Log(t, "Creating secret for Vault license")
```
curtbushko, line 79: I think it is fine to leave this logger message in.

acceptance/tests/vault/vault_wan_fed_test.go
```diff
@@ -31,9 +33,9 @@ import (
 // in the secondary that will treat the Vault server in the primary as an external server.
 func TestVault_WANFederationViaGateways(t *testing.T) {
 	cfg := suite.Config()
-	if cfg.UseKind {
-		t.Skipf("Skipping this test because it's currently flaky on kind")
-	}
+	//if cfg.UseKind {
```
curtbushko, line 36: Reminder: add back?

go.work
```diff
@@ -0,0 +1,12 @@
+go 1.20
```
curtbushko, line 1: I would remove these. We aren't using go.work files as there are problems with it.  I do not remember exactly what the problems were but the team has discussed it before and decided not to use it.

acceptance/tests/vault/vault_wan_fed_test.go
```diff
@@ -517,6 +521,19 @@ func TestVault_WANFederationViaGateways(t *testing.T) {
 
 	logger.Log(t, "checking that connection is successful")
 	k8s.CheckStaticServerConnectionSuccessful(t, primaryCtx.KubectlOptions(t), StaticClientName, "http://localhost:1234")
+
+	if cfg.NoCleanupWanFed {
+		//time.Sleep(1800 * time.Second)
+		intVar, err := strconv.Atoi(cfg.TestDuration)
```
curtbushko, line 527: I don't think you want to use TestDuration as a timeout. Instead:
 
 a) Remove the TestDuration flag.
 b) Add a retry that checks for something successful or several steps that are successful.
 
 If these tests take over 2 hours to run then we will not be able to run them in Github Actions as the runners will timeout...

curtbushko (commented):
Looks good! 

I have left some comments on how to improve this. 

Do you have an idea about what piece is taking so long to complete/setup?  Could some pieces be done in parallel?

acceptance/tests/vault/vault_wan_fed_test.go
kschoche: @curtbushko - I left a PR comment which is sorta in reply to your comments here. If you don't mind maybe have a look at it and see if that makes any sense?

curtbushko:
> I believe the primary purpose of this PR is:
> 
>     * a fix to be able to specify the vault image for enterprise (oss works but ent doesnt because of naming conventions) ✅
> 
>     * add a configurable sleep at the end of a specific test so that a user can manually interact with the clusters already bootstrapped.
> 
> 
> Maybe just the introduction of a `no-cleanup` flag without the sleeps is sufficient enough and clean enough for all of this to work? Then the test can run to completion and the deployments can be left intact so that you would just run something like: `go test -run WANFederationViaGateways -p 1 -timeout 2h -failfast -use-kind -no-cleanup -kubecontext=kind-dc1 -secondary-kubecontext=kind-dc2 -enable-enterprise -enable-multi-cluster -debug-directory=/tmp/debug` ?
> 
> You would need to plumb this into the `acceptance/framework/vault/vault_cluster.go` data structures alongside `noCleanupOnFailure` but it seems potentially cleaner? Here is how it's used:
> 
> https://github.com/hashicorp/consul-k8s/blob/df0e649e95caf9bd9ccf911d7e810c1dd781fa81/acceptance/framework/vault/vault_cluster.go#L227

@kschoche Yes, I think the no-cleanup flag is a much better idea overall and is something that can be used in all environments.

curtbushko (approved):
Tentatively approving once conflicts are resolved and the kind flag check is turned back on in the test.

acceptance/tests/vault/vault_wan_fed_test.go
20sr20: Added back

kschoche (commented):
Looking good, I'm putting a 👍🏽 on the approach.
I see a lot of failures in CI that might be related to the changes, can you take a look and see if you can get them to go green? Sorry I'm not familiar enough with the tests right now to know if they are flaky or if we've introduced a bug on the cleanup path.

andrewstucki:
Hi @20sr20 can you do a follow-up on this? This [block](https://github.com/hashicorp/consul-k8s/pull/2481/files#diff-29adf4fb3f3d19ef876546e19cd2ba8d6d570ebce12925b7a8aba790cb55412aR67-R80) appears to have broken the test as it's trying to create a Kubernetes namespace named `default` (which will always exist):

https://github.com/hashicorp/consul-k8s/blob/9b372a5c11a3cfa32ad8e1fa93ebe5ad287d8554/acceptance/tests/api-gateway/api_gateway_gatewayclassconfig_test.go#L67-L80

david-yu:
I believe this is the follow up: https://github.com/hashicorp/consul-k8s/pull/2631 thanks @pglass !

PR #2487, opened 2023-06-28
api-gateway: Fix nil pointer exception panic

sarahalsmiller:
Changes proposed in this PR:
- Fix nil pointer exception thrown whenever the user uses an httpfilter that does not have RequestHeaderModifier set. 

How I've tested this PR:
- Created unit test that replicated the error and saw that it is resolved with this fix

- Unit tests pass

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.changelog/2487.txt
+++ b/.changelog/2487.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+control-plane: fixed issue where a HTTPRoute would break if a `URLRewrite` filter was defined.
+```
--- a/.changelog/2487.txt
+++ b/.changelog/2487.txt
@@ -1,3 +1,0 @@
-```release-note:bug
-control-plane: fixed issue where a HTTPRoute would break if a `URLRewrite` filter was defined.
-```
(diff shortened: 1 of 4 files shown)


t-eckert (approved):

.changelog/2487.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+control-plane: fixed issue where a HTTPRoute would break if a `URLRewrite` filter was defined.
```
wilkermichael, line 2: This is a new feature for 1.2 correct? If so I don't think we need to include a changelog since the bug was never live

wilkermichael (approved):

.changelog/2487.txt
sarahalsmiller: Deleting!

nathancoleman (approved):

nathancoleman:
The failing test in **Unit test control plane** was added in https://github.com/hashicorp/consul-k8s/pull/2412 and is failing reliably on the `main` branch as well; therefore, it is not attributable to this PR.

PR #2488, opened 2023-06-28
Fix `make consul-enterprise-version` to work for both dev and prod images

t-eckert:
Changes proposed in this PR:
- We run a `make` command to set the enterprise Consul image when running unit tests. For development images the delta from non-enterprise images to enterprise images is to just change the image name from `consul` to `consul-enterprise`. In prod, you must also add a `-ent` to the end of the image tag. This change takes that difference into account so that we can use the `make` command when releasing to production.

### DIFF
--- a/control-plane/build-support/scripts/consul-enterprise-version.sh
+++ b/control-plane/build-support/scripts/consul-enterprise-version.sh
@@ -4,8 +4,11 @@
 FILE=$1
 VERSION=$(yq .global.image $FILE)
 
-if [[ !"${VERSION}" == *"consul:"* ]]; then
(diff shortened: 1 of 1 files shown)


t-eckert:
This change does not need to be backported to branches other than `release/1.2.x` because I already manually cherry-picked it into the release branches in order to get the release out.

zalimeni (approved):
Approving since this is status-quo following the last release.

Curious whether in the future we could look at making the dev + prod tagging consistent so this sort of workaround isn't necessary - but happy to not block on that now.

missylbytes:
This has been covered by https://github.com/hashicorp/consul-k8s/pull/2756

PR #2508, opened 2023-07-03
feature: adding security context and annotations to tls and acl init/cleanup jobs

Chinikins:
Changes proposed in this PR:
- Ability to add annotations to acl init and cleanup jobs
- Ability to add annotations to tls init and cleanup jobs
- Ability to add pod security context to acl init and cleanup jobs
- Ability to add pod security context to tls init and cleanup jobs

How I've tested this PR:

Ran unit tests, helm lint, and also used changes to deploy at work where jobs not having the security context set by default run into regulation issues.

helm lint and unit tests

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/charts/consul/test/unit/server-acl-init-job.bats
+++ b/charts/consul/test/unit/server-acl-init-job.bats
@@ -591,6 +591,22 @@ load _helpers
 }
 
 #--------------------------------------------------------------------
+# server.containerSecurityContext.aclInit
+
+@test "serverACLInit/Job: securityContext is set when server.containerSecurityContext.aclInit is set" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-acl-init-job.yaml  \
+      --set 'global.acls.manageSystemACLs=true' \
+      --set 'server.containerSecurityContext.aclInit.runAsUser=100' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.securityContext.runAsUser' | tee /dev/stderr)
+
+  [ "${actual}" = "100" ]
+
+}
+
+#--------------------------------------------------------------------
 # Vault
 
 @test "serverACLInit/Job: fails when vault is enabled but neither bootstrap nor replication token is provided" {
@@ -2030,7 +2046,7 @@ load _helpers
       --set 'global.cloud.authUrl.secretName=auth-url-name' \
       .
   [ "$status" -eq 1 ]
-  
+
   [[ "$output" =~ "When either global.cloud.authUrl.secretName or global.cloud.authUrl.secretKey is defined, both must be set." ]]
 }
 
@@ -2050,7 +2066,7 @@ load _helpers
       --set 'global.cloud.authUrl.secretKey=auth-url-key' \
       .
   [ "$status" -eq 1 ]
-  
+
   [[ "$output" =~ "When either global.cloud.authUrl.secretName or global.cloud.authUrl.secretKey is defined, both must be set." ]]
 }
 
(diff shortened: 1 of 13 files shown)


david-yu:
Thanks this sounds like a good change. Could you tell us more why you would need annotations on acl init and tls init jobs?

Chinikins:
Sure, it's for flexibility more than anything else - in our case we have istio running on the cluster as well and using consul for just kv purposes, so adding annotations to the jobs to tell istio to ignore sidecar injection into these jobs please. But I can think of a lot of use cases where folks might need to add annotations to a pod depending on what else is on the cluster/use cases. A fairly simple one would be if you have a cluster autoscaler and you're doing an upgrade, annotating the job to also be skipped by the cluster autoscaler.

DanStough (changes requested):
Thank you so much for the PR @Chinikins 🥳 🎉 !

A couple of things to address in the PR before I think we are ready to merge:
1. See the comment about the indent for `containerSecurityContext` 
2. Can we add some unit tests for the annotation fields you added? I believe there is some [prior art here](https://github.com/hashicorp/consul-k8s/blob/1161322e494509ca2642811d88c92b898bb7a4d4/charts/consul/test/unit/server-statefulset.bats#L622-L643).
3. It looks like `.Values.global.acls.nodeSelector` is re-used in the following files. IMHO we should add `.Values.global.acls.annotations` to them as well, along with the respective unit tests.
    1. `gateway-cleanup-job.yaml`
    2. `gateway-resource-job.yaml`

Chinikins:
thank you for the quick review @DanStough !

charts/consul/test/unit/server-acl-init-job.bats
```diff
@@ -2226,3 +2242,27 @@ load _helpers
...
   [ "${actual}" = "bar" ]
 }
+
+#--------------------------------------------------------------------
+# annotations
+
+@test "serverACLInit/Job: no annotations defined by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-acl-init-cleanup-job.yaml \
```
DanStough, line 2252:
```suggestion
      -s templates/server-acl-init-job.yaml \
```

charts/consul/test/unit/server-acl-init-job.bats
```diff
@@ -2226,3 +2242,27 @@ load _helpers
...
+      --set 'global.acls.manageSystemACLs=true' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.metadata.annotations | del(."consul.hashicorp.com/connect-inject") | del(."consul.hashicorp.com/config-checksum")' | tee /dev/stderr)
+  [ "${actual}" = "{}" ]
+}
+
+@test "serverACLInit/Job: annotations can be set" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-acl-init-cleanup-job.yaml \
```
DanStough, line 2262:
```suggestion
      -s templates/server-acl-init-job.yaml \
```

DanStough (changes requested):
Two copy-pasta typos and then we're good to go!

DanStough (approved):

Chinikins:
Is there a way to have the CI tests run in the future that is more friendly to outside commits? It's a bit discouraging to continue contributions in this odd manner. :/

DanStough:
I understand and apologize for the inconvenience. It's an open item for us to find a better solution. We recently migrated chunks of our CI to GH actions and still need to sort out the workflows for forks. 

Thank you again for the PR!

david-yu:
@Chinikins I'll need to follow up with Github to see how they can help us make this easier. Our CI workflow is a bit unique here on consul-k8s so may need some assistance from them.

PR #2514, opened 2023-07-07
Add connectInject.prepareDataplanesUpgrade value

lkysow:
Set to true before performing an upgrade to a consul-dataplanes compatible consul-k8s version. This will ensure that the ACL tokens used to register non-dataplane services won't get deleted during the upgrade to dataplanes. If these ACL tokens are deleted, non-dataplane services won't be able to resolve their upstreams. During a normal non-dataplane upgrade this isn't an issue because the new injector re-registers all services using its new ACL token, but during an upgrade to dataplanes, the injector ignores all non-dataplane services and so doesn't re-register them with its latest token.

How I've tested this PR:
- ran the upgrade and confirmed old non-dataplane services have their upstreams updated

- code

Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/2514.txt
+++ b/.changelog/2514.txt
@@ -1,0 +1,1 @@
+improvement: Add `connectInject.prepareDataplanesUpgrade` setting for help upgrading to dataplanes. This setting is required if upgrading from non-dataplanes to dataplanes when ACLs are enabled. See https://developer.hashicorp.com/consul/docs/k8s/upgrade#upgrading-to-consul-dataplane for more information.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Upcoming
```
david-yu, line 1: This is actually autogenerated now. using the .changelog folder.

david-yu (approved):

PR #2520, opened 2023-07-10
NET-4813: Fix issue where virtual IP saving had insufficient ACLs.

hashi-derek:
When partitions were not enabled, the ACL token for the connect-inject pod would have insufficient privileges to persist virtual IP addresses to Consul. This PR ensures that connect-inject always has `mesh:write` access, so that transparent proxy can correctly utilize KubeDNS names.


lkysow (approved):
LGTM, it should be backported to the 1.16.x branch though right? release/1.2.x?

PR #2524, opened 2023-07-10
first pass making gateway controllers less chatty

missylbytes:
The gateway controllers are very chatty. This change allows for the [log level in the helm chart](https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/values.yaml#L19) to be utilized in the gateway controllers by setting the [verbosity](https://pkg.go.dev/github.com/go-logr/logr#hdr-Verbosity) on the info statements we use.

I think it makes most sense to just put all of our info logging under a different level. I think the info logs are most useful as an all-or-nothing type of thing (besides a couple of exceptions, see below). I think this because if a user really wanted to debug something with these logs, it would make sense to be able to see the whole lifecycle of the reconcile calls:
```
 Reconciling
	Doing deployment
	Doing HTTPRoute
	etc.
Reconciling again
```
On their own they do not make a ton of sense. To turn on this log level, the user just changes `global.logLevel` to `debug`. 

**Exceptions to log-level:**
When a user tries to delete a gatewayClass or a gatewayClassConfig, I left the info statements in saying that they cannot be deleted as they are still in use. These make sense to leave in, as they can happen outside of a general reconciliation loop.
[gatewayClass controller](https://github.com/hashicorp/consul-k8s/blob/main/control-plane/api-gateway/controllers/gatewayclass_controller.go#L76
)
[gatewayClassConfig controller](https://github.com/hashicorp/consul-k8s/blob/main/control-plane/api-gateway/controllers/gatewayclassconfig_controller.go#L59)

#### Room for improvement
Probably need to review the actual logging statements I added, make sure there is no copy-pasta errors

### How I've tested this PR:
Ran the helm chart locally with log levels at debug/info, created resources to see the logging.
Same as above


control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -199,18 +199,28 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			}
 			log.Error(err, "unable to update gateway resources")
 			return ctrl.Result{}, err
 		}
 		r.gatewayCache.EnsureSubscribed(nonNormalizedConsulKey, req.NamespacedName)
 	} else {
 		err := r.deleteGatekeeperResources(ctx, log, &gateway)
 		if err != nil {
+			if isModifiedError(err) {
+				r.Log.V(1).Info("error updating object, will try to re-reconcile")
```
missylbytes, line 215: maybe change this to deleting object?

control-plane/api-gateway/controllers/utils.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) HashiCorp, Inc.
```
missylbytes, line 1: Are we fine with adding a utils file?

control-plane/api-gateway/controllers/utils.go
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package controllers
+
+import (
+	"strings"
+)
+
+func isModifiedError(err error) bool {
+	return strings.Contains(err.Error(), "the object has been modified")
```
sarahalsmiller, line 11: The pattern I typically see for this type of thing is to define the modified error as a variable and then to call Errors.Is. Something like the following.

```
var ErrModifiedError = errors.New("the object has been modified")
errors.Is(err, ErrDivideByZero)
```

But I'm not 100% on if the package throwing this error has a error defined already

missylbytes: Yeah actually I think they do. Will update.

missylbytes: Hmm actually it might just be a specific error message, https://github.com/kubernetes/kubernetes/blob/c95b16b28076e84867ae0521574303ffc058becf/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go#L240C2-L240C24

missylbytes: https://github.com/kubernetes/apiserver/blob/v0.27.3/pkg/registry/generic/registry/store.go#L654

missylbytes: It looks like it would just be a conflict error though, I don't think there is a specific one for this type. So I could just call 
```
if k8serrors.IsConflict(err) {
```
but that might catch other conflict error types.

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -73,7 +73,7 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	var gateway gwv1beta1.Gateway
 
 	log := r.Log.WithValues("gateway", req.NamespacedName)
```
andrewstucki, line 75: Does `V(1)` return a logger? If so, would we just be able to do that here? I know that `Error` calls essentially ignore the log level, and I'd imagine pretty much everything that is info should be less verbose.

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -199,18 +199,28 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 		err := r.updateGatekeeperResources(ctx, log, &gateway, updates.GatewayClassConfig)
 		if err != nil {
+			if isModifiedError(err) {
+				r.Log.V(1).Info("error updating object, will try to re-reconcile")
+
+				return ctrl.Result{}, nil
```
andrewstucki, line 205: I'm assuming that _most_ of the time a modification failure will wind up with a re-reconcile anyway due to the way we trigger things -- but I think you'll want to do something like `return ctrl.Result{Retry: true}, nil` just in case -- this will ensure re-scheduling without the loud errors.

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -199,18 +199,28 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 			return ctrl.Result{}, err
 		}
 		r.gatewayCache.EnsureSubscribed(nonNormalizedConsulKey, req.NamespacedName)
 	} else {
 		err := r.deleteGatekeeperResources(ctx, log, &gateway)
 		if err != nil {
+			if isModifiedError(err) {
+				r.Log.V(1).Info("error updating object, will try to re-reconcile")
+
+				return ctrl.Result{}, nil
```
andrewstucki, line 217: same here

control-plane/api-gateway/controllers/utils.go
andrewstucki: so I think we'll be good to just use `IsConflict` as you mention -- any time we have a conflict I'd imagine we'd want to retry.

andrewstucki (changes requested):
A few changes, but overall looks great. Thanks for this Melisa!

control-plane/api-gateway/controllers/gateway_controller.go
missylbytes: Yes. Ended up doing this for the gateway controller, as all of the info statements there really don't need to be printed out unless someone is debugging. 
I had to do the gatewayClass and gatewayClassConfig controllers a little differently, so that the two special info statements (linked in description of PR) would still be printed out. There might be a reason to make them truly error statements?

control-plane/api-gateway/controllers/gateway_controller.go
missylbytes: Done

control-plane/api-gateway/controllers/gateway_controller.go
missylbytes: done

andrewstucki (approved):
LGTM -- before merging can you add a changelog entry? Can be something simple, maybe under "improvements" something like "make API gateway controller less verbose"?

PR #2526, opened 2023-07-10
docs: self service changelog instructions

DanStough:
Changes proposed in this PR:
- 3rd party contributors should add their own changelog. Link to instructions included.


david-yu:
Thanks this seems to be working well on `hashicorp/consul` so I am supportive of this.

PR #2534, opened 2023-07-10
Refactor test framework to allow for more than two kube contexts

wilkermichael:
**Note to reviewer** I have been very particular about my commits, for an easier time reviewing I suggest going through commit by commit. I have added extra notes, and it breaks up the context of the changes so that it's easier to review.

Changes proposed in this PR:
- To support sameness tests we need to be able to run 3 Consul clusters during the acceptance tests. Instead of adding an addition flag to accomodate this, this PR instead refactors the original flags to instead be flags that accept comma separated lists:

```
- -kubeconfig string
    The path to a kubeconfig file. If this is blank, the default kubeconfig path (~/.kube/config) will be used.
- -kubecontext string
    The name of the Kubernetes context to use. If this is blank, the context set as the current context will be used by default.
- -namespace string
    The Kubernetes namespace to use for tests. (default "default")
- -secondary-kubeconfig string
    The path to a kubeconfig file of the secondary k8s cluster. If this is blank, the default kubeconfig path (~/.kube/config) will be used.
- -secondary-kubecontext string
    The name of the Kubernetes context for the secondary cluster to use. If this is blank, the context set as the current context will be used by default.
- -secondary-namespace string
    The Kubernetes namespace to use in the secondary k8s cluster. (default "default")
+ -kubeconfigs string
    The comma separate list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If this is empty, or fields are blank, then the context set as the current context will be used by default.
+ -kubecontexts string
    The comma separate list of Kubernetes contexts to use (eg. "kind-dc1,kind-dc2"). The first in the list will be treated as the primary context, followed by the secondary, etc. If this is empty, or fields are blank then the default kubeconfig path (~/.kube/config) will be used.
+ -kubenamespaces string
    The comma separate list of Kubernetes namespaces to use (eg. "consul,consul-secondary"). The first in the list will be treated as the primary namespace, followed by the secondary, etc. If the list is empty, or fields are blank, then the "default" namespace will be used.
```

The lists are then combined into lists of the following object for use in tests:
```
type KubeEnv struct {
	KubeConfig    string
	KubeContext   string
	KubeNamespace string
}
```

Example test run command:
```
go test -run "TestPeering/default_installation" -p 1 -timeout 90m -v -failfast -args -use-kind -kubecontexts "kind-dc1,kind-dc2" -enable-enterprise -enable-multi-cluster -consul-image=wilko1989/consul-enterprise@sha256:3cfc9f5cf8a5c49c636bcc26ef0df1a30ac8fd68becdbb1dece192ea20d467a8 -consul-k8s-image=wilko1989/consul-k8s-control-plane-dev@sha256:af2f5035e987b2946ad1c535b37a565e0adce4e01d777681106c5fa7eb012e8b -consul-dataplane-image=wilko1989/consul-dataplane@sha256:b9db7e4c0f96ad110f35104a7b8de0cbc0b94f81f437b83c264d5a1c0f68fca5 -enterprise-license <redacted>
```

How I've tested this PR:

- Ran acceptance tests and verified that they still pass

👀 

Checklist:
- [ ] Tests added

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -666,7 +666,7 @@ Below is the list of available flags:
     This applies only to tests that enable connectInject.
 -enterprise-license
     The enterprise license for Consul.
--kube-configs string
+-kubeconfigs string
     The comma separated list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If the list is empty, or items are blank, then the default kubeconfig path (~/.kube/config) will be used.
 -kube-contexts string
     The comma separated list of Kubernetes contexts to use (eg. "kind-dc1,kind-dc2"). The first in the list will be treated as the primary context, followed by the secondary, etc. If the list is empty, or items are blank, then the current context will be used.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -642,7 +642,7 @@ you may use the following command:
 
     go test ./... -p 1 -timeout 20m \
         -enable-multi-cluster \
-        -kubecontexts="<name of the primary Kubernetes context>,<name of the secondary Kubernetes context>, etc.>"
+        -kube-contexts="<name of the primary Kubernetes context>,<name of the secondary Kubernetes context>, etc.>"
 
 Below is the list of available flags:
 
@@ -666,15 +666,14 @@ Below is the list of available flags:
     This applies only to tests that enable connectInject.
 -enterprise-license
     The enterprise license for Consul.
--kubeconfigs string
-    The comma separate list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If this is empty, or fields are blank, then the context set as the current context will be used by default.
--kubecontexts string
-    The comma separate list of Kubernetes contexts to use (eg. "kind-dc1,kind-dc2"). The first in the list will be treated as the primary context, followed by the secondary, etc. If this is empty, or fields are blank then the default kubeconfig path (~/.kube/config) will be used.
--kubenamespaces string
-    The comma separate list of Kubernetes namespaces to use (eg. "consul,consul-secondary"). The first in the list will be treated as the primary namespace, followed by the secondary, etc. If the list is empty, or fields are blank, then the "default" namespace will be used.
+-kube-configs string
+    The comma separated list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If the list is empty, or items are blank, then the default kubeconfig path (~/.kube/config) will be used.
+-kube-contexts string
+    The comma separated list of Kubernetes contexts to use (eg. "kind-dc1,kind-dc2"). The first in the list will be treated as the primary context, followed by the secondary, etc. If the list is empty, or items are blank, then the current context will be used.
+-kube-namespaces string
+    The comma separated list of Kubernetes namespaces to use (eg. "consul,consul-secondary"). The first in the list will be treated as the primary namespace, followed by the secondary, etc. If the list is empty, or fields are blank, then the current namespace will be used.
 -no-cleanup-on-failure
     If true, the tests will not cleanup Kubernetes resources they create when they finish running.Note this flag must be run with -failfast flag, otherwise subsequent tests will fail.
-
 ```
 
 **Note:** There is a Terraform configuration in the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -642,8 +642,7 @@ you may use the following command:
 
     go test ./... -p 1 -timeout 20m \
         -enable-multi-cluster \
-        -kubecontext=<name of the primary Kubernetes context> \
-        -secondary-kubecontext=<name of the secondary Kubernetes context>
+        -kubecontexts="<name of the primary Kubernetes context>,<name of the secondary Kubernetes context>, etc.>"
 
 Below is the list of available flags:
 
@@ -667,20 +666,15 @@ Below is the list of available flags:
     This applies only to tests that enable connectInject.
 -enterprise-license
     The enterprise license for Consul.
--kubeconfig string
-    The path to a kubeconfig file. If this is blank, the default kubeconfig path (~/.kube/config) will be used.
--kubecontext string
-    The name of the Kubernetes context to use. If this is blank, the context set as the current context will be used by default.
--namespace string
-    The Kubernetes namespace to use for tests. (default "default")
+-kubeconfigs string
+    The comma separate list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If this is empty, or fields are blank, then the context set as the current context will be used by default.
+-kubecontexts string
+    The comma separate list of Kubernetes contexts to use (eg. "kind-dc1,kind-dc2"). The first in the list will be treated as the primary context, followed by the secondary, etc. If this is empty, or fields are blank then the default kubeconfig path (~/.kube/config) will be used.
+-kubenamespaces string
+    The comma separate list of Kubernetes namespaces to use (eg. "consul,consul-secondary"). The first in the list will be treated as the primary namespace, followed by the secondary, etc. If the list is empty, or fields are blank, then the "default" namespace will be used.
 -no-cleanup-on-failure
     If true, the tests will not cleanup Kubernetes resources they create when they finish running.Note this flag must be run with -failfast flag, otherwise subsequent tests will fail.
--secondary-kubeconfig string
-    The path to a kubeconfig file of the secondary k8s cluster. If this is blank, the default kubeconfig path (~/.kube/config) will be used.
--secondary-kubecontext string
-    The name of the Kubernetes context for the secondary cluster to use. If this is blank, the context set as the current context will be used by default.
--secondary-namespace string
-    The Kubernetes namespace to use in the secondary k8s cluster. (default "default")
+
 ```
 
 **Note:** There is a Terraform configuration in the
--- a/acceptance/framework/config/config.go
+++ b/acceptance/framework/config/config.go
@@ -31,7 +31,7 @@ type KubeTestConfig struct {
 
 // NewKubeTestConfigList takes lists of kubernetes configs, contexts and namespaces and constructs KubeTestConfig
 // We validate ahead of time that the lists are either 0 or the same length as we expect that if the length of a list
-// is greater than 0, then the indexes should match. For example: []kubeContexts{"ctx1", "ctx2"} indexes 0, 1 match with []kubeNamespaces{"ns1", "ns2"}
+// is greater than 0, then the indexes should match. For example: []kubeContexts{"ctx1", "ctx2"} indexes 0, 1 match with []kubeNamespaces{"ns1", "ns2"}.
 func NewKubeTestConfigList(kubeConfigs, kubeContexts, kubeNamespaces []string) []KubeTestConfig {
 	// Grab the longest length.
 	l := math.Max(float64(len(kubeConfigs)),
--- a/acceptance/framework/config/config.go
+++ b/acceptance/framework/config/config.go
@@ -22,16 +22,67 @@ const (
 	LicenseSecretKey  = "key"
 )
 
-// TestConfig holds configuration for the test suite.
-type TestConfig struct {
-	Kubeconfig    string
+type KubeEnv struct {
+	KubeConfig    string
 	KubeContext   string
 	KubeNamespace string
+}
 
-	EnableMultiCluster     bool
-	SecondaryKubeconfig    string
-	SecondaryKubeContext   string
-	SecondaryKubeNamespace string
+func KubeEnvListFromStringList(kubeConfigs, kubeContexts, kubeNamespaces []string) []KubeEnv {
+	// Determine the longest list length
+	maxLen := 0
+	lenConf := len(kubeConfigs)
+	if maxLen < lenConf {
+		maxLen = lenConf
+	}
+
+	lenCtx := len(kubeContexts)
+	if maxLen < lenCtx {
+		maxLen = lenCtx
+	}
+
+	lenNs := len(kubeNamespaces)
+	if maxLen < lenNs {
+		maxLen = lenNs
+	}
+
+	// If all are empty, then return a single empty entry
+	out := make([]KubeEnv, 0)
+	if maxLen == 0 {
+		out = append(out, KubeEnv{})
+		return out
+	}
+
+	// Pad lists if required, all list should be made the same length
+	for i := lenConf; i < maxLen; i++ {
+		kubeConfigs = append(kubeConfigs, "")
+	}
+
+	for i := lenCtx; i < maxLen; i++ {
+		kubeContexts = append(kubeContexts, "")
+	}
+
+	for i := lenNs; i < maxLen; i++ {
+		kubeNamespaces = append(kubeNamespaces, "")
+	}
+
+	// Construct the kubeEnv
+	for k, v := range kubeConfigs {
+		kenv := KubeEnv{
+			KubeConfig:    v,
+			KubeContext:   kubeContexts[k],
+			KubeNamespace: kubeNamespaces[k],
+		}
+		out = append(out, kenv)
+	}
+
+	return out
+}
+
+// TestConfig holds configuration for the test suite.
+type TestConfig struct {
+	KubeEnvs           []KubeEnv
+	EnableMultiCluster bool
 
 	EnableEnterprise  bool
 	EnterpriseLicense string
@@ -115,6 +166,23 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
 	setIfNotEmpty(helmValues, "global.imageConsulDataplane", t.ConsulDataplaneImage)
 
 	return helmValues, nil
+}
+
+// IsExpectedClusterCount check that we have at least the required number of clusters to
+// run a test.
+func (t *TestConfig) IsExpectedClusterCount(count int) bool {
+	return len(t.KubeEnvs) >= count
+}
+
+// GetPrimaryKubeEnv returns the primary Kubernetes environment.
+func (t *TestConfig) GetPrimaryKubeEnv() KubeEnv {
+	// Return the first in the list as this is always the primary
+	// kube environment. If empty return an empty kubeEnv
+	if len(t.KubeEnvs) < 1 {
+		return KubeEnv{}
+	} else {
+		return t.KubeEnvs[0]
+	}
 }
 
 type values struct {
--- a/acceptance/framework/config/config.go
+++ b/acceptance/framework/config/config.go
@@ -5,6 +5,7 @@ package config
 
 import (
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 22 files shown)


acceptance/tests/vault/main_test.go
```diff
@@ -14,5 +15,11 @@ var suite testsuite.Suite
 
 func TestMain(m *testing.M) {
 	suite = testsuite.NewSuite(m)
-	os.Exit(suite.Run())
+
+	if suite.Config().EnableMultiCluster && suite.Config().IsExpectedClusterCount(2) {
+		os.Exit(suite.Run())
+	} else {
+		fmt.Println("Skipping vault tests because either -enable-multi-cluster is not set or -disable-peering is set")
```
wilkermichael, line 22: I'll fix this message

hashi-derek (approved):

acceptance/framework/config/config.go
```diff
@@ -22,16 +22,67 @@ const (
 	LicenseSecretKey  = "key"
 )
 
-// TestConfig holds configuration for the test suite.
-type TestConfig struct {
-	Kubeconfig    string
+type KubeEnv struct {
```
curtbushko, line 25: I think having this as `TestConfig` instead of `KubeEnv` is better as it clashes with KubernetesEnvironment from environment.go.  Either that or merge the two types together?

.github/workflows/pr.yml
```diff
@@ -19,6 +19,6 @@ jobs:
       with:
         workflow: test.yml
         repo: hashicorp/consul-k8s-workflows
-        ref: main
+        ref: update-new-context-flags
```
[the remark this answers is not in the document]
wilkermichael: oh that's here: https://github.com/hashicorp/consul-k8s-workflows/pull/24

curtbushko (commented):
So many good things!  I like how your changes didn't result in several tests needing to be refactored.

I left some comments but overall this is pretty solid.

CONTRIBUTING.md
```diff
@@ -667,20 +666,15 @@ Below is the list of available flags:
...
--kubecontext string
-    The name of the Kubernetes context to use. If this is blank, the context set as the current context will be used by default.
--namespace string
-    The Kubernetes namespace to use for tests. (default "default")
+-kubeconfigs string
+    The comma separate list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If this is empty, or fields are blank, then the context set as the current context will be used by default.
+-kubecontexts string
+    The comma separate list of Kubernetes contexts to use (eg. "kind-dc1,kind-dc2"). The first in the list will be treated as the primary context, followed by the secondary, etc. If this is empty, or fields are blank then the default kubeconfig path (~/.kube/config) will be used.
+-kubenamespaces string
+    The comma separate list of Kubernetes namespaces to use (eg. "consul,consul-secondary"). The first in the list will be treated as the primary namespace, followed by the secondary, etc. If the list is empty, or fields are blank, then the "default" namespace will be used.
```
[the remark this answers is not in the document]
wilkermichael: We don't use namespaces anywhere from what I can tell

acceptance/framework/config/config.go
```diff
@@ -22,16 +22,67 @@ const (
...
+	}
+
+	// If all are empty, then return a single empty entry
+	out := make([]KubeEnv, 0)
+	if maxLen == 0 {
+		out = append(out, KubeEnv{})
+		return out
+	}
+
+	// Pad lists if required, all list should be made the same length
```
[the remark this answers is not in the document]
wilkermichael: That is correct, if you wanted config3 paired with ns3 and no ns2 it would be 
```
configs: config1, config2, config3
context: context1, context2
namespaces: ns1, "", ns2
```
```
config1, context1, ns1
config2, context2, ""
config3, "", ns3
```

It's kind funky with the command line inputs

acceptance/framework/config/config.go
```diff
@@ -117,6 +168,23 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
...
 }
 
+// IsExpectedClusterCount check that we have at least the required number of clusters to
+// run a test.
+func (t *TestConfig) IsExpectedClusterCount(count int) bool {
+	return len(t.KubeEnvs) >= count
+}
+
+// GetPrimaryKubeEnv returns the primary Kubernetes environment.
+func (t *TestConfig) GetPrimaryKubeEnv() KubeEnv {
```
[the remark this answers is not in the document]
wilkermichael: We don't use that anywhere, so I didn't make a function for it

acceptance/framework/config/config.go
wilkermichael: `KubeTestConfig` makes the most sense I guess?

acceptance/framework/config/config.go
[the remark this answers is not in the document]
wilkermichael: A bunch of permutations are handled in the unit test

CONTRIBUTING.md
```diff
@@ -667,20 +666,15 @@ Below is the list of available flags:
...
 -enterprise-license
     The enterprise license for Consul.
--kubeconfig string
-    The path to a kubeconfig file. If this is blank, the default kubeconfig path (~/.kube/config) will be used.
--kubecontext string
-    The name of the Kubernetes context to use. If this is blank, the context set as the current context will be used by default.
--namespace string
-    The Kubernetes namespace to use for tests. (default "default")
+-kubeconfigs string
+    The comma separate list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If this is empty, or fields are blank, then the context set as the current context will be used by default.
```
ishustava, line 670:
```suggestion
    The comma separated list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If this is empty, or fields are blank, then the context set as the current context will be used by default.
```

ishustava: I think the ones below should also say `separated`

CONTRIBUTING.md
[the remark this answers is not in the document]
ishustava: We should use the default namespace from your context instead (this is how it works today). So if I call `kubectl config set-context --current --namespace=some-other-namespace`, we should use `some-other-namespace` instead of `default`. This is also how all of kube machinery works.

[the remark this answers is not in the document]
ishustava: We don't set kubenamespaces explicitly anywhere in the pipeline. For openshift tests, we used to set it in the context in the kubeconfig file because you don't want to use the default project on openshift.

CONTRIBUTING.md
```diff
@@ -667,20 +666,15 @@ Below is the list of available flags:
...
+    The comma separate list of Kubernetes namespaces to use (eg. "consul,consul-secondary"). The first in the list will be treated as the primary namespace, followed by the secondary, etc. If the list is empty, or fields are blank, then the "default" namespace will be used.
 -no-cleanup-on-failure
     If true, the tests will not cleanup Kubernetes resources they create when they finish running.Note this flag must be run with -failfast flag, otherwise subsequent tests will fail.
--secondary-kubeconfig string
-    The path to a kubeconfig file of the secondary k8s cluster. If this is blank, the default kubeconfig path (~/.kube/config) will be used.
--secondary-kubecontext string
-    The name of the Kubernetes context for the secondary cluster to use. If this is blank, the context set as the current context will be used by default.
--secondary-namespace string
-    The Kubernetes namespace to use in the secondary k8s cluster. (default "default")
+
```
ishustava, line 677: nit: do we need the extra line?

acceptance/framework/config/config.go
ishustava: I'd be fine with `KubeTestConfig` although the `kube` part is arguably implied by the repo name. 

For context, originally I was thinking that this framework could be reused for other non-kube setups but that never happened. So it probably could make sense to make names more specific.

acceptance/framework/config/config.go
```diff
@@ -22,16 +22,67 @@ const (
...
-// TestConfig holds configuration for the test suite.
-type TestConfig struct {
-	Kubeconfig    string
+type KubeEnv struct {
+	KubeConfig    string
 	KubeContext   string
 	KubeNamespace string
+}
+
+func KubeEnvListFromStringList(kubeConfigs, kubeContexts, kubeNamespaces []string) []KubeEnv {
```
ishustava, line 31: ❓ why don't we just validate that the lists are the same length because otherwise it will get ambiguous. 

So if I want to say `-kube-context=kind-dc1 -kube-context=kind-dc2 -namespace=consul`, we can only interpret it as:
- the first kube-context has the namespace `consul`
- the second kube-context has default namespace of that context

I think enforcing the same length could be simpler because I'm not sure we need this complexity for our use cases anyways. 

So I'm suggesting:
- If you've provided any 2 of the three, their lengths should be equal
- If you've provided all 3, their lengths should be equal as well

Then we just construct the env from those arrays by their position as we don't need to figure out the one with max len.

ishustava (commented):
Thanks so much for doing this Michael! This looks good! 

Leaving some comments I have so far.

acceptance/framework/environment/environment.go
```diff
@@ -21,15 +21,17 @@ import (
 )
 
 const (
-	DefaultContextName   = "default"
-	SecondaryContextName = "secondary"
+	DefaultContextName         = "default"
+	SecondaryContextNamePrefix = "secondary"
```
ishustava, line 25: I think we should probably remove the secondary context const and the specific functions. Now that we have more than one, it'd be nice to change it to:
- get the default (same as today)
- get specific context by index. Ideally the default one should just call the specific one with index `0`

acceptance/framework/flags/flags.go
```diff
@@ -68,13 +65,19 @@ func NewTestFlags() *TestFlags {
...
+
+// String() returns a comma separated list in the form "var1,var2,var3".
+func (f *listFlag) String() string {
+	return strings.Join(*f, ",")
+}
+
+func (f *listFlag) Set(value string) error {
+	*f = strings.Split(value, ",")
+	return nil
+}
```
ishustava, line 78: Might be worth re-using existing implementation somehow: https://github.com/hashicorp/consul-k8s/blob/main/control-plane/subcommand/flags/flag_slice_value.go

acceptance/framework/flags/flags.go
```diff
@@ -86,16 +89,16 @@ func (t *TestFlags) init() {
 	flag.StringVar(&t.flagVaultServerVersion, "vault-server-version", "", "The vault serverversion used for all tests.")
 	flag.StringVar(&t.flagVaultHelmChartVersion, "vault-helm-chart-version", "", "The Vault helm chart used for all tests.")
 
+	flag.Var(&t.flagKubeconfigs, "kubeconfigs", "The list of paths to a kubeconfig files. If this is blank, "+
+		"the default kubeconfig path (~/.kube/config) will be used.")
+	flag.Var(&t.flagKubecontexts, "kubecontexts", "The list of names of the Kubernetes contexts to use. If this is blank, "+
+		"the context set as the current context will be used by default.")
+	flag.Var(&t.flagNamespaces, "namespaces", "The list Kubernetes namespace to use for tests.")
```
ishustava, line 96:
```suggestion
	flag.Var(&t.flagNamespaces, "namespaces", "The list of Kubernetes namespaces to use for tests.")
```

ishustava (commented):
Leaving a few more comments

CONTRIBUTING.md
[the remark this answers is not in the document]
wilkermichael: ok, I can verify that works as before

acceptance/framework/config/config.go
wilkermichael: Ah thanks for the history on that, yeah I think specific works in this case since the use case has narrowed

acceptance/framework/config/config.go
wilkermichael: So basically, if you provide any of the flags `kubecontext` `kubeconfig` and `namespaces` then we enforce theya are the same length. If a flag is not provided/list is empty then we just set it all to empty so it's processed as before?

Here's are examples from what I'm understanding:
```
-kubecontexts="kind-dc1,kind-dc2" -namespaces="consul" # would result in an error
```
```
-kubecontexts="kind-dc1,kind-dc2" -namesapces="consul,consul2" # same length so OK!
# Result Structs (we don't set `kube-config` since it wasn't provided)
[]KubeEnv{
  {
    KubeContext: "kind-dc1"", 
    KubeNamespace: "consul"
  },
  {
    KubeContext: "kind-dc2"", 
    KubeNamespace: "consul2"
  },
}
```

wilkermichael: I think we still need to find out the max length so we can pad the one that's 0 if one of them isn't set?

ishustava: >If a flag is not provided/list is empty then we just set it all to empty so it's processed as before?

Yeah, exactly.

Yep your examples are exactly what I meant.

ishustava: > I think we still need to find out the max length so we can pad the one that's "" if one of them isn't set?

could we rely on defaults in KubeEnv instead? If length is 0 for namespaces for example, then kubeenv for that index will end up not setting the namespace field which should be the same as `""`

wilkermichael: I think no matter how you slice it if you're constructing a new list you'd still have to get the the length of the new list to create, so you're right we don't technically need to pad, we could check to see if the list is zero and if it not index through it... I'm not sure if that's cleaner?

wilkermichael: I guess we could do something like this:
```
	out := make([]KubeTestConfig, 0)
	if len(kubeConfigs) == 0 && len(kubeContexts) == 0 && len(kubeNamespaces) == 0{
		out = append(out, KubeTestConfig{})
		return out
	}
	
	i := 0
	for true{
		kenv := KubeTestConfig{}
		if i != 0 && (i == len(kubeConfigs) || i == len(kubeContexts) || i == len(kubeNamespaces)){
			break
		}
		if len(kubeConfigs) != 0{
			kenv.KubeConfig = kubeConfigs[i]
		}
		if len(kubeContexts) != 0{
			kenv.KubeContext = kubeContexts[i]
		}
		if len(kubeNamespaces) != 0{
			kenv.KubeNamespace = kubeNamespaces[i]
		}
		out = append(out, kenv)
	}
	return out
```

acceptance/framework/flags/flags.go
wilkermichael: So this would be different from what we have I think, this is what it looks like using it with this type:
```
-kube-context="ctx-1" -kube-context="ctx-2"
```
vs what I have
```
-kube-contexts="ctx-1,ctx-2"
```

I think I like mine better, what do you think?

wilkermichael: actually this gets really messy with the namespaces and kubeconfigs, I think we should leave it with the comma separated list

acceptance/framework/config/config.go
ishustava: That's true. You could do something like this which is less code but I dunno if it's that much simpler because you're right you still need to check the length:

```
       l := math.Max(
		float64(len(kubeConfigs)),
		math.Max(float64(len(kubeContexts)), float64(len(kubeNamespaces))))

	// If all are empty, then return a single empty entry
	if l == 0 {
		return []KubeEnv{{}}
	}

	out := make([]KubeEnv, int(l))

	for i := range out {
		if len(kubeConfigs) != 0 {
			out[i].KubeConfig = kubeConfigs[i]
		}

		if len(kubeContexts) != 0 {
			out[i].KubeContext = kubeContexts[i]
		}

		if len(kubeNamespaces) != 0 {
			out[i].KubeNamespace = kubeNamespaces[i]
		}
	}

	return out
```

wilkermichael: oh yeah, `math.Max` is a nice call

wilkermichael: The only thing I changed was checking for `> i` instead of `!=0`

This guarantees we don't index out of range, although we are now validating that lengths are either 0 or equal

CONTRIBUTING.md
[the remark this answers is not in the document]
wilkermichael: talked about this offline, but they are used in the nightly cloud tests

acceptance/framework/flags/flags.go
ishustava: I don't have a super strong opinion between the two, but I'd prefer consistency. IMO also kubecontexts or namespaces look fine, but for kubeconfigs where you need to provide paths it doesn't look as nice.

```
-kubeconfigs=$HOME/path/to/kubeconfig1,$HOME/path/to/kubeconfig2
```

>actually this gets really messy with the namespaces and kubeconfigs, I think we should leave it with the comma separated list

Why does it get messy? They both parse into a list of strings I believe.

wilkermichael: Not messy in the sense of the algorithm, because like you said it stays the same, I just mean UX wise for the user:
-kube-context="ctx-1" -kube-context="ctx-2" -kube-namespace="ns1" -kube-namepsace="ns2"

Maybe it's not that bad, but you have to know that the first `-kube-context` maps to the first `-kube-namepsace`

Mind you maybe it makes it easier if used in this case:
-kube-context="ctx-1" -kube-namespace="ns1" -kube-context="ctx-2" -kube-namespace="ns2"

curtbushko: > Maybe it's not that bad, but you have to know that the first -kube-context maps to the first -kube-namespace

You need to know that with comma separated too so there is no winning.

I prefer comma separated out of all the options as this makes me 😭 

> -kube-context="ctx-1" -kube-context="ctx-2" -kube-context="ctx-3" -kube-context="ctx-4"-kube-context="ctx-5" -kube-context="ctx-6" -kube-namespace="ns1" -kube-namepsace="ns2" -kube-namepsace="ns3" -kube-namepsace="ns4" -kube-namepsace="ns5"

wilkermichael: yeah... my example of keeping track with the multiple arguments one maybe isn't the best. The list at least enforces all the "contexts" for instance are right next to each other, you can technically put the flags anywhere. One after the other, between other flags, etc.

wilkermichael: was talking to Curt, maybe we're at a point where it makes more sense to have a config file for these arguments...

one refactor at a time

ishustava: 💯 a config file will make it much better. 

>one refactor at a time, I'll stick with lists for now

So what did you all decide? They are both lists.

wilkermichael: comma separated lists, sorry for the confusion. I don't think of the multi argument one as a "list input" so I keep referring to the comma separated one as just list 😓 

I realise they both get consumed as lists 😅

acceptance/framework/environment/environment.go
wilkermichael: I'll create a separate commit for this refactor and you can let me know if that's what you had in mind

CONTRIBUTING.md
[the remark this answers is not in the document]
wilkermichael: verified that if the namespace isn't provided it uses whatever you set like in the command Iryna posted:
`kubectl config set-context --current --namespace=some-other-namespace`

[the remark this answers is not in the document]
wilkermichael: Just verified that this doesn't work as expected, this didn't work before either (I checked out main) so we'll have to create an issue and fix this at some point.

Nothing is really exercising the behaviour currently so I guess it broke at some point

CONTRIBUTING.md
```diff
@@ -666,15 +666,14 @@ Below is the list of available flags:
     This applies only to tests that enable connectInject.
 -enterprise-license
     The enterprise license for Consul.
--kubeconfigs string
-    The comma separate list of Kubernetes configs to use (eg. "~/.kube/config,~/.kube/config2"). The first in the list will be treated as the primary config, followed by the secondary, etc. If this is empty, or fields are blank, then the context set as the current context will be used by default.
--kubecontexts string
-    The comma separate list of Kubernetes contexts to use (eg. "kind-dc1,kind-dc2"). The first in the list will be treated as the primary context, followed by the secondary, etc. If this is empty, or fields are blank then the default kubeconfig path (~/.kube/config) will be used.
--kubenamespaces string
-    The comma separate list of Kubernetes namespaces to use (eg. "consul,consul-secondary"). The first in the list will be treated as the primary namespace, followed by the secondary, etc. If the list is empty, or fields are blank, then the "default" namespace will be used.
+-kube-configs string
```
wilkermichael, line 669: this should be kubeconfigs

ishustava (approved):

curtbushko (approved):

Issue #2491, opened 2023-06-29
BUG+FIX: Endpoints controller fails to deregister services


mr-miles:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

We regularly see orphan services "stuck" in the catalog - the services are terminated as expected by k8s but this does not always reflect in the consul service catalog - leaving the last state of the service "frozen".  The impact is catastrophic in the service mesh because if the service is healthy it will continue to be part of envoy clusters and therefore yield intermittent errors to callers.  Worse, if the service is a terminating gateway then those errors will apply to any outbound connection in any service.

We have identified the root cause of this problem as a timing issue in the connect-inject endpoint controller.

The issue occurs in any environment where nodes are dynamically provisioned, for example with EKS+karpenter, or using spot instances.

The issue is mentioned in many issue reports - #2085 #2065 #1817 #1935 #15908 https://github.com/hashicorp/consul/issues/15908 

### Root cause

The controller is triggered by a change event to the endpoint object.
When the controller receives the event, it carries out deregistration with this code:
https://github.com/hashicorp/consul-k8s/blob/920ee3233d41792134a02f7a0e9e1d2d5ed4482d/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go#L1095C44-L1095C44
This is calling the k8s api to retrieve a list of active k8s nodes, and then goes on to find service instances on each of those nodes to true up.

The issue is that if there is any delay in processing the event, the list of nodes that is retrieved may not be the same as when the event was generated.  If a service is registered to a synthetic node in the consul catalog that is not in the node list that the controller retrieves, then it will not be deregistered.

An example scenario:

- A cluster with 5 nodes.  One node running 10 pods is shut down
- K8s generates the endpoint events as expected as pods are terminated
- The controller processes the first 7 events.  The controller retrieves a node list containing all 5 nodes.
- The node is now empty and k8s removes it.
[...]

Issue #2085, opened 2023-04-24
Wrong Instance Count because of Cluster Restart


maheshrajrp:
#### Overview of the Issue
When a Azure Cluster is restarted, the previous pods running inside the mesh are lost and new pods are created. Post restart Consul isn't picking up that the previous pods are deleted/not present. Hence, it is still trying to route to the pods resulting in the following error when I try to consume the service be UI/API.

![image](https://user-images.githubusercontent.com/80001447/232262847-a378318a-6906-4e93-95d9-22aa2908a03b.png)

Further more: 
These are the active pods:
![image](https://user-images.githubusercontent.com/80001447/232262916-4e253b0c-2fee-4f5d-ba52-f1f7b7fcf059.png)

Whereas, Consul UI shows this,
![image](https://user-images.githubusercontent.com/80001447/232262931-500c9d88-dba2-4a37-9760-c3e2004bb893.png)

Notice that there is only one frontend pod is running in AKS whereas UI shows two instances of the service

---

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Create a service mesh within Azure Kubernetes Service.
1. Stop and Start the AKS. 
1.  Notice the previous pod info still exists in Consul UI whereas in reality it doesn't exist in AKS.

-->

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>
/ $ consul info
agent:
        check_monitors = 0
        check_ttls = 0
        checks = 0
        services = 0
build:
        prerelease =
        revision = 7c04b6a0
        version = 1.15.1
        version_metadata =
consul:
        acl = disabled
        bootstrap = true
        known_datacenters = 1
        leader = true
        leader_addr = 10.244.0.12:8300
        server = true
raft:
        applied_index = 2213
        commit_index = 2213
        fsm_pending = 0
        last_contact = 0
        last_log_index = 2213
        last_log_term = 4
        last_snapshot_index = 0
        last_snapshot_term = 0
        latest_configuration = [{Suffrage:Voter ID:b9744a41-cccd-861f-eca2-f3b18496e5b4 Address:10.244.0.12:8300}]
        latest_configuration_index = 0
        num_peers = 0
        protocol_version = 3
        protocol_version_max = 3
        protocol_version_min = 0
        snapshot_version_max = 1
        snapshot_version_min = 0
        state = Leader
        term = 4
runtime:
        arch = amd64
        cpu_count = 4
        goroutines = 259
        max_procs = 4
        os = linux
[...]

Issue #2065, opened 2023-04-18
Nodes and dead services remaining in the consul catalog


Stephani0106:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

We noticed that the issue began to occur in our dev environment, and later in qas, after the adoption of spot instances within our Kubernetes cluster (AKS). We don't rule out the possibility that the mistake has always happened, but we don't notice it as often as we do now.

Because they are machines with idle capacity, sometimes some of our nodes were killed because the machine was "stolen" by Azure, however our environment whenever it identified these node deaths recreated them with new names to always keep a minimum number of nodes available in the environment.

Our issue within the consul began to be identified just when there were these node deaths. When a node is killed, the Consul does not deregister it from the catalog even though it no longer exists within the cluster. This ends up causing "ghost" services to remain in the Consul catalog with failure status even though they no longer exist.

This results in dead service instances in the Consul catalog and healthcheck issues, for example.

![image](https://user-images.githubusercontent.com/73966290/232870056-d8a7b662-c0ae-4045-bd7d-c6b4ac62b11f.png)

### Palliative solution

As a palliative solution to work around the problem, we manually removed dead services from the catalog. For this we open a shell in one of the pods of the Consul server and execute the request below:

```bash
curl http://127.0.0.1:8500/v1/catalog/service/<service-name> | jq
```
After making sure that the node in fact no longer exists in the cluster, we run the command below to unregister all the services present in that node:

```bash
curl --request PUT -d '{"Node":"<node-name>"}' http://127.0.0.1:8500/v1/catalog/deregister
```

This solves the problem temporarily, but every time a node is killed or recreated the problem recurs and again we need to perform the steps above.

### Expected behavior

We expected Consul to deregister all nodes and services that did not respond properly to healthcheck, thus excluding dead services from its catalog.

### Environment details

[...]

Issue #1935, opened 2023-02-22
Daemonset services are listed after the nodes is decommissioned from cluster.


tejnar:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Overview of the Issue

Daemonset services are listed in consul UI and api after the node is decommissioned from cluster.

### Reproduction Steps
1. Install consul server without the agents.
2. Deploy Daemonset(Fluent-bit or node-exporter) to cluster.
3. Delete any node in cluster.

Actual Behaviour: Consul UI is displaying the deleted daemonset service.  API is also listing the services which aren't part of the cluster.
API command: curl -i http://127.0.0.1:8500/v1/catalog/service/fluent-bit

Attached the values.yaml as values.txt file.
[values.txt](https://github.com/hashicorp/consul-k8s/files/10805091/values.txt)

/ $ consul info
agent:
        check_monitors = 0
        check_ttls = 0
        checks = 0
        services = 0
build:
        prerelease = 
        revision = dae670fe
        version = 1.14.4
        version_metadata = 
consul:
        acl = disabled
        bootstrap = false
        known_datacenters = 1
        leader = false
        leader_addr = 100.64.54.123:8300
        server = true
raft:
        applied_index = 69775
        commit_index = 69775
        fsm_pending = 0
        last_contact = 21.178669ms
        last_log_index = 69775
        last_log_term = 8
        last_snapshot_index = 65546
        last_snapshot_term = 8
        latest_configuration = [{Suffrage:Voter ID:72b147fe-690f-0c42-f07e-d42709b9c255 Address:100.64.54.123:8300} {Suffrage:Voter ID:fe9ca4ff-3608-45ee-fc25-d4774c1e6f9b Address:100.64.28.115:8300} {Suffrage:Voter ID:5a56a260-d45e-6748-2c13-fb17877d273b Address:100.64.40.151:8300}]
        latest_configuration_index = 0
        num_peers = 2
        protocol_version = 3
        protocol_version_max = 3
[...]

PR #2571, opened 2023-07-13
Handle errors properly when services are de-registered from the catalog

curtbushko:
Possible fix for: 

https://github.com/hashicorp/consul-k8s/issues/2085 
https://github.com/hashicorp/consul-k8s/issues/2065 
https://github.com/hashicorp/consul/issues/15908 
https://github.com/hashicorp/consul-k8s/issues/1935 
https://github.com/hashicorp/consul-k8s/issues/1817 

which are all the same issue of ghost services left in the catalog when nodes are recycled. 

Associated issue: https://github.com/hashicorp/consul-k8s/issues/2491

Changes proposed in this PR:

- In the past, kubernetes nodes were used as the source of truth to determine the list of services that should exist in Consul. 
- In most cases this was ok but becomes a problem when nodes are quickly deleted from kubernetes such as the case when using spot instances.
- Instead, use consul synthetic-nodes to get the list of services and deregister the services that do not have endpoint addresses.

How I've tested this PR:

- Ran and updated the unit tests so that they pass. This required adding synthetic node meta data to many of the tests.
- Acceptance tests
- Manually tested using my [demo repo](https://github.com/curtbushko/cni-demo/tree/main/consul-on-kind-catalog). 
  - Created a multi-node kind cluster, deployed several static apps to it with a nodeSelector and deleted nodes. 
  - When the node is deleted, endpoints controller detects an endpoint change and deregisters
  - Adding the node back re-registers the service
  - Note: You can simulate node deletions using a combination of `kubectl delete node` + `docker stop <image>`. `docker start <image>` will re-register the node.

:eyes:

Checklist:
- [x] Tests added

### DIFF
--- a/.changelog/2571.txt
+++ b/.changelog/2571.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+control-plane: fix bug in endpoints controller when deregistering services from consul when a node is deleted.
+```
--- a/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
@@ -159,7 +159,6 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	err = r.Client.Get(ctx, req.NamespacedName, &serviceEndpoints)
-
 	// endpointPods holds a set of all pods this endpoints object is currently pointing to.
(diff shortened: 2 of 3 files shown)


mr-miles:
@curtbushko - this is great, thanks for your work on this PR.

natemollica-nm:
@curtbushko I take it we can't backport this to 0.49.x since it's out of the Support window version wise?

PR #2572, opened 2023-07-14
Support running with restricted PSA enforcement enabled (part 1)

pglass:

Support restricted PSA enforcement in a basic setup. This is enough to get a basic setup working and an acceptance test passing (but does not update every component).

This enables running Consul in a basic configuration with PSA enforcement set to restricted on the namespace where Consul is deployed. (This requires deploying the CNI to a different privileged namespace).

On OpenShift, we have the option to set the security context or not. If the security context is unset, then it is set automatically by OpenShift SCCs. However, we prefer to set the security context to avoid useless warnings on OpenShift and to reduce the config difference between OpenShift and plain Kube. By default, OpenShift namespaces have the audit and warn PSA labels set to restricted, so we receive pod security warnings when deploying Consul to OpenShift even though the pods will be able to run.

### Helm chart changes

* Add a helper to define a "restricted" container security context (when pod security policies are not enabled)
* Update the following container securityContexts to use the "restricted" settings (not exhaustive)

  - gateway-cleanup-job.yaml
  - gateway-resources-job.yaml
  - gossip-encryption-autogenerate-job.yaml
  - server-acl-init-cleanup-job.yaml - only if `.Values.server.containerSecurityContext.server.acl-init` is unset
  - server-acl-init-job.yaml - only if `.Values.server.containerSecurityContext.server.acl-init` is unset
  - server-statefulset.yaml:
     - the locality-init container receives the restricted context
     - the consul container receives the restricted context only if `.Values.server.containerSecurityContext.server` is unset
  - tls-init-cleanup-job.yaml - only if `.Values.server.containerSecurityContext.server.tls-init` is unset
  - tls-init-job.yaml - only if `.Values.server.containerSecurityContext.server.tls-init` is unset
  - webhook-cert-manager-deployment.yaml

### Acceptance test changes

* When `-enable-openshift` and `-enable-cni` are set, configure the CNI
  settings correctly for OpenShift, which must look like:

  ```
  connectInject:
    cni:
      enabled: true
      multus: true
      cniBinDir: /var/lib/cni/bin
      cniNetDir: /etc/kubernetes/cni/net.d
  ```

* Add the `-enable-restricted-psa-enforcement` test flag. When this is set,
  the tests assume the Consul namespace has restricted PSA enforcement enabled.
  The tests will deploy the CNI (if enabled) into the `kube-system` namespace.
  Compatible test cases will deploy applications outside of the Consul namespace.
* Update the ConnectHelper to configure the NetworkAttachmentDefinition
  required to be compatible with the CNI on OpenShift.
* Add fixtures for static-client and static-server for OpenShift. This
  is necessary because the deployment configs must reference the network
  attachment definition when using the CNI on OpenShift.
* Update tests in the `acceptance/tests/connect` directory to either
  run or skip based on -enable-cni and -enable-openshift

## How I've tested this PR:

1. Run OpenShift 4.12 locally with CRC. Follow these steps, but give more memory to CRC (`crc start -m 18432`): https://developer.hashicorp.com/consul/tutorials/kubernetes/kubernetes-openshift-red-hat#crc-setup. You will need a non-latest version of CRC to get OpenShift 4.12. (otherwise you'll have 4.13+).

```console
$ crc version
WARN A new version (2.23.0) has been published on https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/2.23.0/crc-macos-installer.pkg
CRC version: 2.19.0+a71226
OpenShift version: 4.12.13
Podman version: 4.4.4
$ crc setup
$ crc start -m 18432
```

2. Login into OpenShift as the kubeadmin:

```console
$ eval $(crc oc-env)
$ crc console --credentials
To login as a regular user, run 'oc login -u developer -p developer https://api.crc.testing:6443'.
To login as an admin, run 'oc login -u kubeadmin -p <password> https://api.crc.testing:6443'
$ oc login -u kubeadmin -p <password> https://api.crc.testing:6443
```

3. I used this script to start the tests. Place this in the `acceptance` directory, and fix the CONSUL_LICENSE environment variable. You may want to increase the test timeout. The script configures three namespaces (cni, consul, app) and sets their PSA enforcement levels appropriately. Then it runs the tests 

<details>
<summary>Script</summary>

```shell
#!/usr/bin/env bash

set -euo pipefail

SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )

export CONSUL_LICENSE=$(cat ~/.consul-ent-license)
export CONSUL_ENT_LICENSE=$CONSUL_LICENSE

# Delete any leftover acceptance test namespaces.
oc get ns | grep ^acceptance | awk '{print $1}' | xargs -n 1 oc delete project || true

function runtest() {
    set -x
    ns_base="acceptance-$1-$RANDOM"

    consul_namespace="${ns_base}-consul"
    oc new-project $consul_namespace

    kubectl label --overwrite ns $consul_namespace \

    cd "${SCRIPT_DIR}/tests/$1"
    rm -rf ./_debug
    mkdir ./_debug
    go test  -v -p 1 -timeout 15m -failfast \
        -consul-k8s-image 'ghcr.io/pglass/consul-k8s-control-plane-dev:ubi' \
        -consul-image 'hashicorp/consul-enterprise:1.16.0-ent-ubi' \
        -debug-directory ./_debug \
        -enable-enterprise \
        -kube-contexts crc-admin \
        -kube-namespaces "$consul_namespace" \
        -enable-openshift \
        -enable-transparent-proxy \
        -enable-cni \
        -enable-restricted-psa-enforcement \
        ./...
}

runtest "connect"
```

</details>

Idk, run OpenShift and try the instructions above if you dare.

Checklist:
- [x] Tests added


acceptance/framework/environment/environment.go
```diff
@@ -59,6 +61,20 @@ func NewKubernetesEnvironmentFromConfig(config *config.TestConfig) *KubernetesEn
 		kenv.contexts[SecondaryContextName] = NewContext(config.SecondaryKubeNamespace, config.SecondaryKubeconfig, config.SecondaryKubeContext)
 	}
 
+	// Optionally, deploy apps into a separate namespace.
+	// Maybe not the right place for this?
```
ishustava, line 65: If this will always be the `consul` namespace then wouldn't we want to deploy the helm chart as well as applications into the restricted one? Or do you think we'll be deploying the helm chart into one and the applications into another one. Also, if it's always the same, we could also consider hard-coding it and setting it whenever `enable-openshift` flag is set.

acceptance/tests/fixtures/cases/static-client-openshift-inject/patch.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: Deployment
+metadata:
+  name: static-client
+spec:
+  template:
+    metadata:
+      annotations:
+        "consul.hashicorp.com/connect-inject": "true"
+        "consul.hashicorp.com/connect-service-upstreams": "static-server:1234"
+        "k8s.v1.cni.cncf.io/networks": '[{ "name":"consul-cni" }]'
```
ishustava, line 14: I think at some point we've added some openshift specific SCCs to static-server or static-client base cases (or both) and with this change, would it make sense to clean those up or do we still need them?

acceptance/framework/environment/environment.go
ishustava: Oh also, maybe of the `*_namespaces*` tests already deploy apps in their own namespaces so that would probably interfere with these settings as they will get overridden by the specific tests. For example: https://github.com/hashicorp/consul-k8s/blob/c790951382235e9a5f10365d1b46fba16c5f2e5a/acceptance/tests/connect/connect_inject_namespaces_test.go#L90-L99

acceptance/tests/fixtures/bases/openshift/network-attachment.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: consul-cni
+spec:
+  config: '{
+            "cniVersion": "0.3.1",
+            "type": "consul-cni",
+            "cni_bin_dir": "/var/lib/cni/bin",
+            "cni_net_dir": "/etc/kubernetes/cni/net.d",
+            "kubeconfig": "ZZZ-consul-cni-kubeconfig",
+            "log_level": "info",
```
curtbushko, line 12: It won't give very much logging output but please set this to 'debug' instead of 'info'.

charts/consul/templates/_helpers.tpl
```diff
@@ -15,6 +15,19 @@ as well as the global.name setting.
 {{- end -}}
 {{- end -}}
 
+{{- define "consul.restrictedSecurityContext" -}}
+{{- if not .Values.global.enablePodSecurityPolicies -}}
+securityContext:
```
curtbushko, line 20: This is a great way to do this! I like it!

charts/consul/templates/server-statefulset.yaml
```diff
@@ -238,6 +238,7 @@ spec:
         volumeMounts:
           - name: extra-config
             mountPath: /consul/extra-config
+        {{- include "consul.restrictedSecurityContext" . | nindent 8 }}
```
curtbushko, line 241: Please add a bats test to test consul.restrictedSecurityContext. Even though the definition is in _helpers we still want to test it at least once and server-statefulset is one of the more common resources to test helpers in. No need to add bats tests for all the other files that use consul.restrictedSecurityContext.

curtbushko (changes requested):
Thanks Paul! 

This is looking pretty good overall but I have some questions/concerns/nits about configs and flags (Iryna does too). If you want to test OpenShift nuances specifically such as SCCs/PSA then you might even want to consider making a specific OpenShift test that does cross namespace magic instead of changing the behaviour of all the tests and how where they are installed. My worry is that you will end up chasing dozens of acceptance test bugs because of hardcoded namespaces in the tests themselves...

acceptance/framework/environment/environment.go
```diff
@@ -59,6 +61,20 @@ func NewKubernetesEnvironmentFromConfig(config *config.TestConfig) *KubernetesEn
 		kenv.contexts[SecondaryContextName] = NewContext(config.SecondaryKubeNamespace, config.SecondaryKubeconfig, config.SecondaryKubeContext)
 	}
 
+	// Optionally, deploy apps into a separate namespace.
```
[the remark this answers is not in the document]
pglass: This is true and I wasn't sure about putting this here, but your comment doesn't feel like sufficient justification because we are already associating namespaces with the existing contexts?

I would need to rework this anyway though to match the recent changes to support N clusters.

acceptance/framework/config/config.go
```diff
@@ -33,14 +34,18 @@ type TestConfig struct {
 	SecondaryKubeContext   string
 	SecondaryKubeNamespace string
 
+	AppNamespace          string
```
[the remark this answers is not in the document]
pglass: > For CNINamespace, we made the design decision that CNI was always installed alongside consul. Othere CNI plugins do it differently and install their plugins in their own namespace. I don't think OpenShift needs the CNI to be installed somewhere else to work?

Right. But if someone wants to run Consul in a namespace with restricted PSA enforcement, then they can't run the CNI in that same namespace because the CNI requires privilege. Eventually, (we assume) restricted PSA enforcement is going to be the OpenShift default, so the goal here was to enable testing with that potential future configuration.

> I'd make AppNamespace/SecondaryAppNamespace test specific or OpenShift specific if needed.

This is still relevant to vanilla K8s because you can enable restricted PSA enforcement there as well (although it likely will never be the default like on OpenShift). So I didn't want to make the app namespace settings OpenShift-specific, because it's useful and (and easier) to be able to test this same configuration on vanilla K8s as well.

acceptance/framework/connhelper/connect_helper.go
```diff
@@ -108,23 +119,47 @@ func (c *ConnectHelper) DeployClientAndServer(t *testing.T) {
...
+	ctx := c.appCtx()
+	opts := ctx.KubectlOptions(t)
+
+	if c.Cfg.EnableCNI && c.Cfg.EnableOpenshift {
+		// On OpenShift with the CNI, we need to create a network attachment definition in the namespace
+		// where the applications are running, and the app deployment configs need to reference that network
+		// attachment definition.
+
+		// TODO: A base fixture is the wrong place for these files
+		k8s.KubectlApply(t, opts, "../fixtures/bases/openshift/")
```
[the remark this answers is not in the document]
pglass: I'm not really clear what you mean. This is in `ConnectHelper` so how does it know the namespace to use?

Right now, it's using a namespace sourced from the test flags via the context (bc that's what I set up). Do you mean I should instead have each test pass a namespace to `ConnectHelper` if needed?

acceptance/tests/fixtures/cases/static-client-openshift-inject/patch.yaml
ishustava: The privileged SCC is needed for when you're not using CNI and so your init container needs elevated privileges to run iptables

acceptance/framework/connhelper/connect_helper.go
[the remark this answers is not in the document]
pglass: I tried something else:

* I added `ConnectHelper.UseAppNamespace = true|false` option
   * If true then it creates, configures, and deploys into a namespace named `<consul-namespace>-apps`.

This way there is a shared/common way to easily create an app namespace that tests can use if needed and there are no more `-app-namespace` flags.

I also added a `-restricted-psa-enforcement-enabled` flag and tests key off of that to determine whether to deploy apps into a separate namespace. (I don't know if I really like this, so I'm open to alternatives)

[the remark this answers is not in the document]
pglass: (Also, I accidentally did all of this while rebasing, so the latest commit that looks like a merge actually contains the changes - sorry!)

acceptance/framework/config/config.go
```diff
@@ -134,10 +136,22 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
...
+		if t.EnableOpenshift {
+			setIfNotEmpty(helmValues, "connectInject.cni.multus", "true")
+			setIfNotEmpty(helmValues, "connectInject.cni.cniBinDir", "/var/lib/cni/bin")
+			setIfNotEmpty(helmValues, "connectInject.cni.cniNetDir", "/etc/kubernetes/cni/net.d")
+		}
+
+		if t.RestrictedPSAEnforcementEnabled {
+			// The CNI requires privilege, so when restricted PSA enforcement is enabled on the Consul
+			// namespace it must be run in a different privileged namespace.
+			setIfNotEmpty(helmValues, "connectInject.cni.namespace", "kube-system")
```
pglass, line 153: Deploying into a privileged namespace (`kube-system`) that already exists seems to work. This allowed me to remove the `-cni-namespace`.

acceptance/framework/environment/environment.go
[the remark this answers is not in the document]
pglass: I did remove the app namespace flags, and no longer use a separate context - so resolving this thread (but leaving others open for discussion)

acceptance/framework/config/config.go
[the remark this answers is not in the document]
pglass: I replaced all these flags with `-restricted-psa-enforcement-enabled` (default false). When true:

1. The Consul namespace(s) passed to the tests via `-kube-namespaces` need to be configured with the restricted PSA enforcement label.
2. The CNI is deployed into the `kube-system` namespace, which is privileged and already exists (so I removed the -cni-namespace flag)
3. I removed the app namespace flags. Instead, ConnectHelper supports a `UseAppNamespace` setting that tests can optionally use (based on whether `-restricted-psa-enforcement-enabled` is set)

Let me know what you think.

acceptance/framework/environment/environment.go
curtbushko: >  My thought was those tests would ultimately use the specified app + app secondary namespaces (if specified) instead of the hard-coded ones.

What happens if a test wants a 3rd app installed in another namespace? 4th? 

We are already seeing a similar thing show up with contexts in the partition tests where we now need 3 clusters and multiple services/namespaces setup...

pglass: That's true. In any case, I've removed the app namespace flags, so that tests can choose what namespaces they use, so resolving this thread.

acceptance/tests/fixtures/bases/openshift/network-attachment.yaml
pglass: I set this to debug.

charts/consul/templates/server-statefulset.yaml
pglass: I updated the bats tests for server-statefulset

acceptance/framework/connhelper/connect_helper.go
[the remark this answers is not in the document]
curtbushko: I like the `-restricted-psa-enforcement-enabled` flag. I think this is a good + global way to do things as I could see us wanting to test this on all the clouds in the future.

acceptance/framework/flags/flags.go
```diff
@@ -115,6 +116,13 @@ func (t *TestFlags) init() {
 	flag.BoolVar(&t.flagEnableCNI, "enable-cni", false,
 		"If true, the test suite will run tests with consul-cni plugin enabled. "+
 			"In general, this will only run against tests that are mesh related (connect, mesh-gateway, peering, etc")
+	flag.BoolVar(&t.flagRestrictedPSAEnforcementEnabled, "restricted-psa-enforcement-enabled", false,
```
curtbushko, line 119: Nit: This flag should be flipped the other way to match the patterns to maybe `-enable-restricted-psa-enforcement`?

I think setting PSA as an 'infra' thing is fine as we have to do the same with CNI; there are specific setups for Kind and the clouds.

curtbushko (approved):
I have a request to change the flag name for PSA enable but I won't block the PR for that.

Feel free to change it in another PR if you want so that you can move onto part 2.

Thanks Paul!

acceptance/framework/flags/flags.go
pglass: Oh, do you mean change the flag name and also have the tests create and configure the consul namespace with PSA enforcement? (rather than requiring the namespace be configured with PSA enforcement before the test is run)

pglass: I changed this flag name (but kept the behavior of needing to set the restricted label on the namespace).

PR #2579, opened 2023-07-17
Mw/net 4260 phase 2 automate the k8s sameness tests

wilkermichael:
*Note:* I've broken up the PR into logical commits and the commit messages contain extra details. So the reviewer can go through this PR commit by commit to make it more manageable.

Changes proposed in this PR:
- This PR adds acceptance tests for sameness failover
    - Tests:
       - service failover
       - prepared query failover (I also technically use DNS to test PQ failover)
       - dns failover  
- The sameness tests will only run in kind because they require 4 clusters to adequately test
- This PR is part 1 of 2, this will, for the most part, only cover the partition failover test case. I will be adding peering scenarios in the next PR.
- To support this PR I need to make changes to the workflow, the commit marked "drop" will be dropped before merging

How I've tested this PR:
- Ran acceptance tests in pipeline and locally

👀 

Checklist:
- [ ] Tests added


acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,393 @@
...
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	secondaryPartition    = "ap1"
+	primaryServer         = "dc1"
```
thisisnotashwin, line 24:
```suggestion
	primaryServerDatacenter        = "dc1"
```

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,393 @@
...
+	peer1                 = "dc2"
+	peer2                 = "dc2"
```
thisisnotashwin, line 26: similarly add Datacenter as a suffix so that it is most technically accurate?

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,393 @@
...
+	keyPrimaryServer = "server"
+	keyPartition     = "partition"
+	keyPeer1         = "peer1"
+	keyPeer2         = "peer2"
+
+	staticServerDeployment = "deploy/static-server"
+	staticClientDeployment = "deploy/static-client"
+
+	dnsLookup   = "static-server.service.ns2.ns.mine.sg.consul"
+	dnsPQLookup = "my-query.query.consul"
```
thisisnotashwin, line 39: IMO these should be defined as constants closer to where they get used so that it is an easier lookup instead of them being global constants.

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,393 @@
...
+				logger.Logf(t, "retrieving partition token secret %s from the server cluster and applying to the client cluster", partitionToken)
+				k8s.CopySecret(t, members[keyPrimaryServer].context, members[keyPartition].context, partitionToken)
+			}
+
+			partitionServiceName := fmt.Sprintf("%s-consul-expose-servers", releaseName)
+			partitionSvcAddress := k8s.ServiceHost(t, cfg, members[keyPrimaryServer].context, partitionServiceName)
+
+			k8sAuthMethodHost := k8s.KubernetesAPIServerHost(t, cfg, members[keyPartition].context)
+
+			secondaryPartitionHelmValues := map[string]string{
```
thisisnotashwin, line 143: in this case where we have peers and partitions, i think it would be a cleaner read to specify that this is also in the same datacenter as the `primaryServer` explicitly

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,393 @@
...
+				}
+			}
+
+			logger.Log(t, "creating exported services")
+			k8s.KubectlApplyK(t, members[keyPrimaryServer].context.KubectlOptions(t), "../fixtures/cases/sameness-exported-services/default-partition")
+			k8s.KubectlApplyK(t, members[keyPartition].context.KubectlOptions(t), "../fixtures/cases/sameness-exported-services/ap1-partition")
+			helpers.Cleanup(t, cfg.NoCleanupOnFailure, func() {
+				k8s.KubectlDeleteK(t, members[keyPrimaryServer].context.KubectlOptions(t), "../fixtures/cases/sameness-exported-services/default-partition")
+				k8s.KubectlDeleteK(t, members[keyPartition].context.KubectlOptions(t), "../fixtures/cases/sameness-exported-services/ap1-partition")
+			})
```
thisisnotashwin, line 283: we can replace these with `appltResources`

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func applyResources(t *testing.T, isCleanupOnFailure bool, kustomizeDir string, opts *terratestk8s.KubectlOptions) {
+	k8s.KubectlApplyK(t, opts, kustomizeDir)
+	helpers.Cleanup(t, isCleanupOnFailure, func() {
+		k8s.KubectlDeleteK(t, opts, kustomizeDir)
+	})
+}
+
+func serviceFailoverCheck(t *testing.T, expectedClusterName string, server *member) {
```
thisisnotashwin, line 360: can you add a comment here as to what exactly is being verified?

thisisnotashwin (commented):
Overall, this looks excellent. I really lke the detail. I have a broader comment on the file structure for the fixtures. For ease of management, can we create a top level folder called sameness under `bases` and `cases` and move all the fixtures introduces into subfolders under `sameness`. I feel like this is causing a sprawl in the folder structure making it slightly tricker to navigate.

acceptance/tests/sameness/sameness_test.go
wilkermichael: ok, I added a comment

thisisnotashwin: sure. i was thinking we can just explicitly specify the dc in the values file.

acceptance/tests/fixtures/bases/sameness/override-ns/payments-service-resolver.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: consul.hashicorp.com/v1alpha1
```
thisisnotashwin, line 1: nit: is the filename supposed to be `payment-service`?

acceptance/tests/fixtures/bases/exportedservices-ap1/exportedservices-ap1.yaml
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+apiVersion: consul.hashicorp.com/v1alpha1
```
thisisnotashwin, line 4: can we move this into `sameness` folder as well?

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,443 @@
...
+				k8s.CopySecret(t, members[keyPrimaryServer].context, members[keyPartition].context, partitionToken)
+			}
+
+			partitionServiceName := fmt.Sprintf("%s-consul-expose-servers", releaseName)
+			partitionSvcAddress := k8s.ServiceHost(t, cfg, members[keyPrimaryServer].context, partitionServiceName)
+
+			k8sAuthMethodHost := k8s.KubernetesAPIServerHost(t, cfg, members[keyPartition].context)
+
+			secondaryPartitionHelmValues := map[string]string{
+				"global.enabled": "false",
```
thisisnotashwin, line 192:
```suggestion
				"global.enabled":     "false",
				"global.datacenter": primaryServerDatacenter,
```

thisisnotashwin: something like this

thisisnotashwin (approved):
Have a few minor nits but overall, this is excellent work!

acceptance/tests/sameness/sameness_test.go
wilkermichael: weird, I lint on save

acceptance/tests/sameness/sameness_test.go
wilkermichael: ah okay, added

acceptance/tests/sameness/sameness_test.go
wilkermichael: oh it's because there is an empty newline, gofmt won't space the values if there's an empty newline

acceptance/tests/fixtures/cases/sameness/static-client/default/patch.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: static-client
+spec:
+  template:
+    metadata:
+      annotations:
+        'consul.hashicorp.com/connect-inject': 'true'
+        consul.hashicorp.com/connect-service-upstreams: "static-server.ns2.default:8080"
```
zalimeni, line 13: ~ More for my own understanding: is there a reason the first key is quoted, and the second one isn't? (same w/ other similar patch files)

acceptance/tests/sameness/main_test.go
```diff
@@ -0,0 +1,31 @@
...
+func TestMain(m *testing.M) {
+	suite = testsuite.NewSuite(m)
+
+	expectedNumberOfClusters := 4
+
+	if suite.Config().EnableMultiCluster && suite.Config().IsExpectedClusterCount(expectedNumberOfClusters) && suite.Config().UseKind {
+		os.Exit(suite.Run())
+	} else if suite.Config().EnableMultiCluster && suite.Config().IsExpectedClusterCount(expectedNumberOfClusters) {
+		fmt.Println(fmt.Sprintf("Skipping sameness tests because either -enable-multi-cluster is "+
+			"not set or the number of clusters did not match the expected count of %d", expectedNumberOfClusters))
```
zalimeni, line 25: It seems the only difference between this and the prior conditional is the first one also checks `&& suite.Config().UseKind`. I think maybe the `&&` here should be an `||`?

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,445 @@
...
+					metav1.ListOptions{LabelSelector: labelSelector})
+				require.NoError(t, err)
+				require.Len(t, podList.Items, 1)
+				require.Len(t, podList.Items[0].Spec.Containers, 2)
+				if labelSelector == "app=static-server" {
+					ip := &podList.Items[0].Status.PodIP
+					require.NotNil(t, ip)
+					logger.Logf(t, "default-static-server-ip: %s", *ip)
+					members[keyPartition].staticServerIP = ip
+				}
```
zalimeni, line 324: 1. Are the IP logs here swapped w.r.t. default vs. partition?
2. (Minor) ~ Would it make sense to break this code out into a helper?

zalimeni (approved):
A few minor questions but overall this looks great, and is really educational to me trying to write my own tests - thank you! (Consider me 👍🏻 once you're satisfied w/ any comment outcomes.)

acceptance/tests/sameness/main_test.go
wilkermichael: hmm yeah, I think I was initially trying to make the logging more meaningful, but I can do that with a single log. I've replaced it with a single `if...else` with:

```Go
		fmt.Println(fmt.Sprintf("Skipping sameness tests because either -enable-multi-cluster is "+
			"not set, the number of clusters did not match the expected count of %d, or --useKind is false. "+
			"Sameness acceptance tests are currently only suopported on Kind clusters", expectedNumberOfClusters))
```

zalimeni: That seemed reasonable to me! Just the logic seemed slightly off - but either way, sounds good 👍🏻

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,445 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	primaryDatacenterPartition = "ap1"
+	primaryServerDatacenter    = "dc1"
+	peer1Datacenter            = "dc2"
+	peer2Datacenter            = "dc2"
```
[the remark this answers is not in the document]
wilkermichael: good catch, this should be dc3

acceptance/tests/sameness/sameness_test.go
wilkermichael: yes you're right about the log swap. I'm not 100% sure what I'm going to do with this when I add peering as well, so that's why I didn't add a helper yet. I'll consider it when I add in peering!

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,445 @@
...
+		// ;; SERVER: <dns-ip>#<dns-port>(<dns-ip>)
+		// ;; WHEN: Mon Aug 10 15:02:40 UTC 2020
+		// ;; MSG SIZE  rcvd: 98
+		//
+		// We assert on the existence of the ANSWER SECTION, The consul-server IPs being present in
+		// the ANSWER SECTION and the DNS IP mentioned in the SERVER: field
+
+		assert.Contains(r, logs, fmt.Sprintf("SERVER: %s", dnsIP))
+		assert.Contains(r, logs, "ANSWER SECTION:")
+		assert.Contains(r, logs, *failover.staticServerIP)
```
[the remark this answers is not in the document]
wilkermichael: TBH, I think that just the IP being present at all is probably sufficient, I copied/pasted these checks from another DNS test in consul-k8s: https://github.com/hashicorp/consul-k8s/blob/7bb0a57aec89dbdab4a6313536ae4de540e43d5f/acceptance/tests/consul-dns/consul_dns_test.go#L92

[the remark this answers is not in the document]
zalimeni: Cool cool. I think you're probably right 👍🏻 SGTM

acceptance/tests/fixtures/cases/sameness/static-client/default/patch.yaml
wilkermichael: no idea actually 😓 these are from Eric's demo

zalimeni: Cool, and totally not a blocker! Seems another is double-quoted, so probably just copy 🍝

acceptance/tests/sameness/sameness_test.go
```diff
@@ -0,0 +1,445 @@
...
+				|    |  |                  +-------+--------+-----------------+------>|                  |        |
+				|    |  |  Static-Server   |       |        |                 |       |  Static-Server   |        |
+				|    |  |                  +-------+---+    |                 |       |                  |        |
+				|    |  |                  |       |   |    |                 |       |                  |        |
+				|    |  |                  |       |   |    |                 |       |                  |        |
+				|    |  |                  +-------+---+----+-------------+   |       |                  |        |
+				|    |  +------------------+       |   |    |             |   |       +------------------+        |
+				|    |  Admin Partitions: Default  |   |    |             |   |                                   |
+				|    |  Name: cluster-01-a         |   |    |             |   |     Admin Partitions: Default     |
+				|    |                             |   |    |             |   |     Name: cluster-03-a            |
```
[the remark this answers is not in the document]
wilkermichael: I'm not sure what you mean exactly? I'm trying to convey that `cluster-01-a` will correspond with the `default` partition and `cluster-01-b` will correspond to the `ap1` partition, both in DC1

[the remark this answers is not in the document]
wilkermichael: oh nevermind, I was looking at the wrong line, yeah I think that makes sense, I'll flip them

acceptance/framework/k8s/kubectl.go
```diff
@@ -16,6 +17,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	kubectlTimeout = "--timeout=90s"
```
curtbushko, line 21: I would go to 120s as I bet this causes a LOT of tests to fail!

acceptance/tests/fixtures/cases/sameness/static-server/default/patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: static-server
+spec:
+  template:
+    metadata:
+      annotations:
+        "consul.hashicorp.com/connect-inject": "true"
+    spec:
+      containers:
+        - name: static-server
+          image: hashicorp/http-echo:latest
```
curtbushko, line 16: This should probably be: `docker.mirror.hashicorp.services/hashicorp/http-echo:alpine` or `docker.mirror.hashicorp.services/hashicorp/http-echo:latest` to avoid pull errors in kind.

I think there were problems with `latest` not working on M1 laptops.

acceptance/tests/fixtures/cases/sameness/static-server/partition/patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: static-server
+spec:
+  template:
+    metadata:
+      annotations:
+        "consul.hashicorp.com/connect-inject": "true"
+    spec:
+      containers:
+        - name: static-server
+          image: hashicorp/http-echo:latest
```
curtbushko, line 16: This should probably be: docker.mirror.hashicorp.services/hashicorp/http-echo:alpine or docker.mirror.hashicorp.services/hashicorp/http-echo:latest to avoid pull errors in kind.

I think there were problems with latest not working on M1 laptops.

curtbushko (approved):
Looks good! 

I have some non-critical things that should be fixed as they will cause random pipeline errors but I will not block you on merging.

acceptance/tests/fixtures/cases/sameness/static-server/default/patch.yaml
wilkermichael: ah okay, I copied this from existing tests, we might want to go through and change those as well

PR #2584, opened 2023-07-18
test(eks): fix deprecated CSI driver terraform

DanStough:
Changes proposed in this PR:
- Replacing the deprecated [`resolve_conflicts`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon#resolve_conflicts) with the new attributes. I don't know if we really need this setting since it is optional and the addon has no user-defined config, but I'm keeping this to keep the behavior consistent.

How I've tested this PR: I did not.

Checklist:
- [ ] ~Tests added~


curtbushko (approved):
LGTM!

DanStough:
I just realized these only run against main, so I'm removing the backport labels.

PR #2592, opened 2023-07-19
Add a check to prevent a nil-pointer dereference on Ingress LB

thisisnotashwin:
Changes proposed in this PR:
- Add a check to prevent a nil-pointer dereference on Ingress LB

How I've tested this PR:
- Deployed an improper Ingress config and validated that it does not crash the sync catalog.

- 👀


curtbushko (approved):
LGTM!

PR #2599, opened 2023-07-19
Differentiate FIPS linux package names

skpratt:
This avoids a name collision in linux package managers, allowing people to get non-FIPSs with i.e. "apt-get install consul-k8s" and FIPS with "apt-get install consul-k8s-fips


zalimeni (approved):
LGTM, tagging in @wilkermichael for some expert 👀 since I’m still coming up to speed on k8s releases

PR #2409, opened 2023-06-20
Fix GatewayClassConfig Test Timing Issue

t-eckert:
Changes proposed in this PR:
- Add retryCheckWithWait func
- Fix retry timing on GatewayClassConfig test

How I've tested this PR:
- Running Acceptance tests locally

How I expect reviewers to test this PR:

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #2631, opened 2023-07-21
Fix TestAPIGateway_GatewayClassConfig

pglass:
Changes proposed in this PR:

* Revert the Namespace creation/cleanup attempt in TestAPIGateway_GatewayClassConfig because it uses the default namespace which should exist

How I've tested this PR:

* Ran the test in isolation locally on Kind
* CI tests

👀 

Checklist:
- [ ] Tests added


curtbushko (approved):

acceptance/tests/api-gateway/api_gateway_gatewayclassconfig_test.go
```diff
@@ -144,10 +128,6 @@ func TestAPIGateway_GatewayClassConfig(t *testing.T) {
 	logger.Log(t, "creating controlled gateway")
 	gateway := createGateway(t, k8sClient, gatewayName, namespace, gatewayClassName, certificateName)
 
-	// make sure it exists
-	logger.Log(t, "checking that gateway one is synchronized to Consul")
-	checkConsulExists(t, consulClient, api.APIGateway, gatewayName)
-
```
pglass, line 150: We combed through https://github.com/hashicorp/consul-k8s/pull/2481 and removed anything that looked like it was added accidentally.

The lines in this file were previously removed in https://github.com/hashicorp/consul-k8s/pull/2409

acceptance/tests/api-gateway/api_gateway_gatewayclassconfig_test.go
20sr20, line 1: Looks good.  Thanks for reverting the accidentally added one.

20sr20 (approved):

pglass:
Note: We believe the acceptance test failure is due to a race condition in API GW. Merging this and we can follow up with additional fixes as need.

PR #2635, opened 2023-07-24
Job shutdown acceptance test

trevorLeonHC:
Changes proposed in this PR:
- Added test for making sure that graceful shutdown works with jobs and that requests are still possible during grace period (if set)
- Made changes to connhelper to allow for creation of test job. 

How I've tested this PR:
Running acceptance tests. 

Running acceptance tests. 

Checklist:
- [x] Tests added


acceptance/framework/k8s/deploy.go
```diff
@@ -72,6 +73,37 @@ func DeployKustomize(t *testing.T, options *k8s.KubectlOptions, noCleanupOnFailu
 	RunKubectl(t, options, "wait", "--for=condition=available", "--timeout=5m", fmt.Sprintf("deploy/%s", deployment.Name))
 }
```
curtbushko, line 75: Missing comment.

This function is bigger so I would keep it as a full function.

acceptance/framework/connhelper/connect_helper.go
```diff
@@ -167,6 +230,17 @@ func (c *ConnectHelper) TestConnectionSuccess(t *testing.T) {
 	}
 }
 
+// Same as TestConnectionSuccess for job-client instead of static-client.
```
curtbushko, line 233: This one could use varadic also.

acceptance/framework/k8s/deploy.go
```diff
@@ -83,6 +115,11 @@ func CheckStaticServerConnection(t *testing.T, options *k8s.KubectlOptions, sour
 
 	CheckStaticServerConnectionMultipleFailureMessages(t, options, sourceApp, expectSuccess, failureMessages, expectedSuccessOutput, curlArgs...)
 }
+func CheckStaticServerConnectionJob(t *testing.T, options *k8s.KubectlOptions, sourceApp string, expectSuccess bool, failureMessages []string, expectedSuccessOutput string, curlArgs ...string) {
```
curtbushko, line 118: This one could use varadic also.

acceptance/framework/k8s/deploy.go
```diff
@@ -123,6 +160,38 @@ func CheckStaticServerConnectionMultipleFailureMessages(t *testing.T, options *k
 		}
 	})
 }
+func CheckStaticServerConnectionMultipleFailureMessagesJob(t *testing.T, options *k8s.KubectlOptions, sourceApp string, expectSuccess bool, failureMessages []string, expectedSuccessOutput string, curlArgs ...string) {
```
curtbushko, line 163: Not needed if you use a varadic TestConnectionSuccess and pass the service through but you would need to change `CheckStaticServerConnectionMultipleFailureMessages` to determine if the sourceApp was the job.

Something like:

```
	resourceType := "deploy"
	if sourceApp == "job-client" {
		resourceType = "jobs"
	}

	args := []string{"exec", resourceType + sourceApp, "-c", sourceApp, "--", "curl", "-vvvsSf"}

```

acceptance/framework/k8s/deploy.go
```diff
@@ -142,6 +211,13 @@ func CheckStaticServerConnectionSuccessful(t *testing.T, options *k8s.KubectlOpt
 	logger.Logf(t, "Took %s to check if static server connection was successful", time.Since(start))
 }
 
+func CheckStaticServerConnectionSuccessfulJob(t *testing.T, options *k8s.KubectlOptions, sourceApp string, curlArgs ...string) {
```
curtbushko, line 214: This one could use varadic also.

acceptance/tests/connect/connect_proxy_lifecycle_test.go
```diff
@@ -212,3 +213,141 @@ func TestConnectInject_ProxyLifecycleShutdown(t *testing.T) {
 		})
 	}
 }
+
+func TestConnectInject_ProxyLifecycleShutdownJob(t *testing.T) {
+	cfg := suite.Config()
+	ver, err := version.NewVersion("1.2.0")
```
curtbushko, line 219: This version check and the corresponding cfg.ConsulDataplaneVersion.LessThan is not needed as all the proxy lifecycle changes have been backported to all of the release branches and and all consul-dataplane versions. It should work consul-dataplane/consul-k8s 1.x and up.

acceptance/tests/fixtures/bases/job-client/privileged-scc-rolebinding.yaml
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) HashiCorp, Inc.
```
curtbushko, line 1: Something went wrong with your rebase. You shouldn't have picked these up these openshift changes in your PR.

acceptance/tests/connect/connect_proxy_lifecycle_test.go
```diff
@@ -212,3 +213,141 @@ func TestConnectInject_ProxyLifecycleShutdown(t *testing.T) {
...
+				Cfg:         cfg,
+				HelmValues: map[string]string{
+					"connectInject.sidecarProxy.lifecycle.defaultShutdownGracePeriodSeconds": strconv.FormatInt(tc.gracePeriod, 10),
+					"connectInject.sidecarProxy.lifecycle.defaultEnabled":                    strconv.FormatBool(tc.gracePeriod > 0),
+				},
+			}
+
+			connHelper.Setup(t)
+			connHelper.Install(t)
+			connHelper.DeployJobAndServer(t)
```
curtbushko, line 251: I think this test needs to be slightly different as you are shouldn't need to install so many times. I am fine if this test only runs for the insecure scenario.

1) Install server only. Split DeployJobAndServer() into DeployServer() and DeployJob(name string).
2) Deploy job 1 (with override graceful annotation X), check connection, check shutdown
3) Deploy job2 (with override graceful annotation Y), check connection, check shutdown
4) Deploy job3 (with override graceful annotation Z), check connection, check shutdown

curtbushko (changes requested):
Congrats on your first PR to consul-k8s @trevorLeonHC !

This is looking pretty good but it will most likely need changes (the first PR always does!)

PR #2652, opened 2023-07-25
Consul client always logs into the local datacenter

thisisnotashwin:
Changes proposed in this PR:
- Update the CONSUL_LOGIN_DATACENTER logic for client daemonset to only login to the local datacenter as it uses a local authmethod and not a global authmethod.
- This led to errors when attempting to deploy clients in a federated cluster.

How I've tested this PR:
- @natemollica-nm was able to replicate the error and this fix fixed it in his federated cluster

- 👀 

Checklist:
- [x] Tests added

### DIFF
--- a/.changelog/2652.txt
+++ b/.changelog/2652.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+helm: fix CONSUL_LOGIN_DATACENTER for consul client-daemonset.
+```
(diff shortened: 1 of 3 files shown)


curtbushko (approved):
LGTM!

(that makes sense...how would a client/agent in dc2 be able to talk to dc1...)

PR #2656, opened 2023-07-25
Increase timeout for acl replication to 60 seconds and poll every 500 ms

thisisnotashwin:
Changes proposed in this PR:
- increase the max timeout for ACL replication to 60 secs from 2 secs. This is because in large federated clusters with external servers and high ping times, the ACL replication can take extended periods of time. This leads to certain operations like mesh-gateway startup timing out which has required users to introduce sleeps of up to 45 seconds to work around this. Increasing this timeout should ensure replication is successful.

How I've tested this PR:
- 👀 
- 👀


curtbushko (approved):
LGTM!

PR #2662, opened 2023-07-25
Do not set securityContext on Openshift < 4.11

pglass:
Changes proposed in this PR:

* Do not set securityContext on OpenShift < 4.11 because the `restricted` SCC disallows setting some of those settings. I broke this in https://github.com/hashicorp/consul-k8s/pull/2572. In OpenShift >= 4.11, setting the securityContext is okay because the new `restricted-v2` SCC is available to all users/accounts and allows setting the fields we want to set, and doing so avoids warnings.
* We determine the OpenShift version in Helm using the Kube version because there is a 1:1 mapping of OpenShift to Kube versions.

How I've tested this PR:

* Ran the connect tests with openshift enabled against 4.10 and 4.11 OpenShift clusters, using instructions similar to those in the description of https://github.com/hashicorp/consul-k8s/pull/2572

👀 

Checklist:
- [ ] Tests added

### DIFF
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -15,8 +15,23 @@ as well as the global.name setting.
 {{- end -}}
 {{- end -}}
 
+
 {{- define "consul.restrictedSecurityContext" -}}
 {{- if not .Values.global.enablePodSecurityPolicies -}}
+{{/*
+To be compatible with the 'restricted' Pod Security Standards profile, we
+should set this securityContext on containers whenever possible.
+
+In OpenShift < 4.11 the restricted SCC disallows setting most of these fields,
+so we do not set any for simplicity (and because that's how it was configured
+prior to adding restricted PSA support here). In OpenShift >= 4.11, the new
+restricted-v2 SCC allows setting these in the securityContext, and by setting
+them we avoid PSA warnings that are enabled by default.
+
+We use the K8s version as a proxy for the OpenShift version because there is a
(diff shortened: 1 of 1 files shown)


charts/consul/templates/_helpers.tpl
```diff
@@ -15,8 +15,23 @@ as well as the global.name setting.
...
+In OpenShift < 4.11 the restricted SCC disallows setting most of these fields,
+so we do not set any for simplicity (and because that's how it was configured
+prior to adding restricted PSA support here). In OpenShift >= 4.11, the new
+restricted-v2 SCC allows setting these in the securityContext, and by setting
+them we avoid PSA warnings that are enabled by default.
+
+We use the K8s version as a proxy for the OpenShift version because there is a
+1:1 mapping of versions. OpenShift 4.11 corresponds to K8s 1.24.x.
+*/}}
+{{- if (or (not .Values.global.openshift.enabled) (and (ge .Capabilities.KubeVersion.Major "1") (ge .Capabilities.KubeVersion.Minor "24"))) -}}
```
pglass, line 34: Technically, we don't need to backport this to consul-k8s 1.2.x because it only supports Kube >= 1.24

I was going to do so anyway because it is safe, and I do need to backport to 1.1.x and 1.0.x anyway (so the backport automation helps with that - but I can re-open this against those release branches. Happy to do it either way).

pglass:
Closed this in favor of https://github.com/hashicorp/consul-k8s/pull/2678

PR #2669, opened 2023-07-26
net-1776,  add job lifecycle test and changes to connhelper

trevorLeonHC:
Changes proposed in this PR:

Added test for making sure that graceful shutdown works with jobs and that requests are still possible during grace period (if set)
Made changes to connhelper to allow for creation of test job.
How I've tested this PR:
Running acceptance tests.

Running acceptance tests.

Checklist:
- [x] Tests added


acceptance/tests/connect/connect_proxy_lifecycle_test.go
```diff
@@ -205,3 +205,167 @@ func TestConnectInject_ProxyLifecycleShutdown(t *testing.T) {
...
+		pods, err := ctx.KubernetesClient(t).CoreV1().Pods(ns).List(
+			context.Background(),
+			metav1.ListOptions{
+				LabelSelector: "app=job-client",
+			},
+		)
+		require.NoError(t, err)
+		require.Len(t, pods.Items, 1)
+		jobName := pods.Items[0].Name
+
```
curtbushko, line 289: I think you might want to check to see if the job can talk to the server before you do the graceful shutdown. That way
you know the proxy is working.

acceptance/tests/connect/connect_proxy_lifecycle_test.go
```diff
@@ -205,3 +205,167 @@ func TestConnectInject_ProxyLifecycleShutdown(t *testing.T) {
...
+func TestConnectInject_ProxyLifecycleShutdownJob(t *testing.T) {
+	cfg := suite.Config()
+
+	defaultGracePeriod := 10
+
+	cases := map[string]int{
+		"../fixtures/cases/job-client-inject":                  defaultGracePeriod,
+		"../fixtures/cases/job-client-inject-grace-period-0s":  0,
+		"../fixtures/cases/job-client-inject-grace-period-30s": 30,
```
curtbushko, line 217: I think 30 seconds is much too long of a test. I'd do 10 as the max test, and 5 as the default.

acceptance/tests/fixtures/cases/job-client-inject-grace-period-30s/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+
+resources:
```
curtbushko, line 2: I left a similar comment somewhere else but I think 30 seconds is too long of a wait. Maybe change this to 10.

curtbushko (changes requested):

acceptance/framework/k8s/deploy.go
```diff
@@ -72,6 +73,33 @@ func DeployKustomize(t *testing.T, options *k8s.KubectlOptions, noCleanupOnFailu
...
+		// objects are deployments, and so when other tests create these
+		// they should have different pod names.
+		WritePodsDebugInfoIfFailed(t, options, debugDirectory, labelMapToString(job.GetLabels()))
+		KubectlDeleteK(t, options, kustomizeDir)
+	})
+	logger.Log(t, "job deployed")
+
+	// The timeout to allow for connect-init to wait for services to be registered by the endpoints controller.
+	//RunKubectl(t, options, "wait", "--for=condition=", "--timeout=5m", fmt.Sprintf("job/%s", job.Name))
+	time.Sleep(30 * time.Second)
```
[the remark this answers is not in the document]
t-eckert: Great call out. We can replace this with a wait that looks for the ready status of Pods created by the job:

```suggestion
	// Because Jobs don't have a "started" condition, we have to check the status of the Pods they create.
	RunKubectl(t, options, "wait", "--for=condition=Ready", "--timeout=5m", "pods", "--selector", fmt.Sprintf("job-name=%s", job.Name))

```

t-eckert (approved):
This looks great! I think you addressed most of @curtbushko's comments and we worked through things together.

curtbushko (approved):

PR #2675, opened 2023-07-27
Fix ordering of licence in templates

thisisnotashwin:


curtbushko (approved):
Thank you for fixing this!

PR #2678, opened 2023-07-27
Do not set securityContext on Openshift < 4.11

pglass:
Changes proposed in this PR:

- This is going into `release/1.1.x`, and will be backported into `release/1.0.x` because both of those Consul K8s versions support K8s < 1.23.x and therefore support OpenShift <= 4.10
- Do not set securityContext on OpenShift < 4.11 because the restricted SCC disallows setting some of those settings. I broke this in https://github.com/hashicorp/consul-k8s/pull/2572. In OpenShift >= 4.11, setting the securityContext is okay because the new restricted-v2 SCC is available to all users/accounts and allows setting the fields we want to set, and doing so avoids warnings.
- We determine the OpenShift version in Helm using the Kube version because there is a 1:1 mapping of OpenShift to Kube versions.

How I've tested this PR:

- Ran the connect tests with openshift enabled against 4.10 and 4.11 OpenShift clusters, using instructions similar to those in the description of https://github.com/hashicorp/consul-k8s/pull/2572

👀 

Checklist:
- [ ] Tests added

### DIFF
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -15,8 +15,23 @@ as well as the global.name setting.
 {{- end -}}
 {{- end -}}
 
+
 {{- define "consul.restrictedSecurityContext" -}}
 {{- if not .Values.global.enablePodSecurityPolicies -}}
+{{/*
+To be compatible with the 'restricted' Pod Security Standards profile, we
+should set this securityContext on containers whenever possible.
+
+In OpenShift < 4.11 the restricted SCC disallows setting most of these fields,
+so we do not set any for simplicity (and because that's how it was configured
+prior to adding restricted PSA support here). In OpenShift >= 4.11, the new
+restricted-v2 SCC allows setting these in the securityContext, and by setting
+them we avoid PSA warnings that are enabled by default.
+
+We use the K8s version as a proxy for the OpenShift version because there is a
+1:1 mapping of versions. OpenShift 4.11 corresponds to K8s 1.24.x.
+*/}}
+{{- if (or (not .Values.global.openshift.enabled) (and (ge .Capabilities.KubeVersion.Major "1") (ge .Capabilities.KubeVersion.Minor "24"))) -}}
 securityContext:
   allowPrivilegeEscalation: false
   capabilities:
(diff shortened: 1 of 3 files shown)


charts/consul/templates/_helpers.tpl
```diff
@@ -15,8 +15,23 @@ as well as the global.name setting.
...
+should set this securityContext on containers whenever possible.
+
+In OpenShift < 4.11 the restricted SCC disallows setting most of these fields,
+so we do not set any for simplicity (and because that's how it was configured
+prior to adding restricted PSA support here). In OpenShift >= 4.11, the new
+restricted-v2 SCC allows setting these in the securityContext, and by setting
+them we avoid PSA warnings that are enabled by default.
+
+We use the K8s version as a proxy for the OpenShift version because there is a
+1:1 mapping of versions. OpenShift 4.11 corresponds to K8s 1.24.x.
```
pglass, line 32: I'm not sure of a better way to check for OpenShift version. As far as we know and based on [their docs](https://access.redhat.com/solutions/4870701) (requires login 😞 ), there's a 1:1 mapping between OpenShift and Kube versions so I think this should be okay.


OpenShift Version | Kubernetes Version
-- | --
4.9 | 1.22
4.10 | 1.23
4.11 | 1.24
4.12 | 1.25
4.13 | 1.26

pglass: I am also unsure how to unit test this (without actually installing different kubectl / kube versions)

pglass: Ah, there is `helm template --kube-version`.

```
$ helm template -h | grep kube-version
      --kube-version string                        Kubernetes version used for Capabilities.KubeVersion
```

I'll add a unit test which uses --kube-version to validate this.

david-yu: @pglass added OpenShift versions here for future reference: https://github.com/hashicorp/consul/pull/18307

curtbushko (approved):
:chefkiss: This is awesome!

Thank you for bats testing both sides of the versioning.. that was going to be the only thing I was concerned about.

PR #2684, opened 2023-07-28
Use controller-gen 0.8.0 for CRDs

thisisnotashwin:
- Add missing license headers.


curtbushko (approved):
Thanks!

Approving assuming that the bats tests pass.

thisisnotashwin:
> Thanks!
> 
> Approving assuming that the bats tests pass.

This should not impact the BATS tests.. AFAIK those tests are 🟢  now..

PR #2702, opened 2023-07-31
Support restricted PSA enforcement part 2

pglass:

Part 2 of https://github.com/hashicorp/consul-k8s/pull/2572.

Update the following to set a "restricted" security context:

* create-federation-secret-job.yaml
* ingress-gateways-deployment.yaml
* ~mesh-gateway-deployment.yaml~
  * This is not included because adding the "restricted" security context settings to MGW makes it impossible to use hostNetwork=true
* partition-init-job.yaml
* terminating-gateways-deployment.yaml
* sync-catalog-deployment.yaml

Also, fix a WAN federation test to run on kind

**How I've tested this PR:**

Run acceptance tests on kind with restricted PSA enforcement enabled on the consul namespace (see script)

* Run the WAN fed test that is fixed in this PR, in order to validate the create federation secret job and mesh gateway deployment
* Run TestTerminatingGateway and TestIngressGateway and manually validate the gateways start and run successfully in the restricted consul namespace 
  * (note: these tests fail when test applications are deployed to a restricted namespace, because the test applications do not support running in a restricted namespace)
* Run TestPartitions_Gateway and manually validate the partitions-init job succeeds

Test script:
  * Run `make kind` and then `./test-psa-kind.sh` (no tproxy) or `./test-psa-kind.sh -tproxy` (tproxy enabled). Toggle comments at the bottom to try to run other tests.
  * Run `make kind-cni` and `./test-psa-kind.sh -cni` to test with tproxy+cni

<details>
 <summary>Test script</summary>

```shell
#!/usr/bin/env bash

set -euo pipefail

SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )

export CONSUL_LICENSE=$(cat ~/.consul-ent-license)
export CONSUL_ENT_LICENSE=$CONSUL_LICENSE

# Cleanup old namespaces
# for context in $(kubectl config get-contexts -o name | grep '^kind-') ; do
for context in kind-dc1 kind-dc2 ; do
    kubectl --context $context get ns \
		| grep ^acceptance | awk '{print $1}' \
		| xargs -n 1 -I '{}' kubectl --context $context delete ns '{}' || true
done

EXTRA_FLAGS=""

while [[ $# -gt 0 ]]; do
    case $1 in
        -tproxy)
            EXTRA_FLAGS+=" -enable-transparent-proxy"
            shift;
            ;;
        -cni)
            EXTRA_FLAGS+=" -enable-cni -enable-transparent-proxy"
            shift;
            ;;
        *)
            echo "Unrecognized argument: '$1'"
            exit 1
    esac
done

function runtest() {
    local testdir=$1
    local runtest=$2

    if [ -n "$runtest" ]; then
        runtest="-run $runtest"
    fi

    # Create consul namespaces with restricted PSA enformcement.
    set -xeuo pipefail

    local ns_base="acceptance-$1-$RANDOM"
    local contexts=""
    local namespaces=""
    #for context in $(kubectl config get-contexts -o name | grep '^kind-') ; do
    for context in kind-dc1 kind-dc2 ; do
        local consul_namespace="${ns_base}-$context"
        kubectl --context $context create ns $consul_namespace
        kubectl --context $context label --overwrite ns $consul_namespace \

        if [ -n "$contexts" ]; then
            contexts+=","
            namespaces+=","
        fi
        contexts+="$context"
        namespaces+="$consul_namespace"
    done

    # Grab the default image versions from the helm values.
    imageK8S=$(cat ../charts/consul/values.yaml | yq -r '.global.imageK8S' \
        | sed 's/-dev/-dev-ubi/')
    imageConsul=$(cat ../charts/consul/values.yaml | yq -r '.global.image' \
        | sed 's/consul:/consul-enterprise:/' | sed 's/-dev/-dev-ubi/')
    imageDataplane=$(cat ../charts/consul/values.yaml | yq -r '.global.imageConsulDataplane' \
        | sed 's/-dev/-dev-ubi/')

    cd "${SCRIPT_DIR}/tests/$testdir"
    rm -rf ./_debug
    mkdir ./_debug
    go test  -v -p 1 -timeout 15m -failfast \
        -consul-k8s-image "$imageK8S" \
        -consul-image "$imageConsul" \
        -consul-dataplane-image "$imageDataplane" \
        -debug-directory ./_debug \
        -enable-enterprise \
        -kube-contexts "$contexts" \
        -kube-namespaces "$namespaces" \
        -enable-multi-cluster -use-kind \
        -enable-restricted-psa-enforcement \
        $EXTRA_FLAGS $runtest \
        ./...
}

runtest "connect" 'TestConnectInject$'
runtest wan-federation 'TestWANFederation/secure'
#runtest terminating-gateway 'TestTerminatingGateway$/secure:_true'
#runtest terminating-gateway 'TestTerminatingGatewaySingleNamespace'
#runtest ingress-gateway 'TestIngressGatewaySingleNamespace'
#runtest ingress-gateway 'TestIngressGateway$/secure:_true'
#runtest cloud 'TestBasicCloud'
#runtest partitions 'TestPartitions_Connect/default_destination_namespace'
```

</details>

Checklist:
- [ ] Tests added


acceptance/framework/k8s/helpers.go
```diff
@@ -139,6 +139,7 @@ func CopySecret(t *testing.T, sourceContext, destContext environment.TestContext
 		secret.ResourceVersion = ""
 		require.NoError(r, err)
 	})
+	secret.Namespace = destContext.KubectlOptions(t).Namespace
```
pglass, line 142: This gave me an error like `the namespace of the provided object does not match the namespace sent on the request` (because I was using different namespaces in cluster 1 and cluster 2)

acceptance/tests/wan-federation/wan_federation_test.go
```diff
@@ -43,10 +43,6 @@ func TestWANFederation(t *testing.T) {
 			env := suite.Environment()
 			cfg := suite.Config()
 
-			if cfg.UseKind {
-				t.Skipf("skipping wan federation tests as they currently fail on Kind even though they work on other clouds.")
-			}
```
pglass, line 48: I don't think I've actually fixed anything here, so I'm curious if this is still broken in CI.

charts/consul/templates/mesh-gateway-deployment.yaml
```diff
@@ -179,6 +180,7 @@ spec:
       containers:
       - name: mesh-gateway
         image: {{ .Values.global.imageConsulDataplane | quote }}
+        {{- include "consul.restrictedSecurityContext" . | nindent 8 }}
```
pglass, line 183: I'm not 100% sure this is safe for mesh-gateway. It can run in at least the cases I've tried, but I notice there is a PodSecurityPolicy that is conditional based on hostNetwork/hostPorts: https://github.com/hashicorp/consul-k8s/blob/8b45de8b153c158d68c972b9e872acc1e4b06825/charts/consul/templates/mesh-gateway-podsecuritypolicy.yaml#L1

I imagine this would fail if I enable hostNetwork or set hostPorts, but I don't understand those settings and need to test.

pglass: I backed out the change to the mesh gateway deployment. Adding this "restricted" security context breaks mesh gateway whenever hostNetwork / hostPorts is enabled, with no workaround (i.e. cannot move it into a separate privileged namespace).

I think we'll need to create an SCC for mesh gateway when hostNetwork / hostPorts are enabled on OpenShift as well, but I don't have time to do this before the next release (PTO).

curtbushko (approved):

PR #2705, opened 2023-08-01
Backport of do not set securityContext on Openshift < 4.11 (#2678)

pglass:
Backport of https://github.com/hashicorp/consul-k8s/pull/2678 to release/1.0.x

Changes proposed in this PR:

- This is going into `release/1.1.x`, and will be backported into `release/1.0.x` because both of those Consul K8s versions support K8s < 1.23.x and therefore support OpenShift <= 4.10
- Do not set securityContext on OpenShift < 4.11 because the restricted SCC disallows setting some of those settings. I broke this in https://github.com/hashicorp/consul-k8s/pull/2572. In OpenShift >= 4.11, setting the securityContext is okay because the new restricted-v2 SCC is available to all users/accounts and allows setting the fields we want to set, and doing so avoids warnings.
- We determine the OpenShift version in Helm using the Kube version because there is a 1:1 mapping of OpenShift to Kube versions.

How I've tested this PR:

- Ran the connect tests with openshift enabled against 4.10 and 4.11 OpenShift clusters, using instructions similar to those in the description of https://github.com/hashicorp/consul-k8s/pull/2572

👀

### DIFF
--- a/.changelog/2678.txt
+++ b/.changelog/2678.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+helm: do not set container securityContexts by default on OpenShift < 4.11
+```
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -15,8 +15,23 @@ as well as the global.name setting.
 {{- end -}}
 {{- end -}}
 
+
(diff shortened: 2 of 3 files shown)


curtbushko (approved):
LGTM!

PR #2709, opened 2023-08-02
Increase timeout while waiting for vault server to be ready

curtbushko:
Changes proposed in this PR:
- Increase the timeout when waiting for the vault server to be started
- Also, the error was backwards so fix the argument order for the pod status check.

How I've tested this PR:
  
- Ran CI 
  
👀 

Checklist:
- [ ] Tests added


zalimeni (approved):
🎉

PR #2716, opened 2023-08-02
Acceptance tests: increase api-gateway retries

curtbushko:
Changes proposed in this PR:
- Increase retry timing and add a retry to getting config entries
- I still saw one flake related to xDS rate limiting but I wasn't able to reproduce it

How I've tested this PR:

- ran it through CI several times

:eyes:

Checklist:
- [ ] Tests added

### DIFF
--- a/acceptance/tests/api-gateway/api_gateway_test.go
+++ b/acceptance/tests/api-gateway/api_gateway_test.go
@@ -101,11 +101,26 @@ func TestAPIGateway_Basic(t *testing.T) {
 			logger.Log(t, "creating target http server")
 			k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.NoCleanup, cfg.DebugDirectory, "../fixtures/cases/static-server-inject")
 
-			logger.Log(t, "patching route to target http server")
-			k8s.RunKubectl(t, ctx.KubectlOptions(t), "patch", "httproute", "http-route", "-p", `{"spec":{"rules":[{"backendRefs":[{"name":"static-server","port":80}]}]}}`, "--type=merge")
+			// We use the static-client pod so that we can make calls to the api gateway
+			// via kubectl exec without needing a route into the cluster from the test machine.
+			logger.Log(t, "creating static-client pod")
+			k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.NoCleanup, cfg.DebugDirectory, "../fixtures/bases/static-client")
 
 			logger.Log(t, "creating target tcp server")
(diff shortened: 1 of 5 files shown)


acceptance/tests/api-gateway/api_gateway_test.go
```diff
@@ -115,11 +120,6 @@ func TestAPIGateway_Basic(t *testing.T) {
 				k8s.RunKubectlAndGetOutputE(t, ctx.KubectlOptions(t), "delete", "-f", "../fixtures/cases/api-gateways/tcproute/route.yaml")
 			})
 
-			// We use the static-client pod so that we can make calls to the api gateway
-			// via kubectl exec without needing a route into the cluster from the test machine.
-			logger.Log(t, "creating static-client pod")
-			k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.NoCleanup, cfg.DebugDirectory, "../fixtures/bases/static-client")
-
```
zalimeni, line 122: ~ Why the relocation above for this block?

acceptance/tests/api-gateway/api_gateway_test.go
```diff
@@ -209,17 +209,25 @@ func TestAPIGateway_Basic(t *testing.T) {
 			checkStatusCondition(t, tcpRoute.Status.Parents[0].Conditions, trueCondition("ConsulAccepted", "Accepted"))
 
 			// check that the Consul entries were created
-			entry, _, err := consulClient.ConfigEntries().Get(api.APIGateway, "gateway", nil)
-			require.NoError(t, err)
-			gateway := entry.(*api.APIGatewayConfigEntry)
+			// On startup, the controller can take upwards of 1m to perform
+			// leader election so we may need to wait a long time for
+			// the reconcile loop to run (hence the 1m timeout here).
```
zalimeni, line 214: It looks like we have a retry above w/ the same comment - do we expect this second block to fail independently after the first passes for the same leader election, or am I missing some nuance?

zalimeni (approved):
Few non-blocking questions, LMK if any of them don't make sense. Otherwise LGTM.

It seems like at worst we'll pick up the difference for a single retry check here in terms of failing more slowly when things flake, so if it buys us fewer timeout-related flakes, that's a win.

acceptance/tests/api-gateway/api_gateway_test.go
curtbushko: I was shuffling things around to add retries & checks to the deployments but it didn't work out.

acceptance/tests/api-gateway/api_gateway_test.go
curtbushko: Removed the comment.

acceptance/tests/api-gateway/api_gateway_test.go
zalimeni: Oh, like we don't need to move this after all?

curtbushko: Not really. It'll give it a few ms to deploy and get ready but it shouldn't make a difference.

PR #2723, opened 2023-08-03
Configure Gateway Deployment Resources

t-eckert:
Adds the ability for users to customize the resource requests and limits for the Pods deployed as part of the Gateway deployment.

Changes proposed in this PR:

- Move `resources` into `managedGatewayClass` in Helm config
- Add resource configuration to `GatewayClassConfig`
- Regenerate CRDs
- Pass resource configuration into the `gateway-resources-job` with a new `gateway-resources-configmap`
- Pull in `resources` from `GatewayClassConfig` in `Gatekeeper`
- Clean up some comments in existing code

How I've tested this PR:

I deployed Consul with the resources for the `managedGatewayClass` configured and saw that the resources were properly configured for the Pods when set, when omitted, and when the stanza was completely deleted.

Try the same, with a deployment, create a Gateway that points to the `consul` GatewayClass and ensure that it has the resources you configured. You can also create your own GatewayClassConfig with custom resource settings.

Checklist:
- [x] Tests added

Closes https://hashicorp.atlassian.net/browse/NET-5155


charts/consul/values.yaml
```diff
@@ -2195,6 +2195,17 @@ connectInject:
         # @type: string
         service: null
 
+      # The resource settings for Pods handling traffic for Gateway API.
```
t-eckert, line 2198: Jeff and I agreed that this made more sense to be set on the `managedGatewayClass`.

charts/consul/values.yaml
```diff
@@ -2226,17 +2237,6 @@ connectInject:
       # @type: string
       annotations: null
 
-    # The resource settings for Pods handling traffic for Gateway API.
```
t-eckert, line 2229: This was unfortunately never wired up.

missylbytes (approved):

control-plane/api-gateway/common/helm_config.go
```diff
@@ -14,13 +14,15 @@ const componentAuthMethod = "k8s-component-auth-method"
 // This is a combination of the apiGateway stanza and other settings that impact api-gateways.
 type HelmConfig struct {
 	// ImageDataplane is the Consul Dataplane image to use in gateway deployments.
-	ImageDataplane             string
+	ImageDataplane string
+	// ImageConsulK8S is the Consul Kubernetes image to use in gateway deployments.
```
nathancoleman, line 18: This might be a little more clear since the docker image is `hashicorp/consul-k8s-control-plane`
```suggestion
	// ImageConsulK8S is the Consul Kubernetes Control Plane image to use in gateway deployments.
```

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -54,6 +54,9 @@ type GatewayClassConfigSpec struct {
 	// Deployment defines the deployment configuration for the gateway.
 	DeploymentSpec DeploymentSpec `json:"deployment,omitempty"`
 
+	// Resources defines the resource requirements for the gateway.
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
nathancoleman, line 58: It seems to me like it might make sense to nest in the `DeploymentSpec`. Thoughts?

control-plane/subcommand/gateway-resources/command.go
```diff
@@ -288,6 +301,35 @@ func (c *Command) validateFlags() error {
...
+		c.resources = defaultResourceRequirements()
+		return nil
+	}
+
+	resources, err := io.ReadAll(file)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Unable to read resources.json, using defaults: %s", err))
+		c.resources = defaultResourceRequirements()
+		return err
+	}
```
nathancoleman, line 321: Looks like you designed this in a way that customers using a newer version of consul-k8s-control plane with an older version of the Helm chart will just get the default requirements ❤️

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.connectInject.enabled }}
+# Configuration of Gateway Resources Job which creates managed Gateway configuration.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "consul.fullname" . }}-gateway-resources-config
```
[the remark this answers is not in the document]
t-eckert: `gateway-resources` is actually the name of the Job that creates the resources the Gateway uses. It's confusing. But this is not isolating it to just the `resources` config in Kubernetes -- it's a reference to the Job name.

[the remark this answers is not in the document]
nathancoleman: ah that is indeed confusing but is how it is. Thanks for clarifying!

control-plane/api/v1alpha1/api_gateway_types.go
t-eckert: I like that! I'll go do that.

control-plane/subcommand/gateway-resources/command.go
t-eckert: Exactly! It will fall back nicely.

charts/consul/values.yaml
```diff
@@ -2195,6 +2195,17 @@ connectInject:
...
+      # The resource settings for Pods handling traffic for Gateway API.
+      # @recurse: false
+      # @type: map
+      resources:
+        requests:
+          memory: "100Mi"
+          cpu: "100m"
+        limits:
+          memory: "100Mi"
+          cpu: "100m"
```
nathancoleman, line 2207: I see these were just moved over from before, but are these the default resource values we want to go out the door with? Probably worth running past @Jeff-Apple 

Edit: If we were to change it here, then we probably also want to change it in `defaultResourceRequirements` below

nathancoleman (commented):
Only the one question about default resource values that I'd like to cement before approving

charts/consul/values.yaml
t-eckert: That's a great call-out. I'm not sure what the best limit would be for these.

charts/consul/templates/crd-gatewayclassconfigs.yaml
```diff
@@ -78,14 +80,72 @@ spec:
                     maximum: 8
                     minimum: 1
                     type: integer
+                  resources:
+                    description: Resources defines the resource requirements for the
+                      gateway.
+                    properties:
+                      claims:
```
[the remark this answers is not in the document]
t-eckert: Yeah, I agree it is strange. Maybe we need to update our corev1 dependency?

charts/consul/values.yaml
t-eckert: These values are the same as the defaults for Ingress Gateway -- so I think that's a good sign that they are solid defaults here.

t-eckert: I checked this out with @lkysow and he gave the thumbs up on these values.

nathancoleman: Thanks @t-eckert !

nathancoleman (approved):

PR #2734, opened 2023-08-04
NET-4978: New CRDs for GW JWT Auth

jm96441n:
Changes proposed in this PR:
- Added CRD for `GatewayPolicy`
- Added CRD for `HTTPRouteAuthFilter`

How I've tested this PR:
Installed CRDs into a kind cluster

read code

Checklist:
- [ ] Tests added


charts/consul/templates/crd-http-route-auth-filter.yaml
```diff
@@ -0,0 +1,135 @@
...
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: crd
+  name: httproute-auth-filters.consul.hashicorp.com
+spec:
+  group: consul.hashicorp.com
+  names:
+    categories:
+      - gateway-api
+    kind: HTTPRouteAuthFilter
```
andrewstucki, line 22: Is this name redundant/does it stutter a bit? Should we just simply rename `AuthFilter` at the CRD level?

charts/consul/templates/crd-gatewaypolicy.yaml
```diff
@@ -0,0 +1,210 @@
+{{- if .Values.connectInject.enabled }}
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
andrewstucki, line 6: did the go structs already make it in or did you hand roll these?

andrewstucki: just wondering b/c normally these are generated from go structs

andrewstucki: Take a look at @sarahalsmiller's PR for an example: https://github.com/hashicorp/consul-k8s/pull/2735/files#diff-ecdac93bf4d514eb648c162d8b54671660e679f0b730829a66ad944e70ca7540

jm96441n: I hand rolled them, talked to @sarahalsmiller on slack and realized that I should be generating them from structs, going to close this and re-open it with the correct setup

charts/consul/templates/crd-http-route-auth-filter.yaml
jm96441n: we could name it `AuthFilter` or maybe even `RouteAuthFilter` if it will also be applied to GRPC routes at some point in the future, I do like the `Route` prefix as it makes it a bit clearer what it applies to

nathancoleman: +1 to the `Route` prefix as it helps make things self-documenting and follows existing conventions w/ `Gateway`, `GatewayClass`, `GatewayClassConfig`

charts/consul/templates/crd-gatewaypolicies.yaml
```diff
@@ -0,0 +1,208 @@
...
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: GatewayPolicySpec defines the desired state of GatewayPolicy
+            properties:
+              default:
+                properties:
+                  jwt:
+                    description: APIGatewayJWTRequirement holds the list of JWT providers
```
missylbytes, line 61: nit, some description fields end in a `.` and others don't. May want to unify.

control-plane/config/rbac/gatewaypolicy_editor_role.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) HashiCorp, Inc.
```
andrewstucki, line 1: Hey John, just for cleanness sake, can you drop these files? Not sure what part of the generation step is adding them, but we generally don't commit any of these to our repos since they're just superfluous.

control-plane/config/crd/kustomization.yaml
```diff
@@ -9,17 +9,23 @@ kind: Kustomization
 # It should be run by config/default
 resources:
 - bases/consul.hashicorp.com_controlplanerequestlimits.yaml
+- bases/consul.hashicorp.com_httprouteauthfilters.yaml
+- bases/consul.hashicorp.com_gatewaypolicies.yaml
```
andrewstucki, line 13: can you also revert this file?

.changelog/2734.txt
```diff
@@ -0,0 +1,3 @@
+```releast-note:feature
+api-gateway: Add CRD for GatewayPolicy and HTTPRouteAuthFilter
```
andrewstucki, line 2: It seems like you're going to separate this into two commits back to `main`? Or were you planning on squashing this along with the actual implementation?

andrewstucki (dismissed):
Some quick thoughts, will review the structs, but, in the meantime, think you can drop the stuff under `control-plane/config/rbac` and revert `control-plane/config/crd/kustomization.yaml`? We generally haven't included those in the repo, and they're unused files.

.changelog/2734.txt
jm96441n: I'm not sure I follow, this includes the implementation for the CRDs, but not necessarily the underlying logic for them to actually do anything, would it make sense to have this changelog on a later commit? (we would still need to handle the translation work and all that to make them usable)

andrewstucki: Yeah, so that was my question, are you going to separate the two:

1. CRDs
2. the plumbing to use them

I'm just wondering if it makes sense to just work off this branch and then do a monolithic feature merge -- just because I envision that the actual controller plumbing here is going to be fairly light (probably roughly the same amount as Sarah's previous [PR](https://github.com/hashicorp/consul-k8s/pull/2735) or a bit larger due to the use of both policies and filters). If you do that, I'd probably just do a single changelog entry, as this commit is pretty much unusable from a user-perspective without the actual implementation paired with it, so not sure if a changelog actually makes sense here.

charts/consul/templates/crd-httprouteauthfilters.yaml
```diff
@@ -0,0 +1,140 @@
...
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: crd
+spec:
+  group: consul.hashicorp.com
+  names:
+    kind: HTTPRouteAuthFilter
```
nathancoleman, line 19: It seems like this should be `RouteAuthFilter` to stay consistent with some shifts in Sarah's PR for timeout

charts/consul/templates/crd-gatewaypolicies.yaml
```diff
@@ -0,0 +1,208 @@
...
+                              type: array
+                          required:
+                          - name
+                          type: object
+                        type: array
+                    required:
+                    - providers
+                    type: object
+                required:
+                - jwt
```
nathancoleman, line 109: Is the `omitEmpty` json tag what adds/removes things from the `required` list in the resulting CRD?

andrewstucki: I believe `kubebuilder` defaults everything to `required`, but you can explicitly mark fields with a `+optional` comment if you want them to be optional in the JSON spec. I believe if the generator sees a `json` tag with `omitempty` it also interprets the field as having a `+optional` comment added to it.

charts/consul/templates/crd-httprouteauthfilters.yaml
jm96441n: so in the spec GrpcRoute filters just reuse the `HTTP` prefixed filters: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1alpha2.GRPCRouteFilter
should we follow suit there or be internally consistent?

nathancoleman: Being internally consistent is the more important thing IMO since these are CRDs.
If y'all want to go back and change `RouteRetryFilter` => `HTTPRouteRetryFilter`, then I'm fine with that. I believe @andrewstucki had stronger feelings about that than I do.

At the end of the day, I care most about internal consistency. Whether that's `Route...` or `HTTPRoute...` is fine with me.

.changelog/2734.txt
jm96441n: so I created a new branch that we're going to work off of that this branch is now based on

charts/consul/templates/crd-gatewaypolicies.yaml
jm96441n: this is now optional

charts/consul/templates/crd-httprouteauthfilters.yaml
jm96441n: either works for me, less typing for our end users is probably better

nathancoleman: Discussed with the team and landed on `RouteAuthFilter`

nathancoleman (approved):

control-plane/api/common/common.go
```diff
@@ -17,6 +17,8 @@ const (
 	SamenessGroup            string = "samenessgroup"
 	JWTProvider              string = "jwtprovider"
 	ControlPlaneRequestLimit string = "controlplanerequestlimit"
+	HTTPRouteAuthFilter      string = "httprouteauthfilter"
```
andrewstucki, line 20: This looks like it needs to be updated still with the new name

control-plane/PROJECT
```diff
@@ -126,4 +126,20 @@ resources:
   kind: RouteTimeoutFilter
   path: github.com/hashicorp/consul-k8s/control-plane/api/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1beta1
+    namespaced: true
+  domain: hashicorp.com
+  group: consul
+  kind: HTTPRouteAuthFilter
```
andrewstucki, line 134: this too even though I'm fairly certain this file is almost entirely unused

control-plane/api/v1alpha1/gatewaypolicy_types.go
```diff
@@ -0,0 +1,117 @@
...
+	SectionName *gwv1beta1.SectionName `json:"sectionName,omitempty"`
+}
+
+type GatewayPolicyConfig struct {
+	//+kubebuilder:validation:Optional
+	JWT *APIGatewayJWTRequirement `json:"jwt,omitempty"`
+}
+
+// APIGatewayJWTRequirement holds the list of JWT providers to be verified against.
+type APIGatewayJWTRequirement struct {
```
andrewstucki, line 90: For consistency are you ok with dropping the `API` part of the naming on these structs? We're "Gateway" spec implementors and happen to call the underlying type of gateway in Consul an "API gateway" - but from the k8s perspective everything is just "Gateway" -- main reason is because these docstrings having "API" in them are part of the CRDs, so I think they'll wind up being somewhat user-facing.

control-plane/PROJECT
jm96441n: I _think_ this file is autogenerated, let me make sure I didn't miss any commands to generate this

control-plane/api/v1alpha1/gatewaypolicy_types.go
jm96441n: I'm okay with that, should we also drop the `API` portion of it on the consul side of things?

control-plane/PROJECT
jm96441n: updated this

control-plane/api/v1alpha1/gatewaypolicy_types.go
andrewstucki: RE Consul side of things, maybe, I think the only place where that'd seem a bit wonky is in the API subpackage assuming someone is creating these directly through the go client.

PR #2735, opened 2023-08-04
Net 4889 implement retry feature on the api gateway

sarahalsmiller:
Changes proposed in this PR:
- Added HTTPRouteRetryFilter logic and HTTPRouteTimeoutFilter retry boilerplate

How I've tested this PR:

Checklist:
- [ X ] Tests added
- [ X ] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)


control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -13,6 +14,7 @@ import (
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
```
jm96441n, line 17: did you mean to add this?

sarahalsmiller: Yes, I was using the errors.Wrap() function in that package. But that'll probably go away after I figure out how to bubble up the errors like I want

jm96441n: gotcha, would the error wrapping in the std lib solve that?

charts/consul/templates/crd-httprouteretryfilters.yaml
```diff
@@ -0,0 +1,122 @@
...
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: HTTPRouteRetryFilterSpec defines the desired state of HTTPRouteRetryFilter
+            properties:
+              numRetries:
```
nathancoleman, line 59: I think it makes sense to require `min: 0` on this, maybe even `min: 1` but I'm less certain about that based on product requirements

charts/consul/templates/crd-httprouteretryfilters.yaml
```diff
@@ -0,0 +1,122 @@
...
+                format: int32
+                type: integer
+              retryOn:
+                items:
+                  type: string
+                type: array
+              retryOnConnectFailure:
+                type: boolean
+              retryOnStatusCodes:
+                items:
```
nathancoleman, line 69: I think it makes sense to add some min-max validation on this field as well

charts/consul/templates/crd-httprouteretryfilters.yaml
```diff
@@ -0,0 +1,122 @@
...
+                  format: int32
+                  type: integer
+                type: array
+            required:
+            - numRetries
+            - retryOn
+            - retryOnConnectFailure
+            - retryOnStatusCodes
```
nathancoleman, line 77: I don't think it makes sense to require all of these be set since you could reasonably set up connect failure retry without setting up status code retry

charts/consul/templates/crd-httproutetimeoutfilters.yaml
```diff
@@ -0,0 +1,118 @@
...
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: HTTPRouteTimeoutFilterSpec defines the desired state of HTTPRouteTimeoutFilter
+            properties:
+              idleTimeout:
```
nathancoleman, line 59: Curious where the description for these guys came from. I think it makes sense to require a `min: 0` here as well. Probs worth checking with Product to see if we should implement an upper bound

nathancoleman (commented):
Some early feedback here - I'll take a deeper look tomorrow

charts/consul/templates/crd-httproutetimeoutfilters.yaml
sarahalsmiller: These files are all autogenerated, but I believe you can set these things with kubebuilder tags. I'll dig into that before I mark this PR as ready to review.

nathancoleman: You're right, today I learned about kubebuilder's [CRD validation](https://book.kubebuilder.io/reference/markers/crd-validation.html) tags 💯 
```go
// +kubebuilder:validation:Minimum
```

control-plane/api-gateway/controllers/gateway_controller.go
sarahalsmiller: Probably, but no longer using the wrap function.

charts/consul/templates/crd-httproutetimeoutfilters.yaml
sarahalsmiller: It turns out you can't use the Minimum tag on time.Duration types.

control-plane/api-gateway/common/filtererrors.go
```diff
@@ -0,0 +1,18 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var ErrInvalidExternalRefType = errors.New("invalid kind")
+var ErrExternalRefNotFound = errors.New("ref not found")
+
+type HTTPRouteFilterError struct {
+	Kind    error
+	Message string
+	client.Object
```
sarahalsmiller, line 17: This makes it so that the httproutefilter error can be passed around our other functions along with the filters themselves, so they can be sent back to the user during bindRoute.

control-plane/api-gateway/common/translation.go
```diff
@@ -4,6 +4,7 @@
 package common
 
 import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
nathancoleman, line 7: #nit import blocking

control-plane/api-gateway/common/resources.go
```diff
@@ -364,6 +365,25 @@ func (s *ResourceMap) ReferenceCountHTTPRoute(route gwv1beta1.HTTPRoute) {
 	s.httpRouteGateways[consulKey] = set
 }
 
+func (s *ResourceMap) AddHTTPRouteExternalFilter(route gwv1beta1.HTTPRoute, filter client.Object) {
+	key := client.ObjectKeyFromObject(&route)
+	consulKey := NormalizeMeta(s.toConsulReference(api.HTTPRoute, key))
+
+	set, ok := s.httpRouteGateways[consulKey]
+	if !ok {
+		s.ReferenceCountHTTPRoute(route)
```
sarahalsmiller, line 374: route map needs to be initialized. This shouldn't happen in normal operations, but I did get a panic without this line during testing

control-plane/api-gateway/common/translation.go
```diff
@@ -163,6 +164,30 @@ func (t ResourceTranslator) ToHTTPRoute(route gwv1beta1.HTTPRoute, resources *Re
...
+				RetryOnStatusCodes:    crdFilter.Spec.RetryOnStatusCodes,
+				RetryOnConnectFailure: crdFilter.Spec.RetryOnConnectFailure,
+			}
+
+			configEntry.Rules = append(configEntry.Rules, api.HTTPRouteRule{Filters: api.HTTPFilters{RetryFilter: &consulFilter}})
+		case *v1alpha1.HTTPRouteTimeoutFilter:
+			//TODO @sarahalsmiller, follow up PR
+
+		default:
+			//TODO @sarahalsmiller throw error
```
nathancoleman, line 186: :eyes:

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -782,6 +795,52 @@ func (c *GatewayController) getRelatedHTTPRoutes(ctx context.Context, gateway ty
...
+				Name:      string(filter.ExtensionRef.Name),
+				Namespace: route.Namespace,
+			}, externalFilter); err != nil {
+				c.Log.Info("\n\n\n error")
+				c.Log.Error(err, "unable to get filter")
+				externalFilter = &common.HTTPRouteFilterError{
+					Kind:    common.ErrExternalRefNotFound,
+					Message: err.Error(),
+				}
+			}
```
nathancoleman: #nit this is the kind of place where I avoid the compound `if` because readability suffers

```suggestion
			err := c.Client.Get(ctx, client.ObjectKey{
				Name:      string(filter.ExtensionRef.Name),
				Namespace: route.Namespace,
			}, externalFilter)
			if err != nil {
				c.Log.Info("\n\n\n error")
				c.Log.Error(err, "unable to get filter")
				externalFilter = &common.HTTPRouteFilterError{
					Kind:    common.ErrExternalRefNotFound,
					Message: err.Error(),
				}
			}
```

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -782,6 +795,52 @@ func (c *GatewayController) getRelatedHTTPRoutes(ctx context.Context, gateway ty
 	return list.Items, nil
 }
 
+func (c *GatewayController) getExternalFiltersForHTTPRoute(ctx context.Context, route gwv1beta1.HTTPRoute, resources *common.ResourceMap) ([]interface{}, error) {
+	var externalFilters []interface{}
+	for _, rule := range route.Spec.Rules {
+		for _, filter := range rule.Filters {
+			c.Log.Info("\n\n filter iterating")
```
nathancoleman: Is the extra logging in this function intended to stick around?

nathancoleman (commented):
Nice work! Putting some initial thoughts/questions here. Excited to use it!

control-plane/api-gateway/binding/result.go
```diff
@@ -183,9 +185,12 @@ func (b bindResults) Condition() metav1.Condition {
...
 				reason = "RefNotPermitted"
-			case errRouteNoMatchingParent:
-				// or if the route declares a parent that we can't find
-				reason = "NoMatchingParent"
+			case common.ErrExternalRefNotFound:
+				reason = "FilterNotFound"
+				message = result.message
+			case common.ErrInvalidExternalRefType:
+				reason = "UnsupportedValue"
+				message = result.message
```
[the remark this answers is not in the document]
sarahalsmiller: Mostly because common needs to throw the error and the binding package calls common and can't import it. The other option would be to pull these error types into their own package but that feels excessive.

control-plane/api-gateway/binding/result.go
```diff
@@ -122,6 +122,7 @@ func (e routeValidationResults) Condition() metav1.Condition {
 type bindResult struct {
 	section gwv1beta1.SectionName
 	err     error
+	message string
```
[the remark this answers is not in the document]
andrewstucki: 👍 the error is meant to be used in the case of "I had a binding error due to something" -- it's not an error meant to bubble up through the controller, but rather used in generating the message set as part of an "error" status on the route. So, ideally we'd keep that same pattern, just return some sane in `err.Error()` and optionally do an error constant/interface if we want to change the status metadata (i.e. type, reason, etc.)

control-plane/api-gateway/controllers/gateway_controller.go
sarahalsmiller: Nope, cleaned it up

control-plane/api-gateway/common/filtererrors.go
andrewstucki: I'm not entirely sure I'm following here? Why not just follow the pattern here?

https://github.com/hashicorp/consul-k8s/blob/0100fa4f066a184de8ae11976ef70b5afa85fe39/control-plane/api-gateway/binding/route_binding.go#L81-L89

Why the need to pass errors around outside of the validation result?

control-plane/api-gateway/common/resources.go
```diff
@@ -364,6 +365,25 @@ func (s *ResourceMap) ReferenceCountHTTPRoute(route gwv1beta1.HTTPRoute) {
...
+		set = s.httpRouteGateways[consulKey]
+	}
+	if set.externalFilters == nil {
+		//initialize
+		set.externalFilters = mapset.NewSet()
+	}
+
+	set.externalFilters.Add(filter)
```
andrewstucki, line 382: Would it make sense to just store this as a map of keys --> kubernetes objects? Also why make this a `client.Object` rather than just passing the CRD type directly? Ideally if we have a bad filter, the whole route is considered invalid and not bound, so no need to store anything overloaded here other than the CRD objects themselves (I believe that mapset also requires object hashing which I'm not sure that `client.Object` will actually implement, so this code might panic).

control-plane/api/v1alpha1/httprouteretryfilter_types.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// HTTPRouteRetryFilter is the Schema for the httprouteretryfilters API
+// +kubebuilder:printcolumn:name="Synced",type="string",JSONPath=".status.conditions[?(@.type==\"Synced\")].status",description="The sync status of the resource with Consul"
+// +kubebuilder:printcolumn:name="Last Synced",type="date",JSONPath=".status.lastSyncedTime",description="The last successful synced time of the resource with Consul"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="The age of the resource"
+type HTTPRouteRetryFilter struct {
```
andrewstucki, line 21: Similar to the conversation around naming with @jm96441n's PR, should this be something more generic like `RouteRetyFilter` assuming we'd use it on GRPC-based routes (if that's a thing in envoy) eventually?

andrewstucki (changes requested):
I think the error/validation handling around this should probably be simplified, let me know if you want to pair. Left comments about ideally what pattern this should follow so that we don't have two ways of doing the same thing.

control-plane/api-gateway/binding/result.go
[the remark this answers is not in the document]
nathancoleman: I believe you can accomplish what you're wanting using `errors.Is` ([docs](https://pkg.go.dev/errors#Is))

```go
switch {
case errors.Is(err, errLocal):
  // TODO
case errors.Is(err, common.ErrInvalidExternalRefType):
  // TODO
...
}

File: control-plane/api/v1alpha1/httprouteretryfilter_types.go (hunk 15 above)
Comment (sarahalsmiller) [reply control-plane/api/v1alpha1/httprouteretryfilter_types.go:21]: Ok, I'll update it and regen

File: control-plane/api-gateway/binding/result.go (hunk 13 above)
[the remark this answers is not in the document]
Comment (sarahalsmiller) [reply control-plane/api-gateway/binding/result.go:124]: Ok I'll give that a go

File: control-plane/api-gateway/common/filtererrors.go (hunk 6 above)
Comment (sarahalsmiller) [reply control-plane/api-gateway/common/filtererrors.go:17]: The issue I was trying to solve for was that the error was occurring when the route was being pulled and I wasn't sure how to account for that, but I think I actually see an easier way to deal with that.

File: control-plane/api-gateway/binding/validation.go [hunk 16]
```diff
@@ -431,6 +431,53 @@ func routeAllowedForListenerHostname(hostname *gwv1beta1.Hostname, hostnames []g
...
+	}
+
+	for _, rule := range route.Spec.Rules {
+		for _, filter := range rule.Filters {
+			if filter.ExtensionRef == nil {
+				continue
+			}
+			filterKey := keyFromLocalObjectReference(filter.ExtensionRef)
+
+			_, ok := extensionRefs[filterKey]
```
andrewstucki, line 452: Rather than doing things like this, I think you probably just want something like a `resources.RouteExtensionRefExists(filter.ExtensionRef)` method.

As it is, this abstraction leaks, where you're creating a map in one context, assuming it's structured a particular way and then recreating the key here to access the elements of the map. Instead, it should be quicker (kind of trivial, but it'll save you at least one loop over the internal mapset) and less leaky to just leverage the internal resourcemap mapset semantics to check for the existence of the key within the context of the `resources.RouteExtensionRefExists` call.

Additionally, I think you should likely be checking the filter type first, and then checking for a corresponding `ExtensionRef`. If the filter type is set to use an extension, but then has no `ExtensionRef` field set, or an invalid `ExtensionRef.Kind` that we don't know how to handle, we should say, "nope, we're invalid".

control-plane/api-gateway/binding/validation.go
```diff
@@ -431,6 +431,53 @@ func routeAllowedForListenerHostname(hostname *gwv1beta1.Hostname, hostnames []g
...
+
+// externalRefsKindAllowedOnRoute
+func externalRefsKindAllowedOnRoute(route *gwv1beta1.HTTPRoute) bool {
+	for _, rule := range route.Spec.Rules {
+		for _, filter := range rule.Filters {
+			if filter.ExtensionRef == nil {
+				continue
+			}
+
+			if filter.ExtensionRef.Kind != "HTTPRouteRetryFilter" {
```
andrewstucki, line 472: Can we use a constant here? Also wondering if we can collapse this with the above method into something like a `validateFilterExtensions` call? That way we just iterate over the rules/filters once and then return an error value corresponding to either an invalid filter type of the filter not existing?

control-plane/api-gateway/common/resources.go
andrewstucki: Since there is an expected order in how you build out these resource dependency maps (i.e. you pull the routes and ref count them first), I'd just see if you can fix up the test instead of doing this here -- I think you'll just need to call the `ReferenceCountHTTPRoute` function prior to calling this function in your test.

control-plane/api-gateway/common/resources.go
```diff
@@ -605,6 +633,34 @@ func (s *ResourceMap) HTTPRouteCanReferenceBackend(route gwv1beta1.HTTPRoute, re
 	return s.referenceValidator.HTTPRouteCanReferenceBackend(route, ref)
 }
 
+func (s *ResourceMap) getConsulKey(route *gwv1beta1.HTTPRoute) api.ResourceReference {
```
andrewstucki, line 636: Be careful with this method, I intentionally kept some of the normalization code a bit heavy in this implementation due to the care that needs to be taken around when we do/don't normalize our lookups based on how non-enterprise Consul freaks out if you pass any sort of Namespace/Partition data back. I think it should be fine in your usage below since you're not trying to sync anything back to Consul and inlining the filters to the route resource, but the easier we make some of this normalized key construction, the easier it is to accidentally break non-enterprise builds.

control-plane/api-gateway/common/resources.go
```diff
@@ -605,6 +633,34 @@ func (s *ResourceMap) HTTPRouteCanReferenceBackend(route gwv1beta1.HTTPRoute, re
...
+	key := client.ObjectKeyFromObject(route)
+	return NormalizeMeta(s.toConsulReference(api.HTTPRoute, key))
+}
+
+func (s *ResourceMap) GetHTTPRouteExtensionRefs(route *gwv1beta1.HTTPRoute) map[string]v1alpha1.ExternalRouteFilter {
```
andrewstucki, line 641: So this is the method I mentioned I'd change above. Instead of returning a map, you can just do the lookup right here, that's how the secret refs code currently works for gateways ensuring they can access secrets IIRC.

That also reminds me... I'm thinking we might need to make all of these resources `ResourceGrant` aware like secret refs are?

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -154,6 +154,15 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
+	//get all custom httprouteFilters these routes reference
+	for _, h := range httpRoutes {
+		_, err := r.getExternalFiltersForHTTPRoute(ctx, h, resources)
```
andrewstucki, line 159: Does it make sense to drop this into the `getRelatedHTTPRoutes` function above? Don't really care either way, but just wondering if it makes more sense for organizational grouping sake.

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -782,6 +791,43 @@ func (c *GatewayController) getRelatedHTTPRoutes(ctx context.Context, gateway ty
...
+	var externalFilters []interface{}
+	for _, rule := range route.Spec.Rules {
+		for _, filter := range rule.Filters {
+			var externalFilter client.Object
+			//check to see if we need to grab this filter
+			if filter.ExtensionRef == nil {
+				continue
+			}
+			switch kind := filter.ExtensionRef.Kind; kind {
+			case "HTTPRouteRetryFilter":
```
andrewstucki, line 804: can we move these to constants?

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -782,6 +791,43 @@ func (c *GatewayController) getRelatedHTTPRoutes(ctx context.Context, gateway ty
 	return list.Items, nil
 }
 
+func (c *GatewayController) getExternalFiltersForHTTPRoute(ctx context.Context, route gwv1beta1.HTTPRoute, resources *common.ResourceMap) ([]interface{}, error) {
+	var externalFilters []interface{}
+	for _, rule := range route.Spec.Rules {
+		for _, filter := range rule.Filters {
+			var externalFilter client.Object
+			//check to see if we need to grab this filter
+			if filter.ExtensionRef == nil {
```
andrewstucki, line 800: think we'll want to check the filter type before this?

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -782,6 +791,43 @@ func (c *GatewayController) getRelatedHTTPRoutes(ctx context.Context, gateway ty
...
+			}
+
+			//get object from API
+			err := c.Client.Get(ctx, client.ObjectKey{
+				Name:      string(filter.ExtensionRef.Name),
+				Namespace: route.Namespace,
+			}, externalFilter)
+
+			if err != nil {
+				c.Log.Error(err, "externalref not found")
```
andrewstucki, line 820: This should probably be something lower than `Error`, just since it'll alarm users when the reality is that they just are missing a configuration object that they should be notified about on the route status.

control-plane/api-gateway/binding/result.go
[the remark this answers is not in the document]
sarahalsmiller: No longer handeing errors this way

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -782,6 +791,43 @@ func (c *GatewayController) getRelatedHTTPRoutes(ctx context.Context, gateway ty
 	return list.Items, nil
 }
 
+func (c *GatewayController) getExternalFiltersForHTTPRoute(ctx context.Context, route gwv1beta1.HTTPRoute, resources *common.ResourceMap) ([]interface{}, error) {
```
andrewstucki, line 794: Aside from this, we'll also likely want to add:

1. A custom cache indexing function to do all of the routes referencing a particular filter
2. A kube resource `Watch` call on these CRDs that triggers the gateway reconciliation through their referencing routes (leveraging the index above), you can probably take some inspiration from the service Endpoint watching code for an idea as to what you can do

andrewstucki (commented):
More comments, I'll take a bit more thorough look next week, but here are some things that stood out. Making progress! 👍

control-plane/api-gateway/common/translation.go
```diff
@@ -163,6 +163,28 @@ func (t ResourceTranslator) ToHTTPRoute(route gwv1beta1.HTTPRoute, resources *Re
...
+			//new filter that needs to be appended
+
+			consulFilter := api.RetryFilter{
+				NumRetries:            crdFilter.Spec.NumRetries,
+				RetryOn:               crdFilter.Spec.RetryOn,
+				RetryOnStatusCodes:    crdFilter.Spec.RetryOnStatusCodes,
+				RetryOnConnectFailure: crdFilter.Spec.RetryOnConnectFailure,
+			}
+
+			configEntry.Rules = append(configEntry.Rules, api.HTTPRouteRule{Filters: api.HTTPFilters{RetryFilter: &consulFilter}})
```
andrewstucki, line 180: Sorry, wound up taking one more look at the implementation and realized a fairly big issue here. So, filters can appear two places in the spec:

1. On an [HTTPBackendRef](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPBackendRef)
2. On an [HTTPRouteRule](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteRule)

A couple things here:

- In general, it looks like this most of what's handled in this PR is geared around the `HTTPRouteRule` filters, we still need to handle the `HTTPBackendRef` filters -- so we'll need to update the validation and fetching logic accordingly.
- We shouldn't actually be appending a **_new_** rule to the ConfigEntry as here, we should be setting the filter on the rule that it's associated with (i.e. we'll want this to get set somewhere in`translateHTTPRouteRule`). The place that looks right to me is specifically under `translateHTTPFilters`. Adding this code there should also work for the translation logic for the backend targeting since `translateHTTPBackendRef` also invokes `translateHTTPFilters`

control-plane/api-gateway/common/filtererrors.go
sarahalsmiller: No longer using this pattern

control-plane/api-gateway/common/translation.go
```diff
@@ -299,10 +303,36 @@ func (t ResourceTranslator) translateHTTPFilters(filters []gwv1beta1.HTTPRouteFi
 			filter.URLRewrite.Path.Type == gwv1beta1.PrefixMatchHTTPPathModifier {
 			urlRewrite = &api.URLRewrite{Path: DerefStringOr(filter.URLRewrite.Path.ReplacePrefixMatch, "")}
 		}
+
+		if filter.ExtensionRef != nil {
+			//get crd from resources map
+			crdFilter, exists := resourceMap.GetExternalFilter(*filter.ExtensionRef)
+			if !exists {
+				//doesn't exist yet, hope this gets handled next reconcillation loop
```
andrewstucki, line 311: can you update this comment assuming the statement below is actually true (I'm still fine with the defensive programming though)? something like?
```suggestion
				// this should never be the case because we only translate a route if it's actually valid, and if we're missing filters during the validation step, then we won't get here
```

andrewstucki (commented):
Couple more thoughts while you're in the thick of things! 😄

control-plane/api-gateway/common/resources.go
```diff
@@ -115,6 +115,7 @@ type ResourceMap struct {
 	tcpRouteGateways      map[api.ResourceReference]*tcpRoute
 	httpRouteGateways     map[api.ResourceReference]*httpRoute
 	gatewayResources      map[api.ResourceReference]*resourceSet
+	externalFilters       map[gwv1beta1.LocalObjectReference]v1alpha1.ExternalRouteFilter
```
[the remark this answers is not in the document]
sarahalsmiller: Updated to use ObjectReference because I was worried about the possibility of two different filter types with the same name but different types.

control-plane/api-gateway/common/resources.go
sarahalsmiller: Went ahead and just removed it since I'm not longer using this function.

control-plane/api-gateway/binding/result.go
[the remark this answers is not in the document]
sarahalsmiller: No longer using this pattern, marking as resolved.

control-plane/api-gateway/binding/validation.go
```diff
@@ -431,6 +431,69 @@ func routeAllowedForListenerHostname(hostname *gwv1beta1.Hostname, hostnames []g
...
+}
+
+// externalRefsKindAllowedOnRoute
+func externalRefsKindAllowedOnRoute(route *gwv1beta1.HTTPRoute) bool {
+	for _, rule := range route.Spec.Rules {
+		for _, filter := range rule.Filters {
+			if !common.FilterIsExternalFilter(filter) {
+				return false
+			}
+		}
```
andrewstucki, line 472:
```suggestion
		if !filtersAllAllowedType(rule.Filters) {
			return false
		}
```

control-plane/api-gateway/common/diff.go
```diff
@@ -149,7 +149,9 @@ func (e entryComparator) httpRouteRulesEqual(a, b api.HTTPRouteRule) bool {
 	return slices.EqualFunc(a.Filters.Headers, b.Filters.Headers, e.httpHeaderFiltersEqual) &&
 		bothNilOrEqualFunc(a.Filters.URLRewrite, b.Filters.URLRewrite, e.urlRewritesEqual) &&
 		slices.EqualFunc(a.Matches, b.Matches, e.httpMatchesEqual) &&
-		slices.EqualFunc(a.Services, b.Services, e.httpServicesEqual)
+		slices.EqualFunc(a.Services, b.Services, e.httpServicesEqual) &&
+		bothNilOrEqualFunc(a.Filters.RetryFilter, b.Filters.RetryFilter, e.retryFiltersEqual) &&
+		bothNilOrEqualFunc(a.Filters.RetryFilter, b.Filters.RetryFilter, e.retryFiltersEqual)
```
andrewstucki, line 154:
```suggestion
		bothNilOrEqualFunc(a.Filters.RetryFilter, b.Filters.RetryFilter, e.retryFiltersEqual) &&
		bothNilOrEqualFunc(a.Filters.TimeoutFilter, b.Filters.TimeoutFilter, e.timeoutFiltersEqual)
```

control-plane/api-gateway/common/helpers.go
```diff
@@ -27,6 +28,25 @@ func NilOrEqual[T ~string](v *T, check string) bool {
...
+	}
+
+	if !DerefEqual(&filter.ExtensionRef.Group, v1alpha1.ConsulHashicorpGroup) {
+		return false
+	}
+
+	switch filter.ExtensionRef.Kind {
+	case v1alpha1.RouteRetryFilterKind:
+		return true
+	case v1alpha1.RouteTimeoutFilterKind:
```
andrewstucki, line 43:
```suggestion
	case v1alpha1.RouteRetryFilterKind, v1alpha1.RouteTimeoutFilterKind:
		return true
```

control-plane/api-gateway/common/translation_test.go
```diff
@@ -354,9 +355,10 @@ func TestTranslator_ToAPIGateway(t *testing.T) {
 func TestTranslator_ToHTTPRoute(t *testing.T) {
 	t.Parallel()
 	type args struct {
-		k8sHTTPRoute gwv1beta1.HTTPRoute
-		services     []types.NamespacedName
-		meshServices []v1alpha1.MeshService
+		k8sHTTPRoute    gwv1beta1.HTTPRoute
+		services        []types.NamespacedName
+		meshServices    []v1alpha1.MeshService
+		externalFilters map[gwv1beta1.LocalObjectReference]v1alpha1.ExternalRouteFilter
```
andrewstucki, line 361: should this be updated to take some arbitrary `ObjectReference` filters in and out of the namespace just to test the namespace alignment checking?

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -777,11 +802,80 @@ func (c *GatewayController) getRelatedHTTPRoutes(ctx context.Context, gateway ty
...
+				c.Log.Info(fmt.Sprintf("externalref %s:%s not found: %v", filter.ExtensionRef.Kind, filter.ExtensionRef.Name, err))
+				//ignore, the validation call should mark this route as error
+			} else {
+				return nil, err
+			}
+		}
+
+		//add external ref (or error) to resource map for this route
+		resources.AddExternalFilter(*filter.ExtensionRef, externalFilter.(v1alpha1.ExternalRouteFilter))
+		externalFilters = append(externalFilters, externalFilter)
```
andrewstucki, line 874: I think you'll want a `continue` here to just not insert a blank filter into the resource map. That way the filter won't be found when you try and validate its presence in the binding logic (Sorry, the diff looks a bit bad, but in case you can't find it, it's inserted under the comment starting with `//ignore`)

```suggestion
			if k8serrors.IsNotFound(err) {
				c.Log.Info(fmt.Sprintf("externalref %s:%s not found: %v", filter.ExtensionRef.Kind, filter.ExtensionRef.Name, err))
				//ignore, the validation call should mark this route as error
				continue
			} else {
				return nil, err
			}
		}

		//add external ref (or error) to resource map for this route
		resources.AddExternalFilter(*filter.ExtensionRef, externalFilter.(v1alpha1.ExternalRouteFilter))
		externalFilters = append(externalFilters, externalFilter)
```

control-plane/config/crd/patches/webhook_in_httprouteretryfilters.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) HashiCorp, Inc.
```
andrewstucki, line 1: I think that some of these files are probably not necessary? Do we have the equivalent generated elsewhere or are we missing/including too many kubebuilder tags?

control-plane/config/crd/external/.yml
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+
```
andrewstucki, line 4: 🙃

andrewstucki (commented):
Great job on some of the changes requested. I think you're basically there, just a few things that look like they might need a bit of cleanup and then I'll take one more pass.

control-plane/api-gateway/controllers/gateway_controller.go
sarahalsmiller: Good catch, definitely meant for there to be a continue there.

control-plane/config/crd/patches/webhook_in_httprouteretryfilters.yaml
sarahalsmiller: I believe they were generated as part of the initial boilerplate generating command. I just went ahead and cleaned them up since they don't appear to be tied to our kubebuilder tags.

control-plane/config/crd/kustomization.yaml
```diff
@@ -6,17 +6,23 @@
 # It should be run by config/default
```
andrewstucki, line 6: think this can get reverted

control-plane/config/rbac/httprouteretryfilter_editor_role.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) HashiCorp, Inc.
```
andrewstucki, line 1: this file can be dropped

control-plane/config/rbac/httprouteretryfilter_viewer_role.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) HashiCorp, Inc.
```
andrewstucki, line 1: this one too

control-plane/config/rbac/httproutetimeoutfilter_editor_role.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) HashiCorp, Inc.
```
andrewstucki, line 1: and here

control-plane/config/rbac/httproutetimeoutfilter_viewer_role.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) HashiCorp, Inc.
```
andrewstucki, line 1: and finally this one

control-plane/api-gateway/common/routefilter.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package common
+
+// +kubebuilder:skipversion
```
andrewstucki, line 6: Wondering what this tag is doing here? I don't think anything in this file involves `kubebuilder`?

control-plane/api-gateway/common/routefilter.go
```diff
@@ -0,0 +1,10 @@
...
+
+type ExternalRouteFilter interface {
```
andrewstucki, line 8: Is there a reason that we need this over say leveraging `client.Object` in the places this is used? Doesn't seem like this has anything route-specific included in it (all `client.Object` instances seem like they'd implent this) and the CRD objects shoud already implement `client.Object`.

control-plane/PROJECT
```diff
@@ -108,4 +108,22 @@ resources:
...
+  version: v1alpha1
+- api:
+    crdVersion: v1beta1
+    namespaced: true
+  controller: true
+  domain: hashicorp.com
+  group: consul
+  kind: HTTPRouteTimeoutFilter
+  path: github.com/hashicorp/consul-k8s/control-plane/api/v1alpha1
+  version: v1alpha1
```
andrewstucki, line 128: do you know what this file is used for? I don't see any of our other CRDs here

.changelog/2735.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+api-gateway: add HTTPRouteRetryFilter and HTTPRouteTimeoutFilter CRDs
```
andrewstucki, line 2: Based off of the current name of the CRDs

```suggestion
api-gateway: add RouteRetryFilter and RouteTimeoutFilter CRDs
```

andrewstucki (commented):
last bit of cleanup that I see

control-plane/config/crd/bases/consul.hashicorp.com_httproutetimeoutfilters.yaml
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) HashiCorp, Inc.
```
andrewstucki, line 1: sorry, think this file needs to be deleted too, looks like it's part of the old CRD before you renamed it.

control-plane/api-gateway/common/routefilter.go
sarahalsmiller: Was an accident, I'll clean it up

control-plane/api-gateway/common/routefilter.go
sarahalsmiller: I think I thought we would need more specific things than just get namespace initially. I'll swap over to client.Object

control-plane/api-gateway/common/resources.go
```diff
@@ -364,6 +365,35 @@ func (s *ResourceMap) ReferenceCountHTTPRoute(route gwv1beta1.HTTPRoute) {
...
 
+func localObjectReferenceToObjectReference(filterRef gwv1beta1.LocalObjectReference, namespace string) corev1.ObjectReference {
+	return corev1.ObjectReference{
+		Kind:      string(filterRef.Kind),
+		Name:      string(filterRef.Name),
+		Namespace: namespace,
+	}
+}
+
+func (s *ResourceMap) AddExternalFilter(filterRef gwv1beta1.LocalObjectReference, filter client.Object) {
```
andrewstucki, line 376: Since you have the full object in this case, you could just construct the key from it a la:

```
 corev1.ObjectReference{
    Kind:      filter.GetKind(),
    Name:      filter.GetName(),
    Namespace: filter.GetNamespace(),
}
```

And drop the need to pass the `filterRef` in here at all. I'll leave it up to you as to whether or not you want to change it though, as it's at least somewhat shaped like the "get" methods.

control-plane/api-gateway/binding/route_binding.go
```diff
@@ -163,6 +163,31 @@ func (r *Binder) bindRoute(route client.Object, boundCount map[gwv1beta1.Section
...
 		})
+
+		httproute, ok := route.(*gwv1beta1.HTTPRoute)
+		if ok {
+			if !externalRefsOnRouteAllExist(httproute, r.config.Resources) {
+				results = append(results, parentBindResult{
+					parent: ref,
+					results: []bindResult{
+						{
+							err: errExternalRefNotFound,
```
andrewstucki, line 174: I don't recall what happens when you have two binding results for the same parent - do the statuses/syncing still work out if the route successfully bound as per above and then has an error here? I'd assume it's an ordering thing and whatever the last bind result is for the parent is what's actually used?

andrewstucki (approved):
Approving, but have a small question/comment or two. Thanks for all of the work @sarahalsmiller!

control-plane/api-gateway/binding/route_binding.go
sarahalsmiller: I believe its the last bind result, yes.

PR #2743, opened 2023-08-08
Net 1784 inject sidecar first

trevorLeonHC:
Changes proposed in this PR:
- If lifecycle setting is enabled, inject consul-dataplane as the first container in the pod. Add unit tests where lifecycle is enabled in multiport enabled/disabled setting. 

How I've tested this PR:
I've added tests that specifically check the container ordering when lifecycle is turned on. 

Checklist:
- [x] Tests added

### DIFF
--- a/.changelog/2743.txt
+++ b/.changelog/2743.txt
@@ -1,5 +1,3 @@
 ```release-note:improvement
-control-plane: Changed the container ordering in connect-inject to insert consul-dataplane
-container first if lifecycle is enabled. Container ordering is unchanged if lifecycle
-is disabled. 
+control-plane: Changed the container ordering in connect-inject to insert consul-dataplane container first if lifecycle is enabled. Container ordering is unchanged if lifecycle is disabled. 
 ```
--- a/.changelog/2743.txt
+++ b/.changelog/2743.txt
@@ -1,0 +1,5 @@
+```release-note:improvement
+control-plane: Changed the container ordering in connect-inject to insert consul-dataplane
+container first if lifecycle is enabled. Container ordering is unchanged if lifecycle
+is disabled. 
+```
--- a/control-plane/connect-inject/webhook/mesh_webhook.go
+++ b/control-plane/connect-inject/webhook/mesh_webhook.go
@@ -333,6 +333,13 @@ func (w *MeshWebhook) Handle(ctx context.Context, req admission.Request) admissi
 			w.Log.Error(err, "checking unsupported cases for multi port pods")
 			return admission.Errored(http.StatusInternalServerError, err)
 		}
+		lifecycleEnabled, ok := w.LifecycleConfig.EnableProxyLifecycle(pod)
+		if ok != nil {
+			lifecycleEnabled = false
+		}
+		//List of sidecar containers for each service. Build list to preserve correct ordering in relation
+		//to services.
+		sidecarContainers := []corev1.Container{}
 		for i, svc := range annotatedSvcNames {
 			w.Log.Info(fmt.Sprintf("service: %s", svc))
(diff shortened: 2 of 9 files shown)


.changelog/2743.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:improvement
+control-plane: Changed the container ordering in connect-inject to insert consul-dataplane
```
curtbushko, line 2: **nitpick:** Please remove line breaks. Your editor probably added them automatically (mine does this all the time).

control-plane/connect-inject/webhook/mesh_webhook.go
```diff
@@ -311,8 +311,14 @@ func (w *MeshWebhook) Handle(ctx context.Context, req admission.Request) admissi
 			w.Log.Error(err, "error configuring injection sidecar container", "request name", req.Name)
 			return admission.Errored(http.StatusInternalServerError, fmt.Errorf("error configuring injection sidecar container: %s", err))
 		}
-		// TODO: invert to start the Envoy sidecar before the application container
-		pod.Spec.Containers = append(pod.Spec.Containers, envoySidecar)
+		//Append the Envoy sidecar before the application container only if lifecycle enabled.
+		lifecycleEnabled, ok := w.LifecycleConfig.EnableProxyLifecycle(pod)
```
curtbushko, line 315: **suggestion:** You call this several times. Instead call it once before the multiport section (around line 297) and use it over and over.

control-plane/connect-inject/webhook/mesh_webhook.go
```diff
@@ -382,9 +395,17 @@ func (w *MeshWebhook) Handle(ctx context.Context, req admission.Request) admissi
 				w.Log.Error(err, "error configuring injection sidecar container", "request name", req.Name)
 				return admission.Errored(http.StatusInternalServerError, fmt.Errorf("error configuring injection sidecar container: %s", err))
 			}
-			// TODO: invert to start the Envoy sidecar container before the
-			// application container
-			pod.Spec.Containers = append(pod.Spec.Containers, envoySidecar)
+			if !lifecycleEnabled {
```
curtbushko, line 398: **suggestion:** Please flip this if statement as all the other if statements are checking for it enabled; this is the only that checks if it is not enabled. Also, please add a little comment about why you are doing this.

control-plane/connect-inject/webhook/mesh_webhook_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	mapset "github.com/deckarep/golang-set"
 	logrtest "github.com/go-logr/logr/testr"
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/lifecycle"
```
curtbushko, line 15: **thought:** Maybe consult with Thomas on this one. I am surprised that we do not have a webhook test that looks at the actual mutated pod itself?

I was thinking that a test could be added that is very similar to `TestHandlerHandle` with your initial pod and then instead of `Patches` you compare against a pod spec.

I won't block on this idea though..

curtbushko (dismissed):
Looking pretty good!  I have a few suggestions and one thought for you to ponder if you want to add it or not...

control-plane/connect-inject/webhook/mesh_webhook_test.go
t-eckert: That is a super fair point. I'll make a note of that to come back and write that unit test in the future.

missylbytes (approved):

PR #2758, opened 2023-08-11
Updates changelog to include 1.0.9

missylbytes:
Changes proposed in this PR:
Updates changelog to include 1.0.9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,36 @@
+## 1.0.9 (Aug 10, 2023)
+
+SECURITY:
+
+* Upgrade to use Go 1.19.11 and `x/net/http` 0.12.0.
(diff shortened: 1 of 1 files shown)


curtbushko (approved):
LGTM!

DanStough (approved):
IMHO we should really put this in descending order since they were released on the same day, i.e. 1.1.4 should be above 1.0.9. What happened to 1.2.1?

missylbytes:
> IMHO we should really put this in descending order since they were released on the same day, i.e. 1.1.4 should be above 1.0.9. What happened to 1.2.1?

1.2.1 will be in [the next pr](https://github.com/hashicorp/consul-k8s/pull/2768), will reorder them.

PR #2771, opened 2023-08-14
Add readOnlyRootFilesystem to security context

mr-miles:
Changes proposed in this PR:
- Explicitly label container security context as having a readonly root filesystem

How I've tested this PR:

I was attempting to run consul in a restricted namespace and our gatekeeper policy tripper on this.  I found I could not add the context by manipulating the helm values and so a PR was required,

I have tested it by deploying my updated package and confirming gatekeeper was not tripped

Confirm it deploys fine in a standard test environment just for sanity: as the components don't write this labelling doesn't affect functionality and so is a no-op for anyone not already blocked by  security context.

Checklist:
- [ ] Tests added

### DIFF
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -19,6 +19,7 @@ as well as the global.name setting.
 {{- if not .Values.global.enablePodSecurityPolicies -}}
 securityContext:
   allowPrivilegeEscalation: false
+  readOnlyRootFilesystem: true
   capabilities:
     drop:
     - ALL
--- a/charts/consul/templates/ingress-gateways-deployment.yaml
+++ b/charts/consul/templates/ingress-gateways-deployment.yaml
@@ -244,6 +244,8 @@ spec:
         resources: {{ toYaml (default $defaults.resources .resources) | nindent 10 }}
         {{- end }}
         volumeMounts:
+        - name: tmp
+          mountPath: /tmp
         - name: consul-service
           mountPath: /consul/service
           readOnly: true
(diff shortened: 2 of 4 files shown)


david-yu:
@mr-miles Could you tell us how you are using GateKeeper and why a policy in gatekeeper is requiring this? Thanks for the PR.

mr-miles:
Sorry @david-yu, you caught this before I'd made a ticket with more explanation in!

Our network team use gatekeeper to enforce their security policy over our k8s cluster and previously consul was exempted from all the checks.  There were a couple of updates to the helm chart recently to allow consul + sidecar to run in a restricted/unprivileged environment so we attempted to re-enabled it. 

On doing so, our "readonlyfilesystem" validation was flagging, and preventing sidecars from starting (I think our restricted environment is a little more restricted than just non-root user accounts).  The change to readonly containers was small - but useful security hardening (plus it has caught some bugs for us in other areas in the past) - hence submitting a PR.  Do you think this is a useful addition?

david-yu:
Sorry did you mean you used the latest changes from https://github.com/hashicorp/consul-k8s/pull/2755 to run in an un-privileged mode, and these were extra settings needed to allow your sidecars to run? In this case gateways are writing to `tempfs ` instead of disk?

mr-miles:
Yes that's it - they are extra settings that I needed to make everything work.  Though it was the changes in #2702 and #2572 that I had been looking at.

Thinking about it, it doesn't need to be medium=memory since the bootstrap file doesn't have anything sensitive in.  Any mount (disk or memory) of /tmp will avoid the "container filesystem is readonly" error message.

pglass (approved):

pglass:
@mr-miles If you can copy over this Helm chart unit test fix: https://github.com/hashicorp/consul-k8s/pull/2788/commits/d38380c088aa6722421f45cc122d29d909577d66, then this PR should be good to merge from my perspective.

(Our CI tests currently won't run successfully for PRs created forks unfortunately 😞)

mr-miles:
@pglass - thanks for looking at this - done!

david-yu:
What version of the Helm chart are you? @mr-miles trying to understand what releases to backport this into. We can at least do consul k8s 1.2.x.

mr-miles:
1.2.x is fine for me - we're on 1.2.1

PR #2776, opened 2023-08-15
Mw/net 4260 add tproxy coverage

wilkermichael:
Changes proposed in this PR:
- Adds test coverage for tproxy

How I've tested this PR:
👀 

👀 

Checklist:
- [ ] Tests added

### DIFF
--- a/acceptance/tests/sameness/sameness_test.go
+++ b/acceptance/tests/sameness/sameness_test.go
@@ -428,21 +428,30 @@ func TestFailover_Connect(t *testing.T) {
 				"../fixtures/cases/sameness/static-server/dc3")
 
 			// Create static client deployments.
+			staticClientDefaultKustomizeDir := "../fixtures/cases/sameness/static-client/default-partition"
+			staticClientAP1KustomizeDir := "../fixtures/cases/sameness/static-client/ap1-partition"
+
+			// If transparent proxy is enabled create clients without explicity upstreams
+			if cfg.EnableTransparentProxy {
+				staticClientDefaultKustomizeDir = fmt.Sprintf("%s-%s", staticClientDefaultKustomizeDir, "tproxy")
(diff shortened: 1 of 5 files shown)


acceptance/tests/sameness/sameness_test.go
```diff
@@ -428,21 +428,30 @@ func TestFailover_Connect(t *testing.T) {
 				"../fixtures/cases/sameness/static-server/dc3")
 
 			// Create static client deployments.
+			staticClientDefaultKustomizeDir := "../fixtures/cases/sameness/static-client/default-partition"
+			staticClientAP1KustomizeDir := "../fixtures/cases/sameness/static-client/ap1-partition"
```
zalimeni, line 432: ~~ Total nit, but it might be easier to scan similar names like these if we put the different segment last, like
```suggestion
			staticClientKustomizeDirDefault := "../fixtures/cases/sameness/static-client/default-partition"
			staticClientKustomizeDirAP1 := "../fixtures/cases/sameness/static-client/ap1-partition"
```

acceptance/tests/sameness/sameness_test.go
```diff
@@ -428,21 +428,30 @@ func TestFailover_Connect(t *testing.T) {
...
+
+			// If transparent proxy is enabled create clients without explicity upstreams
```
zalimeni, line 434:
```suggestion
			// If transparent proxy is enabled create clients without explicit upstreams
```

zalimeni (approved):
One minor comment, but based on what's already in place this addition LGTM!

hashi-derek (approved):
Thanks for finding this gap in the tests and making sure it's covered.

acceptance/tests/sameness/sameness_test.go
wilkermichael: I see Gollum wrote that comment "explicity hobitsys"

PR #2357, opened 2023-06-13
Add CRDs for `ProritizeByLocality` Routing

erichaberkorn:
Changes proposed in this PR:
- Add `prioritizeByLocality` to `ServiceResolver` CRD

How I've tested this PR:
I spun up a consul-k8s cluster and tried setting this field to different values.

How I expect reviewers to test this PR:
👁️ 

Checklist:
- [X] Tests added
- [X] CHANGELOG entry added

PR #2784, opened 2023-08-16
[NET-2880] Add `PrioritizeByLocality` to `ProxyDefaults` CRD

zalimeni:
In addition to service resolver, add this field to the CRD for proxy defaults for parity with Consul config options.

Follow-up to https://github.com/hashicorp/consul-k8s/pull/2357 (mostly replicates that PR +/- shared code moves)

Changes proposed in this PR:
- Add `prioritizeByLocality` field to CRD for proxy defaults
- Move handling code to shared for reuse
- Update and add tests
- Update existing changelog entry to reflect both CRDs (we haven't released or backported this, so using the existing changelog entry should be ok)

How I've tested this PR:
Spun up a cluster and tested setting valid and invalid values

Checklist:
- [x] Tests added


control-plane/api/v1alpha1/shared_types.go
```diff
@@ -310,6 +310,35 @@ func (in *FailoverPolicy) validate(path *field.Path) field.ErrorList {
 	return errs
 }
 
+type PrioritizeByLocality struct {
```
wilkermichael, line 313: Just to conform with the other structs in this file, can you add a comment for this public struct?

wilkermichael (approved):
Looks good to me!

control-plane/api/v1alpha1/shared_types.go
zalimeni: Ah, thanks for catching! Absolutely.

hashi-derek (approved):

PR #2785, opened 2023-08-17
Net 5229 create dedicated argocd stanza

absolutelightning:
Changes proposed in this PR:

* Add support for global.argocd.enabled flag in values.yaml which adds the 
```
        "argocd.argoproj.io/hook": "Sync"
        "argocd.argoproj.io/hook-delete-policy": "HookSucceeded"
```
 in the jobs -
server-acl-init-job

How I've tested this PR:
* Bats Tests
Manual Tests - 
* Created github pages to host helm charts following [this medium article](https://medium.com/@mattiaperi/create-a-public-helm-chart-repository-with-github-pages-49b180dbb417)
* Forked the repo and created a new branch - https://github.com/absolutelightning/consul-minikubes/tree/test-argocd.
* Updated the repo with values from github helm charts created in step 1
* Ran `bash argocd.sh`
* kubectl port-forward svc/argocd-server -n argocd 8087:443
* argocd admin initial-password -n argocd
* Input username - admin and password output from previous step in UI at localhost:8087.
* Recorded video

https://github.com/hashicorp/consul-k8s/assets/134911583/e09b51ff-e02e-44af-a216-440c38275bee

Checklist:
- [x] Tests added


aahel (approved):

Ganeshrockz (approved):
LGTM

lkysow:
I don't think you need those annotations on all the jobs, just the jobs that don't have helm hook annotations already: https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks

> Argo CD supports many (most?) Helm hooks by mapping the Helm annotations onto Argo CD's own hook annotations:

Also how can we ensure Jobs in the future get this hook added? Can you write a test that ensures `Job`'s that don't have helm hook annotations also have this argocd stanza.

absolutelightning:
Hey Luke, I have trouble understanding your comment. 
1. I checked [this link](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks) I need `argocd.argoproj.io/hook: Sync` The documentation doesn't have - it only has PreSync.
2. How can I write test for job which will come in future. Is there a way?

lkysow:
Have you tried this out? I don't think it will work as written. For example, the gateway cleanup job is marked as a pre-delete job but this change maps it to an argocd `Sync` hook. So previously the hook would only run right before the user deletes consul but now it's running every single time they do a deploy of the helm chart.

The gateway-resources job is marked at post-install,post-upgrade which based on the docs is automatically mapped to an argocd `PostSync` but your change makes it `Sync`. Same for `server-acl-init-cleanup`.

I think you don't need to touch the gateway resources or server-acl-init-cleaup jobs but would be best to test that and verify. For the gateway-cleanup job I'm not exactly sure what that does so will need to investigate.

In terms of testing, I was thinking a test similar to https://github.com/hashicorp/consul-k8s/blob/e5ad4475cccc64f5142e434ea165a5f822ca0eec/charts/consul/test/unit/helpers.bats#L90-L134 which checks that all files that have a `Job` either have helm hook annotations _or_ the argocd templating. So that if someone adds a Job in the future without either of those annotations that the test will fail to remind them to add one of the annotations.

curtbushko:
I think this needs to be tested with a Kind cluster and ArgoCD, there are too many unknowns.

[Here](https://github.com/joatmon08/consul-minikubes/blob/argocd/argocd.sh) is a script that you can use as a starting point for setting everything up...

.changelog/2785.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:improvement
+If argocd.enabled is set to true, following annotations are added to job - server-acl-init-job
+ argocd.argoproj.io/hook: Sync
+ argocd.argoproj.io/hook-delete-policy: HookSucceeded
```
lkysow, line 4:
```suggestion
Add new value `global.argocd.enabled`. Set this to `true` when using ArgoCD to deploy this chart.
```

I think best to just say what the improvement gives users.

lkysow (approved):
Great work!

curtbushko (approved):

david-yu:
Added backports to the PR.

david-yu:
@absolutelightning Could you make sure this gets backported? Thank you!

absolutelightning:
sure @david-yu will do it today. Thanks.

PR #2787, opened 2023-08-17
NET-5186 Add NET_BIND_SERVICE capability to Consul's restricted securityContext

nathancoleman:
Changes proposed in this PR:
- Add the `NET_BIND_SERVICE` capability for consul-dataplane related to https://github.com/hashicorp/consul-dataplane/pull/238

How I've tested this PR:
- Install consul w/ ingress-gateway serving on a privileged port such as `443`
- Install consul under restricted pod security policy
- Install consul under restricted[-v2] OpenShift security context constraints

- See above

Checklist:
- [ ] Tests added

### DIFF
--- a/.changelog/2787.txt
+++ b/.changelog/2787.txt
@@ -1,3 +1,3 @@
-```release-note:security
+```release-note:improvement
 Add NET_BIND_SERVICE capability to restricted security context used for consul-dataplane
 ```
--- a/.changelog/2787.txt
+++ b/.changelog/2787.txt
@@ -1,0 +1,3 @@
+```release-note:security
+Add NET_BIND_SERVICE capability to restricted security context used for consul-dataplane
+```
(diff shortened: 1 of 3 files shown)


david-yu (approved):
This sounds like the best solution to allow privileged ports for now. Should work with both GKE Autopilot and OpenShift.

.changelog/2787.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:security
+Add NET_BIND_SERVICE capability to restricted security context used for consul-dataplane
```
nathancoleman, line 2: @david-yu do you have thoughts on how this appears in the changelog? I was torn between `improvement` and `security` as the not category and wasn't sure if it made sense to include any more details here or a link to the related consul-dataplane PR which requires this change.

david-yu: I believe this is more of an improvement than a security change which is typically reserved for addressing CVEs.

nathancoleman: Updated

curtbushko (approved):
Thanks Nathan!

Issue #751, opened 2021-09-29
Set secure defaults for containerSecurityContext fields


tarquin-the-brave:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Is your feature request related to a problem? Please describe.

Consul components running with insecure kubernetes config.

#### Feature Description

Include secure defaults for the containerSecurityContext fields in the Helm chart.

Hi.  A few weeks back I raised #602 and submitted #620 to allow the [securityContext][sc] on the containers in the server and client Pods to be configured through the Helm chart.

[sc]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core

I said once we (Microsoft Azure for Operators) had tested configuring these and found the most secure values they could take, we'd contribute those back as defaults to the chart.

It turns out our use case only needed to deploy the Consul servers via the Helm chart so we've only tested setting `server.containerSecurityContext`.

We found these can be set to:

```yaml
server:
  containerSecurityContext:
    privileged: false
    capabilities:
      drop:
        - ALL
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: true
```

without affecting the functioning of the Consul server.

We haven't tested setting this for the client Pod.  Looking at the `client-daemonset.yaml` template my guess would be that the `consul` container will be OK with the same securityContext as the server above.

For the initContainers `client-acl-init` & `client-tls-init` you might have to relax `readOnlyRootFileSystem` to `false`.  It seems from your Pod spec that these initContainers write files to mounted volumes.  I can't recall if `readOnlyRootFileSystem` prevents that or not, I can't tell from [the docs][sc].  Something to test. 

#### Use Case(s)

All those who deploy with the Helm chart can get the most secure [container secuirtyContext][sc] values by default.

#### Contributions

We could contribute these defaults for the server, but I imagine you'd want to implement the "secure defaults" for the client Pod containers too, which we haven't tested.

david-yu:
[...]

PR #2789, opened 2023-08-17
Add readOnlyRootFilesystem to security context (#2771)

david-yu:
* Bring in PR #2771 and run with CI which fixes https://github.com/hashicorp/consul-k8s/issues/751 

Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added

### DIFF
--- a/.changelog/2781.txt
+++ b/.changelog/2781.txt
@@ -1,3 +1,0 @@
-```release-note:improvement
-helm: Add readOnlyRootFilesystem to the default restricted security context when runnning `consul-k8s` in a restricted namespaces. 
-```
--- a/.changelog/2781.txt
+++ b/.changelog/2781.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+helm: Add readOnlyRootFilesystem to the default restricted security context when runnning `consul-k8s` in a restricted namespaces. 
+```
--- a/.changelog/2789.txt
+++ b/.changelog/2789.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+helm: Add readOnlyRootFilesystem to the default restricted security context when runnning `consul-k8s` in a restricted namespaces. 
+```
--- a/charts/consul/templates/_helpers.tpl
+++ b/charts/consul/templates/_helpers.tpl
@@ -19,6 +19,7 @@ as well as the global.name setting.
 {{- if not .Values.global.enablePodSecurityPolicies -}}
 securityContext:
   allowPrivilegeEscalation: false
+  readOnlyRootFilesystem: true
   capabilities:
     drop:
     - ALL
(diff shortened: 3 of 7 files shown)


pglass (approved):

.changelog/2781.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
david-yu, line 1: @mr-miles Does this sound right?

mr-miles: 👍

david-yu:
@mr-miles It looks like the PRs were able to pass more tests after adding the `tmp` volume mount to the server statefulset as well. Could you perhaps provide us some details on how this tmp volume is used, I'm not exactly sure why that fixes things, is it because the logs of raft logs are now written to tmp instead of folders we can write to?

mr-miles:
With the readonlyrootvolume set to false, the server could write to
anywhere on the filesystem.

With it set to true, the container can only write to volumes that you’ve
mounted.

So my guess is that the server had always even writing some files into /tmp
(maybe ones you don’t care about, as all the important bits are persisted
in datadir) but now you need to make it explicit with the mount.

(If you can shell into the running pod you can see what files are there.
I’m afk otherwise I’d look)

curtbushko (commented):
LGTM!

david-yu:
@mr-miles Thanks again for your contribution, will go ahead and merge and backport to 1.2.x only to reduce blast radius.

curtbushko (approved):

t-eckert:
Update, this is breaking Dataplane and all acceptance tests because Dataplane writes to disk when it starts up: https://github.com/hashicorp/consul-dataplane/blob/da6bf688df5db0e93c8b5c1e9e47227a34e51ea1/pkg/envoy/proxy.go#L337

This is read by Envoy for doing bootstrapping.

mr-miles:
I couldn't work out how the consul-k8s helm chart is being utilised by the
integration tests, but FWIW I think the tests need to be updated to mount a
/tmp volume and so match the helm chart:

https://github.com/hashicorp/consul-dataplane/blob/da6bf688df5db0e93c8b5c1e9e47227a34e51ea1/integration-tests/helpers/dataplane.go#L76

and maybe also

https://github.com/hashicorp/consul-dataplane/blob/da6bf688df5db0e93c8b5c1e9e47227a34e51ea1/integration-tests/helpers/server.go#L82

Then dataplane can write to /tmp in line with what it is expecting

david-yu:
@mr-miles It looks like this was likely not the culprit that broke our tests. I'll re-raise the PR once our pipelines have stabilized.

mr-miles:
Thanks David - no problem (glad not to be implicated!), happy to help if it
is still causing grief after that

PR #2790, opened 2023-08-18
added check if anonymous token policy exists

aahel:
Changes proposed in this PR:

Prevent updating anonymous token if it already exists and is attached to the anonymous token policy

How I've tested this PR:

- setup the deafault partiton and create a non-default partition called my-partition.
- modified the anonymous token policy rule.
- created helm deployment in my-partiton using acl bootstrap token with following policies.
```hcl
acl = "read"
operator = "read"
agent_prefix "" {
  policy = "read"
}
```
```hcl
partition "my-partition" {
  acl = "write"
  mesh = "write"
  peering = "write"
  namespace_prefix "" {
    policy = "write"
  }
  service_prefix "" {
    policy = "write"
  }
}
```

- server-acl-acl init was successfull and anonymous token policy rule didn't get updated

Checklist:
- [x] Tests added

### DIFF
--- a/control-plane/subcommand/server-acl-init/anonymous_token.go
+++ b/control-plane/subcommand/server-acl-init/anonymous_token.go
@@ -7,6 +7,11 @@ import (
 	"github.com/hashicorp/consul/api"
 )
 
+const (
+	AnonymousTokenPolicyName = "anonymous-token-policy"
+	AnonymousTokenAccessorID = "00000000-0000-0000-0000-000000000002"
+)
+
 // configureAnonymousPolicy sets up policies and tokens so that Consul DNS and
 // cross-datacenter Consul connect calls will work.
 func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
@@ -17,6 +22,7 @@ func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
 	if exists {
 		return nil
 	}
+
 	anonRules, err := c.anonymousTokenRules()
 	if err != nil {
 		c.log.Error("Error templating anonymous token rules", "err", err)
@@ -25,7 +31,7 @@ func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
 
 	// Create policy for the anonymous token
 	anonPolicy := api.ACLPolicy{
-		Name:        "anonymous-token-policy",
+		Name:        AnonymousTokenPolicyName,
 		Description: "Anonymous token Policy",
 		Rules:       anonRules,
 	}
@@ -40,7 +46,7 @@ func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
 
 	// Create token to get sent to TokenUpdate
 	aToken := api.ACLToken{
-		AccessorID: "00000000-0000-0000-0000-000000000002",
+		AccessorID: AnonymousTokenAccessorID,
 		Policies:   []*api.ACLTokenPolicyLink{{Name: anonPolicy.Name}},
 	}
 
@@ -53,24 +59,16 @@ func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
 }
 
 func checkIfAnonymousTokenPolicyExists(consulClient *api.Client) (bool, error) {
-	token, _, err := consulClient.ACL().TokenRead("00000000-0000-0000-0000-000000000002", nil)
+	token, _, err := consulClient.ACL().TokenRead(AnonymousTokenAccessorID, nil)
 	if err != nil {
 		return false, err
 	}
-	existingPolicies, _, err := consulClient.ACL().PolicyList(&api.QueryOptions{})
-	if err != nil {
-		return false, err
-	}
-	policyID := ""
-	for _, existingPolicy := range existingPolicies {
-		if existingPolicy.Name == "anonymous-token-policy" && existingPolicy.Description == "Anonymous token Policy" {
-			policyID = existingPolicy.ID
-		}
-	}
+
 	for _, policy := range token.Policies {
-		if policy.ID == policyID {
+		if policy.Name == AnonymousTokenPolicyName {
 			return true, nil
 		}
 	}
+
 	return false, nil
 }
--- a/control-plane/subcommand/server-acl-init/anonymous_token.go
+++ b/control-plane/subcommand/server-acl-init/anonymous_token.go
@@ -8,8 +8,8 @@ import (
 )
 
 const (
-	AnonymousTokenPolicyName = "anonymous-token-policy"
-	AnonymousTokenAccessorID = "00000000-0000-0000-0000-000000000002"
+	anonymousTokenPolicyName = "anonymous-token-policy"
+	anonymousTokenAccessorID = "00000000-0000-0000-0000-000000000002"
 )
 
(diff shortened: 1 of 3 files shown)


control-plane/subcommand/server-acl-init/anonymous_token.go
```diff
@@ -44,3 +51,26 @@ func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
 			return err
 		})
 }
+
+func checkIfAnonymousTokenPolicyExists(consulClient *api.Client) (bool, error) {
+	token, _, err := consulClient.ACL().TokenRead("00000000-0000-0000-0000-000000000002", nil)
```
Ganeshrockz, line 56: You could move the anonymous token ID as a constant since it is being referenced at other places in this file

control-plane/subcommand/server-acl-init/anonymous_token.go
```diff
@@ -44,3 +51,26 @@ func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
...
+	if err != nil {
+		return false, err
+	}
+	existingPolicies, _, err := consulClient.ACL().PolicyList(&api.QueryOptions{})
```
Ganeshrockz, line 60: You could use `consulClient.ACL().PolicyReadByName(anonPolicyName, qopts)` instead of this.

control-plane/subcommand/server-acl-init/anonymous_token.go
```diff
@@ -44,3 +51,26 @@ func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
...
+	if err != nil {
+		return false, err
+	}
+	policyID := ""
+	for _, existingPolicy := range existingPolicies {
+		if existingPolicy.Name == "anonymous-token-policy" && existingPolicy.Description == "Anonymous token Policy" {
+			policyID = existingPolicy.ID
+		}
+	}
+	for _, policy := range token.Policies {
```
Ganeshrockz, line 70: Wondering why this can't be the only check needed to verify the policy's presence? Something like

```go
for _, policy := range token.Policies {
      if policy.Name == anonPolicyName {
                  return true, nil
      }
}
```

Why should we list all the policies and verify with the ID?

control-plane/subcommand/server-acl-init/anonymous_token.go
```diff
@@ -10,6 +10,13 @@ import (
...
 // cross-datacenter Consul connect calls will work.
 func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
+	exists, err := checkIfAnonymousTokenPolicyExists(consulClient)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return nil
+	}
 	anonRules, err := c.anonymousTokenRules()
```
Ganeshrockz, line 20: nit: Add a new line to improve readability.

control-plane/subcommand/server-acl-init/anonymous_token.go
```diff
@@ -44,3 +51,26 @@ func (c *Command) configureAnonymousPolicy(consulClient *api.Client) error {
...
+	if err != nil {
+		return false, err
+	}
+	existingPolicies, _, err := consulClient.ACL().PolicyList(&api.QueryOptions{})
+	if err != nil {
+		return false, err
+	}
+	policyID := ""
+	for _, existingPolicy := range existingPolicies {
+		if existingPolicy.Name == "anonymous-token-policy" && existingPolicy.Description == "Anonymous token Policy" {
```
Ganeshrockz, line 66: `anonymous-token-policy` could be made a constant

control-plane/subcommand/server-acl-init/anonymous_token.go
```diff
@@ -7,9 +7,22 @@ import (
 	"github.com/hashicorp/consul/api"
 )
 
+const (
+	AnonymousTokenPolicyName = "anonymous-token-policy"
```
Ganeshrockz, line 11: Both of these can be a private constants

aahel: done

aahel:
> It would be good to add test for the changes in `configureAnonymousPolicy` function

looks like we have a test called `TestRun_AnonymousTokenPolicy` in command_test.go

control-plane/subcommand/server-acl-init/anonymous_token_test.go
```diff
@@ -0,0 +1,74 @@
...
+	readToken := resp.SecretID
+
+	readOnlyClient, errz := api.NewClient(&api.Config{
+		Address: consulHTTPAddr,
+		Token:   readToken,
+	})
+	require.NoError(t, errz)
+	// does not create/update anonymous token policy
+	erry := cmd.configureAnonymousPolicy(readOnlyClient)
+	require.NoError(t, erry)
```
Ganeshrockz, line 73: Nit: You can use the same `err` variable for consistency

Ganeshrockz (commented):
Looks good to me.  Deferring the approval to someone with more domain context

control-plane/subcommand/server-acl-init/anonymous_token_test.go
```diff
@@ -0,0 +1,74 @@
...
+	require.NoError(t, err)
+	readToken := resp.SecretID
+
+	readOnlyClient, errz := api.NewClient(&api.Config{
+		Address: consulHTTPAddr,
+		Token:   readToken,
+	})
+	require.NoError(t, errz)
+	// does not create/update anonymous token policy
+	erry := cmd.configureAnonymousPolicy(readOnlyClient)
```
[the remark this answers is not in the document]
thisisnotashwin: If the behavior we desire is that the anonymous token is not over-written, i think a more correct assertion for this behavior (if we can), is to check the creation timestamps on it and ensure that it is still the old timestamp even if we use a client that does have write access.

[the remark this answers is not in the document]
aahel: createTime doesn't change on token update.
asserting policy dosen't get updated as discussed on slack.

control-plane/subcommand/server-acl-init/anonymous_token_test.go
```diff
@@ -0,0 +1,68 @@
...
+	require.NoError(t, err)
+
+	err = cmd.configureAnonymousPolicy(consul)
+	require.NoError(t, err)
+
+	newPolicy, _, err := consul.ACL().PolicyReadByName(anonymousTokenPolicyName, nil)
+	require.NoError(t, err)
+
+	// assert policy is still same.
+	require.Equal(t, updatedPolicy, newPolicy)
```
thisisnotashwin, line 67: we should use different naming for the policies. updated and new are confusing. maybe `readOnlyPolicy` and `actualPolicy`

thisisnotashwin (approved):
This looks good. I dont think i need to review this again but I think we should update the names of the policy in the test.

control-plane/subcommand/server-acl-init/anonymous_token_test.go
aahel: done

aahel:
@david-yu  which versions should we backport this to?

david-yu:
Added backports, thank you.

Ganeshrockz (approved):

PR #2791, opened 2023-08-18
[NET-4498] Test locality propagation to services from k8s

zalimeni:
Verify that we propagate locality (region and zone) from standard k8s annotations to services registered by consul-k8s.

This will later be expanded to exercise multi-cluster locality-based failover.

Changes proposed in this PR:
- Add locality labels to k8s cluster nodes
- Assert that registered services receive corresponding locality config

How I've tested this PR: run locally and in CI

Checklist:
- [x] Tests added

### DIFF
--- a/acceptance/tests/sameness/sameness_test.go
+++ b/acceptance/tests/sameness/sameness_test.go
@@ -4,7 +4,7 @@
 package sameness
 
 import (
-	"context"
+	ctx "context"
 	"fmt"
 	"strconv"
 	"strings"
@@ -22,6 +22,7 @@ import (
 	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -50,6 +51,10 @@ const (
 	peerName3a = keyCluster03a
 
 	samenessGroupName = "group-01"
+
+	cluster01Region = "us-east-1"
+	cluster02Region = "us-west-1"
+	cluster03Region = "us-east-2"
 
 	retryTimeout = 5 * time.Minute
 )
@@ -116,8 +121,8 @@ func TestFailover_Connect(t *testing.T) {
 				|    |  +------------------+       |   |    |             |   |       +------------------+        |
 				|    |  Admin Partitions: Default  |   |    |             |   |                                   |
 				|    |  Name: cluster-01-a         |   |    |             |   |     Admin Partitions: Default     |
-				|    |                             |   |    |             |   |     Name: cluster-02-a            |
-				|    +-----------------------------+   |    |             |   |                                   |
+				|    |  Region: us-east-1          |   |    |             |   |     Name: cluster-02-a            |
+				|    +-----------------------------+   |    |             |   |     Region: us-west-1             |
 				|                                      |    |             |   +-----------------------------------+
 				|                            Failover 1|    |  Failover 3 |
 				|   +-------------------------------+  |    |             |   +-----------------------------------+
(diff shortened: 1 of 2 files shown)


acceptance/tests/sameness/sameness_test.go
```diff
@@ -665,16 +685,26 @@ func applyResources(t *testing.T, cfg *config.TestConfig, kustomizeDir string, o
...
+// serviceTargetCheck verifies that curling the `static-server` using the `static-client` responds with the expected
+// cluster name. Each static-server responds with a unique name so that we can verify failover occured as expected.
+func serviceTargetCheck(t *testing.T, server *cluster, expectedName string, curlAddress string) {
 	timer := &retry.Timer{Timeout: retryTimeout, Wait: 5 * time.Second}
 	var resp string
 	var err error
 	retry.RunWith(timer, t, func(r *retry.R) {
 		resp, err = k8s.RunKubectlAndGetOutputE(t, server.clientOpts, "exec", "-i",
-			staticClientDeployment, "-c", staticClientName, "--", "curl", curlAddress)
+			staticClientDeployment, "-c", staticClientName, "--", "curl", "-sS", curlAddress)
```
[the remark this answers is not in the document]
wilkermichael: `-s` I assume is silent, what does sS do?

Also would it be worth adding a comment?

acceptance/tests/sameness/sameness_test.go
```diff
@@ -665,16 +685,24 @@ func applyResources(t *testing.T, cfg *config.TestConfig, kustomizeDir string, o
...
+	require.NoError(t, err)
+	// Get the name of the (only) node from the Kind cluster.
+	node := nodeList.Items[0].Name
+	k8s.KubectlLabel(t, context.KubectlOptions(t), "node", node, corev1.LabelTopologyRegion, c.locality.Region)
+	k8s.KubectlLabel(t, context.KubectlOptions(t), "node", node, corev1.LabelTopologyZone, c.locality.Zone)
+}
+
+// serviceTargetCheck verifies that curling the `static-server` using the `static-client` responds with the expected
+// cluster name. Each static-server responds with a unique name so that we can verify failover occured as expected.
+func serviceTargetCheck(t *testing.T, server *cluster, expectedName string, curlAddress string) {
```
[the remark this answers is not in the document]
wilkermichael: Nice yeah I think that makes sense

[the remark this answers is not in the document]
wilkermichael: Also works with the "non-failover" initial case

acceptance/tests/sameness/sameness_test.go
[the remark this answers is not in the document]
zalimeni: Sure I can do that! `-S` is short for `--show-error`,

> When used with -s, --silent, it makes curl show an error message if it fails.

So basically don't silence errors, just progress bar. Thought TIL `--no-progress-meter` is also a flag! I'm not sure whether there's any subtle difference but probably will keep the shorthand 🤷🏻‍♂️

acceptance/tests/sameness/sameness_test.go
```diff
@@ -776,3 +808,39 @@ func getPeeringAcceptorSecret(t *testing.T, cfg *config.TestConfig, server *clus
...
+
+// localityForRegion returns the full api.Locality to use in tests for a given region string.
+func localityForRegion(r string) api.Locality {
+	return api.Locality{
+		Region: r,
+		Zone:   r + "a",
+	}
+}
+
+// checkLocalities checks the given cluster for client and server service instances matching the locality expected
```
wilkermichael, line 820: I was just thinking about this, but to avoid confusion we should probably refer to all the non server/clients as `static-server` and `static-client`. Everyone in the Consul testing world will understand immediately what's being talked about.

wilkermichael: Might make sense to change it in other places too, wherever it makes sense to use `static-server` and `static-client`

acceptance/tests/sameness/sameness_test.go
```diff
@@ -776,3 +808,39 @@ func getPeeringAcceptorSecret(t *testing.T, cfg *config.TestConfig, server *clus
...
+		for _, svc := range svcs {
+			cs := getCatalogService(t, c, svc, ns, c.partition)
+			assert.NotNil(t, cs.ServiceLocality, "service %s in %s did not have locality set", svc, c.name)
+			assert.Equal(t, c.locality, *cs.ServiceLocality, "locality for service %s in %s did not match expected", svc, c.name)
+		}
+	}
+}
+
+func getCatalogService(t *testing.T, c *cluster, svc, ns, partition string) *api.CatalogService {
+	resp, _, err := c.client.Catalog().Service(svc, "", &api.QueryOptions{Namespace: ns, Partition: partition})
```
wilkermichael, line 842: I wonder if we should add a retry here, just in case?

acceptance/tests/sameness/sameness_test.go
```diff
@@ -665,16 +685,28 @@ func applyResources(t *testing.T, cfg *config.TestConfig, kustomizeDir string, o
 	})
 }
 
-// serviceFailoverCheck verifies that the server failed over as expected by checking that curling the `static-server`
-// using the `static-client` responds with the expected cluster name. Each static-server responds with a uniquue
-// name so that we can verify failover occured as expected.
-func serviceFailoverCheck(t *testing.T, server *cluster, expectedName string, curlAddress string) {
+// setK8sNodeLocality labels the k8s node corresponding to the given cluster with standard labels indicating the
+// locality of that node. These are propagated by connect-inject to registered Consul services.
+func setK8sNodeLocality(t *testing.T, context environment.TestContext, c *cluster) {
```
wilkermichael, line 690: 👨‍🍳

wilkermichael (approved):
Great work Michael! 

You really followed the structure of what I was doing initially and added to it. Excited for the follow up PR :D

Just a couple of very mild comments

acceptance/tests/sameness/sameness_test.go
zalimeni: Good call! Agree that would improve clarity - will do

acceptance/tests/sameness/sameness_test.go
zalimeni: Could do - I'm a little wary of peppering things w/ retries we don't necessarily need, though I get the desire to keep tests as stable as possible. I didn't here bc we have a good number of other non-retried service fetches from the catalog (including in this test for DNS setup), so I figured it was likely to be as stable as those.

zalimeni: @wilkermichael if you're ok w/ it, I might hold off on this and if we see a general need to wrap catalog fetches in this and other tests, do it all at once... LMK if you have different thoughts

wilkermichael: Sound good

PR #2792, opened 2023-08-18
AKS 1.24 is deprecated, update to latest 1.25 patch

curtbushko:
Changes proposed in this PR:
- Updating kubernetes on AKS.
- Nightly acceptance are failing

How I've tested this PR:

Checklist:
- [ ] Tests added


zalimeni (approved):
LGTM

❓ How do we normally decide when to upgrade k8s versions in each provider? Looks like this will have EKS on 1.26 and AKS + GKE on 1.25

curtbushko:
@zalimeni For clouds, we try and keep to the lowest major version + latest minor version supported by the a platform. For Kind we try to be at the latest.

AKS dropped support for 1.24 at the beginning of the month. I saw the error and it's an easy fix.

PR #2812, opened 2023-08-22
Fix peer test flakes.

hashi-derek:
This commit fixes an issue where the peering tests would flake due to the fact that we were concurrently modifying a global map. It also adds in retry logic so that the consul servers have sufficient time to initialize before attempting to generate peering tokens.


curtbushko (approved):
LGTM!

PR #2822, opened 2023-08-22
add sameness testing performance enhancement

wilkermichael:
Changes proposed in this PR:
- Add parallelism in some of the "longer" portions of the testing code, this will dramatically decrease the runtime of the test

Before:
image

After:
image

How I've tested this PR:
- Locally and in pipeline
- Pipeline is no longer working, going to confirm that I ran the sameness tests locally:
- [x] Acceptance
- [x] Acceptance Tproxy
- [x] Acceptance CNI

👀 

Checklist:
- [x] Tests added

### DIFF
--- a/acceptance/tests/sameness/sameness_test.go
+++ b/acceptance/tests/sameness/sameness_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


curtbushko (approved):
This pretty awesome! Nothing is jumping out at me as being incorrect. 

I am worried about kube context mashing from inside client-go though...

What I would have done is I would have:
- Looped through all the installs
- Skipped `k8s.WaitForAllPodsToBeReady` in HelmCluster.Create().
- Checked up on the installs separately (maybe go func that).

As long as we don't see context problems then your ways is awesome and I approve!

PR #2832, opened 2023-08-23
refactor: make space for v2 controllers

DanStough:
Changes proposed in this PR:
- Refactor the inject-connect command to clean up the`Run` function
- Move controller/webhook initialization to separate files
- Incorporate V1/V2 switch to start controllers.

How I've tested this PR:
- CI/CD

- Eyes

Checklist:
- [ ] ~Tests added~
- [ ]~ [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry) ~

### DIFF
--- a/control-plane/consul/resource_client.go
+++ b/control-plane/consul/resource_client.go
@@ -1,29 +1,16 @@
 package consul
 
 import (
-	"context"
 	"fmt"
 
 	"github.com/hashicorp/consul-server-connection-manager/discovery"
 	"github.com/hashicorp/consul/proto-public/pbresource"
-	"github.com/hashicorp/go-hclog"
 )
 
 // NewResourceServiceClient creates a pbresource.ResourceServiceClient for creating V2 Consul resources.
-// It is initialized with a consul-server-connection-manager discovery config to continuously find Consul
+// It is initialized with a consul-server-connection-manager Watcher to continuously find Consul
 // server addresses.
-// The caller should make sure to Stop() the returned `watcher` (preferably with a `defer`) to clean up the gRPC
-// connection and the discovery client.
-// The caller can also set `config.ServerWatchDisabled=false` to prevent subscribing to Consul server address
-// changes, as is the case with single-shot operations.
-func NewResourceServiceClient(ctx context.Context, config discovery.Config, logger hclog.Logger, hack int) (pbresource.ResourceServiceClient, *discovery.Watcher, error) {
-
-	watcher, err := discovery.NewWatcher(ctx, config, logger.Named("consul-server-connection-manager"))
-	if err != nil {
-		return nil, nil, fmt.Errorf("unable to create Consul server watcher: %w", err)
-	}
-
-	go watcher.Run()
+func NewResourceServiceClient(watcher *discovery.Watcher) (pbresource.ResourceServiceClient, error) {
 
 	// We recycle the GRPC connection from the discovery client because it
 	// should have all the necessary dial options, including the resolver that
@@ -31,10 +18,9 @@ func NewResourceServiceClient(ctx context.Context, config discovery.Config, logg
 	// would need to be duplicated
 	state, err := watcher.State()
 	if err != nil {
-		watcher.Stop()
-		return nil, nil, fmt.Errorf("unable to get connection manager state: %w", err)
+		return nil, fmt.Errorf("unable to get connection manager state: %w", err)
 	}
 	resourceClient := pbresource.NewResourceServiceClient(state.GRPCConn)
 
-	return resourceClient, watcher, nil
+	return resourceClient, nil
 }
--- a/control-plane/consul/resource_client_test.go
+++ b/control-plane/consul/resource_client_test.go
@@ -36,12 +36,18 @@ func Test_NewResourceServiceClient(t *testing.T) {
 
 	opts := hclog.LoggerOptions{Name: "resource-service-client"}
 	logger := hclog.New(&opts)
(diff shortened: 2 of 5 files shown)


jmurret (approved):
👏 Nice.  Thank you for doing this as an initial PR so it is not mixed in with the feature stuff coming.  🙏

control-plane/subcommand/inject-connect/command.go
```diff
@@ -410,13 +349,13 @@ func (c *Command) Run(args []string) int {
...
-	watcher, err := discovery.NewWatcher(ctx, serverConnMgrCfg, hcLog)
+
+	resourceClient, watcher, err := consul.NewResourceServiceClient(ctx, serverConnMgrCfg, hcLog)
 	if err != nil {
 		c.UI.Error(fmt.Sprintf("unable to create Consul server watcher: %s", err))
 		return 1
 	}
 
-	go watcher.Run()
 	defer watcher.Stop()
```
zalimeni, line 359: Not certain I'm seeing the whole picture, but: it seems maybe ideal to keep the `Run()` and `Stop()` calls here adjacent to simplify reasoning about lifecycle, and not to duplicate them inside `NewResourceServiceClient`. Could we move the watcher init code back here, or to a helper that could be called independent of client construction, and have `NewResourceServiceClient` be only responsible for constructing the `resourceClient` from a passed `watcher`?

Also curious: is it necessary to create the `resourceClient` outside the v2 controller init? It seems like maybe we could push that concern down rather than handling part of it here, unless there's some future use we anticipate in the "common" code here.

zalimeni (commented):
I like the refactor! Generally LGTM, and my understanding is the only non-refactor functional changes are the flag introduction and new client creation. I have one question about that part.

FWIW, in `NewResourceServiceClient`, re:
```go
	// We recycle the GRPC connection from the discovery client because it
	// should have all the necessary dial options, including the resolver that
	// continuously updates Consul server addresses. Otherwise, a lot of code from consul-server-connection-manager
	// would need to be duplicated
```
I'm not sure `NewResourceServiceClient` is actually reusing a connection - my read of this is that `discovery.NewWatcher` is creating a brand new managed gRPC connection, and our use _seems_ to align w/ the `consul-server-connection-manager` docs. Maybe I'm missing something though.

The only thing that jumps out to me as potentially problematic, is if the v2 controller is run simultaneously w/ v1 in the future, and floods the gRPC connection w/ requests, we could hit the stream limit for the shared connection and block traffic for v1. We could look at creating an independent second watcher if so - but r/n that doesn't seem necessary since the two are either-or.

control-plane/subcommand/inject-connect/command.go
DanStough: Yes, I agree. I didn't anticipate sharing the watcher around in the previous PR. It makes to have it create it outside of the resourceClient and push creating the that to the V2 controller logic.

DanStough:
Not that the Acceptance Test 4 (vault, etc.) is currently failing on `main`. If a fix doesn't go in before this is reviewed I am going to merge it as-is.

control-plane/consul/resource_client_test.go
```diff
@@ -36,7 +36,15 @@ func Test_NewResourceServiceClient(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NotNil(t, watcher)
+
+	defer watcher.Stop()
+	go watcher.Run()
+
+	client, err := NewResourceServiceClient(watcher)
 	require.NoError(t, err)
 	require.NotNil(t, client)
 	require.NotNil(t, watcher)
```
zalimeni, line 50: ~ Might be a diff issue on my end, but looks like maybe there's a duplicate `defer watcher.Stop()` under this block of code (outside comment range)?

control-plane/consul/resource_client.go
```diff
@@ -32,9 +19,9 @@ func NewResourceServiceClient(ctx context.Context, config discovery.Config, logg
 	state, err := watcher.State()
 	if err != nil {
 		watcher.Stop()
```
zalimeni, line 21: ~ Do we need to `Stop()` here since the `watcher` is managed outside this file? Might be cleaner just to return the error and let the caller deal w/ shutdown

control-plane/subcommand/inject-connect/v2controllers.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/hashicorp/consul-server-connection-manager/discovery"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+func (c *Command) configureV2Controllers(ctx context.Context, mgr manager.Manager, watcher *discovery.Watcher) error {
+
+	//resourceClient, err := consul.NewResourceServiceClient(watcher)
+	//if err != nil {
+	//	return fmt.Errorf("unable to create Consul server watcher: %w", err)
```
zalimeni, line 17:
```suggestion
	//	return fmt.Errorf("unable to create Consul resource service client: %w", err)
```

zalimeni (approved):
Few minor non-blockers, LGTM!

control-plane/consul/resource_client_test.go
DanStough: Correct! Fixed.

control-plane/consul/resource_client.go
DanStough: You are right! I forgot to change this after the re-factor.

PR #2835, opened 2023-08-24
Gateway policy translation NET 4980

sarahalsmiller:
Changes proposed in this PR:
- Added gateway policy translations
-

How I've tested this PR:
- Currently testing

Checklist:
- [x] Tests added


control-plane/api-gateway/common/translation.go
```diff
@@ -141,6 +147,47 @@ func ToContainerPort(portNumber gwv1beta1.PortNumber, mapPrivilegedContainerPort
...
+
+	overridePolicy := &api.APIGatewayPolicy{
+		JWT: t.translateJWTRequirement(policy.Spec.Override.JWT),
+	}
+	return defaultPolicy, overridePolicy
+}
+
+func (t ResourceTranslator) translateJWTRequirement(crdRequirement *v1alpha1.GatewayJWTRequirement) *api.APIGatewayJWTRequirement {
+	apiRequirement := api.APIGatewayJWTRequirement{}
+	for _, provider := range crdRequirement.Providers {
```
jm96441n, line 167: I think we need a nil check the requirement since it's a pointer

control-plane/api-gateway/common/translation.go
```diff
@@ -141,6 +147,47 @@ func ToContainerPort(portNumber gwv1beta1.PortNumber, mapPrivilegedContainerPort
 	return int(portNumber) + int(mapPrivilegedContainerPorts)
 }
 
+func (t ResourceTranslator) translateGatewayPolicy(policy *v1alpha1.GatewayPolicy) (*api.APIGatewayPolicy, *api.APIGatewayPolicy) {
+	if policy == nil {
+		return nil, nil
+	}
+
+	defaultPolicy := &api.APIGatewayPolicy{
+		JWT: t.translateJWTRequirement(policy.Spec.Default.JWT),
```
jm96441n, line 156: I think we need a `nil` check on both `policy.Spec.Default` and `policy.Spec.Override` since both are optional

control-plane/api-gateway/common/translation.go
jm96441n, line 167: similar here, `Providers` is a pointer so we need a nil check on this

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -167,6 +167,13 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
+	// get all http routes referencing this gateway
```
jm96441n, line 170: is this comment correct?

control-plane/api-gateway/controllers/index.go
```diff
@@ -325,3 +331,15 @@ func filtersForHTTPRoute(o client.Object) []string {
 	}
 	return filters
 }
+
+func gatewayForGatewayPolicy(o client.Object) []string {
+	gatewayPolicy := o.(*v1alpha1.GatewayPolicy)
+
+	targetGateway := gatewayPolicy.Spec.TargetRef
```
jm96441n, line 338: is this the target gateway? the TargetRef should be a listener on a gateway right?

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -581,6 +593,26 @@ func (r *GatewayController) transformConsulHTTPRoute(ctx context.Context) func(e
 	}
 }
 
+// transformGatewayPolicy will return a list of all gateways that need to be reconcilled.
+func (r *GatewayController) transformGatewayPolicy(ctx context.Context) func(object client.Object) []reconcile.Request {
+	return func(o client.Object) []reconcile.Request {
+		gatewayPolicy := o.(*v1alpha1.GatewayPolicy)
+
+		//gateway policy to gateway is 1to1
```
jm96441n, line 601: this isn't necessarily the case, it's `gatewayPolicy:gatewayListener` is 1:1, so multiple policies can reference a gateway at this point, but each listener can only be referenced once

control-plane/api-gateway/common/resources.go
```diff
@@ -116,6 +116,7 @@ type ResourceMap struct {
 	httpRouteGateways     map[api.ResourceReference]*httpRoute
 	gatewayResources      map[api.ResourceReference]*resourceSet
 	externalFilters       map[corev1.ObjectReference]client.Object
+	gatewayPolicies       map[corev1.ObjectReference]*v1alpha1.GatewayPolicy
```
jm96441n, line 119: I think we might need to use `api.ResourceReference` as the key here so that we can reference the policy by the listener it applies to, what do you think?

control-plane/api-gateway/common/translation.go
jm96441n, line 156: I think we need a nil check on both policy.Spec.Default and policy.Spec.Override since both are optional

control-plane/api-gateway/common/translation.go
jm96441n, line 167: similar here, Providers is a pointer so we need a nil check on this

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -167,6 +167,13 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
+	// get all http routes referencing this gateway
+	policy, err := r.getRelatedGatewayPolicy(ctx, req.NamespacedName, resources)
```
jm96441n, line 171: is this comment correct?

control-plane/api-gateway/controllers/gateway_controller.go
jm96441n, line 601: this isn't necessarily the case, it's gatewayPolicy:gatewayListener is 1:1, so multiple policies can reference a gateway at this point, but each listener can only be referenced once

control-plane/api-gateway/controllers/index.go
jm96441n, line 338: ~~is this the target gateway? the TargetRef should be a listener on a gateway right?~~ nevermind, the target ref will be a gateway with the listener as part of the section name

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -877,6 +909,28 @@ func (c *GatewayController) filterFiltersForExternalRefs(ctx context.Context, ro
...
+	var list v1alpha1.GatewayPolicyList
+
+	if err := c.Client.List(ctx, &list, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(Gatewaypolicy_GatewayIndex, gateway.String()),
+	}); err != nil {
+		return nil, err
+	}
+
+	if len(list.Items) > 0 {
+		//TODO should we error here or assume that this will happen in validation webhook
```
jm96441n, line 922: this should be fine if we grab all the gateways referenced and then grab all the policies that reference the gateway and then loop through to find the one referencing the specific listener we want

jm96441n (changes requested):
few comments, but overall looks good!

control-plane/api-gateway/common/resources.go
sarahalsmiller: I'll give it a try. I think I initially thought it was 1to1 and wasn't accounting for the listeners.

control-plane/api-gateway/binding/binder.go
```diff
@@ -61,6 +61,8 @@ type BinderConfig struct {
 	// Resources is a map containing all service targets to verify
 	// against the routing backends.
 	Resources *common.ResourceMap
+
+	Policies []v1alpha1.GatewayPolicy
```
t-eckert, line 65: Can we add a comment here?

control-plane/api-gateway/common/resources.go
```diff
@@ -401,6 +402,38 @@ func (s *ResourceMap) ExternalFilterExists(filterRef gwv1beta1.LocalObjectRefere
 	return ok
 }
 
+func (s *ResourceMap) AddGatewayPolicy(gatewayPolicy *v1alpha1.GatewayPolicy) *v1alpha1.GatewayPolicy {
+	sectionName := ""
+	if gatewayPolicy.Spec.TargetRef.SectionName != nil {
```
t-eckert, line 407: Can any of the parent values to `SectionName` be `nil`? If so, we should `nil` check them to avoid a deref error.

control-plane/api-gateway/common/resources.go
```diff
@@ -401,6 +402,38 @@ func (s *ResourceMap) ExternalFilterExists(filterRef gwv1beta1.LocalObjectRefere
...
+	if s.gatewayPolicies == nil {
+		s.gatewayPolicies = make(map[api.ResourceReference]*v1alpha1.GatewayPolicy)
+	}
+
+	s.gatewayPolicies[key] = gatewayPolicy
+
+	return s.gatewayPolicies[key]
+}
+
+func (s *ResourceMap) GetPolicyForGatewayListener(gateway gwv1beta1.Gateway, gatewayListener gwv1beta1.Listener) *v1alpha1.GatewayPolicy {
```
t-eckert, line 426: Retrieval from a map like this will return an optional boolean. So you can improve the API for this function by returning `(*v1alpha1.GatewayPolicy, bool)`. That way, when you use it, you can have a quick check if the value was found in the map:

```go
if policy, exists := s.GetPolicyForGatewayListener(gateway, gatewayListener); exists {
    [... insert great code here ...]
}
```

control-plane/api-gateway/common/translation.go
```diff
@@ -113,6 +113,10 @@ func (t ResourceTranslator) toAPIGatewayListener(gateway gwv1beta1.Gateway, list
 		}
 	}
 
+	//grab policy if it exists
```
t-eckert, line 116: I think the linter generally prefers comments that are sentence cased and end in a period.

control-plane/api-gateway/common/translation.go
```diff
@@ -141,6 +147,60 @@ func ToContainerPort(portNumber gwv1beta1.PortNumber, mapPrivilegedContainerPort
...
+	if crdProvider == nil {
+		return nil
+	}
+
+	apiProvider := api.APIGatewayJWTProvider{
+		Name: crdProvider.Name,
+	}
+	for _, verifyClaim := range crdProvider.VerifyClaims {
+		apiProvider.VerifyClaims = append(apiProvider.VerifyClaims, t.translateVerifyClaims(verifyClaim))
+	}
```
t-eckert, line 189: You can replace this with `ConvertSliceFunc`.

control-plane/api-gateway/common/translation_test.go
```diff
@@ -1643,3 +1643,149 @@ func TestResourceTranslator_translateHTTPFilters(t1 *testing.T) {
...
+	}
+	tests := []struct {
+		name     string
+		args     args
+		policies []v1alpha1.GatewayPolicy
+		want     api.APIGatewayListener
+		want1    bool
+	}{
+		{
+			name: "listener with jwt auth",
```
jm96441n, line 1666: do we need the table test here if we're only running one case?

control-plane/api-gateway/controllers/index.go
```diff
@@ -325,3 +331,15 @@ func filtersForHTTPRoute(o client.Object) []string {
...
+	//gateway policy is 1to1
```
jm96441n, line 339: this comment looks like it's missing the rest

control-plane/api-gateway/controllers/index.go
```diff
@@ -325,3 +331,15 @@ func filtersForHTTPRoute(o client.Object) []string {
...
+	if targetGateway.Group == v1alpha1.ConsulHashicorpGroup && targetGateway.Kind == common.KindGateway {
+		return []string{common.IndexedNamespacedNameWithDefault(targetGateway.Name, &targetGateway.Namespace, gatewayPolicy.Namespace).String()}
+	}
```
jm96441n, line 342: looking at this more, I don't think this works as expect because the `IndexedNamespacedNameWithDefault` expects the second argument to be nil-able but `targetGateway.Namespace` will never be nil so it will always use that value no matter what is in it (even an empty string), I think something like the following would work
```suggestion
func gatewayForGatewayPolicy(o client.Object) []string {
	gatewayPolicy := o.(*v1alpha1.GatewayPolicy)

	targetGateway := gatewayPolicy.Spec.TargetRef
	// gateway policy is 1to1
	if targetGateway.Group == "gateway.networking.k8s.io/v1beta1" && targetGateway.Kind == common.KindGateway {
		policyNamespace := gatewayPolicy.Namespace
		if policyNamespace == "" {
			policyNamespace = "default"
		}
		targetNS := targetGateway.Namespace
		if targetNS == "" {
			targetNS = policyNamespace
		}
		return []string{types.NamespacedName{Name: targetGateway.Name, Namespace: targetNS}.String()}
	}

	return []string{}
}
```

sarahalsmiller: Truly after looking at this I think I was just trying to avoid having to build out a namespacedname object, since I don't think there ever actually is a reason for it to use the policies namespace here.

sarahalsmiller: I'll just build it out so things are clearer

control-plane/api-gateway/common/resources.go
sarahalsmiller: I don't believe they can be since Spec and TargetRef are both non nillable.

control-plane/api-gateway/controllers/index.go
jm96441n: >  I don't think there ever actually is a reason for it to use the policies namespace here.
I think we would use the policies namespace if it's left out of the target ref since the usual behavior there is to assume it's the same as the parent resource

sarahalsmiller: I see. Fair enough, I'll add that in.

jm96441n (approved):
LGTM! Awesome work!

PR #2836, opened 2023-08-24
Translate HTTPAuthFilter onto HTTPRoute

t-eckert:
> [!NOTE]  
> This PR merges into #2835 

Changes proposed in this PR:
- Translate JWT filter CRDs to config entry format on HTTP Route.
- Re-use JWT configuration from API Gateways.
- Split out filtering into helper functions.

How I've tested this PR:
- Added unit test

- Running unit tests

Checklist:
- [x] Tests added

I don't think this needs a CHANGELOG entry because it will get swept into the broader changelog for the feature.


control-plane/api-gateway/common/translation.go
```diff
@@ -168,10 +168,11 @@ func (t ResourceTranslator) ToHTTPRoute(route gwv1beta1.HTTPRoute, resources *Re
 }
 
 func (t ResourceTranslator) translateHTTPRouteRule(route gwv1beta1.HTTPRoute, rule gwv1beta1.HTTPRouteRule, resources *ResourceMap) (api.HTTPRouteRule, bool) {
-	services := ConvertSliceFuncIf(rule.BackendRefs, func(ref gwv1beta1.HTTPBackendRef) (api.HTTPService, bool) {
-
-		return t.translateHTTPBackendRef(route, ref, resources)
-	})
+	services := ConvertSliceFuncIf(
```
t-eckert, line 171: I found this formatting made the code easier to read YMMV.

control-plane/api-gateway/common/translation.go
```diff
@@ -311,38 +317,68 @@ func (t ResourceTranslator) translateHTTPFilters(filters []gwv1beta1.HTTPRouteFi
...
+}
+
+func makeJWTFilter(routeJWTFilter *v1alpha1.RouteAuthFilter) *api.JWTFilter {
+	if routeJWTFilter.Spec.JWT == nil || routeJWTFilter.Spec.JWT.Providers == nil {
+		return nil
+	}
+
+	// For each provider, translate the claims to the Consul config entry format.
+	providers := make([]*api.APIGatewayJWTProvider, 0, len(routeJWTFilter.Spec.JWT.Providers))
+	for _, provider := range routeJWTFilter.Spec.JWT.Providers {
```
jm96441n, line 364: we might be able to use https://github.com/hashicorp/consul-k8s/pull/2835/files#diff-4eabdf1a10d46efb8f48e81965e976da7abe91a8ebb6fc643871c8e2592e3612R173 as to handle the translation since they're both translating the same types

jm96441n (changes requested):
minor comment so we don't have two implementations of how to translate a struct. otherwise good work!

control-plane/api-gateway/common/translation.go
t-eckert: This is a great comment. I'll rebase on @sarahalsmiller's PR to grab it.

charts/consul/templates/crd-gateways.yaml
```diff
@@ -7,14 +7,15 @@ kind: CustomResourceDefinition
...
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/1538
-    gateway.networking.k8s.io/bundle-version: v0.6.2
-    gateway.networking.k8s.io/channel: experimental
   labels:
     app: {{ template "consul.name" . }}
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
     component: crd
+    gateway.networking.k8s.io/bundle-version: v0.6.2
```
sarahalsmiller, line 16: Did you mean to touch all of these templates?

t-eckert: I had to when I regen'd the CRDs

t-eckert: I think the person who did this before didn't use the `0.8.0` version that is standard for the repo.

control-plane/api-gateway/common/translation_test.go
```diff
@@ -1267,7 +1272,21 @@ func TestTranslator_ToHTTPRoute(t *testing.T) {
 										ExtensionRef: &gwv1beta1.LocalObjectReference{
 											Name:  "test",
 											Kind:  v1alpha1.RouteRetryFilterKind,
-											Group: "consul.hashicorp.com/v1alpha1",
+											Group: ConsulHashicorpGroup,
```
andrewstucki, line 1275: Since `github.com/hashicorp/consul-k8s/control-plane/api/v1alpha1` is already imported, I think you can just do something like this to avoid redefining the constant everywhere:

```suggestion
											Group: v1alpha1.GroupVersion.String(),
```

Same with the other places where this is used.

control-plane/config/crd/bases/consul.hashicorp.com_gatewayclassconfigs.yaml
```diff
@@ -75,13 +75,22 @@ spec:
                     minimum: 1
                     type: integer
                 type: object
+              mapPrivilegedContainerPorts:
+                description: The value to add to privileged ports ( ports < 1024)
+                  for gateway containers
+                format: int32
+                type: integer
```
andrewstucki, line 82: uh-oh, did the generators not get re-run? these fields seems to be from the addition of some openshift specific stuff, which isn't in-scope in this PR.

control-plane/api-gateway/common/translation.go
```diff
@@ -358,38 +398,52 @@ func (t ResourceTranslator) translateHTTPFilters(filters []gwv1beta1.HTTPRouteFi
...
+		}
+
+		providers = append(providers, &api.APIGatewayJWTProvider{
+			Name:         provider.Name,
+			VerifyClaims: claims,
+		})
+	}
+
+	return &api.JWTFilter{
+		Providers: providers,
```
andrewstucki, line 446: To keep with some of the way a lot of this file is done functionally (in part to break down the error handling/internal nil pointer checking) and since the helpers already exist, wondering if it'd be a bit more terse/standardized to do something like:

```go
func makeJWTFilter(filter *v1alpha1.RouteAuthFilter) *api.JWTFilter {
	if filter.Spec.JWT == nil {
		return nil
	}

	return &api.JWTFilter{
		Providers: ConvertSliceFunc(filter.Spec.JWT.Providers, t.translateJWTProvider),
	})
}
```

control-plane/api-gateway/common/translation.go
```diff
@@ -147,6 +147,38 @@ func ToContainerPort(portNumber gwv1beta1.PortNumber, mapPrivilegedContainerPort
...
+}
+
+func (t ResourceTranslator) translateRouteTimeoutFilter(routeTimeoutFilter *v1alpha1.RouteTimeoutFilter) *api.TimeoutFilter {
+	return &api.TimeoutFilter{
+		RequestTimeout: routeTimeoutFilter.Spec.RequestTimeout,
+		IdleTimeout:    routeTimeoutFilter.Spec.IdleTimeout,
+	}
+}
+
+func (t ResourceTranslator) translateRouteJWTFilter(routeJWTFilter *v1alpha1.RouteAuthFilter) *api.JWTFilter {
```
andrewstucki, line 166: Is there a diff between this and the `make*` function below? I'm assuming they should probably just be the same function? Also, to avoid the additional `append` boilerplate, could probably just use the generic `ConvertSliceFunc` mentioned previously here too.

control-plane/config/crd/bases/consul.hashicorp.com_gatewayclassconfigs.yaml
t-eckert: Potentially. I just ran them for my CRD changes.

t-eckert: Looking at what changed, I think the previous compilation used the wrong version of KubeBuilder. The version we should be using is `0.8.0`. I think the previous change was made with `0.11`

control-plane/api-gateway/common/translation.go
t-eckert: Nope, that was my initial function. When I saw how Sarah was writing her's as `translate` instead of `make`, I switched to that wording here and forgot to remove the `make` one.

control-plane/config/crd/bases/consul.hashicorp.com_gatewayclassconfigs.yaml
jm96441n: can we only include the CRD changes that are necessary for this PR and clean up the rest as part of a PR against `main`?

t-eckert: Yes, I will manually fix these.

jm96441n (approved):
overall LGTM!

PR #2842, opened 2023-08-25
fix: lifecycle enabled iptables mismatch

DanStough:
Changes proposed in this PR:
- Fixed the indexing between the iptables rules and the overwriteProbes to reference the same number regardless of when dataplane is injected.

How I've tested this PR: 
- I wrote a unit test that failed w/o this fix, then added the fix (TDD)
- consul-k8s CLI test is now passing because it explicitly enables tproxy

Checklist:
- [X] Tests added
- [ ]~ [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry) ~

### DIFF
--- a/acceptance/framework/connhelper/connect_helper.go
+++ b/acceptance/framework/connhelper/connect_helper.go
@@ -11,16 +11,17 @@ import (
 	"time"
 
 	terratestK8s "github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/hashicorp/consul-k8s/acceptance/framework/config"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/consul"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/environment"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
 	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
-	"github.com/hashicorp/consul/api"
-	"github.com/hashicorp/consul/sdk/testutil/retry"
-	"github.com/stretchr/testify/require"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
--- a/control-plane/connect-inject/webhook/mesh_webhook.go
+++ b/control-plane/connect-inject/webhook/mesh_webhook.go
@@ -15,13 +15,6 @@ import (
 
 	mapset "github.com/deckarep/golang-set"
 	"github.com/go-logr/logr"
-	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/common"
-	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
-	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/lifecycle"
-	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/metrics"
-	"github.com/hashicorp/consul-k8s/control-plane/consul"
-	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
-	"github.com/hashicorp/consul-k8s/control-plane/version"
 	"gomodules.xyz/jsonpatch/v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -30,6 +23,14 @@ import (
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/common"
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/lifecycle"
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/metrics"
+	"github.com/hashicorp/consul-k8s/control-plane/consul"
+	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
+	"github.com/hashicorp/consul-k8s/control-plane/version"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


acceptance/framework/connhelper/connect_helper.go
```diff
@@ -395,6 +396,7 @@ func (c *ConnectHelper) helmValues() map[string]string {
 		"global.acls.manageSystemACLs": strconv.FormatBool(c.Secure),
 		"dns.enabled":                  "true",
 		"dns.enableRedirection":        "true",
+		"global.imageK8S":              "ttl.sh/dans/consul-k8s-dev-iptables:24h",
```
DanStough, line 399: will fix this

control-plane/connect-inject/webhook/redirect_traffic.go
```diff
@@ -62,20 +63,24 @@ func (w *MeshWebhook) iptablesConfigJSON(pod corev1.Pod, ns corev1.Namespace) (s
 	}
 
 	if overwriteProbes {
-		for i, container := range pod.Spec.Containers {
-			// skip the "envoy-sidecar" container from having its probes overridden
+		// We don't use the loop index because this need to line up w.overwriteProbes(),
```
wilkermichael, line 66:
```suggestion
		// We don't use the loop index because this needs to line up w.overwriteProbes(),
```

control-plane/connect-inject/webhook/mesh_webhook_test.go
```diff
@@ -1129,6 +1132,231 @@ func TestHandlerHandle(t *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tt := range cases {
+		t.Run(tt.Name, func(t *testing.T) {
+			tt.Webhook.ConsulConfig = &consul.Config{HTTPPort: 8500}
+			ctx := context.Background()
+			resp := tt.Webhook.Handle(ctx, tt.Req)
+			if (tt.Err == "") != resp.Allowed {
```
wilkermichael, line 1291: nit: This took me a bit longer to wrap my head around, would it be clearer to write:
`if tt.Err != "" && resp.Allowed`

control-plane/connect-inject/webhook/mesh_webhook_test.go
```diff
@@ -1129,6 +1132,231 @@ func TestHandlerHandle(t *testing.T) {
...
+						valueMap, ok := value[0].(map[string]any)
+						require.True(t, ok)
+
+						envs, ok := valueMap["env"].([]any)
+						require.True(t, ok)
+
+						redirectEnv, ok := envs[8].(map[string]any)
+						require.True(t, ok)
+
+						require.Equal(t, redirectEnv["name"].(string), "CONSUL_REDIRECT_TRAFFIC_CONFIG")
```
wilkermichael, line 1317: Not for you to do, but I'm surprised "CONSUL_REDIRECT_TRAFFIC_CONFIG" isn't a constant somewhere

control-plane/connect-inject/webhook/mesh_webhook_test.go
```diff
@@ -1129,6 +1132,231 @@ func TestHandlerHandle(t *testing.T) {
...
+
+								portNum, ok := port.(float64)
+								require.True(t, ok)
+
+								overwritePorts = append(overwritePorts, strconv.Itoa(int(portNum)))
+							}
+						}
+					}
+
+					actual[i].Value = nil
```
wilkermichael, line 1351: I'm not 100% clear on this step, would it be useful to add a comment explaining the need to set the Value to nil?

control-plane/connect-inject/webhook/mesh_webhook_test.go
```diff
@@ -1129,6 +1132,231 @@ func TestHandlerHandle(t *testing.T) {
...
+				require.Contains(t, resp.Result.Message, tt.Err)
+				return
+			}
+
+			var iptablesCfg iptables.Config
+			var overwritePorts []string
+			actual := resp.Patches
+			if len(actual) > 0 {
+				for i := range actual {
+					if actual[i].Path == "/spec/initContainers" {
```
wilkermichael, line 1304: For the purposes of this test, do we care about the checks here? It seems like we only need to accumulate the overwritePorts.

If you found it useful in debugging stuff though, then we can leave it

control-plane/connect-inject/webhook/mesh_webhook_test.go
```diff
@@ -1129,6 +1132,231 @@ func TestHandlerHandle(t *testing.T) {
...
+						require.NoError(t, err)
+					}
+
+					if strings.Contains(actual[i].Path, "/spec/containers/1") {
+						valueMap, ok := actual[i].Value.(map[string]any)
+						require.True(t, ok)
+
+						for k, v := range valueMap {
+							if strings.Contains(k, "Probe") {
+								probe, ok := v.(map[string]any)
```
wilkermichael, line 1331: It took me a second to realise that we weren't really validating any of these values, but just building up the map to get to building our list of `overwritePorts` maybe just a few comments to explain what's going on.

wilkermichael (approved):
Great work Dan, thank you so much for doing this.

A few comments, but nothing blocking! :D

control-plane/connect-inject/webhook/mesh_webhook_test.go
DanStough: Yeah, this is copy-paste from another test. TBH I kind of like it w/ the parens to help with the order of operations.

control-plane/connect-inject/webhook/mesh_webhook_test.go
DanStough: Yeah, I can remove them. Was using them for debug.

control-plane/connect-inject/webhook/mesh_webhook_test.go
DanStough: Hmm, good point. I will check.

control-plane/connect-inject/webhook/mesh_webhook_test.go
DanStough: Will add some comments!

control-plane/connect-inject/webhook/mesh_webhook_test.go
DanStough: Yeah, it's another copy-paste artifact. I also found this hard to understand. We don't actually validate the output is correct, just that the right keys change. I will add a comment though.

control-plane/connect-inject/webhook/mesh_webhook_test.go
DanStough: Nope, not in `constants`. I'm going to leave that for a future PR.

control-plane/connect-inject/webhook/mesh_webhook.go
```diff
@@ -533,20 +534,24 @@ func (w *MeshWebhook) overwriteProbes(ns corev1.Namespace, pod *corev1.Pod) erro
 	}
 
 	if tproxyEnabled && overwriteProbes {
-		for i, container := range pod.Spec.Containers {
+		// We don't use the loop index because this needs to line up w.withiptablesConfigJSON,
+		// which is performed before the sidecar is injected.
```
t-eckert, line 538: Great comment. I think it would be easy to get confused on this otherwise.

t-eckert (approved):
Thank you so much for catching and fixing this.

wilkermichael:
@DanStough can you also just verify if this needs to be backported?

PR #2843, opened 2023-08-25
Fix issue where CLI install test was running Tproxy manually

t-eckert:
Changes proposed in this PR:
- Removes the case where transparent proxy is enabled in the test -- it should be set outside of the test with the flag

How I've tested this PR:
- Acceptance

- Acceptance

Checklist:
- [ ] Tests added


wilkermichael (approved):
Thanks for doing this!

Can we backport this change back as well to `1.0.x`, `1.1.x` and `1.2.x`

DanStough (approved):
LGTM

PR #2868, opened 2023-08-30
feat: add v2 pod controller w/ workload lifecycle

DanStough:
This PR is part of a series of PRs that add functionality to the V2 Pod Controller. For the MVP task to be complete, there will be one additional PR to fill in the lifecycle of HealthStatus and ProxyConfiguration resources. 

`TODO(dans)` comments are needed for the MVP tasks. Other `TODOs` are captured in future stories.

Changes proposed in this PR:
- Add the V2 Pod Controller
- Add enough logic in the controller to manage Consul Workloads
- pod controller test framework, borrowed from the endpoints controller.

How I expect reviewers to test this PR: 👓 

Checklist:
- [X] Tests added

### DIFF
--- a/control-plane/connect-inject/constants/constants.go
+++ b/control-plane/connect-inject/constants/constants.go
@@ -7,6 +7,12 @@ const (
 	// ConsulCAFile is the location of the Consul CA file inside the injected pod.
 	ConsulCAFile = "/consul/connect-inject/consul-ca.pem"
 
+	// DefaultConsulNS is the default Consul namespace name.
+	DefaultConsulNS = "default"
+
+	// DefaultConsulPartition is the default Consul partition name.
+	DefaultConsulPartition = "default"
+
 	// ProxyDefaultInboundPort is the default inbound port for the proxy.
 	ProxyDefaultInboundPort = 20000
 
--- a/control-plane/connect-inject/controllers/endpoints/consul_client_health_checks.go
+++ b/control-plane/connect-inject/controllers/endpoints/consul_client_health_checks.go
@@ -6,12 +6,13 @@ package endpoints
 import (
 	"fmt"
 
-	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
-	"github.com/hashicorp/consul-k8s/control-plane/consul"
 	"github.com/hashicorp/consul-server-connection-manager/discovery"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-version"
 	corev1 "k8s.io/api/core/v1"
+
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	"github.com/hashicorp/consul-k8s/control-plane/consul"
 )
 
 const minSupportedConsulDataplaneVersion = "v1.0.0-beta1"
@@ -19,7 +20,12 @@ const minSupportedConsulDataplaneVersion = "v1.0.0-beta1"
 // isConsulDataplaneSupported returns true if the consul-k8s version on the pod supports
 // consul-dataplane architecture of Consul.
 func isConsulDataplaneSupported(pod corev1.Pod) bool {
-	if anno, ok := pod.Annotations[constants.AnnotationConsulK8sVersion]; ok {
+	anno, ok := pod.Annotations[constants.LegacyAnnotationConsulK8sVersion]
+	if !ok {
+		anno, ok = pod.Annotations[constants.AnnotationConsulK8sVersion]
+	}
+
+	if ok {
 		consulK8sVersion, err := version.NewVersion(anno)
 		if err != nil {
 			// Only consul-k8s v1.0.0+ (including pre-release versions) have the version annotation. So it would be
--- a/control-plane/connect-inject/controllers/pod/pod_controller.go
+++ b/control-plane/connect-inject/controllers/pod/pod_controller.go
@@ -1,0 +1,337 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package pod
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	mapset "github.com/deckarep/golang-set"
+	"github.com/go-logr/logr"
+	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v1alpha1"
+	"github.com/hashicorp/consul/proto-public/pbresource"
+	"github.com/hashicorp/go-multierror"
+	"google.golang.org/protobuf/types/known/anypb"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/common"
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/metrics"
+	"github.com/hashicorp/consul-k8s/control-plane/consul"
+	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
+)
+
+const (
+	metaKeyManagedBy    = "managed-by"
+	tokenMetaPodNameKey = "pod"
+)
+
+type Controller struct {
+	client.Client
+	// ConsulClientConfig is the config for the Consul API client.
+	ConsulClientConfig *consul.Config
+	// ConsulServerConnMgr is the watcher for the Consul server addresses.
+	ConsulServerConnMgr consul.ServerConnectionManager
+	// Only pods in the AllowK8sNamespacesSet are reconciled.
+	AllowK8sNamespacesSet mapset.Set
+	// Pods in the DenyK8sNamespacesSet are ignored.
+	DenyK8sNamespacesSet mapset.Set
+	// EnableConsulPartitions indicates that a user is running Consul Enterprise
+	EnableConsulPartitions bool
+	// ConsulPartition is the Consul Partition to which this controller belongs
+	ConsulPartition string
+	// EnableConsulNamespaces indicates that a user is running Consul Enterprise
+	EnableConsulNamespaces bool
+	// ConsulDestinationNamespace is the name of the Consul namespace to create
+	// all config entries in. If EnableNSMirroring is true this is ignored.
+	ConsulDestinationNamespace string
+	// EnableNSMirroring causes Consul namespaces to be created to match the
+	// k8s namespace of any config entry custom resource. Config entries will
+	// be created in the matching Consul namespace.
+	EnableNSMirroring bool
+	// NSMirroringPrefix is an optional prefix that can be added to the Consul
+	// namespaces created while mirroring. For example, if it is set to "k8s-",
+	// then the k8s `default` namespace will be mirrored in Consul's
+	// `k8s-default` namespace.
+	NSMirroringPrefix string
+
+	// TODO: EnableWANFederation
+
+	// AuthMethod is the name of the Kubernetes Auth Method that
+	// was used to login with Consul. The Endpoints controller
+	// will delete any tokens associated with this auth method
+	// whenever service instances are deregistered.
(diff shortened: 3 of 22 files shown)


control-plane/consul/resource_client.go
```diff
@@ -6,14 +6,13 @@ package consul
...
 
-	"github.com/hashicorp/consul-server-connection-manager/discovery"
 	"github.com/hashicorp/consul/proto-public/pbresource"
 )
 
 // NewResourceServiceClient creates a pbresource.ResourceServiceClient for creating V2 Consul resources.
 // It is initialized with a consul-server-connection-manager Watcher to continuously find Consul
 // server addresses.
-func NewResourceServiceClient(watcher *discovery.Watcher) (pbresource.ResourceServiceClient, error) {
+func NewResourceServiceClient(watcher ServerConnectionManager) (pbresource.ResourceServiceClient, error) {
```
DanStough, line 15: This interface allows for using the mock during tests.

control-plane/connect-inject/webhook/mesh_webhook.go
```diff
@@ -620,6 +620,7 @@ func (w *MeshWebhook) defaultAnnotations(pod *corev1.Pod, podJson string) error
 		}
 	}
 	pod.Annotations[constants.AnnotationOriginalPod] = podJson
+	pod.Annotations[constants.LegacyAnnotationConsulK8sVersion] = version.GetHumanVersion()
```
DanStough, line 623: I added the new annotation and kept the old one for backward compatibility.

control-plane/connect-inject/constants/constants.go
```diff
@@ -7,6 +7,12 @@ const (
 	// ConsulCAFile is the location of the Consul CA file inside the injected pod.
 	ConsulCAFile = "/consul/connect-inject/consul-ca.pem"
 
+	// DefaultConsulNS is the default Consul namespace name
+	DefaultConsulNS = "default"
+
+	// DefaultConsulPartition is the default Consul partition name
+	DefaultConsulPartition = "default"
```
DanStough, line 14: resources currently need these set all the time

control-plane/connect-inject/controllers/endpoints/consul_client_health_checks.go
```diff
@@ -6,20 +6,26 @@ package endpoints
...
 
 // isConsulDataplaneSupported returns true if the consul-k8s version on the pod supports
 // consul-dataplane architecture of Consul.
 func isConsulDataplaneSupported(pod corev1.Pod) bool {
-	if anno, ok := pod.Annotations[constants.AnnotationConsulK8sVersion]; ok {
+	anno, ok := pod.Annotations[constants.LegacyAnnotationConsulK8sVersion]
+	if !ok {
+		anno, ok = pod.Annotations[constants.AnnotationConsulK8sVersion]
+	}
+
```
DanStough, line 27: Check for new an legacy annotations.

control-plane/connect-inject/controllers/pod/pod.go
```diff
@@ -0,0 +1,361 @@
...
+		return true
+	}
+
+	// Ignores if not in allow list or allow list is not *.
+	if !allowSet.Contains("*") && !allowSet.Contains(namespace) {
+		return true
+	}
+
+	return false
+}
```
zalimeni, line 272: ~ Move this and the v1 endpoints version to common? Would be nice to consolidate tests for it too

zalimeni (commented):
This is looking great to me, and I really appreciate having your test setup as an example while I'm trying to write my own.

Some non-blocking comments about potential shared helpers, but overall what you have so far looks reasonable to me. Will look forward to seeing any other feedback around functional req's.

control-plane/connect-inject/controllers/pod/pod.go
```diff
@@ -0,0 +1,361 @@
...
+		Identity: pod.Spec.ServiceAccountName,
+		Locality: locality,
+		NodeName: common.ConsulNodeNameFromK8sNode(pod.Spec.NodeName),
+		Ports:    workloadPorts,
+	}
+
+	proto, err := anypb.New(workload)
+	if err != nil {
+		return fmt.Errorf("could not serialize workload: %w", err)
+	}
```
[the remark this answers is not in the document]
DanStough: IMHO, I'd rather have a log than a panic, otherwise, I agree the helper makes sense. I'll add a TODO in my PR to come back and refactor this when you merge it.

control-plane/connect-inject/controllers/pod/pod.go
```diff
@@ -0,0 +1,361 @@
...
+)
+
+type Controller struct {
+	client.Client
+	// ConsulClientConfig is the config for the Consul API client.
+	ConsulClientConfig *consul.Config
+	// ConsulServerConnMgr is the watcher for the Consul server addresses.
+	ConsulServerConnMgr consul.ServerConnectionManager
+	// ConsulResourceServiceClient is a gRPC client for the resource service.
+	ConsulResourceServiceClient pbresource.ResourceServiceClient
```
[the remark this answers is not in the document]
DanStough: You're right, we don't need both. I'll refactor to only pass in the `ConsulServerConnMgr`.

control-plane/connect-inject/controllers/pod/pod.go
```diff
@@ -0,0 +1,361 @@
...
+
+	return r.ConsulDestinationNamespace
+}
+
+func (r *Controller) getPartition() string {
+	if !r.EnableConsulPartitions || r.Partition == "" {
+		return constants.DefaultConsulPartition
+	}
+	return r.Partition
+}
```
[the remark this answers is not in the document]
DanStough: This is using all the Controller structs fields from the receiver, so I think it will be less verbose to keep it as-is than a free function.

control-plane/connect-inject/controllers/pod/pod.go
```diff
@@ -0,0 +1,361 @@
...
+
+	if region == "" {
+		return nil
+	}
+
+	return &pbcatalog.Locality{
+		Region: region,
+		Zone:   zone,
+	}
+}
```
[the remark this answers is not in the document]
DanStough: This differs slightly from the one in endpoints since it returns a `pbcatalog.Locality`, and I'm not sure any other controllers need it, so I'll keep it as-is for now.

control-plane/connect-inject/controllers/pod/pod.go
```diff
@@ -0,0 +1,361 @@
...
+	if !r.EnableConsulNamespaces {
+		return constants.DefaultConsulNS
+	}
+
+	// Mirroring takes precedence.
+	if r.EnableNSMirroring {
+		return fmt.Sprintf("%s%s", r.NSMirroringPrefix, kubeNamespace)
+	}
+
+	return r.ConsulDestinationNamespace
```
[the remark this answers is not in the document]
DanStough: From a slack thread, it sounds like the behavior _should_ change in the future. Considering I don't know the exact timing, I'll opt for the former change.

control-plane/connect-inject/controllers/pod/pod.go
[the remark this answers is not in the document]
zalimeni: Ah, yeah that's fair. SGTM!

control-plane/go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/google/go-cmp v0.5.9
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/hashicorp/consul/api v1.10.1-0.20230821180813-217d305b38d5
-	github.com/hashicorp/consul/proto-public v0.1.2-0.20230821180813-217d305b38d5 // this points to a commit on Consul main
+	github.com/hashicorp/consul/proto-public v0.1.2-0.20230829221456-f8812eddf1ef // this points to a commit on Consul main
```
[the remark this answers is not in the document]
zalimeni: `:jinx-coke-intensifies:`

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,352 @@
...
+	// NSMirroringPrefix is an optional prefix that can be added to the Consul
+	// namespaces created while mirroring. For example, if it is set to "k8s-",
+	// then the k8s `default` namespace will be mirrored in Consul's
+	// `k8s-default` namespace.
+	NSMirroringPrefix string
+
+	// TODO: EnableWANFederation
+
+	// Partition is the Consul Partition to which this controller belongs
+	Partition string
```
zalimeni, line 79: ~ Maybe put this below `EnableConsulPartitions` and rename to `ConsulPartition`, similar to the namespace args?

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,352 @@
...
+	ConsulServerConnMgr consul.ServerConnectionManager
+	// Only pods in the AllowK8sNamespacesSet are reconciled.
+	AllowK8sNamespacesSet mapset.Set
+	// Pods in the DenyK8sNamespacesSet are ignored.
+	DenyK8sNamespacesSet mapset.Set
+	// EnableConsulPartitions indicates that a user is running Consul Enterprise
+	// with version 1.11+ which supports Admin Partitions.
+	EnableConsulPartitions bool
+	// EnableConsulNamespaces indicates that a user is running Consul Enterprise
+	// with version 1.7+ which supports namespaces.
```
zalimeni, line 61: ~ I think we can drop the version comments here since we won't support this controller w/ earlier versions of Consul

control-plane/helper/test/test_util.go
```diff
@@ -257,6 +271,27 @@ func SetupK8sAuthMethodWithNamespaces(t *testing.T, consulClient *api.Client, se
 	require.NoError(t, err)
 }
 
+// ResourceHasPersisted checks that a recently written resource exists in the Consul
+// state store with a valid version. This must be true before a resource is overwritten
+// or deleted.
```
zalimeni, line 276: AFAIK this shouldn't actually be the case for unversioned writes - should we clarify that this is specifically intended to stabilize tests because currently there are issues w/ deletion CAS race conditions? (I'm assuming this is related to [this](https://hashicorp.slack.com/archives/C04AM35BFQT/p1693260271895209) thread)

zalimeni (approved):

control-plane/connect-inject/controllers/pod/pod_controller.go
DanStough: I think partition is unambiguously a Consul thing but NBD to change the name anyway.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -0,0 +1,1035 @@
...
+			ConsulClientConfig:    testClient.Cfg,
+			ConsulServerConnMgr:   testClient.Watcher,
+			AllowK8sNamespacesSet: mapset.NewSetWith("*"),
+			DenyK8sNamespacesSet:  mapset.NewSetWith(),
+			ResourceClient:        resourceClient,
+		}
+
+		err = pc.writeWorkload(context.Background(), *tc.pod)
+		require.NoError(t, err)
+
```
ndhanushkodi, line 162: Does there need to be a ResourceHasPersisted check here before the read in L175?

ndhanushkodi (approved):
The testing is 🎉

control-plane/connect-inject/controllers/pod/pod_controller_test.go
DanStough: I _think_ it's not needed here because we are doing a read after write, and in this case, we don't care if the Consul controller has reconciled the resource yet. In the other cases, we are overwriting/deleting resources, which had the CAS behavior at the time of testing.

DanStough: I could also totally be mistaken, so please correct me. I didn't see any flakes with the test, unlike the others.

ndhanushkodi: >In the other cases, we are overwriting/deleting resources, which had the CAS behavior at the time of testing.

Ah ok, so the delete wasn't going through sometimes, unless you know consul has persisted? As opposed to in this case, the read is being done from some higher level client/cache?

I think that makes sense, I just wanted to understand the difference between this case and the others! If you ran something like a -count=100 without flakes I think it's fine!

PR #2882, opened 2023-08-31
Support telemetry.disable_hostname

DanielYWoo:
Changes proposed in this PR:
- Support telemetry.disable_hostname

How I've tested this PR:
with the unit test in `server-config-configmap.bats`
however, for some reason there are other failed test cases, I am not sure if they are related to this PR.

help me fix the unrelated unit test

Checklist:
- [X] Tests added


esthu (approved):

david-yu:
@DanielYWoo Could you tell us the use case for this PR (service mesh or service discovery)? Also what was the problem the PR was looking to address? Thanks that helps for context when reviewing the PR.

PR #2883, opened 2023-08-31
feat: add v2 endpoints controller

zalimeni:
_Follows pattern of iterative review in https://github.com/hashicorp/consul-k8s/pull/2868, and currently targets that branch due to dependence on common code added there_

------------

This PR is part of a series of PRs that add functionality to the V2 Endpoints Controller. For the MVP task to be complete, there will be a few additional changes to resolve TODOs, debounce upsert writes if needed, and add further tests.

Changes proposed in this PR:

- Add the V2 Endpoints Controller
- Add enough logic in the controller to manage Consul Services
- Endpoints controller test framework, borrowed from linked Pod Controller PR (thanks @DanStough!)

How I expect reviewers to test this PR: 👁️‍🗨️👃🏻👁️‍🗨️ 

Checklist:
- [x] Tests added

### DIFF
--- a/control-plane/connect-inject/common/common.go
+++ b/control-plane/connect-inject/common/common.go
@@ -9,10 +9,13 @@ import (
 	"strings"
 
 	mapset "github.com/deckarep/golang-set"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v1alpha1"
 )
 
 // DetermineAndValidatePort behaves as follows:
@@ -115,4 +118,45 @@ func ShouldIgnore(namespace string, denySet, allowSet mapset.Set) bool {
 
 func ConsulNodeNameFromK8sNode(nodeName string) string {
 	return fmt.Sprintf("%s-virtual", nodeName)
+}
+
+// ********************
+// V2 Exclusive Common Code
+// ********************
+
+// ToProtoAny is a convenience function for converting proto.Message values to anypb.Any without error handling.
+// This should _only_ be used in cases where a nil or valid proto.Message value is _guaranteed_, else it will panic.
+// If the type of m is *anypb.Any, that value will be returned unmodified.
+func ToProtoAny(m proto.Message) *anypb.Any {
+	switch v := m.(type) {
+	case nil:
+		return nil
+	case *anypb.Any:
+		return v
+	}
+	a, err := anypb.New(m)
+	if err != nil {
+		panic(fmt.Errorf("unexpected error: failed to convert proto message to anypb.Any: %w", err))
+	}
+	return a
+}
+
+// GetPortProtocol matches the Kubernetes EndpointPort.AppProtocol or ServicePort.AppProtocol (*string) to a supported
+// Consul catalog port protocol. If nil or unrecognized, the default of `PROTOCOL_UNSPECIFIED` is returned.
+func GetPortProtocol(appProtocol *string) pbcatalog.Protocol {
+	if appProtocol == nil {
+		return pbcatalog.Protocol_PROTOCOL_UNSPECIFIED
+	}
+	switch *appProtocol {
+	case "tcp":
+		return pbcatalog.Protocol_PROTOCOL_TCP
+	case "http":
+		return pbcatalog.Protocol_PROTOCOL_HTTP
+	case "http2":
+		return pbcatalog.Protocol_PROTOCOL_HTTP2
+	case "grpc":
+		return pbcatalog.Protocol_PROTOCOL_GRPC
+	}
+	// If unrecognized or empty string, return default
+	return pbcatalog.Protocol_PROTOCOL_UNSPECIFIED
 }
--- a/control-plane/connect-inject/common/common_test.go
+++ b/control-plane/connect-inject/common/common_test.go
@@ -7,12 +7,17 @@ import (
 	"testing"
 
 	mapset "github.com/deckarep/golang-set"
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/testing/protocmp"
+	"google.golang.org/protobuf/types/known/anypb"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
 	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
+	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v1alpha1"
+	"github.com/hashicorp/consul/proto-public/pbresource"
 )
 
 func TestCommonDetermineAndValidatePort(t *testing.T) {
@@ -310,6 +315,86 @@ func TestShouldIgnore(t *testing.T) {
 	for _, tt := range cases {
 		t.Run(tt.name, func(t *testing.T) {
 			actual := ShouldIgnore(tt.namespace, tt.denySet, tt.allowSet)
+			require.Equal(t, tt.expected, actual)
+		})
+	}
+}
+
+func TestToProtoAny(t *testing.T) {
+	t.Parallel()
+
+	t.Run("nil gets nil", func(t *testing.T) {
+		require.Nil(t, ToProtoAny(nil))
+	})
+
+	t.Run("anypb.Any gets same value", func(t *testing.T) {
+		testMsg := &pbresource.Resource{Id: &pbresource.ID{Name: "foo"}}
+		testAny, err := anypb.New(testMsg)
+		require.NoError(t, err)
+
+		require.Equal(t, testAny, ToProtoAny(testAny))
+	})
+
+	t.Run("valid proto is successfully serialized", func(t *testing.T) {
+		testMsg := &pbresource.Resource{Id: &pbresource.ID{Name: "foo"}}
+		testAny, err := anypb.New(testMsg)
+		require.NoError(t, err)
+
+		if diff := cmp.Diff(testAny, ToProtoAny(testMsg), protocmp.Transform()); diff != "" {
+			t.Errorf("unexpected difference:\n%v", diff)
+		}
+	})
+}
+
+func TestGetPortProtocol(t *testing.T) {
+	t.Parallel()
+	toStringPtr := func(s string) *string {
+		return &s
+	}
+	cases := []struct {
+		name     string
+		input    *string
+		expected pbcatalog.Protocol
+	}{
+		{
+			name:     "nil gets UNSPECIFIED",
+			input:    nil,
+			expected: pbcatalog.Protocol_PROTOCOL_UNSPECIFIED,
+		},
+		{
+			name:     "tcp gets TCP",
+			input:    toStringPtr("tcp"),
+			expected: pbcatalog.Protocol_PROTOCOL_TCP,
+		},
+		{
+			name:     "http gets HTTP",
+			input:    toStringPtr("http"),
+			expected: pbcatalog.Protocol_PROTOCOL_HTTP,
+		},
+		{
+			name:     "http2 gets HTTP2",
+			input:    toStringPtr("http2"),
+			expected: pbcatalog.Protocol_PROTOCOL_HTTP2,
+		},
+		{
+			name:     "grpc gets GRPC",
+			input:    toStringPtr("grpc"),
+			expected: pbcatalog.Protocol_PROTOCOL_GRPC,
+		},
+		{
+			name:     "case sensitive",
+			input:    toStringPtr("gRPC"),
+			expected: pbcatalog.Protocol_PROTOCOL_UNSPECIFIED,
+		},
+		{
+			name:     "unknown gets UNSPECIFIED",
+			input:    toStringPtr("foo"),
+			expected: pbcatalog.Protocol_PROTOCOL_UNSPECIFIED,
+		},
+	}
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			actual := GetPortProtocol(tt.input)
 			require.Equal(t, tt.expected, actual)
 		})
 	}
--- a/control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
@@ -1,0 +1,369 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+package endpointsv2
+
+import (
+	"context"
+	"net"
+	"strings"
+
+	mapset "github.com/deckarep/golang-set"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/common"
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	"github.com/hashicorp/consul-k8s/control-plane/consul"
+	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
+	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v1alpha1"
+	"github.com/hashicorp/consul/proto-public/pbresource"
+	"github.com/hashicorp/go-multierror"
+)
+
+const (
+	metaKeyManagedBy = "managed-by"
+	kindReplicaSet   = "ReplicaSet"
+)
+
+type Controller struct {
+	client.Client
+	// ConsulServerConnMgr is the watcher for the Consul server addresses used to create Consul API v2 clients.
+	ConsulServerConnMgr consul.ServerConnectionManager
+	// Only endpoints in the AllowK8sNamespacesSet are reconciled.
+	AllowK8sNamespacesSet mapset.Set
+	// Endpoints in the DenyK8sNamespacesSet are ignored.
+	DenyK8sNamespacesSet mapset.Set
+	// EnableConsulPartitions indicates that a user is running Consul Enterprise.
+	EnableConsulPartitions bool
+	// ConsulPartition is the Consul Partition to which this controller belongs.
+	ConsulPartition string
+	// EnableConsulNamespaces indicates that a user is running Consul Enterprise.
+	EnableConsulNamespaces bool
+	// ConsulDestinationNamespace is the name of the Consul namespace to create
+	// all config entries in. If EnableNSMirroring is true this is ignored.
+	ConsulDestinationNamespace string
+	// EnableNSMirroring causes Consul namespaces to be created to match the
+	// k8s namespace of any config entry custom resource. Config entries will
+	// be created in the matching Consul namespace.
+	EnableNSMirroring bool
+	// NSMirroringPrefix is an optional prefix that can be added to the Consul
+	// namespaces created while mirroring. For example, if it is set to "k8s-",
+	// then the k8s `default` namespace will be mirrored in Consul's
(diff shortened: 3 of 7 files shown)


control-plane/connect-inject/common/common_test.go
```diff
@@ -314,3 +319,75 @@ func TestShouldIgnore(t *testing.T) {
 		})
 	}
 }
+
+func TestToProtoAny(t *testing.T) {
```
zalimeni, line 323: Full disclosure, I didn't test panics in this function intentionally because 1) I'm not really sure how to trigger one easily and 2) the purpose of the function is use when doing things that should never panic.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -0,0 +1,351 @@
...
+						podExactNames[podName.Name] = true
+					}
+				}
+			}
+		}
+	}
+
+	//TODO(zalimeni) Do we need to check service-enable label here on service/deployments/replicasets?
+	// Should we expect owner is annotated, and if so, is that consistent for non-replicaset workloads?
+	// If so, update to register and deregister conditionally.
```
zalimeni, line 151: [This](https://docs.google.com/document/d/1q01OHEnQ6Ir40y5GeVq53X5iN-5YExHshX_pashLvlQ/edit?disco=AAAA2PmcWks) 🧵 discusses maybe not needing explicit exclusion for services. I'm not 100% certain whether this controller should expect any opt-outs that will impact service registration, outside of disabled k8s namespaces from config, already checked above.

Input welcome! Lacking a strong opinion, I'll probably convert this to a follow-up ticket as a placeholder for a pre-GA decision.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -0,0 +1,351 @@
...
+func getOwnerPrefixFromPodName(podName string) string {
+	podNameParts := strings.Split(podName, "-")
+	return strings.Join(podNameParts[:len(podNameParts)-1], "-")
+}
+
+// getOwnerPrefixFromPod returns the common name prefix of the pod, if the pod is a member of a set with a unique name
+// prefix. Currently, this only applies to ReplicaSets.
+//
+// We have to fetch the owner and check its type because pod names cannot be disambiguated from pod owner names due to
+// the `-` delimiter and unique ID parts also being valid name components.
```
zalimeni, line 184: Feel free to correct me if anyone knows better here. 🧵 for context [here](https://hashicorp.slack.com/archives/C04MG2UT3PA/p1692203149392719?thread_ts=1692199919.770389&cid=C04MG2UT3PA).

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -0,0 +1,351 @@
...
+		service.Name, // Consul and Kubernetes service name will always match
+		r.getConsulNamespace(service.Namespace),
+		r.getConsulPartition(),
+		getServiceMeta(service),
+	)
+
+	r.Log.Info("registering service with Consul", getLogFieldsForResource(serviceResource.Id)...)
+	//TODO(zalimeni) Read from Consul or some other state to determine whether a write is needed.
+	// For now, we blindly write state on each reconcile to simplify implementation.
+	_, err := resourceClient.Write(ctx, &pbresource.WriteRequest{Resource: serviceResource})
```
zalimeni, line 213: My thinking on this r/n is that we should just try to read the existing Service, and if we find it, check for total equality minus resource version. If there's a diff, rewrite.

If we move to an `EndpointSlices`/`Pods`-based approach in the future, we'll need to do some sort of checking for stale workload selector values, but since `Endpoints` gives us a full snapshot up to the 1K pod limit, we should be able to just upsert today w/o being concerned with drift.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -0,0 +1,351 @@
...
+
+// getServicePorts converts Kubernetes Service ports data into Consul service ports.
+func getServicePorts(service corev1.Service) []*pbcatalog.ServicePort {
+	ports := make([]*pbcatalog.ServicePort, 0, len(service.Spec.Ports)+1)
+
+	for _, p := range service.Spec.Ports {
+		ports = append(ports, &pbcatalog.ServicePort{
+			VirtualPort: uint32(p.Port),
+			TargetPort:  p.TargetPort.String(),
+			Protocol:    common.GetPortProtocol(p.AppProtocol),
```
zalimeni, line 254: I'd appreciate a spot check on this mapping JIC I'm missing some nuance - cc @ishustava in case you have input

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -0,0 +1,351 @@
...
+		})
+	}
+
+	// Append Consul service mesh port in addition to discovered ports.
+	//TODO(zalimeni) Omit if zero mesh ports present in service endpoints? Or could there be some
+	// use of mesh-inject at the service level in the webhook as well?
+	ports = append(ports, &pbcatalog.ServicePort{
+		TargetPort: "mesh",
+		Protocol:   pbcatalog.Protocol_PROTOCOL_MESH,
+	})
```
zalimeni, line 264: Assuming we support discovery sync (non-mesh use cases) in the future, I think we'd probably want to exclude this port addition - but it's probably something we can defer. Marked `TODO(zalimeni)` JIC we already know the answer before I ticket it as a follow-up.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	return r.ConsulPartition
+}
+
+func getLogFieldsForResource(id *pbresource.ID) []any {
+	return []any{
+		"name", id.Name,
+		"namespace", id.Tenancy.Namespace,
+		"partition", id.Tenancy.Partition,
+	}
+}
```
zalimeni, line 351: Maybe overkill, maybe share-able 🤷🏻‍♂️

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller_test.go
```diff
@@ -0,0 +1,404 @@
...
+	err = anypb.UnmarshalTo(expectedResource.Data, expectedService, proto.UnmarshalOptions{})
+	require.NoError(t, err)
+
+	actualService := &pbcatalog.Service{}
+	err = res.GetResource().GetData().UnmarshalTo(actualService)
+	require.NoError(t, err)
+
+	if diff := cmp.Diff(actualService, expectedService, protocmp.Transform()); diff != "" {
+		t.Errorf("unexpected difference:\n%v", diff)
+	}
```
zalimeni, line 356: @DanStough you might be interested in adding this to your comp fn - it's what we use in goldens and is really nice for a visual diff

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller_test.go
```diff
@@ -0,0 +1,404 @@
...
+
+func createServicePodOwnedBy(ownerKind, ownerName string) *corev1.Pod {
+	return createServicePod(ownerKind, ownerName, randomKubernetesId())
+}
+
+func createServicePod(ownerKind, ownerName, podId string) *corev1.Pod {
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%s", ownerName, podId),
+			Namespace: "default",
```
zalimeni, line 367: `TODO`: I'll need to fix this for non-default ns tests

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	//TODO(zalimeni) This is adapted from v1 pod-based meta injection - verify correct and needed
+	for k, v := range service.Annotations {
+		if strings.HasPrefix(k, constants.AnnotationMeta) && strings.TrimPrefix(k, constants.AnnotationMeta) != "" {
+			if v == "$SERVICE_NAME" {
+				meta[strings.TrimPrefix(k, constants.AnnotationMeta)] = service.Name
+			} else {
+				meta[strings.TrimPrefix(k, constants.AnnotationMeta)] = v
+			}
+		}
+	}
```
zalimeni, line 296: I wasn't sure whether we wanted to keep allowing for custom meta via annotations, but seemed feasible to adapt from the v1 controller. Thoughts? Happy to cut if we're not sure about supporting it, or the interpolation of `$SERVICE_NAME`.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
DanStough: Was there a specific case you were trying to address by taking the diff and rewriting it? There was some discussion of running multiple versions of the controller concurrently in the future, and I wonder if this could create flapping with different workers fighting over the same service.

I think I naively assumed that if the write doesn't error, the resource service has persisted the service. Some future work is planned for later releases to sync the status of Consul resources back to Kubernetes resources. Do you think this would address any of your concerns?

control-plane/connect-inject/common/common.go
```diff
@@ -116,3 +119,40 @@ func ShouldIgnore(namespace string, denySet, allowSet mapset.Set) bool {
...
+	}
+	switch *appProtocol {
+	case "tcp":
+		return pbcatalog.Protocol_PROTOCOL_TCP
+	case "http":
+		return pbcatalog.Protocol_PROTOCOL_HTTP
+	case "http2":
+		return pbcatalog.Protocol_PROTOCOL_HTTP2
+	case "grpc":
+		return pbcatalog.Protocol_PROTOCOL_GRPC
```
DanStough, line 154: I wonder if we technically need to include prefixes with some of these protocols. I was reading the API docs for [appProtocol](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#serviceport-v1-core) and it mentions only[ IANA names](https://www.iana.org/assignments/service-names) can be unprefixed. Only my first time seeing this so I could be mistaken.

I also think an annotation on the service or pod might be easier for customers to support. e.g. the [Bitnami Postgres helmchart](https://github.com/bitnami/charts/tree/main/bitnami/postgresql) exposes annotations for services and certain pods but not a change of appProtocol.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
DanStough: I left this out for now, but I don't have a strong opinion.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
DanStough: I don't think `targetPort` will work in the case it is a number, but the pod defines the service with a name. We will need to look up the corresponding port name in the pod.  I'm open to changing how this works on the Pod side but not sure how you get around the name being optional and `targetPort` working on either.

DanStough (commented):
I didn't get a chance to review the test case yet, but everything else looks great! I wanted to send you my open-ended comments in advance, like how `targetPort` works and the `appProtocol`.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: > Was there a specific case you were trying to address by taking the diff and rewriting it?

Mainly just avoiding needless writes - TBF I'm not fully up to speed on the resource API and state store changes, but my basic understanding is reads are cheaper, and we are theoretically capped in the 2-3K/s for writes. This seemed like an easy way to avoid needless clobbering, esp. when potentially many reconciles loops attempt to write at once w/o versions and the baked-in CAS retries compete w/ each other.

>  There was some discussion of running multiple versions of the controller concurrently in the future, and I wonder if this could create flapping with different workers fighting over the same service.

I think the flapping would occur regardless of pre-reads? But that's a good callout. I'd hope we'd come up w/ some one-way version tagging similar or identical to the `managed-by` tag if needed to avoid issues from that. (Do you have a link to that conversation by chance?)

> I think I naively assumed that if the write doesn't error, the resource service has persisted the service. Some future work is planned for later releases to sync the status of Consul resources back to Kubernetes resources. Do you think this would address any of your concerns?

That's my assumption too 👍🏻 ATM I'm only thinking of debouncing, though bidirectional sync would add complexity to that.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: > I don't think targetPort will work in the case it is a number, but the pod defines the service with a name

(_This might be wrong or obvious, feel free to correct me_) It seems to me from the k8s [docs](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service) that there's an inherent tradeoff that applies generally: you need to pick whether to use a name consistently across service and pods (which enables mixed pod port numbers under one name), or you need to pick a consistent pod port number and use that in the service. Either way, it seems selected pods need to adhere to the service's chosen int-or-string value for `targetPort`.

Our docs for Consul `Service.TargetPort` say `target_port is the name of the workload port`, so I took that to mean we defer totally to the value used by the k8s service.

It seems wrong to me to reach into the pods to infer anything beyond the workload selector, since the `targetPort` mapping isn't really a `consul-k8s`-specific limitation... But I might be missing a reason we'd want to do that.

> I'm open to changing how this works on the Pod side

What did you have in mind here?

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
DanStough: I think it was in a casual Zoom, so I don't happen to have a link to the convo.

control-plane/connect-inject/common/common_test.go
DanStough: Let's keep it for now.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
DanStough: > It seems wrong to me to reach into the pods to infer anything beyond the workload selector, since the targetPort mapping isn't really a consul-k8s-specific limitation

Yeah, I don't like the idea of reaching in either. The pods may be slightly different, and the names don't align, making the lookup challenging. But I think I disagree that there isn't something K8s-specific here. We generate a Consul port name per port, even if one wasn't defined in the Kube spec, and Kubernetes doesn't necessarily require you to use any name at all, so there is a mismatch of strong (consul) vs. weak (k8s) references. I think I had the opposite reaction that since the VM case is straightforward, we'd have to "fix it" in consul-k8s.

>> I'm open to changing how this works on the Pod side
No specific ideas here. I'm just trying not to look like I'm suggesting a singular solution.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: Asked 🧵 [here](https://hashicorp.slack.com/archives/C04AM35BFQT/p1694005306847929) to seek wisdom

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: Started 🧵 [here](https://hashicorp.slack.com/archives/C04MG2UT3PA/p1694005024255509) to figure out how we want this to work

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: Marking this as a figure-out-later item for now

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: Omitted for brevity 🚀 TODO'd for follow-up

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller_test.go
zalimeni: Leaving as follow-up in the interest of early testing using default ns

zalimeni (commented):
@DanStough @thisisnotashwin ok, I think this is ready for a next pass. I've added more tests around the essential CrUD bits as well as the pod fetching behavior since that seemed valuable to lock in.

Assuming no major changes are needed, I'll plan to merge this to unblock testing while I follow up on remaining TODOs and Slack 🧵s.

control-plane/connect-inject/common/common_test.go
zalimeni: Added handling for `anypb.Any` input -> return unmodified and test case to match.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+// getOwnerPrefixFromPod returns the common name prefix of the pod, if the pod is a member of a set with a unique name
+// prefix. Currently, this only applies to ReplicaSets.
+//
+// We have to fetch the owner and check its type because pod names cannot be disambiguated from pod owner names due to
+// the `-` delimiter and unique ID parts also being valid name components.
+//
+// If the pod owner does not have a unique name, the empty string is returned.
+func getOwnerPrefixFromPod(pod *corev1.Pod) string {
```
DanStough, line 192: Just an interesting fact, I noticed there is a field [generateName](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#objectmeta-v1-meta) that includes the shared prefixes that pods used. I think the current approach is more robust though.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller_test.go
DanStough, line 1: WDYT about an additional test with some `NotReadyAddresses` in the Endpoint Subset to test that they are included in the name generation?

DanStough (approved):
I think based on all the tradeoffs we discussed to get a version of this ready to test, LGTM.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: Ah yeah, thanks for pointing out! I think I ran across some doc referring to that field early on and had the same thought; it didn't seem guaranteed enough to hang implementation on over the owner ref.

For fun, here's an example:
```yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2023-08-28T15:55:38Z"
  generateName: fluentd-elasticsearch-85996f8dbd-
  labels:
    app: nginx
    pod-template-hash: 85996f8dbd
  name: fluentd-elasticsearch-85996f8dbd-9r67q
  namespace: default
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: ReplicaSet
    name: fluentd-elasticsearch-85996f8dbd
    uid: 610a6961-944c-4863-9738-7d783dd1992c
  resourceVersion: "1841"
  uid: 28e1fdd5-0107-4ddc-890b-e07065d6000f
```

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller_test.go
zalimeni: Great call, I'll do that 👍🏻

thisisnotashwin (approved):
BEAUTIFUL!!!

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: Summary from Slack: not entirely conclusive, but we agree early optimization should be generally avoided until we get a better idea of write performance in the wild.

I think it's fair to argue that debouncing rapid identical writes falls under low-hanging fruit to reduce needless load, but is leading me to additional questions about how we'll handle state in resources that may be mutated by multiple sources/data flows (not sure how much of this there will be - could resource `Meta` tagging be an example?), since we also can't just blindly rewrite records in that scenario.

Regardless, this is tagged for follow-up outside this PR.

control-plane/connect-inject/common/common.go
zalimeni: Added this as a bullet in follow-up ticket so we don't lose track but probably don't need to figure out right this second.

zalimeni: Circling back to this one to tie up the loose end: @ishustava, do you happen to have perspective on either of Dan's questions above?

My gut is to leave the unprefixed protocols in place for now since there seems to be precedent in other meshes, and wait to see if we need annotations in addition to just `appProtocol`. If that's not ideal, we can fix now or plan to follow up in 1.18.

PR #2881, opened 2023-08-31
[NET-5346] Expose JWKCluster fields in jwt-provider config entry

roncodingenthusiast:
Changes proposed in this PR:
- After exposing JWKCluster https://github.com/hashicorp/consul/pull/17978 we want to ensure the CRD for jwt-provider config entry matches the one in consul core

How I've tested this PR:
- Manually
- Unit tests

How I expect reviewers to test this PR:

TODO:
- [x] Tests added
- [x] changelog entry

PR #2887, opened 2023-09-01
Fixing diff resulting from make copy-crds-to-chart

roncodingenthusiast:
Changes proposed in this PR:
- When working on https://github.com/hashicorp/consul-k8s/pull/2881, I noticed there was an additional diff when trying to copy generated CRD YAML into charts/consul
- `make copy-crds-to-chart` seems to be creating a diff locally on main. My assumption is: folks are manually creating the charts. Not sure what the expectation is  

How I've tested this PR:
- Running `make copy-crds-to-chart` locally 

- If needed, provide any obvious context I may be missing. 
- Callout any expected tests I need to update


roncodingenthusiast:
closing this as it is super outdated

PR #2892, opened 2023-09-04
made server locality-init configurable

aahel:
Changes proposed in this PR:
- Made storing server locality configurable through helm chart.

How I've tested this PR:
installed consul using server.locality = true/false. 

Checklist:
- [ ] Tests added

### DIFF
--- a/charts/consul/templates/server-statefulset.yaml
+++ b/charts/consul/templates/server-statefulset.yaml
@@ -229,6 +229,7 @@ spec:
       priorityClassName: {{ .Values.server.priorityClassName | quote }}
       {{- end }}
       initContainers:
+      {{- if .Values.server.locality }}
       - name: locality-init
         image: {{ .Values.global.imageK8S }}
         env:
@@ -245,6 +246,7 @@ spec:
           - name: extra-config
             mountPath: /consul/extra-config
         {{- include "consul.restrictedSecurityContext" . | nindent 8 }}
+      {{- end }}
       containers:
         - name: consul
           image: "{{ default .Values.global.image .Values.server.image }}"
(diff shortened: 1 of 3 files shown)


charts/consul/templates/server-statefulset.yaml
```diff
@@ -229,6 +229,7 @@ spec:
       priorityClassName: {{ .Values.server.priorityClassName | quote }}
       {{- end }}
       initContainers:
+      {{- if .Values.server.locality }}
```
Ganeshrockz, line 232: This renders a block that looks something like (when locality is disabled)

```yaml
initContainers:
containers:
- name: consul
....
```

Does this cause syntax errors? I guess it does.

charts/consul/values.yaml
```diff
@@ -776,6 +776,10 @@ server:
   # @type: string
   logLevel: ""
 
+  # If false locality of server nodes won't be stored 
+  # @type: boolean
+  locality: true
```
Ganeshrockz, line 781: Should this be (in case we decide to further parameterize the init container's logic)

```
locality:
   enabled: true
```
instead?

charts/consul/templates/server-statefulset.yaml
aahel: no

charts/consul/values.yaml
aahel: we can.

aahel: https://github.com/hashicorp/consul-k8s/pull/2892/commits/f52bdd72824ba2f2f1488f14450dc95bccd227d2

Issue #1285, opened 2022-06-16
Cannot create Ingress Gateways with the same name in different namespaces


thomashashi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

When using the Helm chart version 0.42.0 (haven't been able to try it with the latest version) I am unable to create two ingress gateways with the same name but in different namespaces. I could do that manually if I were outside of Kubernetes, but cannot with the Helm chart — I believe we make some assumptions in how we name the Deployments and Pods created which don't take into account the namespace into which an IGW is provisioned.

### Reproduction Steps

Create a Helm values file with the following stanza:

```
ingressGateways:
  enabled: true
  defaults:
    annotations: |
      "service.beta.kubernetes.io/aws-load-balancer-internal": "true"
      "service.beta.kubernetes.io/aws-load-balancer-scheme": "internal"
      "service.beta.kubernetes.io/aws-load-balancer-type": "nlb"
  gateways:
    - name: ingressgw
      consulNamespace: gateways
      service:
        type: ClusterIP
        ports:
          - port: 80
          - port: 443
    - name: ingressgw
      consulNamespace: k8s-122-gateways
      service:
        type: ClusterIP
        ports:
          - port: 80
          - port: 443
```
and attempt to deploy.

### Logs

Output from `helm upgrade consul hashicorp/consul -f consul-helm.yaml --namespace consul --version 0.42.0` (where `consul-helm.yaml` is the values file containing the stanza in the previous section): 

```
Error: UPGRADE FAILED: template: consul/templates/ingress-gateways-deployment.yaml:18:3: executing "consul/templates/ingress-gateways-deployment.yaml" at <fail (cat "ingress gateways must have unique names but found duplicate name" $innerIngressVal.name)>: error calling fail: ingress gateways must have unique names but found duplicate name ingressgw
```

### Environment details

* Helm chart version 0.42.0
* Consul Enterprise 1.11.4

jmurret:
[...]

PR #2903, opened 2023-09-05
Address issue #1285 via docs

david-yu:
Changes proposed in this PR:
- Addresses issue https://github.com/hashicorp/consul-k8s/issues/1285#issuecomment-1160461529. Ingress Gateways are deprecated so the limitation will remain. 

How I've tested this PR:

Checklist:
- [ ] Tests added

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -3072,7 +3072,7 @@ ingressGateways:
 
   # Gateways is a list of gateway objects. The only required field for
   # each is `name`, though they can also contain any of the fields in
-  # `defaults`. Names must be provided for each and they must be
+  # `defaults`. Names must be provided for each ingress gateway and they must be
   # unique, regardless of whether they are in different namespaces. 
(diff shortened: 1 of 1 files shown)


charts/consul/values.yaml
```diff
@@ -3072,8 +3072,9 @@ ingressGateways:
 
   # Gateways is a list of gateway objects. The only required field for
   # each is `name`, though they can also contain any of the fields in
-  # `defaults`. Values defined here override the defaults except in the
-  # case of annotations where both will be applied.
+  # `defaults`. Names must be provided for each ingress gateway and they must be
+  # unique, regardless of whether they are in different namespaces. 
+  # Values defined here override the defaults except in the case of annotations where both will be applied.
```
im2nguyen, line 3077:
```suggestion
  # `defaults`. You must provide a unique name for each ingress gateway. These names 
  # must be unique across different namespaces. 
  # Values defined here override the defaults, except in the case of annotations where both will be applied.
```

im2nguyen (approved):

PR #2904, opened 2023-09-05
NET-5531 Translate response header modifier(s) from HTTPRoute onto Consul config entry

nathancoleman:
> **Note**
> This PR depends on changes in https://github.com/hashicorp/consul/pull/18646

Translate the `responseHeaderModifier` that already exists on the K8s `HTTPRoute` onto the `http-route` config entry in Consul, which accepts response header modifiers as of https://github.com/hashicorp/consul/pull/18646.

This PR also addresses the fact that we write an empty header modifier to Consul today for request header modifications when the K8s `HTTPRoute` does not specify any header modifications. Instead, we now appropriately write an empty list of header modifications to Consul when the `HTTPRoute` does not specify any.

**How I've tested this PR:**
Create a K8s `HTTPRoute` with various response header modifier filter(s) and verify that the filter syncs into Consul and functions correctly. See [video](https://github.com/hashicorp/consul-k8s/assets/3476400/5be69969-821b-4413-9771-61b5366e8084) of my own testing.

In addition, conformance tests exist [here](https://github.com/kubernetes-sigs/gateway-api/blob/fbf223030a7b3dd8ef15a5b22ed64e08ec3d5385/conformance/tests/httproute-response-header-modifier.go) for this functionality that should pass where they previously failed. Here is the output of those tests passing.
CleanShot 2023-09-12 at 11 54 23@2x

See above

**Checklist:**
- [x] Tests added


control-plane/api-gateway/common/translation.go
```diff
@@ -275,26 +278,65 @@ func (t ResourceTranslator) translateHTTPQueryMatch(match gwv1beta1.HTTPQueryPar
...
+	// Multiple filters applying the same or conflicting operations are allowed but may
+	// result in unexpected behavior.
 	for _, filter := range filters {
 		if filter.RequestHeaderModifier != nil {
-			consulFilter.Remove = append(consulFilter.Remove, filter.RequestHeaderModifier.Remove...)
+			newFilter := api.HTTPHeaderFilter{}
 
-			for _, toAdd := range filter.RequestHeaderModifier.Add {
-				consulFilter.Add[string(toAdd.Name)] = toAdd.Value
+			if len(filter.RequestHeaderModifier.Remove) > 0 {
```
jm96441n, line 297: why do we need the len checks here? won't the append effectively be a no-op if the remove slice is emtpy?

control-plane/go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/hashicorp/consul-k8s/control-plane
 
 // TODO: remove when the SDK is released for Consul 1.17
 // The replace directive is needed be because `api` requires 0.14.1 of SDK and is both a direct and indirect dependency
-replace github.com/hashicorp/consul/sdk v0.14.1 => github.com/hashicorp/consul/sdk v0.4.1-0.20230825164720-ecdcde430924
+replace (
+	github.com/hashicorp/consul/proto-public v0.4.1 => github.com/hashicorp/consul/proto-public v0.1.2-0.20230911164019-a69e901660bd
```
jm96441n, line 6: is there a reason why we're adding a new replace here? if it's to work with an unreleased version should we update this comment as well?

jm96441n (approved):
few minor questions but overall LGTM

control-plane/api-gateway/common/translation.go
nathancoleman: Good point, addressing...

control-plane/go.mod
nathancoleman: I added a comment explaining the replace directive that I added as well as making the TODO more robust

PR #2910, opened 2023-09-06
[NET-5399] Improve token fetching performance for endpoints controller.

hashi-derek:
Prior to this change, the endpoints controller would list all ACL tokens in a namespace when a service instance is being deleted. This commit improves the performance by querying only the necessary subset of tokens by service-identity / service-name.

Changes proposed in this PR:
- Update the Consul API version
- Modify the endpoints controller to utilize the new API and fetch fewer tokens.

How I've tested this PR:

- Ensured that calling `/v1/acl/tokens?servicename=mysvc` does not throw an error on prior versions of Consul 1.16 and simply returns the full list of tokens. This means that updating consul-k8s to use the new query param will still be backwards compatible (because it can still loop and filter on the full list of tokens in-memory, rather than looping on the subset).
- Ensured that calling `/v1/acl/tokens?servicename=mysvc` returns the proper subset of tokens on a consul-k8s install using this new build.
- Created a consul-k8s cluster with services and ensured that tokens are still created / deleted properly.

Checklist:
- [x] Tests added

### DIFF
--- a/acceptance/tests/connect/connect_inject_test.go
+++ b/acceptance/tests/connect/connect_inject_test.go
@@ -134,6 +134,24 @@ func TestConnectInject_CleanupKilledPods(t *testing.T) {
 			require.Len(t, pods.Items, 1)
 			podName := pods.Items[0].Name
 
+			// Ensure the token exists
+			if secure {
+				retry.Run(t, func(r *retry.R) {
+					tokens, _, err := consulClient.ACL().TokenListFiltered(
+						api.ACLTokenFilterOptions{ServiceName: "static-client"}, nil)
+					require.NoError(r, err)
+					// Ensure that the tokens exist. Note that we must iterate over the tokens and scan for the name,
+					// because older versions of Consul do not support the filtered query param and will return
+					// the full list of tokens instead.
+					count := 0
+					for _, t := range tokens {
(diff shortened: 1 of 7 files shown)


hashi-derek:
This PR is marked as no-backport due to go.mod changes. There will be manual backports created for 1.0.X -> 1.2.X

acceptance/tests/connect/connect_inject_test.go
```diff
@@ -134,6 +134,24 @@ func TestConnectInject_CleanupKilledPods(t *testing.T) {
...
+					// Ensure that the tokens exist. Note that we must iterate over the tokens and scan for the name,
+					// because older versions of Consul do not support the filtered query param and will return
+					// the full list of tokens instead.
+					count := 0
+					for _, t := range tokens {
+						if len(t.ServiceIdentities) > 0 && t.ServiceIdentities[0].ServiceName == "static-client" {
+							count++
+						}
+					}
+					require.Greater(r, count, 0)
```
hashi-derek, line 152: Another assertion exists further down that ensures the token is deleted properly. This was just added here to make sure that the token exists in the first place.

wilkermichael (approved):
LGTM too!

PR #2911, opened 2023-09-06
feat: add v2 pod controller healthstatus and proxyconfiguration

DanStough:
Changes proposed in this PR:
- Adds support in the V2 Pod Controller for HealthStatuses and ProxyConfigurations.
- Enterprise tests and logging will come in a subsequent PR.

How I've tested this PR:
- Unit and integration tests as included in the PR

Checklist:
- [X] Tests added


control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -110,7 +120,13 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	if k8serrors.IsNotFound(err) {
 
-		if err := r.deleteWorkload(ctx, req.NamespacedName); err != nil {
+		if err := r.deleteHealthStatus(ctx, req.NamespacedName); err != nil {
+			errs = multierror.Append(errs, err)
+		}
+
+		// TODO: clean up ACL Tokens
+
+		if err := r.deleteProxyConfiguration(ctx, req.NamespacedName); err != nil {
```
zalimeni, line 129: Does order matter at all here on the Consul side before the workload is deleted (i.e. should we fast-fail at any point)?

Also wondering whether we can rely on Consul controllers to do these deletes for us - I'd expect clients to be able to rely on Consul to manage cascading deletes of dependent resources via `Owner` w/o managing that part of the lifecycle explicitly. Is that something we're not planning on ATM? (I can't find any references in the RFCs that describe owner links, but it seems implied by the k8s-oriented data model)

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -210,33 +233,238 @@ func (r *Controller) writeWorkload(ctx context.Context, pod corev1.Pod) error {
...
-//func (r *Controller) writeHealthStatus(pod corev1.Pod) error {
-//	return nil
-//}
+func (r *Controller) getTproxyMode(ctx context.Context, pod corev1.Pod) (pbmesh.ProxyMode, error) {
+	// A user can enable/disable tproxy for an entire namespace.
+	var ns corev1.Namespace
+	err := r.Client.Get(ctx, types.NamespacedName{Name: pod.GetNamespace()}, &ns)
+	if err != nil {
+		return pbmesh.ProxyMode_PROXY_MODE_DEFAULT, fmt.Errorf("could not get namespace info for %s: %w", pod.GetNamespace(), err)
+	}
```
zalimeni, line 311: ~ Should we plumb the ns through higher up to simplify testing and so we don't accidentally introduce redundant fetches later in other parts of the code?

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -210,33 +233,238 @@ func (r *Controller) writeWorkload(ctx context.Context, pod corev1.Pod) error {
...
+			if originalContainer.Name == mutatedContainer.Name {
+				paths, err := getContainerExposePaths(originalPod, originalContainer, mutatedContainer)
+				if err != nil {
+					return nil, fmt.Errorf("error getting container expose path for %s: %w", originalContainer.Name, err)
+				}
+
+				exposeConfig.ExposePaths = append(exposeConfig.ExposePaths, paths...)
+			}
+		}
+	}
```
zalimeni, line 355: Just to call out, this looks like a straight port of Endpoints V1, so I assumed it was meant to be 1-1 and didn't read closely. If this needs more thorough review happy to try myself or ask for a hand.

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -210,33 +233,238 @@ func (r *Controller) writeWorkload(ctx context.Context, pod corev1.Pod) error {
...
+			LocalPathPort: originalStartupPort,
+			Path:          mutatedContainer.StartupProbe.HTTPGet.Path,
+		}
+		paths = append(paths, newPath)
+	}
+	return paths, nil
+}
+
+func (r *Controller) getBootstrapConfig(pod corev1.Pod) (*pbmesh.BootstrapConfig, error) {
+
```
zalimeni, line 408: ~ Nit
```suggestion
```

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -210,33 +233,238 @@ func (r *Controller) writeWorkload(ctx context.Context, pod corev1.Pod) error {
...
+	bootstrap := &pbmesh.BootstrapConfig{}
+
+	// If metrics are enabled, the BootstrapConfig should set envoy_prometheus_bind_addr to a listener on 0.0.0.0 on
+	// the PrometheusScrapePort that points to a metrics backend. The backend for this listener will be determined by
+	// the envoy bootstrapping command (consul connect envoy) or the consul-dataplane GetBoostrapParams rpc.
+	// If there is a merged metrics server, the backend would be that server.
+	// If we are not running the merged metrics server, the backend should just be the Envoy metrics endpoint.
```
zalimeni, line 415: Same here re: Endpoints V1 code port skimming

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -321,6 +551,40 @@ func parseLocality(node corev1.Node) *pbcatalog.Locality {
...
+}
+
+// getHealthStatusReason takes Consul's health check status (either passing or critical)
+// and the pod to return a descriptive output for the HealthStatus Output.
+func getHealthStatusReason(state pbcatalog.Health, pod corev1.Pod) string {
+	if state == pbcatalog.Health_HEALTH_PASSING {
+		return constants.KubernetesSuccessReasonMsg
+	}
+
+	return fmt.Sprintf("Pod \"%s/%s\" is not ready", pod.GetName(), pod.GetNamespace())
```
zalimeni, line 585: ~ Should name and ns be swapped here?

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -322,20 +298,519 @@ func TestWorkloadWrite(t *testing.T) {
...
+			name:                 "not ready pod",
+			pod:                  createPod("foo", "10.0.0.1", "foo", true, false),
+			expectedHealthStatus: createCriticalHealthStatus(),
+		},
+		{
+			name: "pod with no condition",
+			pod:  createPod("foo", "10.0.0.1", "foo", true, true),
+			podModifier: func(pod *corev1.Pod) {
+				pod.Status.Conditions = []corev1.PodCondition{}
+			},
```
zalimeni, line 492: Nice 👍🏻

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -322,20 +298,519 @@ func TestWorkloadWrite(t *testing.T) {
...
+			require.Equal(t, codes.NotFound, s.Code())
+			return
+		}
+
+		require.NoError(t, err)
+		require.NotNil(t, actualRes)
+
+		require.Equal(t, tc.pod.GetName(), actualRes.GetResource().GetId().GetName())
+		require.Equal(t, constants.DefaultConsulNS, actualRes.GetResource().GetId().GetTenancy().GetNamespace())
+		require.Equal(t, constants.DefaultConsulPartition, actualRes.GetResource().GetId().GetTenancy().GetPartition())
```
zalimeni, line 687: ~~ Could maybe add an `ID` equals helper for the tests that need it, just for better legibility

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -565,29 +1008,23 @@ func TestReconcileCreatePod(t *testing.T) {
 			podName: "foo",
 			k8sObjects: func() []runtime.Object {
 				pod := createPod("foo", "10.0.0.1", "foo", true, true)
+				podBytes, _ := json.Marshal(pod)
+				pod.Annotations[constants.AnnotationOriginalPod] = string(podBytes)
+
+				// Fake the probe changes that would be added by the mesh webhook
+				pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Port = intstr.FromInt(20300)
+				pod.Spec.Containers[0].LivenessProbe.HTTPGet.Port = intstr.FromInt(20400)
+				pod.Spec.Containers[0].StartupProbe.HTTPGet.Port = intstr.FromInt(20500)
```
zalimeni, line 1017: ~ Here and elsewhere: Should we `TODO` a follow-up to use a shared helper so these tests don't drift from the hook impl? If not, maybe a local helper for reuse?

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -1015,17 +1620,76 @@ func expectedWorkloadMatches(t *testing.T, client pbresource.ResourceServiceClie
...
+	require.NoError(t, err)
+	require.NotNil(t, res)
+
+	require.Equal(t, name, res.GetResource().GetId().GetName())
+	require.Equal(t, constants.DefaultConsulNS, res.GetResource().GetId().GetTenancy().GetNamespace())
+	require.Equal(t, constants.DefaultConsulPartition, res.GetResource().GetId().GetTenancy().GetPartition())
+
+	require.NotNil(t, res.GetResource().GetData())
+
+	actualWorkload := &pbcatalog.HealthStatus{}
```
zalimeni, line 1645:
```suggestion
	actualHealthStatus := &pbcatalog.HealthStatus{}
```

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -1015,17 +1620,76 @@ func expectedWorkloadMatches(t *testing.T, client pbresource.ResourceServiceClie
...
+	err = res.GetResource().GetData().UnmarshalTo(actualWorkload)
+	require.NoError(t, err)
+
+	require.True(t, proto.Equal(actualWorkload, expectedHealthStatus))
+}
+
+func expectedProxyConfigurationMatches(t *testing.T, client pbresource.ResourceServiceClient, name string, expectedProxyConfiguration *pbmesh.ProxyConfiguration) {
```
zalimeni, line 1652: ~~ I feel like maybe we could convert these to a generic "resource compare" fn, but no strong feeling about doing it now

control-plane/go.mod
```diff
@@ -49,6 +47,11 @@ require (
 	sigs.k8s.io/gateway-api v0.7.1
 )
 
+require (
+	github.com/hashicorp/consul-k8s/control-plane/cni v0.0.0-20230905204616-c6b703d1deff
+	github.com/hashicorp/consul-server-connection-manager v0.1.4
+)
```
zalimeni, line 53: ~ Maybe just me not understanding `go mod tidy` magic: why the new `require` block here? I'm used to only seeing direct and indirect deps split out.

zalimeni (approved):
A few minor suggestions and one major [question](https://github.com/hashicorp/consul-k8s/pull/2911#discussion_r1318952241) about cascading deletes - otherwise LGTM!

Approving to unblock based on whether you think the comments are actionable; if so happy to re-review as needed.

control-plane/go.mod
DanStough: Yeah, not sure. I moved to 1.21 locally, and I've seen a bit of this moving around. I didn't see anything in the release notes about it. I'll probably leave it for now and assume `go mod tidy` knows what it likes.

zalimeni: Sounds good; CI should complain if `1.20` `go mod tidy` doesn't like the format.

control-plane/connect-inject/controllers/pod/pod_controller.go
DanStough: Yeah, IMHO, order definitely matters. I was trying to remove the health check first to prevent routing to the workload. ProxyConfiguration should go last for deletion. Of course, since everything is eventually consistent, perhaps it's less important than I think.

That could be where Consul comes in. I have observed that Consul does handle cascading deletes. I kept the explicit deletions to be safe, but it might help with the consistency to let Consul do it. I will update it so HealthStatus is deleted implicitly.  

Thinking about the eventual consistency, though, maybe we need to wait until the workload is removed to delete the proxy configuration, lest we change the behavior before the Consul reconcile completes?

control-plane/connect-inject/controllers/pod/pod_controller.go
DanStough: Yep, this is verbatim V1 Endpoints code, shifted around a bit to make it easier to read.

control-plane/connect-inject/controllers/pod/pod_controller.go
DanStough: Yep, good catch!

control-plane/connect-inject/controllers/pod/pod_controller.go
DanStough: I think it's a good idea, but I'll defer optimizing when we need this data in some other context.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
DanStough: 👍 I added a local helper to simplify.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
DanStough: I like this idea and started playing with it. Implementing it is a little messier than I'd like, so I'm thinking I'll revisit it as part of the ENT tests.

DanStough:
@zalimeni maybe you can just take a peek again at the bit where I add the PeerName and make sure it makes sense to you?

control-plane/connect-inject/controllers/pod/pod_controller.go
zalimeni: That’s a good point about ProxyConfiguration. It makes me feel like we should have that point to a Workload as owner too, so there’s a cascading deletion relationship, and perhaps validate on the server side that selector and owner are 1-1 when owner is specified. IMO it’s not super robust to rely on this controller to manage that lifecycle since ProxyConfiguration could easily be orphaned if the pod goes away when the controller is erroring or otherwise MIA. Do you think that makes sense as a question to pose or is the lack of guaranteed GC a constraint we’ve chosen to live with explicitly somewhere?

zalimeni (approved):
LGTM re: PeerName addition, though wouldn’t hurt to add a comment explaining why it’s (temporarily) needed. I feel like the follow up volume just keeps growing and won’t age well… 😬

control-plane/connect-inject/controllers/pod/pod_controller.go
DanStough: It's a good question to ask. It seems like if we start by deleting the owner, then deleting the orphans in GC, that would be sufficient for what we need. I think the problem with using an Owner ref is that it would be optional. I'm not sure either way if the controller arch supports it.

PR #2912, opened 2023-09-06
Added validating webhook for gateway policy

jm96441n:
"NOTE: this targets an open PR
Changes proposed in this PR:
- adds infra for validating webhooks
- adds validating webhook for incoming gateway policy

How I've tested this PR:
- unit tests
- ran locally with https://github.com/jm96441n/consul-experiments/tree/main/k8s/jwts and added the `consul/gatewaypolicy.yaml` spec and then should see the `consul/bad_policy.yaml` apply fail

- read the code

Checklist:
- [X] Tests added


control-plane/api-gateway/controllers/index.go
```diff
@@ -336,9 +338,18 @@ func gatewayForGatewayPolicy(o client.Object) []string {
 	gatewayPolicy := o.(*v1alpha1.GatewayPolicy)
 
 	targetGateway := gatewayPolicy.Spec.TargetRef
-	//gateway policy is 1to1
-	if targetGateway.Group == v1alpha1.ConsulHashicorpGroup && targetGateway.Kind == common.KindGateway {
-		return []string{common.IndexedNamespacedNameWithDefault(targetGateway.Name, &targetGateway.Namespace, gatewayPolicy.Namespace).String()}
+	// gateway policy is 1to1
+	fmt.Println()
+	if targetGateway.Group == "gateway.networking.k8s.io/v1beta1" && targetGateway.Kind == common.KindGateway {
```
jm96441n, line 343: mad some changes here because this was not working as expected

control-plane/api/v1alpha1/gatewaypolicy_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/hashicorp/consul-k8s/control-plane/api/common"
+)
+
+const Gatewaypolicy_GatewayIndex = "__gatewaypolicy_referencing_gateway"
```
jm96441n, line 19: had to copy this to avoid circular imports

control-plane/api-gateway/controllers/index.go
t-eckert: Let's add `"gateway.networking.k8s.io/v1beta1"` to common/constants.go so we can change it more easily when we upgrade in the future.

t-eckert (approved):
This is really clean. I like that you also improved the existing code organization.

control-plane/api-gateway/controllers/index.go
jm96441n: good catch! this constant exists in the k8s sig package, I thought I caught all the refs I had hardcoded (this will also probably change in @sarahalsmiller PR where she's implementing it

control-plane/api/v1alpha1/gatewaypolicy_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+	})
+
+	if err != nil {
+		v.Logger.Error(err, "error getting list of policies referencing gateway")
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	for _, policy := range list.Items {
+		if differentPolicySameTarget(resource, policy) {
+			return admission.Denied("policy targets a gateway listener that is already the target of an existing policy")
```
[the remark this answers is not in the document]
t-eckert: Yes, given that it's readily available. That will just help people debug.

PR #2844, opened 2023-08-25
Add RateLimit config to serviceDefaults

kisunji:
Changes proposed in this PR:
- Adds the RateLimits field to `serviceDefaults` CRD to match the addition in https://github.com/hashicorp/consul/pull/18583

How I've tested this PR:
- Added unit tests

How I expect reviewers to test this PR:

Checklist:
- [x] Tests added
- [x] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #2932, opened 2023-09-08
Add local rate limiting acceptance test

cthain:
Changes proposed in this PR:

This PR adds a local rate limiting acceptance test to the `connect` folder. It is a follow on to #2844. 

How I've tested this PR:
- Successfully ran the acceptance tests locally

:eyes: 

Checklist:
- [x] Tests added

### DIFF
--- a/acceptance/tests/connect/local_rate_limit_test.go
+++ b/acceptance/tests/connect/local_rate_limit_test.go
@@ -49,7 +49,7 @@ func TestConnectInject_LocalRateLimiting(t *testing.T) {
 
 	opts := newRateLimitOptions(t, ctx)
 
-	// Ensure that all requests from static-client to static-server succeed.
+	// Ensure that all requests from static-client to static-server succeed (no rate limiting set).
 	for addr, rps := range rateLimitMap {
 		opts.rps = rps
 		assertRateLimits(t, opts, addr)
--- a/acceptance/tests/connect/local_rate_limit_test.go
+++ b/acceptance/tests/connect/local_rate_limit_test.go
@@ -1,0 +1,126 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package connect
+
+import (
+	"testing"
+	"time"
+
+	terratestK8s "github.com/gruntwork-io/terratest/modules/k8s"
+
+	"github.com/hashicorp/consul-k8s/acceptance/framework/connhelper"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/consul"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/environment"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+)
+
+// TestConnectInject_LocalRateLimiting tests that local rate limiting works as expected between services.
+func TestConnectInject_LocalRateLimiting(t *testing.T) {
+	cfg := suite.Config()
+	ctx := suite.Environment().DefaultContext(t)
+
+	releaseName := helpers.RandomName()
+	connHelper := connhelper.ConnectHelper{
+		ClusterKind:     consul.Helm,
+		Secure:          false,
+		ReleaseName:     releaseName,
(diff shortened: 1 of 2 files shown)


acceptance/tests/connect/local_rate_limit_test.go
```diff
@@ -0,0 +1,126 @@
...
+	// Map the static-server URL and path to the rate limits defined in the service defaults at:
+	// ../fixtures/cases/local-rate-limiting/service-defaults-static-server.yaml
+	rateLimitMap := map[string]int{
+		"http://localhost:1234":             2,
+		"http://localhost:1234/exact":       3,
+		"http://localhost:1234/prefix-test": 4,
+		"http://localhost:1234/regex":       5,
+	}
+
+	opts := newRateLimitOptions(t, ctx)
```
wilkermichael, line 50: Just a thought, but maybe it would be more straightforward to create an array of opts and add the addr/rate limit to each one? Would keep the configuration all in one place.

Then you could just loop on the opts

acceptance/tests/connect/local_rate_limit_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Ensure that all requests from static-client to static-server succeed.
```
wilkermichael, line 52:
```suggestion
	// Ensure that all requests from static-client to static-server succeed (no rate limiting set)
```

acceptance/tests/connect/local_rate_limit_test.go
```diff
@@ -0,0 +1,126 @@
...
+	enforced        bool
+	retryTimer      *retry.Timer
+	curlOpts        string
+}
+
+func newRateLimitOptions(t *testing.T, ctx environment.TestContext) *assertRateLimitOptions {
+	return &assertRateLimitOptions{
+		resourceType:    "deploy/",
+		successOutput:   "hello world",
+		rateLimitOutput: "curl: (22) The requested URL returned error: 429",
```
wilkermichael, line 120: I might even consider changing this to just contain `429` in case any part of this string were to change in the future, 429 would be the most likely part to stay consistent

acceptance/tests/connect/local_rate_limit_test.go
```diff
@@ -0,0 +1,126 @@
...
+	// Ensure that going over the limit causes the static-server to apply rate limiting and
+	// reply with 429 Too Many Requests
+	opts.enforced = true
+	for addr, reqPerSec := range rateLimitMap {
+		opts.rps = reqPerSec
+		assertRateLimits(t, opts, addr)
+	}
+}
+
+func assertRateLimits(t *testing.T, opts *assertRateLimitOptions, curlArgs ...string) {
```
wilkermichael, line 70: Just an observation that I think this mostly just tests `requestMaxBursts`. Do we want to add tests for throttling based on `requestsPerSecond`?

wilkermichael (approved):
Thanks for adding this Chris! Looks great!

Mostly just some observations, nothing critical here.

kisunji (approved):
This is great! Love your assertion that the bursts happened in 1s.

acceptance/tests/connect/local_rate_limit_test.go
cthain: Yeah that might be a nice update. My only concern with that is we would still have to do bits of the `opts` setup in the loop. I think I'll stick with the host/rps map for now.

acceptance/tests/connect/local_rate_limit_test.go
cthain: It's possible that this could change but I think it's unlikely :shrug:. If curl changes its output format/messages we'll likely have to change things elsewhere. I followed what was used in the `connhelper` pkg for testing failed connections: https://github.com/hashicorp/consul-k8s/blob/3d3bd70b7168c346e55be0dbfdae7175111ad25a/acceptance/framework/connhelper/connect_helper.go#L417-L419

acceptance/tests/connect/local_rate_limit_test.go
cthain: Yes this test is bursty. It was pretty difficult to setup a reliable burst test and with the current framework I think it would be even more difficult to reliably test throttling. We're using `curl` via `kubectl exec` and everything has to happen in under 1 second because we don't allow tweaking the `fill_interval`.

I'm open to ideas if you have thoughts on how we could do this :thinking:

PR #2950, opened 2023-09-12
[NET-5017] APIGW Status Conditions for Gateway for JWT/Reconcile on JWTProvider Changes

jm96441n:
Changes proposed in this PR:
- Adds status conditions translation for JWT related issues on Gateways (gateway policy will be a follow up PR)
- Sets up the reconcile workflow for when changes to a referenced JWT provider occur

How I've tested this PR:
1. after that finishes run `kubectl describe gateway api-gateway` to see all the statuses as valid
1. run `kubectl apply ./consul/gatewaypolicy-okta.yaml`, this will add a policy to the gateway listener that references a non-existent jwt provider named `okta`
1. run `kubectl describe gateway api-gateway`  again to see the gateway has an invalid jwt providers status as well as the listener having an invalid jwt provider status
1. run `kubectl describe gateway api-gateway`  again to see the gateway back in a healthy status

Read the code
Run against kind cluster using steps above

Checklist:
- [x] Tests added

### DIFF
--- a/control-plane/api-gateway/binding/result.go
+++ b/control-plane/api-gateway/binding/result.go
@@ -240,6 +240,7 @@ var (
 	errListenerInvalidCertificateRef_InvalidData      = errors.New("certificate is invalid or does not contain a supported server name")
 	errListenerInvalidCertificateRef_NonFIPSRSAKeyLen = errors.New("certificate has an invalid length: RSA Keys must be at least 2048-bit")
 	errListenerInvalidCertificateRef_FIPSRSAKeyLen    = errors.New("certificate has an invalid length: RSA keys must be either 2048-bit, 3072-bit, or 4096-bit in FIPS mode")
+	errListenerJWTProviderNotFound                    = errors.New("policy referencing this listener references unknown JWT provider")
 	errListenerInvalidRouteKinds                      = errors.New("allowed route kind is invalid")
 	errListenerProgrammed_Invalid                     = errors.New("listener cannot be programmed because it is invalid")
 
@@ -269,7 +270,7 @@ type listenerValidationResult struct {
 	// status type: Conflicted
 	conflictedErr error
 	// status type: ResolvedRefs
-	refErr error
+	refErrs []error
 	// status type: ResolvedRefs (but with internal validation)
 	routeKindErr error
 }
@@ -281,7 +282,7 @@ func (l listenerValidationResult) programmedCondition(generation int64) metav1.C
 	now := timeFunc()
 
 	switch {
-	case l.acceptedErr != nil, l.conflictedErr != nil, l.refErr != nil, l.routeKindErr != nil:
+	case l.acceptedErr != nil, l.conflictedErr != nil, len(l.refErrs) != 0, l.routeKindErr != nil:
 		return metav1.Condition{
 			Type:               "Programmed",
 			Status:             metav1.ConditionFalse,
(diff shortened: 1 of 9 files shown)


control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -466,6 +472,10 @@ func SetupGatewayControllerWithManager(ctx context.Context, mgr ctrl.Manager, co
 			&source.Channel{Source: c.Subscribe(ctx, api.InlineCertificate, r.transformConsulInlineCertificate(ctx)).Events()},
 			&handler.EnqueueRequestForObject{},
 		).
+		Watches(
```
jm96441n, line 475: we only care about when consul receives the JWTProvider so we only watch for the entry to come from consul

control-plane/api-gateway/cache/consul.go
```diff
@@ -222,7 +223,7 @@ func (c *Cache) updateAndNotify(ctx context.Context, once *sync.Once, kind strin
 
 	for _, entry := range entries {
 		meta := entry.GetMeta()
-		if meta[constants.MetaKeyKubeName] == "" || meta[constants.MetaKeyDatacenter] != c.datacenter {
+		if (meta[constants.MetaKeyKubeName] == "" || meta[constants.MetaKeyDatacenter] != c.datacenter) && kind != api.JWTProvider {
```
jm96441n, line 226: I kinda feel icky about this but I'm not sure of another way to skip this if we get a JWTProvider back

control-plane/api-gateway/common/resources.go
```diff
@@ -429,6 +435,23 @@ func (s *ResourceMap) AddGatewayPolicy(gatewayPolicy *v1alpha1.GatewayPolicy) *v
...
 
+func (s *ResourceMap) AddJWTProvider(provider *v1alpha1.JWTProvider) {
+	key := api.ResourceReference{
+		Kind: provider.Kind,
+		Name: provider.Name,
+	}
+	s.jwtProviders[key] = provider
+}
+
+func (s *ResourceMap) GetJWTProviderForProvider(provider *v1alpha1.GatewayJWTProvider) (*v1alpha1.JWTProvider, bool) {
```
nathancoleman, line 446: The name of this function causes me to rethink what's happening every time it's used. I kind of wish our `v1alpha1.GatewayJWTProvider` was `...JWTProviderReference` and this function was `GetJWTProviderForReference`. Is this just me?

control-plane/api-gateway/cache/consul.go
nathancoleman: Agree. I think nesting the `if` might make this easier to comprehend?

control-plane/api-gateway/common/resources.go
```diff
@@ -437,6 +460,7 @@ func (s *ResourceMap) GetPolicyForGatewayListener(gateway gwv1beta1.Gateway, gat
 		Namespace:   gateway.Namespace,
 	}
 
+	fmt.Printf("%#v\n", s.gatewayPolicies)
```
nathancoleman, line 463: Is this intended to stick around

control-plane/api-gateway/common/resources.go
jm96441n: yeah, I've been hitting more and more weirdness with the name, I'm down with changing the name, though `JWTProviderReference` isn't as descriptive because this also contains the claim information about what we should be verifying

control-plane/api-gateway/common/resources.go
jm96441n: nooooope, missed that one

control-plane/api-gateway/common/resources.go
```diff
@@ -429,6 +435,23 @@ func (s *ResourceMap) AddGatewayPolicy(gatewayPolicy *v1alpha1.GatewayPolicy) *v
...
+	s.jwtProviders[key] = provider
+}
+
+func (s *ResourceMap) GetJWTProviderForProvider(provider *v1alpha1.GatewayJWTProvider) (*v1alpha1.JWTProvider, bool) {
+	key := api.ResourceReference{
+		Name: provider.Name,
+		Kind: "JWTProvider",
+	}
+	value, exists := s.jwtProviders[key]
+	return value, exists
```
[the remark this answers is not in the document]
jm96441n: can do, my editor linter does noooot like this but it compiles!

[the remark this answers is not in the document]
jm96441n: I lied, this actually doesn't build

control-plane/api-gateway/common/resources.go
nathancoleman: That makes sense. Maybe even `GetJWTProviderForGatewayJWTProvider` at the risk of being too verbose. It's not a blocker, just a stumbling point when reading code

jm96441n: I can do that, also down to have more of a brainstorming session to see if we can name these a bit better

control-plane/api-gateway/binding/result.go
```diff
@@ -281,7 +282,7 @@ func (l listenerValidationResult) programmedCondition(generation int64) metav1.C
 	now := timeFunc()
 
 	switch {
-	case l.acceptedErr != nil, l.conflictedErr != nil, l.refErr != nil, l.routeKindErr != nil:
+	case l.acceptedErr != nil, l.conflictedErr != nil, len(l.refErrs) != 0, l.routeKindErr != nil:
```
t-eckert, line 285: I think you need to `nil` check `l.refErrs`.

t-eckert: Nevemind, I was wrong. `len(nil)` just returns 0.

control-plane/api-gateway/binding/result.go
```diff
@@ -382,59 +383,75 @@ func (l listenerValidationResult) conflictedCondition(generation int64) metav1.C
...
-		return metav1.Condition{
-			Type:               "ResolvedRefs",
-			Status:             metav1.ConditionFalse,
-			Reason:             "RefNotPermitted",
-			ObservedGeneration: generation,
-			Message:            l.refErr.Error(),
-			LastTransitionTime: now,
+	for _, refErr := range l.refErrs {
+		switch refErr {
+		case errListenerInvalidCertificateRef_NotFound, errListenerInvalidCertificateRef_NotSupported, errListenerInvalidCertificateRef_InvalidData, errListenerInvalidCertificateRef_NonFIPSRSAKeyLen, errListenerInvalidCertificateRef_FIPSRSAKeyLen:
```
t-eckert, line 404: For ease of reading: 
```suggestion
		case errListenerInvalidCertificateRef_NotFound,
		     errListenerInvalidCertificateRef_NotSupported,
		     errListenerInvalidCertificateRef_InvalidData,
		     errListenerInvalidCertificateRef_NonFIPSRSAKeyLen,
		     errListenerInvalidCertificateRef_FIPSRSAKeyLen:
```

control-plane/api-gateway/binding/result.go
```diff
@@ -382,59 +383,75 @@ func (l listenerValidationResult) conflictedCondition(generation int64) metav1.C
...
 func (l listenerValidationResult) Conditions(generation int64) []metav1.Condition {
-	return []metav1.Condition{
+	conditions := []metav1.Condition{
 		l.acceptedCondition(generation),
 		l.programmedCondition(generation),
 		l.conflictedCondition(generation),
-		l.resolvedRefsCondition(generation),
 	}
+	conditions = append(conditions, l.resolvedRefsConditions(generation)...)
+	return conditions
```
t-eckert, line 454: You can cut a line here 🤷🏻 
```suggestion
	return append(conditions, l.resolvedRefsConditions(generation)...)
```

control-plane/api-gateway/binding/validation.go
```diff
@@ -226,6 +227,32 @@ func validateTLS(gateway gwv1beta1.Gateway, tls *gwv1beta1.GatewayTLSConfig, res
...
+	policy, _ := resources.GetPolicyForGatewayListener(gateway, listener)
+	if policy == nil {
+		return nil
+	}
+
+	if policy.Spec.Override != nil && policy.Spec.Override.JWT != nil {
+		for _, provider := range policy.Spec.Override.JWT.Providers {
+			_, ok := resources.GetJWTProviderForGatewayJWTProvider(provider)
+			if !ok {
+				return errListenerJWTProviderNotFound
```
t-eckert, line 240: Here and below, you can add specific information about the name of the provider not found:

```suggestion
				return fmt.Errorf("%w, %s", errListenerJWTProviderNotFound, provider)
```

t-eckert: You could combine this with returning multiple errors to return all issues with providers for both default and overrides in one go. That way if a user has issues in both sections, they can find and fix them all at once.

Basically, not returning early and instead returning a slice of errors.

t-eckert (approved):

control-plane/api-gateway/binding/validation.go
jm96441n: good call!

jm96441n: actually for this we don't care about that, the gateway policy is going to be the one with all the information about what is wrong, for this all we care about is if any are incorrect

PR #2952, opened 2023-09-13
GKE Autopilot support

curtbushko:
Changes proposed in this PR:
- This adds a first pass of support for running on GKE Autopilot.
- We won't know how things fully work until we are certified with 1.2.2 (next patch release)
- I tried to do this without any helm chart values changes but it would have been considerably more messy than it is now.

How I've tested this PR:

- went through an initial acceptance pass with the Autopilot team
- installed consul-k8s using the following helm values:
```
global:
  name: consul
  datacenter: dc1
  logLevel: "debug"
  tls:
    enabled: false
ingress:
  enabled: false
server:
  replicas: 1
  resources:
    requests:
      memory: "200Mi"
      cpu: "500m"
    limits:
      memory: "200Mi"
      cpu: "500m"
connectInject:
  enabled: true
  default: true
  replicas: 1
  transparentProxy:
    defaultEnabled: true
  apiGateway:
    manageExternalCRDs: false
    manageCustomCRDs: true
dns:
  enabled: false
ui:
  enabled: true
  service:
    enabled: true
controller:
  enabled: true
```
	

Checklist:
- [ ] Tests added

### DIFF
--- a/acceptance/framework/config/config.go
+++ b/acceptance/framework/config/config.go
@@ -157,7 +157,7 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
 	}
 
 	// UseGKEAutopilot is a temporary hack that we need in place as GKE Autopilot is already installing
-	// Gateway CRDs in the clusters. There are still other CRDs we need to install though (see helm cluster installE)
+	// Gateway CRDs in the clusters. There are still other CRDs we need to install though (see helm cluster install)
 	if t.UseGKEAutopilot {
 		setIfNotEmpty(helmValues, "global.server.resources.requests.cpu", "500m")
 		setIfNotEmpty(helmValues, "global.server.resources.limits.cpu", "500m")
--- a/acceptance/framework/config/config.go
+++ b/acceptance/framework/config/config.go
@@ -99,10 +99,11 @@ type TestConfig struct {
 	NoCleanup          bool
 	DebugDirectory     string
 
-	UseAKS  bool
-	UseEKS  bool
-	UseGKE  bool
-	UseKind bool
+	UseAKS          bool
+	UseEKS          bool
+	UseGKE          bool
+	UseGKEAutopilot bool
+	UseKind         bool
 
 	helmChartPath string
 }
@@ -153,6 +154,15 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
 			// namespace it must be run in a different privileged namespace.
 			setIfNotEmpty(helmValues, "connectInject.cni.namespace", "kube-system")
 		}
+	}
+
+	// UseGKEAutopilot is a temporary hack that we need in place as GKE Autopilot is already installing
+	// Gateway CRDs in the clusters. There are still other CRDs we need to install though (see helm cluster installE)
+	if t.UseGKEAutopilot {
+		setIfNotEmpty(helmValues, "global.server.resources.requests.cpu", "500m")
+		setIfNotEmpty(helmValues, "global.server.resources.limits.cpu", "500m")
+		setIfNotEmpty(helmValues, "connectInject.apiGateway.manageExternalCRDs", "false")
+		setIfNotEmpty(helmValues, "connectInject.apiGateway.manageCustomCRDs", "true")
 	}
 
 	setIfNotEmpty(helmValues, "connectInject.transparentProxy.defaultEnabled", strconv.FormatBool(t.EnableTransparentProxy))
--- a/acceptance/framework/config/config.go
+++ b/acceptance/framework/config/config.go
@@ -162,7 +162,7 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
 		setIfNotEmpty(helmValues, "global.server.resources.requests.cpu", "500m")
 		setIfNotEmpty(helmValues, "global.server.resources.limits.cpu", "500m")
 		setIfNotEmpty(helmValues, "connectInject.apiGateway.manageExternalCRDs", "false")
-		setIfNotEmpty(helmValues, "connectInject.apiGateway.manageCustomCRDs", "true")
+		setIfNotEmpty(helmValues, "connectInject.apiGateway.manageNonStandardCRDs", "true")
 	}
 
 	setIfNotEmpty(helmValues, "connectInject.transparentProxy.defaultEnabled", strconv.FormatBool(t.EnableTransparentProxy))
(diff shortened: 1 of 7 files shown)


acceptance/framework/config/config.go
```diff
@@ -155,6 +156,15 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
 		}
 	}
 
+	// UseGKEAutopilot is a temporary hack that we need in place as GKE Autopilot is already installing
+	// Gateway CRDs in the clusters. There are still other CRDs we need to install though (see helm cluster installE)
```
nathancoleman, line 160:
```suggestion
	// Gateway CRDs in the clusters. There are still other CRDs we need to install though (see helm cluster install)
```

charts/consul/values.yaml
```diff
@@ -2185,6 +2185,11 @@ connectInject:
     # If this setting is false, you will need to install the Gateway API CRDs manually.
     manageExternalCRDs: true
 
+    # Enables Consul on Kubernets to manage only the custom CRDs used for Gateway API. If manageExternalCRDs is true 
+    # then all CRDs will be installed. If manageCustomerCRDs is true then only TCPRoute, gatewayclassconfigs and meshservice
```
nathancoleman, line 2189:
```suggestion
    # then all CRDs will be installed. If manageCustomCRDs is true then only TCPRoute, GatewayClassConfig and MeshService
```

charts/consul/values.yaml
```diff
@@ -2185,6 +2185,11 @@ connectInject:
...
+    # will be installed.
+    manageCustomCRDs: false
```
nathancoleman, line 2191: If I'm understanding correctly, the combinations that matter are

```yaml
# manageCustomCRDs doesn't matter here and could be true or false, all CRDs will be installed
manageExternalCRDs: true
# manageCustomCRDs: false
```

```yaml
# This only installs TCPRoute, GatewayClassConfig and MeshService
manageExternalCRDs: false
manageCustomCRDs: true
```

```yaml
# No CRDs at all will be installed
manageExternalCRDs: false
manageCustomCRDs: false 
```

nathancoleman (changes requested):
I think it would be better to avoid having "custom" refer to both our own CRDs -- which are "custom" for Consul -- and some of the CRDs defined by the Kubernetes SIG that have not graduated to "standard" yet but most likely will in the future.

These are the groups as I see them
- "Standard" CRDs defined by Gateway API [here](https://github.com/kubernetes-sigs/gateway-api/tree/main/config/crd/standard) (`Gateway`, `HTTPRoute`, etc.)
- "Experimental" CRDs defined by Gateway API [here](https://github.com/kubernetes-sigs/gateway-api/tree/main/config/crd/experimental) (`TCPRoute` in this case)
- "Consul" CRDs defined by us (`GatewayClassConfig`, `MeshService`)

Here are a couple of slightly different config structures that I think make more sense given that grouping. Let me know what you think.

```yaml
manageExternalCRDs: false # Let GKE manage the "standard" CRDs
manageNonStandardCRDs: true # Install TCPRoute, GatewayClassConfig + MeshService
```

```yaml
manageExternalCRDs: false # Let GKE manage the "standard" CRDs
# This is a higher bar for the config since this set *must* be installed right now; however, it
# avoids grouping things together and acknowledges that TCPRoute, for example, might one
# day be considered "standard". Our upgrade instructions would need to include
#   - "remove TCPRoute if..."
#   - "add SomeNewConsulCRD if..."
manageIndividualCRDs:
- TCPRoute
- GatewayClassConfig
- MeshService
```

charts/consul/values.yaml
curtbushko: That looks correct.

nathancoleman:
I'm fairly certain @Jeff-Apple will have thoughts on the UX here

curtbushko:
> I think it would be better to avoid having "custom" refer to both our own CRDs -- which are "custom" for Consul -- and some of the CRDs defined by the Kubernetes SIG that have not graduated to "standard" yet but most likely will in the future.
> 
> These are the groups as I see them
> 
>     * "Standard" CRDs defined by Gateway API [here](https://github.com/kubernetes-sigs/gateway-api/tree/main/config/crd/standard) (`Gateway`, `HTTPRoute`, etc.)
> 
>     * "Extended" CRDs defined by Gateway API [here](https://github.com/kubernetes-sigs/gateway-api/tree/main/config/crd/experimental) (`TCPRoute` in this case)
> 
>     * "Consul" CRDs defined by us (`GatewayClassConfig`, `MeshService`)
> 
> 
> Here are a couple of slightly different config structures that I think make more sense given that grouping. Let me know what you think.
> 
> ```yaml
> manageExternalCRDs: false # Let GKE manage the "standard" CRDs
> manageNonStandardCRDs: true # Install TCPRoute, GatewayClassConfig + MeshService
> ```
> 
> ```yaml
> manageExternalCRDs: false # Let GKE manage the "standard" CRDs
> # This is a higher bar for the config since this set *must* be installed right now; however, it
> # avoids grouping things together and acknowledges that TCPRoute, for example, might one
> # day be considered "standard". Our upgrade instructions would include "remove TCPRoute if..."
> manageIndividualCRDs:
> - TCPRoute
> - GatewayClassConfig
> - MeshService
> ```

I am fine with `manageNonStandardCRD`, but not the list. I do not want to get into that game for something that you will probably change in the future...

david-yu:
I'm ok with `manageNonStandardCRD` as well seems to be appropriate.

charts/consul/values.yaml
```diff
@@ -2185,10 +2185,10 @@ connectInject:
     # If this setting is false, you will need to install the Gateway API CRDs manually.
     manageExternalCRDs: true
 
-    # Enables Consul on Kubernets to manage only the custom CRDs used for Gateway API. If manageExternalCRDs is true 
-    # then all CRDs will be installed. If manageCustomCRDs is true then only TCPRoute, GatewayClassConfig and MeshService
+    # Enables Consul on Kubernets to manage only the non-standard CRDs used for Gateway API. If manageExternalCRDs is true 
+    # then all CRDs will be installed. If manageNonStandardCRDs is true then only TCPRoute, GatewayClassConfig and MeshService
```
nathancoleman, line 2189:
```suggestion
    # then all CRDs will be installed; otherwise, if manageNonStandardCRDs is true then only TCPRoute, GatewayClassConfig and MeshService
```

nathancoleman (approved):

PR #2955, opened 2023-09-13
[NET-5017] APIGW Status Conditions for Gateway Policies

jm96441n:
Changes proposed in this PR:
- Add status conditions for gateway policies

How I've tested this PR:
1. run `kubectl apply ./consul/gatewaypolicy-okta.yaml`
1. after that finishes run `kubectl describe gatewaypolicy gw-policy-2` to see all the statuses as invalid because the referenced provider does not exist
1. run `kubectl describe gatewaypolicy gw-policy-2`  again to see the policy back in a healthy status

Checklist:
- [x] Tests added


control-plane/api/v1alpha1/gatewaypolicy_types.go
```diff
@@ -115,3 +115,21 @@ type GatewayJWTClaimVerification struct {
 	//   that this value matches.
 	Value string `json:"value"`
 }
+
+// GatewayPolicyStatus defines the observed state of the gateway
+type GatewayPolicyStatus struct {
```
jm96441n, line 120: we needed a specific status type in order to add the status, similar to how the `Gateway` type in the k8s spec has a specific type for `GatewayStatus`

control-plane/api-gateway/binding/binder.go
```diff
@@ -238,6 +243,19 @@ func (b *Binder) Snapshot() *Snapshot {
...
 			snapshot.Kubernetes.StatusUpdates.Add(&b.config.Gateway)
 		}
+
+		for idx, policy := range b.config.Policies {
+			policy := policy
+
+			var policyStatus v1alpha1.GatewayPolicyStatus
+
+			policyStatus.Conditions = policyValidation.Conditions(policy.Generation, idx)
+			// only mark the policy as needing a status update if there's a diff with it's old status
```
t-eckert, line 253: "its" is possessive, "it's" is "it is".

```suggestion
			// only mark the policy as needing a status update if there's a diff with its old status
```

control-plane/api-gateway/binding/result.go
```diff
@@ -582,3 +582,94 @@ func (l gatewayValidationResult) Conditions(generation int64, listenersInvalid b
...
+	return result.Conditions(generation)
+}
+
+func (g gatewayPolicyValidationResult) Conditions(generation int64) []metav1.Condition {
+	conditions := make([]metav1.Condition, 0)
+
+	conditions = append(conditions, g.acceptedCondition(generation))
+	conditions = append(conditions, g.resolvedRefsConditions(generation)...)
+	return conditions
+}
```
t-eckert, line 611: 🤷🏻 

```suggestion
func (g gatewayPolicyValidationResult) Conditions(generation int64) []metav1.Condition {
	return append([]metav1.Condition{g.acceptedCondition(generation)}, g.resolvedRefsConditions(generation)...)
}
```

t-eckert (approved):

charts/consul/templates/crd-gatewaypolicies.yaml
```diff
@@ -199,42 +199,93 @@ spec:
...
+                - lastTransitionTime: "1970-01-01T00:00:00Z"
+                  message: Waiting for controller
+                  reason: Pending
+                  status: Unknown
+                  type: Accepted
+                - lastTransitionTime: "1970-01-01T00:00:00Z"
+                  message: Waiting for controller
+                  reason: Pending
+                  status: Unknown
+                  type: Programmed
```
andrewstucki, line 215: So, this default doesn't match the valid conditions that the controller sets, I think you'll want to change it to `ResolvedRefs`

jm96441n: good  catch! I had copied/pasted this and forgot to change it :facepalm:

control-plane/api-gateway/binding/binder.go
jm96441n: words are hard

PR #2956, opened 2023-09-13
feat: add namespace controller

DanStough:
Changes proposed in this PR:
- Add support for Namespace Controller to sync nodes from K8s to Consul. 
- Removes Consul namespace creation from the mesh webhook, because now there are other generated resources (e.g. ServiceIdentities and Services), that are not tied to this lifecycle.

Proposing the backoff behavior needed in the other controllers for when the namespace is still reconciling will be in a separate PR.

How I expect reviewers to test this PR: 👁️ 👃  👁️ 

Checklist:
- [X] Tests added


control-plane/connect-inject/namespace/namespace_controller_ent_test.go
```diff
@@ -0,0 +1,413 @@
...
+			nsMirrorPrefix:          "k8s-",
+			enableNSMirroring:       true,
+			existingConsulNamespace: getNamespace("k8s-foo", "", false),
+		},
+		{
+			name:                    "mirrored namespaces with prefix, non-default partition",
+			partition:               "baz",
+			nsMirrorPrefix:          "k8s-",
+			enableNSMirroring:       true,
+			existingConsulNamespace: getNamespace("k8s-foo", "", false),
```
zalimeni, line 346: I think this one needs a partition in existing?

```suggestion
			existingConsulNamespace: getNamespace("k8s-foo", "bar", false),
```

control-plane/namespaces/namespaces_test.go
```diff
@@ -159,6 +159,51 @@ func TestEnsureExists_CreatesNS(tt *testing.T) {
...
+					Name: name,
+				}
+				_, _, err = consulClient.Namespaces().Create(&ns, nil)
+				require.NoError(t, err)
+			}
+
+			err = EnsureDeleted(consulClient, name)
+			require.NoError(t, err)
+
+			// Ensure it was created.
```
zalimeni, line 197:
```suggestion
			// Ensure it was deleted.
```

control-plane/namespaces/namespaces_test.go
```diff
@@ -159,6 +159,51 @@ func TestEnsureExists_CreatesNS(tt *testing.T) {
...
+				Address: consul.HTTPAddr,
+			})
+			require.NoError(t, err)
+
+			if c.NamespaceExists {
+				ns := capi.Namespace{
+					Name: name,
+				}
+				_, _, err = consulClient.Namespaces().Create(&ns, nil)
+				require.NoError(t, err)
```
zalimeni, line 191: ~ Should we make sure we can read it first before deleting?

wilkermichael (approved):
Great work Dan!

Some minor comments

control-plane/connect-inject/namespace/namespace_controller.go
```diff
@@ -0,0 +1,130 @@
...
+		For(&corev1.Namespace{}).
+		Complete(r)
+}
+
+// consulNamespace returns the Consul destination namespace for a provided Kubernetes namespace
+// depending on Consul Namespaces being enabled and the value of namespace mirroring.
+func (r *Controller) getConsulNamespace(kubeNamespace string) string {
+	ns := namespaces.ConsulNamespace(
+		kubeNamespace,
+		true,
```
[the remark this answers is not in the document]
DanStough: Yep, that was my thought. There is no `EnableNamespaces` field on this controller because it only runs if that is true.

control-plane/namespaces/namespaces.go
```diff
@@ -55,6 +55,29 @@ func EnsureExists(client *capi.Client, ns string, crossNSAClPolicy string) (bool
 	return true, err
 }
 
+// EnsureDeleted ensures a Consul namespace with name ns is deleted. If it is already not found
+// the call to delete will be skipped.
+func EnsureDeleted(client *capi.Client, ns string) error {
```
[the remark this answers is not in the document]
DanStough: Yes, I intended it to be the former, so I would vote to keep it `EnsureDeleted`.

control-plane/connect-inject/namespace/namespace_controller_ent_test.go
DanStough: Yeah, it should be `baz`. I guess it succeeds since the delete works even if the resource doesn't exist (some kind of idempotency?).

control-plane/namespaces/namespaces_test.go
DanStough: :goodthinking:

control-plane/connect-inject/namespace/namespace_controller_ent_test.go
zalimeni: Lol oops.

But yeah, w/ idempotent deletes and lacking some feedback that it caused a change to occur, it's a bit tricky to rule out false negatives. Could do a "list all namespaces -> count" check in the run function to rule those out?

PR #2960, opened 2023-09-14
feat(control-plane): v2 backoff on missing namespace

DanStough:
Changes proposed in this PR:
- Add requeue logic when the namespace is missing, which could be the case before the namespace controller has reconciled it.

How I've tested this PR: Unit tests on the helper func, but not the controller. We have a task to come back to the enterprise tests. I'm also fine lumping that into this PR.

Checklist:
- [ ] ~Tests added~

### DIFF
--- a/control-plane/connect-inject/common/common.go
+++ b/control-plane/connect-inject/common/common.go
@@ -9,14 +9,17 @@ import (
 	"strings"
 
 	mapset "github.com/deckarep/golang-set"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
-	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
 	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v1alpha1"
+
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
 )
 
 // DetermineAndValidatePort behaves as follows:
@@ -182,6 +185,22 @@ func PortValueFromIntOrString(pod corev1.Pod, port intstr.IntOrString) (uint32, 
 // Does not apply to V1 pods, which use a different key (`constants.KeyInjectStatus`).
 func HasBeenMeshInjected(pod corev1.Pod) bool {
 	if anno, ok := pod.Annotations[constants.KeyMeshInjectStatus]; ok && anno == constants.Injected {
+		return true
+	}
+	return false
+}
+
+// ConsulNamespaceIsNotFound checks the gRPC error code and message to determine
+// if a namespace does not exist. If the namespace exists this function returns false, true otherwise.
+func ConsulNamespaceIsNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+	s, ok := status.FromError(err)
+	if !ok {
+		return false
+	}
+	if codes.InvalidArgument == s.Code() && strings.Contains(s.Message(), "namespace resource not found") {
 		return true
 	}
 	return false
--- a/control-plane/connect-inject/common/common_test.go
+++ b/control-plane/connect-inject/common/common_test.go
@@ -4,20 +4,28 @@
 package common
 
 import (
+	"context"
+	"fmt"
 	"testing"
 
 	mapset "github.com/deckarep/golang-set"
 	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/consul/sdk/testutil"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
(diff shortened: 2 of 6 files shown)


control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -145,10 +145,22 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			}
 			errs = multierror.Append(errs, err)
 		}
 
 		if err := r.writeWorkload(ctx, pod); err != nil {
+			if common.ConsulNamespaceIsNotFound(err) {
+				r.Log.Info("Consul namespace not found; re-queueing request",
+					"pod", req.Name, "ns", req.Namespace, "consul-ns", r.getConsulNamespace(req.Namespace))
+				return ctrl.Result{Requeue: true}, nil
+			}
```
DanStough, line 163: In theory is the previous write succeeds, this is unnecessary, but I kept it in case it gets refactored later.

control-plane/connect-inject/common/common.go
```diff
@@ -177,3 +180,19 @@ func PortValueFromIntOrString(pod corev1.Pod, port intstr.IntOrString) (uint32,
...
+// error indicates that the namespace of for a resource does not yet exist.
+func ConsulNamespaceIsNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+	s, ok := status.FromError(err)
+	if !ok {
+		return false
+	}
+	if codes.InvalidArgument == s.Code() && strings.Contains(s.Message(), "namespace resource not found") {
```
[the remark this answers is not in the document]
wilkermichael: Do we need to verify the message? Or would checking for the `InvalidArgument` code be sufficient?

zalimeni (approved):

control-plane/connect-inject/common/common_test.go
```diff
@@ -399,3 +403,41 @@ func TestGetPortProtocol(t *testing.T) {
...
+		{
+			name:                   "grpc code is InvalidArgument, but the message is not for namespaces",
+			input:                  status.Error(codes.InvalidArgument, "blurg resource not found"),
+			expectMissingNamespace: false,
+		},
+		{
+			name:                   "namespace is missing",
+			input:                  status.Error(codes.InvalidArgument, "namespace resource not found"),
+			expectMissingNamespace: true,
+		},
```
wilkermichael, line 434: Might as well also cover the case where error is nil

control-plane/connect-inject/controllers/pod/pod_controller.go
wilkermichael: Should we add a comment to that effect? That way someone doesn't come through here and "optimize" not realising your intent.

wilkermichael (approved):
Looks great Dan!

Love all the retries 😍 

Just some minor comments/questions

control-plane/connect-inject/common/common.go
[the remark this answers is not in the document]
zalimeni: > Do we need to verify the message? Or would checking for the InvalidArgument code be sufficient?

I wouldn't rely just on that code in particular to identify "missing namespace" since it can show up in a number of cases - though TBF, delete is more limited than most. We could end up misreporting the actual reason for failure instead of propagating the underlying error.

Though, as a side note, it's probably worth including the original error message in the log @DanStough so we can at least catch a misidentification bug if it occurs.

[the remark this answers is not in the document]
DanStough: ❤️ 
* Add original error message
* Add integration test with the server to test the message.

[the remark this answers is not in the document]
DanStough: @zalimeni added the Int test. Let me know if that's what you were thinking. We would also catch it on acceptance tests if Consul changed.

control-plane/connect-inject/common/common_test.go
```diff
@@ -399,3 +407,104 @@ func TestGetPortProtocol(t *testing.T) {
...
+	id := &pbresource.ID{
+		Name: "foo",
+		Type: &pbresource.Type{
+			Group:        "catalog",
+			GroupVersion: "v1alpha1",
+			Kind:         "Workload",
+		},
+		Tenancy: &pbresource.Tenancy{
+			Partition: constants.DefaultConsulPartition,
+			Namespace: "i-dont-exist-but-its-ok-we-will-meet-again-someday",
```
zalimeni, line 474: 😂

control-plane/connect-inject/common/common_test.go
```diff
@@ -399,3 +407,104 @@ func TestGetPortProtocol(t *testing.T) {
...
+		Data: data,
+	}
+
+	_, err = resourceClient.Write(context.Background(), &pbresource.WriteRequest{Resource: resource})
+	require.Error(t, err)
+
+	s, ok := status.FromError(err)
+	require.True(t, ok)
+	require.Equal(t, codes.InvalidArgument, s.Code())
+	require.Contains(t, s.Message(), "namespace resource not found")
```
zalimeni, line 509: Love this test! Only suggestion I have is to also assert `ConsulNamespaceIsNotFound` returns true here

zalimeni (approved):

thisisnotashwin (approved):

zalimeni (approved):

PR #2961, opened 2023-09-14
[NET-5017] APIGW Status Conditions for RouteAuthFilter and Routes wrt JWT

jm96441n:
Changes proposed in this PR:
- Adds status conditions for RouteAuthFilter and Routes for JWT validation

How I've tested this PR:
1. after that finishes run `kubectl describe http-route http-route` to see the invalid accepted status on the route for the missing external filter
1. run `kubectl apply ./consul/httprouteauthfilter-okta.yaml`
1. after that finishes run `kubectl describe http-route http-route` to see the invalid accepted status on the route for the auth filter for the missing JWT reference
1. run `kubectl describe routeauthfilter auth-filter-okta` to the invalid stasus for the missing jwt provider
1. run `kubectl describe http-route http-route`  again to see the route back in a healthy status
1. run `kubectl describe routeauthfilter auth-filter-okta`  again to see the filter back in a healthy status

Read the code
Run the above steps

Checklist:
- [X] Tests added


control-plane/api/v1alpha1/gatewaypolicy_types.go
```diff
@@ -131,5 +131,5 @@ type GatewayPolicyStatus struct {
 	// +listMapKey=type
 	// +kubebuilder:validation:MaxItems=8
 	// +kubebuilder:default={{type: "Accepted", status: "Unknown", reason:"Pending", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"},{type: "ResolvedRefs", status: "Unknown", reason:"Pending", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}
-	Conditions []metav1.Condition `json:"conditions,omitemtpy"`
```
jm96441n, line 134: liiiiiittle bit of a typo

control-plane/api-gateway/binding/route_binding.go
```diff
@@ -177,6 +178,17 @@ func (r *Binder) bindRoute(route client.Object, boundCount map[gwv1beta1.Section
 				})
 			}
 
+			if authFilterReferencesMissingJWTProvider(httproute, r.config.Resources) {
+				results = append(results, parentBindResult{
```
jm96441n, line 182: this feels weird to have on the parent bind result, but we put the other statuses for external filters there and we don't have really any other place to put a status like this

andrewstucki: Yep, this is one of the things that I'm not a fan of with the spec, if the status of a route has nothing to do with a particular gateway (though that's how it's related to our controller), you still wind up scoping the status to the given gateway, so a filter missing and multiple parent gateways == multiple statuses saying the same thing, "the filter is missing". We already have this issue with backendRefs too... but frankly it's dumb.

andrewstucki: Acually, just wondering... is this the status for the filter not having a referenced JWT provider on it? If so, couldn't you just store the status of that missing provider on the filter itself?

jm96441n: so we do both: on the route we have a status that just tells the user that one of their auth filters is  referencing an invalid JWT provider and then on the filter itself is where we say what providers are missing

andrewstucki: Got it, that makes sense. Haven't looked too deeply into what the actual message used on the Route itself is, but I think it'd be fine to do something kind of generic -- like `"filter 'default/myFilter' is invalid"` or something, but I'm cool if it's more specific too

jm96441n: yeah right now the message is `externalRef references missing JWT provider` but I can make it a bit more generic

jm96441n: updated the message to be more generic

charts/consul/templates/crd-gatewaypolicies.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.3
+    controller-gen.kubebuilder.io/version: v0.8.0
```
nathancoleman, line 7: Is this intentional? I suspect it's a change due to the version of tooling installed locally

control-plane/api-gateway/binding/result.go
```diff
@@ -36,6 +36,7 @@ var (
 	errRouteNoMatchingParent                = errors.New("no matching parent")
 	errInvalidExternalRefType               = errors.New("invalid externalref filter kind")
 	errExternalRefNotFound                  = errors.New("ref not found")
+	errFilterInvalid         = errors.New("filter invalid")
```
nathancoleman, line 39: gofmt

control-plane/api-gateway/binding/validation.go
```diff
@@ -564,6 +564,51 @@ func externalRefsOnRouteAllExist(route *gwv1beta1.HTTPRoute, resources *common.R
 	return true
 }
 
+func checkIfReferencesMissingJWTProvider(filter gwv1beta1.HTTPRouteFilter, resources *common.ResourceMap, namespace string, invalidFilters map[string]struct{}) {
+	if filter.Type == gwv1beta1.HTTPRouteFilterExtensionRef {
```
nathancoleman, line 568: #nit I generally prefer early returns to avoid nesting the whole function body

control-plane/api-gateway/binding/validation.go
```diff
@@ -564,6 +564,51 @@ func externalRefsOnRouteAllExist(route *gwv1beta1.HTTPRoute, resources *common.R
...
+		}
+		authFilter, ok := externalFilter.(*v1alpha1.RouteAuthFilter)
+		if !ok {
+			return
+		}
+
+		for _, provider := range authFilter.Spec.JWT.Providers {
+			_, ok := resources.GetJWTProviderForGatewayJWTProvider(provider)
+			if !ok {
+				invalidFilters[fmt.Sprintf("%s/%s", authFilter.Name, namespace)] = struct{}{}
```
nathancoleman, line 581: We generally go `<namespace>/<resourcename>` since the namespace is the outer scope

control-plane/api-gateway/binding/validation.go
```diff
@@ -564,6 +564,51 @@ func externalRefsOnRouteAllExist(route *gwv1beta1.HTTPRoute, resources *common.R
...
+		}
+	}
+
+	invalidList := make([]string, 0, len(invalidFilters))
+
+	for name := range invalidFilters {
+		invalidList = append(invalidList, name)
+	}
+
+	return invalidList
```
nathancoleman, line 609: Can just `return maps.Keys(invalidFilters)` ([docs](https://pkg.go.dev/golang.org/x/exp/maps#Keys))

control-plane/config/crd/bases/consul.hashicorp.com_routeauthfilters.yaml
```diff
@@ -6,7 +6,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.11.3
```
nathancoleman, line 9: weird that this is the inverse of the change I commented on earlier 😕

nathancoleman (changes requested):

charts/consul/templates/crd-gatewaypolicies.yaml
jm96441n: yeah it is, there were changes added to main that specify the version of this tool to be `v0.8.0` that were pulled in on a rebase, my version previously was 0.11.3

control-plane/api-gateway/binding/result.go
jm96441n: I really gotta figure out why my editor has been doing this format after save recently rather than before save

control-plane/api-gateway/binding/validation.go
jm96441n: TIL

control-plane/api/v1alpha1/routeauthfilter_types.go
```diff
@@ -26,8 +26,8 @@ type RouteAuthFilter struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   RouteAuthFilterSpec `json:"spec,omitempty"`
-	Status `json:"status,omitempty"`
+	Spec   RouteAuthFilterSpec   `json:"spec,omitempty"`
+	Status RouteAuthFilterStatus `json:"status,omitempty"`
```
sarahalsmiller, line 30: Is there a reason to have a specific Status type here? The other types in consul all seem to use the common corev1.Status struct.

sarahalsmiller, line 30:
```suggestion
	Status            `json:"status,omitempty"`
```

Is how I see it defined for the other CRDs

control-plane/api-gateway/binding/validation.go
```diff
@@ -644,6 +689,35 @@ func routeKindIsAllowedForListenerExplicit(allowedRoutes *gwv1alpha2.AllowedRout
...
+
+	for _, filter := range authFilters {
+		if filter == nil {
+			continue
+		}
+		var result authFilterValidationResult
+		missingJWTProviders := make([]string, 0)
+		for _, provider := range filter.Spec.JWT.Providers {
+			if _, ok := resources.GetJWTProviderForGatewayJWTProvider(provider); !ok {
+				missingJWTProviders = append(missingJWTProviders, provider.Name)
```
nathancoleman, line 703: Does it make sense to include the namespace here in case the user has providers with the same name in different namespaces?

control-plane/config/crd/bases/consul.hashicorp.com_routeauthfilters.yaml
jm96441n: missed this during the rebasing

control-plane/api-gateway/binding/validation.go
jm96441n: jwt providers are partition scoped so namespace shouldn't matter here

control-plane/api-gateway/binding/route_binding.go
```diff
@@ -177,6 +181,17 @@ func (r *Binder) bindRoute(route client.Object, boundCount map[gwv1beta1.Section
 				})
 			}
 
+			if invalidFilterNames := authFilterReferencesMissingJWTProvider(httproute, r.config.Resources); len(invalidFilterNames) > 0 {
+				results = append(results, parentBindResult{
+					parent: ref,
+					results: []bindResult{
+						{
+							err: fmt.Errorf("%w: %s", errFilterInvalid, strings.Join(invalidFilterNames, ",")),
```
nathancoleman, line 189: It looks like this code path doesn't have any test coverage. If that's true, can you add coverage in this PR or queue up a ticket to add test coverage in the next couple of weeks? I know we're on a tight timeline

control-plane/api/v1alpha1/routeauthfilter_types.go
jm96441n: that's true, but for the gateway CRDs we include the `generation` in the condition which is not part of our `Status` type (and this follows in line with how the other gateway types handle status in the sig https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/gateway_types.go#L45, https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/httproute_types.go#L43

charts/consul/templates/crd-gatewaypolicies.yaml
nathancoleman: TIL

control-plane/api-gateway/binding/route_binding.go
jm96441n: good catch I can add some to cover this path to https://github.com/hashicorp/consul-k8s/blob/main/control-plane/api-gateway/binding/binder_test.go

nathancoleman (approved):

control-plane/api-gateway/binding/route_binding.go
nathancoleman: @jm96441n queued up a ticket

PR #2962, opened 2023-09-14
APIGW NS JWT Auth

jm96441n:
Changes proposed in this PR:
- Enable NS JWT authn/authz for API Gateway

How I've tested this PR:
Running through configurations in this repo: https://github.com/jm96441n/consul-experiments/tree/main/k8s/jwts

Read the PR

Checklist:
- [X] Tests added


charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -30,11 +30,14 @@ rules:
   - controlplanerequestlimits
   - routeretryfilters
   - routetimeoutfilters
+  - httprouteauthfilters
```
jm96441n, line 33: this should be `routeauthfilters`

charts/consul/test/unit/crd-routeauthfilters.bats
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "httproute-auth-filters/CustomResourceDefinition: enabled by default" {
```
jm96441n, line 5: these names should probably just be `routeauth-filters`

control-plane/api-gateway/binding/result.go
```diff
@@ -382,59 +386,78 @@ func (l listenerValidationResult) conflictedCondition(generation int64) metav1.C
...
+			errListenerInvalidCertificateRef_NonFIPSRSAKeyLen,
+			errListenerInvalidCertificateRef_FIPSRSAKeyLen:
+			conditions = append(conditions, metav1.Condition{
+				Type:               "ResolvedRefs",
+				Status:             metav1.ConditionFalse,
+				Reason:             "InvalidCertificateRef",
+				ObservedGeneration: generation,
+				Message:            refErr.Error(),
+				LastTransitionTime: now,
+			})
```
t-eckert, line 419: Not for this PR, but something to consider looping back around on: maybe we should have a function that creates `metav1.Condition` objects with the current timestamp given how much of this file is defining those objects. Then we could reduce the amount of code by having a function call that builds the condition.

```go
func newCondition(type string, status metav1.Status, reason string, generation int64, message: string) metav1.Condition {
    return metav1.Condition{
		Type:               type,
		Status:             status,
		Reason:             reason,
		ObservedGeneration: generation,
		Message:            message,
		LastTransitionTime: timeFunc(),
	}
}
```

t-eckert (approved):
Spectacular work! I'm so happy to see this going into `main`!

PR #2968, opened 2023-09-15
Fix endpoints controller v2 tests

zalimeni:
The missing constant somehow slipped by in CI. Are we not running v2 tests? Fixing first before digging.

Also fixes test that is breaking due to a protocol defaulting fix in Consul merged today.


DanStough (approved):

zalimeni:
https://github.com/hashicorp/consul-k8s-workflows/actions/runs/6202272584/job/16840895107#step:11:1258 <- this should not have passed (or compiled). Wondering whether we're caching PR branches in `consul-k8s-workflows`... Will look more Monday.

PR #2969, opened 2023-09-15
fix(control-plane): Only register v2 service when pods injected

zalimeni:
We are currently registering all services in k8s regardless of whether they represent mesh-injected workloads. This is both creating "junk" registrations for Consul and k8s components, but additionally can create issues in Consul core when generating routes with TProxy enabled, since these services will not have endpoints.

To solve for both of these issues, selectively sync k8s services to Consul in the v2 Endpoints controller only when at least one of its pods is injected.

Follow-up work will address edge cases where we want to maintain a service entry even without workloads, such as when the global inject flag is set, and when a service temporarily loses endpoints but is already registered and part of the mesh.

Changes proposed in this PR:
- Selectively register services based on whether pods are mesh-injected
- Move helper for checking v2 pod inject status to common

How I've tested this PR: Updated and added tests

Checklist:
- [x] Tests added


ishustava (approved):

zalimeni:
Rebased to resolve merge conflict in commented-out test; no functional changes.

PR #2977, opened 2023-09-20
Mw/net 5340 add support for explicit upstreams

wilkermichael:
Changes proposed in this PR:
*Note to Reviewer* I cleaned up the commits, so see each commit for more details in the tab above.

This PR adds support for V2 pod annotations (in both cases the order matters):

## Labeled Style: 

[service-port-name].port.[service-name].svc.[service-namespace].ns.[service-dc].dc:[bind-port]
eg. `myPort.port.mySVC.svc.ns2.ns.dc1.dc:1234`

## Unlabeled Style:

[service-port-name].[service-name].[service-namespace].[service-partition]:[bind-port]:[optional datacenter]
eg. `myPort.mySVC.ns2.dc1:1234`

Since we will not be supporting dcs and peers at first in the annotations, there is a final commit that addresses this and adds errors if those keys are used.

How I've tested this PR:
- unit and integration tests

👀 

Checklist:
- [x] Tests added

### DIFF
--- a/control-plane/connect-inject/controllers/pod/pod_controller.go
+++ b/control-plane/connect-inject/controllers/pod/pod_controller.go
@@ -498,7 +498,7 @@ func (r *Controller) processUpstreams(pod corev1.Pod) (*pbmesh.Upstreams, error)
 			labeledFormat := false
 			keys := []string{"port", "svc", "ns", "ap", "peer", "dc"}
 			for _, v := range keys {
-				if strings.Contains(raw, fmt.Sprintf(".%s.", v)) {
+				if strings.Contains(raw, fmt.Sprintf(".%s.", v)) || strings.Contains(raw, fmt.Sprintf(".%s:", v)) {
 					labeledFormat = true
 					break
 				}
@@ -511,7 +511,11 @@ func (r *Controller) processUpstreams(pod corev1.Pod) (*pbmesh.Upstreams, error)
 					return &pbmesh.Upstreams{}, err
 				}
 			} else {
-				upstream = r.processUnlabeledUpstream(pod, raw)
+				var err error
+				upstream, err = r.processUnlabeledUpstream(pod, raw)
+				if err != nil {
+					return &pbmesh.Upstreams{}, err
+				}
 			}
 
 			upstreams.Upstreams = append(upstreams.Upstreams, upstream)
@@ -526,6 +530,7 @@ func (r *Controller) processUpstreams(pod corev1.Pod) (*pbmesh.Upstreams, error)
 // [service-port-name].port.[service-name].svc.[service-namespace].ns.[service-partition].ap:[port]
 // [service-port-name].port.[service-name].svc.[service-namespace].ns.[service-datacenter].dc:[port].
 // peer/ap/dc are mutually exclusive. At minimum service-port-name and service-name are required.
+// TODO: enable dc and peer support when ready, currently return errors if set.
 func (r *Controller) processLabeledUpstream(pod corev1.Pod, rawUpstream string) (*pbmesh.Upstream, error) {
 	parts := strings.SplitN(rawUpstream, ":", 3)
 	var port int32
@@ -541,11 +546,15 @@ func (r *Controller) processLabeledUpstream(pod corev1.Pod, rawUpstream string) 
 			end := strings.TrimSpace(pieces[7])
 			switch end {
 			case "peer":
-				peer = strings.TrimSpace(pieces[6])
+				// TODO: uncomment and remove error when peers supported
+				//peer = strings.TrimSpace(pieces[6])
+				return &pbmesh.Upstream{}, fmt.Errorf("upstream currently does not support peers: %s", rawUpstream)
 			case "ap":
 				partition = strings.TrimSpace(pieces[6])
 			case "dc":
-				datacenter = strings.TrimSpace(pieces[6])
+				// TODO: uncomment and remove error when datacenters are supported
+				//datacenter = strings.TrimSpace(pieces[6])
+				return &pbmesh.Upstream{}, fmt.Errorf("upstream currently does not support datacenters: %s", rawUpstream)
 			default:
 				return &pbmesh.Upstream{}, fmt.Errorf("upstream structured incorrectly: %s", rawUpstream)
 			}
@@ -577,13 +586,18 @@ func (r *Controller) processLabeledUpstream(pod corev1.Pod, rawUpstream string) 
 			end := strings.TrimSpace(pieces[5])
 			switch end {
 			case "peer":
-				peer = strings.TrimSpace(pieces[4])
+				// TODO: uncomment and remove error when peers supported
+				//peer = strings.TrimSpace(pieces[4])
+				return &pbmesh.Upstream{}, fmt.Errorf("upstream currently does not support peers: %s", rawUpstream)
 			case "dc":
-				datacenter = strings.TrimSpace(pieces[4])
+				// TODO: uncomment and remove error when datacenter supported
+				//datacenter = strings.TrimSpace(pieces[4])
+				return &pbmesh.Upstream{}, fmt.Errorf("upstream currently does not support datacenters: %s", rawUpstream)
 			default:
 				return &pbmesh.Upstream{}, fmt.Errorf("upstream structured incorrectly: %s", rawUpstream)
 			}
-			fallthrough
+			// TODO: uncomment and remove error when datacenter and/or peers supported
+			//fallthrough
 		case 4:
 			if strings.TrimSpace(pieces[3]) == "svc" {
 				svcName = strings.TrimSpace(pieces[2])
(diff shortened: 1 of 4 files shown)


wilkermichael:
@DanStough I assume we're doing one big changelog at the end? So I didn't add a changelog, but let me know if I assumed incorrectly.

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -432,10 +445,266 @@ func (r *Controller) writeHealthStatus(ctx context.Context, pod corev1.Pod) erro
...
+					PeerName:  peer,
+				},
+				Name: svcName,
+			},
+			DestinationPort: portName,
+			Datacenter:      datacenter,
+			ListenAddr: &pbmesh.Upstream_IpPort{
+				IpPort: &pbmesh.IPPortAddress{
+					Port: uint32(port),
+				},
```
DanStough, line 638: I *think* we'll need an address here. I'm guessing loopback `127.0.0.1`.

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -432,10 +445,266 @@ func (r *Controller) writeHealthStatus(ctx context.Context, pod corev1.Pod) erro
...
+	}
+
+	var upstream pbmesh.Upstream
+	if port > 0 {
+		upstream = pbmesh.Upstream{
+			DestinationRef: &pbresource.Reference{
+				Tenancy: &pbresource.Tenancy{
+					Partition: partition,
+					Namespace: namespace,
+					PeerName:  peer,
```
DanStough, line 629: Since Consul V2 is weird about defaults, we'll want to check if these are empty and set them explicitly to constant values.

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -432,10 +445,266 @@ func (r *Controller) writeHealthStatus(ctx context.Context, pod corev1.Pod) erro
...
+			} else {
+				return &pbmesh.Upstream{}, fmt.Errorf("upstream structured incorrectly: %s", rawUpstream)
+			}
+		default:
+			return &pbmesh.Upstream{}, fmt.Errorf("upstream structured incorrectly: %s", rawUpstream)
+		}
+	}
+
+	var upstream pbmesh.Upstream
+	if port > 0 {
```
DanStough, line 623: ~ We can probably check this much earlier and bail with an error.

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -432,10 +445,266 @@ func (r *Controller) writeHealthStatus(ctx context.Context, pod corev1.Pod) erro
...
+		upstream = pbmesh.Upstream{
```
DanStough, line 624: We'll also need a `Type` as part of the reference. This should be `catalog.v1alpha1.Service`.

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -432,10 +445,266 @@ func (r *Controller) writeHealthStatus(ctx context.Context, pod corev1.Pod) erro
...
+	return &upstream, nil
+}
 
-//func (r *Controller) deleteUpstreams(pod corev1.Pod) error
+// processUnlabeledUpstream processes an upstream in the format:
+// [service-port-name].[service-name].[service-namespace].[service-partition]:[port]:[optional datacenter].
+// There is no unlabeled field for peering.
+// TODO: enable dc and peer support when ready, currently return errors if set. We also most likely won't need to return an error at all.
+func (r *Controller) processUnlabeledUpstream(pod corev1.Pod, rawUpstream string) (*pbmesh.Upstream, error) {
+	var portName, datacenter, svcName, namespace, partition, peer string
```
DanStough, line 651: I think like your other edits in V1, `peer` isn't supported.

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -432,10 +445,266 @@ func (r *Controller) writeHealthStatus(ctx context.Context, pod corev1.Pod) erro
...
+			Datacenter:      datacenter,
+			ListenAddr: &pbmesh.Upstream_IpPort{
+				IpPort: &pbmesh.IPPortAddress{
+					Port: uint32(port),
+				},
+			},
+		}
+	}
+	return &upstream, nil
+}
```
DanStough, line 707: Same comments here as the labeled-style:
1. need explicit tenenacy
2. need Type in DestinationRef
3. need IP in the listener

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -761,11 +763,1161 @@ func TestProxyConfigurationDelete(t *testing.T) {
...
-// TODO
-// func TestUpstreamsDelete(t *testing.T)
+	const podName = "pod1"
+
+	cases := []struct {
+		name                    string
+		pod                     func() *corev1.Pod
+		expected                *pbmesh.Upstreams
+		expErr                  string
+		configEntry             func() api.ConfigEntry
```
DanStough, line 778: Is this field used in any of the tests? The V2 changes shouldn't respond to any V1 config entries.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -761,11 +763,1161 @@ func TestProxyConfigurationDelete(t *testing.T) {
...
+		consulUnavailable       bool
```
DanStough, line 779: Same question here. I didn't see anywhere in the test that consumes this.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -1446,6 +2664,34 @@ func createProxyConfiguration() *pbmesh.ProxyConfiguration {
...
+		Workloads: &pbcatalog.WorkloadSelector{
+			Names: []string{"foo"},
+		},
+		Upstreams: []*pbmesh.Upstream{
+			{
+				DestinationRef: &pbresource.Reference{
+					Tenancy: &pbresource.Tenancy{
+						Partition: "",
+						Namespace: "",
+						PeerName:  "",
```
DanStough, line 2679: These should be explicit all the time. Part of this is historical that Consul V2 didn't default them, but if you set one of them, the others won't correctly inherit the default.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -761,11 +763,1161 @@ func TestProxyConfigurationDelete(t *testing.T) {
...
+			//					Port: uint32(1234),
+			//				},
+			//			},
+			//		},
+			//	},
+			//},
+			consulUnavailable:       true,
+			consulNamespacesEnabled: false,
+			consulPartitionsEnabled: false,
+		},
```
DanStough, line 1357: I suspect this test case doesn't really function as expected for V2.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -761,11 +763,1161 @@ func TestProxyConfigurationDelete(t *testing.T) {
...
+			consulNamespacesEnabled: true,
+			consulPartitionsEnabled: false,
+		},
+		{
+			name: "labeled multiple annotated upstreams",
+			pod: func() *corev1.Pod {
+				pod1 := createPod(podName, "1.2.3.4", "foo", true, true)
+				pod1.Annotations[constants.AnnotationMeshDestinations] = "myPort.port.upstream1.svc.ns1.ns.dc1.dc:1234, myPort2.port.upstream2.svc:2234, myPort3.port.upstream3.svc.ns1.ns:3234, myPort4.port.upstream4.svc.ns1.ns.peer1.peer:4234"
+				return pod1
+			},
```
DanStough, line 1234: I think we should have one functional case of multiple labeled upstream. Better yet, one that has both labeled and unlabeled would also be sufficient.

DanStough (changes requested):
I think there are few critical fixes with the tenancy and reference-type, but overall this looks amazing and it's 98% there! 🚀 💨

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -146,10 +151,10 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		}
 
-		// TODO: create explicit upstreams
-		//if err := r.writeUpstreams(ctx, pod); err != nil {
-		//	errs = multierror.Append(errs, err)
-		//}
+		// Create explicit upstreams (if any exist)
+		if err := r.writeUpstreams(ctx, pod); err != nil {
+			errs = multierror.Append(errs, err)
+		}
```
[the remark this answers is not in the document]
wilkermichael: ah yes I did! I noticed it after

control-plane/connect-inject/controllers/pod/pod_controller.go
wilkermichael: yeah, a lot of it is from existing code I didn't modify more than I needed to for the most part.

But this is a good callout, I'll move it

control-plane/connect-inject/controllers/pod/pod_controller_test.go
wilkermichael: Doesn't look like it, copied it over from the old test

control-plane/connect-inject/controllers/pod/pod_controller_test.go
wilkermichael: nope, not used

control-plane/connect-inject/controllers/pod/pod_controller_test.go
wilkermichael: I'll remove it

control-plane/connect-inject/controllers/pod/pod_controller_test.go
wilkermichael: oh yeah, I guess those did get removed with the ap/dcs being removed

DanStough (approved):
Bellissimo!

PR #2978, opened 2023-09-20
[NET-5674] v2: Conditional target port when numeric in k8s

zalimeni:
When choosing a Consul service TargetPort (name) in the v2 Endpoints Controller, attempt to find the best match among endpoint container ports when the K8s service target port value is a number.

This bridges an existing gap between Consul (which always expects a Workload port to be named), and K8s (where container ports need not be named, and names are ignored when the K8s service targets by number).

This change will be mostly reverted in a future release once Consul's v2 data model allows for target ports to be a name or number in alignment w/ K8s behavior: [NET-5585](https://hashicorp.atlassian.net/browse/NET-5585)

Changes proposed in this PR:
- Conditionally choose Consul target port based on container port names when K8s target port is numeric.
- Small `nil` safety check for mesh-inject detection.

How I've tested this PR:
* Added unit tests ✅ 
* Run new v2 mesh acceptance test w/ numeric target ports to verify bug and fix ✅ 

Checklist:
- [x] Tests added

[NET-5585]: https://hashicorp.atlassian.net/browse/NET-5585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/acceptance/tests/fixtures/bases/v2-multiport-app/service.yaml
+++ b/acceptance/tests/fixtures/bases/v2-multiport-app/service.yaml
@@ -11,8 +11,8 @@ spec:
   ports:
     - name: web
       port: 8080
-      # TODO: (v2/nitya) ensure this works with numeric target ports also once support for that is merged.
       targetPort: web
     - name: admin
       port: 9090
-      targetPort: admin
+      # Test with a mix of named and numeric target ports.
+      targetPort: 9090
--- a/control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
@@ -15,6 +15,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -60,7 +61,6 @@ func (r *Controller) SetupWithManager(mgr ctrl.Manager) error {
 // Reconcile reads the state of an Endpoints object for a Kubernetes Service and reconciles Consul services which
 // correspond to the Kubernetes Service. These events are driven by changes to the Pods backing the Kube service.
 func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	var errs error
 	var endpoints corev1.Endpoints
 	var service corev1.Service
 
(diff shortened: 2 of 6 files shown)


control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -96,38 +96,74 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-			if common.ConsulNamespaceIsNotFound(err) {
-				r.Log.Info("Consul namespace not found; re-queueing request",
-					"service", service.GetName(), "ns", req.Namespace,
-					"consul-ns", r.getConsulNamespace(req.Namespace), "err", err.Error())
-				return ctrl.Result{Requeue: true}, nil
-			}
-			errs = multierror.Append(errs, err)
+	if consulSvc.Workloads == nil {
+		return ctrl.Result{}, nil
+	}
```
zalimeni, line 111: This replaces the check above - where we used to create just the workload selector here and let the `registerService` call create the service for us, we now do the heavy lifting inside `getConsulService `.

The surrounding code in this `Reconcile` func is effectively unchanged.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -96,38 +96,74 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	return ctrl.Result{}, nil
 }
 
-// getWorkloadSelectorFromEndpoints calculates a Consul service WorkloadSelector from Endpoints based on pod names and
-// owners.
-func (r *Controller) getWorkloadSelectorFromEndpoints(ctx context.Context, pf PodFetcher, endpoints *corev1.Endpoints) (*pbcatalog.WorkloadSelector, error) {
-	podPrefixes := make(map[string]any)
-	podExactNames := make(map[string]any)
+func (r *Controller) getConsulService(ctx context.Context, pf PodFetcher, service corev1.Service, endpoints corev1.Endpoints) (*pbcatalog.Service, error) {
+	prefixedPods, exactNamePods, err := r.getWorkloadDataFromEndpoints(ctx, pf, endpoints)
```
zalimeni, line 131: We now use these two compound data sets for both `WorkloadSelector` creation and target port setting below. Previously, they were mapped directly into a `WorkloadSelector` and nothing else.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -96,38 +96,74 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+type podSetData struct {
+	podCount  int
+	samplePod *corev1.Pod
+}
+
+// selectorPodData represents data for each set of pods represented by a WorkloadSelector value.
+// The data may be for several pods (prefix) or a single pod (exact name).
+// This is used for choosing the ideal Consul service TargetPort value when the K8s service target port is numeric.
+type selectorPodData map[string]*podSetData
```
zalimeni, line 152: Added a type for this to simplify passing it around and make the code more legible.

Below, we use the `podSetData` struct to accumulate information about the pod count and container port name (values), in addition to prefix and exact name sets (keys) for the `WorkloadSelector`.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -285,6 +340,102 @@ func getServicePorts(service corev1.Service) []*pbcatalog.ServicePort {
 	return ports
 }
 
+func getEffectiveTargetPort(targetPort intstr.IntOrString, prefixedPods selectorPodData, exactNamePods selectorPodData) string {
```
zalimeni, line 343: This func attempts to encapsulate the complicated edge cases for picking a target port name from container ports. The happy path is the early return directly below when the K8s service target port is a string, which will be used without modification.

thisisnotashwin (approved):
Whew!! For the fact that the logic for computing the target port is pretty convoluted, this code made it pretty easy to parse! great job here!!

ndhanushkodi (approved):
Were you thinking to also change the acceptance test based on this to have one string/one numeric port on the multiport service? Totally fine for that to come separately or if you think that is covered in the unit tests fine! I thoroughly reviewed the code but not the endpoints controller test, but giving it a thumbs up since you have another approval anyways!

Also thanks for chatting with me through port things!

zalimeni:
Thanks @ndhanushkodi ! 

> Were you thinking to also change the acceptance test based on this to have one string/one numeric port on the multiport service? Totally fine for that to come separately or if you think that is covered in the unit tests fine!

The unit tests should cover this, but yes I'm more than happy to update the acceptance test too now that it's merged (I missed that)! Pushed and rebased w/ the update.

acceptance/tests/fixtures/bases/v2-multiport-app/service.yaml
```diff
@@ -11,8 +11,8 @@ spec:
   ports:
     - name: web
       port: 8080
-      # TODO: (v2/nitya) ensure this works with numeric target ports also once support for that is merged.
       targetPort: web
     - name: admin
       port: 9090
-      targetPort: admin
+      # Test with a mix of named and numeric target ports.
+      targetPort: 9090
```
zalimeni, line 18: @ndhanushkodi LMK if this doesn't look right to you and I'll adjust

ndhanushkodi: looks perfect!

zalimeni:
Added a comment to the container port name in fixture to call out it needs to be set.

PR #2985, opened 2023-09-21
[NET-4144] add gateway pdb

kkavish:
Changes proposed in this PR:
- Add PDB for all gateways
- 

How I've tested this PR:
manually

- Deploy consul k8s cluster using kind with gateways, check kubectl get poddisruptionbudget -n <namespace> for PDBs
- Unit tests

Checklist:
- [x] Tests added


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Unplanned Release
+
+BUG FIXES:
```
Ganeshrockz, line 3: I don't think you should update this file. Adding `.changelog/*.txt` should be suffice.

.changelog/2985.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
Ganeshrockz, line 1: Is this a bug? Should we call it as an enhancement?

charts/consul/values.yaml
```diff
@@ -66,7 +66,7 @@ global:
   # image: "hashicorp/consul-enterprise:1.10.0-ent"
   # ```
   # @default: hashicorp/consul:<latest version>
-  image: docker.mirror.hashicorp.services/hashicorppreview/consul:1.17-dev
+  image: docker.mirror.hashicorp.services/hashicorppreview/consul:1.16.1
```
Ganeshrockz, line 69: I guess this should still be `1.17-dev`

charts/consul/values.yaml
```diff
@@ -2272,7 +2272,7 @@ connectInject:
     # If true, then all traffic redirection setup uses the consul-cni plugin.
     # Requires connectInject.enabled to also be true.
     # @type: boolean
-    enabled: false
+    enabled: true
```
Ganeshrockz, line 2275: Any reason to enable CNI?

charts/consul/values.yaml
```diff
@@ -2138,7 +2138,7 @@ connectInject:
   # [injection annotation](https://developer.hashicorp.com/consul/docs/k8s/connect#consul-hashicorp-com-connect-inject)
   # to opt-in to Connect injection. If this is true, pods can use the same annotation
   # to explicitly opt-out of injection.
-  default: false
+  default: true
```
Ganeshrockz, line 2141: Is this change required?

charts/consul/values.yaml
```diff
@@ -3098,6 +3119,22 @@ terminatingGateways:
...
+    # Enables registering a PodDisruptionBudget for the terminating gateways
+    # If enabled, it only registers the budget so long as
+    # the terminating gateways is enabled. To disable, set to `false`.
+    enabled: true
+
+    # The maximum number of unavailable pods. By default, this will be
+    # automatically computed based on the `server.replicas` value to be `(n/2)-1`.
+    # If you need to set this to `0`, you will need to add a
+    # --set 'server.terminatingGateways.maxUnavailable=0'` flag to the helm chart installation
+    # command because of a limitation in the Helm templating language.
```
Ganeshrockz, line 3134: > command because of a limitation in the Helm templating language.

What limitation is this?

charts/consul/values.yaml
```diff
@@ -2753,6 +2753,27 @@ meshGateway:
     # DNS entry to point to your mesh gateways.
     static: ""
 
+  # This configures the [`PodDisruptionBudget`](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)
```
Ganeshrockz, line 2756: May be I am missing something, but I don't see a `disruptionBudget` stanza for `apiGateways` but it seems to be referenced in the helm template files. Is this intentional?

charts/consul/templates/crd-gateways-disruptionbudget.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if .Values.connectInject.enabled }}
```
Ganeshrockz, line 1: What are CRD gateways?

Ganeshrockz (commented):
Skimmed through the changes and have some basic questions. Will go through the PR again in detail once those queries get resolved.

charts/consul/templates/crd-gateways-disruptionbudget.yaml
kkavish: custom gateways,  name: gateways.gateway.networking.k8s.io

charts/consul/values.yaml
kkavish: please check line #3276, this is for consul api gateway, connect inject uses CRD gateway as per my understanding, it has its own default pdb yaml.

charts/consul/values.yaml
kkavish: ah, seems to be a copied comment not applicable here

CHANGELOG.md
kkavish: noted.

.changelog/2985.txt
kkavish: reported as a bug, I too had this confusion.

charts/consul/values.yaml
kkavish: yes, it was for local testing, reverting

charts/consul/values.yaml
kkavish: nope, it was for local testing, reverting

charts/consul/values.yaml
kkavish: nope, it was for local testing, reverting

charts/consul/values.yaml
Ganeshrockz: Ahh my bad 🤦 . The diff confused me. Thanks for pointing out

.changelog/2985.txt
Ganeshrockz: It could be an improvement

Ganeshrockz (commented):
Looks good mostly. Deferring the approval to the `service-mesh`/`gateways` team who owns the codebase.

.changelog/2985.txt
curtbushko: I'd call it an 'improvement' also.

kkavish: updated

mr-miles:
Sorry to drop in out of nowhere!  But any chance to add a pdb for the ingress gateways too?  I know ingress gateways themselves are deprecated for replacement by api-gateways, but they are still supported.  Its confusing if the gateways are inconsistent in their config, unless they're going to be removed from the chart completely.

kkavish:
> Sorry to drop in out of nowhere! But any chance to add a pdb for the ingress gateways too? I know ingress gateways themselves are deprecated for replacement by api-gateways, but they are still supported. Its confusing if the gateways are inconsistent in their config, unless they're going to be removed from the chart completely.

-  thanks for suggesting it, makes sense, we'd add it to ingress as well.

charts/consul/templates/crd-gateways-disruptionbudget.yaml
nathancoleman, line 1: Based on the resolved comment below, it seems like this is intended to target Consul API Gateway pods resulting from a user creating a `Gateway` resource like this:

```yaml
apiVersion: gateways.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: my-gateway
spec:
  gatewayClassName: nathan-gateway
```

This won't work in practice though and doesn't match how Gateways are structured.
Gateways specify a `GatewayClass`. For Consul, each `GatewayClass` has a `GatewaysClassConfig` (which we define -- [docs](https://developer.hashicorp.com/consul/docs/api-gateway/configuration/gatewayclassconfig)) tied to it that controls configuration specific to that `GatewayClass` ([docs](https://gateway-api.sigs.k8s.io/api-types/gatewayclass/)). You can have up to `n` GatewayClasses, each with their own GatewayClassConfig.

I'd expect this configuration to live on the `GatewayClassConfig` defined [here](https://github.com/hashicorp/consul-k8s/blob/012227d5f0ebf23f582e6f0301e73d161e1a3863/charts/consul/templates/crd-gatewayclassconfigs.yaml). The resulting GatewayClassConfig with support for `PodDisruptionBudget` might look like:

```yaml
kind: GatewayClassConfig
metadata:
  name: nathan-gateway
spec:
  podDisruptionBudget:
    maxUnavailable: 2
  ...
```

API Gateway resources -- such as Deployment, Service, ServiceAccount, etc. -- are then created in K8s dynamically when a user creates a `Gateway` resource. I'd expect the logic to create a `PodDisruptionBudget` to live in [the `gatekeeper` package](https://github.com/hashicorp/consul-k8s/tree/012227d5f0ebf23f582e6f0301e73d161e1a3863/control-plane/api-gateway/gatekeeper) in a new file called `poddisruptionbudget.go`.

charts/consul/values.yaml
```diff
@@ -2753,6 +2753,25 @@ meshGateway:
...
+    # The maximum number of unavailable pods. By default, minAvailable will be set to 1
+    # If you need to set this to some value, you will need to add a
+    # --set 'meshGateway.disruptionBudget.maxUnavailable=0'` flag to the helm chart installation
+    # @type: integer
+    maxUnavailable: null
+
+    # The minimum number of available pods.
+    # Takes precedence over maxUnavailable if set.
+    # @type: integer
+    minAvailable: null
```
nathancoleman, line 2773: I don't see this value actually being consumed anywhere. I'd expect we either:
1. allow the user to set `minAvailable` for every gateway type, or
2. never allow the user to set `minAvailable` for any gateway type

nathancoleman (changes requested):

control-plane/api-gateway/gatekeeper/poddisruptionbudget.go
```diff
@@ -0,0 +1,113 @@
...
+func (g *Gatekeeper) upsertPodDisruptionBudget(ctx context.Context, gateway gwv1beta1.Gateway, gcc v1alpha1.GatewayClassConfig, config common.HelmConfig) error {
+	if gcc.Spec.PodDisruptionBudgetSpec == nil || !gcc.Spec.PodDisruptionBudgetSpec.Enabled {
+		return g.deletePodDisruptionBudget(ctx, types.NamespacedName{Namespace: gateway.Namespace, Name: gateway.Name})
+	}
+
+	pdb := g.podDisruptionBudget(ctx, gateway, gcc, config)
+	mutated := pdb.DeepCopy()
+	mutator := newPodDisruptionBudgetMutator(pdb, mutated, gateway, g.Client.Scheme())
+
+	result, err := controllerutil.CreateOrUpdate(ctx, g.Client, mutated, mutator)
```
nathancoleman, line 27: Let's follow the pattern found in `role.go` where we check for an existing `PodDisruptionBudget` and, if one exists, verify that it's owned by our controller.

In the future, I'd like to use that pattern for all resources that we create, but that is outside the scope of this PR.

nathancoleman (changes requested):

control-plane/api-gateway/gatekeeper/poddisruptionbudget.go
kkavish: my thought process was:

The configs remain same for resources that we create and the resources created through templates. i.e. they all have - enable, minAvailable and maxUnavailable thus making them homogenous.
For the resources we create, if the end-user/dev needs to disable PDB she/he could just go ahead and make enabled = false instead of deleting the whole stanza and adding it back again when they require it.
Shall I go ahead and remove the enabled flag anyways?

control-plane/api/v1alpha1/api_gateway_types.go
```diff
@@ -65,6 +66,15 @@ type GatewayClassConfigSpec struct {
 
 	// The value to add to privileged ports ( ports < 1024) for gateway containers
 	MapPrivilegedContainerPorts int32 `json:"mapPrivilegedContainerPorts,omitempty"`
+
+	// The value to add pod disruption budgets, will prevent gateway pods from being evicted all at once.
+	PodDisruptionBudgetSpec *PodDisruptionBudgetSpec `json:"podDisruptionBudget,omitempty"`
+}
+
+type PodDisruptionBudgetSpec struct {
+	Enabled        bool                `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
kkavish: my thought process was:

1. The configs remain same for resources that we create and the resources created through templates. i.e. they all have - enable, minAvailable and maxUnavailable thus making them homogenous.
2. For the resources we create, if the end-user/dev needs to disable PDB she/he could just go ahead and make enabled = false instead of deleting the whole stanza and adding it back again when they require it.

Shall I go ahead and remove the enabled flag anyways?

david-yu:
@nathancoleman I chatted with @kkavish he said this might need re-review? However I'm not sure if it's ready given that it looks like most of the tests are still failing?

nathancoleman:
Re-reviewing is on my To Do list; however, it's lower in priority than work related to Consul 1.17.0 releasing next week. I confirmed with @kkavish that we're **not** trying to get this change into 1.17.0.

david-yu:
Sure not a rush. Let's revisit after 1.17, I also asked @kkavish to brush up this PR to get the tests to pass as well.

david-yu:
@kkavish Could you try re-basing to get the unit tests on control plane to pass?

david-yu:
I'm guess the PR is still failing to pass any acceptance tests?

kkavish:
> I'm guess the PR is still failing to pass any acceptance tests?

@david-yu sorry, I've been on a vacation, I'll try to get this fixed this week.

PR #2994, opened 2023-09-21
add health status syncing check in acceptance test

ndhanushkodi:
Changes proposed in this PR:
- ~~needed to tweak pod controller to report unhealthy if any container is unhealthy, and healthy only if all containers are healthy~~
- add health status check in acceptance test. This was failing before the pod controller fix and the consul fix here: https://github.com/hashicorp/consul/pull/18958

How I've tested this PR:

built images off of this branch and this consul branch with a fix in consul: https://github.com/hashicorp/consul/pull/18958 and the acceptance test passes.

Checklist:
- [x] Tests added

### DIFF
--- a/acceptance/tests-v2/mesh/mesh_inject_test.go
+++ b/acceptance/tests-v2/mesh/mesh_inject_test.go
@@ -38,7 +38,7 @@ func TestMeshInject_MultiportService(t *testing.T) {
 
 			helmValues := map[string]string{
 				"global.image":                "ndhanushkodi/consul-dev:multiport37",
-				"global.imageK8S":             "ndhanushkodi/consul-k8s-dev:multiport23",
+				"global.imageK8S":             "ndhanushkodi/consul-k8s-dev:multiport25",
 				"global.imageConsulDataplane": "hashicorppreview/consul-dataplane:1.3-dev",
 				"global.experiments[0]":       "resource-apis",
 				// The UI is not supported for v2 in 1.17, so for now it must be disabled.
--- a/acceptance/tests-v2/mesh/mesh_inject_test.go
+++ b/acceptance/tests-v2/mesh/mesh_inject_test.go
@@ -19,6 +19,8 @@ import (
 	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
 )
 
+const multiport = "multiport"
+
 // Test that mesh sidecar proxies work for an application with multiple ports. The multiport application is a Pod listening on
 // two ports. This tests inbound connections to each port of the multiport app, and outbound connections from the
 // multiport app to static-server.
@@ -35,8 +37,8 @@ func TestMeshInject_MultiportService(t *testing.T) {
 			ctx := suite.Environment().DefaultContext(t)
 
 			helmValues := map[string]string{
-				"global.image":                "ndhanushkodi/consul-dev:multiport36",
-				"global.imageK8S":             "ndhanushkodi/consul-k8s-dev:multiport20",
+				"global.image":                "ndhanushkodi/consul-dev:multiport37",
+				"global.imageK8S":             "ndhanushkodi/consul-k8s-dev:multiport23",
 				"global.imageConsulDataplane": "hashicorppreview/consul-dataplane:1.3-dev",
 				"global.experiments[0]":       "resource-apis",
 				// The UI is not supported for v2 in 1.17, so for now it must be disabled.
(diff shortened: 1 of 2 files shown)


thisisnotashwin (approved):

DanStough (approved):
Nice!

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -811,11 +811,12 @@ func getHealthStatusFromPod(pod corev1.Pod) pbcatalog.Health {
...
 	for _, condition := range pod.Status.Conditions {
-		if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
-			return pbcatalog.Health_HEALTH_PASSING
+		if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionFalse {
+			return pbcatalog.Health_HEALTH_CRITICAL
 		}
 	}
-	return pbcatalog.Health_HEALTH_CRITICAL
+
+	return pbcatalog.Health_HEALTH_PASSING
```
ishustava, line 819: Wouldn't it mean that if pod is not ready we'll return a passing health status? Also, are there unit tests for this that should be updated?

ndhanushkodi: We should return in line 815 if any of the ready conditions fail

I don't think there are unit tests but I can certainly add some for this function now!

ishustava: I'm also curious, are you seeing multiple pod ready status conditions per container? There should be only one per pod, right? So I'm curious why we needed that change.

ndhanushkodi: ya it's iterating through these pod.Status.Conditions here (this is the multiport pod during the acceptance test when we mark it unhealthy), so we're looking at anything with type: ready and then what the status is:
```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2023-09-21T20:13:01Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2023-09-21T20:13:37Z"
    message: 'containers with unready status: [multiport multiport-admin]'
    reason: ContainersNotReady
    status: "False"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: "2023-09-21T20:13:37Z"
    message: 'containers with unready status: [multiport multiport-admin]'
    reason: ContainersNotReady
    status: "False"
    type: ContainersReady
  - lastProbeTime: null
    lastTransitionTime: "2023-09-21T20:12:54Z"
    status: "True"
    type: PodScheduled

```

ndhanushkodi: on the unit tests- it does look like theres one failure that i need to fix so there might be some coverage already actually! will go back and update

ishustava: >We should return in line 815 if any of the ready conditions fail 

What if the pod doesn't have a ready condition yet for some reason?

>ya it's iterating through these pod.Status.Conditions here (this is the multiport pod during the acceptance test when we mark it unhealthy), so we're looking at anything with type: ready and then what the status is:

Hmm so there's only one Ready type condition which is false. Does the old code return that the health status is passing in this case? I'd expect it to return critical because it would iterate through all the conditions and would find none that are both type `Ready` and status `true` and so it'd return status critical by default. Maybe I'm missing something, but a unit test would probably help just to make sure we're covering all cases correctly.

ndhanushkodi: Ah I see what youre saying, I didn't realize there is only one ready status. let me try testing with just the consul fix and not the k8s changes here, and see if it still works. In either case, I can add unit tests

ndhanushkodi: I've reverted my changes to this function since it was correct in the first place. Thanks for catching @ishustava!

I'm seeing there is full unit test coverage of the health status behavior, its tested in TestHealthStatusWrite which calls out to writeHealthStatus which calls getHealthStatusFromPod. It checks all 3 cases- when ready condition is true, false, and nonexistent, and the logic is correct there. 

I didn't look closely enough at the health statuses to realize there was only one ready status, and the test passes without changes to pod controller and just the consul fixes!

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -815,6 +815,7 @@ func getHealthStatusFromPod(pod corev1.Pod) pbcatalog.Health {
 			return pbcatalog.Health_HEALTH_PASSING
 		}
 	}
+
```
zalimeni, line 818: ~ Can drop this line I think so this file goes totally unchanged
```suggestion
```

acceptance/tests-v2/mesh/mesh_inject_test.go
```diff
@@ -79,6 +81,22 @@ func TestMeshInject_MultiportService(t *testing.T) {
...
 			// Check connection from static-client to multiport-admin.
 			k8s.CheckStaticServerConnectionSuccessfulWithMessage(t, ctx.KubectlOptions(t), connhelper.StaticClientName, "hello world from 9090 admin", "http://multiport:9090")
+
+			// Test that kubernetes readiness status is synced to Consul. This will make the multi port pods unhealthy
+			// and check inbound connections to the multi port pods' services.
+			// Create the files so that the readiness probes of the multi port pod fails.
+			logger.Log(t, "testing k8s -> consul health checks sync by making the multiport unhealthy")
+			k8s.RunKubectl(t, ctx.KubectlOptions(t), "exec", "deploy/"+multiport, "-c", "multiport", "--", "touch", "/tmp/unhealthy-multiport")
+			logger.Log(t, "testing k8s -> consul health checks sync by making the multiport-admin unhealthy")
+			k8s.RunKubectl(t, ctx.KubectlOptions(t), "exec", "deploy/"+multiport, "-c", "multiport-admin", "--", "touch", "/tmp/unhealthy-multiport-admin")
```
zalimeni, line 91: ~ Non-blocking thought: Would it make sense to check mixed success (`multiport` fails, `multiport-admin` succeeds) after the first of these two commands?

That might give us just a bit more coverage to know that port connections fail independently on their respective health checks.

zalimeni (approved):
LGTM, thanks for chasing this fix down and double checking on test coverage!

acceptance/tests-v2/mesh/mesh_inject_test.go
ndhanushkodi: will address this next time I'm adding to acceptance tests!

PR #2996, opened 2023-09-21
refactored annotation processing

wilkermichael:
Changes proposed in this PR:
- Refactored annotation processing so it can be reused for the environment variable injection in this follow up PR: https://github.com/hashicorp/consul-k8s/pull/2995
- mostly just moved the processing code from the pod controller into common, it's mostly the same with some modifications to support the new package it's in
- I moved the "get" of the constants to the constants package as I felt in hindsight this actually made more sense

How I've tested this PR:
- added unit tests

👀 

Checklist:
- [ ] Tests added


control-plane/connect-inject/constants/constants.go
```diff
@@ -54,3 +54,27 @@ const (
 
 	KubernetesSuccessReasonMsg = "Kubernetes health checks passing"
 )
+
+func GetDefaultConsulNamespace(ns string) string {
```
wilkermichael, line 58: I'll add comments to these public methods

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -903,35 +665,3 @@ func getUpstreamsID(name, namespace, partition string) *pbresource.ID {
 		},
 	}
 }
-
-func upstreamReferenceType() *pbresource.Type {
```
wilkermichael, line 907: moved to constants

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -903,35 +665,3 @@ func getUpstreamsID(name, namespace, partition string) *pbresource.ID {
...
-	return &pbresource.Type{
-		Group:        "catalog",
-		GroupVersion: "v1alpha1",
-		Kind:         "Service",
-	}
-}
```
zalimeni, line 913: Heads up, this will get deleted when https://github.com/hashicorp/consul-k8s/pull/2993 is merged, at which point you'll want to replace the new one too. Since it needs an export here, you're probably stuck w/ a conflict either way, so probably just a `TODO` comment.

wilkermichael: Thanks for the heads up, I'll also have to make the other changes made in that PR to these functions.

control-plane/connect-inject/constants/constants.go
```diff
@@ -54,3 +54,33 @@ const (
 
 	KubernetesSuccessReasonMsg = "Kubernetes health checks passing"
 )
+
+// GetDefaultConsulNamespace returns the default namespace if the passed namespace
+// is empty, otherwise returns back the passed in namespace.
+func GetDefaultConsulNamespace(ns string) string {
```
zalimeni, line 60: ~ I feel like the names for these are a bit confusing. WDYT about something like `GetNormalizedConsulNamespace`?

The name implies that we're getting the default, but that's only true when `ns` is empty.

zalimeni (approved):
High-level, LGTM assuming there's no functional diff between the moved tests and functions (if so can you call out and I'll have a closer look?)

control-plane/connect-inject/common/annotation_processor.go
```diff
@@ -0,0 +1,275 @@
...
+	enableNamespaces bool
+}
+
+// NewPodAnnotationProcessor constructs a PodAnnotationProcessor for processing pod annotations into pbmesh.Upsreams.
+func NewPodAnnotationProcessor(enablePartitions, enableNamespaces bool) PodAnnotationProcessor {
+	return PodAnnotationProcessor{
+		enablePartitions: enablePartitions,
+		enableNamespaces: enableNamespaces,
+	}
+}
```
zalimeni, line 30: ~ I'm not sure this abstraction is super necessary since it only serves a single function call, unless it sets you up for more complex logic later - I might consider just adding this data as args to `ProcessUpstreams`

control-plane/connect-inject/constants/constants.go
wilkermichael: yeah, that sounds good

PR #2998, opened 2023-09-22
Add acceptance test for traffic permissions

thisisnotashwin:
Changes proposed in this PR:
- Add a test that ensure traffic permissions behave as expected. Currently since ACLs dont work, we explicitly create a deny permission, validate that it stops the connection from working and then set that permission to allow and let the rest of the test run validating the traffic was correctly moderated.

How I've tested this PR:
- pipeline

- 👀 

Checklist:
- [x] Tests added


ndhanushkodi (approved):
🎉

Issue #1516, opened 2022-09-19
Servers go through rolling-deploy for config changes on Consul servers


dnlopes:
Hello,

I have been trying out multiple deployment modes for Consul. I have sucessfully deployed a multi-DC via WAN federation using autoscaling groups in AWS, and now I’m moving to try out deploying on top of K8s. The experience was pretty smooth with ASG; in the end I got a pretty stable setup where I could add/remove nodes at will and the datacenters reacted smoothly (e.g, consensus was impeccable).

However, I keep having a lot of stability issues on the Raft consensus on top of K8s:

#### Question

1. increasing from 3 replicas to 5 replicas for some reason makes Raft lose consensus. I don’t undertstand this, why would Raft lose consensus when increasing the replicas?
2. with a 5-node deployment, changing a Consul setting (e.g., log rotation) and then upgrading helm once again makes the consensus to be lost.

One of the issues I believe I detected is that, because pods are named like consul-server-1, consul-server-2, etc, when pods are refreshing, they come up with the same name as the previous one (instead of generating a random suffix for example). This makes some members of the consensus protocol to detect two nodes with the same name (e.g., consul-server-2) running on different IPs (one from the new pod, and one from the old pod that was just replaced). But because this happens very quickly, the nodes don’t have time to “forget” the old pod causing a naming conflict.

Other more general questions:
1. how can I automate the servers upgrade process without downtime? [Official documentation](https://www.consul.io/docs/k8s/upgrade#upgrading-consul-servers) mentions that I should manipulate the `server.partitions` setting in multiple phases. However, in a “real” scenario, in which deploys are managed by CI/CD tools, does it mean I need to do multiple commits and multiple deploys to ensure the servers all receive the upgrade? It does sound a bit unproductive. Are there any other alternatives to this, while still using the official Helm chart?
2. the helm chart is using deprecated settings, both from Consul as well as K8s settings (e.g., TLS settings and PodSecurityPolicy). Is this a known issue?

**I’m probably missing something that could explain these issues, as I am fairly new working with K8s.**

### Helm Configuration

```yaml
global:
  enabled: true
  logLevel: "debug"
  logJSON: false
  name: "dlo"
  datacenter: "dlo"
  consulAPITimeout: "5s"
  enablePodSecurityPolicies: true
  recursors: []
  tls:
    enabled: true
    enableAutoEncrypt: true
    serverAdditionalDNSSANs: []
    serverAdditionalIPSANs: []
    verify: true
[...]

PR #3000, opened 2023-09-22
Set leave_on_terminate=true for servers and hardcode maxUnavailable=1

lkysow:
When leave_on_terminate=false (default), rolling the statefulset is
disruptive because the new servers come up with the same node IDs but
different IP addresses. They can't join the server cluster until the old
server's node ID is marked as failed by serf. During this time, they continually
start leader elections because they don't know there's a leader. When
they eventually join the cluster, their election term is higher, and so
they trigger a leadership swap. The leadership swap happens at the same
time as the next node to be rolled is being stopped, and so the cluster
can end up without a leader.

With leave_on_terminate=true, the stopping server cleanly leaves the
cluster, so the new server can join smoothly, even though it has the
same node ID as the old server. This increases the speed of the rollout
and in my testing eliminates the period without a leader.

The downside of this change is that when a server leaves gracefully, it
also reduces the number of raft peers. The number of peers is used to
calculate the quorum size, so this can unexpectedly change the fault
tolerance of the cluster. When running with an odd number of servers, 1
server leaving the cluster does not affect quorum size. E.g. 5 servers
=> quorum 3, 4 servers => quorum still 3. During a rollout, Kubernetes
only stops 1 server at a time, so the quorum won't change. During a
voluntary disruption event, e.g. a node being drained, Kubernetes uses
the pod disruption budget to determine how many pods in a statefulset
can be made unavailable at a time. That's why this change hardcodes this
number to 1 now.

Also set autopilot min_quorum to min quorum and disable autopilot
upgrade migration since that's for blue/green deploys.

How I've tested:
* Acceptance test that restarts the servers and ensures there's no loss of quorum
* Manual acceptance testing bumping server versions and ensuring the rollout is smooth (https://gist.github.com/lkysow/d310e61dfd207c438528f393e709a0f8). I didn't add this to the test suite because we need to run a helm install with an older version and then upgrade to the current version which requires more thought/work for how that test and versions will get updated.
* manual testing of wan federation, restarting servers on both sides and ensuring the new nodes are found and the connection still works
* manual testing of cluster peering, restarting servers on both sides, communication between svcs has no issue, no errors in logs

### DIFF
--- a/.changelog/2962.txt
+++ b/.changelog/2962.txt
@@ -1,3 +1,3 @@
-```releast-note:feature
+```release-note:feature
 api-gateway: (Consul Enterprise) Add JWT authentication and authorization for API Gateway and HTTPRoutes.
 ```
--- a/acceptance/tests/server/server_test.go
+++ b/acceptance/tests/server/server_test.go
@@ -1,0 +1,91 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package server
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/consul-k8s/acceptance/framework/consul"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
+	"github.com/hashicorp/go-multierror"
+	"github.com/stretchr/testify/require"
+)
+
+// Test that when servers are restarted, they don't lose leadership.
+func TestServerRestart(t *testing.T) {
+	cfg := suite.Config()
+	if cfg.EnableCNI || cfg.EnableTransparentProxy {
+		t.Skipf("skipping because -enable-cni or -enable-transparent-proxy is set and server restart " +
+			"is already tested without those settings and those settings don't affect this test")
+	}
+
+	ctx := suite.Environment().DefaultContext(t)
+	replicas := 3
+	releaseName := helpers.RandomName()
+	helmValues := map[string]string{
+		"global.enabled":        "false",
+		"connectInject.enabled": "false",
+		"server.enabled":        "true",
+		"server.replicas":       fmt.Sprintf("%d", replicas),
+		"server.affinity":       "null", // Allow >1 pods per node so we can test in minikube with one node.
+	}
+	consulCluster := consul.NewHelmCluster(t, helmValues, suite.Environment().DefaultContext(t), suite.Config(), releaseName)
+	consulCluster.Create(t)
+
+	// Start a separate goroutine to check if at any point more than one server is without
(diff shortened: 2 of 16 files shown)


.changelog/2962.txt
```diff
@@ -1,3 +1,3 @@
-```releast-note:feature
+```release-note:feature
```
lkysow, line 1: fix typo

wilkermichael:
@lkysow since you're adding a new acceptance test, in order to get it to run in the pipeline you need to add it to the yaml matrices: https://github.com/hashicorp/consul-k8s/tree/lkysow/server-restart/acceptance/ci-inputs

charts/consul/values.yaml
```diff
@@ -1005,8 +1005,12 @@ server:
     # the server cluster is enabled. To disable, set to `false`.
     enabled: true
 
-    # The maximum number of unavailable pods. By default, this will be
-    # automatically computed based on the `server.replicas` value to be `(n/2)-1`.
+    # The maximum number of unavailable pods. In most cases you should not change this. This setting has been kept
+    # as to not break backwards compatibility.
```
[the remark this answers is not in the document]
wilkermichael:
```suggestion
    # unset as to not break backwards compatibility with those using this value.yaml file.
```

wilkermichael (changes requested):
Looks good to me, just a couple nits (you can ignore if they don't make sense) and you need to add the `server` acceptance package to the acceptance test matrics (see above).

Thanks for all the great comments too, it was very easy to see your reasoning behind things.

.changelog/2962.txt
wilkermichael: Thanks!

charts/consul/values.yaml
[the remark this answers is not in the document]
lkysow: No, it's been kept around as a stanza in values.yaml so that you could set it. I'll make that clearer.

acceptance/tests/server/server_test.go
```diff
@@ -0,0 +1,84 @@
...
+	require.NoError(t, err, out)
+
+	// Wait for restart to finish.
+	start := time.Now()
+	out, err = k8s.RunKubectlAndGetOutputE(t, ctx.KubectlOptions(t), "rollout", "status", "--timeout", "5m", "--watch", fmt.Sprintf("statefulset/%s-consul-server", releaseName))
+	require.NoError(t, err, out, "rollout status command errored, this likely means the rollout didn't complete in time")
+
+	// Check results
+	require.NoError(t, unmarshallErrs, "there were some json unmarshall errors, this is likely a bug")
+	logger.Logf(t, "restart took %s, there were %d instances where than one server had no leader", time.Since(start), timesWithoutLeader)
```
wilkermichael, line 82:
```suggestion
	logger.Logf(t, "restart took %s, there were %d instances where more than one server had no leader", time.Since(start), timesWithoutLeader)
```

wilkermichael (approved):

david-yu:
@lkysow Thanks for working on this, could we close this issue? https://github.com/hashicorp/consul-k8s/issues/1516 I believe that issue is now resolved by the PR.

PR #3029, opened 2023-10-03
[NET-5718] feat(control-plane): ServiceAccount v2 backoff on missing Consul NS

zalimeni:
Add backoff for missing Consul NS similar to other v2 controllers.

Changes proposed in this PR:
- Add backoff for Service Account controller as follow-up to https://github.com/hashicorp/consul-k8s/pull/2960

How I've tested this PR: Testing follow-up ticket for non-default NS covers these similar to original PR.

Checklist:
- [ ] Tests added


DanStough (approved):
🚀

PR #3032, opened 2023-10-03
Fix Flaky Wan Federation Failover Test

wilkermichael:
Changes proposed in this PR:
- This PR addresses a flaky test that resulted from performing a CURL without a retry

How I've tested this PR:
👀 / pipeline

👀 

Checklist:
- [ ] Tests added

### DIFF
--- a/acceptance/tests/wan-federation/wan_federation_test.go
+++ b/acceptance/tests/wan-federation/wan_federation_test.go
@@ -423,20 +423,21 @@ func serviceFailoverCheck(t *testing.T, options *terratestK8s.KubectlOptions, po
 	var resp string
 	var err error
 
-	f := func(ft require.TestingT) {
+	// Retry until we get the response we expect, sometimes you get back the previous server until things stabalize
+	retry.RunWith(timer, t, func(r *retry.R) {
 		resp, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", "-i",
 			staticClientDeployment, "-c", connhelper.StaticClientName, "--", "curl", fmt.Sprintf("localhost:%s", port))
-		require.NoError(ft, err)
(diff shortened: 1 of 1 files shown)


acceptance/tests/wan-federation/wan_federation_test.go
```diff
@@ -423,20 +423,21 @@ func serviceFailoverCheck(t *testing.T, options *terratestK8s.KubectlOptions, po
...
 		resp, err = k8s.RunKubectlAndGetOutputE(t, options, "exec", "-i",
 			staticClientDeployment, "-c", connhelper.StaticClientName, "--", "curl", fmt.Sprintf("localhost:%s", port))
-		require.NoError(ft, err)
-		assert.Contains(ft, resp, expectedName)
-	}
+		assert.NoError(r, err)
+		assert.Contains(r, resp, expectedName)
+	})
 
+	// Try again to rule out load-balancing. Errors can still happen so retry
```
zalimeni, line 434: ~ I know this is status-quo so non-blocking: Do we know that LB would be round-robin in this scenario? Wondering whether multiple follow-up requests would be appropriate to prove we're not getting lucky. I did that in [this](https://github.com/hashicorp/consul-enterprise/blob/df2e31f83af2178aa75b53480a2f41f55390bb9d/test/integration/consul-container/test/basic/locality_aware_routing_ent_test.go#L130) test for similar reasons.

zalimeni (approved):

zalimeni:
Thank you for fixing this!

acceptance/tests/wan-federation/wan_federation_test.go
wilkermichael: good idea, added something similar

PR #3033, opened 2023-10-03
[NET-5805] Fixing validating cert webhooks, fixing replace statements for go mod

jm96441n:
Changes proposed in this PR:
- Fix replace statements for proto-public to point to an existing commit
- Fix cert issue on validating webhooks

How I've tested this PR:
Ran setup locally using https://github.com/hashicorp/shared-consul-examples/tree/main/k8s/jwts and saw prior to these changes we had webhook cert errors, after we no longer have them

Checklist:
- [ ] Tests added

### DIFF
--- a/control-plane/helper/webhook-configuration/webhook_configuration.go
+++ b/control-plane/helper/webhook-configuration/webhook_configuration.go
@@ -29,14 +29,20 @@ func UpdateWithCABundle(ctx context.Context, clientset kubernetes.Interface, web
 		return err
 	}
 
+	if !k8serrors.IsNotFound(err) {
 	err = updateMutatingWebhooksWithCABundle(ctx, clientset, mutatingWebhookCfg, caCert)
 	if err != nil {
 		return err
 	}
+    }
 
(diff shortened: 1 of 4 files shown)


control-plane/helper/webhook-configuration/webhook_configuration.go
```diff
@@ -25,11 +25,12 @@ func UpdateWithCABundle(ctx context.Context, clientset kubernetes.Interface, web
 	}
 
 	mutatingWebhookCfg, err := clientset.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(ctx, webhookConfigName, metav1.GetOptions{})
-	if err == nil {
-		return updateMutatingWebhooksWithCABundle(ctx, clientset, mutatingWebhookCfg, caCert)
+	if err != nil && !k8serrors.IsNotFound(err) {
```
jm96441n, line 28: so the TLDR is that rather than just return out from the mutating webhook update we should update both the validating webhooks and mutating webhooks, I was initially under the impression that each was named after their path but it looks like there's just a single `consul-connect-injector` config name that all the validating and mutating webhooks are just paths on

nathancoleman (approved):

sarahalsmiller (approved):

PR #3037, opened 2023-10-04
Mw/fix partitions test flakiness

wilkermichael:
Changes proposed in this PR:
- This is adds more server replicas to get around a potential bug, will remove this when the bug is fixed.
-

How I've tested this PR:

Checklist:
- [ ] Tests added

### DIFF
--- a/acceptance/tests/partitions/partitions_connect_test.go
+++ b/acceptance/tests/partitions/partitions_connect_test.go
@@ -87,9 +87,9 @@ func TestPartitions_Connect(t *testing.T) {
 			secondaryPartitionClusterContext := env.Context(t, 1)
 
 			commonHelmValues := map[string]string{
-				"sglobal.adminPartitions.enabled": "true",
-				"global.enableConsulNamespaces":   "true",
-				"global.logLevel":                 "debug",
+				"global.adminPartitions.enabled": "true",
+				"global.enableConsulNamespaces":  "true",
+				"global.logLevel":                "debug",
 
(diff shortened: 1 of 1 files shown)


acceptance/tests/partitions/partitions_connect_test.go
```diff
@@ -91,6 +91,11 @@ func TestPartitions_Connect(t *testing.T) {
 				"global.enableConsulNamespaces":  "true",
 				"global.logLevel":                "debug",
 
+				// Currently there is a bug with xds limits being reached for a single server, so need to increase
+				// the server replicas to 3
+				// TODO: remove when NET-5819 is fixed
+				"global.replicas": "3",
```
curtbushko, line 97: **question:** Should this be for CNI only or does it make all the partition tests better?

curtbushko (approved):
I have a question below but I will not block on it.

acceptance/tests/partitions/partitions_connect_test.go
wilkermichael: it could be for CNI only, I don't "think" the test was flaking on tests other than CNI... however this doesn't hurt anything so thought it would be simpler to not have a separate check for CNI being enabled.

hashi-derek: It's worth noting that this will make local test runs fail unless people are spinning up their kind clusters with 3+ nodes, but I don't see it being a huge ordeal.

wilkermichael:
I won't merge this as it doesn't really fix the problem

PR #3039, opened 2023-10-04
[NET-5704] feat(control-plane): v2 only register services for injected pods

zalimeni:
Rather than indiscriminately registering any service that is not excluded due to NS allow/deny-listing, selectively register services that target injected pods (i.e. in alignment with the mesh webhook).

Do not deregister services that already exist when endpoints are empty; rather, predicate registration on pod injection and deletion on service deletion from Kubernetes. In the future, we can layer on explicit allow/deny annotations for services as needed.

This is already implemented by existing workload selector logic that filters on mesh-inject status; this change removes `TODO`s, updates comments, and adds tests for the desired behavior.

Changes proposed in this PR:
- Update `TODO`s and comments
- Add tests verifying behavior

How I've tested this PR: Added tests

Checklist:
- [x] Tests added


DanStough (approved):
LGTM!

PR #3045, opened 2023-10-05
[NET-5818] Fix Status for Gateway Policy 

jm96441n:
Changes proposed in this PR:
- Fixes status for Gateway Policies when the gateway they reference is removed. Previously we set some default values, now we show no status to mimic the behavior on routes when a gateway they reference cannot be found 

How I've tested this PR:
Ran locally

Checklist:
- [X] Tests added


t-eckert (approved):

NicoletaPopoviciu:
@jm96441n since this is going in the RC, cc @david-yu

david-yu:
Sounds good. Thank you.

PR #3049, opened 2023-10-05
[NET-5681] v2: Debounce unnecessary Service writes to Consul

zalimeni:
Endpoints Controller v2 should not write services every time endpoints change, as this can occur far more frequently and be subject to rapid updates in the case of large clusters or flapping pod health.

Instead, track writes by storing a fingerprint of the written payload along with the Consul resource generation, and compare each potential write and existing Consul resource to them, writing only if a discrepancy is detected or the fingerprint read/compare itself fails (pessimistically fall back to blind writes).

Changes proposed in this PR:
- Add `WriteCache` to Endpoints Controller V2 and use to debounce writes from redundant reconciles
- Add test hook to verify write/no-write scenarios

How I've tested this PR: New tests, acceptance test for V2 should continue to pass

Checklist:
- [x] Tests added

### DIFF
--- a/control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
@@ -5,6 +5,7 @@ package endpointsv2
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"net"
 	"sort"
@@ -14,6 +15,10 @@ import (
 	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v2beta1"
 	"github.com/hashicorp/consul/proto-public/pbresource"
 	"github.com/hashicorp/go-multierror"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -42,6 +47,13 @@ type Controller struct {
 	// ConsulTenancyConfig manages settings related to Consul namespaces and partitions.
 	common.ConsulTenancyConfig
 
+	// WriteCache keeps track of records already written to Consul in order to enable debouncing of writes.
+	// This is useful in particular for this controller which will see potentially many more reconciles due to
+	// endpoint changes (e.g. pod health) than changes to service data written to Consul.
+	// It is intentionally simple and best-effort, and does not guarantee against all redundant writes.
+	// It is not persistent across restarts of the controller process.
+	WriteCache WriteCache
+
 	Log logr.Logger
 
 	Scheme *runtime.Scheme
@@ -53,6 +65,9 @@ func (r *Controller) Logger(name types.NamespacedName) logr.Logger {
 }
 
 func (r *Controller) SetupWithManager(mgr ctrl.Manager) error {
+	if r.WriteCache == nil {
+		return fmt.Errorf("WriteCache was not configured for Controller")
+	}
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Endpoints{}).
 		Complete(r)
@@ -80,7 +95,7 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// we need to deregister that service from Consul.
 	err = r.Client.Get(ctx, req.NamespacedName, &endpoints)
 	if k8serrors.IsNotFound(err) {
-		err = r.deregisterService(ctx, resourceClient, req.Name, r.getConsulNamespace(req.Namespace), r.getConsulPartition())
+		err = r.deregisterService(ctx, resourceClient, req)
 		return ctrl.Result{}, err
 	} else if err != nil {
 		r.Log.Error(err, "failed to get Endpoints", "name", req.Name, "ns", req.Namespace)
@@ -111,7 +126,13 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	// Register the service in Consul.
-	if err = r.registerService(ctx, resourceClient, service, consulSvc); err != nil {
+	id := getServiceID(
+		service.Name, // Consul and Kubernetes service name will always match
+		r.getConsulNamespace(service.Namespace),
+		r.getConsulPartition())
+	meta := getServiceMeta(service)
+	k8sUid := string(service.UID)
+	if err = r.ensureService(ctx, &defaultResourceReadWriter{resourceClient}, k8sUid, id, meta, consulSvc); err != nil {
 		// We could be racing with the namespace controller.
 		// Requeue (which includes backoff) to try again.
 		if inject.ConsulNamespaceIsNotFound(err) {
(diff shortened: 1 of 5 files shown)


control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -510,6 +545,22 @@ func (r *Controller) getConsulPartition() string {
...
+	// We aggressively delete from the cache whenever an invalidating event occurs, and rely on writes
+	// for eventual consistency. See also WriteCache.hasMatch.
+	return serviceName.String()
+}
+
+// getWriteHash gets a hash of the given resource to deduplicate writes to Consul.
+// This hash is not intended to be cryptographically secure, only deterministic and collision-resistent for tens of
+// thousands of values.
+func getWriteHash(resource *pbresource.Resource) []byte {
+	h := sha256.Sum256(resource.Data.Value)
```
rboyer, line 560: I wonder if encoding an `any.Any` field into `[]byte` will do a plain Marshal or the Deterministic=true one: https://pkg.go.dev/google.golang.org/protobuf/proto#MarshalOptions

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -262,34 +274,54 @@ func getOwnerPrefixFromPod(pod *corev1.Pod) string {
...
-		r.getConsulPartition(),
-		getServiceMeta(k8sService),
-	)
+		r.getConsulPartition())
+
+	// We use the locally-created Resource and ID (without Uid and Version)
+	// when writing so that it behaves as an upsert rather than CAS.
+	consulSvcResource := &pbresource.Resource{
+		Id:       id,
+		Data:     inject.ToProtoAny(consulSvc),
```
rboyer, line 289: I think you need to change the underlying `inject.ToProtoAny` code to not use `any.New` in favor of `any.MarshalFrom` so you can set `Deterministic=true` which should up the chances of actually debouncing.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
rboyer: See my other comment.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: Ah, TIL `Deterministic=true`, thanks for catching this! I checked some other docs and it _looked_ like marshalling was default deterministic within a given version of the language proto impl - but this does looks necessary here.

zalimeni: Given `inject.ToProtoAny` is general-purpose and forces sorting, I opted for a direct marshal just for this case, but I think this should work correctly now.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
zalimeni: Yep, I think that was the right fix you suggested - that way we avoid deterministically encoding a non-deterministic payload 😅

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -510,6 +554,22 @@ func (r *Controller) getConsulPartition() string {
...
+	// for eventual consistency. See also WriteCache.hasMatch.
+	return serviceName.String()
+}
+
+// getWriteHash gets a hash of the given resource to deduplicate writes to Consul.
+// This hash is not intended to be cryptographically secure, only deterministic and collision-resistent for tens of
+// thousands of values.
+func getWriteHash(data *anypb.Any) []byte {
+	h := sha256.Sum256(data.Value)
+	return h[:]
```
DanStough, line 570: neat!

DanStough (commented):
Preliminary `LGTM`. We already discussed my comment about `multierror` and it looks like it's resolved.

zalimeni:
Pushed additional tests but waiting on fixes from https://github.com/hashicorp/consul-k8s/pull/3069 - tests are passing locally on 1.17 RC `consul` binary.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -510,6 +586,31 @@ func (r *Controller) getConsulPartition() string {
...
+// This hash is not intended to be cryptographically secure, only deterministic and collision-resistent
+// for tens of thousands of values.
+//
+// If an error occurs marshalling the resource for the hash, returns a nil hash value and the error.
+// error will be returned.
+func getWriteHash(r *pbresource.Resource) ([]byte, error) {
+	// We convert the entire resource (not just its own Data, which is also an Any)
+	// to Any in order to take advantage of the deterministic marshal offered by proto
+	// and include fields like Meta, which are not part of the resource Data.
+	data := new(anypb.Any)
```
rboyer, line 606: There is a toplevel `proto` function you can call to do this without the `anypb` step. We already do that variant in `agent/xds/delta.go` for reference.

zalimeni: You beat me to pushing - thanks for the tip! Fixed.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller_test.go
```diff
@@ -54,9 +54,11 @@ type reconcileCase struct {
 	targetConsulNs        string
 	targetConsulPartition string
 	expErr                string
+	caseFn                func(*testing.T, *reconcileCase, *Controller, pbresource.ResourceServiceClient)
 }
 
 // TODO(NET-5716): Allow/deny namespaces for reconcile tests
+// TODO(NET-5932): Add tests for consistently sorting repeated output fields (getConsulService, getServicePorts)
```
zalimeni, line 61: I added this to improve guards around determinism. Test coverage should be sufficient as-is, so I wanted to avoid blocking review further given time constraints and treat this as an ideal follow-up if there's time.

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller_test.go
```diff
@@ -1325,6 +1327,174 @@ func TestReconcile_UpdateService(t *testing.T) {
...
+						Prefixes: []string{"service-created-rs-abcde"},
+					},
+					VirtualIps: []string{"172.18.0.1"},
+				}),
+				Metadata: map[string]string{
+					constants.MetaKeyKubeNS:    constants.DefaultConsulNS,
+					constants.MetaKeyManagedBy: constants.ManagedByEndpointsValue,
+				},
+			},
+			caseFn: func(t *testing.T, tc *reconcileCase, ep *Controller, resourceClient pbresource.ResourceServiceClient) {
```
zalimeni, line 1442: I considered moving this special case to the separate test, but wanted to fully exercise the debounce flow via reconciles just to ensure the method tests don't drift from reality.

zalimeni:
Going to ask about why adding sleeps to tests was necessary to de-flake tomorrow, but for now, this should be ready for review again.

control-plane/connect-inject/controllers/endpointsv2/write_cache_test.go
```diff
@@ -0,0 +1,237 @@
...
+					return testGeneration
+				},
+				testK8sUid,
+			},
+			setupFn: func(args args, cache WriteCache) {
+				cache.update(args.key, args.hash, args.generationFetchFn(), args.k8sUid)
+				cache.update(args.key, args.hash, args.generationFetchFn(), "")
+			},
+			want: true,
+		},
```
DanStough, line 193: Not sure what I would expect from the behavior here if someone wants to update the cache with junk values. In that case, I think we'd want to fail on the side of always wanting to invalidate it. No strong opinion here.

DanStough (approved):
Looks great! Awesome improvement. Your only regret will be never knowing how many customers would have run into issues here :)

control-plane/connect-inject/controllers/endpointsv2/write_cache_test.go
zalimeni: That's a fair callout. I think here it will be 6+half-dozen due to the fact that either way, the next write is guaranteed (we didn't match and wrote to Consul, hence the cache write - so skipping the cache update should be equivalent to invalidating). I expect the only solution in either case will be to notice the error logs or tons of writes from the controller and investigate.

PR #3057, opened 2023-10-06
Fix local rate limiting acceptance tests

cthain:
Changes proposed in this PR:
- This PR fixes the failing local rate limiting tests for tproxy and CNI.
- There was a bug in the test when tproxy was enabled where it would try to reach the `static-server` via `localhost:1234` (explict upstream) instead of using the service name. 

How I've tested this PR:
- Locally
- CI

:eyes: 

Checklist:
- [x] Tests updated
- [x] ~[CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)~


curtbushko (approved):
Awesome fix!

(Michael is out)

kisunji (approved):

PR #3070, opened 2023-10-11
NET-5947 Add NET_BIND_SERVICE capability in security context for api-gateway pod on OpenShift

nathancoleman:
**Why this change is needed:**
This capability became a requirement for consul-dataplane, which api-gateway uses under the hood, as of https://github.com/hashicorp/consul-dataplane/pull/238. Since the `securityContext` created for each consul-dataplane `Pod` was not requesting the `NET_BIND_SERVICE` capability specifically when deploying onto OpenShift, it was not being granted, and the `Pod` was failing to start with a permission denied error.

- Always add required `NET_BIND_SERVICE` capability to the `securityContext` for each api-gateway `Pod`

**How I've tested this PR:**
- Added unit test coverage verifying `securityContext` is _always_ set appropriately on `Pod`
- @missylbytes tested it live on OpenShift 4.11 and 4.12 clusters, verifying that the `Pod` starts up correctly now

- 🤖 tests passing

**Checklist:**
- [x] Tests added

### DIFF
--- a/.changelog/3070.txt
+++ b/.changelog/3070.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api-gateway: add required `NET_BIND_SERVICE` capability to the `securityContext` for every api-gateway `Pod`
+```
--- a/.changelog/3070.txt
+++ b/.changelog/3070.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-api-gateway: fix issue where missing `NET_BIND_SERVICE` capability prevented api-gateway `Pod` from starting up
+api-gateway: fix issue where missing `NET_BIND_SERVICE` capability prevented api-gateway `Pod` from starting up when deployed to OpenShift
 ```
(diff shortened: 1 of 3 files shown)


t-eckert (approved):

.changelog/3070.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+api-gateway: add required `NET_BIND_SERVICE` capability to the `securityContext` for every api-gateway `Pod`
```
nathancoleman, line 2:
```suggestion
api-gateway: fix issue where missing `NET_BIND_SERVICE` capability prevented api-gateway `Pod` from starting up
```

curtbushko (approved):
I don't really like it but I understand why it is needed. Approve!

control-plane/api-gateway/gatekeeper/dataplane.go
```diff
@@ -19,7 +18,7 @@ import (
 )
 
 const (
-	allCapabilities              = "all"
+	allCapabilities              = "ALL"
```
nathancoleman, line 21: This was causing an odd merge when the OpenShift controller added `ALL` alongside our `all` and resulted in
```yaml
drop:
- ALL
- all
```

missylbytes: Fixed: 
image

missylbytes (approved):

PR #3077, opened 2023-10-12
Fix ENT Tests Now that They Are Running Again 🏃 

DanStough:
ENT tests were just turned back on in CI. This revealed some failing test cases. This PR attempts to get everything in working order.

### DIFF
--- a/control-plane/config-entries/controllers/configentry_controller_ent_test.go
+++ b/control-plane/config-entries/controllers/configentry_controller_ent_test.go
@@ -294,13 +294,13 @@ func TestConfigEntryController_updatesEntConfigEntry(t *testing.T) {
 			},
 			updateF: func(resource common.ConfigEntryResource) {
 				sg := resource.(*v1alpha1.SamenessGroup)
-				sg.Spec.IncludeLocal = false
+				sg.Spec.DefaultForFailover = false
 			},
 			compare: func(t *testing.T, consulEntry capi.ConfigEntry) {
 				resource, ok := consulEntry.(*capi.SamenessGroupConfigEntry)
 				require.True(t, ok, "cast error")
-				require.Equal(t, true, resource.DefaultForFailover)
-				require.Equal(t, false, resource.IncludeLocal)
+				require.Equal(t, false, resource.DefaultForFailover)
+				require.Equal(t, true, resource.IncludeLocal)
 				require.Equal(t, "dc1", resource.Members[0].Peer)
 				require.Equal(t, "", resource.Members[0].Partition)
 			},
--- a/control-plane/config-entries/controllersv2/meshconfig_controller_test.go
+++ b/control-plane/config-entries/controllersv2/meshconfig_controller_test.go
@@ -145,6 +145,11 @@ func TestMeshConfigController_createsMeshConfig(t *testing.T) {
 			resourceClient, err := consul.NewResourceServiceClient(testClient.Watcher)
 			require.NoError(t, err)
 
+			require.Eventually(t, func() bool {
+				_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
+				return err == nil
+			}, 5*time.Second, 500*time.Millisecond)
+
 			r := c.reconciler(fakeClient, testClient.Cfg, testClient.Watcher, logrtest.New(t))
 			namespacedName := types.NamespacedName{
 				Namespace: metav1.NamespaceDefault,
@@ -284,6 +289,12 @@ func TestMeshConfigController_updatesMeshConfig(t *testing.T) {
 			})
 			resourceClient, err := consul.NewResourceServiceClient(testClient.Watcher)
 			require.NoError(t, err)
+
+			require.Eventually(t, func() bool {
+				_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
+				return err == nil
+			}, 5*time.Second, 500*time.Millisecond)
+
 			// We haven't run reconcile yet, so we must create the MeshConfig
 			// in Consul ourselves.
 			{
@@ -400,6 +411,11 @@ func TestMeshConfigController_deletesMeshConfig(t *testing.T) {
 			resourceClient, err := consul.NewResourceServiceClient(testClient.Watcher)
 			require.NoError(t, err)
 
+			require.Eventually(t, func() bool {
+				_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
+				return err == nil
+			}, 5*time.Second, 500*time.Millisecond)
+
 			// We haven't run reconcile yet, so we must create the config entry
 			// in Consul ourselves.
 			{
@@ -465,6 +481,11 @@ func TestMeshConfigController_errorUpdatesSyncStatus(t *testing.T) {
 	testClient := test.TestServerWithMockConnMgrWatcher(t, func(c *testutil.TestServerConfig) {
 		c.Experiments = []string{"resource-apis"}
 	})
+
+	require.Eventually(t, func() bool {
+		_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
+		return err == nil
+	}, 5*time.Second, 500*time.Millisecond)
 
 	// Stop the server before calling reconcile imitating a server that's not running.
 	_ = testClient.TestServer.Stop()
@@ -547,6 +568,11 @@ func TestMeshConfigController_setsSyncedToTrue(t *testing.T) {
 	resourceClient, err := consul.NewResourceServiceClient(testClient.Watcher)
 	require.NoError(t, err)
 
+	require.Eventually(t, func() bool {
+		_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
+		return err == nil
+	}, 5*time.Second, 500*time.Millisecond)
+
 	reconciler := &TrafficPermissionsController{
 		Client: fakeClient,
 		Log:    logrtest.New(t),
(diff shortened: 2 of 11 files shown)


control-plane/connect-inject/controllers/endpoints/endpoints_controller_ent_test.go
```diff
@@ -641,6 +641,9 @@ func TestReconcileUpdateEndpointWithNamespaces(t *testing.T) {
 					{
 						Node:    consulNodeName,
 						Address: consulNodeAddress,
+						NodeMeta: map[string]string{
+							metaKeySyntheticNode: "true",
+						},
```
DanStough, line 646: These changes were made in a PR to the OSS tests but missed here.

control-plane/config-entries/controllers/configentry_controller_ent_test.go
```diff
@@ -294,13 +294,13 @@ func TestConfigEntryController_updatesEntConfigEntry(t *testing.T) {
 			},
 			updateF: func(resource common.ConfigEntryResource) {
 				sg := resource.(*v1alpha1.SamenessGroup)
-				sg.Spec.IncludeLocal = false
+				sg.Spec.DefaultForFailover = false
```
DanStough, line 297: You're not allowed to change this field or you get a 500.

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -321,6 +322,11 @@ func TestRun_ACLPolicyUpdates(t *testing.T) {
 			require.NoError(t, err)
 
 			// Check that the expected policies were created.
+			// There will be more policies returned in the List API that are defaults
+			// existing in Consul on startup, including but not limited to:
+			// * global-management
+			// * builtin/global-read-only
+			// * agent-token
```
DanStough, line 329: We added a default policy and the counting broke. I don't see the point in testing the count - we already test that the policies we expect to create exist below. The fact that there are defaults is extraneous and brittle.

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -998,6 +995,7 @@ partition "default" {
 // Test that server-acl-init used the local auth method to create the desired token in the primary datacenter.
 // The test works by running the login command and then ensuring that the token
 // returned has the correct role for the component.
+// TODO(dans): these flake, fix
```
DanStough, line 998: I fixed this with the timeout. I need to remove this comment

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -1050,7 +1048,7 @@ func TestRun_NamespaceEnabled_ValidateLoginToken_PrimaryDatacenter(t *testing.T)
 				clientset: k8s,
 			}
 			cmdArgs := append([]string{
-				"-timeout=500ms",
+				"-timeout=1m",
```
DanStough, line 1051: All the other tests use 1m, and this was causing a flake in some cases.

control-plane/config-entries/controllersv2/meshconfig_controller_test.go
```diff
@@ -145,6 +145,14 @@ func TestMeshConfigController_createsMeshConfig(t *testing.T) {
...
 			require.NoError(t, err)
 
+			require.Eventually(t, func() bool {
+				_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
+				if err != nil {
+					return false
+				}
+				return true
+			}, 5*time.Second, 500*time.Millisecond)
+
```
DanStough, line 155: Ideally, this moves to Consul's `sdk`, but it needs to be done in a way that works across V1 and V2 that will require some more thought, more than cleaning up test failure on `main`.

DanStough:
There are still flakes in these tests, but it would seem not enough that would prevent them from passing with successive gotestsum runs.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
```diff
@@ -602,6 +622,9 @@ func TestProxyConfigurationWrite(t *testing.T) {
 							},
 						},
 					},
+					TransparentProxy: &pbmesh.TransparentProxy{
+						OutboundListenerPort: 15001,
+					},
```
zalimeni, line 627: Just curious: Newly required when using TProxy?

control-plane/subcommand/server-acl-init/command_ent_test.go
```diff
@@ -386,15 +386,10 @@ func TestRun_ACLPolicyUpdates(t *testing.T) {
 				"anothertgw-policy",
 				"partitions-token",
 			}
+			time.Sleep(1 * time.Second)
 			policies, _, err = consul.ACL().PolicyList(nil)
 			require.NoError(t, err)
```
zalimeni, line 391: ~ `require.EventuallyWithT` instead of sleep?

control-plane/config-entries/controllersv2/meshconfig_controller_test.go
```diff
@@ -145,6 +145,11 @@ func TestMeshConfigController_createsMeshConfig(t *testing.T) {
 			resourceClient, err := consul.NewResourceServiceClient(testClient.Watcher)
 			require.NoError(t, err)
 
+			require.Eventually(t, func() bool {
+				_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
+				return err == nil
+			}, 5*time.Second, 500*time.Millisecond)
```
zalimeni, line 151: ~~ Could helper-ify this since it's needed a bunch of places

control-plane/connect-inject/controllers/pod/pod_controller_ent_test.go
```diff
@@ -674,6 +675,14 @@ func runControllerTest(t *testing.T, tc testCase) {
 	resourceClient, err := consul.NewResourceServiceClient(testClient.Watcher)
 	require.NoError(t, err)
 
+	require.Eventually(t, func() bool {
+		_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
+		if err != nil {
+			return false
+		}
+		return true
```
zalimeni, line 683:
```suggestion
		_, _, err := testClient.APIClient.Partitions().Read(context.Background(), constants.DefaultConsulPartition, nil)
		return err == nil
```

zalimeni (approved):
Thank you for fixing these!

control-plane/connect-inject/controllers/peering/peering_acceptor_controller_test.go
```diff
@@ -508,6 +509,7 @@ func TestReconcile_CreateUpdatePeeringAcceptor(t *testing.T) {
 			// Create test consul server.
 			testClient := test.TestServerWithMockConnMgrWatcher(t, nil)
 			consulClient := testClient.APIClient
+			testClient.TestServer.WaitForActiveCARoot(t)
```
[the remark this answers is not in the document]
DanStough: In the peering case, we want to wait for the CA certs specifically.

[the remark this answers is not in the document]
zalimeni: Ah, got it. Thanks

control-plane/config-entries/controllersv2/meshconfig_controller_test.go
DanStough: That's a fair comment, but I didn't want to find a home for this if eventually I can get it back in the SDK.

control-plane/connect-inject/controllers/pod/pod_controller_test.go
DanStough: Yes, Iryna changed the mutating webhook for ProxyConfiguration to add this.

control-plane/subcommand/server-acl-init/command_ent_test.go
DanStough: Ah, good catch. I was just playing around here, so I'll see if I can get away with removing it. There still might be an optimization here to wait for the bootstrap token, but it passes consistency on consecutive local runs.

DanStough:
@zalimeni re: confused emoji: `gotestsum` reruns any failed tests up to 3 times, so it can flake occasionally but not block CI. I wouldn't say to rely on it, but trying to balance out the investment in fixing this before a release.

zalimeni:
> @zalimeni re: confused emoji: 

Thank you for the context! I was actually just looking for the closest thing to a sad face, bc there are still flakes despite all your efforts here. I think that approach is super reasonable.

thisisnotashwin (approved):
:chef-kiss:

zalimeni:
Looking at some of these failures to lend a hand since they're still looking weird.

```
=== FAIL: subcommand/fetch-server-region TestRun_FlagValidation/missing_node_name (0.00s)
    command_test.go:46: 
        	Error Trace:	/home/runner/actions-runner/_work/consul-k8s-workflows/consul-k8s-workflows/control-plane/subcommand/fetch-server-region/command_test.go:46
        	Error:      	"-output-file is required\n-output-file is required\n" does not contain "-node-name is required"
        	Test:       	TestRun_FlagValidation/missing_node_name
```
Just noting here that this looks like a race w/ the synced buffer underlying `MockUI` (? why is this necessary?)... Going to leave that be until we see more of them and focus on the pod controller tests.

control-plane/subcommand/server-acl-init/command_ent_test.go
nathancoleman, line 329: I'm not sure if it makes sense to backport this entire PR to 1.2.x or not, but I have been running into a failing test there as well for this exact reason. Thought I'd mention it and let you decide. For my own backport PR, I've stolen your changes to get this test passing again.

nathancoleman: Running into more and more failures -- I'm thinking it does make sense to backport this PR to 1.2.x. Thoughts?

zalimeni: @nathancoleman we'd need to cut out the v2 tests fixed here (they don't exist in 1.2.x), but backporting the common fixes through n-2 sounds like the right thing to do.

I'll set a reminder for us to revisit Monday unless you already have it covered.

zalimeni:
Tests are passing now, with @DanStough's fixes and nothing disabled.

I believe inconsistent failures came from `main` and `release/1.17.x` on `consul` sharing a `VERSION` file, leading them to race each other to overwrite `hashicorppreview/consul-1.17-dev`. That plus some missing backports earlier today would have been enough to flip-flop expected behavior without these tests changing.

No-op amended the commit to get a fresh SHA so none of the GHA results were old.

control-plane/subcommand/server-acl-init/command_ent_test.go
nathancoleman: For future readers, see linked PRs below for manual backports into `release/1.1.x` and `release/1.2.x`

PR #3079, opened 2023-10-13
NET-5943 - Fix ACL not found error in K8 application services

nvanthao:
Changes proposed in this PR:
- When deleting ACL token in Endpoint Controller Reconcile loop, only delete tokens that created before the Reconcile start time. 

This will prevent this scenario from happening: we have 2 tokens for same pod name due to `StatefulSet` consistent pod naming.

Before the PR, both old and new token will be deleted (based on pod name).
After the PR, only old token will be deleted since new token is created after current Reconcile loop

How I've tested this PR:

- Manual test with synthetic delay of `15` seconds before retrieving all ACL tokens in `deleteACLTokensForServiceInstance`

Checklist:
- [ ] Tests added


aahel (commented):
the two timestamps(token create time and reconciliation start time) may not be equivalent since they might come from two different nodes.

We  may be able to compare by storing pod uid in the tokenMeta. And comparing using that.

nvanthao:
Thanks for the review, @aahel!

Could you suggest a way to retrieve pod uid during `ACL.Login` in container?

https://github.com/hashicorp/consul-k8s/blob/488e8acb842e270181f6e75dab8bc191aa4ab73e/control-plane/subcommand/connect-init/command.go#L127

aahel:
> Thanks for the review, @aahel!
> 
> Could you suggest a way to retrieve pod uid during `ACL.Login` in container?
> 
> https://github.com/hashicorp/consul-k8s/blob/488e8acb842e270181f6e75dab8bc191aa4ab73e/control-plane/subcommand/connect-init/command.go#L127

i think we should be able to pass uid as flag in connect-init command along with pod_name

``
{
				Name: "POD_UID",
				ValueFrom: &corev1.EnvVarSource{
					FieldRef: &corev1.ObjectFieldSelector{FieldPath: "metadata.uid"},
				},
			},
``

nvanthao:
Thanks @aahel ,

I've updated the PR with your suggestion. We now only delete token that has metadata Pod UID matching Pod UID from Service Registration metadata.

E.g.

```
# token
AccessorID:       e2b86025-1554-577d-ce30-11fb89d6190e
SecretID:         a5c71ce6-2bad-d2cd-2007-d80f993351a8
Description:      token created via login: {"pod":"default/counting-6ddbc5d9b-dk2nx","pod-uid":"60d33906-0e90-44fe-b264-7dcd7cbe2abf"}
Local:            true
Auth Method:      consul-k8s-auth-method (Namespace: )
Create Time:      2023-10-26 00:27:14.759119632 +0000 UTC
Legacy:           false
Service Identities:
   counting (Datacenters: all)

# service registration
        "ServiceMeta": {
            "k8s-namespace": "default",
            "k8s-service-name": "counting",
            "managed-by": "consul-k8s-endpoints-controller",
            "pod-name": "counting-6ddbc5d9b-dk2nx",
            "pod-uid": "60d33906-0e90-44fe-b264-7dcd7cbe2abf"
        },
```

Please let me know how it goes!

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -807,6 +809,7 @@ func (r *EndpointsController) deleteACLTokensForServiceInstance(client *api.Clie
 		// Only delete tokens that:
 		// * have been created with the auth method configured for this endpoints controller
 		// * have a single service identity whose service name is the same as 'serviceName'
+		// * created before start time of current reconcile loop
```
aahel, line 812: this comment can be removed

aahel (changes requested):
the changes look good. 
Couple of things :-
- please take a look why the unit tests are failing.
- newer versions also seem to filter the token via pod name before deletion. 
If this bug is still there. we should create a pr against main and backport to all the other versions

aahel (approved):

control-plane/connect-inject/endpoints_controller.go
```diff
@@ -818,7 +820,7 @@ func (r *EndpointsController) deleteACLTokensForServiceInstance(client *api.Clie
 			tokenPodName := strings.TrimPrefix(tokenMeta[TokenMetaPodNameKey], k8sNS+"/")
 
 			// If we can't find token's pod, delete it.
-			if tokenPodName == podName {
+			if tokenPodName == podName && tokenMeta[MetaKeyPodUID] == podUID {
```
Ganeshrockz, line 823: How is backward compatibility maintained with this logic? Assuming that customers upgrade their existing installation to the latest release, `tokenMeta[MetaKeyPodUID]` could be an empty string for pods that got created before pushing this change. Won't we miss deleting tokens of such pods/service instances in such cases?

nvanthao: Many thanks for reviewing @Ganeshrockz and that's a great catch!

I've updated the logic to cater for backward compatibility.

https://github.com/hashicorp/consul-k8s/blob/2942824bd1154093afd059f213a001230812862b/control-plane/connect-inject/endpoints_controller.go#L822-L826

Please let me know what you think.

Ganeshrockz: Looks good to me

Ganeshrockz (approved):
Thanks for addressing the feedback. Do look into the failing tests as well

nvanthao:
Many thanks @Ganeshrockz,

I think the failing tests are acceptance tests that currently flaky/not relevant to this branch.

hashi-derek:
I believe there's a chance this issue could occur on all implementations, not just the 0.49.x branch. Is there a reason why we are patching just this version?

And most of the acceptance tests should be passing. We should not be ignoring them if they are all red.

nvanthao:
Hi @hashi-derek,

Apologies for the late reply!

I plan to create another PR for 1.x version after this PR is verified and merged.
We can't backport with single PR because there're significant changes for Endpoint controller between 0.49.x and 1.x.

For the acceptance tests, I notice all tests are failing with

```
flag provided but not defined: -consul-dataplane-image
```

which made me believe the test is not for 0.49.x since there is no dataplane in this version. Please let me know if this is correct understanding!

hashi-derek:
@nvanthao Oh, you're right. Someone must have accidentally backported something that is incorrectly referencing dataplane. That's odd -- I'll ask some people on my team. Also, I think 0.49.x is now considered to be out of support officially, so even if we add the bugfix to this branch, I don't think it will end up being officially published.

hashi-derek:
Adding @david-yu

aahel:
closing this in favour of https://github.com/hashicorp/consul-k8s/pull/3210. since we decided not to merge into 0.49

PR #3090, opened 2023-10-17
Backport of Fix ENT Tests Now that They Are Running Again 🏃 into release/1.0.x

zalimeni:
## Backport

This PR is manually generated from #3077 to be assessed for backporting.

The below text is copied from the body of the original PR.

---

ENT tests were just turned back on in CI. This revealed some failing test cases. This PR attempts to get everything in working order.


nathancoleman (approved):
LGTM. We do technically support this version for another couple of weeks. Thanks for backporting!

zalimeni:
> LGTM. We do technically support this version for another couple of weeks. Thanks for backporting!

@nathancoleman sure thing! Ran into failures in another PR and realized I needed this too.

PR #3103, opened 2023-10-19
test(v2): check for token deletion in acceptance test

DanStough:
Changes proposed in this PR:
- Add a check to the acceptance test that makes sure tokens are deleted.

How I've tested this PR: 🟢 

Checklist:
- [X] Tests added

### DIFF
--- a/acceptance/tests/mesh_v2/mesh_inject_test.go
+++ b/acceptance/tests/mesh_v2/mesh_inject_test.go
@@ -8,7 +8,9 @@ import (
 	"fmt"
 	"strconv"
 	"testing"
+	"time"
 
+	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


acceptance/tests/mesh_v2/mesh_inject_test.go
```diff
@@ -50,6 +52,26 @@ func TestMeshInject_MultiportService(t *testing.T) {
...
+				// because golang will execute them in reverse order i.e. the last registered
+				// cleanup function will be executed first.
+				t.Cleanup(func() {
+					retrier := &retry.Timer{Timeout: 5 * time.Minute, Wait: 1 * time.Second}
+					retry.RunWith(retrier, t, func(r *retry.R) {
+						tokens, _, err := consulClient.ACL().TokenList(nil)
+						require.NoError(r, err)
+						for _, token := range tokens {
+							require.NotContains(r, token.Description, multiport)
+							require.NotContains(r, token.Description, connhelper.StaticClientName)
```
zalimeni, line 69: ~ Is there any constant or ID we can borrow directly from the test to make sure this check doesn't drift from token contents -> lead to a false negative?

zalimeni (approved):
My initial gut feel was this might make more sense as a plain check at the end of the test, but it seems like if we want to retain the cleanup of resources that leads to the deletion we're checking for, it needs to also be part of the cleanup chain. TIL you can fail a test from `Cleanup`!

acceptance/tests/mesh_v2/mesh_inject_test.go
DanStough: It's a great question - unfortunately, the only thing that contains the login metadata is the description. There are also other tokens in the system so we can't check for an empty set or a specific number.

PR #3110, opened 2023-10-19
Mw/port dans changes from prepare 1 4 dev (#3104)

wilkermichael:
* build: remove last set-output from yaml

* add patch version for dataplane, now required.

---------

Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -332,8 +332,6 @@ jobs:
       - name: Copy LICENSE
         run:
           cp LICENSE ./control-plane
-      - uses: hashicorp/actions-docker-build@v1
-
       # This naming convention will be used ONLY for per-commit dev images
       - name: Set docker dev tag
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,3 +1,5 @@
+# Copyright (c) HashiCorp, Inc.
+
 name: build
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -320,6 +333,17 @@ jobs:
         run:
           cp LICENSE ./control-plane
       - uses: hashicorp/actions-docker-build@v1
```
DanStough, line 335: Is this step empty now?

wilkermichael: I guess, weird, this wasn't a change I made

wilkermichael: oh maybe I did... one sec

DanStough: But a new step with that action was added after the fact. Probably, there was a conflict here.

wilkermichael: yeah, I didn't get a conflict, I guess this is how it merged the cherry pick

wilkermichael: Ok, I think I just eliminate this uses, and the docker build gets done down lower

wilkermichael: that was a weird diff 😅

wilkermichael: @DanStough should be fixed

DanStough (approved):

PR #3116, opened 2023-10-23
Update Envoy

cthain:
This PR updates Envoy to the latest patch releases to address [CVE-2023-44487](https://github.com/envoyproxy/envoy/security/advisories/GHSA-jhv4-f7mr-xx76).

Note that the Envoy image version is only applicable to the legacy API Gateway since the Envoy sidecars are now packaged directly with Consul dataplane.

Release branch PRs:
- #3117
- #3118
- #3119
- #3120
- #3121

Checklist:
- [x] ~Tests added~

### DIFF
--- a/charts/consul/Chart.yaml
+++ b/charts/consul/Chart.yaml
@@ -22,7 +22,7 @@ annotations:
     - name: consul-dataplane
       image: docker.mirror.hashicorp.services/hashicorppreview/consul-dataplane:1.4-dev
     - name: envoy
-      image: envoyproxy/envoy:v1.26.2
+      image: envoyproxy/envoy:v1.25.11
   artifacthub.io/license: MPL-2.0
   artifacthub.io/links: |
     - name: Documentation
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -3233,7 +3233,7 @@ terminatingGateways:
   gateways:
     - name: terminating-gateway
 
-# [DEPRECATED] Use connectInject.apiGateway instead. This stanza will be removed with the release of Consul 1.17
+# [DEPRECATED] Use connectInject.apiGateway instead.
 # Configuration settings for the Consul API Gateway integration
 apiGateway:
   # When true the helm chart will install the Consul API Gateway controller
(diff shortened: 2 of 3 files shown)


charts/consul/Chart.yaml
```diff
@@ -22,7 +22,7 @@ annotations:
     - name: consul-dataplane
       image: docker.mirror.hashicorp.services/hashicorppreview/consul-dataplane:1.4-dev
     - name: envoy
-      image: envoyproxy/envoy:v1.26.2
+      image: envoyproxy/envoy:v1.25.11
```
cthain, line 25: This is an annotation that gets added to the Consul Helm chart on the artifacthub.io page. It should match the default Envoy version for the legacy `apiGateway.imageEnvoy` field.

charts/consul/values.yaml
```diff
@@ -3233,7 +3233,7 @@ terminatingGateways:
   gateways:
     - name: terminating-gateway
 
-# [DEPRECATED] Use connectInject.apiGateway instead. This stanza will be removed with the release of Consul 1.17
+# [DEPRECATED] Use connectInject.apiGateway instead.
```
cthain, line 3236: Per the service mesh team, removal of this field has been postponed to a later Consul release.

charts/consul/values.yaml
```diff
@@ -3248,7 +3248,7 @@ apiGateway:
   # The name (and tag) of the Envoy Docker image used for the
   # apiGateway. For other Consul compoenents, imageEnvoy has been replaced with Consul Dataplane.
   # @default: envoyproxy/envoy:<latest supported version>
-  imageEnvoy: "envoyproxy/envoy:v1.25.1"
+  imageEnvoy: "envoyproxy/envoy:v1.25.11"
```
cthain, line 3251: v1.25.x is the latest version that I saw for the legacy API Gateway. I'm not sure how/why the value on this branch is so out-of-date, but I've bumped it to the latest for consistency with the other branches.

david-yu (approved):

charts/consul/Chart.yaml
zalimeni: I'm surprised this is meant to match `apiGateway.imageEnvoy`, as I'd heard otherwise and also intuitively thought it was meant to match the current version in use by default (provided by `consul-dataplane`), as with the other annos. Thanks a bunch for digging this up.

Since the intended use seems to be representing the deprecated API Gateway Envoy version, can we also add a comment here to indicate that so it's not updated incorrectly in the future? I can also update the runbook.

zalimeni (approved):

charts/consul/Chart.yaml
zalimeni: Discussed offline; to minimize change for these bumps we can add this to the runbook instead or follow up later w/ a comment if needed.

PR #3118, opened 2023-10-23
release/1.3.x - Update Envoy

cthain:
This PR updates Envoy to the latest patch releases to address [CVE-2023-44487](https://github.com/envoyproxy/envoy/security/advisories/GHSA-jhv4-f7mr-xx76).

Note that the Envoy image version is only applicable to the legacy API Gateway since the Envoy sidecars are now packaged directly with Consul dataplane.

Checklist:
- [x] ~Tests added~

### DIFF
--- a/charts/consul/Chart.yaml
+++ b/charts/consul/Chart.yaml
@@ -22,7 +22,7 @@ annotations:
     - name: consul-dataplane
       image: docker.mirror.hashicorp.services/hashicorppreview/consul-dataplane:1.3-dev
     - name: envoy
-      image: envoyproxy/envoy:v1.26.2
+      image: envoyproxy/envoy:v1.25.11
   artifacthub.io/license: MPL-2.0
   artifacthub.io/links: |
     - name: Documentation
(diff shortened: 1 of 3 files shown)


david-yu (approved):

charts/consul/Chart.yaml
```diff
@@ -22,7 +22,7 @@ annotations:
     - name: consul-dataplane
       image: docker.mirror.hashicorp.services/hashicorppreview/consul-dataplane:1.3-dev
     - name: envoy
-      image: envoyproxy/envoy:v1.26.2
```
david-yu, line 25: Is there a reason why we downgraded from 1.26.2 to 1.25.11?

david-yu: Ah I see this was an inaccurate value that needed to be updated, please ignore.

zalimeni: Yup. x-post for posterity https://github.com/hashicorp/consul-k8s/pull/3116#discussion_r1368872800

zalimeni (approved):

PR #3122, opened 2023-10-23
Stop running api-gateway as root in vanilla k8s

nathancoleman:
Changes proposed in this PR:
- Set `runAsNonRoot`
- Stop explicitly setting `runAsUser`

How I've tested this PR:

Checklist:
- [ ] Tests added

### DIFF
--- a/control-plane/api-gateway/gatekeeper/dataplane.go
+++ b/control-plane/api-gateway/gatekeeper/dataplane.go
@@ -105,18 +105,14 @@ func consulDataplaneContainer(config common.HelmConfig, gcc v1alpha1.GatewayClas
 		container.Resources = *gcc.Spec.DeploymentSpec.Resources
 	}
 
-	// If running in vanilla K8s, run as root to allow binding to privileged ports;
-	// otherwise, allow the user to be assigned by OpenShift.
 	container.SecurityContext = &corev1.SecurityContext{
 		ReadOnlyRootFilesystem: pointer.Bool(true),
+		RunAsNonRoot:           pointer.Bool(true),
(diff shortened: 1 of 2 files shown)


control-plane/api-gateway/gatekeeper/dataplane.go
```diff
@@ -105,18 +105,15 @@ func consulDataplaneContainer(config common.HelmConfig, gcc v1alpha1.GatewayClas
 		container.Resources = *gcc.Spec.DeploymentSpec.Resources
 	}
 
-	// If running in vanilla K8s, run as root to allow binding to privileged ports;
-	// otherwise, allow the user to be assigned by OpenShift.
 	container.SecurityContext = &corev1.SecurityContext{
-		ReadOnlyRootFilesystem: pointer.Bool(true),
-		// Drop any Linux capabilities you'd get as root other than NET_BIND_SERVICE.
+		AllowPrivilegeEscalation: pointer.Bool(false),
```
nathancoleman, line 109: Setting this will force practitioners to use the `mapPrivilegedContainerPorts` setting that we added recently rather than binding to a privileged port such as `80` which works prior to this change.

nathancoleman: One thing to consider: Should we change the default value of `mapPrivilegedContainerPorts` to something other than `0`?

sarahalsmiller: Do we know what other people tend to do by default?

nathancoleman:
Closing as this is several years old at this point, and I no longer work on consul

PR #3065, opened 2023-10-09
Fix Gateway API CRDs

t-eckert:
It looks like the `copywrite` CLI deleted the Gateway CRDs we install. This would have broken us if it was released. I regenerated the external CRDs and added them to the Copywrite ignore given they are just pulled in from the Gateway SIG anyway and are not our IP.

Changes proposed in this PR:
- Regenerate external Gateway CRDs
- Remove errant .yml
- Add external CRDs to copywrite ignore

PR #3124, opened 2023-10-23
Remove copyright headers from external crds

nathancoleman:
**Context**
#3065 removed the copyright headers in the generated external CRD definitions because they are defined by the Gateway API Kubernetes SIG, not by HashiCorp.

#3019 then accidentally re-introduced the headers that were removed due to unfortunate timing.

These headers would have been re-introduced later anyway due to the fact that the exclude path in `.copywrite.hcl` was a little off.

In addition, the comment in `.copywrite.hcl` was a little misleading as the problem is caused by the _tool that generates the copyright headers_ as opposed to the copyright headers themselves.

**How I've tested this PR:**
- Visual inspection
- `$ make generate-external-crds` produces no diff

See above

**Checklist:**
- [ ] Tests added

### DIFF
--- a/.copywrite.hcl
+++ b/.copywrite.hcl
@@ -11,7 +11,7 @@ project {
 
     # ignoring charts templates as adding copyright headers breaks all tests
     "charts/consul/templates/**",
-    # we don't own these and copyright headers break them
+    # we don't own these and the tool that adds copyright headers breaks them
     "control-plane/config/crds/external/**",
 
   ]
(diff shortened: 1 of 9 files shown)


t-eckert (approved):
Great change! I think this comment is much clearer.

.copywrite.hcl
```diff
@@ -11,8 +11,8 @@ project {
 
     # ignoring charts templates as adding copyright headers breaks all tests
     "charts/consul/templates/**",
-    # we don't own these and copyright headers break them
-    "control-plane/config/crds/external/**",
+    # we don't own these and the tool that adds copyright headers breaks them
+    "control-plane/config/crd/external/**",
```
nathancoleman, line 15: This path did not match the location of the files below, so I updated it

PR #3132, opened 2023-10-25
WIP Replace api-gateway manual policy with templated policy

roncodingenthusiast:
Changes proposed in this PR:
- remove the manual creation of a policy and use templated policies  

- Folks more familiar with api-gateway could point out anything I am missing 

Checklist:
- [ ] Tests added

### DIFF
--- a/control-plane/api-gateway/cache/consul.go
+++ b/control-plane/api-gateway/cache/consul.go
@@ -4,12 +4,10 @@
 package cache
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"strings"
 	"sync"
-	"text/template"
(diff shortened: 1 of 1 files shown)


control-plane/api-gateway/cache/consul.go
```diff
@@ -23,33 +21,12 @@ import (
 	"github.com/hashicorp/consul/api"
 )
 
-func init() {
-	gatewayTpl = template.Must(template.New("root").Parse(strings.TrimSpace(gatewayRulesTpl)))
-}
+func init() {}
```
roncodingenthusiast, line 24: do i still need this init function?

PR #3137, opened 2023-10-26
Add refreshes and retries to server-acl-init job

curtbushko:
Changes proposed in this PR:
- When the server-acl-init job runs it boostraps ACL tokens and sets a bunch of things on the server up
- It gets the static IPs for the servers, creates a boostrap token and sets a policy using that IP
- Then consul-server-connection-manager runs and other ACLs get set up using the consul client
- Usually this is fine.
- But during upgrades the servers statefulsets can change and thus the server IPs change.
- Our code was set to retry over and over again and would eventually timeout after 10 minutes
- This does not work as it forces customers to run the upgrades several times in hopes that the server IPs do not change.
- I've created a wrapper around the consul client called DynamicClient that allows you to refresh the IPs using consul-server-connection-manager
- The DynamicClient will be useful in other place also.
- A customer reported a single error but in my testing I discovered about 5 other places where the IPs would be incorrect and we needed to refresh.
- The key areas are around 'untilSucceeds()' as these consul calls would fail forever

How I've tested this PR:

- Added a unit test to DynamicClient that simulates a bad server ip, a refresh with a good ip and a client call.
- The race condition does not occur all the time but I have been able to reproduce it manually by upgrading over and over again, which sometimes forces the servers to change IPs.

👀 

Checklist:
- [x] Tests added

### DIFF
--- a/control-plane/subcommand/server-acl-init/command.go
+++ b/control-plane/subcommand/server-acl-init/command.go
@@ -115,6 +115,9 @@ type Command struct {
 	ctx           context.Context
 	retryDuration time.Duration
 
+	// the amount of time to contact the Consul API before timing out
+	apiTimeoutDuration time.Duration
+
 	// log
 	log hclog.Logger
 
@@ -233,6 +236,12 @@ func (c *Command) init() {
 	if c.retryDuration == 0 {
 		c.retryDuration = 1 * time.Second
 	}
+
+	// Most of the API calls are in an infinite loop until the command cancels. This timeout
+	// allows us to refresh the server IPs so that calls will succeed.
+	if c.apiTimeoutDuration == 0 {
+		c.apiTimeoutDuration = 2 * time.Minute
+	}
 }
 
 func (c *Command) Synopsis() string { return synopsis }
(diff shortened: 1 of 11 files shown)


control-plane/subcommand/server-acl-init/command.go
```diff
@@ -1032,6 +1055,25 @@ func (c *Command) quitVaultAgent() {
...
+		if err != nil {
+			c.log.Error("Error resolving IP Address", "err", err)
+			return err
+		}
+		c.log.Info("Refreshing server IP addresses", "addresses", ipAddrs)
+		return nil
+	}, exponentialBackoffWithMaxInterval()); err != nil {
+		return nil, err
+	}
+	return ipAddrs, nil
```
DanStough, line 1074: Do we want the nested backoff/retry? There may be a case for resetting the parent loop, but it might be easier to reason about with one-level.

DanStough (approved):
I'm not sure how I feel about the nest backoff/retry, but otherwise looks great! Thank you!

control-plane/subcommand/server-acl-init/command.go
curtbushko: I am not a fan of the style of it but we are using a similar retry style during the bootstrap (line 330) and I went with that for consistency.

PR #3172, opened 2023-11-03
fix(controller): v2 pod controller errors for acl deletion

DanStough:
Changes proposed in this PR:
- Don't error out if a token is missing metadata.
- Make sure dataplane is injected with the correct metadata so that the ACL token is created with metadata on login.
- Explicitly call stop when in the init container when re-direction is enabled.

How I've tested this PR: manual K8S testing plus unit tests.

How I expect reviewers to test this PR: 👁️ 👃 👁️ 

Checklist:
- [X] Tests added


thisisnotashwin (approved):
nice work!

Issue #1815, opened 2022-12-21
Enable Leader Election for HA Mode


Dentrax:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Is your feature request related to a problem? Please describe.

Imagine you have running big Kubernetes cluster(s). And assume that one of the nodes that contain your consul-sync-catalog pod has crashed. What would be happened? Pod couldn't do its job properly until the broken node gets fixed. This is where we hit the need of leader election.

One of the missing crucial part of `consul-sync-catalog` is leader election. This is a must-have feature to make the deployment run with multiple replicas on multiple nodes. Otherwise, setting the ReplicaSet to more than one, resulting in data-race between sync-catalog pods due to inmem-sync-cache-inconsistency. (Each pod tries to sync simultaneously.) It's essential to make one pod is leader and others as follower.

See the related issue: https://github.com/hashicorp/consul-k8s/issues/58

#### Feature Description

* Implement [leaderelection](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection) client
  * One of the pods should be leader
  * It should set `leases.coordination.k8s.io`
* Ability to set more than `1` pods for deployment

This is **NOT** something that involves even deeper changes; Kubernetes already provides out-of-box `leaderelection` package. Implementing that would resolve the issue.

#### Use Case(s)

* To run `consul-sync-catalog` with multiple replicas on different nodes using `podAntiAffinity` assigned.

#### Contributions

`-`

/cc @david-yu @jacksontj

david-yu:
Thanks we'll have someone take a look and see if we can provide feedback.

Dentrax:
Hey @david-yu, did you have a chance to review this? Is there any progress or plan recently?

Dentrax:
Kind ping @david-yu

Dentrax:
We are running almost 1k+ consul-sync-catalog instances on _control plane_ nodes. And highly affecting due to this missing feature.

[...]

PR #3173, opened 2023-11-06
Enable leader election for Consul sync catalog

ThanhNguyen-Vungle:
Changes proposed in this PR:
- Enabling leader election for Consul catalog sync, following the discussion in #1815 

How it works:
- Number of replicas for sync catalog can be configured.
- There is only 1 leader at one time who holds the lease, the rest will stay idle and keep checking if current leader is still active, and start voting for a new leader when the current leader is no longer responsive.
- Implementation is built on top of the [leaderelection](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection) package.

How I've tested this PR:
- Run a local cluster using [kind](https://kind.sigs.k8s.io/), deploy this change there and observe the behavior of replicas under different condition - on startup, during voting, new leader is determined, new leader is gone (pod deleted for example).

- Same as above.

Checklist:
- [ ] Tests added

### DIFF
--- a/charts/consul/templates/sync-catalog-clusterrole.yaml
+++ b/charts/consul/templates/sync-catalog-clusterrole.yaml
@@ -45,4 +45,12 @@ rules:
   - get
   - list
   - watch
+- apiGroups: ["coordination.k8s.io"]
+  resources: ["leases"]
+  verbs: ["create"]
+- apiGroups: ["coordination.k8s.io"]
+  resources: ["leases"]
+  resourceNames:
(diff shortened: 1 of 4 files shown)


erulabs:
Heya @david-yu, could you take a peek at this? It works for me, and would be an *excellent* feature to land! I'd love to use this in prod without having to make my own cut!

tommynguyen-vungle:
Hi @david-yu, could we prioritise this since it's quite useful and address the pain point

tommynguyen-vungle:
cc @missylbytes

PR #3176, opened 2023-11-06
Allow to use mTLS to connect to Consul

7fELF:
- Add the necessary parameters allowing to connect to Consul using mutual TLS (client certificates)

This used to be possible with consul-k8s 0.26.0 using the `CONSUL_CLIENT_CERT`/`CONSUL_CLIENT_KEY` variables, but after attempting an upgrade to 1.xx I noticed it didn't work anymore. 

### How I've tested this PR:

I started catalog-sync with the environments variables `CONSUL_CLIENT_{CERT,KEY}_FILE` set to a valid certificate and key file in my testing Kubernetes+Consul environment. My Consul server has the `tls.defaults.verify_incoming` setting set to true. 

It manages to connect to the server without any issue. 

Generate a CA certificate and a client certificate for that CA (using, for instance, OpenSSL).

Start a Consul cluster with TLS enabled and the following settings:
- `tls.defaults.verify_incoming` set to true
- `tls.defaults.ca_file` set to your generated CA

Install catalog-sync in a test Kubernetes cluster using the client certificate you created earlier by pointing to its location using the `CONSUL_CLIENT_CERT_FILE` and `CONSUL_CLIENT_KEY_FILE` environment variables.

### Checklist:
- [x] Tests added


7fELF:
@thisisnotashwin @curtbushko Can you take a look? This is an easy fix for a regression from consul-k8s 0.x.x

david-yu:
Hi @7fELF We removed the requirement with Clients as of 1.0.x for Catalog Sync and instead use GRPC TLS to communicate back to the servers, since it removes a lot of issues related to deployment especially if the servers and clusters using sync catalog are not in the same LAN. Although it is not mTLS, we use ACL login from Catalog Sync to ensure that the API Calls are authorized.

7fELF:
Hello @david-yu, thank you for your answer.
I understand that you might want to make things simpler for people using the recommended deployment methods (using your guides and the helm chart). However, consul server supports mTLS for both the HTTP and GRPC so I think the client applications and tool should do the same.

7fELF:
Pretty please :pray:  @david-yu
This is already supported in the code, I'm only adding the flags to use it!

PR #3181, opened 2023-11-07
NET-5392 Replace dev dependencies w/ latest release tags

nathancoleman:
Some dependencies were previously pinned to utilize unreleased features of the Consul SDK.
This replaces those with the tagged releases containing those features.

In addition, there were some breaking changes in the API submodule that were never released but didn't yet have the consuming end updated in control-plane. I've addressed those as well.

**How I've tested this PR:**
- Robot tests pass
- Visual inspection

See above

**Checklist:**
- [ ] Tests added

### DIFF
--- a/control-plane/api-gateway/common/translation_test.go
+++ b/control-plane/api-gateway/common/translation_test.go
@@ -1428,10 +1428,10 @@ func TestTranslator_ToHTTPRoute(t *testing.T) {
 							Headers:    []api.HTTPHeaderFilter{},
 							URLRewrite: nil,
 							RetryFilter: &api.RetryFilter{
-								NumRetries:            pointer.Uint32(3),
+								NumRetries:            3,
 								RetryOn:               []string{"cancelled"},
 								RetryOnStatusCodes:    []uint32{500, 502},
-								RetryOnConnectFailure: pointer.Bool(false),
+								RetryOnConnectFailure: false,
 							},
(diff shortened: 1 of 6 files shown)


control-plane/api-gateway/common/translation_test.go
```diff
@@ -1428,10 +1428,10 @@ func TestTranslator_ToHTTPRoute(t *testing.T) {
 							Headers:    []api.HTTPHeaderFilter{},
 							URLRewrite: nil,
 							RetryFilter: &api.RetryFilter{
-								NumRetries:            pointer.Uint32(3),
+								NumRetries:            3,
 								RetryOn:               []string{"cancelled"},
 								RetryOnStatusCodes:    []uint32{500, 502},
-								RetryOnConnectFailure: pointer.Bool(false),
+								RetryOnConnectFailure: false,
 							},
```
DanStough, line 1435: Looks like the `api` module had a breaking change. I suppose we can stick with it until someone else runs into it.

DanStough (approved):

control-plane/api-gateway/common/translation_test.go
nathancoleman: It was technically a breaking change within a single release -- it was never released as a breaking change to users.

The consuming end was just missed until now. Still catching it before it goes out to customers

PR #3188, opened 2023-11-08
Add CRD for failover policy

thisisnotashwin:
Changes proposed in this PR:
- Add CRD for FailoverPolicy in catalog.consul.hashicorp.com
- Fix RBAC for existing CRDs and add permissions for the new CRD
- Fix an error where many of the webhooks were not registered with the mutating webhook.
- Add missing license headers.
- Going forward, we will no longer write validations for the CRDs. A following PR with drop validations on consul-k8s side since Consul persons aspects of this already. Additionally, we have future work in consul to expose RPC endpoints which will persons mutations and validations on the CRDs which will be invoked from the webhooks and therefore we will not need to perform them.

How I've tested this PR:
- deployed on cluster locally
- unit tests
- 👀 

Checklist:
- [x] Tests added


control-plane/go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/hashicorp/consul-k8s/control-plane/cni v0.0.0-20230825213844-4ea04860c5ed
 	github.com/hashicorp/consul-server-connection-manager v0.1.6
 	github.com/hashicorp/consul/api v1.26.1
-	github.com/hashicorp/consul/proto-public v0.5.1
+	github.com/hashicorp/consul/proto-public v0.1.2-0.20231107172454-028f1d8c502f
```
t-eckert, line 14: You should be able to rebase on `main` as of this morning to use a replace directive for this. I handled it in #3194. We want to use a replace directive instead of swapping this value because `api` depends on `proto-public` `v0.5.1` and you can run into some dependency errors if you just swap the values in place.

charts/consul/templates/crd-tcproutes-external.yaml
```diff
@@ -1,21 +1,18 @@
 {{- if and .Values.connectInject.enabled .Values.connectInject.apiGateway.manageExternalCRDs }}
```
ndhanushkodi, line 1: Are the license file changes intended? In a few places it's getting removed and a few places it's getting added. And should it be the other license?

cli/cmd/proxy/stats/command_test.go
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
```
ndhanushkodi, line 3: For example here- is this intended?

PR #3184, opened 2023-11-08
Fix consul-telemetry-collector deployment templates

jjti:
Changes proposed in this PR:
- remove `foo` arg
- remove the quotes around `"default"` auth login namespace that break auth 

How I've tested this PR:

I reproduced the behavior described by @david-yu 

I did a Helm install using a template that includes `global.acls.manageSystemACLs=true`: https://github.com/jjti/consul-experiments/blob/main/telemetry-collector/helm/consul.yaml

I `kubectl apply` a manually created deployment (based on a fixed version of helm's output): https://github.com/jjti/consul-experiments/blame/e2da8fe2d768472b549780f8eea2cc10e7da04e4/telemetry-collector/resources/consul-telemetry-collector.yaml#L231

I saw the telemetry collector start and push metrics:

```
# josh @ josh-XWGWHJ0JV2 in ~/GitHub/jjti/consul-exp/telemetry-collector on git:main x [21:21:02]
$ k get pods
NAME                                           READY   STATUS    RESTARTS   AGE
consul-connect-injector-74d6fb76b8-87pck       1/1     Running   0          15m
consul-server-0                                1/1     Running   0          15m
consul-telemetry-collector-9d858d459-5c6gq     2/2     Running   0          7m47s
consul-terminating-gateway-7f5877fdb8-rgtcj    1/1     Running   0          15m
consul-webhook-cert-manager-7c5bc5fb59-ln2m4   1/1     Running   0          15m
prometheus-server-695744d587-k9k6l             2/2     Running   0          15m
```

How I expect reviewers to test this PR:

Checklist:
- [ ] Tests added
- [ ] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #3192, opened 2023-11-08
Fix consul-telemetry-collector deployment when enableConsulNamespaces 

jjti:
### Issues

@david-yu reported multiple issues with the Consul Telemetry Collector deployment. A different PR fixed a couple of consul-dataplane args: https://github.com/hashicorp/consul-k8s/pull/3184

This attempts to fix some of the connect-inject issues with namespaces:

> Once I create the namespace manually, the connect-init container is not able  find the service because there are a couple of things in the deployment template that I think is not right: ... should be the namespace where the OTEL is deployed, that by default is consul. It is taking the namespace used the the SyncCatalog which is wrong because I am using just the connectInject.consulNamespaces.mirroringK8s=true, so the SyncCatalog is false and that value is default then

My guess at what happened when this was written was, we were trying to replicate the namespace selection logic of the Consul Dataplane Sidecar webhook: https://github.com/hashicorp/consul-k8s/blob/b84339050bb2c4b62b60cec96275f74952b0ac9d/control-plane/connect-inject/webhook/consul_dataplane_sidecar.go#L237

We run into issues tho, doing that. Partly because that's using `syncCatalog`. But also because, if CTC is the first service in a namespace, we fail to inject it because the namespace does not exist.

### Changes

This PR is an attempt at fixing some of those namespace/partition issues.

1. adds a new pod label. If set, we ignore all other namespace logic and just deploy the Consul Telemetry Collector to the default namespace. This is the same as what mesh-gateway does: https://github.com/hashicorp/consul-k8s/blob/e11143561551f86e0e7897a3f2830daa67ef5d74/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go#L692
2. adds `-service-partition` arg to the consul-dataplane container

#### How I've tested this PR:

- bats
- the scenarios below

##### Non-default namespaces

- I set `connectInject.consulNamespaces.mirroringK8s=true` and deployed the consul-telemetry-collector to a `namespace: test-namespace` after creating a deployment from this template change: https://github.com/jjti/consul-experiments/blob/ec80caaeded716a1dfbc34343d0da5db6142ad19/telemetry-collector/resources/consul-telemetry-collector.yaml#L33

```bash
$ k get pods --namespace test-namespace
NAME                                           READY   STATUS    RESTARTS   AGE
consul-connect-injector-74c6b4b7dc-5twg9       1/1     Running   0          14m
...
consul-telemetry-collector-68b7b446fc-nzwgb    2/2     Running   0          2m10s
```

##### Non-default partitions

- I followed this guide shared by @david-yu: https://developer.hashicorp.com/consul/docs/enterprise/admin-partitions#install-on-the-non-default-partition-clusters-running-workloads
- I deployed the consul-telemetry-collector into a non-default partition (in addition to the default partition) using a template created with the changes here: https://github.com/jjti/consul-experiments/blob/main/telemetry-collector/resources/consul-telemetry-collector-test.yaml

#### How I expect reviewers to test this PR:

- fact check my assumptions + approaches about consul-k8s-auth-method
- fact check whether I missed some combination of logic in connect-inject so this will fail with some combo of namespace/partition settings that I'm ignorant of

Checklist:
- [] Tests added


charts/consul/templates/telemetry-collector-deployment.yaml
```diff
@@ -93,24 +93,31 @@ spec:
...
+            value: {{ .Values.global.datacenter }}
           - name: CONSUL_LOGIN_META
             value: "component=consul-telemetry-collector,pod=$(NAMESPACE)/$(POD_NAME)"
-          {{- end }}
-          - name: CONSUL_NODE_NAME
-            value: $(NODE_NAME)-virtual
-          {{- include "consul.consulK8sConsulServerEnvVars" . | nindent 10 }}
           {{- if .Values.global.enableConsulNamespaces }}
-          - name: CONSUL_NAMESPACE
-            value: {{ .Values.syncCatalog.consulNamespaces.consulDestinationNamespace }}
```
jjti, line 107: this removes all ref to `.Values.syncCatalog.consulNamespaces.consulDestinationNamespace`. AFAICT, `.Values.syncCatalog.enabled` might be `false` but we'll still insert it here. My guess is that this is a bug that was from copying this setting from the Sync Catalog Deployment: https://github.com/hashicorp/consul-k8s/blob/768e29d2b7693e7900f75f7d0be1fd079d0c1244/charts/consul/templates/sync-catalog-deployment.yaml#L177

charts/consul/templates/telemetry-collector-deployment.yaml
```diff
@@ -285,23 +298,30 @@ spec:
...
+          - -service-namespace={{ .Values.connectInject.consulNamespaces.mirroringK8SPrefix }}{{ .Release.Namespace }}
+          {{- else }}
+          - -service-namespace={{ .Values.connectInject.consulNamespaces.consulDestinationNamespace }}
+          {{- end }}
+          {{- if .Values.global.acls.manageSystemACLs }}
+          {{- if .Values.connectInject.consulNamespaces.mirroringK8S }}
           - -login-namespace=default
           {{- else }}
-          - -login-namespace={{ .Values.syncCatalog.consulNamespaces.consulDestinationNamespace }}
+          - -login-namespace={{ .Values.connectInject.consulNamespaces.consulDestinationNamespace }}
```
jjti, line 313: this is one of the fixes: login-name is from connectInject destination namespace not syncCatalog

david-yu:
@jjti When you say you failed to test with non-default partitions is this because non-default partition support is not available in the telemetry collector? Or is it because you would need to help with partitions on K8s?

jjti:
> @jjti When you say you failed to test with non-default partitions is this because non-default partition support is not available in the telemetry collector? Or is it because you would need to help with partitions on K8s?

At this point it's a me problem. If you have a ref values.yaml for that that'd be appreciated. It think it'll work, but the changes here are copied from adjacent deployments like ingress-gateways

charts/consul/templates/telemetry-collector-deployment.yaml
```diff
@@ -285,23 +298,30 @@ spec:
           - -credential-type=login
           - -login-bearer-token-path=/var/run/secrets/kubernetes.io/serviceaccount/token
           - -login-auth-method={{ template "consul.fullname" . }}-k8s-auth-method
+          {{- end }}
+          # service and login namespace
           {{- if .Values.global.enableConsulNamespaces }}
-          {{- if .Values.syncCatalog.consulNamespaces.mirroringK8S }}
+          {{- if .Values.connectInject.consulNamespaces.mirroringK8S }}
+          - -service-namespace={{ .Values.connectInject.consulNamespaces.mirroringK8SPrefix }}{{ .Release.Namespace }}
```
jjti, line 305: change here to use k8s namespace prefixing like we do in the webhook

david-yu:
@jjti Does this help? https://developer.hashicorp.com/consul/docs/enterprise/admin-partitions#install-on-the-non-default-partition-clusters-running-workloads or do you need a writeup specifically written for HCP?

jjti:
> @jjti Does this help? https://developer.hashicorp.com/consul/docs/enterprise/admin-partitions#install-on-the-non-default-partition-clusters-running-workloads or do you need a writeup specifically written for HCP?

It does and thanks. I've been trying to hack helm templates together but lack the background knowledge to be efficient. Thanks! I'll draft this until that's done

PR #3194, opened 2023-11-09
Add replace directive in Go mod for Control Plane so that we pull in the latest Consul submodules

t-eckert:
This is a change that we should only keep until we publish the next version of the `proto-public` submodule. 

Changes proposed in this PR:
- Reference the latest `main` commit for the `proto-public` submodule.

How I've tested this PR:
I am now able to reference code that is in the latest version of the submodule.

### DIFF
--- a/control-plane/go.mod
+++ b/control-plane/go.mod
@@ -1,6 +1,7 @@
 module github.com/hashicorp/consul-k8s/control-plane
 
 // TODO: Remove this when the next version of the submodule is released.
+// We need to use a replace directive instead of directly pinning because `api` requires version `0.5.1` and will clobber the pin, but not the replace directive.
 replace github.com/hashicorp/consul/proto-public => github.com/hashicorp/consul/proto-public v0.1.2-0.20231109213314-40c57f10a0fb
 
 require (
--- a/control-plane/go.mod
+++ b/control-plane/go.mod
@@ -1,5 +1,8 @@
 module github.com/hashicorp/consul-k8s/control-plane
 
+// TODO: Remove this when the next version of the submodule is released.
+replace github.com/hashicorp/consul/proto-public => github.com/hashicorp/consul/proto-public v0.1.2-0.20231109213314-40c57f10a0fb
+
 require (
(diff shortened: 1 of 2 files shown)


control-plane/go.mod
```diff
@@ -1,5 +1,8 @@
 module github.com/hashicorp/consul-k8s/control-plane
 
+// TODO: Remove this when the next version of the submodule is released.
+replace github.com/hashicorp/consul/proto-public => github.com/hashicorp/consul/proto-public v0.1.2-0.20231109213314-40c57f10a0fb
+
```
zalimeni, line 5: I think we've typically targeted SHA versions as direct dependencies, rather than via `replace` directives, when an unreleased version of a sub is needed (there's a few recent examples in git history). ❓ Is there a reason for using `replace` in this case?

FWIW, I think @thisisnotashwin is planning to do something similar in https://github.com/hashicorp/consul-k8s/pull/3188/files#diff-b85c459442bcfeb4bb66362e683f1648d6d097e9478a95a43c40bffc82769b4aR14

t-eckert: That's fine, I thought a replace model was easier to remove when we get to the next submodule release.

nathancoleman: `replace` has been used most recently -- I undid it as part of the 1.3.0 release. The reason was lost from the comment that previously explained how `/api` will overwrite `proto-public` if you just do a simple pin

https://github.com/hashicorp/consul-k8s/commit/0f81470bde7c00d6da419068ef23a15f60e436cb#diff-b85c459442bcfeb4bb66362e683f1648d6d097e9478a95a43c40bffc82769b4a

t-eckert: Oh yeah, if you just pin right now and then do a `tidy` it will clobber the pinned version and you will get this: 

image

zalimeni: Ah, I forgot about that comment. Thanks y'all! Appreciate the restoration for clarity (clearly I need it... 😆)

zalimeni (approved):

PR #3200, opened 2023-11-10
NET-6406 Adds GatewayClassConfig and MeshGateway resources to the gateway-reso…

missylbytes:
…urces-configmap.yaml in the Helm chart

Changes proposed in this PR:
- This adds the MeshGateway and GatewayClassConfigs to the gateway-resources-configmap.yaml. 

How I've tested this PR:
Added bats tests, also manual testing.

```
helm template consul . -s templates/gateway-resources-configmap.yaml --set 'connectInject.enabled=true' --set 'global.experiments[0]=resource-apis' --set 'ui.enabled=false'
``` 
and check the output

Checklist:
- [x] Tests added

### DIFF
--- a/charts/consul/templates/gateway-resources-configmap.yaml
+++ b/charts/consul/templates/gateway-resources-configmap.yaml
@@ -19,13 +19,17 @@ data:
   {{- if (mustHas "resource-apis" .Values.global.experiments) }}
   config.yaml: |
     gatewayClassConfigs:
-      - apiVersion: consul.hashicorp.com/v2beta1
+      - apiVersion: mesh.consul.hashicorp.com/v2beta1
         metadata:
           name: consul-mesh-gateway
           namespace: {{ .Release.Namespace }}
-        deployment:
-          resources:
-            {{ .Values.meshGateway.resources }}
+        kind: gatewayClassConfig
+        spec:
+          deployment:
+            resources:
+              {{ .Values.meshGateway.resources }}
+          nodeSelector: {{ .Values.meshGateway.nodeSelector }}
+          serviceType: {{ .Values.meshGateway.service.type }}
     meshGateways:
       - name: mesh-gateway
         spec:
--- a/charts/consul/templates/gateway-resources-configmap.yaml
+++ b/charts/consul/templates/gateway-resources-configmap.yaml
@@ -16,4 +16,19 @@ data:
   resources.json: |
     {{ toJson .Values.connectInject.apiGateway.managedGatewayClass.resources }}
   {{- end }}
+  {{- if (mustHas "resource-apis" .Values.global.experiments) }}
+  config.yaml: |
+    gatewayClassConfigs:
+      - apiVersion: consul.hashicorp.com/v2beta1
+        metadata:
+          name: consul-mesh-gateway
+          namespace: {{ .Release.Namespace }}
+        deployment:
+          resources:
+            {{ .Values.meshGateway.resources }}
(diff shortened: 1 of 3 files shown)


charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -16,4 +16,19 @@ data:
   resources.json: |
     {{ toJson .Values.connectInject.apiGateway.managedGatewayClass.resources }}
   {{- end }}
+  {{- if (mustHas "resource-apis" .Values.global.experiments) }}
+  config.yaml: |
+    gatewayClassConfigs:
+      - apiVersion: consul.hashicorp.com/v2beta1
```
nathancoleman, line 22: I believe this will be `mesh.consul.hashicorp...`

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -16,4 +16,19 @@ data:
...
+        metadata:
+          name: consul-mesh-gateway
+          namespace: {{ .Release.Namespace }}
+        deployment:
+          resources:
+            {{ .Values.meshGateway.resources }}
```
nathancoleman, line 28: Is the intention that the entire resource should be present here? If so, this should be `spec.deployment` and have a `kind: GatewayClassConfig`

nathancoleman: I'm thinking it's likely I just overlooked ^ in the ticket 😆

charts/consul/templates/gateway-resources-job.yaml
```diff
@@ -51,7 +51,7 @@ spec:
             - -heritage={{ .Release.Service }}
             - -release-name={{ .Release.Name }}
             - -component=api-gateway
-            {{- if .Values.apiGateway.enabled }} # Overide values from the old stanza. To be removed in 1.17 (t-eckert 2023-05-19)
+            {{- if .Values.apiGateway.enabled }} # Override values from the old stanza. To be removed after ~1.18 (t-eckert 2023-05-19) NET-6263
```
nathancoleman, line 54: 🏆

charts/consul/test/unit/gateway-resources-configmap.bats
nathancoleman, line 1: Nice work adding these tests!

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -16,4 +16,19 @@ data:
...
+    meshGateways:
+      - name: mesh-gateway
+        spec:
+          gatewayClassName: consul-mesh-gateway
```
jm96441n, line 32: is this needed to eventually consolidate the gateways?

nathancoleman: Yep, we're aiming to make `MeshGateway` function as closely as possible to `Gateway` without actually using the `Gateway` resource with its current limations

charts/consul/templates/gateway-resources-configmap.yaml
missylbytes: Yes I missed that, but I think I will leave some of the other fields out until they are needed for development. IRT maybe modifying what the GatewayClassConfig looks like

missylbytes: Updated

nathancoleman (approved):

PR #3202, opened 2023-11-10
Add validation to account for type change

sarahalsmiller:
Changes proposed in this PR:
- At some point prior to release the types on filters changed from time.Duration to metav1.Duration
- This is a perfectly valid change, but it modified the behavior of how the CRD was being parsed from as an integer to a string. This feature is not broken, but is currently fragile based on user input. This adds additional safeguards to make sure the user won't accidentally pass anything that will cause the controller to break. 
- https://github.com/hashicorp/consul/pull/19601 This PR updates the docs to match the behavior change as well. 

How I've tested this PR:
- Built locally and applied new filters ensuring that the validation failed

- CI passes. 

Checklist:
- [ ] Tests added


nathancoleman (approved):
Excellent find!

PR #3204, opened 2023-11-10
NET-6401 Stub MeshGateway controller

nathancoleman:
Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added

### DIFF
--- a/charts/consul/templates/connect-inject-clusterrole.yaml
+++ b/charts/consul/templates/connect-inject-clusterrole.yaml
@@ -98,6 +98,7 @@ rules:
   resources:
     - grpcroutes
     - httproutes
+    - meshgateways
     - tcproutes
     - proxyconfigurations
   verbs:
@@ -113,6 +114,7 @@ rules:
   resources:
     - grpcroutes/status
     - httproutes/status
+    - meshgateways/status
     - tcproutes/status
     - proxyconfigurations/status
   verbs:
--- a/control-plane/config-entries/controllersv2/mesh_gateway_controller.go
+++ b/control-plane/config-entries/controllersv2/mesh_gateway_controller.go
@@ -1,0 +1,44 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package controllersv2
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
+)
+
+// MeshGatewayController reconciles a MeshGateway object.
+type MeshGatewayController struct {
+	client.Client
+	Log                  logr.Logger
+	Scheme               *runtime.Scheme
+	MeshConfigController *MeshConfigController
+}
+
+// +kubebuilder:rbac:groups=mesh.consul.hashicorp.com,resources=meshgateway,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=mesh.consul.hashicorp.com,resources=meshgateway/status,verbs=get;update;patch
+
+func (r *MeshGatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// TODO NET-6392 NET-6393 NET-6394 NET-6395
+	return r.MeshConfigController.ReconcileEntry(ctx, r, req, &meshv2beta1.MeshGateway{})
+}
+
+func (r *MeshGatewayController) Logger(name types.NamespacedName) logr.Logger {
+	return r.Log.WithValues("request", name)
+}
+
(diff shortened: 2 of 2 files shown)


control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -0,0 +1,44 @@
...
+// MeshGatewayController reconciles a MeshGateway object.
+type MeshGatewayController struct {
+	client.Client
+	Log                  logr.Logger
+	Scheme               *runtime.Scheme
+	MeshConfigController *MeshConfigController
+}
+
+// +kubebuilder:rbac:groups=mesh.consul.hashicorp.com,resources=meshgateway,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=mesh.consul.hashicorp.com,resources=meshgateway/status,verbs=get;update;patch
```
nathancoleman, line 27: I'm not sure if these actually do anything anymore, but all of the other controllers in this package have them.
They imply they'd be used to generate the `ClusterRole` for the connect-inject-controller, but the `ClusterRole` doesn't appear to be generated, and I can't find any command that would do so either.

Maybe @thisisnotashwin knows?

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (r *MeshGatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// TODO NET-6392 NET-6393 NET-6394 NET-6395
```
nathancoleman, line 30: Leaving a trail of breadcrumbs for future implementation

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -0,0 +1,44 @@
...
+	return r.MeshConfigController.ReconcileEntry(ctx, r, req, &meshv2beta1.MeshGateway{})
```
nathancoleman, line 31: This function is interesting to dig into. It essentially takes the `spec` of the CRD, wraps it into a resource, and writes that resource through the Consul API

charts/consul/templates/connect-inject-clusterrole.yaml
```diff
@@ -98,6 +98,7 @@ rules:
   resources:
     - grpcroutes
     - httproutes
+    - meshgateways
```
nathancoleman, line 101: Will add bats tests for this before merge, but about to drop off for the weekend

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (r *MeshGatewayController) Logger(name types.NamespacedName) logr.Logger {
+	return r.Log.WithValues("request", name)
```
jm96441n, line 35: minor (and probably not huge since this is a stub) but should the key here be `"mesh-gateway"` rather than `"request"`?

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
thisisnotashwin: yeah.. so this does create/updates the controller in the `config` folder under `control-plane` but for the most part, the RBAC is self managed. my only argument for continuing to keep it is to sanity check if we have in fact updated the RBAC for connect-inject correctly, but there are other ways we can tackle that. 

my only request would be that if we do delete those two lines, can we do there for all the V2 CRDs as well to avoid confusion for anybody trying to reference these files?

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
nathancoleman: This mimics what happens today for `httproute`, `tcproute`, and `grpcroute`. Gonna leave this as a broader change for a separate PR if that's cool with you @jm96441n 

https://github.com/hashicorp/consul-k8s/blob/f4ccbaa9bc9152722fdb8ff032220f1bc95475e2/control-plane/config-entries/controllersv2/http_route_controller.go#L34

https://github.com/hashicorp/consul-k8s/blob/f4ccbaa9bc9152722fdb8ff032220f1bc95475e2/control-plane/config-entries/controllersv2/tcp_route_controller.go#L34

https://github.com/hashicorp/consul-k8s/blob/f4ccbaa9bc9152722fdb8ff032220f1bc95475e2/control-plane/config-entries/controllersv2/grpc_route_controller.go#L34

charts/consul/templates/connect-inject-clusterrole.yaml
nathancoleman: Added in https://github.com/hashicorp/consul-k8s/pull/3204/commits/59cc948afc181b5b7ccceea691ecf83939229a32

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
nathancoleman: I've regenerated the role in the `config` folder that @thisisnotashwin referenced

thisisnotashwin: thanks so much Nathan! Ill give this PR a look shortly.. sorry have been pulled in many directions the last couple days

missylbytes (approved):

PR #3210, opened 2023-11-14
fixed acl deletetion in endpoints controller

aahel:
Changes proposed in this PR:

When deleting ACL token in Endpoint Controller Reconcile loop, delete the correct tokens.
This will prevent this scenario from happening: we have 2 tokens for same pod name due to StatefulSet consistent pod naming.
This is done by utilizing the pod uid to filter the token instead of relying on pod name

How I've tested this PR:
Manually tested it  by creating a stateful set and deleting it and ensuring  all tokens get cleaned up correctly

Checklist:
- [x] Tests added


Ganeshrockz (approved):

control-plane/connect-inject/controllers/endpoints/endpoints_controller_test.go
```diff
@@ -715,6 +715,7 @@ func TestReconcileCreateEndpoint_MultiportService(t *testing.T) {
 						constants.MetaKeyKubeNS:  "default",
```
ndhanushkodi, line 715: Should we add a case for TestReconcileUpdateEndpoint or TestReconcileDeleteEndpoint or even a new test case that's a bit more end to end that ensures that with a non-empty pod uid meta, only acl token with matching pod uid is deleted?

ndhanushkodi (commented):
What do you think of this additional test case? Otherwise it looks great!

ndhanushkodi (approved):
Approving to not block on my comment.

control-plane/connect-inject/controllers/endpoints/endpoints_controller_test.go
aahel: updated the TestReconcileDeleteEndpoint to support deletion using pod uid

ndhanushkodi (approved):
looks great!!

aahel:
> looks great!!

found bug in the previous approach while manually testing it. so have changed how we are passing pod-uid. so that its gets set as part of token meta

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -120,13 +120,21 @@ func (w *MeshWebhook) consulDataplaneSidecar(namespace corev1.Namespace, pod cor
 			},
 			{
 				Name:  "DP_CREDENTIAL_LOGIN_META",
-				Value: "pod=$(POD_NAMESPACE)/$(POD_NAME),pod-uid=$(POD_UID)",
+				Value: "pod=$(POD_NAMESPACE)/$(POD_NAME)",
+			},
+			{
+				Name:  "DP_CREDENTIAL_LOGIN_META",
+				Value: "pod-uid=$(POD_UID)",
```
ndhanushkodi, line 127: Will this not overwrite the value of the environment variable setting the meta value `pod=`? Do you see both pieces of meta on the token after this change

aahel: removed it. do you know why we have used the same value for DP_CREDENTIAL_LOGIN_META and DP_CREDENTIAL_LOGIN_META1 ?

ndhanushkodi: Ah, so I'm seeing [here](https://github.com/hashicorp/consul-dataplane/blob/main/cmd/consul-dataplane/flags.go#L47-L56) that logic allows the environment variable to be specified with a suffix of {1-9} so that you can specify multiple values. So I think the key that makes this PR work is DP_CREDENTIAL_LOGIN_META2 where you've added the pod uid.

ndhanushkodi (approved):

PR #3218, opened 2023-11-15
Add validation that externalServers.hosts is not set to HCP-managed cluster's addresses when global.cloud.enabled

jjti:
### Overview

If we set `externalServers.hosts` to the public or private address of an HCP-manged cluster, we run into the following because we're hardcoding `-tls-server-name` to `server.{{ .Values.global.datacenter}}.{{ .Values.global.domain}}`:

> 2023-11-10T01:32:31.386Z [ERROR] consul-server-connection-manager: connection error: error="rpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for 517429dd-fcab-4fe6-bc4a-57750f69df7e.aws.hashicorp.cloud, consul-cluster.consul.e4065767-bd61-4f00-8b2b-81f631f7d4d1.aws.hashicorp.cloud, consul-cluster.private.consul.e4065767-bd61-4f00-8b2b-81f631f7d4d1.aws.hashicorp.cloud, not server.consul-cluster.consul\”"

We [hard-code `-tls-server-name` in many places](https://github.com/search?q=repo%3Ahashicorp%2Fconsul-k8s%20%22-tls-server-name%3Dserver.%7B%7B%20.Values.global.datacenter%7D%7D.%7B%7B%20.Values.global.domain%7D%7D%22&type=code) like:

https://github.com/hashicorp/consul-k8s/blob/4d7a1873b2c6049efc9769e3a03406db1a8f1380/charts/consul/templates/_helpers.tpl#L229-L231

- Add validation that users don't set `global.cloud.enabled` while also using `externalServers.hosts` that point at an HCP-managed cluster (any address that has the suffix ".hashicorp.cloud")
- Add note to `values.yaml` describing how `global.cloud` is really for _linking_ clusters, not for HCP-managed cloud clusters.

### How I've tested this PR:
- bats, new bats test, ran it locally

### Checklist:
- [x] Tests added


jjti:
I cc'ed @hanshasselberg who I hope can keep me honest w/ this. My understanding is `global.cloud.enabled` only makes sense when a set of servers in a consul helm deployment are being linked to HCP. `externalServers` implies there aren't any in the deployment

That assumption would be wrong if we used `global.cloud` for clients or anything else w/o servers

lkysow (changes requested):
You only need to put fail conditionals in one file. And then you also don't need a helper for it.

lkysow:
I see that `global.cloud` is used also for the partition-init job to set the tls server name. When using partitions you won't have servers enabled so I don't think this change works.

jjti:
From talking with @freddygv , wondering we can check the URL and see if it's HCP-managed. Eg check if it's `aws.hashicorp.cloud`: `https://consul-cluster.consul.e4065767-bd61-4f00-8b2b-81f631f7d4d1.aws.hashicorp.cloud`

freddygv:
The reason we can't allow `global.cloud.enabled` with HCP-managed clusters is that `global.cloud.enabled` will hardcode the `server.<datacenter>.<domain>` as the TLS server name to verify against. 

However, HCP managed clusters do no use that server name for TLS connections to the public gRPC port, which consul-dataplane uses. Those certs come from lets encrypt and will have some other name. That's what leading to the validation errors above.

jjti:
> You only need to put fail conditionals in one file. And then you also don't need a helper for it.

I moved this to a single spot at the top of `charts/consul/templates/connect-inject-deployment.yaml`

hanshasselberg (commented):
LGTM

hanshasselberg (approved):

lkysow (approved):
LGTM. Should we think about renaming `global.cloud` to `global.cloudLinked`? We can make it backwards compatible via templating.

PR #1440, opened 2022-08-22
Add named metrics port to envoy sidecar

hamishforbes:
Relates to #671 

Changes proposed in this PR:
- Adds a [containerPort](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#containerport-v1-core) configuration to the injected Envoy sidecar for the metrics port 
- TODO: A test! draft until i add that

The [targetPort](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#podmetricsendpoint) field on Prometheus Operator ports accepts a port **name**, using an integer port number is deprecated.

Without any named ports on the injected container it is not possible to configure a `ServiceMonitor` or `PodMonitor` resource (without using a deprecated feature) in order to scrape metrics from a Prometheus Operator managed Prometheus instance

How I've tested this PR:
I've been running this patch in production for nearly a year

How I expect reviewers to test this PR:
Confirm the injected envoy sidecar on a Connect enabled pod has a ports configuration:

```
  name: envoy-sidecar
  ports:
  - containerPort: 20200
    name: envoy-metrics
    protocol: TCP
```

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #3222, opened 2023-11-16
feat: add named prom port to dataplane sidecar

DanStough:
Changes proposed in this PR:
- This port of #1440 is updated for consul-dataplane and the v2 API. It adds a named port, `prometheus` to the container. This can be used for the Prometheus operator.

This could easily be backported to previous versions, but I figured there was no point in the risk, considering how long this issue has been around. Let me know if anyone disagrees. I will need to do it manually to remove the V2 changes.

Checklist:
- [X] Tests added


david-yu:
@DanStough I would backport to at least 1.3.0 since right now there is no option for folks to use this feature and they will have to wait until v2 is GA.

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -162,6 +162,15 @@ func (w *MeshWebhook) consulDataplaneSidecar(namespace corev1.Namespace, pod cor
 		container.VolumeMounts = append(container.VolumeMounts, volumeMounts...)
 	}
 
+	// Container Ports
+	ports, err := w.getContainerPorts(pod)
+	if err != nil {
+		return corev1.Container{}, err
+	}
+	if ports != nil {
+		container.Ports = ports
```
curtbushko, line 171: Doesn't this incorrectly wipe out all the ports already set? Should it be an append?

On line 149:

		container.Ports = append(container.Ports, corev1.ContainerPort{
			Name:          fmt.Sprintf("%s-%d", "proxy-health", mpi.serviceIndex),
			ContainerPort: int32(constants.ProxyDefaultHealthPort + mpi.serviceIndex),

curtbushko (changes requested):
Thanks Dan!  Looks pretty good but I think that I may? have found a bug?

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
DanStough: Great catch! 🙇

curtbushko (approved):
LGTM!

PR #3230, opened 2023-11-17
Reduce permissions of API gateway policy

roncodingenthusiast:
Changes proposed in this PR:
- The reason api-gateway previously required service write on everything because it would create intentions for any service that the gateway was configured to route to. This is no longer the case as we only need service write on the api-gateway
- This PR suggest only give service write on the API gateway 
- This is the first step into removing this completely to use templated policies

### DIFF
--- a/control-plane/api-gateway/cache/consul.go
+++ b/control-plane/api-gateway/cache/consul.go
@@ -29,6 +29,7 @@ func init() {
 
 type templateArgs struct {
 	EnableNamespaces bool
+	APIGatewayName   string
 }
 
 var (
@@ -41,7 +42,10 @@ mesh = "read"
 		node_prefix "" {
 			policy = "read"
 		}
-    service_prefix "" {
+		service_prefix "" {
+			policy = "read"
+		}
+    service "{{.APIGatewayName}}" {
       policy = "write"
     }
 {{- if .EnableNamespaces }}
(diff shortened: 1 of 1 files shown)


control-plane/api-gateway/cache/consul.go
```diff
@@ -29,6 +29,7 @@ func init() {
 
 type templateArgs struct {
 	EnableNamespaces bool
+	ApiGatewayName   string
```
nathancoleman, line 32:
```suggestion
	APIGatewayName   string
```

nathancoleman (commented):
This LGTM. Have you had a chance to test it yet? If not, I can help out with that

nathancoleman:
I have confirmed that this works for native Consul API Gateway using the setup [here](https://github.com/nathancoleman/consul-lab/tree/main/k8s/api-gateway/basic), resulting in:
```shell
$ http http://34.148.240.250 -j
HTTP/1.1 200 OK
content-length: 255
content-type: text/plain; charset=utf-8
date: Thu, 07 Dec 2023 20:00:41 GMT
server: envoy
x-envoy-upstream-service-time: 5

{
    "body": "Hello World",
    "code": 200,
    "duration": "146.643µs",
    "end_time": "2023-12-07T20:00:41.767760",
    "ip_addresses": [
        "10.40.2.8"
    ],
    "name": "backend",
    "start_time": "2023-12-07T20:00:41.767613",
    "type": "HTTP",
    "uri": "/"
}
```

I have also confirmed that everything still works for legacy Consul API Gateway using the setup [here](https://github.com/nathancoleman/consul-lab/tree/main/k8s/api-gateway/legacy), resulting in:
```shell
$ http http://104.196.141.50 -j
HTTP/1.1 200 OK
content-length: 256
content-type: text/plain; charset=utf-8
date: Thu, 07 Dec 2023 20:44:36 GMT
server: envoy
x-envoy-upstream-service-time: 10

{
    "body": "Hello World",
    "code": 200,
    "duration": "463.781µs",
    "end_time": "2023-12-07T20:44:36.868101",
    "ip_addresses": [
        "10.40.5.10"
    ],
    "name": "backend",
    "start_time": "2023-12-07T20:44:36.867637",
    "type": "HTTP",
    "uri": "/"
}
```

nathancoleman (approved):

PR #3237, opened 2023-11-21
[NET-6617] security: Bump github.com/golang-jwt/jwt/v4 to 4.5.0

zalimeni:
This version is accepted by Prisma/Twistlock, resolving scan results for issue PRISMA-2022-0270. Chosen over later versions to avoid a major version with breaking changes that is otherwise unnecessary.

Note that in practice this is a false positive (see https://github.com/golang-jwt/jwt/issues/258), but we should update the version to aid customers relying on scanners that flag it.

c.f. Consul change here: https://github.com/hashicorp/consul/pull/19705

Changes proposed in this PR:
- Resolves proprietary scanner reported vulnerability.

I've reviewed the changelog from `4.2.0` to `4.5.0` and don't see any evidence of breaking changes.

Checklist:
- [ ] Tests added

### DIFF
--- a/.changelog/3237.txt
+++ b/.changelog/3237.txt
@@ -1,0 +1,3 @@
+```release-note:security
+Update `github.com/golang-jwt/jwt/v4` to v4.5.0 to address [PRISMA-2022-0270](https://github.com/golang-jwt/jwt/issues/258).
+```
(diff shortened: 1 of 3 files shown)


curtbushko (approved):
LGTM

PR #3244, opened 2023-11-21
NET-6394 Create/update/delete ServiceAccount on MeshGateway reconcile

nathancoleman:
When reconciling a `MeshGateway` resource in the v2 API, create/update/delete a corresponding `ServiceAccount`.
Notably, we need to ensure we aren't updating or deleting resources that our controller didn't create simply because they have the expected name.

**How I've tested this PR:**
Added unit test coverage

See above

**Checklist:**
- [x] Tests added

### DIFF
--- a/control-plane/config-entries/controllersv2/mesh_gateway_controller.go
+++ b/control-plane/config-entries/controllersv2/mesh_gateway_controller.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
 	"github.com/hashicorp/consul-k8s/control-plane/gateways"
@@ -73,15 +74,12 @@ func (r *MeshGatewayController) SetupWithManager(mgr ctrl.Manager) error {
 func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Request, resource *meshv2beta1.MeshGateway) error {
 	builder := gateways.NewMeshGatewayBuilder(resource)
 
-	ifNew := func(ctx context.Context, object client.Object) error {
-		return r.Create(ctx, object)
+	upsertOp := func(ctx context.Context, _, object client.Object) error {
+		_, err := controllerutil.CreateOrUpdate(ctx, r.Client, object, func() error { return nil })
+		return err
 	}
 
-	ifExists := func(ctx context.Context, object client.Object) error {
-		return r.Update(ctx, object)
-	}
-
-	err := r.opIfNewOrOwned(ctx, resource, &corev1.ServiceAccount{}, builder.ServiceAccount(), ifNew, ifExists)
+	err := r.opIfNewOrOwned(ctx, resource, &corev1.ServiceAccount{}, builder.ServiceAccount(), upsertOp)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -69,11 +72,83 @@ func (r *MeshGatewayController) SetupWithManager(mgr ctrl.Manager) error {
...
-	// TODO NET-6392 NET-6393 NET-6394 NET-6395
-	return errors.New("onCreateUpdate not implemented")
+	builder := gateways.NewMeshGatewayBuilder(resource)
+
+	ifNew := func(ctx context.Context, object client.Object) error {
+		return r.Create(ctx, object)
+	}
+
+	ifExists := func(ctx context.Context, object client.Object) error {
+		return r.Update(ctx, object)
```
andrewstucki, line 82: Just wondering, but would we need to take into account things like managed kubernetes service annotation additions or anything that mutates some sort of metadata here? Not sure if any implementations mutate `ServiceAccount` objects or not... but I know there's some stuff added to `Service` objects by EKS and (I believe?) GKE that we need to ignore to avoid infinite reconciliation problems.

Accordingly, just wondering if maybe it's worthwhile to make the `ifExists` closure here take an `oldEntry, newEntry client.Object` argument so we can possibly do some merging of resources on an update?

nathancoleman: Good callout! I believe we only do this merging logic for [Service](https://github.com/hashicorp/consul-k8s/blob/71cdbc24fa8ad3bf3571b4c71744a5171cafdd50/control-plane/api-gateway/gatekeeper/service.go#L37-L43) and [Deployment](https://github.com/hashicorp/consul-k8s/blob/085c812761439c0a1be774dc7158eb441901099f/control-plane/api-gateway/gatekeeper/deployment.go#L58-L64) today.

Tomorrow, I'll make sure we have a path forward on that. Might wait to actually introduce that code in the PR where it's needed.

nathancoleman: Wound up modifying the operation func to receive the old object and the new object to allow for merging for other resources @andrewstucki

andrewstucki (commented):
So, the only thing I really see missing here is triggering reconciliation on owned object changes. For that I think you'll probably need to extend

https://github.com/hashicorp/consul-k8s/blob/66382615a81b18418b65859052a5f7f06d8cbbac/control-plane/config-entries/controllersv2/meshconfig_controller.go#L209

and add some ability to specify additional `Owns` references (maybe just make a closure for doing additional arbitrary resource watch configuration before calling `Complete`?), similar to how we do in our current impl:

https://github.com/hashicorp/consul-k8s/blob/66382615a81b18418b65859052a5f7f06d8cbbac/control-plane/api-gateway/controllers/gateway_controller.go#L420-L423

otherwise any changes to the managed `ServiceAccount` won't re-trigger reconciliation and you'll never hit the `ifExists` branch and will get configuration drift.

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -69,11 +72,89 @@ func (r *MeshGatewayController) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	return nil
 }
 
 func (r *MeshGatewayController) onDelete(ctx context.Context, req ctrl.Request, resource *meshv2beta1.MeshGateway) error {
-	// TODO NET-6392 NET-6393 NET-6394 NET-6395
-	return errors.New("onDelete not implemented")
+	builder := gateways.NewMeshGatewayBuilder(resource)
+
+	deleteOp := func(ctx context.Context, _, object client.Object) error {
```
andrewstucki, line 95: I'm thinking we can just let the garbage collector clean these up? By setting the owner reference field, if the owner is deleted, these will actually get cleaned up automatically. If we're cool with just letting the GC do its work, then the only thing we have to worry about on deletion is cleaning up any side-effects that we may have done in our controller upsert path.

nathancoleman: My assumption was that there's a reason we do explicit cleanup in our v1 implementation [here](https://github.com/hashicorp/consul-k8s/blob/671675d49fbca59272fbbd54becd6366620704a5/control-plane/api-gateway/gatekeeper/gatekeeper.go#L60-L86), but I don't know what that might have been. Any ideas?

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -65,15 +68,96 @@ func (r *MeshGatewayController) UpdateStatus(ctx context.Context, obj client.Obj
...
+	if err != nil {
+		return err
+	}
+
+	// TODO NET-6392 NET-6393 NET-6395
+
+	return nil
+}
+
+// ownedObjectOp represent an operation that needs to be applied
```
missylbytes, line 112: nit `represents`

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
nathancoleman: I've dropped the explicit deletes in favor of allowing the k8s garbage collector to do its thing.
If we discover a reason we need to delete explicitly like we do in the `gatekeeper` package today, then we can add the explicit deletion at a later date.

nathancoleman:
The enterprise control-plane tests are failing because our self-hosted macos runner doesn't have `yarn` installed. Working to resolve this outside of this PR.

PR #3254, opened 2023-11-22
[NET-6462] Modify gateway resources job for class and class config

jm96441n:
Changes proposed in this PR:
- Modify Gateway Resources job to create gateway class and gateway class config when mesh gateway is enabled

How I've tested this PR:
:bat: :bat: 
:eye: 

:eye: 

Checklist:
- [ ] Tests added

### DIFF
--- a/charts/consul/templates/gateway-resources-job.yaml
+++ b/charts/consul/templates/gateway-resources-job.yaml
@@ -1,4 +1,4 @@
-{{- if or .Values.connectInject.enabled .Values.meshGateway.enabled }}
+{{- if .Values.connectInject.enabled }}
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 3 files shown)


charts/consul/templates/gateway-resources-job.yaml
```diff
@@ -113,6 +113,34 @@ spec:
             - name: config
               mountPath: /consul/config
               readOnly: true
+        {{ if .Values.meshGateway.enabled }}
+        - name: mesh-gateway-resources
```
andrewstucki, line 117: So, all of this stuff should probably just be passed as conditional flags to the job in the above  container definition instead of making a secondary container definition. You may also want to adjust the flag names on the job accordingly since this is now for both mesh stuff and api gateway stuff, which I'd imagine means you may want to prefix any api gateway specific flags with something like `api-gateway`, so something like `-service-type` would be `-api-gateway-service-type`. Mesh gateway specific stuff could just be prefixed with `mesh-gateway` and then anything shared wouldn't have a prefix.

andrewstucki (changes requested):
So, gave some feedback on the helm chart stuff. Additionally, what you'll want to do is modify the actual job go code to do something similar to:

https://github.com/hashicorp/consul-k8s/blob/98d6b90d56ee4176f25ff749e1e52fb3c7adf5fb/control-plane/subcommand/gateway-resources/command.go#L209-L248

But with the new CRDs we're shipping that aren't part of the upstream spec.

PR #3283, opened 2023-11-29
NET-6722 Rename MeshConfigController to ConsulResourceController

t-eckert:
The names `MeshConfigController` and `MeshConfig` were initially chosen to represent the generic Consul Resource mapping from Kubernetes into Consul.

Previously, these resources were referred to as "config-entries". With the upcoming implementation of Mesh Gateways V2, there was a name conflict with that implementation's `MeshConfiguration` resource and `MeshConfigurationController`. 

It was decided that the existing generic `MeshConfig` usage be replaced with `ConsulResource`. This PR adopts that naming scheme.

This change is not user-facing.

Changes proposed in this PR:
- Refactor MeshConfigController to ConsulResouceController
- Rename controller file
- Rename MeshConfig to ConsulResource
- Rename Controller to ResourceController

How I've tested this PR:

- Ensuring existing tests continue to pass.

Checklist:
- [ ] Tests added


control-plane/api/common/consul_resource_webhook_test.go
```diff
@@ -55,7 +55,7 @@ func TestValidateMeshConfig(t *testing.T) {
 			expErrMessage: "invalid",
 		},
 		"duplicate name": {
-			existingResources: []MeshConfig{&mockMeshConfig{
+			existingResources: []ConsulResource{&mockMeshConfig{
```
thisisnotashwin, line 58: in the spirit of completeness, would you mind updating the name of the mock as well?

control-plane/subcommand/inject-connect/v2controllers.go
```diff
@@ -127,80 +127,80 @@ func (c *Command) configureV2Controllers(ctx context.Context, mgr manager.Manage
 		}
 	}
 
-	meshConfigReconciler := &controllersv2.MeshConfigController{
+	meshConfigReconciler := &controllersv2.ConsulResourceController{
```
thisisnotashwin, line 130: can we rename the var to `consulResourceReconciler` as well

control-plane/config-entries/controllersv2/gateway_class_config_controller.go
```diff
@@ -18,16 +18,16 @@ import (
 // GatewayClassConfigController reconciles a GatewayClassConfig object.
 type GatewayClassConfigController struct {
 	client.Client
-	Log                  logr.Logger
-	Scheme               *runtime.Scheme
-	MeshConfigController *MeshConfigController
+	Log                      logr.Logger
+	Scheme                   *runtime.Scheme
+	ConsulResourceController *ConsulResourceController
```
thisisnotashwin, line 23: nit: rename the var in the struct `Controller` to simplify its usage since it is not a conflicting name here? We can do this for all the Controllers.

thisisnotashwin (commented):
This looks really good!! thanks for taking care of it so quick. i left a few comments inline.

control-plane/subcommand/inject-connect/v2controllers.go
t-eckert: Sure thing. Is there a reason this is referred to as "Reconciler" instead of "Controller"?

control-plane/config-entries/controllersv2/gateway_class_config_controller.go
t-eckert: I think that's a good nit.

thisisnotashwin:
looks like there are some linter errors that need cleaning up.. based on the cursory look, it seems fairly straightforward. let me know if you need hellp with those

missylbytes (approved):
LGTM

control-plane/subcommand/inject-connect/v2controllers.go
t-eckert: I gotcha, I'll change it to controller just for clarity.

PR #3285, opened 2023-11-29
Allow meshGateway.enabled when resource-apis experiment is enabled

nathancoleman:
Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added


NiniOak (commented):
Code looks good, I'm going to manually test the change manually before smashing the ✅ button

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -17,6 +17,7 @@ data:
     {{ toJson .Values.connectInject.apiGateway.managedGatewayClass.resources }}
   {{- end }}
   {{- if (mustHas "resource-apis" .Values.global.experiments) }}
+  {{- if .Values.meshGateway.enabled }}
```
jm96441n, line 20: should we just `and` these two statements rather than two `if` statements?

NiniOak: +1

nathancoleman: 😆 I considered that, but the one combined `if` winds up being harder to digest IMHO. If it's something you feel strongly about, I'm down to change it

jm96441n: I don't feel particularly strongly either way tbh, not combining them might make it easier to find all places where the experiments check is used and remove them when we need to do that later

NiniOak: Here's my +1 for using `and` for both statements, .`Values.meshGateway.enabled` I'm assuming based on my manual tests cannot be used to enable v2 without `  experiments: - "resource-apis"` added to the yaml, so I think we should add the check here to prevent runtime errors

NiniOak (approved):
non blocking comment - looks good otherwise.

charts/consul/templates/gateway-resources-configmap.yaml
nathancoleman: It's slightly different: `experiments: [resource-apis]` enables v2. `meshGateway.enabled` can be set in v1 or v2

PR #3289, opened 2023-11-30
Use retry test context within retry blocks

t-eckert:
- Use my modified Retry as a dependency
- Use Terratest's TestingT interface'
- Fix cli_cluster
- Fix helm_cluster
- Fix deploy
- Fix helpers
- Fix kubectl
- Fix connect_helper and port_forward
- Fix vault_cluster
- Undo changes to Logf
- Change t to r throughout multiple files
- Fix command_test

Changes proposed in this PR:
-
-

How I've tested this PR:

Checklist:
- [ ] Tests added


control-plane/go.mod
```diff
@@ -4,6 +4,8 @@ module github.com/hashicorp/consul-k8s/control-plane
 // We need to use a replace directive instead of directly pinning because `api` requires version `0.5.1` and will clobber the pin, but not the replace directive.
 replace github.com/hashicorp/consul/proto-public => github.com/hashicorp/consul/proto-public v0.1.2-0.20231128204611-fd1d97c3349d
 
+replace github.com/hashicorp/consul/sdk => github.com/hashicorp/consul/sdk v0.4.1-0.20231130173447-f44a9e1729df
```
t-eckert, line 7: This points to my branch where I extended the Retry struct. When [this PR](https://github.com/hashicorp/consul/pull/19813) gets merged.

acceptance/go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/hashicorp/consul-k8s/acceptance
 
 go 1.20
 
+replace github.com/hashicorp/consul/sdk => github.com/hashicorp/consul/sdk v0.4.1-0.20231130173447-f44a9e1729df
```
t-eckert, line 5: This points to my branch where I extended the Retry struct. When https://github.com/hashicorp/consul/pull/19813 gets merged.

acceptance/tests/config-entries/config_entries_test.go
```diff
@@ -85,12 +85,12 @@ func TestController(t *testing.T) {
...
 					// endpoint fails initially.
-					out, err := k8s.RunKubectlAndGetOutputE(t, ctx.KubectlOptions(t), "apply", "-k", "../fixtures/bases/crds-oss")
+					out, err := k8s.RunKubectlAndGetOutputE(r, ctx.KubectlOptions(r), "apply", "-k", "../fixtures/bases/crds-oss")
 					require.NoError(r, err, out)
-					helpers.Cleanup(t, cfg.NoCleanupOnFailure, cfg.NoCleanup, func() {
+					helpers.CleanupWithOnFailure(r, cfg.NoCleanupOnFailure, cfg.NoCleanup, func() {
 						// Ignore errors here because if the test ran as expected
 						// the custom resources will have been deleted.
-						k8s.RunKubectlAndGetOutputE(t, ctx.KubectlOptions(t), "delete", "-k", "../fixtures/bases/crds-oss")
+						k8s.RunKubectlAndGetOutputE(r, ctx.KubectlOptions(r), "delete", "-k", "../fixtures/bases/crds-oss")
```
mkeeler, line 93: This probably wont execute in the manner you expect. The `CleanupWithOnFailure` method will include the `r` within a closure. As creation succeeded this retry block will not fail its final attempt. When the cleanup method is invoked it will call `r.Failed()` which in this case will always return false and so this will never be run.

I _think_ to correct the behavior we should move the `helpers.CleanupWithOnFailure` call outside of the `retry.Run`

acceptance/framework/helpers/helpers.go
```diff
@@ -84,11 +86,10 @@ func SetupInterruptHandler(cleanup func()) {
...
 }
 
-// Cleanup will both register a cleanup function with t
-// and SetupInterruptHandler to make sure resources get cleaned up
-// if an interrupt signal is caught.
-func Cleanup(t *testing.T, noCleanupOnFailure bool, noCleanup bool, cleanup func()) {
-	t.Helper()
+// CleanupWithOnFailure will both register a cleanup function with t and SetupInterruptHandler to make sure resources
+// get cleaned up if an interrupt signal is caught.
+func CleanupWithOnFailure(t terratesting.TestingT, noCleanupOnFailure bool, noCleanup bool, cleanup func()) {
```
mkeeler, line 91: Note that if `t` is a `retry.R` then `Failed` will only return true if the last retry attempt failed. This means that you probably shouldn't add a cleanup func like this in a retry block **unless you use the `retry.WithImmediateCleanup` method to run cleanup funcs between each retry attempt.

PR #3297, opened 2023-12-01
NET-6758 Use kubebuilder for v2 GatewayClassConfig CRD

nathancoleman:
We've run into a wall that prevents us from parsing the `GatewayClassConfig` from yaml when the `GatewayClassConfig` was generated from Consul proto. @jm96441n can hopefully comment with more details since he's led the investigation.

Since Consul doesn't actually need any information from the `GatewayClassConfig`, we'd like to move forward using Go structs w/ kubebuilder annotations to unblock us for now. We can revisit this issue later if we decide it's worthwhile.

**How I've tested this PR:**
`helm install` with this branch and verify that you can `kubectl apply` a `GatewayClassConfig`.

<details>
<summary>Example GatewayClassConfig</summary>

```yaml
kind: GatewayClassConfig
metadata:
  name: my-gateway
spec:
  copyAnnotations:
    service:
    - copy-me 
  deployment:
    defaultInstances: 1
    maxInstances: 1
    minInstances: 1
    resources:
      limits:
        cpu: 100m
        memory: 100Mi
      requests:
        cpu: 100m
        memory: 100Mi
  serviceType: LoadBalancer
```
</details>

See above

**Checklist:**
- [ ] Tests added


missylbytes (approved):

control-plane/api/mesh/v2beta1/gateway_class_config_types.go
```diff
@@ -38,115 +25,79 @@ type GatewayClassConfig struct {
...
-			// Because we are explicitly defining NS/partition, this will not default and must be explicit.
-			// At a future point, this will move out of the Tenancy block.
-			PeerName: constants.DefaultConsulPeer,
-		},
-	}
-}
+	// +kubebuilder:validation:Enum=ClusterIP;NodePort;LoadBalancer
+	ServiceType *corev1.ServiceType `json:"serviceType,omitempty"`
 
-func (in *GatewayClassConfig) Resource(namespace, partition string) *pbresource.Resource {
```
jm96441n, line 69: do we still need all these functions like `ResourceID`, `Resource`, etc to maintain compatibility with other v2 resources?

nathancoleman: I intentionally removed these so that no kubebuilder-based `GatewayClassConfig` can ever accidentally be used as a Consul resource and will produce a compile-time error instead of a runtime error

jm96441n (approved):
LGTM

PR #3298, opened 2023-12-01
NET-6759 Restructure v2 GatewayClassConfig CRD

nathancoleman:
This PR reshapes the `GatewayClassConfig` to match where we're at today, having evolved significantly since the first version was introduced for API Gateway.

The general idea is that there is one config stanza for each Kubernetes resource that we create for a given `MeshGateway` (and `Gateway` and `TerminatingGateway` in the future). Any extensions to the config should match the hierarchy of the corresponding Kubernetes resources.

Notable features:
- A top-level setting for `annotations` and `labels` that will apply to all resources we create for the `MeshGateway` utilizing this `GatewayClass`, regardless of type
- The ability to inherit annotations and labels from the `MeshGateway` resource still exists
- Support for container-specific configuration -- `resources`, notably -- on the container and the init container

**How I've tested this PR:**
Create a `GatewayClassConfig` with this new structure

<details>
<summary>Example GatewayClassConfig</summary>

```yaml
kind: GatewayClassConfig
metadata:
  name: consul-mesh-gateway
spec:
  annotations:
    set:
      put-this-annotation-on-everything: value
    inheritFromGateway:
    - copy-this-annotation-from-the-gateway
  labels:
    set:
      put-this-label-on-everything: value
    inheritFromGateway:
    - copy-this-label-from-the-gateway
  deployment:
    annotations:
      set:
        put-this-annotation-on-the-deployment: value
    labels:
      set:
        put-this-label-on-the-deployment: value
    replicas:
      default: 3
      min: 2
      max: 8
    container:
      consul:
        logging:
          level: debug
      resources:
        limits:
          cpu: 500m
          memory: 128Mi
        requests:
          cpu: 250m
          memory: 64Mi
    initContainer:
      consul:
        logging:
          level: warn
      resources: {}
    securityContext: {}
  role:
    annotations:
      set:
        put-this-annotation-on-the-role: value
  roleBinding: {}
  service:
    annotations:
      set:
        service.beta.kubernetes.io/aws-load-balancer-type: external
      inheritFromGateway:
      - service.beta.kubernetes.io/aws-load-balancer-name
    type: LoadBalancer
  serviceAccount: {}
```
</details>

**Checklist:**
- [ ] Tests added


control-plane/api/mesh/v2beta1/gateway_class_config_types.go
```diff
@@ -31,66 +31,98 @@ type GatewayClassConfig struct {
...
+}
 
-	// Tolerations allow the scheduler to schedule nodes with matching taints.
-	// More Info: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
-	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+type GatewayClassContainerConfig struct {
+	// Consul specifies configuration for the consul-dataplane container
+	Consul GatewayClassConsulConfig `json:"consul,omitempty"`
+	// Resources specifies the resource requirements for the created Deployment's container
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
missylbytes, line 78: Is this where things like meshGateway.replicas would get set?

control-plane/api/mesh/v2beta1/gateway_class_config_types.go
```diff
@@ -31,66 +31,98 @@ type GatewayClassConfig struct {
...
+	SecurityContext *corev1.PodSecurityContext `json:"securityContext,omitempty"`
+	// Tolerations specifies the tolerations to use on the created Deployment
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+}
 
-	// NodeSelector is a selector which must be true for the pod to fit on a node.
-	// Selector which must match a node's labels for the pod to be scheduled on that node.
-	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
-	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+type GatewayClassInitContainerConfig struct {
```
missylbytes, line 67: Nit/question, some of these fields are just `GatewayClass...` instead of `GatewayClassConfig...`. Is that for brevity since their shouldn't be these same fields on the actual `GatewayClass` so there won't be confusion?

control-plane/api/mesh/v2beta1/gateway_class_config_types.go
nathancoleman: This draft doesn't have all fields set yet. I'd expect it to go where you linked to 💯

nathancoleman: Add replicas configuration that looks like
```yaml
deployment:
  replicas:
    default: 3
    min: 2
    max: 8
```

missylbytes (approved):

PR #3306, opened 2023-12-05
[NET-6404] Create Mesh Gateway Resource from Gateway Resources Job

jm96441n:
Changes proposed in this PR:
- create mesh gateway resource from gateway resources job
-

How I've tested this PR:
- :eye:  
- ran in kind cluster

:eye: 
run in kind cluster

Checklist:
- [X] Tests added

### DIFF
--- a/charts/consul/templates/gateway-resources-clusterrole.yaml
+++ b/charts/consul/templates/gateway-resources-clusterrole.yaml
@@ -14,6 +14,15 @@ rules:
     - consul.hashicorp.com
     - mesh.consul.hashicorp.com
     resources:
+    - meshgateways
+    verbs:
+    - get
+    - update
+    - create
+  - apiGroups:
+    - consul.hashicorp.com
+    - mesh.consul.hashicorp.com
+    resources:
     - gatewayclassconfigs
     verbs:
     - get
--- a/charts/consul/templates/gateway-resources-clusterrole.yaml
+++ b/charts/consul/templates/gateway-resources-clusterrole.yaml
@@ -11,7 +11,6 @@ metadata:
     component: gateway-resources
 rules:
   - apiGroups:
-    - consul.hashicorp.com
     - mesh.consul.hashicorp.com
     resources:
     - meshgateways
--- a/charts/consul/templates/gateway-resources-configmap.yaml
+++ b/charts/consul/templates/gateway-resources-configmap.yaml
@@ -77,7 +77,9 @@ data:
               set: {{ toJson .Values.meshGateway.serviceAccount.annotations }}
           {{- end }}
     meshGateways:
-      - metadata:
+      - apiVersion: mesh.consul.hashicorp.com/v2beta1
+        kind: meshGateway
+        metadata:
           name: mesh-gateway
           namespace: {{ .Release.Namespace }}
         spec:
--- a/charts/consul/templates/gateway-resources-configmap.yaml
+++ b/charts/consul/templates/gateway-resources-configmap.yaml
@@ -78,7 +78,7 @@ data:
           {{- end }}
     meshGateways:
       - apiVersion: mesh.consul.hashicorp.com/v2beta1
-        kind: meshGateway
+        kind: MeshGateway
         metadata:
           name: mesh-gateway
           namespace: {{ .Release.Namespace }}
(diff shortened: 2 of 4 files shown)


control-plane/subcommand/gateway-resources/command.go
```diff
@@ -288,7 +290,15 @@ func (c *Command) Run(args []string) int {
...
 	if len(c.gatewayConfig.GatewayClassConfigs) > 0 {
-		err = c.createV2GatewayClassAndClassConfigs(context.Background(), "consul-mesh-gateway", "consul-mesh-gateway-controller")
+		err = c.createV2GatewayClassAndClassConfigs(context.Background(), meshGatewayComponent, "consul-mesh-gateway-controller")
+		if err != nil {
+			c.UI.Error(err.Error())
+			return 1
+		}
+	}
+
+	if len(c.gatewayConfig.MeshGateways) > 0 {
```
jm96441n, line 300: I thought about a check here that an entry in the gatewayClassConfig list must exist for each mesh gateway and panic if not (as this would be a programmer error on our end), but left it out for now

charts/consul/templates/gateway-resources-clusterrole.yaml
```diff
@@ -10,6 +10,15 @@ metadata:
     release: {{ .Release.Name }}
     component: gateway-resources
 rules:
+  - apiGroups:
+    - consul.hashicorp.com
```
nathancoleman, line 14: MeshGateway doesn't actually exist in this API group right?

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -30,8 +30,9 @@ data:
           nodeSelector: {{ .Values.meshGateway.nodeSelector }}
           serviceType: {{ .Values.meshGateway.service.type }}
     meshGateways:
-      - name: mesh-gateway
-        namespace: {{ .Release.Namespace }}
+      - metadata:
+          name: mesh-gateway
+          namespace: {{ .Release.Namespace }}
```
nathancoleman, line 35: Big fan of this change. Thoughts on just making it the whole resource like we did with `GatewayClassConfig`? I know we can derive the `apiVersion` and `kind` from the fact that it's in this list, but the organization part of my brain prefers we be consistent anyways. It also makes slurping up the config a lil more straightforward

nathancoleman (commented):
Looks good! A couple of questions

control-plane/subcommand/gateway-resources/command.go
missylbytes: Can you clarify what you mean? Are you saying you thought about:
```
if len(c.gatewayConfig.MeshGateways) > 0 {
  if cantFindMatch(c.gatewayConfig.MeshGateways, c.gatewayConfig.GatewayClassConfigs){
     error("cannot find matching gatewayClassConfig for `my-mesh-gw`, `my-mesh-gw-cc` does not exist")
    }
....
```
I don't know how we would want a user to try and fix that, maybe we need to catch it before it gets here? idk like bats tests or something. I feel like if this is failing then someone messed with the gatewayResources configMap and shouldn't have.
I don't have any suggestions here, just talking it out.

control-plane/subcommand/gateway-resources/command.go
```diff
@@ -442,8 +453,28 @@ func (c *Command) createV2GatewayClassAndClassConfigs(ctx context.Context, compo
 		if err != nil {
 			return err
 		}
+
```
missylbytes: nit: spacing

charts/consul/templates/gateway-resources-clusterrole.yaml
jm96441n: ah no it does not, that's a bad copy pasta

charts/consul/templates/gateway-resources-configmap.yaml
jm96441n: can do!

nathancoleman (approved):

control-plane/subcommand/gateway-resources/command.go
missylbytes: Yeah definitely something for us to keep in mind.

PR #3307, opened 2023-12-05
[NET-6588] Remove abandoned virtual nodes from Consul Catalog

curtbushko:
Changes proposed in this PR:
- When deregistering a service in Consul we would leave the virtual node sitting around
- This PR checks if a node has an empty service list and if so deletes 

How I've tested this PR:

- manually testing in kind to make sure nodes are cleaned up
- unit tests. The enpoints controller tests have a lot of coverage for service interactions.

👀 

Checklist:
- [ ] Tests added

### DIFF
--- a/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
@@ -292,7 +292,7 @@ func (r *Controller) registerServicesAndHealthCheck(apiClient *api.Client, pod c
 
 		// Register the service instance with Consul.
 		r.Log.Info("registering service with Consul", "name", serviceRegistration.Service.Service,
-			"id", serviceRegistration.ID)
+			"id", serviceRegistration.Service.ID)
 		_, err = apiClient.Catalog().Register(serviceRegistration, nil)
 		if err != nil {
 			r.Log.Error(err, "failed to register service", "name", serviceRegistration.Service.Service)
@@ -308,7 +308,7 @@ func (r *Controller) registerServicesAndHealthCheck(apiClient *api.Client, pod c
 		}
 
 		// Register the proxy service instance with Consul.
-		r.Log.Info("registering proxy service with Consul", "name", proxyServiceRegistration.Service.Service)
+		r.Log.Info("registering proxy service with Consul", "name", proxyServiceRegistration.Service.Service, "id", proxyServiceRegistration.Service.ID)
 		_, err = apiClient.Catalog().Register(proxyServiceRegistration, nil)
 		if err != nil {
 			r.Log.Error(err, "failed to register proxy service", "name", proxyServiceRegistration.Service.Service)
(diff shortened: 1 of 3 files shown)


control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -975,13 +975,41 @@ func (r *Controller) deregisterService(apiClient *api.Client, k8sSvcName, k8sSvc
...
 	return errs
 
 }
 
+// deregisterNode removes a node if it does not have any associated services attached to it.
+// When using Consul Enterprise, serviceInstancesForK8SServiceNameAndNamespace will search across all namespaces
+// (wildcard) to determine if there any other services associated with the Node. We also only search for nodes that have
+// the correct kubernetes metadata (managed-by-endpoints-controller and synthetic-node).
+func (r *Controller) deregisterNode(apiClient *api.Client, k8sServiceName, k8sServiceNamespace, nodeName string) error {
+	nodeServices, err := r.serviceInstancesForK8SServiceNameAndNamespace(apiClient, k8sServiceName, k8sServiceNamespace, nodeName)
```
DanStough, line 998: I wonder if this is the wrong function because it filters on the `k8sServiceName`, but to clean the node we need to know if ANY service is associated with the node.

curtbushko: We are getting all the services associated with that node with:

If I am reading the filter correctly it is:

give me the node with:
 meta {
 "k8s-service" = "static-server"
 "k8s-namespace" = "default"
 "managed-by" = "endpoints-controller"
 }

`filter := fmt.Sprintf(`Meta[%q] == %q and Meta[%q] == %q and Meta[%q] == %q`,
		metaKeyKubeServiceName, k8sServiceName, constants.MetaKeyKubeNS, k8sServiceNamespace, metaKeyManagedBy, constants.ManagedByValue)`

`serviceList, _, err = apiClient.Catalog().NodeServiceList(nodeName, &api.QueryOptions{Filter: filter})`

I think we do it with a filter like this instead of node because Consul would be slow to return without the filter.

komapa:
Let me know if you want me to run this in a real cluster which can reproduce this problem many times per day :)

curtbushko:
> Let me know if you want me to run this in a real cluster which can reproduce this problem many times per day :)

@komapa That would amazing!

I just built my testing image and you should be able to pull it from `curtbushko/consul-k8s-control-plane-dev:NET-6588-abandoned-nodes`

curtbushko:
> Let me know if you want me to run this in a real cluster which can reproduce this problem many times per day :)

@komapa Thank you!

You can grab the image at: `docker.io/curtbushko/consul-k8s-control-plane-dev:abandoned-nodes`

(I tagged you late on Friday but I found a bug that I had to fix. It should be better now)

thisisnotashwin (approved):
I really like this. I have minor concerns about the performance of the endpoints controller given we have steadily increased what happens during the reconciles, but i think that requires a different, more thorough introspection into what we do in the endpoints controller. This is great work right here though.

curtbushko:
> I really like this. I have minor concerns about the performance of the endpoints controller given we have steadily increased what happens during the reconciles, but i think that requires a different, more thorough introspection into what we do in the endpoints controller. This is great work right here though.

@thisisnotashwin Zalimeni has a PR in draft to help with the endpoints controller performance. I do think that my PR might help with performance as we would have fewer abandoned nodes to scan through on each reconcile.

thisisnotashwin:
> > I really like this. I have minor concerns about the performance of the endpoints controller given we have steadily increased what happens during the reconciles, but i think that requires a different, more thorough introspection into what we do in the endpoints controller. This is great work right here though.
> 
> @thisisnotashwin Zalimeni has a PR in draft to help with the endpoints controller performance. I do think that my PR might help with performance as we would have fewer abandoned nodes to scan through on each reconcile.

yayyy!! that definitely helps. i think this multipronged approach is the way forward!

PR #3315, opened 2023-12-06
Add validation that externalServers.hosts is not set to HCP-managed cluster's addresses when global.cloud.enabled

zalimeni:
This reintroduces #3218 with an additional small change to ensure Helm template validation is applied with necessary short-circuits. I've added the fix as a separate commit to simplify review based on the original.

Original PR description below:

---------------------------------

### Overview

If we set `externalServers.hosts` to the public or private address of an HCP-manged cluster, we run into the following because we're hardcoding `-tls-server-name` to `server.{{ .Values.global.datacenter}}.{{ .Values.global.domain}}`:

> 2023-11-10T01:32:31.386Z [ERROR] consul-server-connection-manager: connection error: error="rpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for 517429dd-fcab-4fe6-bc4a-57750f69df7e.aws.hashicorp.cloud, consul-cluster.consul.e4065767-bd61-4f00-8b2b-81f631f7d4d1.aws.hashicorp.cloud, consul-cluster.private.consul.e4065767-bd61-4f00-8b2b-81f631f7d4d1.aws.hashicorp.cloud, not server.consul-cluster.consul\”"

We [hard-code `-tls-server-name` in many places](https://github.com/search?q=repo%3Ahashicorp%2Fconsul-k8s%20%22-tls-server-name%3Dserver.%7B%7B%20.Values.global.datacenter%7D%7D.%7B%7B%20.Values.global.domain%7D%7D%22&type=code) like:

https://github.com/hashicorp/consul-k8s/blob/4d7a1873b2c6049efc9769e3a03406db1a8f1380/charts/consul/templates/_helpers.tpl#L229-L231

- Add validation that users don't set `global.cloud.enabled` while also using `externalServers.hosts` that point at an HCP-managed cluster (any address that has the suffix ".hashicorp.cloud")
- Add note to `values.yaml` describing how `global.cloud` is really for _linking_ clusters, not for HCP-managed cloud clusters.

### How I've tested this PR:
- bats, new bats test, ran it locally

### Checklist:
- [x] Tests added

### DIFF
--- a/.changelog/3315.txt
+++ b/.changelog/3315.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+helm: add validation that global.cloud.enabled is not set with externalServers.hosts set to HCP-managed clusters
+```
--- a/charts/consul/templates/connect-inject-deployment.yaml
+++ b/charts/consul/templates/connect-inject-deployment.yaml
@@ -1,6 +1,7 @@
 {{- if and .Values.global.peering.enabled (not .Values.connectInject.enabled) }}{{ fail "setting global.peering.enabled to true requires connectInject.enabled to be true" }}{{ end }}
 {{- if and .Values.global.peering.enabled (not .Values.global.tls.enabled) }}{{ fail "setting global.peering.enabled to true requires global.tls.enabled to be true" }}{{ end }}
 {{- if and .Values.global.peering.enabled (not .Values.meshGateway.enabled) }}{{ fail "setting global.peering.enabled to true requires meshGateway.enabled to be true" }}{{ end }}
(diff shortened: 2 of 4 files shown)


jjti (approved):
Thank you @zalimeni for this fix! I'm checking since I didn't last time: are these acceptance test failures an issue or possibly broken because of this change? https://github.com/hashicorp/consul-k8s-workflows/actions/runs/7119514288

zalimeni:
@jjti sure thing! Re: the acceptance test failure, I think you are ok. Given the vanilla acceptance suite passed and this change isn't specific to CNI, I expect you just hit a flake, which unfortunately isn't uncommon r/n w/ the acceptance tests (particularly CNI).

Consider me "approved" even though I opened the PR, and feel free to merge + backport when it makes sense.

PR #3316, opened 2023-12-06
Specify required RoleRef on RoleBinding

nathancoleman:
The `RoleRef` field must be set on the `RoleBinding`. Since we don't currently set this, K8s returns an error when the reconciler attempts to create the `RoleBinding`.

```log
Reconciler error    {"controller": "meshgateway", "controllerGroup": "mesh.consul.hashicorp.com", "controllerKind": "MeshGateway", "MeshGateway": {"name":"mesh-gateway","namespace":"consul"}, "na │
│ mespace": "consul", "name": "mesh-gateway", "reconcileID": "80c9985f-28ed-4872-b330-7fcf87dffcde", "error": "unable to create role binding: unsupported role reference kind: \"\""}
```

**How I've tested this PR:**
`helm upgrade --install` from this branch, then verify that the expected `RoleBinding` was created:
```shell
$ kubectl get rolebinding -A
```

See above

**Checklist:**
- [ ] Tests added

### DIFF
--- a/control-plane/gateways/role.go
+++ b/control-plane/gateways/role.go
@@ -25,10 +25,16 @@ func (b *meshGatewayBuilder) RoleBinding() *rbacv1.RoleBinding {
 		},
 		Subjects: []rbacv1.Subject{
 			{
+				APIGroup:  "",
 				Kind:      rbacv1.ServiceAccountKind,
 				Name:      b.gateway.Name,
 				Namespace: b.gateway.Namespace,
 			},
+		},
+		RoleRef: rbacv1.RoleRef{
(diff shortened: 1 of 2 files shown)


control-plane/gateways/role.go
```diff
@@ -25,10 +25,16 @@ func (b *meshGatewayBuilder) RoleBinding() *rbacv1.RoleBinding {
 		},
 		Subjects: []rbacv1.Subject{
 			{
+				APIGroup:  "",
```
nathancoleman, line 28: Just setting this to be explicit

jm96441n (approved):
LGTM

PR #3318, opened 2023-12-06
[NET-6769] Update GatewayClassConfig for PortModifier

jm96441n:
- Updates gateway class config to be able to read in the port modifier value

- ran tests

- run tests
- :eye: 

### Checklist ###
- [X] Tests added

### DIFF
--- a/charts/consul/templates/crd-gatewayclassconfigs.yaml
+++ b/charts/consul/templates/crd-gatewayclassconfigs.yaml
@@ -107,6 +107,10 @@ spec:
                                 type: string
                             type: object
                         type: object
+                      portModifier:
+                        description: PortModifier specifies the value to be added
+                          to every port value for listeners on this gateway
+                        type: integer
                       resources:
                         description: Resources specifies the resource requirements
                           for the created Deployment's container
--- a/charts/consul/templates/crd-gatewayclassconfigs.yaml
+++ b/charts/consul/templates/crd-gatewayclassconfigs.yaml
@@ -109,7 +109,9 @@ spec:
                         type: object
                       portModifier:
                         description: PortModifier specifies the value to be added
-                          to every port value for listeners on this gateway
+                          to every port value for listeners on this gateway. This
+                          is generally used to avoid binding to privileged ports in
+                          the container.
                         type: integer
                       resources:
                         description: Resources specifies the resource requirements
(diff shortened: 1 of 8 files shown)


control-plane/gateways/role.go
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) HashiCorp, Inc.
```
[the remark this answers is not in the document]
nathancoleman: yep yep

control-plane/gateways/deployment_dataplane_container.go
```diff
@@ -107,6 +114,13 @@ func consulDataplaneContainer(config GatewayConfig, resources *corev1.ResourceRe
 		ContainerPort: int32(constants.DefaultWANPort),
 	})
 
+	if containerConfig != nil {
```
nathancoleman, line 117: I think this should be in an `if-else` with the port above.

if containerConfig, define the port named `wan` using the `PortModifier`
if nil containerConfig, define the port named `wan` using the `DefaultWANPort`

charts/consul/templates/crd-gatewayclassconfigs.yaml
```diff
@@ -107,6 +107,10 @@ spec:
                                 type: string
                             type: object
                         type: object
+                      portModifier:
+                        description: PortModifier specifies the value to be added
+                          to every port value for listeners on this gateway
```
nathancoleman, line 112: I think this requires a regen to include the added bit about avoiding privileged ports

control-plane/api/mesh/v2beta1/gateway_class_config_types.go
```diff
@@ -89,6 +89,9 @@ type GatewayClassContainerConfig struct {
 	Consul GatewayClassConsulConfig `json:"consul,omitempty"`
 	// Resources specifies the resource requirements for the created Deployment's container
 	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+	// PortModifier specifies the value to be added to every port value for listeners on this gateway
+	// this is generally used to avoid binding to privileged ports in the container
+	PortModifier int `json:"portModifier,omitempty"`
```
nathancoleman, line 94:
```suggestion
	// PortModifier specifies the value to be added to every port value for listeners on this gateway.
	// This is generally used to avoid binding to privileged ports in the container.
	PortModifier int `json:"portModifier,omitempty"`
```

nathancoleman (approved):

PR #3328, opened 2023-12-08
NET-6393: Create/update/delete Service on MeshGateway reconcile

NiniOak:
When reconciling a MeshGateway resource in the v2 API, create/update/delete a corresponding Service.

- Unit Test in code for review

**Manual Test Steps Below**
- Checkout and build consul-k8s using this branch;
- Run `make control-plane-dev-docker` to ensure you're using a `local` control-plane-dev image with changes
- [If required] Update consul values.yaml with ` imageK8S: consul-k8s-control-plane-dev:local`
- Then do a local kind helm install/uninstall of Consul
- Add `v2beta1` gateway resource

**RESULT**
```
$ k get gatewayclass.mesh -A
gateway-class2                               32m
```
Then get the gatewayclass.mesh resource yaml output
```
kind: GatewayClass
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"mesh.consul.hashicorp.com/v2beta1","kind":"GatewayClass","metadata":{"annotations":{},"name":"gateway-class2"},"spec":{"parametersRef":{"group":"mesh.consul.hashicorp.com","kind":"GatewayClassConfig","name":"gateway-class-config2"}}}
  creationTimestamp: "2023-12-13T02:21:02Z"
  generation: 1
  name: gateway-class2
  resourceVersion: "1658"
  uid: e17be12f-2afa-4e57-bc65-86733ef075ae
spec:
  parametersRef:
    group: mesh.consul.hashicorp.com
    kind: GatewayClassConfig
    name: gateway-class-config2
```
Next, validate service was deployed
```
$ k get service -n consul
mesh-gateway              LoadBalancer   10.96.67.159   <pending>     443:32130/TCP  
```
Then get the yaml output
```
$ k get service mesh-gateway -o yaml -n consul
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2023-12-11T20:39:52Z"
  labels:
    mesh.consul.hashicorp.com/managed-by: consul-k8s
  name: mesh-gateway
  namespace: consul
  ownerReferences:
  - apiVersion: mesh.consul.hashicorp.com/v2beta1
    blockOwnerDeletion: true
    controller: true
    kind: MeshGateway
    name: mesh-gateway
    uid: 369c3fe7-4db7-4b20-ba4d-82da2f1b9b96
  resourceVersion: "1086"
  uid: 35706f9f-ba96-440f-8746-4984f711f483
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 10.96.9.244
  clusterIPs:
  - 10.96.9.244
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: wan
    nodePort: 30271
    port: 443
    protocol: TCP
    targetPort: 8443
  selector:
    mesh.consul.hashicorp.com/managed-by: consul-k8s
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer: {}
```
Follow testing steps above

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/config-entries/controllersv2/mesh_gateway_controller.go
+++ b/control-plane/config-entries/controllersv2/mesh_gateway_controller.go
@@ -289,7 +289,7 @@ func (r *MeshGatewayController) getGatewayClassForGateway(ctx context.Context, g
 }
 
 func areServicesEqual(a, b *corev1.Service) bool {
-
+	// If either service "a" or "b" is nil, don't want to try and merge the nil service
 	if a == nil || b == nil {
 		return true
 	}
--- a/control-plane/config-entries/controllersv2/mesh_gateway_controller.go
+++ b/control-plane/config-entries/controllersv2/mesh_gateway_controller.go
@@ -12,6 +12,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -128,7 +129,28 @@ func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Req
 		return fmt.Errorf("unable to create role binding: %w", err)
 	}
 
-	// TODO NET-6393
+	//Create Service
+
+	mergeServiceOp := func(ctx context.Context, existingObject, object client.Object) error {
+		existingService, ok := existingObject.(*corev1.Service)
+		if !ok && existingService != nil {
+			return fmt.Errorf("unable to infer existing service type")
+		}
+		builtService, ok := object.(*corev1.Service)
+		if !ok {
+			return fmt.Errorf("unable to infer built service type")
+		}
+
+		mergedService := mergeService(existingService, builtService)
+
+		_, err := controllerutil.CreateOrUpdate(ctx, r.Client, mergedService, func() error { return nil })
+		return err
+	}
+
+	err = r.opIfNewOrOwned(ctx, resource, &corev1.Service{}, builder.Service(), mergeServiceOp)
+	if err != nil {
+		return fmt.Errorf("unable to create service: %w", err)
+	}
 
 	// Create deployment
 
@@ -264,4 +286,41 @@ func (r *MeshGatewayController) getGatewayClassForGateway(ctx context.Context, g
 		return nil, client.IgnoreNotFound(err)
 	}
 	return &gatewayClass, nil
+}
+
+func areServicesEqual(a, b *corev1.Service) bool {
+
+	if a == nil || b == nil {
+		return true
+	}
+
+	if !equality.Semantic.DeepEqual(a.Annotations, b.Annotations) {
+		return false
+	}
+
+	if len(b.Spec.Ports) != len(a.Spec.Ports) {
+		return false
+	}
+
+	for i, port := range a.Spec.Ports {
+		otherPort := b.Spec.Ports[i]
+		if port.Port != otherPort.Port || port.Protocol != otherPort.Protocol {
+			return false
+		}
+	}
+	return true
+}
+
+// mergeService is used to keep annotations and ports from the `from` Service
+// to the `to` service. This prevents an infinite reconciliation loop when
+// Kubernetes adds this configuration back in.
+func mergeService(from, to *corev1.Service) *corev1.Service {
+	if areServicesEqual(from, to) {
+		return to
+	}
+
+	to.Annotations = from.Annotations
+	to.Spec.Ports = from.Spec.Ports
+
+	return to
 }
--- a/control-plane/gateways/service.go
+++ b/control-plane/gateways/service.go
@@ -1,0 +1,50 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package gateways
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
+)
+
+const port = 443
+
+func (b *meshGatewayBuilder) Service() *corev1.Service {
+	var (
+		containerConfig *meshv2beta1.GatewayClassContainerConfig
+		portModifier    = 0
+		serviceType     = corev1.ServiceType("")
+	)
+
+	if b.gcc != nil {
+		containerConfig = b.gcc.Spec.Deployment.Container
+		portModifier = containerConfig.PortModifier
+		serviceType = *b.gcc.Spec.Service.Type
+	}
+
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        b.gateway.Name,
+			Namespace:   b.gateway.Namespace,
+			Labels:      b.Labels(),
+			Annotations: b.Annotations(),
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: b.Labels(),
+			Type:     serviceType,
+			Ports: []corev1.ServicePort{
+				{
+					Name: "wan",
+					Port: port,
+					TargetPort: intstr.IntOrString{
+						IntVal: int32(port + portModifier),
+					},
+				},
+			},
+		},
+	}
+}
--- a/control-plane/subcommand/gateway-resources/command.go
+++ b/control-plane/subcommand/gateway-resources/command.go
@@ -436,6 +436,7 @@ func (c *Command) createV2GatewayClassAndClassConfigs(ctx context.Context, compo
 			return err
 		}
 
+		// TODO: NET-6934 remove hardcoded values;
 		class := &v2beta1.GatewayClass{
 			ObjectMeta: metav1.ObjectMeta{Name: cfg.Name, Labels: labels},
 			TypeMeta:   metav1.TypeMeta{Kind: "GatewayClass"},
--- a/control-plane/subcommand/gateway-resources/command.go
+++ b/control-plane/subcommand/gateway-resources/command.go
@@ -436,7 +436,7 @@ func (c *Command) createV2GatewayClassAndClassConfigs(ctx context.Context, compo
 			return err
 		}
 
-		// TODO: NET-6934 remove hardcoded values;
+		// TODO: NET-6934 remove hardcoded values
 		class := &v2beta1.GatewayClass{
 			ObjectMeta: metav1.ObjectMeta{Name: cfg.Name, Labels: labels},
 			TypeMeta:   metav1.TypeMeta{Kind: "GatewayClass"},
--- a/control-plane/api/mesh/v2beta1/gateway_class_config_types.go
+++ b/control-plane/api/mesh/v2beta1/gateway_class_config_types.go
@@ -113,6 +113,7 @@ type GatewayClassServiceConfig struct {
 	GatewayClassAnnotationsAndLabels `json:",inline"`
 
 	// Type specifies the type of Service to use (LoadBalancer, ClusterIP, etc.)
+	// +kubebuilder:validation:Enum=ClusterIP;NodePort;LoadBalancer
 	Type *corev1.ServiceType `json:"type,omitempty"`
 }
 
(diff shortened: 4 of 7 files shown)


control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -126,7 +124,11 @@ func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Req
 		return fmt.Errorf("unable to create role binding: %w", err)
 	}
 
-	// TODO NET-6393
+	//Create Service
+	err = r.opIfNewOrOwned(ctx, resource, &corev1.Service{}, builder.Service(), upsertOp)
```
nathancoleman, line 128: `builderService()` isn't currently implemented. I'd expect this implementation to live in `gateways/service.go`, analogous to [gateways/serviceaccount.go](https://github.com/hashicorp/consul-k8s/blob/278509109e664014280bba062754c87c7d2e4f56/control-plane/gateways/serviceaccount.go)

nathancoleman (changes requested):

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
NiniOak: Whooops, it didn't 🤦‍♀️

nathancoleman, line 128: The `Service` will need its own merge op based on [the v1 implementation](https://github.com/hashicorp/consul-k8s/blob/3f5486177031e7a75e45885fad499195f22894ae/control-plane/api-gateway/gatekeeper/service.go#L118-L127) -- similar to `Deployment` below -- instead of using `upsertOp`.

control-plane/gateways/service.go
```diff
@@ -0,0 +1,50 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const port = 443
+
+func (b *meshGatewayBuilder) Service() *corev1.Service {
+	var (
+		containerConfig *meshv2beta1.GatewayClassContainerConfig
+		PortModifier    = 0
```
nathancoleman, line 18:
```suggestion
		portModifier    = 0
```

control-plane/gateways/service.go
```diff
@@ -0,0 +1,50 @@
...
+			Labels:    b.Labels(),
+			//TODO: Ask about annotations, shoule we set this?!
+			// Annotations: annotations,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: b.Labels(),
+			Type:     serviceType,
+			Ports: []corev1.ServicePort{
+				{
+					Name: b.gateway.Name,
```
nathancoleman, line 41: I think we should hardcode this to `wan` to match what we used for the container port that's targeted below

nathancoleman (changes requested):
Nice work! Would you mind adding some elementary resource ownership tests to the table-driven tests in `mesh_gateway_controller_test.go`?

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -126,7 +125,67 @@ func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Req
 		return fmt.Errorf("unable to create role binding: %w", err)
 	}
 
-	// TODO NET-6393
+	//Create Service
+
+	areServicesEqual := func(a, b *corev1.Service) bool {
+
+		if a == nil || b == nil {
+			return true
```
jm96441n, line 133: should this be false if either is nil? like if a exists and b is nil then the would not be equal?

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -126,7 +125,67 @@ func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Req
...
+		}
+
+		if !equality.Semantic.DeepEqual(a.Annotations, b.Annotations) {
+			return false
+		}
+		if len(b.Spec.Ports) != len(a.Spec.Ports) {
```
jm96441n, line 139: nit: newline above this for readability

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -126,7 +125,67 @@ func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Req
...
+		}
+
+		for i, port := range a.Spec.Ports {
+			otherPort := b.Spec.Ports[i]
+			if port.Port != otherPort.Port {
+				return false
+			}
+			if port.Protocol != otherPort.Protocol {
+				return false
+			}
```
jm96441n, line 150: nit:
```suggestion
			if port.Port != otherPort.Port || port.Protocol != otherPort.Protocol {
				return false
			}
```

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -126,7 +125,67 @@ func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Req
...
+		}
+		return true
+	}
+
+	// mergeService is used to keep annotations and ports from the `from` Service
+	// to the `to` service. This prevents an infinite reconciliation loop when
+	// Kubernetes adds this configuration back in.
+	mergeService := func(from, to *corev1.Service) *corev1.Service {
```
jm96441n, line 158: minor/style pref: could we define this and the `areServicesEqual` as separate named functions rather than inline/anonymous functions (anonymous might not be the right word here, but splitting them out into named functions would allow them to be more easily tested)

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
NiniOak: Yes, if "a" exists, and "b" is nil then service would not be equal. The idea here was if service "a"  or "b" is nil, then there really isn't a need to merge from nil to an existing service. But after re-reading the `mergeServiceOp` function, we do check that usecase and return an error. So I'll return false here. Good catch!

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -126,7 +125,67 @@ func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Req
...
-	// TODO NET-6393
+	//Create Service
+
+	areServicesEqual := func(a, b *corev1.Service) bool {
+
+		if a == nil || b == nil {
+			return true
+		}
+
+		if !equality.Semantic.DeepEqual(a.Annotations, b.Annotations) {
```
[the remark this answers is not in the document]
NiniOak: I had a conversation with @sarahalsmiller about this. From the unit tests, we know that sometimes the maps come unordered and will return false even if the maps have the same values. So using `DeepEqual` in this case prevent a false positive/negative comparison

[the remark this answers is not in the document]
jm96441n: > we know that sometimes the maps come unordered and will return false even if the maps have the same values

gotcha, so maps will always be unordered, we could check the lengths of the maps and then ensure that keys from both are in the other and the values for those keys are the same, so something like:
```
if len(a.Annotations) != len(b.Annotations) {
  return false
}
for k, v in a.Annotations {
  v2, ok := b.Annotations[k]
  if !ok || v != v2 {
     return false
   }
}
```
but I'm fine leaving this as is

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
jm96441n: do we care if both a and b are nil?

NiniOak: We do! Technically (I think), service "b" or builtService object should never be nil as that's the service we want to create/update.

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
NiniOak: done :)

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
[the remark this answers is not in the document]
NiniOak: I'll leave it as is ,and change if we ever run into issues. Thanks for the suggestion TIL!

control-plane/gateways/service.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: b.Labels(),
+			Type:     serviceType,
+			Ports: []corev1.ServicePort{
+				{
+					Name: "wan",
+					Port: port,
+					TargetPort: intstr.IntOrString{
+						IntVal: int32(port + portModifier),
```
nathancoleman, line 43: Perfect 🙌

jm96441n (approved):
LGTM! Great work!

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -124,7 +125,28 @@ func (r *MeshGatewayController) onCreateUpdate(ctx context.Context, req ctrl.Req
...
+			return fmt.Errorf("unable to infer existing service type")
+		}
+		builtService, ok := object.(*corev1.Service)
+		if !ok {
+			return fmt.Errorf("unable to infer built service type")
+		}
+
+		mergedService := mergeService(existingService, builtService)
+
+		_, err := controllerutil.CreateOrUpdate(ctx, r.Client, mergedService, func() error { return nil })
```
t-eckert, line 142: Out of curiosity, what is the role of this no-op function?

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -261,3 +283,40 @@ func (r *MeshGatewayController) getGatewayClassForGateway(ctx context.Context, g
 	}
 	return &gatewayClass, nil
 }
+
+func areServicesEqual(a, b *corev1.Service) bool {
+
+	if a == nil || b == nil {
```
t-eckert, line 289: Should this be `&&`? Or do you accept them as equal if either one is `nil`?

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
```diff
@@ -261,3 +283,40 @@ func (r *MeshGatewayController) getGatewayClassForGateway(ctx context.Context, g
...
+		if port.Port != otherPort.Port || port.Protocol != otherPort.Protocol {
+			return false
+		}
+	}
+	return true
+}
+
+// mergeService is used to keep annotations and ports from the `from` Service
+// to the `to` service. This prevents an infinite reconciliation loop when
+// Kubernetes adds this configuration back in.
```
t-eckert, line 312: Love this comment!

t-eckert (approved):
Excellent! I'm really digging the clarity of your code.

control-plane/subcommand/gateway-resources/command.go
```diff
@@ -430,6 +430,8 @@ func (c *Command) createV2GatewayClassAndClassConfigs(ctx context.Context, compo
 			return err
 		}
 
+		//TODO: remove hardcoded values;
+		//add constants for GatewayClass and GatewayClassConfig Kinds
```
missylbytes, line 434: Can you add a JIRA story and link it in this TODO please?

nathancoleman (approved):

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
sarahalsmiller: If either one of them is nil, you don't want to try and merge the nil service, so we consider them to be equal.

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
NiniOak: I guess it does carry out an action `CreateOrUpdate creates or updates the given object in the Kubernetes
cluster`, here I just don't log the result of the operation.

t-eckert: Ah! I should have been clearer. The `func() error { return nil }` no-op function.

control-plane/config-entries/controllersv2/mesh_gateway_controller.go
t-eckert: Gotcha! Maybe there could be a comment above the line to clarify.

control-plane/subcommand/gateway-resources/command.go
```diff
@@ -436,6 +436,7 @@ func (c *Command) createV2GatewayClassAndClassConfigs(ctx context.Context, compo
 			return err
 		}
 
+		// TODO: NET-6934 remove hardcoded values
```
missylbytes, line 439: 🙌

missylbytes (approved):

PR #3330, opened 2023-12-08
fixed incorrect test validation with peering test

wilkermichael:
- test was performing the wrong connection check, since http should be checking for a '403'
- added cluster creation in parallel to speed up peering tests

-
-

### Checklist ###
- [ ] Tests added

### DIFF
--- a/acceptance/tests/peering/peering_connect_test.go
+++ b/acceptance/tests/peering/peering_connect_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


acceptance/tests/peering/peering_connect_test.go
```diff
@@ -350,7 +368,7 @@ func TestPeering_Connect(t *testing.T) {
 				// If ACLs are enabled, test that deny intentions prevent connections.
 				if c.ACLsEnabled {
 					logger.Log(t, "testing intentions prevent connections through the terminating gateway")
-					k8s.CheckStaticServerConnectionFailing(t, staticClientOpts, staticClientName, externalServerHostnameURL)
```
wilkermichael, line 353: @t-eckert this was the line causing the issue with the test. It should be checking for 403

t-eckert (approved):
This is great! It'll make the tests run faster too!

PR #3347, opened 2023-12-11
Add `consul.hashicorp.com/proxy-config-map` annotation.

hashi-derek:
This introduces an annotation `consul.hashicorp.com/proxy-config-map` that allows for generically setting data in the opaque config map for proxy service registrations. Use of this annotation is not encouraged in most situations, but it is necessary for certain circumstances.

Notably, this new annotation allows users to specify the `xds_fetch_timeout_ms` configuration during service registration, which is sometimes needed by proxies with a large number of upstreams.

### DIFF
--- a/.changelog/3347.txt
+++ b/.changelog/3347.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+control-plane: Add new `consul.hashicorp.com/proxy-config-map` annotation that allows for setting values in the opaque config map for proxy service registrations.
+```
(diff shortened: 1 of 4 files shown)


david-yu:
@hashi-derek Could you also add a PR for docs for this new annotation here: https://developer.hashicorp.com/consul/docs/k8s/annotations-and-labels#consul-service-mesh

thisisnotashwin (approved):
nicely done!! LGTM!!

erichaberkorn (approved):

PR #3362, opened 2023-12-12
[Net-5510][Net-5455]: CRD controller should only patch the finalizer in the metadata of a CRD, rather than the whole object

ndhanushkodi:
Issue: ArgoCD will flap when configured with replace mode, which causes zero values to get defaulted and the applied config to differ from the state in K8s.

This is how the ArgoCD flapping problem happens:
1. Argo triggers a replace
2. A replace operation will remove all finalizers, including the one that we add to all config entries in the config entry controller
3. This triggers our config entry controller to run a reconcile to add the finalizer
4. When our controller adds the finalizer, it updates the whole object, causing the zero values to get defaulted. (json omit empty does not kick in because the object is not being serialized/deserialized like it does when you first apply the config).

This PR: 
- Only patches finalizers in the metadata of the CRD, rather than updating the entire entry, which had caused 0 values to be defaulted when using kubectl with replace.

Added @hashi-derek as a coauthor for his work on a finalizer patcher POC that I used for this.

From this repo https://github.com/ndhanushkodi/argocd-example-apps/tree/nd/debug-consul-config/consul-counting

Lightweight test steps with kubectl replace:
`kubectl apply -f counting.yaml -f dashboard.yaml -f splitter.yaml`

Test steps with Argo CD:
1. Followed [these instructions](https://argo-cd.readthedocs.io/en/stable/getting_started/) to setup argoCD, pointing to the consul-counting directory, set up the sync in manual mode with the "replace" setting.
2. Hit the sync button in argoCD a few times.
3. Before this PR, there would always be a diff that showed up in argoCD when using the "replace" setting with the config, because of the omitted values getting a default 0 value set.
4. After this PR, there is no longer a diff in argoCD.

👀 

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/config-entries/controllers/configentry_controller.go
+++ b/control-plane/config-entries/controllers/configentry_controller.go
@@ -31,6 +31,7 @@ import (
 const (
 	FinalizerName                = "finalizers.consul.hashicorp.com"
 	ConsulAgentError             = "ConsulAgentError"
+	ConsulPatchError             = "ConsulPatchError"
 	ExternallyManagedConfigError = "ExternallyManagedConfigError"
 	MigrationFailedError         = "MigrationFailedError"
 )
@@ -38,8 +39,13 @@ const (
 // Controller is implemented by CRD-specific controllers. It is used by
 // ConfigEntryController to abstract CRD-specific controllers.
 type Controller interface {
-	// Update updates the state of the whole object.
-	Update(context.Context, client.Object, ...client.UpdateOption) error
+	// AddFinalizersPatch creates a patch with the original finalizers with new ones appended to the end.
+	AddFinalizersPatch(obj client.Object, finalizers ...string) *FinalizerPatch
+	// RemoveFinalizersPatch creates a patch to remove a set of finalizers, while preserving the order.
+	RemoveFinalizersPatch(obj client.Object, finalizers ...string) *FinalizerPatch
+	// Patch patches the object. This should only ever be used for updating the metadata of an object, and not object
+	// spec or status. Updating the spec could have unintended consequences such as defaulting zero values.
+	Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error
 	// UpdateStatus updates the state of just the object's status.
 	UpdateStatus(context.Context, client.Object, ...client.SubResourceUpdateOption) error
 	// Get retrieves an obj for the given object key from the Kubernetes Cluster.
@@ -125,7 +131,13 @@ func (r *ConfigEntryController) ReconcileEntry(ctx context.Context, crdCtrl Cont
 		// then let's add the finalizer and update the object. This is equivalent
 		// registering our finalizer.
 		if !containsString(configEntry.GetFinalizers(), FinalizerName) {
-			configEntry.AddFinalizer(FinalizerName)
+			addPatch := crdCtrl.AddFinalizersPatch(configEntry, FinalizerName)
+			err := crdCtrl.Patch(ctx, configEntry, addPatch)
+			if err != nil {
+				return r.syncFailed(ctx, logger, crdCtrl, configEntry, ConsulPatchError,
+					fmt.Errorf("adding finalizer: %w", err))
+			}
+
 			if err := r.syncUnknown(ctx, crdCtrl, configEntry); err != nil {
 				return ctrl.Result{}, err
 			}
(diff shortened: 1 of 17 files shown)


control-plane/config-entries/controllers/configentry_controller.go
```diff
@@ -38,8 +38,13 @@ const (
...
 type Controller interface {
-	// Update updates the state of the whole object.
-	Update(context.Context, client.Object, ...client.UpdateOption) error
+	// AddFinalizersPatch creates a patch with the original finalizers with new ones appended to the end.
+	AddFinalizersPatch(obj client.Object, finalizers ...string) *FinalizerPatch
+	// Remove
+	RemoveFinalizersPatch(obj client.Object, finalizers ...string) *FinalizerPatch
+	// Patch patches the object. This should only ever be used for updating the metadata of an object, and not object
+	// spec or status. Updating the spec could have unintended consequences like defaulting zero values.
+	Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error
```
ndhanushkodi, line 47: @hashi-derek does this interface look ok? I tried keeping it as AddFinalizers and RemoveFinalizers, implemented by the FinalizerPatcher in finalizer_patch.go but if I do the patching within those functions, I'd need to pass in the k8s client to to the FinalizerPatcher. Keeping Patch as a function on the controller allows the FinalizerPatcher to not need another k8s client that's already being embedded in the controller implementations.

control-plane/subcommand/inject-connect/v1controllers.go
```diff
@@ -161,6 +161,7 @@ func (c *Command) configureV1Controllers(ctx context.Context, mgr manager.Manage
 		CrossNSACLPolicy:           c.flagCrossNamespaceACLPolicy,
 	}
 	if err := (&controllers.ServiceDefaultsController{
+		FinalizerPatcher:      controllers.FinalizerPatcher{},
```
ndhanushkodi, line 164: These can be removed

control-plane/config-entries/controllers/configentry_controller.go
hashi-derek: Yeah, this is fine. It's simpler than having the client passed in on each function call.

control-plane/config-entries/controllers/configentry_controller.go
```diff
@@ -125,7 +130,9 @@ func (r *ConfigEntryController) ReconcileEntry(ctx context.Context, crdCtrl Cont
 		// then let's add the finalizer and update the object. This is equivalent
 		// registering our finalizer.
 		if !containsString(configEntry.GetFinalizers(), FinalizerName) {
-			configEntry.AddFinalizer(FinalizerName)
+			addPatch := crdCtrl.AddFinalizersPatch(configEntry, FinalizerName)
+			crdCtrl.Patch(ctx, configEntry, addPatch)
```
hashi-derek, line 134: Is there supposed to be error handling here? I'm on the fence about this, but do you think we should maybe do the `Patch()` call inside of `AddFinalizerPatch()` ?

control-plane/config-entries/controllers/finalizer_patch.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, err
+	}
+
+	p, err := jsonpatch.CreateMergePatch([]byte(`{}`), newData)
+	return p, err
+}
+
+// Type implements client.Patch.
+func (fp *FinalizerPatch) Type() types.PatchType {
+	return types.MergePatchType
```
hashi-derek, line 69: Perhaps put in a note above this or the `FinalizerPatch` struct that mentions the patch will replace the entire list, rather than appending / removing in the patch operation itself and that there is a very small race condition where we could overwrite recently added finalizers by doing this. Because these are custom CRDs, Kubernetes does not allow for more advanced patch (StrategicMergePatch) operations that could alleviate this race.

control-plane/config-entries/controllers/configentry_controller.go
ndhanushkodi: Yes, just added the error handling, thank you for catching!

Definitely correct me if I'm wrong, but I don't think there's a way to do that without having an implementation of AddFinalizerPatch() for each controller. Since AddFinalizerPatch is a function on the FinalizerPatcher which is embedded in each controller right now, it doesn't have access to the Patch() function on the client. 

If we choose to un-embed the FinalizerPatcher and implement the AddFinalizersPatch for each controller, we could do the patch inside of AddFinalizerPatch. Do you have a preference? I don't have a strong one, just chose this for less code.

hashi-derek: This seems fine. Changing it would shuffle complexity into other locations, so we can leave this as-is.

control-plane/config-entries/controllers/finalizer_patch.go
ndhanushkodi: Added!

hashi-derek (approved):

PR #3363, opened 2023-12-12
[NET-6756] Tolerations for MeshGW

jm96441n:
- Adds support for tolerations for mesh gw

- :eye: 
- tests
- ran local kind cluster with tolerations set and checked the deployment to ensure it got copied over

- :eye: 
- tests
- ran local kind cluster with tolerations set and checked the deployment to ensure it got copied over

### Checklist ###
- [X] Tests added

### DIFF
--- a/charts/consul/templates/gateway-resources-configmap.yaml
+++ b/charts/consul/templates/gateway-resources-configmap.yaml
@@ -66,7 +66,7 @@ data:
               min: {{ .Values.meshGateway.replicas }}
               max: {{ .Values.meshGateway.replicas }}
             {{- if .Values.meshGateway.tolerations }}
-            tolerations: {{ toJson .Values.meshGateway.tolerations }}
+            tolerations: {{ fromYamlArray .Values.meshGateway.tolerations | toJson }}
             {{- end }}
           service:
             {{- if .Values.meshGateway.service.annotations }}
--- a/control-plane/gateways/deployment.go
+++ b/control-plane/gateways/deployment.go
@@ -36,23 +36,26 @@ func (b *meshGatewayBuilder) deploymentSpec() (*appsv1.DeploymentSpec, error) {
 		return nil, err
 	}
 
-	var containerConfig *meshv2beta1.GatewayClassContainerConfig
-	var deploymentConfig meshv2beta1.GatewayClassDeploymentConfig
-
+	var (
+		containerConfig  *meshv2beta1.GatewayClassContainerConfig
+		nodeSelector     map[string]string
+		tolerations      []corev1.Toleration
+		deploymentConfig meshv2beta1.GatewayClassDeploymentConfig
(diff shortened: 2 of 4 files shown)


control-plane/gateways/deployment.go
```diff
@@ -36,22 +36,26 @@ func (b *meshGatewayBuilder) deploymentSpec() (*appsv1.DeploymentSpec, error) {
 		return nil, err
 	}
 
-	var containerConfig *meshv2beta1.GatewayClassContainerConfig
+	var (
+		containerConfig *meshv2beta1.GatewayClassContainerConfig
+		nodeSelector    map[string]string
```
NiniOak, line 41: I wonder if this should be perhaps be changed to `nodeSelector = map[string]string{}`, because if b.gcc is empty, I believe this will panic as opposed to setting the value as nil.

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -63,7 +63,7 @@ data:
               min: {{ .Values.meshGateway.replicas }}
               max: {{ .Values.meshGateway.replicas }}
             {{- if .Values.meshGateway.tolerations }}
-            tolerations: {{ toJson .Values.meshGateway.tolerations }}
+            tolerations: {{ fromYamlArray .Values.meshGateway.tolerations | toJson }}
```
NiniOak, line 66: 🙌 nice!

control-plane/gateways/deployment.go
jm96441n: I _think_ it should work because you can assign a nil value to the field in the deployment and it should "just work :tm:" but I'm going to add the test just  to be sure

control-plane/subcommand/gateway-resources/command_test.go
```diff
@@ -460,6 +465,14 @@ func TestRun_loadGatewayConfigs(t *testing.T) {
...
 					"beta.kubernetes.io/os":   "linux",
 				},
+				Tolerations: []corev1.Toleration{
+					{
+						Key:      "key1",
+						Operator: "Equal",
+						Value:    "value1",
+						Effect:   "NoSchedule",
+					},
+				},
```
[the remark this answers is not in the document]
jm96441n: good call on this test! it actually caught a bug where we weren't setting the default resource requirements

control-plane/gateways/deployment.go
jm96441n: so yeah, this would panic if we were doing anything more than assigning the variables to a field on the deployment struct, if we tried accessing anything in them is where we'd have that nil panic issue

NiniOak: Nice! Good catch

NiniOak (approved):
LGTM 🚀

PR #3365, opened 2023-12-12
NET-6663 Set gateway-kind in Workload metadata when it represents a xGateway Pod

nathancoleman:
By setting the gateway-kind annotation on the Pods for a MeshGateway, we indicate to the Pod controller in consul-k8s that the Pod represents a mesh gateway (or api/terminating in the future). The Pod controller then passes this along as metadata on the Workload that it creates in Consul.

The end result is that the sidecar and gateway proxy controllers can determine which Workloads they should generate ProxyStateTemplates for and which should be left to the other.

When combined with https://github.com/hashicorp/consul/pull/19902, the sidecar proxy controller should skip any Workloads created for a `MeshGateway` when `meshGateway.enabled=true` and `experiments: [resource-apis]` in values.yaml.

See above

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/connect-inject/controllers/pod/pod_controller.go
+++ b/control-plane/connect-inject/controllers/pod/pod_controller.go
@@ -664,10 +664,16 @@ func parseLocality(node corev1.Node) *pbcatalog.Locality {
 
 func metaFromPod(pod corev1.Pod) map[string]string {
 	// TODO: allow custom workload metadata
-	return map[string]string{
+	meta := map[string]string{
 		constants.MetaKeyKubeNS:    pod.GetNamespace(),
 		constants.MetaKeyManagedBy: constants.ManagedByPodValue,
 	}
+
+	if gatewayKind := pod.Annotations[constants.AnnotationGatewayKind]; gatewayKind != "" {
+		meta[constants.MetaGatewayKind] = gatewayKind
+	}
+
+	return meta
 }
 
 // getHealthStatusFromPod checks the Pod for a "Ready" condition that is true.
--- a/control-plane/gateways/deployment.go
+++ b/control-plane/gateways/deployment.go
@@ -16,7 +16,8 @@ import (
 )
 
 const (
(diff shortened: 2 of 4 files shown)


DanStough (approved):
LGTM

control-plane/gateways/deployment.go
```diff
@@ -69,6 +70,7 @@ func (b *meshGatewayBuilder) deploymentSpec() (*appsv1.DeploymentSpec, error) {
 				Labels: b.Labels(),
 				Annotations: map[string]string{
 					"consul.hashicorp.com/mesh-inject": "false",
+					constants.AnnotationGatewayKind:    "mesh",
```
sarahalsmiller, line 73: should "mesh" also be declared as a constant?

sarahalsmiller (approved):

control-plane/connect-inject/controllers/pod/pod_controller.go
```diff
@@ -664,10 +664,16 @@ func parseLocality(node corev1.Node) *pbcatalog.Locality {
 
 func metaFromPod(pod corev1.Pod) map[string]string {
 	// TODO: allow custom workload metadata
-	return map[string]string{
+	meta := map[string]string{
 		constants.MetaKeyKubeNS:    pod.GetNamespace(),
 		constants.MetaKeyManagedBy: constants.ManagedByPodValue,
 	}
+
+	if gatewayKind := pod.Annotations[constants.AnnotationGatewayKind]; gatewayKind != "" {
```
jm96441n, line 672: should this be using the comma ok idiom to guard against the key not existing in the map?

nathancoleman: not necessary since `gatewayKind` will be an empty string in that case as well

control-plane/gateways/deployment.go
nathancoleman: done!

PR #3371, opened 2023-12-13
[NET-6721] Replicas support for MeshGW v2 Deployments

jm96441n:
- Add support for replicas config for meshgw v2 deployments
- currently draft because base branch https://github.com/hashicorp/consul-k8s/pull/3363 has an open PR

- :eye: 
- ran tests
- deployed kind cluster with replicas, saw them show up on deployment

- :eye: 
- ran tests
- deployed kind cluster with replicas, saw them show up on deployment

### Checklist ###
- [X] Tests added

### DIFF
--- a/control-plane/gateways/deployment.go
+++ b/control-plane/gateways/deployment.go
@@ -155,18 +155,18 @@ func compareDeployments(a, b *appsv1.Deployment) bool {
 
 func deploymentReplicaCount(replicas *meshv2beta1.GatewayClassReplicasConfig, currentReplicas *int32) *int32 {
 	// if we have the replicas config, use it
-	if replicas != nil && currentReplicas == nil {
+	if replicas != nil && replicas.Default != nil && currentReplicas == nil {
 		return replicas.Default
 	}
 
 	// if we have the replicas config and the current replicas, use the min/max to ensure
 	// the current replicas are within the min/max range
 	if replicas != nil && currentReplicas != nil {
-		if *currentReplicas > *replicas.Max {
+		if replicas.Max != nil && *currentReplicas > *replicas.Max {
 			return replicas.Max
 		}
 
-		if *currentReplicas < *replicas.Min {
+		if replicas.Min != nil && *currentReplicas < *replicas.Min {
 			return replicas.Min
 		}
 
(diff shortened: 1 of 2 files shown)


NiniOak (commented):
General testing question, if one replica is deleted does the reconciliation process work correctly? That is, does it get things back to the desired state? (2/2)

jm96441n:
@NiniOak that should work as that's baked into k8s, the mesh gateways aren't currently getting into a running state (which is expected) so I can't test killing one and seeing it come back up just yet

control-plane/gateways/deployment.go
```diff
@@ -156,11 +153,31 @@ func compareDeployments(a, b *appsv1.Deployment) bool {
...
+func deploymentReplicaCount(replicas *meshv2beta1.GatewayClassReplicasConfig, currentReplicas *int32) *int32 {
+	// if we have the replicas config, use it
+	if replicas != nil && currentReplicas == nil {
+		return replicas.Default
+	}
+
+	// if we have the replicas config and the current replicas, use the min/max to ensure
+	// the current replicas are within the min/max range
+	if replicas != nil && currentReplicas != nil {
+		if *currentReplicas > *replicas.Max {
```
t-eckert, line 165: Even though we know `replicas` is not `nil`, do we still have a risk of `nil` pointer deref if `replicas.Max` is `nil`?

control-plane/gateways/deployment.go
```diff
@@ -156,11 +153,31 @@ func compareDeployments(a, b *appsv1.Deployment) bool {
...
+			return replicas.Max
+		}
+
+		if *currentReplicas < *replicas.Min {
```
t-eckert, line 169: Same thought as above here.

t-eckert (approved):
I left a note about a potential `nil` pointer deref, but otherwise this looks good.

control-plane/gateways/deployment.go
nathancoleman: yes, it's possible that a user wants to configure a min but not a max, for example, in which case `replicas.Max` would be nil

jm96441n: so we don't give that much control to the user, `replicas` is just a single number https://developer.hashicorp.com/consul/docs/k8s/helm#v-meshgateway-replicas

jm96441n: and to see how we handle it: https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/gateway-resources-configmap.yaml#L65, we could do the nil checks here to future proof it though

PR #3380, opened 2023-12-14
[NET-6789] DNSPolicy support for meshgw v2 deployments

jm96441n:
- Add support for DNSPolicy in mesh gw v2 deployments
- 

- ran kind cluster setting values for DNSPolicy/no value to ensure it would default to `ClusterFirst` as the k8s docs specify
- ran tests

- :eye: 
- tests

### Checklist ###
- [X] Tests added


charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -66,6 +66,9 @@ data:
             {{- with .Values.meshGateway.hostNetwork }}
             hostNetwork: {{ . }}
             {{- end }}
+            {{- with .Values.meshGateway.dnsPolicy }}
+            dnsPolicy: {{ . }}
+            {{- end }}
```
nathancoleman, line 71: What does the `with` here do vs. something like
```yaml
{{- if .Values.meshGateway.dnsPolicy }}
dnsPolicy: {{ .Values.meshGateway.dnsPolicy }}
{{- end }}
```

control-plane/api/mesh/v2beta1/gateway_class_config_types.go
```diff
@@ -59,19 +59,21 @@ type GatewayClassDeploymentConfig struct {
...
+	// Tolerations specifies the tolerations to use on the created Deployment.
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
-	// HostNetwork specifies whether the gateway pods should run on the host network
+	// HostNetwork specifies whether the gateway pods should run on the host network.
 	HostNetwork bool `json:"hostNetwork,omitempty"`
 	// TopologySpreadConstraints is a feature that controls how pods are spead across your topology.
 	// More info: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// DNSPolicy specifies the dns policy to use. These are set on a per pod basis.
+	DNSPolicy string `json:"dnsPolicy,omitempty"`
```
nathancoleman, line 76: Is there a reason to prefer `string` here over `corev1.DNSPolicy` with a `+kubebuilder:validation:Enum` like we do for `<GatewayClassServiceConfig>.Type`?

charts/consul/templates/gateway-resources-configmap.yaml
jm96441n: pretty much the same thing, it just allows you to reference the value with `.` rather than needing to reference the value again https://pkg.go.dev/text/template#hdr-Actions, quoted the relevant section here:
> {{with pipeline}} T1 {{end}}
	If the value of the pipeline is empty, no output is generated;
	otherwise, dot is set to the value of the pipeline and T1 is
	executed.

control-plane/api/mesh/v2beta1/gateway_class_config_types.go
jm96441n: none in particular, let me change this so we get that validation as well

charts/consul/templates/gateway-resources-configmap.yaml
nathancoleman: ohhhhh I like that a lot.

TIL. Thanks for sharing!

control-plane/api/mesh/v2beta1/gateway_class_config_types.go
```diff
@@ -73,7 +73,8 @@ type GatewayClassDeploymentConfig struct {
 	// More info: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
 	// DNSPolicy specifies the dns policy to use. These are set on a per pod basis.
-	DNSPolicy string `json:"dnsPolicy,omitempty"`
+	// +kubebuilder:validation:Enum=Default;ClusterFirst;ClusterFirstWithHostNet;None
+	DNSPolicy corev1.DNSPolicy `json:"dnsPolicy,omitempty"`
```
nathancoleman, line 77: Is it safe to assume that this validation only applies if the key is included in the yaml? So the fact that it would obviously default to `""` if not set would be fine?

control-plane/gateways/deployment.go
```diff
@@ -108,6 +108,7 @@ func (b *meshGatewayBuilder) deploymentSpec() (*appsv1.DeploymentSpec, error) {
 				HostNetwork:               deploymentConfig.HostNetwork,
 				Tolerations:               deploymentConfig.Tolerations,
 				ServiceAccountName:        b.serviceAccountName(),
+				DNSPolicy:                 corev1.DNSPolicy(deploymentConfig.DNSPolicy),
```
nathancoleman, line 111: Assuming this is unnecessary now

control-plane/api/mesh/v2beta1/gateway_class_config_types.go
jm96441n: yup just tested it, this only applies if the value is specified (and k8s defaults to using `ClusterFirst` here if the value is not set https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy)

nathancoleman (approved):

PR #3381, opened 2023-12-14
NET-6788 - Mesh Gateway Host Port Support

sarahalsmiller:
- Added hostPort support

- Ran local consul test

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/gateways/deployment_dataplane_container.go
+++ b/control-plane/gateways/deployment_dataplane_container.go
@@ -27,17 +27,14 @@ const (
 	volumeName                   = "consul-connect-inject-data"
 )
 
-func consulDataplaneContainer(config GatewayConfig, containerConfig *v2beta1.GatewayClassContainerConfig, name, namespace string) (corev1.Container, error) {
+func consulDataplaneContainer(config GatewayConfig, containerConfig v2beta1.GatewayClassContainerConfig, name, namespace string) (corev1.Container, error) {
 	// Extract the service account token's volume mount.
 	var (
 		err             error
 		bearerTokenFile string
 	)
 
-	var resources *corev1.ResourceRequirements
-	if containerConfig != nil {
-		resources = containerConfig.Resources
-	}
+	resources := containerConfig.Resources
 
 	if config.AuthMethod != "" {
 		bearerTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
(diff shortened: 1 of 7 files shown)


control-plane/gateways/deployment_dataplane_container.go
```diff
@@ -112,11 +109,10 @@ func consulDataplaneContainer(config GatewayConfig, containerConfig *v2beta1.Gat
 	wanPort := corev1.ContainerPort{
 		Name:          "wan",
 		ContainerPort: int32(constants.DefaultWANPort),
+		HostPort:      containerConfig.HostPort,
 	}
 
-	if containerConfig != nil {
-		wanPort.ContainerPort = int32(443 + containerConfig.PortModifier)
-	}
+	wanPort.ContainerPort = 443 + containerConfig.PortModifier
```
NiniOak, line 115: Just ranting, why is hostPort int32 and containerPort just int?

control-plane/gateways/deployment_dataplane_container.go
```diff
@@ -112,11 +109,10 @@ func consulDataplaneContainer(config GatewayConfig, containerConfig *v2beta1.Gat
 	wanPort := corev1.ContainerPort{
 		Name:          "wan",
 		ContainerPort: int32(constants.DefaultWANPort),
+		HostPort:      containerConfig.HostPort,
```
NiniOak, line 112: Thinking out loud, should we potentially set a default value for `hostPort` if we happen to run into an issue where a host port isn't exposed? Super hypothetical

control-plane/gateways/deployment_dataplane_container.go
sarahalsmiller: It is int32, I updated the value to match what the object is expecting. https://github.com/hashicorp/consul-k8s/pull/3381/files#diff-3a9113026018fc0f58d311cb39bf0de995570740f1ebde6fa3cb1dcbb4e2c64bR101

control-plane/gateways/deployment_dataplane_container.go
sarahalsmiller: I believe the defaultValue of host port as it exists today is 0, in which case the port won't be exposed. I could explicitly set that, but we get that behavior baked in when we initialize an empty struct.

control-plane/gateways/deployment_dataplane_container.go
NiniOak: nice!

NiniOak (approved):

PR #3397, opened 2023-12-18
[NET-6766] Implement computing of annotations + labels for all created resources

nathancoleman:
- All resources created for v2 mesh gateways have annotations and labels computed based on a hierarchical structure.
- This PR implements that computation and assigns the output to each of the resources as appropriate

Create a mesh gateway by `helm install` w/ `experiments = [resource-apis]` and `meshGateway.enabled`.
Inspect the `Deployment`, `Service`, etc. and verify that they have the expected labels and annotations based on the `GatewayClassConfig`.

👀 

### Checklist ###
- [ ] Tests added


control-plane/gateways/metadata.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+const labelManagedBy = "mesh.consul.hashicorp.com/managed-by"
+
+var defaultLabels = map[string]string{labelManagedBy: "consul-k8s"}
+
+// Annotations returns the computed set of annotations for a resource by inheriting
+// from the MeshGateway's annotations and adding specified key-values from the
+// GatewayClassConfig based on the Kubernetes object type.
+func (b *meshGatewayBuilder) Annotations(object client.Object) map[string]string {
```
jm96441n, line 20: what are your thoughts on having separate `AnnotationsForX` methods (and the others) rather than the switch to make calling a bit more intuitive? like it reads a little strange to me to pass in an empty object at the call site to figure out what labels/annotations are needed

nathancoleman: That sounds perfectly cool to me!

jm96441n (approved):
LGTM

PR #3420, opened 2023-12-21
Support WAN Address Annotations

t-eckert:

- Add wanAddress configuration to the configmap
- Set the annotations on the mesh gateway CRD
- Patch through the annotations from the Mesh Gateway

I added BATs and I ran a deployment with the following configuration:

```yaml
global:
  imageK8S: consul-k8s-control-plane:local
  experiments:
    - resource-apis

meshGateway:
  enabled: true

ui:
  enabled: false
```

Which produced a mesh gateway with the correct configuration.

Screenshot 2023-12-21 at 1 44 17 PM

### Checklist ###
- [x] Tests added

### DIFF
--- a/charts/consul/templates/gateway-resources-configmap.yaml
+++ b/charts/consul/templates/gateway-resources-configmap.yaml
@@ -1,4 +1,9 @@
 {{- if .Values.connectInject.enabled }}
+
+# Validation
+# For meshGateway.wanAddress, static must be set if source is "Static"
+{{if (and (eq .Values.meshGateway.wanAddress.source "Static") (eq .Values.meshGateway.wanAddress.static ""))}}{{fail ".meshGateway.wanAddress.static must be set to a value if .meshGateway.wanAddress.source is Static"}}{{ end }}
+
 # Configuration of Gateway Resources Job which creates managed Gateway configuration.
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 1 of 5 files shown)


t-eckert:
The original story for this wanted the values to be set on the GatewayClassConfig by extending its configuration. Talking it over with Nate and Melisa, we decided to add the annotations directly on the Mesh Gateway CRD and patch those through. That way, we don't get mesh gateway specific config on a "general gateway" configuration object. 

Also, we aren't sure if we will need these annotations in V2 and if we don't, this model is a bit easier to remove.

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -1,4 +1,9 @@
 {{- if .Values.connectInject.enabled }}
+
+# Validation
+# For meshGateway.wanAddress, static must be set if source is "Static"
+{{if (and (eq .Values.meshGateway.wanAddress.source "Static") (eq .Values.meshGateway.wanAddress.static ""))}}{{fail ".meshGateway.wanAddress.static must be set to a value if .meshGateway.wanAddress.source is Static"}}{{ end }}
```
t-eckert, line 5: I added validation here that will block the invalid config of a "Static" address with no address set.

jm96441n (approved):
LGTM! nice work!

missylbytes (approved):

PR #3430, opened 2024-01-03
Pass gateway init resources from config to the container

t-eckert:
- Make initConfig an associated function to the builder
- Pass in init container resources if they are set
- Use values from the builder now that initContainer is associated with it

Updated the happy path unit test to exercise the resource setting.

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/gateways/deployment_init_container.go
+++ b/control-plane/gateways/deployment_init_container.go
@@ -146,7 +146,7 @@ func (b *meshGatewayBuilder) initContainer() (corev1.Container, error) {
 			Value: consulNamespace,
 		})
 
-	if config.TLSEnabled {
+	if b.config.TLSEnabled {
 		container.Env = append(container.Env,
 			corev1.EnvVar{
 				Name:  constants.UseTLSEnvVar,
@@ -154,11 +154,11 @@ func (b *meshGatewayBuilder) initContainer() (corev1.Container, error) {
 			},
 			corev1.EnvVar{
 				Name:  constants.CACertPEMEnvVar,
-				Value: config.ConsulCACert,
+				Value: b.config.ConsulCACert,
 			},
 			corev1.EnvVar{
 				Name:  constants.TLSServerNameEnvVar,
-				Value: config.ConsulTLSServerName,
+				Value: b.config.ConsulTLSServerName,
 			})
 	}
 
--- a/control-plane/gateways/deployment_init_container.go
+++ b/control-plane/gateways/deployment_init_container.go
@@ -11,6 +11,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 
+	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
 	"github.com/hashicorp/consul-k8s/control-plane/namespaces"
 )
@@ -18,6 +19,7 @@ import (
 const (
 	injectInitContainerName      = "consul-mesh-init"
 	initContainersUserAndGroupID = 5996
+	defaultBearerTokenFile       = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 )
 
 type initContainerCommandData struct {
@@ -51,7 +53,7 @@ func (b *meshGatewayBuilder) initContainer(config GatewayConfig, name, namespace
 
 	var bearerTokenFile string
 	if config.AuthMethod != "" {
-		bearerTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+		bearerTokenFile = defaultBearerTokenFile
 	}
 
 	// Render the command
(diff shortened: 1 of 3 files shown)


control-plane/gateways/deployment_init_container.go
```diff
@@ -30,15 +32,15 @@ type initContainerCommandData struct {
 	LogJSON  bool
 }
 
-// containerInit returns the init container spec for connect-init that polls for the service and the connect proxy service to be registered
+// initContainer returns the init container spec for connect-init that polls for the service and the connect proxy service to be registered
 // so that it can save the proxy service id to the shared volume and boostrap Envoy with the proxy-id.
-func initContainer(config GatewayConfig, name, namespace string) (corev1.Container, error) {
+func (b *meshGatewayBuilder) initContainer() (corev1.Container, error) {
```
t-eckert, line 37: I either had to add a param to this function to pass in gateway class config or associate the function with the builder. I felt the association was cleaner because we are now using so much of the struct. 

It doesn't really need to be a pointer receiver. I think that might make it faster to not copy, but nothing is changed on the struct by the function.

jm96441n: I think having consistency with the rest of the method set (where it's all defined as a pointer receiver) is another big reason to keep this as a pointer receiver

jm96441n, line 37: also, what do you think about changing the name of this method to something like `buildInitContainer` or `newInitContainer`, something to specify that this is creating the init container config rather than initializing a container

control-plane/gateways/deployment_init_container.go
```diff
@@ -50,8 +52,8 @@ func initContainer(config GatewayConfig, name, namespace string) (corev1.Contain
 	}
 
 	var bearerTokenFile string
-	if config.AuthMethod != "" {
-		bearerTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	if b.config.AuthMethod != "" {
+		bearerTokenFile = defaultBearerTokenFile
```
jm96441n, line 56: :+1:

control-plane/gateways/deployment_init_container.go
sarahalsmiller: Yes, I like the pointer reviever/pulling the params from the builder to keep everything in line with the rest of the package. 🚀

control-plane/gateways/deployment_init_container.go
```diff
@@ -143,17 +146,25 @@ func initContainer(config GatewayConfig, name, namespace string) (corev1.Contain
...
 
 	return container, nil
 }
 
+func initContainerResourcesOrDefault(gcc *meshv2beta1.GatewayClassConfig) corev1.ResourceRequirements {
+	if gcc != nil && gcc.Spec.Deployment.InitContainer != nil && gcc.Spec.Deployment.InitContainer.Resources != nil {
+		return *gcc.Spec.Deployment.InitContainer.Resources
+	}
+
+	return corev1.ResourceRequirements{}
```
sarahalsmiller, line 165: I'm assuming if this is empty k8s just uses whatever the defaults are?

sarahalsmiller (approved):
Looks good to me, but consider John's suggestion about renaming the initcontainer function.

control-plane/gateways/deployment_init_container.go
t-eckert: Yeah, it just sets it to empty and there are no requests/limits. It's not recommended, but we can set real defaults later.

control-plane/gateways/deployment_init_container.go
t-eckert: Yes. I like it. I think I'll do it all at once in another PR though.

nathancoleman: the only downside is that this logic is then specifically associated with the `meshGatewayBuilder` when it will likely need to be consumed by `apiGatewayBuilder` and/or `terminatingGatewayBuilder` as well in the future. We can figure that out when the time comes though 👍

control-plane/gateways/deployment_init_container.go
```diff
@@ -30,15 +32,15 @@ type initContainerCommandData struct {
...
-		AuthMethod:         config.AuthMethod,
-		LogLevel:           config.LogLevel,
-		LogJSON:            config.LogJSON,
-		ServiceName:        name,
-		ServiceAccountName: name,
+		AuthMethod:         b.config.AuthMethod,
+		LogLevel:           b.config.LogLevel,
+		LogJSON:            b.config.LogJSON,
+		ServiceName:        b.gateway.Name,
+		ServiceAccountName: b.gateway.Name,
```
nathancoleman, line 43: Like to use this everywhere since we have it -- in theory, changing the naming logic would then only have to happen in one place.

```suggestion
		ServiceAccountName: b.serviceAccountName(),
```

nathancoleman (approved):

PR #3431, opened 2024-01-04
NET-6768: Set port on MeshGateway

NiniOak:
Add support for .Values.meshGateway.service.port in service

- Update `GatewayClassServiceConfig` to store the port value
- Update `service` in configmap.yaml

- unit test passes

MANUAL TEST FAILING SO FAR
1. Install consul with custom meshGateway service port in values.yaml
```
meshGateway:
  enabled: true
  service:
    port: 8182
```
**RESULT:** 
1. Get logs from connect-inject pod to ensure no nil pointer errors
```
PORT 0 <= should be whatever custom value is set in values.yaml (8182)
```
2.  Get meshgateway service.yaml logs
```
apiVersion: v1
kind: Service
...
spec:
 ...
  ports:
  - name: wan
    nodePort: 31432
    port: 443 <= INCORRECT should be 8182
    protocol: TCP
    targetPort: 704 <= INCORRECT should be 8182 + 8000
...
```

- Same as above
- Run unit test for [service.go](gateways/service_test.go)

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/gateways/service.go
+++ b/control-plane/gateways/service.go
@@ -4,6 +4,8 @@
 package gateways
 
 import (
+	"fmt"
+
 	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -19,10 +21,14 @@ func (b *meshGatewayBuilder) Service() *corev1.Service {
 	)
 
 	if b.gcc != nil {
+		fmt.Println("PORT", b.gcc.Spec.Service.Port)
+
+		if b.gcc.Spec.Service.Port != int32(0) {
+			port = b.gcc.Spec.Service.Port
+		}
 		containerConfig = b.gcc.Spec.Deployment.Container
 		portModifier = containerConfig.PortModifier
 		serviceType = *b.gcc.Spec.Service.Type
-		port = b.gcc.Spec.Service.Port
 	}
 
 	return &corev1.Service{
--- a/control-plane/gateways/service.go
+++ b/control-plane/gateways/service.go
@@ -4,26 +4,25 @@
 package gateways
 
 import (
+	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
-
-	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
 )
-
-const port = int32(443)
 
 func (b *meshGatewayBuilder) Service() *corev1.Service {
 	var (
-		containerConfig *meshv2beta1.GatewayClassContainerConfig
 		portModifier    = int32(0)
+		port            = int32(443)
 		serviceType     = corev1.ServiceType("")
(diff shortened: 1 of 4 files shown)


control-plane/gateways/metadata.go
```diff
@@ -143,3 +147,15 @@ func computeAnnotationsOrLabels(inheritFrom map[string]string, primary, secondar
...
 }
+
+// portForService converts "AnnotationGatewayWANPort" whose value is the
+// WAN port for the mesh-gateway service registration to int32
+func (b *meshGatewayBuilder) portForService(annotations map[string]string) int32 {
+	wanPort, err := strconv.Atoi(annotations[constants.AnnotationGatewayWANPort])
+	// do we care about handling this error?
+	if err != nil {
+		fmt.Println(err)
+	}
```
NiniOak, line 158: Should we handle this error?

control-plane/gateways/service.go
```diff
@@ -4,26 +4,25 @@
...
+		port            = int32(443)
 		serviceType     = corev1.ServiceType("")
+		containerConfig *meshv2beta1.GatewayClassContainerConfig
 	)
 
 	if b.gcc != nil {
 		containerConfig = b.gcc.Spec.Deployment.Container
 		portModifier = containerConfig.PortModifier
 		serviceType = *b.gcc.Spec.Service.Type
+		port = b.gcc.Spec.Service.Port
```
t-eckert, line 25: This is where you are getting that error you posted. The type of `Port` is `int32`. This will initialize by default to `0`. So you need an additional check when you determine the port. You'll need to check not only that `gcc` is not `nil`, but also that `b.gcc.Spec.Service.Port` is not `0`. 

What's happening in this case is you are setting `port` to `int32(443)`, then you are checking if `gcc` is not `nil`. It isn't, but `b.gcc.Spec.Service.Port == int32(0)` so you overwrite what you set `port` to. When the `Service` is created, you get an error because the port cannot be `0`.

control-plane/gateways/service.go
```diff
@@ -4,23 +4,28 @@
...
 	var (
-		containerConfig *meshv2beta1.GatewayClassContainerConfig
 		portModifier    = int32(0)
+		port            = int32(443)
 		serviceType     = corev1.ServiceType("")
+		containerConfig *meshv2beta1.GatewayClassContainerConfig
 	)
 
 	if b.gcc != nil {
+		fmt.Println("PORT", b.gcc.Spec.Service.Port)
```
NiniOak, line 24: Required for manual testing, will clean up

control-plane/gateways/service.go
```diff
@@ -4,23 +4,28 @@
...
+
+		if b.gcc.Spec.Service.Port != int32(0) {
```
jm96441n, line 26: I don't think you'll need the `int32(0)` here, the compiler should be smart enough to infer that a `0` is a `int32` (small example https://go.dev/play/p/NsXMC_WN0iI)

control-plane/gateways/service_test.go
```diff
@@ -83,9 +84,9 @@ func Test_meshGatewayBuilder_Service(t *testing.T) {
 					Ports: []corev1.ServicePort{
 						{
 							Name: "wan",
-							Port: int32(443),
```
jm96441n, line 86: should we have a separate path in here showing that when the gcc is not nil and port is not set that we default to 443 as the port?

NiniOak: Great idea! On it now

NiniOak:
Closed. See details [here](https://hashicorp.atlassian.net/browse/NET-6768?focusedCommentId=379100)

PR #3434, opened 2024-01-04
Net 6822 - consul consul k8s add listeners to mesh gateway consume for service

sarahalsmiller:
- Add listeners to meshgateway CRD

- Local install, updating unit tests

- CI passes
- Local helm install

### Checklist ###
- [ X ] Tests added

### DIFF
--- a/charts/consul/templates/crd-meshgateways.yaml
+++ b/charts/consul/templates/crd-meshgateways.yaml
@@ -65,6 +65,7 @@ spec:
                       format: int32
                       type: integer
                   type: object
+                minItems: 1
                 type: array
             type: object
           status:
--- a/charts/consul/templates/crd-meshgateways.yaml
+++ b/charts/consul/templates/crd-meshgateways.yaml
@@ -63,6 +63,8 @@ spec:
                       type: string
                     port:
                       format: int32
+                      maximum: 65535
+                      minimum: 0
                       type: integer
                   type: object
                 minItems: 1
--- a/charts/consul/templates/crd-meshgateways.yaml
+++ b/charts/consul/templates/crd-meshgateways.yaml
@@ -56,6 +56,16 @@ spec:
                 description: GatewayClassName is the name of the GatewayClass used
                   by the MeshGateway
                 type: string
+              listeners:
+                items:
+                  properties:
+                    name:
+                      type: string
+                    port:
+                      format: int32
+                      type: integer
+                  type: object
+                type: array
             type: object
           status:
             properties:
--- a/charts/consul/templates/gateway-resources-configmap.yaml
+++ b/charts/consul/templates/gateway-resources-configmap.yaml
@@ -76,7 +76,7 @@ data:
             {{- end }}
             {{- if .Values.meshGateway.topologySpreadConstraints }}
             topologySpreadConstraints:
-              {{ tpl .Values.meshGateway.topologySpreadConstraints . | nindent 16 | trim }}
+            {{ tpl .Values.meshGateway.topologySpreadConstraints . | nindent 16 | trim }}
             {{- end }}
             {{- if .Values.meshGateway.affinity }}
             affinity: 
@@ -121,5 +121,8 @@ data:
             {{- end }}
         spec:
           gatewayClassName: consul-mesh-gateway
+          listeners: 
+          - name: "wan"
+            port: {{ .Values.meshGateway.service.port }}
   {{- end }}
 {{- end }}
--- a/charts/consul/templates/gateway-resources-configmap.yaml
+++ b/charts/consul/templates/gateway-resources-configmap.yaml
@@ -76,7 +76,7 @@ data:
             {{- end }}
             {{- if .Values.meshGateway.topologySpreadConstraints }}
             topologySpreadConstraints:
-            {{ tpl .Values.meshGateway.topologySpreadConstraints . | nindent 16 | trim }}
+              {{ tpl .Values.meshGateway.topologySpreadConstraints . | nindent 16 | trim }}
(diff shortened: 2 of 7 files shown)


control-plane/gateways/service.go
```diff
@@ -4,15 +4,14 @@
 package gateways
 
 import (
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
```
nathancoleman, line 7: #nit import blocking

control-plane/gateways/service.go
```diff
@@ -36,15 +35,39 @@ func (b *meshGatewayBuilder) Service() *corev1.Service {
...
+
+// Ports build a list of ports from the listener objects. In theory there should only ever be a WAN port on
+// mesh gateway but building the ports from a list of listeners will
+func (b *meshGatewayBuilder) Ports(portModifier int32) []corev1.ServicePort {
+
+	ports := []corev1.ServicePort{}
+
+	if len(b.gateway.Spec.Listeners) == 0 {
+		//If empty use the default value. This should always be set, but in case it's not, this check
+		//will prevent a panic.
```
nathancoleman, line 51: Can we instead add a minimum length requirement to the listeners field on our CRD using kubebuilder annotations? This would make our code simpler

nathancoleman (commented):
So happy to see this! Noice

control-plane/gateways/service.go
NiniOak: But wouldn't that make it harder to test?

sarahalsmiller: I personally would argue that we should be doing both, since the struct where I will add the kubetag is in a different repository but I will defer to others opinions here.

sarahalsmiller: https://github.com/hashicorp/consul/pull/20120

PR to take care of the kubetag

charts/consul/templates/crd-meshgateways.yaml
```diff
@@ -56,6 +56,16 @@ spec:
                 description: GatewayClassName is the name of the GatewayClass used
                   by the MeshGateway
                 type: string
+              listeners:
+                items:
+                  properties:
+                    name:
+                      type: string
+                    port:
+                      format: int32
```
andrewstucki, line 65: Hmm... will there be issues without setting a min here? In the proto definition we have an unsigned integer, which openapi spec doesn't support. It seems like setting `-1` as a listener port at the CRD level would subsequently pass the spec, and I'm guessing depending on how the json unmarshaler handles things we'd either get some weird error unmarshaling the payload and just continually retry, or we'd get some sort of underflow and assign an invalid port number.

Thinking it'd make sense to pin a min/max of 0/65535 for this field.

sarahalsmiller: Working through setting the min, this PR needs to be merged first. https://github.com/hashicorp/consul/pull/20120

sarahalsmiller: https://github.com/hashicorp/consul-k8s/pull/3434/commits/ef8e44ae44fc131d3bab5decf3fd9bbce4a3bc33#diff-ba562f0a244dfbd95791e3d82a24c65b87721dbf89ea2144c4c4b30d2b2cbae1R68 Added

nathancoleman: @sarahalsmiller I think Andrew is saying we should add a second kubebuilder validation for this `port` field to set the min/max value for it

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -74,9 +74,9 @@ data:
             {{- with .Values.meshGateway.dnsPolicy }}
             dnsPolicy: {{ . }}
             {{- end }}
-            {{- if .Values.meshGateway.topologySpreadConstraints }}
+            {{- with .Values.meshGateway.topologySpreadConstraints }}
             topologySpreadConstraints:
-              {{ tpl .Values.meshGateway.topologySpreadConstraints . | nindent 16 | trim }}
+              {{ fromYamlArray . | toJson }}
```
sarahalsmiller, line 79: Small fix to an issue I was having parsing when testing.

charts/consul/templates/crd-meshgateways.yaml
sarahalsmiller: Ah my bad! I was in a hurry and misread. https://github.com/hashicorp/consul/pull/20149 <- to take care of that kubetag

charts/consul/templates/gateway-resources-configmap.yaml
```diff
@@ -121,5 +121,8 @@ data:
             {{- end }}
         spec:
           gatewayClassName: consul-mesh-gateway
+          listeners: 
+          - name: "wan"
```
NiniOak, line 125: Should this value be hardcoded here or in the service.go file?

sarahalsmiller: https://hashicorp.atlassian.net/browse/NET-6822 From the ticket, I believe we want it to be defined on the CRD

charts/consul/templates/crd-meshgateways.yaml
sarahalsmiller: https://github.com/hashicorp/consul-k8s/pull/3434/files#diff-ba562f0a244dfbd95791e3d82a24c65b87721dbf89ea2144c4c4b30d2b2cbae1R66 added

nathancoleman (approved):
Nice work! So cool seeing it all come together 🕺

PR #3463, opened 2024-01-10
Update to supporting 1.13.x

thisisnotashwin:
- Add support for Consul 1.13

- Acceptance tests Consul Helm

### Checklist ###
- [x] Tests added


hack/crds-to-consul-helm/main.go
```diff
@@ -77,7 +77,7 @@ func realMain(helmPathAbs string) error {
 			`    release: {{ .Release.Name }}`,
 			`    component: crd`,
 		}
-		withLabels := append(splitOnNewlines[0:9], append(labelLines, splitOnNewlines[9:]...)...)
+		withLabels := append(splitOnNewlines[0:7], append(labelLines, splitOnNewlines[7:]...)...)
```
thisisnotashwin, line 80: This ensure the labels are correctly copied into the YAML template that is created in helm

wilkermichael (approved):
I witnessed Ashwin do the changes so I don't have too much to add.

Since a lot of it was copy/pasting and I saw the acceptance tests pass locally, I'm approving this from my perspective.

api/v1alpha1/serviceresolver_types_test.go
```diff
@@ -319,7 +320,7 @@ func TestServiceResolver_ToConsul(t *testing.T) {
 							FieldValue: "value",
 							CookieConfig: &capi.CookieConfig{
 								Session: true,
-								TTL:     1,
+								TTL:     1000000000,
```
hashi-derek, line 323: Is this converting correctly?

api/v1alpha1/ingressgateway_types.go
```diff
@@ -193,9 +247,10 @@ func (in *IngressGateway) ToConsul(datacenter string) capi.ConfigEntry {
 	return &capi.IngressGatewayConfigEntry{
 		Kind:      in.ConsulKind(),
 		Name:      in.ConsulName(),
-		TLS:       in.Spec.TLS.toConsul(),
+		TLS:       *in.Spec.TLS.toConsul(),
```
hashi-derek, line 250: Do we need to worry about nil pointer dereferences here?

hashi-derek (commented):
Can we put in a little bit more description here about how we got all the code for these changes? Which branch was it mostly copied from, etc? We should also note that the partition / namespace is not expected to really work in this version due to the open-source clients used in addition to other unsupported behavior that we know of.

api/v1alpha1/ingressgateway_types.go
thisisnotashwin: this file was copy-pasted over, so i think we should be in the clear?

hashi-derek (approved):

PR #3464, opened 2024-01-10
Reorg v2 controllers

jm96441n:
- Reorganize v2 resource controllers from the `config-entries` package to the `resources` package to better indicate what that they are for controlling resources and not config-entries
- rename `ReconcileEntry` method on `ResourcesController` to be `ReconcileResource`

- ran tests

- run tests
- :eye: 

### Checklist ###
- [ ] Tests added


thisisnotashwin (approved):
nice!! i like the re-org!

DanStough (commented):
I wonder if instead of:
* `control-plane/resources/controllers`
* `control-plane/config-entries/controllers`
it would make more sense to have:
* `control-plane/controllers/resources`
* `control-plane/controllers/config-entries`

I think this would be cleaner and meets the original intent, which is controllers that are shared with catalog sync live outside the `connect-inject` folder. I suspect that other v2 controllers will migrate here as well as we come back to catalog sync. I'll take @thisisnotashwin to verify the catalog-sync shared point because he knows infinitely more about it than I do 🧠 .

jm96441n:
@DanStough I like that in that it's a bit cleaner around "controllers live here" and then which controllers are specified by the package

DanStough (approved):
LGTM, conditional on the unit* test that's failing. Looks like maybe it needs to reference the new controller location.

PR #3465, opened 2024-01-11
NET-7179: Update MeshGateway to use new proto with workload selector

NiniOak:

- import the workload selector via go.mod  to be accessible in consul-k8s. See proto PR [here](https://github.com/hashicorp/consul/pull/20159) in Consul
- Ensure workloadSelector field is added to gateways CRD

local install and uninstall of Consul

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/api/mesh/v2beta1/mesh_gateway_types.go
+++ b/control-plane/api/mesh/v2beta1/mesh_gateway_types.go
@@ -70,13 +70,7 @@ func (in *MeshGateway) ResourceID(_, partition string) *pbresource.ID {
 func (in *MeshGateway) Resource(namespace, partition string) *pbresource.Resource {
 	return &pbresource.Resource{
 		Id: in.ResourceID(namespace, partition),
-		Data: inject.ToProtoAny(&pbmesh.MeshGateway{
-			GatewayClassName: in.Spec.GatewayClassName,
-			Listeners:        in.Spec.Listeners,
-			Workloads: &pbcatalog.WorkloadSelector{
-				Prefixes: []string{in.Name},
-			},
-		}),
+		Data: inject.ToProtoAny(&in.Spec),
 		Metadata: meshConfigMeta(),
 	}
 }
--- a/control-plane/api/mesh/v2beta1/mesh_gateway_types.go
+++ b/control-plane/api/mesh/v2beta1/mesh_gateway_types.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v2beta1"
 	pbmesh "github.com/hashicorp/consul/proto-public/pbmesh/v2beta1"
 	"github.com/hashicorp/consul/proto-public/pbresource"
 	"google.golang.org/protobuf/testing/protocmp"
@@ -66,10 +67,19 @@ func (in *MeshGateway) ResourceID(_, partition string) *pbresource.ID {
 	}
 }
 
+// Resource converts the `MeshGateway` CRD into the equivalent Consul resource.
+// This includes adding a pbcatalog.WorkloadSelector that will target the Pods created
+// for this MeshGateway.
 func (in *MeshGateway) Resource(namespace, partition string) *pbresource.Resource {
 	return &pbresource.Resource{
-		Id:       in.ResourceID(namespace, partition),
-		Data:     inject.ToProtoAny(&in.Spec),
+		Id: in.ResourceID(namespace, partition),
+		Data: inject.ToProtoAny(&pbmesh.MeshGateway{
+			GatewayClassName: in.Spec.GatewayClassName,
+			Listeners:        in.Spec.Listeners,
+			Workloads: &pbcatalog.WorkloadSelector{
+				Prefixes: []string{in.Name},
+			},
+		}),
 		Metadata: meshConfigMeta(),
 	}
 }
--- a/control-plane/api/mesh/v2beta1/mesh_gateway_types.go
+++ b/control-plane/api/mesh/v2beta1/mesh_gateway_types.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v2beta1"
 	pbmesh "github.com/hashicorp/consul/proto-public/pbmesh/v2beta1"
 	"github.com/hashicorp/consul/proto-public/pbresource"
 	"google.golang.org/protobuf/testing/protocmp"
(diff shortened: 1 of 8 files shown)


control-plane/go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/hashicorp/consul-k8s/control-plane/cni v0.0.0-20230825213844-4ea04860c5ed
 	github.com/hashicorp/consul-server-connection-manager v0.1.6
 	github.com/hashicorp/consul/api v1.26.1
-	github.com/hashicorp/consul/proto-public v0.5.1
+	github.com/hashicorp/consul/proto-public v0.1.2-0.20240110225000-635cdcf6e070
```
nathancoleman, line 21: Once you've updated the `replace` pin like you did above, you can just drop this change

nathancoleman (commented):
Nice! You got through the hard part.

To update the CRD definitions, you can just run `make ctrl-manifests` from here and see the changes to the 2 `...meshgateways.yaml` files. You should see the `workload` bits pop in after running that command.

The interesting thing is that we actually **don't** want the workload bits to appear in the CRD, so part of this work is figuring out how to make that happen. I suspect that we have 2 options:
- some kubebuilder annotation that tells the CRD generator to skip the field, or
- making our `MeshGateway` Go type a little more custom to omit the workload selector

control-plane/api/mesh/v2beta1/mesh_gateway_types.go
```diff
@@ -38,7 +38,7 @@ type MeshGateway struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   pbmesh.MeshGateway `json:"spec,omitempty"`
+	Spec   pbmesh.MeshGatewaySpec `json:"spec,omitempty"`
```
missylbytes, line 41: So we need to also include the WorkloadSelector on this type, so that we can send it in when we register to Consul. The tricky part is making sure this is not added to the CRD, even though it will be present here.

control-plane/api/mesh/v2beta1/mesh_gateway_types.go
```diff
@@ -4,6 +4,7 @@ package v2beta1
 
 import (
 	"fmt"
+	pbcatalog "github.com/hashicorp/consul/proto-public/pbcatalog/v2beta1"
```
nathancoleman, line 7: #nit import blocking

control-plane/api/mesh/v2beta1/mesh_gateway_types.go
```diff
@@ -38,10 +39,16 @@ type MeshGateway struct {
...
 
-	Spec   pbmesh.MeshGateway `json:"spec,omitempty"`
+	Spec   MeshGatewaySpec `json:"spec,omitempty"`
 	Status `json:"status,omitempty"`
 }
 
+type MeshGatewaySpec struct {
+	// GatewayClassName is the name of the GatewayClass used by the MeshGateway
+	GatewayClassName string                        `json:"gatewayClassName"`
+	Listeners        []*pbmesh.MeshGatewayListener `json:"listeners"`
```
nathancoleman, line 49: The `json` tags wind up differing over in the [generated Consul proto](https://github.com/hashicorp/consul/pull/20159/files#diff-790a8f054f856c26912a0292e2c2a3778edaf3ffc19460037b58eb50fa088ee3R34-R36), which is what we'll map this stuff into.

I think it might be best to just straight up copy these lines with their tags. Thoughts?

control-plane/api/mesh/v2beta1/mesh_gateway_types.go
```diff
@@ -68,8 +75,14 @@ func (in *MeshGateway) ResourceID(_, partition string) *pbresource.ID {
 
 func (in *MeshGateway) Resource(namespace, partition string) *pbresource.Resource {
```
nathancoleman, line 76:
```suggestion

// Resource converts the `MeshGateway` CRD into the equivalent Consul resource.
// This includes adding a pbcatalog.WorkloadSelector that will target the Pods created
// for this MeshGateway.
func (in *MeshGateway) Resource(namespace, partition string) *pbresource.Resource {
```

control-plane/gateways/deployment_test.go
```diff
@@ -64,7 +64,7 @@ func Test_meshGatewayBuilder_Deployment(t *testing.T) {
 							constants.AnnotationGatewayWANAddress: "",
 						},
 					},
-					Spec: pbmesh.MeshGateway{
+					Spec: pbmesh.MeshGatewaySpec{
```
nathancoleman, line 67: I think these tests will need to be updated based on the changes in https://github.com/hashicorp/consul/pull/20159

nathancoleman (changes requested):
Nice work!

control-plane/api/mesh/v2beta1/mesh_gateway_types.go
NiniOak: I'll do that now, would be the best option I think

control-plane/api/mesh/v2beta1/mesh_gateway_types.go
```diff
@@ -66,10 +67,19 @@ func (in *MeshGateway) ResourceID(_, partition string) *pbresource.ID {
...
-		Id:       in.ResourceID(namespace, partition),
-		Data:     inject.ToProtoAny(&in.Spec),
+		Id: in.ResourceID(namespace, partition),
+		Data: inject.ToProtoAny(&pbmesh.MeshGateway{
+			GatewayClassName: in.Spec.GatewayClassName,
+			Listeners:        in.Spec.Listeners,
+			Workloads: &pbcatalog.WorkloadSelector{
+				Prefixes: []string{in.Name},
+			},
+		}),
```
nathancoleman, line 82: I think it makes sense to restore this to its previous state

control-plane/api/mesh/v2beta1/mesh_gateway_types.go
```diff
@@ -66,10 +67,19 @@ func (in *MeshGateway) ResourceID(_, partition string) *pbresource.ID {
 	}
 }
 
+// Resource converts the `MeshGateway` CRD into the equivalent Consul resource.
+// This includes adding a pbcatalog.WorkloadSelector that will target the Pods created
+// for this MeshGateway.
```
nathancoleman, line 72: Since we won't be overwriting the workload selector here

```suggestion
```

nathancoleman (changes requested):
couple of nits on review, about to test it out.

for anyone not in the conversations outside of this PR, we've decided to include the workload selector in the CRD for now and come back to remove it later. This lets us spend our time in the most valuable place possible right now, which is getting TCP traffic working through mesh-gateway.

nathancoleman (approved):
- [x] Confirmed that `MeshGateway` CRD created in consul-k8s includes the workload selector
- [x] Confirmed that `MeshGateway` resource created in consul includes the workload selector
- [x] Confirmed that the `ServiceEndpoints` resource in consul for the mesh gateway still selects the correct `Workloads`
- [x] Confirmed that the `protocol` field on the listeners is written all the way down

PR #3477, opened 2024-01-16
Mesh Gateway Deployment Configuration

t-eckert:
This PR largely wraps a bow on the configuration of deployment resources for Mesh Gateways.

It completes:
- https://hashicorp.atlassian.net/browse/NET-7065
- https://hashicorp.atlassian.net/browse/NET-7064
- https://hashicorp.atlassian.net/browse/NET-6933

- Add BATs for Gateway Log Level Configuration
- Pass logLevel into init-container and dataplane-container
- Add BATs for extraLabels
- Test that extraLabels get set on the deployment
- BATs for annotations
- Use config for log level over gcc if available
- Make consulDataplaneContainer an assoc func to builder
- Use logLevelForDataplaneContainer func
- Test annotations getting set
- Add comments for Builder obj
- Rename config.go to gateway_config.go
- Add comments to gateway_config
- Move commands closer to their configuration
- Extract some constants

- BATS
- Helm generation to verify values get passed int
- Unit testing the configuration generated for deployment resources

This can be reviewed mostly visually. There are certain clean ups I took the liberty of doing that are worth validating.

### Checklist ###
- [x] Tests added


control-plane/gateways/constants.go
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package gateways
+
+const (
+	// General environment variables
+	envPodName      = "POD_NAME"
```
sarahalsmiller, line 8: 🚀

control-plane/gateways/deployment_dataplane_container.go
```diff
@@ -27,7 +27,7 @@ const (
 	volumeName                   = "consul-mesh-inject-data"
 )
 
-func consulDataplaneContainer(config GatewayConfig, containerConfig v2beta1.GatewayClassContainerConfig, name, namespace string) (corev1.Container, error) {
+func (b *meshGatewayBuilder) consulDataplaneContainer(containerConfig v2beta1.GatewayClassContainerConfig) (corev1.Container, error) {
```
sarahalsmiller, line 30: You might be able to get rid of the containerConfig being passed in too, since I believe thats pulled from the GatewayConfig

sarahalsmiller (approved):
Love the cleanups here

control-plane/gateways/metadata_test.go
```diff
@@ -259,6 +259,52 @@ func TestNewMeshGatewayBuilder_Labels(t *testing.T) {
 	}
 }
 
+// The LogLevel for deployment containers may be set on the Gateway Class Config or the Gateway Config.
+// If it is set on both, the Gateway Config takes precedence.
+func TestMeshGatewayBuilder_LogLevel(t *testing.T) {
+	expected := "debug"
```
jm96441n, line 265: minor, but can we name this variable `debug` or something like that, it's a little difficult to parse it being called `expected` and being passed into the test cases

jm96441n (approved):
minor comment but shouldn't hold this back, nice work!

control-plane/gateways/deployment_dataplane_container.go
t-eckert: It is, but there is logic in the Deployment func that sets an override for defaults and I wanted to keep that configuration there so it can be done once and then we don't have to worry about the override behavior within the dataplane container func or have any repetition/accidentally forget to do the override elsewhere.

This way, we know the override work is done and can just use it without thinking too hard about it.

control-plane/gateways/constants.go
t-eckert: TYSMBFF

PR #3522, opened 2024-01-26
feature: [syncCatalog] Add available Region and Zone Kubernetes Topology Metadata when Syncing Kubernetes Services

kolorful:
- Add available Region and Zone Kubernetes Topology Metadata when Syncing Kubernetes Services

- Unit test
- Running locally compiled version on a production Kubernetes cluster and verified it worked (see attached screenshot)

svc1
svc2

- Same as how I tested

### Checklist ###
- [x] Tests added

### DIFF
--- a/.changelog/3522.txt
+++ b/.changelog/3522.txt
@@ -1,3 +1,3 @@
 ```release-note:feature
-control-plane: Add Region and Zone Kubernetes Topology Metadata to Synced Node Port Services.
+control-plane: Add available Region and Zone Kubernetes Topology Metadata when Syncing Kubernetes Services.
 ```
--- a/.changelog/3522.txt
+++ b/.changelog/3522.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+control-plane: Add Region and Zone Kubernetes Topology Metadata to Synced Node Port Services.
+```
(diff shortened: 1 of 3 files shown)


david-yu:
Thanks for the PR. We'll try to see if we can review this after our 1.18 GA is out.

david-yu:
@kolorful Would this PR instead work for you? It incorporates some of your changes as well: https://github.com/hashicorp/consul-k8s/pull/3693

kolorful:
@david-yu yes, I'll close mine since https://github.com/hashicorp/consul-k8s/pull/3693 includes my change and added more features to it. Thanks for letting me know.

david-yu:
@kolorful could you re-raise this PR? We may want to get this in as https://github.com/hashicorp/consul-k8s/pull/3693 is probably going to take a while to get in since there is a lot more work to do endpoint slice reconciliation, I believe your work gets the topology info from the node as opposed to the endpoint slice which is probably ok.

kolorful:
@david-yu, sure I reopend this PR

jukie:
Would it be acceptable to add zone/region for all service types?

kolorful:
@jukie sure I added the code and please take a look.

jukie:
I'm not a maintainer but one thing I was thinking about in my PR as well was how to avoid calling a get node for each endpoint. Would that be a concern for large services and add too much strain on the api server?
What would you think about creating a map of nodenames and endpoints or addresses and then doing a single get nodes call and then grabbing topology info from the result? That may add more complexity than you'd prefer.

kolorful:
This can be closed since https://github.com/hashicorp/consul-k8s/pull/3693 has the change.

PR #3526, opened 2024-01-29
[NET-7492, NET-7495] Support -server-watch-disabled, use -proxy-* args instead of -service-*

nathancoleman:
- Adjust envars for consul-dataplane to use the `-proxy-` args instead of the deprecated `-service-` args. This is important because it's either-or in consul-dataplane [here](https://github.com/hashicorp/consul-dataplane/blob/b80f2bf4a4876789ae0b524275b8201e692b7177/cmd/consul-dataplane/config.go#L270-L288) which is biting us because we set the `-proxy-id` arg today which pushes into the initial `if` case and ignores our `-service-*` flags.
- Add support for disabling server watch, which is used the the consul-servers are behind a load balancer, such as when a second admin partition communicates via the `expose-consul-servers` K8s `Service`.

Deploy a second admin partition with `externalServers.skipServerWatch: true` in your `values.yaml`. Verify that the arg is set on the mesh gateway's dataplane container and that the dataplane container is only trying to connect to the public IP of the `expose-consul-servers` K8s `Service (via logs).

See above

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/gateways/deployment_dataplane_container.go
+++ b/control-plane/gateways/deployment_dataplane_container.go
@@ -93,10 +93,6 @@ func (b *meshGatewayBuilder) consulDataplaneContainer(containerConfig v2beta1.Ga
 				Name:  envDPCredentialLoginMeta,
 				Value: "pod=$(POD_NAMESPACE)/$(DP_PROXY_ID)",
 			},
-			{
-				Name:  envDPServiceNodeName,
-				Value: "$(NODE_NAME)-virtual",
-			},
 		},
 		VolumeMounts: []corev1.VolumeMount{
 			{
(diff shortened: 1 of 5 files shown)


control-plane/gateways/deployment_dataplane_container.go
```diff
@@ -93,10 +93,6 @@ func (b *meshGatewayBuilder) consulDataplaneContainer(containerConfig v2beta1.Ga
 				Name:  envDPCredentialLoginMeta,
 				Value: "pod=$(POD_NAMESPACE)/$(DP_PROXY_ID)",
 			},
-			{
-				Name:  envDPServiceNodeName,
-				Value: "$(NODE_NAME)-virtual",
-			},
```
nathancoleman, line 99: The flags in consul-dataplane [indicate](https://github.com/hashicorp/consul-dataplane/blob/b80f2bf4a4876789ae0b524275b8201e692b7177/cmd/consul-dataplane/main.go#L59C1-L61C57) that this envar is deprecated and should use `proxy-node-name` instead; however, the docs for `proxy-node-name` indicate that it's ignored for the v2 API.

sarahalsmiller (approved):
Good catch

PR #3528, opened 2024-01-30
[NET-7534] v2: Make port names in consul-k8s compatible with NET-5586

ndhanushkodi:
- [This change in consul](https://github.com/hashicorp/consul/pull/20371) involves now interpreting whether xRoute/FailoverPolicy/DestinationPolicy resource service references use either the service port (virtualPort in consul) or service target port (targetPort in consul). To make this decision unambiguously:
> This change updates our interpretation of these reference fields/keys (parent, backend, destination), s.t.:
>
> * A numeric value will be exclusively interpreted to indicate a ServicePort.virtual_port
> * A non-numeric value will be exclusively interpreted to indicate a ServicePort.target_port (this supports VMs/Nomad and other cases where network virtual ports are not used, and port names are expected to be in reference to workload ports, not service ports)

- If a K8s service targetport is allowed to be the stringified version of a number, it will be ambiguous in consul what to interpret the string "portID" as. 

- This change makes it such that the string port can never be a number, and will always also have alpha characters by prefixing "cslport-" to the workload port if the workload port name is unspecified.

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/connect-inject/common/common.go
+++ b/control-plane/connect-inject/common/common.go
@@ -82,7 +82,7 @@ func WorkloadPortName(port *corev1.ContainerPort) string {
 		isNum = true
 	}
 	if name == "" || isNum {
-		name = "cslport-" + strconv.Itoa(int(port.ContainerPort))
+		name = constants.UnnamedWorkloadPortNamePrefix + strconv.Itoa(int(port.ContainerPort))
 	}
 	return name
 }
--- a/control-plane/connect-inject/common/common.go
+++ b/control-plane/connect-inject/common/common.go
@@ -73,6 +73,20 @@ func PortValue(pod corev1.Pod, value string) (int32, error) {
 	return int32(raw), err
 }
 
+// WorkloadPortName returns the container port's name if it has one, and if not, constructs a name from the port number
+// and adds a constant prefix. The port name must be 1-15 characters and must have at least 1 alpha character.
+func WorkloadPortName(port *corev1.ContainerPort) string {
+	name := port.Name
+	var isNum bool
+	if _, err := strconv.Atoi(name); err == nil {
+		isNum = true
+	}
+	if name == "" || isNum {
+		name = "cslport-" + strconv.Itoa(int(port.ContainerPort))
+	}
+	return name
+}
+
 // TransparentProxyEnabled returns true if transparent proxy should be enabled for this pod.
 // It returns an error when the annotation value cannot be parsed by strconv.ParseBool or if we are unable
 // to read the pod's namespace label when it exists.
--- a/control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/go-logr/logr"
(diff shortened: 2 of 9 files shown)


thisisnotashwin (approved):

control-plane/connect-inject/controllers/endpointsv2/endpoints_controller.go
```diff
@@ -493,7 +498,7 @@ func getEffectiveTargetPort(targetPort intstr.IntOrString, prefixedPods selector
 
 	// If still no match for the target port, fall back to string-ifying the target port name, which
 	// is what the PodController will do when converting unnamed ContainerPorts to Workload ports.
-	return targetPort.String()
+	return "cslport-" + targetPort.String()
```
rboyer, line 501: This feels like it belongs in some constants package.

ndhanushkodi: Ah yep will refactor

control-plane/connect-inject/common/common.go
```diff
@@ -73,6 +73,20 @@ func PortValue(pod corev1.Pod, value string) (int32, error) {
...
+// WorkloadPortName returns the container port's name if it has one, and if not, constructs a name from the port number
+// and adds a constant prefix. The port name must be 1-15 characters and must have at least 1 alpha character.
+func WorkloadPortName(port *corev1.ContainerPort) string {
+	name := port.Name
+	var isNum bool
+	if _, err := strconv.Atoi(name); err == nil {
+		isNum = true
+	}
+	if name == "" || isNum {
+		name = "cslport-" + strconv.Itoa(int(port.ContainerPort))
```
jm96441n, line 85: should we use the constant `UnnamedWorkloadPortNamePrefix` here?

control-plane/connect-inject/common/common.go
```diff
@@ -73,6 +73,20 @@ func PortValue(pod corev1.Pod, value string) (int32, error) {
 	return int32(raw), err
 }
 
+// WorkloadPortName returns the container port's name if it has one, and if not, constructs a name from the port number
+// and adds a constant prefix. The port name must be 1-15 characters and must have at least 1 alpha character.
+func WorkloadPortName(port *corev1.ContainerPort) string {
```
jm96441n, line 78: can we add some unit tests for this function?

control-plane/connect-inject/webhookv2/mesh_webhook.go
```diff
@@ -251,6 +252,16 @@ func (w *MeshWebhook) Handle(ctx context.Context, req admission.Request) admissi
 
 	w.Log.Info("received pod", "name", req.Name, "ns", req.Namespace)
 
+	// Validate that none of the pod ports start with the prefix "cslport-" as that may result in conflicts with ports
+	// created by the pod controller when creating workloads.
+	for _, c := range pod.Spec.Containers {
+		for _, p := range c.Ports {
+			if strings.HasPrefix(p.Name, constants.UnnamedWorkloadPortNamePrefix) {
```
jm96441n, line 259: not necessarily for this PR but are we tracking anywhere restrictions like this that we should document for end users?

control-plane/connect-inject/common/common.go
ndhanushkodi: ah yep i forgot to add unit tests on this side after refactoring

control-plane/connect-inject/common/common.go
ndhanushkodi: yup updated thanks

control-plane/connect-inject/webhookv2/mesh_webhook.go
ndhanushkodi: For this particular case I'd not expect users to run into this and would expect if they do that it's easily discoverable through the pod not being accepted for injection. It definitely feels like a thing that'd be glazed over in any docs. Any other restrictions (so far) are enforced through kube validations

jm96441n (approved):
LGTM! Thanks for jumping on this!

control-plane/connect-inject/common/common.go
```diff
@@ -73,6 +73,20 @@ func PortValue(pod corev1.Pod, value string) (int32, error) {
...
+	name := port.Name
+	var isNum bool
+	if _, err := strconv.Atoi(name); err == nil {
```
rboyer, line 81: codegolf, nonblocking, unnecessary optimization:

You don't have to try to `Atoi` the empty string, since you know it won't work.

ndhanushkodi: Ah right, makes sense, I may skip for now because just trying to get the tests green to merge before code freeze 🤞

zalimeni:
For posterity, ACK this change and thank you @ndhanushkodi for handling it!

PR #3532, opened 2024-02-01
[NET-6417] Fix meshgw tests

jm96441n:
- fixes tests for meshgw that were broken by adding https://github.com/hashicorp/consul/pull/20425 validation

ran tests

### Checklist ###
- [X] Tests added

### DIFF
--- a/control-plane/controllers/resources/mesh_gateway_controller_test.go
+++ b/control-plane/controllers/resources/mesh_gateway_controller_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 
 	logrtest "github.com/go-logr/logr/testr"
-	"github.com/hashicorp/consul/sdk/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


control-plane/controllers/resources/mesh_gateway_controller_test.go
```diff
@@ -444,7 +586,7 @@ func TestMeshGatewayController_Reconcile(t *testing.T) {
 
 			res, err := controller.Reconcile(context.Background(), testCase.request)
 			if testCase.expectedErr != nil {
-				//require.EqualError(t, err, testCase.expectedErr.Error())
+				// require.EqualError(t, err, testCase.expectedErr.Error())
```
nathancoleman, line 589: Should this be here? Out of scope for this PR, just curious

control-plane/controllers/resources/mesh_gateway_controller_test.go
```diff
@@ -50,6 +52,16 @@ func TestMeshGatewayController_Reconcile(t *testing.T) {
 						Namespace: "consul",
 						Name:      "mesh-gateway",
 					},
+					Spec: pbmesh.MeshGateway{
+						GatewayClassName: "consul",
+						Listeners: []*pbmesh.MeshGatewayListener{
+							{
+								Name:     "wan",
+								Port:     8443,
+								Protocol: "mesh",
```
nathancoleman, line 61: This is currently required to be `TCP` [here](https://github.com/hashicorp/consul/blob/6a8554317502dc54fb794ce950db8678ec352c55/proto-public/pbmesh/v2beta1/mesh_gateway.proto#L29-L30). Will be explicitly `mesh` in the future; however, I'd make it TCP for now in these tests so as not to run into any surprises.

control-plane/controllers/resources/mesh_gateway_controller_test.go
jm96441n: honestly I don't know, this was gofumpt editing the spacing on the comment

nathancoleman (approved):
Thanks for picking up and fixing this issue so quickly! 🥇

PR #3538, opened 2024-02-01
[NET-7156] Generalize MeshGatewayBuilder to just GatewayBuilder

jm96441n:
- Generalize the `MeshGatewayBuilder` to be a generic `GatewayBuilder`

Ran tests

### Checklist ###
- [X] Tests added


control-plane/gateways/builder.go
```diff
@@ -5,21 +5,31 @@ package gateways
...
+	GetLabels() map[string]string
+}
+
+// gatewayBuilder is a helper struct for building the Kubernetes resources for a mesh gateway.
 // This includes Deployment, Role, Service, and ServiceAccount resources.
 // Configuration is combined from the MeshGateway, GatewayConfig, and GatewayClassConfig.
-type meshGatewayBuilder struct {
-	gateway *meshv2beta1.MeshGateway
-	config  GatewayConfig
+type gatewayBuilder[T Gateway] struct {
```
jm96441n, line 23: the generic here allows us to restrict  this to only gateway types, I wasn't a huge fan of making all these use getters in the code, but the current generics implementation doesn't allow for struct fields. the other option I considered was having the `gatewayBuilder` take the name/namespace/annotations/labels as arguments to the builder rather than passing the entire gateway, but there is the issue of the listeners where the behavior of converting listeners to ports for mesh gateways is slightly different from api gateways.

jm96441n:
unrelated: github does not allow commenting on unchanged lines but I need to update https://github.com/hashicorp/consul-k8s/pull/3538/files#diff-abdfd222e5e5dbd68aefadbed3a94141d93833e7977af4832d035538aef9fe4eL120 for building the container

sarahalsmiller (approved):
Great work! Thanks for walking me through it

PR #3540, opened 2024-02-02
Net 7279 consul k8s write failing acceptance test for tcp route

sarahalsmiller:
- Write acceptance test to run TCPRoute feature against when feature is complete 

- Test builds run and fails when TEST_APIGW_V2=true
- Test skips when TEST_APIGW_V2 is not set

- Look over

### Checklist ###
- [ X ] Tests added


acceptance/tests/fixtures/bases/api-gateway/kustomization.yaml
```diff
@@ -5,5 +5,5 @@ resources:
   - gatewayclassconfig.yaml
   - gatewayclass.yaml
   - apigateway.yaml
-  - httproute.yaml
+  - tcproute.yaml
```
jm96441n, line 8: is there a reason we needed to change the existing setups?

acceptance/tests/fixtures/bases/api-gateway-v2/meshservice.yaml
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+apiVersion: consul.hashicorp.com/v1alpha1
```
jm96441n, line 4: I'm not sure we need the MeshService here because we're not talking to any services outside of the cluster

acceptance/tests/fixtures/bases/api-gateway-v2/apigateway.yaml
```diff
@@ -0,0 +1,31 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+apiVersion: mesh.consul.hashicorp.com/v2beta1
+kind: APIGateway
+metadata:
+  name: gateway
+spec:
+  gatewayClassName: gateway-class
+  listeners:
+  - protocol: HTTP
```
jm96441n, line 11: I think for now we can just have the tcp listener since we're not implementing any of the http listener/route functionality just yet

acceptance/tests/api-gateway_v2/api_gateway_v2_test.go
```diff
@@ -0,0 +1,186 @@
...
+			// We use the static-client pod so that we can make calls to the api gateway
+			// via kubectl exec without needing a route into the cluster from the test machine.
+			logger.Log(t, "creating static-client pod")
+			k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.NoCleanup, cfg.DebugDirectory, "../fixtures/bases/static-client")
+
+			logger.Log(t, "creating target tcp server")
+			k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.NoCleanup, cfg.DebugDirectory, "../fixtures/bases/static-server-tcp")
+
+			logger.Log(t, "creating tcp-route")
+			k8s.RunKubectl(t, ctx.KubectlOptions(t), "apply", "-f", "../fixtures/cases/api-gateways/tcproute/route.yaml")
```
jm96441n, line 103: I think this route needs to be `api-gateways-v2` right? because this will end up using the kubesig route?

acceptance/tests/api-gateway/api_gateway_v2_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
+	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Test that api gateway basic functionality works in a default installation and a secure installation for V2
+func TestAPIGateway_V2_Basic(t *testing.T) {
```
jm96441n, line 29: do we need this in both the api-gateway and api-gateway-v2 directories?

acceptance/tests/api-gateway_v2/main_test.go
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package apigateway_v2
```
jm96441n, line 4: minor but can we name this package `apigatewayv2` https://go.dev/doc/effective_go#package-names

acceptance/tests/api-gateway_v2/api_gateway_v2_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/hashicorp/consul-k8s/acceptance/framework/k8s"
+	"github.com/hashicorp/consul-k8s/acceptance/framework/logger"
+	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Test that api gateway basic functionality works in a default installation and a secure installation for V2
+func TestAPIGateway_V2_Basic(t *testing.T) {
```
jm96441n, line 28: should we have an env var check here and skip if it's not set/empty?

sarahalsmiller: When I tested the env check in the [main](https://github.com/hashicorp/consul-k8s/pull/3540/files#diff-73333c5946f976c570d1673d58af6ed82da06fa4cdb446c5680901c1ac862d21R17) was sufficient for the entire package to be skipped.

acceptance/tests/api-gateway/api_gateway_v2_test.go
sarahalsmiller: Nope. Missed deleting that, good catch

acceptance/tests/fixtures/bases/api-gateway/kustomization.yaml
sarahalsmiller: I believe goland must have gone this automatically when I was renaming things. I have reverted it.

PR #3541, opened 2024-02-02
updated script to point at RC version correctly

wilkermichael:
-
-

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/build-support/scripts/consul-enterprise-version.sh
+++ b/control-plane/build-support/scripts/consul-enterprise-version.sh
@@ -6,8 +6,6 @@ VERSION=$(yq .global.image $FILE)
 
 if [[ !"${VERSION}" == *"hashicorp/consul:"* ]]; then
 	VERSION=$(echo ${VERSION} | sed "s/consul:/consul-enterprise:/g" | sed "s/$/-ent/g")
-elif [[ !"${VERSION}" == *"-rc"* ]]; then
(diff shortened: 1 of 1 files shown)


control-plane/build-support/scripts/consul-enterprise-version.sh
```diff
@@ -6,8 +6,6 @@ VERSION=$(yq .global.image $FILE)
 
 if [[ !"${VERSION}" == *"hashicorp/consul:"* ]]; then
```
andrewstucki, line 7: So, this line is still problematic, not sure if you want to fix it here. Right now let's say you have `VERSION=foobar`, what this check is going to be testing the equality of is:

```
"!foobar" == *"hashicorp/consul:"*
```

the globs still work fine, but the bang prefix is just going to get appended as a string token, you can check that with:

```
if [[ !"${VERSION}" == "!${VERSION}" ]]; then
  echo "bang"
fi
```

which should print out an echo. Problem is if you fix this, then the reference to `hashicorppreview/consul` in the default values file is properly going to hit this conditional and add a `-ent` suffix.

I'm cool with dropping the below conditional, but not sure what you want to do with this line -- it's not functioning as intended, but if functioning as intended, it'll break the build due to the lack of `-ent` tags on the preview repo. So... do you know what this line is _supposed_ to be for?

PR #3548, opened 2024-02-05
Ensure signals are passed to commands

lkysow:
Change `/bin/sh -ec "<command>"` to
`/bin/sh -ec "exec <command>"`. Adding `exec` ensures that `<command>` is not executed as a child process but replaces the `/bin/sh` process. This ensure that `<command>` receives any signals.

Specifically this is an issue when attempting to trap SIGTERMs as part of graceful pod shutdown. Without this change, we weren't receiving any signals because they aren't passed down by `/bin/sh -c`.

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.changelog/3548.txt
+++ b/.changelog/3548.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+helm: Change `/bin/sh -ec "<command>"` to `/bin/sh -ec "exec <command>"` in helm deployments
+```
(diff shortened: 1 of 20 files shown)


thisisnotashwin:
waiting for the helm tests to go green before it being ok to merge

lkysow:
yeah I'm waiting for the acceptance tests to finish before pushing the unit test fix

thisisnotashwin:
acceptance tests: https://github.com/hashicorp/consul-k8s-workflows/actions/runs/7788341139

lkysow:
All good, thanks! How far do you think we should backport?

thisisnotashwin:
we will cut patch releases all the way back to 1.1 before we release 1.4 so all 4 branches.

PR #3549, opened 2024-02-05
Add NET_BIND_SERVICE to the security context in the deployment of Mesh Gateway (NET-6463)

missylbytes:
Adds the NET_BIND_SERVICE to the security context in the deployment of the Mesh Gateway for the Consul Dataplane container.

This is related to this PR https://github.com/hashicorp/consul-dataplane/pull/238 in the Consul Dataplane
NET-6463
- Manual testing to see the security context is added to the Consul Dataplane container

```
kubectl get deployment -n consul consul-consul-mesh-gateway -o yaml
# Below has been edited for brevity
apiVersion: apps/v1
kind: Deployment
metadata:
  name: consul-consul-mesh-gateway
  namespace: consul
spec:
  template:
    spec:
      containers:
        - args:
            - -addresses=consul-consul-server.consul.svc
            - -grpc-port=8502
            - -proxy-service-id-path=/consul/service/proxy-id
            - -ca-certs=/consul/tls/ca/tls.crt
            - -credential-type=login
            - -login-bearer-token-path=/var/run/secrets/kubernetes.io/serviceaccount/token
            - -login-auth-method=consul-consul-k8s-component-auth-method
            - -login-partition=default
            - -service-partition=default
            - -log-level=info
            - -log-json=false
          command:
            - consul-dataplane
          image: docker.io/hashicorppreview/consul-dataplane:1.2.5-ubi
          imagePullPolicy: IfNotPresent
          securityContext:
            capabilities:
              add:
                - NET_BIND_SERVICE
              drop:
                - ALL
 ```
## How I expect others to test this PR ##
You cant test this without openshift, just install the branch with MeshGateways enabled and check to see if there is a security context.

Someone else was kind enough to test on Openshift:
image

### Checklist ###
- [x] Tests added

### CODE REVIEW

**Comment by sophie-gairo** [comment]
Added @lkysow as an additional optional reviewer being one of the few who has openshift context in consul. This is for a 5/3rd fix looking to go out with the next RC

**Review by jm96441n — APPROVED** [review]
LGTM

File: charts/consul/test/unit/mesh-gateway-deployment.bats [hunk 1]
```diff
@@ -1783,4 +1783,21 @@ key2: value2' \
   local actual=$(echo "$cmd" |
     yq 'any(contains("-log-level=warn"))' | tee /dev/stderr)
   [ "${actual}" = "true" ]
+}
+
+#--------------------------------------------------------------------
+# security context
```
lkysow, line 1789: We don't need to add tests for things that aren't conditional since it's basically just a tautology at that point. We just add them for things that are conditional

lkysow:
This isn't openshift specific is it? securityContext on the container is a regular kube setting I thought

lkysow:
Is there a reason we're not copying ingress and terminating gateways and doing:
```
        {{- include "consul.restrictedSecurityContext" $ | nindent 8 }}
```

lkysow:
Ah, I found the reason here (https://github.com/hashicorp/consul-k8s/pull/2702)

> This is not included because adding the "restricted" security context settings to MGW makes it impossible to use hostNetwork=true

Would this change also affect this?

david-yu:
Is there a reason there is no changelog for this PR?

missylbytes:
> Is there perhaps a reason there is no changelog for this PR?

@david-yu I just haven't made one yet and didn't want CI to yell at me.

charts/consul/templates/mesh-gateway-deployment.yaml
```diff
@@ -184,6 +184,12 @@ spec:
       containers:
       - name: mesh-gateway
         image: {{ .Values.global.imageConsulDataplane | quote }}
+        securityContext:
+          capabilities:
+            drop:
+              - ALL
```
nathancoleman, line 190: Noting that we talked a few minutes ago about gating this `drop: [ALL]` behind `not .Values.meshGateway.hostNetwork` so that we still allow that to work for the few users who require it.

This will leave us with more work to do to lock down the security context specifically for that `hostNetwork` use case but will unblock all other use cases who can't upgrade right now due to the fact that mesh gateway doesn't request `NET_BIND_SERVICE` when it's actually required.

lkysow: Sounds great (in this case we would want a unit test to test the conditional. That makes a lot of sense to me, not sure why that wasn't done in the original PR.

missylbytes: Updated with fixes from these comments/discussions

charts/consul/test/unit/mesh-gateway-deployment.bats
```diff
@@ -1783,4 +1783,21 @@ key2: value2' \
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/mesh-gateway-deployment.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'meshGateway.hostNetwork=true' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].securityContext' | tee /dev/stderr)
+
+  [ $(echo "${actual}" | yq -r '.capabilities.add[0]') = "NET_BIND_SERVICE" ]
```
lkysow, line 1801: this doesn't test that we don't set `drop: all`, just that we set `add`

lkysow (approved):
LGTM after test is fixed.

charts/consul/test/unit/mesh-gateway-deployment.bats
lkysow: should be two tests: one that tests we set drop: all by default, another that tests we don't set it when hostnetwork=true

PR #3558, opened 2024-02-06
Ensure signals are passed to commands (#3548)

thisisnotashwin:
* Ensure signals are passed to commands

Change `/bin/sh -ec "<command>"` to
`/bin/sh -ec "exec <command>"`. Adding `exec` ensures that `<command>` is not executed as a child process but replaces the `/bin/sh` process. This ensure that `<command>` receives any signals.

Specifically this is an issue when attempting to trap SIGTERMs as part of graceful pod shutdown. Without this change, we weren't receiving any signals because they aren't passed down by `/bin/sh -c`.

* Fix broken bats tests and add changelog

---------

-
-

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.changelog/3548.txt
+++ b/.changelog/3548.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+helm: Change `/bin/sh -ec "<command>"` to `/bin/sh -ec "exec <command>"` in helm deployments
+```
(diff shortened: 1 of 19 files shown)


thisisnotashwin:
Closes #3554

wilkermichael (approved):

PR #3559, opened 2024-02-06
[NET-7657,NET-6934] Define v2 GatewayClass + GatewayClassConfig locally

nathancoleman:
The `spec` field for v2 `GatewayClass`es currently uses the proto-public definitions from Consul; however, we don't want to do that anymore since Consul has no need for this type at all.

We've already broken this dependency for `GatewayClassConfig` in the past but need to do the same for `GatewayClass` so that we can remove the proto definition from Consul altogether.

- `helm install` with `.meshGateway.enabled=true` and `.global.imageK8S` set to the build of this branch
- verify that everything comes up correctly.

See above

### Checklist ###
- [x] Tests updated


control-plane/api/mesh/v2beta1/gateway_class_types.go
```diff
@@ -50,97 +37,23 @@ type GatewayClassList struct {
 	Items           []*GatewayClass `json:"items"`
 }
 
-func (in *GatewayClass) ResourceID(namespace, partition string) *pbresource.ID {
```
nathancoleman, line 53: These funcs all relate to write/reading the type from Consul; however, we don't and will not be doing that. Removing the interface implementation helps prevent someone from accidentally trying to write this type through to Consul

control-plane/api/mesh/v2beta1/gateway_class_types.go
```diff
@@ -37,7 +24,7 @@ type GatewayClass struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   pbmesh.GatewayClass `json:"spec,omitempty"`
+	Spec   GatewayClassSpec `json:"spec,omitempty"`
```
nathancoleman, line 27: This is the crux of the whole PR where we no longer consume the proto type from Consul

missylbytes (approved):

PR #3574, opened 2024-02-07
[NET-7156] Gateways Controllers Reusability

jm96441n:
- adds indices onto the gateway resources in k8s for faster lookups as well as easier to read code
- makes the setup for a gateway controller generic so it can be reused across gateway types

- ran mesh gateways before and after, all works the same

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/controllers/resources/gateway_controller_setup.go
+++ b/control-plane/controllers/resources/gateway_controller_setup.go
@@ -124,7 +124,7 @@ func getGatewayClassesReferencingGatewayClassConfig(ctx context.Context, k8sClie
 }
 
 // getGatewaysReferencingGatewayClass queries all xGateway resources in the cluster
-// and returns any that reference the given GatewayClass.
+// and returns any that reference the given GatewayClass by name.
 func getGatewaysReferencingGatewayClass[T gatewayList](ctx context.Context, k8sClient client.Client, className string, index indexName) (T, error) {
 	var allGateways T
 	if err := k8sClient.List(ctx, allGateways, &client.ListOptions{
--- a/control-plane/controllers/resources/gateway_controller_setup.go
+++ b/control-plane/controllers/resources/gateway_controller_setup.go
@@ -8,7 +8,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -22,7 +21,7 @@ type gatewayList interface {
 	ReconcileRequests() []reconcile.Request
 }
 
-func setupGatewayControllerWithManager[L gatewayList](mgr ctrl.Manager, obj client.Object, k8sClient client.Client, gwc reconcile.Reconciler) error {
+func setupGatewayControllerWithManager[L gatewayList](mgr ctrl.Manager, obj client.Object, k8sClient client.Client, gwc reconcile.Reconciler, index indexName) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(obj).
 		Owns(&appsv1.Deployment{}).
@@ -33,7 +32,12 @@ func setupGatewayControllerWithManager[L gatewayList](mgr ctrl.Manager, obj clie
 		Watches(
 			source.NewKindWithCache(&meshv2beta1.GatewayClass{}, mgr.GetCache()),
 			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
-				gateways, err := getGatewaysReferencingGatewayClass[L](context.Background(), k8sClient, o.(*meshv2beta1.GatewayClass))
+				gc := o.(*meshv2beta1.GatewayClass)
+				if gc == nil {
+					return nil
+				}
+
+				gateways, err := getGatewaysReferencingGatewayClass[L](context.Background(), k8sClient, gc.Name, index)
 				if err != nil {
 					return nil
 				}
@@ -43,14 +47,23 @@ func setupGatewayControllerWithManager[L gatewayList](mgr ctrl.Manager, obj clie
 		Watches(
 			source.NewKindWithCache(&meshv2beta1.GatewayClassConfig{}, mgr.GetCache()),
 			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
-				classes, err := getGatewayClassesReferencingGatewayClassConfig(context.Background(), k8sClient, o.(*meshv2beta1.GatewayClassConfig))
+				gcc := o.(*meshv2beta1.GatewayClassConfig)
+				if gcc == nil {
+					return nil
+				}
+
+				classes, err := getGatewayClassesByGatewayClassConfigName(context.Background(), k8sClient, gcc.Name)
 				if err != nil {
 					return nil
 				}
 
 				var requests []reconcile.Request
 				for _, class := range classes.Items {
-					gateways, err := getGatewaysReferencingGatewayClass[L](context.Background(), k8sClient, class)
+					if class == nil {
+						continue
+					}
+
+					gateways, err := getGatewaysReferencingGatewayClass[L](context.Background(), k8sClient, class.Name, index)
 					if err != nil {
 						continue
 					}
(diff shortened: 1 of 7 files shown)


control-plane/controllers/resources/gateway_indices.go
```diff
@@ -0,0 +1,70 @@
...
+		indexerFunc: gatewayClassForGateway,
+	},
+}
+
+// gatewayClassConfigForGatewayClass creates an index of every GatewayClassConfig referenced by a GatewayClass.
+func gatewayClassConfigForGatewayClass(o client.Object) []string {
+	gc := o.(*meshv2beta1.GatewayClass)
+
+	pr := gc.Spec.ParametersRef
+	if pr != nil && pr.Kind == "GatewayClassConfig" {
```
nathancoleman, line 59: there's a constant for this now over in `gateway_class_config_types.go`

control-plane/controllers/resources/gateway_indices.go
```diff
@@ -0,0 +1,70 @@
...
+		return []string{pr.Name}
+	}
+
+	return []string{}
+}
```
nathancoleman, line 64: I'd update the docstring and function names here to reflect that this function isn't actually doing any indexing itself and doesn't return the full type, just the name of one.

I'd also be fine with inlining it above since it doesn't appear to be re-used anywhere

```suggestion
// mapGatewayClassToGatewayClassConfigName returns the name of the GatewayClassConfig referenced by a GatewayClass.
func mapGatewayClassToGatewayClassConfigName(o client.Object) []string {
	gc := o.(*meshv2beta1.GatewayClass)

	pr := gc.Spec.ParametersRef
	if pr != nil && pr.Kind == "GatewayClassConfig" {
		return []string{pr.Name}
	}

	return []string{}
}
```

control-plane/controllers/resources/gateway_indices.go
```diff
@@ -0,0 +1,70 @@
...
+
+// gatewayClassForGateway creates an index of every GatewayClass referenced by a Gateway.
+func gatewayClassForGateway(o client.Object) []string {
+	g := o.(*meshv2beta1.APIGateway)
+	return []string{string(g.Spec.GatewayClassName)}
+}
```
nathancoleman, line 70: Similarly here

```suggestion
// mapGatewayToGatewayClassName returns the name of the GatewayClass referenced by a Gateway
func mapGatewayToGatewayClassName(o client.Object) []string {
	g := o.(*meshv2beta1.APIGateway)
	return []string{string(g.Spec.GatewayClassName)}
}
```

control-plane/controllers/resources/gateway_controller_setup.go
```diff
@@ -0,0 +1,141 @@
...
+
+func getGatewayClassConfigForGatewayClass(ctx context.Context, k8sClient client.Client, gatewayClass *meshv2beta1.GatewayClass) (*meshv2beta1.GatewayClassConfig, error) {
+	if gatewayClass == nil {
+		// if we don't have a gateway class we can't fetch the corresponding config
+		return nil, nil
+	}
+
+	config := &meshv2beta1.GatewayClassConfig{}
+	if ref := gatewayClass.Spec.ParametersRef; ref != nil {
+		if ref.Group != meshv2beta1.MeshGroup || ref.Kind != "GatewayClassConfig" {
```
nathancoleman, line 74: There's a constant for this now in `gateway_class_config_types.go`

control-plane/controllers/resources/gateway_controller_setup.go
```diff
@@ -0,0 +1,141 @@
...
+
+		if err := k8sClient.Get(ctx, types.NamespacedName{Name: ref.Name}, config); err != nil {
+			return nil, client.IgnoreNotFound(err)
+		}
+	}
+	return config, nil
+}
+
+func getGatewayClassForGateway(ctx context.Context, k8sClient client.Client, className string) (*meshv2beta1.GatewayClass, error) {
+	var gatewayClass meshv2beta1.GatewayClass
```
nathancoleman, line 87: I find the name here confusing since it doesn't actually see or care about the `Gateway` at all. It's really just a fetcher, `getGatewayClass` or `getGatewayClassByName` or something like that

control-plane/controllers/resources/gateway_controller_setup.go
```diff
@@ -0,0 +1,141 @@
...
+// getGatewayClassesReferencingGatewayClassConfig queries all GatewayClass resources in the
+// cluster and returns any that reference the given GatewayClassConfig.
+func getGatewayClassesReferencingGatewayClassConfig(ctx context.Context, k8sClient client.Client, config *meshv2beta1.GatewayClassConfig) (*meshv2beta1.GatewayClassList, error) {
+	if config == nil {
+		return nil, nil
+	}
+
+	allClasses := &meshv2beta1.GatewayClassList{}
+	if err := k8sClient.List(ctx, allClasses, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(GatewayClass_GatewayClassConfigIndex, config.Name),
```
nathancoleman, line 104: Does this actually work? I checked last week and saw an outstanding proposal for FieldSelector support with CRDs

https://github.com/kubernetes/kubernetes/pull/122717

nathancoleman: I guess this maybe only works here _because of_ the indexers?

control-plane/controllers/resources/gateway_controller_setup.go
```diff
@@ -0,0 +1,141 @@
...
+	if err := k8sClient.List(ctx, allGateways, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(Gateway_GatewayClassIndex, class.Name),
+	}); err != nil {
+		return nil, client.IgnoreNotFound(err)
+	}
+
+	return allGateways, nil
+}
+
+func getGatewayClassConfigForGateway(ctx context.Context, k8sClient client.Client, className string) (*meshv2beta1.GatewayClassConfig, error) {
```
nathancoleman, line 129: Similarly to the above, this doesn't actually see or care about the `Gateway` at all.

Some of these functions accept the full `GC` or `GCC`. Others accept only the class name from the `GC` or `GCC`. Might be good to do one or the other consistently if that makes sense.

nathancoleman (changes requested):
Nice work! Excited for this change to go in

control-plane/controllers/resources/gateway_controller_setup.go
jm96441n: yeah, this is the same pattern we use for v1 (in fact this is pretty much copy pasted from v1), it works because of the indexers

control-plane/controllers/resources/gateway_controller_setup.go
```diff
@@ -0,0 +1,137 @@
...
+	if err := k8sClient.List(ctx, allClasses, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(string(GatewayClass_GatewayClassConfigIndex), configName),
+	}); err != nil {
+		return nil, client.IgnoreNotFound(err)
+	}
+
+	return allClasses, nil
+}
+
+// getGatewaysReferencingGatewayClass queries all MeshGateway resources in the cluster
```
nathancoleman, line 126:
```suggestion
// getGatewaysReferencingGatewayClass queries all xGateway resources in the cluster
```

control-plane/controllers/resources/gateway_controller_setup.go
```diff
@@ -0,0 +1,137 @@
...
+//
+//	if err := k8sClient.Get(ctx, types.NamespacedName{Name: className}, &gatewayClass); err != nil {
+//		return nil, client.IgnoreNotFound(err)
+//	}
+//	return &gatewayClass, nil
+//}
+
+// getGatewayClassesByGatewayClassConfigName queries all GatewayClass resources in the
+// cluster and returns any that reference the given GatewayClassConfig.
+func getGatewayClassesByGatewayClassConfigName(ctx context.Context, k8sClient client.Client, configName string) (*meshv2beta1.GatewayClassList, error) {
```
nathancoleman, line 115: For consistency with the below

```suggestion
// getGatewayClassesReferencingGatewayClassConfig queries all GatewayClass resources in the
// cluster and returns any that reference the given GatewayClassConfig by name.
func getGatewayClassesReferencingGatewayClassConfig(ctx context.Context, k8sClient client.Client, configName string) (*meshv2beta1.GatewayClassList, error) {
```

control-plane/controllers/resources/gateway_controller_setup.go
```diff
@@ -0,0 +1,137 @@
...
+		FieldSelector: fields.OneTermEqualSelector(string(GatewayClass_GatewayClassConfigIndex), configName),
+	}); err != nil {
+		return nil, client.IgnoreNotFound(err)
+	}
+
+	return allClasses, nil
+}
+
+// getGatewaysReferencingGatewayClass queries all MeshGateway resources in the cluster
+// and returns any that reference the given GatewayClass.
```
nathancoleman, line 127:
```suggestion
// and returns any that reference the given GatewayClass by name.
```

control-plane/controllers/resources/gateway_indices.go
```diff
@@ -0,0 +1,75 @@
...
+	GatewayClass_GatewayClassConfigIndex indexName = "__v2_gatewayclass_referencing_gatewayclassconfig"
+
+	APIGateway_GatewayClassIndex  indexName = "__v2_api_gateway_referencing_gatewayclass"
+	MeshGateway_GatewayClassIndex indexName = "__v2_mesh_gateway_referencing_gatewayclass"
+)
+
+// RegisterFieldIndexes registers all of the field indexes for the API gateway controllers.
+// These indexes are similar to indexes used in databases to speed up queries.
+// They allow us to quickly find objects based on a field value.
+func RegisterFieldIndexes(ctx context.Context, mgr ctrl.Manager) error {
```
nathancoleman, line 29:
```suggestion
// RegisterGatewayFieldIndexes registers all of the field indexes for the xGateway controllers.
// These indexes are similar to indexes used in databases to speed up queries.
// They allow us to quickly find objects based on a field value.
func RegisterGatewayFieldIndexes(ctx context.Context, mgr ctrl.Manager) error {
```

nathancoleman (approved):
LGTM with a few more cleanup requests. Is it reasonable to add unit test coverage for this, or is it more sensible at the integration level where the k8s API is actually up and running?

jm96441n:
so we have unit tests covering the MeshGateway controller, being as this affects the setup entirely I think it's better covered by the integration tests since it's really just setting up watches for k8s

control-plane/controllers/resources/gateway_controller_setup.go
jm96441n: https://pkg.go.dev/sigs.k8s.io/controller-runtime#hdr-Clients_and_Caches this details a bit that field indexers can be used to look up objects

PR #3531, opened 2024-02-01
[NET-6465] Respect connectInject.initContainer.resources for v1 API gateways

nathancoleman:
### Changes proposed in this PR ###  
-
-

### Checklist ###
- [ ] Tests added
- [ ] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #3575, opened 2024-02-07
[NET-6465] Consider init container resources when determining if existing + desired deployments are equal

nathancoleman:
> [!NOTE]
> This is a followup to #3531 

When there's an existing `Deployment` in the cluster and we're comparing it against our desired state to determine if there's an update to be made, the comparison needs to consider whether the resources for the init container(s) have been updated.

1. Create a `Gateway` on a cluster having `helm install`-ed with or without a value for `.Values.connectInject.initContainer.resources`.
2. Change the value of `.Values.connectInject.initContainer.resources` and `helm upgrade`
3. Verify that cycling the gateway `Pod` triggers an update to the resources of the init container to match your values. This cycling requirement is consistent with how regular service Deployments in Consul mesh behave due to needing some action to trigger the reconciliation process.

See above

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/api-gateway/gatekeeper/deployment_test.go
+++ b/control-plane/api-gateway/gatekeeper/deployment_test.go
@@ -1,0 +1,216 @@
+package gatekeeper
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/hashicorp/consul-k8s/control-plane/api-gateway/common"
+)
+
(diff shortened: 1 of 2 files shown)


NiniOak (approved):

control-plane/api-gateway/gatekeeper/deployment_test.go
```diff
@@ -0,0 +1,216 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+			shouldBeEqual: false,
+		},
+		{
+			name: "different init container resources",
```
nathancoleman, line 94:
```suggestion
			name: "same init container resources",
```

PR #3580, opened 2024-02-07
[NET-7450] setup crud hooks for APIGateway v2

jm96441n:
- modifies the api-gateway v2 controller to utilize the generic crud hooks
-

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/controllers/resources/api-gateway-controller.go
+++ b/control-plane/controllers/resources/api-gateway-controller.go
@@ -14,14 +14,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	meshv2beta1 "github.com/hashicorp/consul-k8s/control-plane/api/mesh/v2beta1"
+	"github.com/hashicorp/consul-k8s/control-plane/gateways"
 )
 
 // APIGatewayController reconciles a APIGateway object.
 type APIGatewayController struct {
 	client.Client
-	Log        logr.Logger
-	Scheme     *runtime.Scheme
-	Controller *ConsulResourceController
+	Log           logr.Logger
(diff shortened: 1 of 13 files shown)


control-plane/controllers/resources/api-gateway-controller.go
```diff
@@ -14,14 +14,16 @@ import (
...
 // APIGatewayController reconciles a APIGateway object.
 type APIGatewayController struct {
 	client.Client
-	Log        logr.Logger
-	Scheme     *runtime.Scheme
-	Controller *ConsulResourceController
+	Log           logr.Logger
+	Scheme        *runtime.Scheme
+	Controller    *ConsulResourceController
+	GatewayConfig gateways.GatewayConfig
```
andrewstucki, line 26: Just to be clear on this -- is this derived from the global helm configuration? The idea of `GatewayConfig` just seems kind of close to `GatewayClassConfig` to me, so the naming is slightly confusing.

andrewstucki (approved):
Generally seems fine to me, though I'm sure we may need to iterate some. I do think we may want to rename the `GatewayConfig` struct at some point though, but that can wait.

Only other real question that I have is how we envision handling naming collisions between API Gateways and Mesh if we're creating named deployments under the hood that overlap... do we eventually need some sort of validating webhook to ensure you can't have colliding names? But that can probably wait too.

control-plane/controllers/resources/api-gateway-controller.go
jm96441n: yeah, this is an existing struct we have that takes the gateway specific stuff from the helm chart, I'm down for renaming it as it is very similar to `GatewayClassConfig`

PR #3615, opened 2024-02-12
[DONOTMERGE] tmp: run acceptance tests targeting Envoy 1.26.7 via custom Dataplane

zalimeni:
To vet the Consul 1.15 / consul-k8s 1.1 upgrade from Envoy 1.25 to 1.26.7, run acceptance tests using these new versions.

### Checklist ###
- [ ] Tests added


zalimeni:
Note that we should also be getting this now on `release/1.1.x` builds of `consul-k8s`; this forces a version update just to make sure we've run against latest since we use a floating tag.

zalimeni:
Passed, closing PR

PR #3625, opened 2024-02-13
[NET-2420] security: Upgrade helm containerd and several other dependencies

zalimeni:
This upgrades several dependencies to address open CVEs in the `cli` module (caused by our consumption of 3rd-party libs for API clients that are bundled w/ unused runtimes, where a non-impacting vulnerability actually exists) and cross-compatibility between upgraded modules.

**This will allow us to re-enable release blocks on Go module scan results in CRT**, previously disabled in:
* https://github.com/hashicorp/consul-k8s/pull/1974
* https://github.com/hashicorp/consul-k8s/pull/1160

Upgrade the following (split by commit to aid review):
- `helm/v3` 3.9.4 -> 3.11.3
  - Later versions have introduced breaking changes. I believe we will eventually want to match our helm version to K8s per [this](https://helm.sh/docs/topics/version_skew/) matrix, but the current goal is to address open CVEs, not correct the version alignment.
- `k8s.io/` dependencies 1.26.? -> 1.26.10 to match `controller-runtime` / `helm`
- `containerd` 1.3.0 -> 1.7.13 (latest)
- `docker/docker` 23.0.3+incompatible -> 25.0.2+incompatible (latest)
- `oras.land/oras-go` 1.2.2 -> 1.2.5 to fix docker incompatibility
- `docker/distribution` 2.8.1+incompatible -> 2.8.3+incompatible (latest)
- `cyphar/filepath-securejoin` 0.2.3 -> 0.2.4 (latest 0.x patch)

Automated backports will almost certainly fail, but I've added the labels JIC they're successful and to make it clear this should be backported.

CI and acceptance tests continue to pass.

👀 + 💭 

### Checklist ###
- [ ] Tests added


cli/go.mod
```diff
@@ -13,61 +13,60 @@ require (
...
 	k8s.io/apimachinery v0.26.12
-	k8s.io/cli-runtime v0.24.3
+	k8s.io/cli-runtime v0.26.10
 	k8s.io/client-go v0.26.12
-	k8s.io/utils v0.0.0-20221107191617-1a15be271d1d
+	k8s.io/utils v0.0.0-20230220204549-a5ecb0141aa5
 	sigs.k8s.io/yaml v1.3.0
 )
 
-require go.opentelemetry.io/proto/otlp v0.19.0 // indirect
```
zalimeni, line 31: Moved below into multiline block

cli/go.mod
```diff
@@ -13,61 +13,60 @@ require (
...
+	github.com/emicklei/go-restful/v3 v3.10.1 // indirect
 	github.com/envoyproxy/go-control-plane v0.11.1 // indirect
 	github.com/envoyproxy/go-control-plane/xdsmatcher v0.0.0-20230524161521-aaaacbfbe53e // indirect
-	github.com/envoyproxy/protoc-gen-validate v1.0.1 // indirect
-	github.com/evanphx/json-patch v4.12.0+incompatible // indirect
+	github.com/envoyproxy/protoc-gen-validate v1.0.2 // indirect
+	github.com/evanphx/json-patch v5.6.0+incompatible // indirect
 	github.com/exponent-io/jsonpath v0.0.0-20151013193312-d6023ce2651d // indirect
-	github.com/fsnotify/fsnotify v1.5.4 // indirect
+	github.com/felixge/httpsnoop v1.0.3 // indirect
```
zalimeni, line 66: Brought in by helm dependencies, it's an http [metrics library](https://github.com/felixge/httpsnoop). Checked bc of the name 😄

wilkermichael:
I would just wait for all the acceptance tests to pass before merging

zalimeni:
> I would just wait for all the acceptance tests to pass before merging

Fully agree. Thanks for the review @wilkermichael !

DanStough (approved):
LGTM 👍

PR #3693, opened 2024-02-28
Catalog: Use EndpointSlice and propagate Kubernetes Topology information to synced consul service

jukie:
- Switch to EndpointSlice vs Endpoints
  - Necessary for accessing endpoint zone information but it's also the preferred API to use. Ref: https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#motivation
- Adds topology zone from Endpoint to Cosul service meta
- Adds topology region from Endpoint's associated Node to Consul service meta
- Adds get, list, and watch permissions for EndpointSlice to sync catalog's ClusterRole
- Removes get, list, and watch permissions for Endpoints from sync catalog's ClusterRole

- Existing and new tests
- Built and deployed catalog service to verify

- Enable service sync and validate the registered instances have a listed meta entry for `external-k8s-topology-zone` and `external-k8s-topology-region`

### Checklist ###
- [x] Tests added


jukie:
@david-yu Similar end result yes, but that PR specifically targets NodePort services and uses underlying node labels. My change doesn't currently include pulling in region information as that's not directly available on EndpointSlice.

jukie:
My PR now includes region info for NodePort services as well so encompasses everything from the mentioned PR plus the addition of zone info for all other service types

david-yu:
Thank you, I will try to have someone review sometime later next week.

jukie:
@david-yu took it further to include all service types now

jmurret:
👋 @jukie thank you so much for this work!  It looks great.  I can see that the some of the unit tests in `catalog` package in `control-plane` are failing locally and in CI (you may not have access to see CI runs).  Do these pass for you?  If you have any time to resolve them, that would be a great help.  Otherwise, I can try to take a look at the end of the week.  🙏 
Screenshot 2024-03-06 at 9 57 54 AM

jukie:
Found the problem, had to make sure the nodes referenced by the fake endpoints in the test also exist.

jmurret:
Updated somebroken helm chart tests andalso fixed logic in sync-catalog-clusterrole.  TestSyncCatalogNamespaces and TestSyncCatalogNamespaces acceptance tests are currently failing.  will try to take a look tomorrow.

jmurret:
If you run without the clusterrole permission and deploy a service to verify that catalog sync does in fact sync, you will it does not and in the log it says:
```
2024-03-07T23:32:54.070Z [DEBUG] to-k8s/sink: sync triggered: create=0 update=0 delete=0
W0307 23:32:55.647788       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.12/tools/cache/reflector.go:169: failed to list *v1.Endpoints: endpoints is forbidden: User "system:serviceaccount:default:test-i1asnm-consul-sync-catalog" cannot list resource "endpoints" in API group "" at the cluster scope
```

jukie:
Let me double check this but I deployed a version of this prior to the commit adding support for region topology and endpoint permissions are no longer needed. There shouldn't be any piece of code left that still need to interact with Endpoints via the k8s api's.

jukie:
@jmurret Would it be possible to also enable the automatic pr tests as well?

jmurret:
hey @jukie 
a couple of things:
- I think i mixed up the image i was using locally running acceptance tests.  CI does not keep the k8s logs and only logs the tests.  so I was seeing a failure and then locally seeing a different error.  you are right, that clusterrolle permission is not needed.  sorrya bout the confusion
- for CI, we have this complex set up for this repo where it is run in a private github actions repo that is meant to make protecting secrets used to run CI more secure but it causes headaches when community members / non-hashicorp org members create PRs.  I had to copy your commits to another branch and open a draft PR so CI would run.  That PR is here: https://github.com/hashicorp/consul-k8s/pull/3716
- lastly, i spent quite a bit of time debugging the acceptance test failures and found that endpointslices get reconciled asynchronously in a different way than endpoints.  an initial endpoint slice is created with nil values for `Endpoints` and `Ports`.  A race condition occurs where the catalog sync is only listening for service changes, so when it gets a Service change watch notification it queries for endpoint slices.  If the `Endpoints` and `Ports` have not been updated on a given endpoint slice at that time, then this current version of catalog sync will not loop through the endpoints to get the address and add the item to the `consulMap` to be synced with Consul.  Below is the output from the watch of endpoints via `kubectl get endpointslices -n sync -o json -w` while the `TestSyncCatalogNamespaces` test was running.  You can see there are two different entries returned from the watch where it changed states by adding the `Endpoints` and `Ports` values.
```
$ kubectl get endpointslices -n sync -o json -w
{
    "addressType": "IPv4",
    "apiVersion": "discovery.k8s.io/v1",
    "endpoints": null,
    "kind": "EndpointSlice",
    "metadata": {
        "annotations": {
            "endpoints.kubernetes.io/last-change-trigger-time": "2024-03-08T03:51:26Z"
        },
        "creationTimestamp": "2024-03-08T03:51:26Z",
        "generateName": "static-server-",
        "generation": 1,
        "labels": {
            "endpointslice.kubernetes.io/managed-by": "endpointslice-controller.k8s.io",
            "kubernetes.io/service-name": "static-server"
        },
        "name": "static-server-xpz59",
        "namespace": "sync",
        "ownerReferences": [
            {
                "apiVersion": "v1",
                "blockOwnerDeletion": true,
                "controller": true,
                "kind": "Service",
                "name": "static-server",
                "uid": "ccdfcc87-4349-4129-83f9-51f6c340c7b8"
            }
        ],
        "resourceVersion": "920",
        "uid": "67b423f6-68ff-4c2d-9744-ee227d074d69"
    },
    "ports": null
}
{
    "addressType": "IPv4",
    "apiVersion": "discovery.k8s.io/v1",
    "endpoints": [
        {
            "addresses": [
                "10.244.0.11"
            ],
            "conditions": {
                "ready": true,
                "serving": true,
                "terminating": false
            },
            "nodeName": "dc1-control-plane",
            "targetRef": {
                "kind": "Pod",
                "name": "static-server-548bdffcb8-4tsp5",
                "namespace": "sync",
                "uid": "7f1100ea-9c6a-437e-bc9b-bfd4d811f088"
            }
        }
    ],
    "kind": "EndpointSlice",
    "metadata": {
        "annotations": {
            "endpoints.kubernetes.io/last-change-trigger-time": "2024-03-08T03:51:30Z"
        },
        "creationTimestamp": "2024-03-08T03:51:26Z",
        "generateName": "static-server-",
        "generation": 2,
        "labels": {
            "endpointslice.kubernetes.io/managed-by": "endpointslice-controller.k8s.io",
            "kubernetes.io/service-name": "static-server"
        },
        "name": "static-server-xpz59",
        "namespace": "sync",
        "ownerReferences": [
            {
                "apiVersion": "v1",
                "blockOwnerDeletion": true,
                "controller": true,
                "kind": "Service",
                "name": "static-server",
                "uid": "ccdfcc87-4349-4129-83f9-51f6c340c7b8"
            }
        ],
        "resourceVersion": "950",
        "uid": "67b423f6-68ff-4c2d-9744-ee227d074d69"
    },
    "ports": [
        {
            "name": "http",
            "port": 8080,
            "protocol": "TCP"
        }
    ]
}
```

I think quite a bit more logic is going to be required to add watches to sync-catalog for EndpointSlices in addition to watching services.  I am assuming that previously this call would error if endpoints were not ready and the event would re-queue to be processed again until the endpoints were ready.  I think this is more complicated to try to do with just processing Service records and trying to detected missing endpoints when we now have an array of EndpointSlices that each contain an array of endpoints.  I would assume that each endpoint slice gets reconciled by k8s separately, so trying to detect errors in this same place would require looping through all endpoint slices looking for missing endpoints and ports and then failing the whole service to be requeued.   This makes me think that catalog sync needs to watch endpoint slices in addition to services instead so that it can be notified rather than requeueing and querying (especially when there is also already a requeue max limit that would need to be changed to accomodate more failures when a service has many endpointslices that could possibly not all be ready in several queueings of the original k8s Service change).
```
endpoints, err := t.Client.CoreV1().
			Endpoints(service.Namespace).
			Get(t.Ctx, service.Name, metav1.GetOptions{})
		if err != nil {
			t.Log.Warn("error loading initial endpoints",
				"key", key,
				"err", err)
		}
``` 

I am going to pause and see what thoughts you have?

jmurret:
Also, if the main motivation is to get region and zone topology synced to consul services, we could move forward with https://github.com/hashicorp/consul-k8s/pull/3522 and forgo the EndpointSlices stuff for a different time.

jukie:
> due to the catalog sync is only listening for service changes
Could you also help me understand this more? I may be misreading things but it looks like catalog starts 3 background informers to watch [Services](https://github.com/hashicorp/consul-k8s/blob/5dcb4537b3b68ae00d105bef6d316a1fb5c53560/control-plane/catalog/to-consul/resource.go#L181), [Endpoints](https://github.com/hashicorp/consul-k8s/blob/5dcb4537b3b68ae00d105bef6d316a1fb5c53560/control-plane/catalog/to-consul/resource.go#L851) (EndpointSlices with my change), and [Ingresses](https://github.com/hashicorp/consul-k8s/blob/5dcb4537b3b68ae00d105bef6d316a1fb5c53560/control-plane/catalog/to-consul/resource.go#L935). There is certainly a race condition upon initial Service creation where an empty EndpointSlice resource is created but I'd think an event from the EndpointSlice informer would resolve that shortly after.

jukie:
Alright I got it working. The `key` names and my logic for collecting slices was causing issues but will be pushing something with soon with the fixes.

jukie:
@jmurret this should be good now, acceptance tests as well this time.
* My setup of `endpointSlicesMap` was incorrect
* I also needed to account for the differences with `Service` naming vs `EndpointSlice` naming which includes an extra UUID suffix and was causing issues.

jukie:
Could you please share the cli flags for the tproxy tests? Running locally with `-enable-transparent-proxy` things are working.

1. Build image and run kind-load
2. `go test ./... -p 1 -timeout 2h -failfast -use-kind -kube-contexts=kind-dc2  -debug-directory=/tmp/debug/all  -kubeconfigs=/Users/isaac.wilson/.kube/ttd -consul-k8s-image=consul-k8s-control-plane:local -enable-transparent-proxy=true` is working with my image locally.

control-plane/catalog/to-consul/resource.go
```diff
@@ -143,9 +146,11 @@ type ServiceResource struct {
...
 	serviceMap map[string]*corev1.Service
 
-	// endpointsMap uses the same keys as serviceMap but maps to the endpoints
-	// of each service.
-	endpointsMap map[string]*corev1.Endpoints
+	// endpointSlicesMap tracks EndpointSlices associated with services that are being synced to Consul.
+	// The outer map's keys represent service identifiers in the same format as serviceMap and maps
+	// each service to its related EndpointSlices. The inner map's keys are EndpointSlice name keys
+	// the format "<kube namespace>/<kube endpointslice name>".
+	endpointSlicesMap map[string]map[string]discoveryv1.EndpointSlice
```
jukie, line 153: I can switch to `map[string]map[string]*discoveryv1.EndpointSlice` instead here if preferred

jmurret:
@jukie thank you!  this is awesome work!  We appreciate all of your efforts.  
The failing test was on api-gateway at this line cleaning up resources: https://github.com/hashicorp/consul-k8s/blob/main/acceptance/tests/api-gateway/api_gateway_external_servers_test.go#L85
I've rerun failed jobs and we will see if this clears up as just a flake.

Related to the region information...I have not had much time to give it much thought or look into it but will have likely have time at the end of the week.

jukie:
@jmurret Could you share the cli flags for that one? If it does fail again I'd like to debug it more.

jmurret:
It passed.  It was called with this below.  basically packages are split into groupings to go on runners and then are used to set the PACKAGES envvar here.  So, all packages will have their tests run this way.
```
PACKAGES="api-gateway ingress-gateway sync example consul-dns"
  echo "Running packages: ${PACKAGES}"
  for PKG in ${PACKAGES}
  do
    FULLPKG="github.com/hashicorp/consul-k8s/acceptance/tests/${PKG}"
    echo "Testing package: ${FULLPKG}"
    go run gotest.tools/gotestsum@v1.11.0 \
      --format=github-actions \
      --jsonfile="jsonfile-${PKG////-}" \
      --packages="${PKG}" \
      --junitfile="/tmp/test-results/acceptance/${PKG}-gotestsum-report.xml" \
      -- "${FULLPKG}" -p 1 -timeout 2h -failfast \
      -use-kind \
      -kube-contexts="kind-dc1,kind-dc2,kind-dc[3](https://github.com/hashicorp/consul-k8s-workflows/actions/runs/8327109167/job/22784276823#step:32:3),kind-dc[4](https://github.com/hashicorp/consul-k8s-workflows/actions/runs/8327109167/job/22784276823#step:32:4)" \
      -enable-enterprise \
      -enable-multi-cluster \
      -debug-directory=/tmp/test-results/debug \
      -consul-image=docker.mirror.hashicorp.services/hashicorppreview/consul-enterprise:1.19-dev \
      -consul-k8s-image=docker.io/hashicorppreview/consul-k8s-control-plane:1.[5](https://github.com/hashicorp/consul-k8s-workflows/actions/runs/8327109167/job/22784276823#step:32:5).0-dev-pr-5606e615d94262ac878a0b3be374b7f19565256a \
      -consul-dataplane-image=docker.mirror.hashicorp.services/hashicorppreview/consul-dataplane:1.5-dev \
      -enable-enterprise -enable-multi-cluster -enable-transparent-proxy
  done
```

jukie:
@kolorful did you have a use case for region or would topology zone be good enough?

jukie:
Main difference is the zone information being available on the endpoints directly. It's probably not a big deal but was curious if there was any concern over the amount of extra api server calls to get each node. That already happens for NodePort type services so again, probably not a big deal to expand that to all service types.

jmurret:
@jukie thanks for raising your scalability concerns about also grabbing node information while processing endpoint slices.  I think we would want a way to turn that off via configuration/helm chart since we don't have load tests asserting its performance.  I propose we remove that from this PR since there is no current requirement for Region and when it is needed we can add the appropriate helm configuration.  SOund ok?

This PR looks great.  I think the two things needed to approve and merge are:
- remove the Region/node lookups
- make the endpointslice pointer change that you mentioned.

Thanks again for your efforts on this!

control-plane/catalog/to-consul/resource.go
jmurret: please do.  🙏

jukie:
Done. I've also taken another stab at dropping endpoints permission from the catalog cluster role because I really don't think we need it but you can add that back in again if preferred.

jukie:
@jmurret could please kick off the acceptance tests again?

kolorful:
@jukie I see, thanks for explaining. I think not having region info is fine, because usually it's part of the zone info anyway.

jmurret (approved):
🚀 🙏 thank you @jukie for sticking with this and getting it done.  We appreciate it!

jukie:
Awesome! Will this automatically land on the -dev release for the image and helm charts?

jmurret:
@jukie there is no -dev channel/artifact for dev releases.  you'll have to use the helm chart on main until the next 1.4 patch release.  the values.yaml has a link to an internal dev preview that I think you should have access to: https://github.com/hashicorp/consul-k8s/blob/5279204558f76229a8d0dce98e99ed986407b3a6/charts/consul/values.yaml#L69

PR #3700, opened 2024-02-29
release: add \n to end of NOTE for releases 

david-yu:
- NOTE is clobbering BREAKING changes in release notes

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/build-support/functions/10-util.sh
+++ b/control-plane/build-support/functions/10-util.sh
@@ -761,7 +761,7 @@ function set_changelog {
         fi
         compatibility_note="
 
-> NOTE: Consul K8s ${version_short}.x is compatible with Consul ${consul_version_short}.x and Consul Dataplane ${consul_dataplane_version_short}.x. Refer to our [compatibility matrix](https://developer.hashicorp.com/consul/docs/k8s/compatibility) for more info."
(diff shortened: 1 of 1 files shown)


control-plane/build-support/functions/10-util.sh
```diff
@@ -761,7 +761,7 @@ function set_changelog {
         fi
         compatibility_note="
 
-> NOTE: Consul K8s ${version_short}.x is compatible with Consul ${consul_version_short}.x and Consul Dataplane ${consul_dataplane_version_short}.x. Refer to our [compatibility matrix](https://developer.hashicorp.com/consul/docs/k8s/compatibility) for more info."
+> NOTE: Consul K8s ${version_short}.x is compatible with Consul ${consul_version_short}.x and Consul Dataplane ${consul_dataplane_version_short}.x. Refer to our [compatibility matrix](https://developer.hashicorp.com/consul/docs/k8s/compatibility) for more info. \n"
```
zalimeni, line 764: I think we actually need a true line break here. Fortunately this is a multiline string, so it should just work w/ a real newline (tested and seems good)
```suggestion
> NOTE: Consul K8s ${version_short}.x is compatible with Consul ${consul_version_short}.x and Consul Dataplane ${consul_dataplane_version_short}.x. Refer to our [compatibility matrix](https://developer.hashicorp.com/consul/docs/k8s/compatibility) for more info.
"
```

zalimeni (approved):
Should be ok skipping backport since this only impacts `.0` releases.

Approved to unblock after fix above - thank you!

PR #3705, opened 2024-02-29
[NET-8174] security: add triage alias for GO-2024-2554

zalimeni:
This vulnerability was already triaged [via its GHSA alias](https://osv.dev/vulnerability/GO-2024-2554), but the scanner is flagging it under this name, so adding an explicit entry.

Scanner now passes locally.

👀 

cc @picatz @dduzgun-security for awareness since I haven't seen aliases re-emerge weeks later before - maybe out of our control?


DanStough (approved):
👍

PR #3717, opened 2024-03-07
test upgraded library

NiniOak:
-
-

### Checklist ###
- [ ] Tests added


acceptance/go.mod
```diff
@@ -1,10 +1,19 @@
...
-go 1.20
+go 1.21.1
+
+toolchain go1.22.0
+
+replace (
+	k8s.io/api => k8s.io/api v0.26.4
+	k8s.io/apimachinery => k8s.io/apimachinery v0.26.4
+	k8s.io/client-go => k8s.io/client-go v0.26.4
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.14.6
```
jm96441n, line 11: is there a reason we need to have the replace statements?

acceptance/framework/helpers/helpers.go
```diff
@@ -185,3 +187,19 @@ func RegisterExternalService(t *testing.T, consulClient *api.Client, namespace,
...
+	Args       []string          // The args to pass to the command
+	WorkingDir string            // The working directory
+	Env        map[string]string // Additional environment variables to set
+	Logger     *terratestLogger.Logger
+}
+
+func RunCommand(r *retry.R, command Command) (string, error) {
+	r.Helper()
+	// // this command needs a brief delay to avoid flakiness
+	// time.Sleep(500 * time.Millisecond)
```
jm96441n, line 202: should this still be here?

NiniOak: Oh yh, taking this out

acceptance/go.mod
NiniOak: Yes, by upgrading this library `github.com/gruntwork-io/terratest v0.31.2` -> `v0.46.7`, the other dependencies were upgraded to version `k8s.io/apimachinery v0.28.4` resulting in this known [error](https://github.com/operator-framework/operator-sdk/issues/6396)
```
# sigs.k8s.io/controller-runtime/pkg/cache
../../../go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.7/pkg/cache/multi_namespace_cache.go:308:9: cannot use handles (variable of type map[string]"k8s.io/client-go/tools/cache".ResourceEventHandlerRegistration) as "k8s.io/client-go/tools/cache".ResourceEventHandlerRegistration value in return statement: 
map[string]"k8s.io/client-go/tools/cache".ResourceEventHandlerRegistration does not implement "k8s.io/client-go/tools/cache".ResourceEventHandlerRegistration (missing method HasSynced)
FAIL	github.com/hashicorp/consul-k8s/acceptance/tests/api-gateway [build failed]
```
 
The only [solution](https://github.com/operator-framework/operator-sdk/issues/6396) currently is to downgrade the versions of the dependencies OR not upgrade `github.com/gruntwork-io/terratest` library. So i picked the latter

missylbytes (approved):

NiniOak:
> I think this looks good, but do we want to backport at all? Might be worth it for release time on older versions.

Good call, I'll add the backport tags now

PR #3718, opened 2024-03-07
Removes Legacy API Gateway Stanza that was deprecated in Consul 1.16

missylbytes:
- This removes the legacy API Gateway stanza from the Helm chart, as well as the templates associated with it. This gateway was deprecated in Consul 1.16 (consul-k8s 1.2.0) and replaced by `.connectInject.api-gateway`

- Cannot set .Values.apiGateway.enabled anymore
- Looked for all usages of these values
- Checked that our docs do not show the usage of .Values.apiGateway, and instead use `.connectInject.api-gateway`

Try to Helm install with 
`helm upgrade --install consul ./charts/consul -f ./local-values.yaml`
```
# local-values.yaml
global:
  name: consul

apiGateway:
  enabled: true
```
It should error out since the stanza does not exist anymore.

### Checklist ###
- [x] Tests added
- [x] Update this page: https://developer.hashicorp.com/consul/docs/upgrading/upgrade-specific
  - [x] https://github.com/hashicorp/consul/pull/20844


missylbytes:
Adding Do not merge to be sure we don't miss anything.

.changelog/3718.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:deprecation
+api-gateway: The api-gateway stanza located under .Values.api-gateway is deprecated and being removed in favor of
```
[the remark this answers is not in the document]
missylbytes: Yes will add that.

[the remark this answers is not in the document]
missylbytes: Updated

charts/consul/templates/gateway-resources-job.yaml
```diff
@@ -1,3 +1,4 @@
+{{ template "consul.validate_deprecated_apigateway_not_set" . }}
```
missylbytes, line 1: I think this is the best place to put this?

charts/consul/templates/server-acl-init-job.yaml
```diff
@@ -304,10 +304,6 @@ spec:
             -partition-token-file=/vault/secrets/partition-token \
             {{- end }}
 
-            {{- if .Values.apiGateway.enabled }}
-            -api-gateway-controller=true \
```
sarahalsmiller, line 308: Did we not have any spots in the actual job binary that were using stanza or was it all in the templating?

charts/consul/templates/_helpers.tpl
```diff
@@ -685,4 +681,17 @@ Usage: {{ template "consul.versionInfo" }}
...
 {{- end -}}
 {{- printf "%s" $sanitizedVersion | quote }}
+{{- end -}}
+
+{{/*
+Validate that the depraced stanze for .Values.apiGateway is not set.
+Cleanup server.extraConfig entries to avoid conflicting entries:
+    - server.enableAgentDebug:
+      - `enable_debug` should not exist in extraConfig
+Usage: {{ template "consul.validateExtraConfig" . }}
```
lkysow, line 691: I don't think you need to use a helper template, we do an `if blah, fail` directly in a lot of the manifests already so you could just have:

```
{{- if .Values.apiGateway}}{{fail "[DEPRECATED and REMOVED]  apiGateway stanza is no longer supported as of Consul 1.19.0. Use connectInject.apiGateway instead."}}{{ end }}
```

at the top of the gateway-jobs yaml.

Also, I think say "apiGateway stanza" instead of ".Values.apiGateway" because users won't know what .Values is

charts/consul/templates/gateway-resources-job.yaml
lkysow: I'd say here or in connect-inject-deployment where we moved the stanza to.

lkysow (approved):
Nice work!

PR #3736, opened 2024-03-12
fix(control-plane): acl tokens deleted while pods in graceful shutdown

DanStough:
- Users have reported that when GracefulShutdown is enabled, ACL tokens are deleted immediately when a pod goes in to TERMINATING, which can disrupt consul DNS and an app's shutdown procedure if that is required.
- The new behavior is that if a pod still exists and has GracefulShutdown enabled, we will skip ACL and node cleanup, instead re-queue the event for the longest shutdown period, plus some buffer to cut down on noise.

- Manual testing in a local kube cluster, deploying an app w/ ACLs and graceful shutdown. I delete a pod and watch the control-plane and consul-dataplane logs to verify when the ACL token is cleaned up.
- Additional unit tests added.

👀 ☕ 

### Checklist ###
- [X] Tests added

### DIFF
--- a/.changelog/3736.txt
+++ b/.changelog/3736.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+control-plane: fix an issue where ACL token cleanup did not respect a pod's GracefulShutdownPeriodSeconds and
+tokens were invalidated immediately on pod entering Terminating state.
+```
--- a/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
@@ -6,10 +6,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net"
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	mapset "github.com/deckarep/golang-set"
 	"github.com/go-logr/logr"
@@ -25,6 +27,7 @@ import (
 
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/common"
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/lifecycle"
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/metrics"
 	"github.com/hashicorp/consul-k8s/control-plane/consul"
 	"github.com/hashicorp/consul-k8s/control-plane/helper/parsetags"
@@ -94,6 +97,8 @@ type Controller struct {
 	// any created Consul namespaces to allow cross namespace service discovery.
 	// Only necessary if ACLs are enabled.
 	CrossNSACLPolicy string
+	// Lifecycle config set graceful startup/shutdown defaults for pods.
+	LifecycleConfig lifecycle.Config
 	// ReleaseName is the Consul Helm installation release.
 	ReleaseName string
 	// ReleaseNamespace is the namespace where Consul is installed.
@@ -153,18 +158,14 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	err = r.Client.Get(ctx, req.NamespacedName, &serviceEndpoints)
-	// endpointPods holds a set of all pods this endpoints object is currently pointing to.
-	// We use this later when we reconcile ACL tokens to decide whether an ACL token in Consul
-	// is for a pod that no longer exists.
-	endpointPods := mapset.NewSet()
 
 	// If the endpoints object has been deleted (and we get an IsNotFound
 	// error), we need to deregister all instances in Consul for that service.
 	if k8serrors.IsNotFound(err) {
 		// Deregister all instances in Consul for this service. The function deregisterService handles
 		// the case where the Consul service name is different from the Kubernetes service name.
-		err = r.deregisterService(apiClient, req.Name, req.Namespace, nil)
-		return ctrl.Result{}, err
+		requeueAfter, err := r.deregisterService(ctx, apiClient, req.Name, req.Namespace, nil)
+		return ctrl.Result{RequeueAfter: requeueAfter}, err
 	} else if err != nil {
 		r.Log.Error(err, "failed to get Endpoints", "name", req.Name, "ns", req.Namespace)
 		return ctrl.Result{}, err
@@ -177,8 +178,8 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	if isLabeledIgnore(serviceEndpoints.Labels) {
 		// We always deregister the service to handle the case where a user has registered the service, then added the label later.
 		r.Log.Info("ignoring endpoint labeled with `consul.hashicorp.com/service-ignore: \"true\"`", "name", req.Name, "namespace", req.Namespace)
-		err = r.deregisterService(apiClient, req.Name, req.Namespace, nil)
-		return ctrl.Result{}, err
+		requeueAfter, err := r.deregisterService(ctx, apiClient, req.Name, req.Namespace, nil)
+		return ctrl.Result{RequeueAfter: requeueAfter}, err
 	}
 
 	// endpointAddressMap stores every IP that corresponds to a Pod in the Endpoints object. It is used to compare
@@ -213,12 +214,13 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				}
 
 				if hasBeenInjected(pod) {
-					endpointPods.Add(address.TargetRef.Name)
 					if isConsulDataplaneSupported(pod) {
-						if err = r.registerServicesAndHealthCheck(apiClient, pod, serviceEndpoints, healthStatus, endpointAddressMap); err != nil {
+						if err = r.registerServicesAndHealthCheck(apiClient, pod, serviceEndpoints, healthStatus); err != nil {
 							r.Log.Error(err, "failed to register services or health check", "name", serviceEndpoints.Name, "ns", serviceEndpoints.Namespace)
 							errs = multierror.Append(errs, err)
 						}
+						// Build the endpointAddressMap up for deregistering service instances later.
+						endpointAddressMap[pod.Status.PodIP] = true
 					} else {
 						r.Log.Info("detected an update to pre-consul-dataplane service", "name", serviceEndpoints.Name, "ns", serviceEndpoints.Namespace)
 						nodeAgentClientCfg, err := r.consulClientCfgForNodeAgent(apiClient, pod, serverState)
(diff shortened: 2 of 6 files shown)


control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -999,8 +1008,87 @@ func (r *Controller) deregisterService(apiClient *api.Client, k8sSvcName, k8sSvc
...
+			"id", svc.ServiceID, "pod", podName, "k8sNamespace", pod.Namespace)
+		_, err = apiClient.Catalog().Register(serviceRegistration, nil)
+		if err != nil {
+			r.Log.Error(err, "failed to update service health status", "name", svc.ServiceName, "pod", podName)
+			return 0, fmt.Errorf("failed to update service health status for pod %s/%s: %w", pod.Namespace, podName, err)
+		}
+
+		// Return the duration for which the pod should be re-queued. We add 20% to the shutdownSeconds to account for
+		// any potential delay in the pod killed.
+		return time.Duration(shutdownSeconds+int(float64(shutdownSeconds)*0.2)) * time.Second, nil
```
zalimeni, line 1074: ~ Would it make sense to take the `math.Ceil` of `float64(shutdownSeconds)*0.2` instead of casting to `int`? That would guarantee an additional 1s no matter how small the value

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -213,9 +213,11 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				}
 
 				if hasBeenInjected(pod) {
-					endpointPods.Add(address.TargetRef.Name)
 					if isConsulDataplaneSupported(pod) {
-						if err = r.registerServicesAndHealthCheck(apiClient, pod, serviceEndpoints, healthStatus, endpointAddressMap); err != nil {
+						// Build the endpointAddressMap up for deregistering service instances later.
+						endpointAddressMap[pod.Status.PodIP] = true
```
jmurret, line 218: should this be after the registration call so that we know it was successful? (same below in `isGatway(pod)` case)

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -934,40 +933,50 @@ func getHealthCheckStatusReason(healthCheckStatus, podName, podNamespace string)
...
+	k8sSvcName string,
+	k8sSvcNamespace string,
+	endpointsAddressesMap map[string]bool) (time.Duration, error) {
+
 	// Get services matching metadata from Consul
 	serviceInstances, err := r.serviceInstances(apiClient, k8sSvcName, k8sSvcNamespace)
 	if err != nil {
 		r.Log.Error(err, "failed to get service instances", "name", k8sSvcName)
-		return err
+		return 0, err
```
jmurret, line 947: [micro-nit] should we update the method docstring to describe the requeueAfter value?

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -153,18 +157,14 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	err = r.Client.Get(ctx, req.NamespacedName, &serviceEndpoints)
-	// endpointPods holds a set of all pods this endpoints object is currently pointing to.
-	// We use this later when we reconcile ACL tokens to decide whether an ACL token in Consul
-	// is for a pod that no longer exists.
-	endpointPods := mapset.NewSet()
```
zalimeni, line 159: Nice, thanks for removing this unused one

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -999,8 +1008,87 @@ func (r *Controller) deregisterService(apiClient *api.Client, k8sSvcName, k8sSvc
 		}
 	}
 
-	return errs
+	if requeueAfter > 0 {
+		r.Log.Info("re-queueing event", "requeueAfter", requeueAfter)
```
jmurret, line 1012: should we include service and namespace or other meta?  you could grab it from the log line above, but if you were looking at this in a structured log viewer and quering for requeued events, it might be helpful to have that.  Also, it might be helpful to say `re-queueing <event type> event` to help differentiate it from other requeuese we might do during registration.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -999,8 +1008,87 @@ func (r *Controller) deregisterService(apiClient *api.Client, k8sSvcName, k8sSvc
...
+				Type:      constants.ConsulKubernetesCheckType,
+				Status:    api.HealthCritical,
+				ServiceID: svc.ServiceID,
+				Output:    fmt.Sprintf("Pod \"%s/%s\" is terminating", pod.Namespace, podName),
+				Namespace: r.consulNamespace(pod.Namespace),
+			},
+			SkipNodeUpdate: true,
+		}
+
+		r.Log.Info("updating health status of service with Consul", "name", svc.ServiceName,
```
jmurret, line 1064: should we add a few more here to say updating to critical in order to drain traffic?  basically something like the comment you have at the top of this if.

jmurret (approved):
🔥 this is so great.  thank you!  left a few very minor questions.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -999,8 +1008,87 @@ func (r *Controller) deregisterService(apiClient *api.Client, k8sSvcName, k8sSvc
...
+
+	var pod corev1.Pod
+	err := r.Client.Get(ctx, types.NamespacedName{Name: podName, Namespace: k8sNamespace}, &pod)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		r.Log.Error(err, "failed to get terminating pod", "name", podName, "k8sNamespace", k8sNamespace)
+		return 0, fmt.Errorf("failed to get terminating pod %s/%s: %w", k8sNamespace, podName, err)
+	}
+	if k8serrors.IsNotFound(err) {
+		return 0, nil
+	}
```
zalimeni, line 1037: ~ ❓ Could we simplify this to:
```suggestion
	if k8serrors.IsNotFound(err) {
		return 0, nil
	}
	if err != nil {
		r.Log.Error(err, "failed to get terminating pod", "name", podName, "k8sNamespace", k8sNamespace)
		return 0, fmt.Errorf("failed to get terminating pod %s/%s: %w", k8sNamespace, podName, err)
	}
```

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
zalimeni, line 1012: ~ It might help to call out why we're requeueing so operators know it's expected, and due to graceful shutdown

.changelog/3736.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+control-plane: fix an issue where ACL token cleanup did not respect a pod's GracefulShutdownPeriodSeconds and
+were being cleaned up immediately on going in to the Terminating state.
```
zalimeni, line 3:
```suggestion
control-plane: fix an issue where ACL token cleanup did not respect a pod's GracefulShutdownPeriodSeconds and
tokens were invalidated immediately on pod entering Terminating state.
```

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -934,40 +933,50 @@ func getHealthCheckStatusReason(healthCheckStatus, podName, podNamespace string)
...
+				requeueAfter = podShutdownDuration
+			}
+			if podShutdownDuration > 0 {
+				continue
+			}
+
+			// If the service address is not in the Endpoints addresses, deregister it.
 			r.Log.Info("deregistering service from consul", "svc", svc.ServiceID)
-			_, err := apiClient.Catalog().Deregister(&api.CatalogDeregistration{
+			_, err = apiClient.Catalog().Deregister(&api.CatalogDeregistration{
```
zalimeni, line 979: I think this is ok and not a concerning change in behavior, but wanted to spell out my understanding:

- This is within the scope of a single service (endpoints) being reconciled; that means the only time we expect heterogeneity between pods' graceful shutdown timings, is when deployments are being updated or in weirder pets-not-cattle pod treatment
- Even if one pod manages to artificially delay the deregistration of another's service instance, any pod that has finished terminating will be deregistered immediately
- Any pod without graceful shutdown configured is unaffected by this change

zalimeni (approved):
LGTM! Couple minor comments but nothing blocking.

I think it might be worth a final 👍🏻 from someone really familiar w/ K8s requeue + debouncing behavior JIC, unless you feel good on that front (just calling out that I'm not sure I have enough experience there to be 100% confident in this use of requeues).

Thanks for cleaning up the code while you were here as well 🙏🏻

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
DanStough: Yeah, I like that more now that I have separated this from the registration functions.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
DanStough: I think this is a good summary with maybe one clarification: 
> Even if one pod manages to artificially delay the deregistration of another's service instance, any pod that has finished terminating will be deregistered immediately

I don't think in this case it will be deregistered immediately, but:
1. When the pod finishes normally, consul-server-connection-manager should clean up the ACL token immediately
1. Even if the pod is force-killed by k8s, the consul service instance will be `critical` health, so it should not be routable.
1. Any other endpoint reconciliation for that service would cause that instance to be deregistered prior to the `requeueAfter` time. 

Let me know if you see any problems with this approach.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
DanStough: Good idea. I'll add the meta data and clarify this is for graceful shutdown.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
DanStough: Yeah, I like that idea!

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
zalimeni: > Any other endpoint reconciliation for that service would cause that instance to be deregistered prior to the requeueAfter time.

Good clarification. I guess when I said "immediately", I meant "during the next reconcile", which I expect would follow that state transition (though maybe that's actually a behavioral difference between Endpoints and EndpointSlices). Given the other safeguards mentioned, this seems ok to me.

One other question I have about the `critical` health update: is there a possibility that status would flap on a subsequent health check of the service, if it isn't yet self-reporting as unhealthy due to shutdown?

DanStough: > One other question I have about the critical health update: is there a possibility that status would flap on a subsequent health check of the service, if it isn't yet self-reporting as unhealthy due to shutdown?

My thought here is that consul-k8s is the only source of health check information and this is only otherwise set during service registration. Since the pod should be removed from the endpoints object indefinitely, I would expect that we would never get back to a registration event that would toggle it. At least that's what I observed doing testing.

PR #3758, opened 2024-03-15
[net-8411] bug: fix premature token and service instance deletion due to pod fetch errors

ndhanushkodi:
- When a pod fetch fails, we used to not add it to the endpointAddressMap resulting in it getting deregistered and its token deleted. 
- This PR makes the cases we deregister in more explicit-- only if the pod does not exist, we will deregister the service instance and delete the token. This PR also switches endpointAddressMap to a map that marks endpoints explicitly for deregistration (or not). And only entries explicitly marked in the map for deregistration are deregistered. 

Unit tests, Manually testing by changing the code to simulate pod fetch errors and with the fix in this PR it doesn't deregister the instance and delete the token

### Checklist ###
- [x] Tests added

### DIFF
--- a/.changelog/3758.txt
+++ b/.changelog/3758.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+control-plane: fix an issue where ACL token cleanup did not respect a pod's GracefulShutdownPeriodSeconds and
+tokens were invalidated immediately on pod entering Terminating state.
+```
--- a/.changelog/3758.txt
+++ b/.changelog/3758.txt
@@ -1,4 +1,4 @@
 ```release-note:bug
-control-plane: fix an issue where ACL token cleanup did not respect a pod's GracefulShutdownPeriodSeconds and
-tokens were invalidated immediately on pod entering Terminating state.
+control-plane: fix an issue where ACL tokens would prematurely be deleted and services would be deregistered if there
+was a K8s API error fetching the pod.
 ```
--- a/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
@@ -197,6 +197,7 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 					if k8serrors.IsNotFound(err) {
 						deregisterEndpointAddress[address.IP] = true
 					} else {
+						deregisterEndpointAddress[address.IP] = false
 						r.Log.Error(err, "failed to get pod", "name", address.TargetRef.Name)
 						errs = multierror.Append(errs, err)
 					}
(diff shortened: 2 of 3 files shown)


control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -1621,10 +1626,15 @@ func getMultiPortIdx(pod corev1.Pod, serviceEndpoints corev1.Endpoints) int {
...
+	if deregisterEndpointAddress == nil {
 		return true
 	}
-	_, ok := endpointsAddressesMap[address]
-	return !ok
+	deregister, ok := deregisterEndpointAddress[address]
+	if ok {
+		return deregister
+	}
+	return false
```
DanStough, line 1639: Don't we want to return `true` if it's not in the map, like in the case an service instance no longer appears in the endpoints subset?

control-plane/connect-inject/controllers/endpoints/endpoints_controller_test.go
```diff
@@ -2264,6 +2265,242 @@ func TestParseLocality(t *testing.T) {
...
+	client.WithWatch
+}
+
+func (c fakeClientWithPodCustomization) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	if key.Name == "pod1" {
+		return errors.New("some fake error while fetching pod")
+	}
+	err := c.WithWatch.Get(ctx, key, obj, opts...)
+	return err
+}
```
DanStough, line 2502: Nice!

DanStough (commented):
One blocking question about the `deregister` function.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
ndhanushkodi: Thanks for this-- I intended the change to be that only if you explicitly add the address to the map for deregistration then you should deregister, but I think there's too many cases where we do rely on the implicit deregistration when a service instance no longer appears in the endpoints, and a bunch of tests were failing. So I just changed the logic to be very specific to the pod-fetch-error case, so only if there is a k8s api related pod fetch error, we explicitly choose to add it to the map and say "don't deregister this", and now the deregister function does return true if it's not in the map!

.changelog/3758.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+control-plane: fix an issue where ACL token cleanup did not respect a pod's GracefulShutdownPeriodSeconds and
+tokens were invalidated immediately on pod entering Terminating state.
```
DanStough, line 3: I think this might be a copy-paste error from the other PR; we want something specific here, right?

DanStough (approved):
LGTM! Just one non-blocking note about the changelog.

.changelog/3758.txt
ndhanushkodi: LOOL yes thanks!

PR #3766, opened 2024-03-15
Add support for Nomad transparent proxy

tgross:
_Work-in-progress, see https://github.com/hashicorp/nomad/issues/10628_

Nomad will implement support for Connect transparent proxy. Unlike in K8s, the
CNI plugin can't contact the Nomad API to read allocation metadata (pod labels)
to get the iptables configuration, and doesn't use the rest of the Consul-K8s
control plane to inject that metadata. Instead, Nomad will pass the iptables
configuration JSON-serialized in the CNI arguments.

This changeset implements the behavior switch by detecting the `IPTABLES_CONFIG`
argument in the CNI arguments. This hypothetically allows for non-Nomad
workflows to use the same code path, if desired.

Ref: https://github.com/hashicorp/nomad/issues/10628

### Checklist ###
- [x] Tests added


tgross:
Closing in lieu of https://github.com/hashicorp/consul-k8s/pull/3795 which was pushed directly, now that we've got GitHub permissions sorted.

PR #3768, opened 2024-03-18
upgraded helm v3 to address GHSA-jw44-4f3j-q396

wilkermichael:
- This PR addresses [CVE GHSA-jw44-4f3j-q396 ](https://github.com/advisories/GHSA-jw44-4f3j-q396). For now we will disable the scan to get the pipeline in order as there is currently no fix. `helm.sh/helm/v3` 3.14.4 will supposedly contain only bug fixes and will be released on April 10, 2024. So maybe it will be included with that.

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.release/security-scan.hcl
+++ b/.release/security-scan.hcl
@@ -39,7 +39,8 @@ binary {
         # NET-8174 (2024-02-26): Missing YAML Content Leads To Panic (requires malicious plugin)
         "GHSA-r53h-jv2g-vpx6", 
         "CVE-2024-26147", # alias
-        "GHSA-jw44-4f3j-q396", # alias
+        "GHSA-jw44-4f3j-q396", # Tracked in NET-8174
+        "CVE-2019-25210" # alias
       ]
     }
   }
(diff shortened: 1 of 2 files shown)


.release/security-scan.hcl
```diff
@@ -39,6 +39,7 @@ binary {
         # NET-8174 (2024-02-26): Missing YAML Content Leads To Panic (requires malicious plugin)
         "GHSA-r53h-jv2g-vpx6", 
         "CVE-2024-26147", # alias
+        "GHSA-jw44-4f3j-q396", # alias
```
zalimeni, line 42: ~ The idea here is to put in a comment referencing the NET ticket that tracks the resolution of the triage entry. We can add one or more aliases to the list if the scanner is hitting the same issue under a different ID (e.g. the GHSA vs. CVE ID).

This new GHSA seems to be https://osv.dev/vulnerability/CVE-2019-25210, so I don't think it's in the list yet. Can we add a new description comment for it in this and the other file?

It's probably also worth calling out that we aren't affected directly by this in our use of `helm/v3` as a library (it's a `helm` CLI issue from what I can see). We can probably track it under NET-8174 (3.14 upgrade) as well rather than w/ a separate issue.

zalimeni (approved):
One comment on comments (😄), but approving to unblock once that's resolved.

.release/security-scan.hcl
wilkermichael: @zalimeni does the latest commit address your comments?

zalimeni (approved):
@wilkermichael totally, thanks a bunch for updating here and in the ticket!

PR #3779, opened 2024-03-19
[NET-8412] Fix order of APIGW ACL policy/role creation

jm96441n:
- cache role/policy for api-gateway so we know if we need to create so we avoid getting errors in consul around trying to create a role/policy that already exists

### Checklist ###
- [ ] Tests added

### DIFF
--- a/control-plane/api-gateway/cache/consul.go
+++ b/control-plane/api-gateway/cache/consul.go
@@ -83,6 +83,12 @@ type Cache struct {
 	subscribers     map[string][]*Subscription
 	subscriberMutex *sync.Mutex
 
+	gatewayNameToPolicy map[string]*api.ACLPolicy
+	policyMutex         *sync.Mutex
+
+	gatewayNameToRole map[string]*api.ACLRole
+	aclRoleMutex      *sync.Mutex
+
 	namespacesEnabled       bool
 	crossNamespaceACLPolicy string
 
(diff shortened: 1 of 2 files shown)


control-plane/api-gateway/cache/consul.go
```diff
@@ -83,6 +83,12 @@ type Cache struct {
 	subscribers     map[string][]*Subscription
 	subscriberMutex *sync.Mutex
 
+	gatewayNameToPolicy map[string]*api.ACLPolicy
+	policyMutex         *sync.Mutex
+
+	gatewayNameToRole map[string]*api.ACLRole
```
jm96441n, line 89: I debated having the values here just be the aclpolicy name or id (same for the role) but decided on keeping the entire role/policy on there in case we need more from there in the future but don't feel too strongly about it

control-plane/api-gateway/cache/consul.go
```diff
@@ -339,21 +349,46 @@ func (c *Cache) Write(ctx context.Context, entry api.ConfigEntry) error {
...
+
+		if existing == nil {
+			return createPolicy()
+		}
+
+		if err != nil {
+			return "", err
+		}
+
+		return existing.ID, nil
```
andrewstucki, line 388: Only real suggestion is to make this a bit less nested and more linear like the role code below, it's easier to follow in the below function

andrewstucki (approved):
Looks like it needs a changelog entry, but generally LGTM

pawellegowski89:
I don't think this fix is ​​working properly.

After adding a CRD in the form of API gateway and then uninstalling it in consul, the role remains and adding such a CRD again causes an error.

The error occurs even without any intervention, if we create and raise the environment, the API gateway will no longer be raised, but will hang on INIT due to the desire to re-add an existing role, i.e. the same error again:

` Reconciler error        {"controller": "gateway", "controllerGroup": "gateway.networking.k8s.io", "controllerKind": "Gateway", "Gateway": {"name":"mesh-api-gateway","namespace":"data"}, "namespace": "data", "name": "mesh-api-gateway", "reconcileID": "739cd7fb-540e-46f2-b6dd-653baf933f1a", "error": "Unexpected response code: 500 (Invalid Role: A Role with Name \"managed-gateway-acl-role-mesh-api-gateway\" already exists)"}
`

PR #3785, opened 2024-03-19
Bugfix: Don't recreate servicemap for catalog sync

jukie:
- I recently contributed https://github.com/hashicorp/consul-k8s/pull/3693 only to discover a bug that will add every service to the serviceMap and trigger a registration in consul. This removes that bug.

Existing tests and a manual build/deploy in my environment to confirm we don't register everything

With my previous change, if spinning up consul catalog sync you'll see every k8s service become registered regardless of configuration. With this change you should only see intended services become registered.

### Checklist ###
- [ ] Tests added


jukie:
@jmurret if you can take a look. My last PR introduced a bug it seems.

jmurret (approved):
🙏 thank you for quickly finding and fixing this.

PR #3795, opened 2024-03-20
Cni iptables config from cni args

tgross:
_Work-in-progress, not ready for review. See https://github.com/hashicorp/nomad/issues/10628. Currently includes #3794_.

Nomad will implement support for Connect transparent proxy. Unlike in K8s, the
CNI plugin can't contact the Nomad API to read allocation metadata (pod labels)
to get the iptables configuration, and doesn't use the rest of the Consul-K8s
control plane to inject that metadata. Instead, Nomad will pass the iptables
configuration JSON-serialized in the CNI arguments.

This changeset implements the behavior switch by detecting the `IPTABLES_CONFIG`
argument in the CNI arguments. This hypothetically allows for non-Nomad
workflows to use the same code path, if desired.

Ref: https://github.com/hashicorp/nomad/issues/10628

### Checklist ###
- [x] Tests added


control-plane/cni/main.go
```diff
@@ -132,9 +135,11 @@ func (c *Command) cmdAdd(args *skel.CmdArgs) error {
 
 	podNamespace := string(cniArgs.K8S_POD_NAMESPACE)
 	podName := string(cniArgs.K8S_POD_NAME)
+	cniArgsIPTablesCfg := string(cniArgs.IPTABLES_CONFIG)
```
tgross, line 138: Note for reviewers: this arg name was what we landed on in the [RFC NMD-191 (internal link)](https://docs.google.com/document/d/1GTC-j6sIgQ3LtYcuoDxnAcluenKZU4WHybth2ngcoOc/edit) so that it's generic rather than specific to Nomad. But if we think this is likely to have a namespace collision with other CNI args in the ecosystem, we might want this to be HashiCorp-specific. I don't have a good handle on whether ecosystem for k8s cares much about this, so I'll defer to experts on that here. I'll just need to update the equivalent argument name in https://github.com/hashicorp/nomad/pull/20183

ndhanushkodi: > But if we think this is likely to have a namespace collision with other CNI args in the ecosystem, we might want this to be HashiCorp-specific.

That's a good callout-- I don't know of any specific other CNI's that take in an IPTABLES_CONFIG arg, but it seems like since it's easy enough to make the name more specific we might as well do it now rather than have the potential to run into collisions? Since that is quite a generic arg name and another CNI could plausibly be configuring iptables.

Also an idea is it could be feature/tproxy specific instead of hashicorp specific and that way company name doesn't have to make it into an argument name.

control-plane/cni/main.go
```diff
@@ -220,15 +230,21 @@ func (c *Command) cmdAdd(args *skel.CmdArgs) error {
 		return fmt.Errorf("could not apply iptables setup: %v", err)
 	}
 
-	// We do not throw an error here because kubernetes will often throw a benign error where the pod has been
-	// updated in between the get and update of the annotation. Eventually kubernetes will update the annotation
-	ok = c.updateTransparentProxyStatusAnnotation(podName, podNamespace, complete)
-	if !ok {
-		logger.Info("unable to update %s pod annotation to complete", keyTransparentProxyStatus)
+	if cniArgsIPTablesCfg == "" {
```
ndhanushkodi, line 233: Maybe somewhere near the top we can add a comment that if `cniArgsIPTablesCfg` is empty then we are on K8s and otherwise we are on Nomad

control-plane/cni/main_test.go
```diff
@@ -355,3 +377,65 @@ const missingIPsStdinData = `{
     "name": "consul-cni",
     "type": "consul-cni"
 }`
+
+// TODO: need to make this match what Nomad emits
```
ndhanushkodi, line 381: Is this resolved since you mentioned you tested it end to end?

ndhanushkodi (approved):
🎉 Nothing blocking, just thoughts on the comment you made

control-plane/cni/main.go
tgross: Good call, will do.

control-plane/cni/main_test.go
tgross: :facepalm: Yes, thanks. Will remove!

control-plane/cni/main.go
tgross: How does `TPROXY_IPTABLES_CONFIG` sound?

ndhanushkodi: Both seem fine, maybe `CONSUL_IPTABLES_CONFIG` because we sometimes reserve the consul prefix in other cases like for metadata.

PR #3800, opened 2024-03-20
improve reliability of acceptance tests

NiniOak:
- When are command fails to run with a error, include the output with the error code see [example](https://github.com/hashicorp/consul-k8s-workflows/actions/runs/8353236571/job/22864767746#step:20:2917)
- Wait for pods to rollout with timeout before moving on to next step of testing. 

CI should pass

CI should pass

### Checklist ###
- [ ] Tests added

### DIFF
--- a/acceptance/tests/api-gateway/api_gateway_test.go
+++ b/acceptance/tests/api-gateway/api_gateway_test.go
@@ -107,11 +107,14 @@ func TestAPIGateway_Basic(t *testing.T) {
 			logger.Log(t, "creating static-client pod")
 			k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.NoCleanup, cfg.DebugDirectory, "../fixtures/bases/static-client")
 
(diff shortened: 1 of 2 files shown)


acceptance/tests/api-gateway/api_gateway_test.go
```diff
@@ -107,11 +107,14 @@ func TestAPIGateway_Basic(t *testing.T) {
 			logger.Log(t, "creating static-client pod")
 			k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.NoCleanup, cfg.DebugDirectory, "../fixtures/bases/static-client")
 
+			k8s.RunKubectl(t, ctx.KubectlOptions(t), "wait", "--for=condition=available", "--timeout=5m", fmt.Sprintf("deploy/%s", "static-server"))
```
nathancoleman, line 110: If at the end of the 5m timeout the pods still aren't available, will the test continue on or fail?

NiniOak: The test would just continue, it will fail further down the line when other conditions are not met.

nathancoleman (approved):

PR #3811, opened 2024-03-26
API gateway metrics

andrewstucki:
Adds metric scraping ability to API Gateway pods. The way this works is by writing values into the GatewayClassConfiguration that configures the metrics publishing capabilities. When the values are set, the proxy service gets registered with an extra `ProxyConfiguration` map that sets the `envoy_prometheus_bind_addr` value documented [here](https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#envoy_prometheus_bind_addr). That value, when set, causes the [dataplane bootstrapping code](https://github.com/hashicorp/consul-dataplane/blob/be5f2fedf24967e7edc896ea01424f2cecda0a2a/internal/bootstrap/bootstrap_config.go#L64-L76) to spin up an extra listener for prometheus metrics. In addition, we annotate the gateway with the proper prometheus upstream scraping annotations so that its kubernetes implementation will automatically ingest metrics, similar to our other proxies/gateways.

Spun up a stack with metrics enabled and watched them go into Prometheus.

### Checklist ###
- [ ] Tests added


.changelog/3811.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+api-gateway: Expose prometheus scrape metrics on api-gateway pods.
```
nathancoleman, line 2:
```suggestion
api-gateway: Expose Prometheus scrape metrics on api-gateway pods.
```

control-plane/api-gateway/common/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+	Enabled bool
+	Path    string
+	Port    int
+}
+
+func gatewayMetricsEnabled(gateway gwv1beta1.Gateway, gcc v1alpha1.GatewayClassConfig, config HelmConfig) bool {
+	// first check our annotations, if something is there, then it means we've explicitly
+	// annotated metrics collection
+	if scrape, isSet := gateway.Annotations[constants.AnnotationEnableMetrics]; isSet {
+		return scrape == "true"
```
nathancoleman, line 29: Should we compare after lowercasing, or do we want to be particular about the annotation value being `true` exactly (vs. `True`)?

control-plane/api-gateway/common/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+	return config.DefaultPrometheusScrapePort
+}
+
+func gatewayMetricsPort(gateway gwv1beta1.Gateway, gcc v1alpha1.GatewayClassConfig, config HelmConfig) int {
+	portString := fetchPortString(gateway, gcc, config)
+
+	port, err := strconv.Atoi(portString)
+	if err != nil {
+		// if we can't parse the port string, just use the default
+		// TODO: log an error
```
nathancoleman, line 63: Are you intending to merge with this TODO in place?

control-plane/api-gateway/common/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+		return defaultScrapePort
+	}
+
+	if port < 1024 || port > 65535 {
+		// if we requested a privileged port, use the default
+		// TODO: log an error
```
nathancoleman, line 69: Same question here

control-plane/api-gateway/common/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+		return path
+	}
+
+	// if it's not set on the annotation, then we check to see if it's set on the GatewayClassConfig
+	if gcc.Spec.Metrics.Path != nil {
+		return *gcc.Spec.Metrics.Path
+	}
+
+	// otherwise, fallback to the global helm setting
+	return config.DefaultPrometheusScrapePath
```
nathancoleman, line 89: Should we be doing any validation on the path here?

control-plane/subcommand/gateway-resources/command.go
```diff
@@ -604,6 +636,35 @@ func exponentialBackoffWithMaxIntervalAndTime() *backoff.ExponentialBackOff {
...
+func getMetricsEnabled(v string) (bool, bool) {
+	if v == "true" {
+		return true, true
+	}
+	if v == "false" {
+		return false, true
+	}
+	return false, false
+}
+
```
nathancoleman, line 667: It feels odd to me that we're validating in 2 places: once here for the managed gateway class and once in `control-plane/api-gateway/common/metrics.go` that applies to all consumed `GatewayClasses`.

I don't have super strong feelings about it but am more generally concerned about having two pieces of similar code to maintain if we ever need to change things. Just for your consideration.

nathancoleman (commented):
Just a few questions. Nice work!

control-plane/api-gateway/common/metrics.go
andrewstucki: Ideally, yes, but from what I can tell we don't really have solid path validation anywhere else in our implementation (unless you can point me to a prior art?). Was thinking we'd just keep things equivalent for now, but we could see if we could `url.Parse` or regex match the url string if we wanted to? I believe that if this is invalid you'll just get a dataplane process that fails to start up envoy.

control-plane/api-gateway/common/metrics.go
andrewstucki: Yeah, for now just to get this out the door, we can go back later and pass a logger instance into this if we want to log on an invalid port.

control-plane/subcommand/gateway-resources/command.go
andrewstucki: So, I suppose we can refactor this into some generic validation block that could be re-used between the different subcommands, but it is validating two different methods of input currently:

1. The raw helm values themselves.
2. Values that may come from helm, from the spec of the `GatewayClassConfig`, or from an annotation.

Maybe I can do a quick follow-up pass after merge to consolidate the validation? Just thinking that we could keep as-is for now if we need this merged by later today.

control-plane/api-gateway/common/metrics.go
andrewstucki: I can update this to a `strconv.ParseBool` call.

control-plane/subcommand/gateway-resources/command.go
nathancoleman: I'm okay with keeping as-is for now, just wanna make sure we're intentional about it

nathancoleman (approved):

PR #3813, opened 2024-03-26
Remove anyuid SCC requirement for OpenShift

curtbushko:
- OpenShift does not want you to use a hardcoded user id and group when running containers.
- We currently hardcode 5995 and 5996 for the init and inject containers.
- OpenShift tells you what ids you can use by adding an annotation to the namespace you are deploying to. 
- We now read that annotation from the namespace and change the ids at run time.
- This removes the requirement of having to run `oc adm policy add-scc-to-group anyuid system:serviceaccounts:$TARGET_NAMESPACE` when deploying to OpenShift

- Added some helper code to connect-inject/common that reads from the namespace and gets the UID and Group to be used. This code has some fallbacks based from previous versions of OpenShift
- NOTE: The word 'range' is used in the OpenShift annotation but it is not really a range (anymore). 
- If on OpenShift, use the IDs for init and connect inject.
- For CNI, we add an annotation to the pod with the redirect config. It includes the UID in the json blob. CNI uses this annotation/blob during Pod Network setup.
- Acceptance tests were creating the anyuid policy when running on OpenShift and that is not longer needed. (Files dropped/kustomize changes).

- Unit tests
- Created a local [CRC](https://access.redhat.com/documentation/en-us/red_hat_codeready_containers/1.0/html/getting_started_guide/getting-started-with-codeready-containers_gsg) OpenShift cluster, deployed consul and static server to manually check the IDs.
- Had @wilkermichael run the connect acceptance tests against a real OpenShift cluster using my control plane image.

👀 

### Checklist ###
- [x] Tests added


acceptance/framework/consul/helm_cluster.go
```diff
@@ -702,16 +702,14 @@ func configureNamespace(t *testing.T, client kubernetes.Interface, cfg *config.T
...
 func configureSCCs(t *testing.T, client kubernetes.Interface, cfg *config.TestConfig, namespace string) {
-	const anyuidClusterRole = "system:openshift:scc:anyuid"
 	const privilegedClusterRole = "system:openshift:scc:privileged"
-	anyuidRoleBinding := "anyuid-test"
 	privilegedRoleBinding := "privileged-test"
 
 	// A role binding to allow default service account in the installation namespace access to the SCCs.
 	{
-		for clusterRoleName, roleBindingName := range map[string]string{anyuidClusterRole: anyuidRoleBinding, privilegedClusterRole: privilegedRoleBinding} {
+		for clusterRoleName, roleBindingName := range map[string]string{privilegedClusterRole: privilegedRoleBinding} {
```
wilkermichael, line 712: I guess technically this doesn't need to be a loop anymore

control-plane/connect-inject/common/openshift.go
```diff
@@ -0,0 +1,129 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Get the user id from the OpenShift annotation 'openshift.io/sa.scc.uid-range'.
+func GetOpenShiftUID(ns *corev1.Namespace) (int64, error) {
```
wilkermichael, line 28: nit: Since this is a public method (and you have a function comment), should put the name first in the function comment, same for below

control-plane/connect-inject/common/openshift.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	return uid, nil
+}
+
+// Parse the UID "range" from the annotation string. The annotation can either have a '/' or '-' as a separator.
+// '-' is the old style of UID from when it used to be an actual range.
+func parseOpenShiftUID(val string) (int64, error) {
```
wilkermichael, line 47: maybe add an example UID so it's easy to tell what the parsing is parsing

control-plane/connect-inject/common/openshift_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestOpenShiftUID(t *testing.T) {
```
wilkermichael, line 15: I think you're just missing the case where the openshiftID is found but is empty

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -214,33 +218,70 @@ func (w *MeshWebhook) consulDataplaneSidecar(namespace corev1.Namespace, pod cor
 	// When transparent proxy is enabled, then consul-dataplane needs to run as our specific user
 	// so that traffic redirection will work.
 	if tproxyEnabled || !w.EnableOpenShift {
-		if pod.Spec.SecurityContext != nil {
-			// User container and consul-dataplane container cannot have the same UID.
-			if pod.Spec.SecurityContext.RunAsUser != nil && *pod.Spec.SecurityContext.RunAsUser == sidecarUserAndGroupID {
-				return corev1.Container{}, fmt.Errorf("pod's security context cannot have the same UID as consul-dataplane: %v", sidecarUserAndGroupID)
+		// In non-OpenShift environments we set the User and group ID for the sidecare to our values.
```
wilkermichael, line 221:
```suggestion
		// In non-OpenShift environments we set the User and group ID for the sidecar to our values.
```

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -214,33 +218,70 @@ func (w *MeshWebhook) consulDataplaneSidecar(namespace corev1.Namespace, pod cor
...
-		container.SecurityContext = &corev1.SecurityContext{
-			RunAsUser:                pointer.Int64(sidecarUserAndGroupID),
-			RunAsGroup:               pointer.Int64(sidecarUserAndGroupID),
-			RunAsNonRoot:             pointer.Bool(true),
-			AllowPrivilegeEscalation: pointer.Bool(false),
-			ReadOnlyRootFilesystem:   pointer.Bool(true),
 		}
 	}
 
+	// When running tproxy and OpenShift,
```
wilkermichael, line 274: remove comment?

wilkermichael (approved):

acceptance/framework/consul/helm_cluster.go
curtbushko: Done!

missylbytes (approved):

PR #3816, opened 2024-03-27
Respect the configuration when configuring the GRPC port.

slackfan:
- The helm template client-daemonsets.yaml respects the client.grpc configuration when configuring and exposing the GRPC port 8502 for the consul client container. 

- Tests happened manually by deploying the server-statefulset and client-daemonset onto the same Kubernetes nodes. I confirmed that the clients were able to start without port collisions while the server.exposeGossipAndRPCPorts flag was enabled and the client.grpc flag was disabled. 

- Confirm that use cases I currently do not see are covered and still functional. 

### Checklist ###
- [x] Tests added

### DIFF
--- a/charts/consul/templates/client-daemonset.yaml
+++ b/charts/consul/templates/client-daemonset.yaml
@@ -443,9 +443,11 @@ spec:
               hostPort: 8501
               name: https
             {{- end }}
+            {{- if .Values.client.grpc }}
             - containerPort: 8502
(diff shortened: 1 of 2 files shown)


rexopl (approved):

justinas-wix:
When could we expect this to be merged?

grizzlybite:
Very necessary revision,please make a review  :)

PR #3428, opened 2024-01-03
build: Create arm64 packages as well

dekimsey:
### Changes proposed in this PR ###  

During the CRT on-boarding, packaging for other Linux architectures (arm64) was
not enabled. This change adds packaging support for those architectures. I've
specifically opted not to include 32-bit.

Other related updates:

 - To make future support a bit easier, I've enabled the build workflow from
   releng prefixed branches.
 - Using qemu emulation for testing package installs on other architectures,
   thus allowing us to validate the binaries work as intended
 - Minor alteration to the package install tests to use yum instead of rpm

Note: I tried to add 386/arm packages but pivoted away. There's something suspicious with the 386 binary specifically (see comments). Since usage of that architecture is low, I punted.

### How I've tested this PR ###

Please note, these packages have not been extensively tested on their respective platforms. I did update the pipeline to run the smoke test (consul version) against the arm64 packages in an arm64 user-space. So I think it's in a good place.

### How I expect reviewers to test this PR ###

Download builds and test in the same manner existing binaries are tested. Docker containers would be sufficient.

### Checklist ###
- [x] ~~Tests added~~
- [x] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry) 

### Related ###

See #1132.
Related to hashicorp/releng-support#178.

PR #3821, opened 2024-03-27
1.1.11 sync to 1.1.x

kisunji:
-
-

### Checklist ###
- [ ] Tests added


zalimeni (approved):
Versions LGTM, but wondering whether #3428 should have been included in backports - we might want to wait on this one until that's verified.

Approving to unblock in case I'm offline and wrong about ^

PR #3829, opened 2024-03-28
fix version output for `consul-cni`

tgross:
The `consul-cni` plugin emits "version unknown" because the CNI library's `PluginMain` uses a global variable that isn't being set as part of our build process. Import the `control-plane/version` package so that we have an identical version in builds across both binaries.

Before:

```
$ ./control-plane/cni/bin/consul-cni --version
CNI consul-cni plugin version unknown
CNI protocol versions supported: 0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 1.0.0
```

After:

```
$ make control-plane-dev
...
$ ./control-plane/cni/bin/consul-cni --version
CNI consul-cni plugin v1.5.0-dev
CNI protocol versions supported: 0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 1.0.0
```

As above

### Checklist ###
- [x] ~Tests added~ n/a

---

Note for reviewers: an alternative approach I considered was to "vendor" the `control-plane/version` package into the `consul-cni` plugin. This seems to work but doesn't make a different in binary size, and makes for a second place folks would need to update in releases, so I opted for an import. The nice thing about this (maybe?) is that it also keeps the indirect dependencies up-to-date in `consul-cni`. Happy to change approaches here though, if you'd like.

Nomad fingerprints CNI binary versions to ensure that the node we place workloads onto can actually support the requested networking configuration, so we'll need to this for Nomad's transparent proxy support https://github.com/hashicorp/nomad/issues/10628


ndhanushkodi (approved):
Thanks for leaving your note on the approach, sounds/looks good to me!

PR #3860, opened 2024-04-03
[1.3.4] Remove Duplicate UBI Tags (#3265)

kisunji:
- Amalgamate UBI with Dockerhub and Redhat tags into one step
- Avoids a production incident that errors on duplicate tags: hashicorp/releng-support#123

-
-

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -329,7 +329,7 @@ jobs:
 
   build-docker-ubi:
     name: Docker ${{ matrix.arch }} ${{ matrix.fips }} UBI builds
-    needs: [get-product-version, build]
+    needs: [get-product-version, get-go-version, build]
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -327,9 +327,9 @@ jobs:
           extra_build_args: |
             GOLANG_VERSION=${{ needs.get-go-version.outputs.go-version }}
 
-  build-docker-ubi-redhat-registry:
-    name: Docker ${{ matrix.arch }} ${{ matrix.fips }} UBI build for RedHat Registry
-    needs: [get-product-version, get-go-version, build]
+  build-docker-ubi:
+    name: Docker ${{ matrix.arch }} ${{ matrix.fips }} UBI builds
+    needs: [get-product-version, build]
```
zalimeni, line 332: I think we still need `get-go-version` here as a dependency due to the `extra_build_args` that relies on it

zalimeni (approved):
LGTM other than the `get-go-version` missing. I also updated the PR title since this is a point release backport and we'll want to make sure we don't forget `.x` as well.

If needed, we can add back the Dockerhub dev tags [missing](https://github.com/hashicorp/consul-k8s/pull/3265/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L474) for FIPS builds in a separate PR and backport all the way down.

Issue #3406, opened 2023-12-19
Only error for config entries from different datacenters when the config entries are different


komapa:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Is your feature request related to a problem? Please describe.

We get a lot of reconciliation errors right now because we have ServiceRouter (for example) with the same content for the same app in two different datacenters where that app is deployed.

#### Use Case(s)

We apply the same config entries to different datacenters because of the way our automation works and we would like to not error when the end result in Consul is the same.

#### Contributions

```diff
Index: control-plane/controllers/configentry_controller.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/control-plane/controllers/configentry_controller.go b/control-plane/controllers/configentry_controller.go
--- a/control-plane/controllers/configentry_controller.go	(revision 0bd21a7814d69f39f82abd24db59bc22d2a46d47)
+++ b/control-plane/controllers/configentry_controller.go	(date 1703008805097)
@@ -213,25 +213,25 @@
 	requiresMigration := false
 	sourceDatacenter := entry.GetMeta()[common.DatacenterKey]
 
-	// Check if the config entry is managed by our datacenter.
-	// Do not process resource if the entry was not created within our datacenter
-	// as it was created in a different cluster which will be managing that config entry.
-	if sourceDatacenter != r.DatacenterName {
+	if !configEntry.MatchesConsul(entry) {
+		// Check if the config entry is managed by our datacenter.
+		// Do not process resource if the entry was not created within our datacenter
+		// as it was created in a different cluster which will be managing that config entry.
+		if sourceDatacenter != r.DatacenterName {
 
-		// Note that there is a special case where we will migrate a config entry
-		// that wasn't created by the controller if it has the migrate-entry annotation set to true.
-		// This functionality exists to help folks who are upgrading from older helm
-		// chart versions where they had previously created config entries themselves but
[...]

PR #3873, opened 2024-04-03
GH-3406 - Only error for config entries from different datacenters when the config entries are different

jmurret:
Fixes https://github.com/hashicorp/consul-k8s/issues/3406

-
-

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.changelog/3873.txt
+++ b/.changelog/3873.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ConfigEntries controller: Only error for config entries from different datacenters when the config entries are different
+```
--- a/control-plane/controllers/configentries/configentry_controller.go
+++ b/control-plane/controllers/configentries/configentry_controller.go
@@ -225,27 +225,26 @@ func (r *ConfigEntryController) ReconcileEntry(ctx context.Context, crdCtrl Cont
 
 	requiresMigration := false
 	sourceDatacenter := entry.GetMeta()[common.DatacenterKey]
-	matchesConsul := configEntry.MatchesConsul(entry)
 
-	// Check if the config entry is managed by our datacenter.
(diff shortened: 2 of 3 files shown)


zalimeni (approved):

zalimeni:
Confirmed this will only need backporting in `release/1.4.x` (file DNE prior)

komapa:
LGTM, but can we backport all the way to 1.1.x? Thank you!

jmurret:
It looks like there is more logic to consider with the migrate-entry annotation (`TestConfigEntryController_Migration`) that needs to be considered more closely.  will take a look tomorrow.

david-yu:
@komapa I assume the same config entry is applied perhaps via Argo on different DCs that have the same name and same config? Also to be clear the CRD is applied on different clusters that are connected to different external servers is that correct? If so I agree that they should be marked as duplicate since they are different. If you have a test case we should explicitly check for that would be helpful.

komapa:
> @komapa I assume the same config entry is applied perhaps via Argo on different DCs that have the same name and same config? Also to be clear the CRD is applied on different clusters that are connected to different external servers is that correct? If so I agree that they should be marked as duplicate since they are different. If you have a test case we should explicitly check for that would be helpful.

Yes, this is the exact use case. We have developers deploy from shared Kustomize templates and they can deploy to one region (one consul datacenter) or another region (another consul datacenter) or to both (the datacenters are WAN federated). So we do not really have the ability to only put the CRDs in one or the other.

I realize this is not idea, but only because of the way the Kubernetes functionality is not a great use-case for WAN federated clusters. We are kind of asking here to meet us in the middle and not hard fail reconciliation when the contents of the config entry are identical. You should still fail like you do not if they ever get out of sync.

PR #3874, opened 2024-04-04
Sync Catalog: Add k8s endpoint health state to consul health check

### LINKED ISSUE #3898: Sync Catalog: Utilize EndpointSlice conditions when endpoints are updated
### Community Note

---

#### Is your feature request related to a problem? Please describe.

Sync Catalog currently doesn't account for an Endpoint's condition state. As a result, an endpoint that is in `terminating` state will remain being synced to Consul until it's fully removed. This can cause issues in environments where a user may want to healthily wait for connections to drop off on a pod (e.g. the synced consul instance) before fully terminating or the opposite use case of not wanting to register the Endpoint in Consul until it's deemed `ready`.

#### Feature Description

With https://github.com/hashicorp/consul-k8s/pull/3693 Sync Catalog changed from using `Endpoints` to `EndpointSlices`. This was mostly done to provide access to topology information but another feature of EndpointSlices is [Conditions](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#conditions) which expose additional state about whether the endpoint is `ready`, `serving`, and/or `terminating`.
Sync Catalog could use this extra information to enhance existing behavior and only register `Endpoint`'s that are marked as `ready` and clean up Endpoints once they are marked as `terminating`. 

#### Use Case(s)

- I'm rolling a new replicaset for my Kubernetes Deployment and my workload has a long initialization phase. I don't want it to be registered to Consul until all readiness checks have passed and the underlying Endpoint is marked as `ready` or `serving`.
- I'm using consul-template to add consul instances as backend targets in an external system. It would be ideal to have these Consul instances updated with some state information or removed whenever the pod is being torn down to allow for a buffer period and any open connections to complete.  

#### Contributions

https://github.com/hashicorp/consul-k8s/pull/3874 adds this feature

jukie:
I recently contributed a change here from using Endpoints to EndpointSlices. 
EndpointSlices offer additional state information such as readiness information but in the current approach we aren't taking advantage of this and instead only add/remove endpoints based on creation or deletions.
We can utilize the Endpoint conditions to accurately represent a registered service's health-state instead of always registering as healthy.

- Added tests
- Manual build/test in live environment

An easy way to test would be to launch a pod with a long terminationGracePeriod that stays up and initiate a deployment rollout or delete the pod. Once the pod enters terminating state you'll see the same state in the relevant `EndpointSlice`.
At this point the registered consul service instance will have its health check updated to "critical" phase.

### Checklist ###
- [x] Tests added

fixes #3898


jukie:
@david-yu could you possibly help get a review on this please?

jukie:
bumpety bump in case anyone's watching

ndhanushkodi:
@jukie taking a look at this now

control-plane/catalog/to-consul/resource.go
```diff
@@ -739,67 +744,72 @@ func (t *ServiceResource) registerServiceInstance(
...
-					addr = t.serviceHostnameMap[key].hostName
-					epPort = int(t.serviceHostnameMap[key].port)
-				} else {
-					addr = endpointAddr
-					if addr == "" && useHostname {
-						addr = *endpoint.Hostname
+			// Only consider endpoints that are ready
+			// nil represents an unknown state that can be interpreted as a non-terminating endpoint (assume ready or at least a state we shouldn't ignore)
+			// Ref: https://github.com/kubernetes/api/blob/5147c1a32f6a0b9b155bb84e59f933e0ff8a3792/discovery/v1/types.go#L128-L151
+			if endpoint.Conditions.Ready == nil || *endpoint.Conditions.Ready {
```
jukie, line 750: @ndhanushkodi this is probably the part that I'd specifically like to callout that someone should review. I couldn't find any documented examples where `Ready|Terminating` would be nil and the edge cases that would entail so could remove this if preferred but the docs do imply this is "safe".

ndhanushkodi: > nil represents an unknown state that can be interpreted as a non-terminating endpoint (assume ready or at least a state we shouldn't ignore)

This makes sense to me, I'd lean towards keeping the nil check as you've done since it may be worth being safe and considering that endpoint rather than ignoring it.

control-plane/catalog/to-consul/resource_test.go
```diff
@@ -2154,11 +2154,22 @@ func createEndpointSlice(t *testing.T, client *fake.Clientset, serviceName strin
...
+					Addresses: []string{"3.3.3.3"},
+					Conditions: discoveryv1.EndpointConditions{
+						Ready:       pointer.Bool(false),
+						Serving:     pointer.Bool(false),
+						Terminating: pointer.Bool(true),
+					},
+					TargetRef: &targetRef,
+					NodeName:  &node1,
+					Zone:      pointer.String("us-west-2a"),
+				},
```
ndhanushkodi, line 2172: Can you add one more address with `Ready: pointer.Bool(true),`?  Just to cover all cases.

ndhanushkodi (commented):
Confirmed that the relevant acceptance tests are passing, and the changes look good to me. Just one small comment to address!

Tomorrow I want to make sure the failing peering acceptance tests are unrelated, and if you're able to make that one small test case change then I think we can merge after.

control-plane/catalog/to-consul/resource_test.go
jukie: Sure I'll add more but could you elaborate please? I think all 3 possible values for `Ready` should be covered here and `Ready: pointer.Bool(true)` is there on [line 2155](https://github.com/hashicorp/consul-k8s/pull/3874/files#diff-49e18150b63f1fa627a95f2f6c502f9f554a15ed8c9ac740787e3b7fa1ede710R2155).

ndhanushkodi: Ah you're right, I didn't see that in the diff initially.

ndhanushkodi (approved):

control-plane/catalog/to-consul/resource_test.go
jukie: No worries! That line was already there

ndhanushkodi:
Hey @jukie I was about to merge yesterday but then I was thinking more about the use cases you listed in the issue, and since Consul supports health checks, should we instead use Consul healthchecks to reflect the readiness of an endpoint, rather than ignoring it entirely if its non-ready? That way the user can configure Consul as they wish based on the health checks. I talked to the team and the consensus was that it makes more sense to use the health checks to reflect the readiness status. 

If you're willing to make that change, I'll keep an eye on this PR (or a new PR if you make one, feel free to tag me) for those changes, or I can bring up the changes for prioritization within the team and work on it myself (but I'm not sure when I'll be able to commit to having this done). Thanks so much for your effort on this.

jukie:
@ndhanushkodi could you explain how that would work? I think that may be in addition to or an extension of this change vs instead of. It's not guaranteed that consul servers (I might be misunderstanding where the checks would execute from) would have network access back to the registered endpoints.
Since consul-k8s is already watching for updates to the EndpointSlice objects that seems like the most efficient place to react to changes in readiness in real time.

jukie:
I'd be willing to make a new PR to support consul health checks as well but I'd like to understand the opinion that it wouldn't be a separate feature contribution. The existing behavior reacts to Service Endpoint changes and this extends that to make use of the additional state provided by EndpointSlices.

ndhanushkodi:
> I'd like to understand the opinion that it wouldn't be a separate feature contribution. The existing behavior reacts to Service Endpoint changes and this extends that to make use of the additional state provided by EndpointSlices.

Thanks for your willingness to contribute and so sorry for the late response!

The existing behavior seems to register all endpoints in the endpoint slice, regardless of the condition of that endpoint, so if I created a service where one of the endpoints went into a non-ready state, that would still get registered as a service instance in Consul. This PR would change the behavior to no longer register service instances in Consul for endpoints that are in a non-ready state. My thought with the suggestion above was we should continue to register non-ready endpoints into Consul as we do today, but instead mark the health check unhealthy. Rather than changing the behavior to not register the non-ready endpoints, I'm proposing the health check change as the way to deal with non-ready endpoints.

For the cases that use the function `registerServiceInstance`, I'm imagining making a change [here](https://github.com/hashicorp/consul-k8s/blob/main/control-plane/catalog/to-consul/resource.go#L797) where we change the health based on the endpoint condition rather than always marking it as passing. For [Nodeport services](https://github.com/hashicorp/consul-k8s/blob/main/control-plane/catalog/to-consul/resource.go#L613) which doesn't use `registerServiceInstance`, I'm thinking we'd have to manually add a check [here](https://github.com/hashicorp/consul-k8s/blob/main/control-plane/catalog/to-consul/resource.go#L658) and line 679.

Let me know if I'm misunderstanding anything you're saying, or if this makes sense. 

> If a user wants to always publish endpoints they can also set spec.publishNotReadyAddresses and that would already be supported.

This is for K8s Services right? Whether the endpoint addresses should exist in the K8s Endpoint itself? This change is about whether the K8s Endpoint addresses that exist should get synced to Consul if they are non-ready so I think that feature doesn't quite match the behavior I mentioned in my suggestion. 

Again, let me know if I'm misunderstanding what you mean!

jukie:
@ndhanushkodi Sorry for the late reply as well but coming back to this. You are correct that it would change behavior but I think that behavior matches the intention given the way that services would already be registered with details stating that "Kubernetes health checks passing" https://github.com/hashicorp/consul-k8s/blob/main/control-plane/catalog/to-consul/resource.go#L44-L48

If you'd like to include a way that replicates prior behavior for backwards compatibility I could instead add a new option that would always register an endpoint no matter what.

This change is to achieve the desired behavior of what catalog sync says it's doing. As shown in the link above when services are registered it is already assumed that they are "ready" and skipping non-ready endpoints looks to be a limitation in how Endpoints worked since that state information wasn't available. Now that EndpointSlices are in use we have more granular control and should make use of it.

jukie:
Thinking more, ignoring might not be the best path but instead adjusting the health check state that's added to the registered consul service.

jukie:
@ndhanushkodi I've made that adjustment to instead include endpoint health info and retain the logic of still registering all services. Could you have another look please?

jukie:
Not sure if @ndhanushkodi is still an active maintainer so adding @zalimeni as well

control-plane/catalog/to-consul/resource_test.go
```diff
@@ -2141,7 +2141,7 @@ func createEndpointSlice(t *testing.T, client *fake.Clientset, serviceName strin
 				{
```
ndhanushkodi, line 2141: nit: description of createEndpointSlice says it creates an endpoint slice with 2 endpoints and this now creates 3

control-plane/catalog/to-consul/resource_test.go
```diff
@@ -2154,11 +2154,22 @@ func createEndpointSlice(t *testing.T, client *fake.Clientset, serviceName strin
...
+					Addresses: []string{"3.3.3.3"},
+					Conditions: discoveryv1.EndpointConditions{
+						Ready:       ptr.To(false),
+						Serving:     ptr.To(false),
+						Terminating: ptr.To(true),
+					},
+					TargetRef: &targetRef,
+					NodeName:  &node1,
+					Zone:      ptr.To("us-west-2a"),
+				},
```
ndhanushkodi, line 2172: With the updates to this function I think we'd need to update assertions. Can you run the tests in this file locally? I think a bunch will fail.

For example `TestServiceResource_nodePort` uses `createEndpointSlice`, but it checks that the list of endpoints registered has length 2 and doesn't have any assertions on the health of those endpoints. I think to all the functions that use `createEndpointSlice`, we need to update the length check to 3 and add assertions for the 3rd endpoint, and add health assertions for all 3 endpoints. 

```go
	actual := syncer.Registrations
	require.Len(r, actual, 3) // this should be 3 instead of 2
	require.Equal(r, "foo", actual[0].Service.Service)
	require.Equal(r, "1.2.3.4", actual[0].Service.Address)
	require.Equal(r, 30000, actual[0].Service.Port)
	require.Equal(r, "k8s-sync", actual[0].Node)
	require.Equal(r, "foo", actual[1].Service.Service)
	require.Equal(r, "2.3.4.5", actual[1].Service.Address)
	require.Equal(r, 30000, actual[1].Service.Port)
	require.Equal(r, "k8s-sync", actual[1].Node)
	// these should get added 
	require.Equal(r, "foo", actual[2].Service.Service)
	require.Equal(r, "1.2.3.4", actual[2].Service.Address) // since it's on node 1
	require.Equal(r, 30000, actual[2].Service.Port)
	require.Equal(r, "k8s-sync", actual[2].Node)
        // we should also assert on the health status being registered to consul
```

ndhanushkodi (commented):
Sorry for the late response I was working on some other tickets. Thanks for making these changes, I agree with this approach. I think the resource_test.go file will have failing tests (when I pulled those changes locally, it doesn't work)-- are you able to update those? Once that's in a passing state I can run our pipelines against your changes and we can get it merged

control-plane/catalog/to-consul/resource_test.go
jukie: Fixed

jukie:
@ndhanushkodi sorry for my lack of checking beforehand! Updated the tests and re-ran acceptance ones as well.

control-plane/catalog/to-consul/resource_test.go
jukie: Updated these and everything passes now

jukie:
@ndhanushkodi was there any other changes you'd like to see here or is it ready to merge?

ndhanushkodi:
@jukie Running our pipelines in https://github.com/hashicorp/consul-k8s/pull/4381

jukie:
Could you share the failure output from the sameness acceptance test? That's passing for me locally

poblahblahblah:
We are _very_ interested in having this functionality at my company!

jukie:
@ndhanushkodi what release will this be included in?

gaeulautumn:
@jukie @ndhanushkodi 
Hi! I hope this feature can be released ASAP.
Could you please let me know when it will be available?

ndhanushkodi:
Hey all, sorry for missing your comments, it will go out in the next consul-k8s release in 1.7.0, which is in april

Issue #3876, opened 2024-04-04
Datadog autodiscovery annotations are hardcoding the server service name


dgdelahera:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

The Datadog autodiscovery annotations are hardcoding the server service name to `consul-server` (See https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/server-statefulset.yaml#L157 ). If we take a look to the [server service](https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/server-service.yaml#L10), this should be `{{ template "consul.fullname" . }}-server`

### Reproduction Steps

1. When running helm install with the following `values.yml`:
```yaml
global:
  name:  my-awesome-consul
  metrics:
    datadog:
      enabled: true
```

2. We can see that the Datadog agent is trying to connect to a service that doesn't exist:

```
2024-04-04 09:44:13 UTC | CORE | ERROR | (pkg/collector/worker/check_logger.go:70 in Error) | check:consul | 
Error running check: [{"message": "HTTPConnectionPool(host='consul-server.consul.svc', port=8500): 
Max retries exceeded with url: /v1/agent/metrics?format=prometheus 
(Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fb9db1b3d60>: 
Failed to establish a new connection: [Errno -2] Name or service not known'))", 
"traceback": "Traceback (most recent call last):

(...)
```

3. The actual service name is `my-awesome-consul-server.consul.svc` instead of `consul-server.consul.svc`

  --->

### Expected behavior

The annotations should use the correct service name.

PR #3877, opened 2024-04-04
Helm: fix hardcoded datadog ad url

dgdelahera:
- Address https://github.com/hashicorp/consul-k8s/issues/3876

Result can be checked in the test. Now the ad URL will always point to the correct server service name.

### Checklist ###
- [X] Tests added


dgdelahera:
I'm unable to sign the CLA:

image

david-yu:
Hi @dgdelahera We have a PR in the works that should resolve this! https://github.com/hashicorp/consul-k8s/pull/3685 Thanks for filing. We'll resolve your issue when that PR is closed.

PR #3880, opened 2024-04-04
[1.4.x] Post-release steps

kisunji:
-
-

### Checklist ###
- [ ] Tests added

### DIFF
--- a/charts/consul/Chart.yaml
+++ b/charts/consul/Chart.yaml
@@ -3,8 +3,8 @@
 
 apiVersion: v2
 name: consul
-version: 1.4.1
-appVersion: 1.18.1
+version: 1.4.2-dev
+appVersion: 1.18-dev
 kubeVersion: ">=1.22.0-0"
 description: Official HashiCorp Consul Chart
 home: https://www.consul.io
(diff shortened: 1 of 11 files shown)


charts/consul/Chart.yaml
```diff
@@ -18,7 +18,7 @@ annotations:
     - name: consul
       image: docker.mirror.hashicorp.services/hashicorppreview/consul:1.18-dev
     - name: consul-k8s-control-plane
-      image: docker.mirror.hashicorp.services/hashicorppreview/consul-k8s-control-plane:1.4-dev
+      image: docker.mirror.hashicorp.services/hashicorppreview/consul-k8s-control-plane:1.4.2-dev
```
zalimeni, line 21: I think we want to keep this at the floating tag similar to [`main`](https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/Chart.yaml)

charts/consul/values.yaml
```diff
@@ -86,7 +86,7 @@ global:
   # image that is used for functionality such as catalog sync.
   # This can be overridden per component.
   # @default: hashicorp/consul-k8s-control-plane:<latest version>
-  imageK8S: docker.mirror.hashicorp.services/hashicorppreview/consul-k8s-control-plane:1.4-dev
+  imageK8S: docker.mirror.hashicorp.services/hashicorppreview/consul-k8s-control-plane:1.4.2-dev
```
zalimeni, line 89: I think we want to keep this at the floating tag similar to [`main`](https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/values.yaml#L89) (same as above comment)

PR #3928, opened 2024-04-19
[NET-6466] Remove secrets from termgw role

jm96441n:
- remove secrets access from terminating gateway as this is no longer needed

1. build a local version of the control plane by running `make dev-docker` from the root of this repo
1. clone the following gist https://gist.github.com/jm96441n/8671cc5e73bf950955c96f7cd3281694
2. checkout this branch in consul-k8s repo on your machine
3. in `start.sh` replace the value for `CONSUL_K8S_CHART_LOCATION` with the path to the helm charts in consul-k8s on your system
4. run `start.sh`
5. once all the services are up exec into the bender service using:
```
kubectl exec -i -t $(kubectl get pods --selector=app=bender -o jsonpath='{.items[*].metadata.name}') -c bender -- /bin/sh
```
6. install `curl` onto the container using `apk add curl`
7. contact the zoidberg service fronted by the termgw using `curl zoidberg.virtual.consul`

Read the code, follow the above steps

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.changelog/3928.txt
+++ b/.changelog/3928.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+terminating-gateways: Remove unnecessary permissions from terminating gateways role
+```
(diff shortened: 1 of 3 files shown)


NiniOak (approved):
Tests pass now, thanks for the testing notes

nathancoleman (changes requested):
It looks like there is a bats test failing for the terminating gateway role, probably asserting that this `Secret` access is there

jm96441n:
@nathancoleman bats tests should be all green now

nathancoleman (approved):

PR #3957, opened 2024-04-29
NET-6704 Modify-server-acl-init-role

sarahalsmiller:
- Update server-acl-init-role to use 'core' instead of '""'

### Checklist ###
- [ ] Tests added


jm96441n:
overall LGTM but I think we should probably add a changelog here

PR #3959, opened 2024-05-01
[NET-6464] lock down init container

sarahalsmiller:
- Lock down security context in init container and remove open shift gating

### Checklist ###
- [ ] Tests added


jm96441n:
LGTM, but similar here I think we should probably have a changelog entry for this

PR #3974, opened 2024-05-07
activate tproxy mode even when a cluster IP is not assigned to pod

dhiaayachi:
Consul service should be create with mode transparent proxy even when a cluster IP is not assigned to the service.

Manually tested using an EKS cluster

NA


blake (approved):

jmurret (approved):
👏 thank you!

PR #3978, opened 2024-05-08
NET-9178-Consul-api-gateway-not-starting-after-restart

sarahalsmiller:
- If role already exists, don't error out of connect inject. 

- Local build, create api gateway, delete connect inject. Before this change, connect inject would crash with 'role already exists', after this change, connect inject ignores that the role already exists and continues to operate as expected. 

- Local Build, CI passes

### Checklist ###
- [ ] Tests added


NiniOak (approved):

NiniOak:
Followed the testing steps, looks good! Thanks for the fix 🚀

Issue #3831, opened 2024-03-28
ArgoCD hook annotations on the server-acl-init-job do not work


0011blindmice:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

While installing the Consul chart via ArgoCD, there are continuous Out-of-Sync warnings due to the init job being cleaned by a separate hook, leaving ArgoCD under the assumption that there is drift from the intended manifest.

There is currently a values entry to add ArgoCD's specific annotations for cleaning up hooks post-success, but they are being applied at the pod level of the spec.

i.e. https://github.com/hashicorp/consul-k8s/blob/caee250283e2d9e88f194f61da72243ba1fa8203/charts/consul/templates/server-acl-init-job.yaml#L50

It is my understanding this block of the template should actually be in the root `.metadata.annotations` of the job.

### Reproduction Steps

<!--- 

In order to effectively and quickly resolve the issue, please provide exact steps that allow us the reproduce the problem. If no steps are provided, then it will likely take longer to get the issue resolved. An example that you can follow is provided below. 

Steps to reproduce this issue, eg:

1. When running helm install with the following `values.yml`:
```
global:
  argocd:
    enabled: true
```

### Logs

![image](https://github.com/hashicorp/consul-k8s/assets/4370189/33e15d83-c3f3-42f2-8091-368203df5c39)

### Expected behavior

Ideally, the hook annotation would be at the Job's top level, as that is the resource that appears to be getting destroyed and triggering the sync disparity.

### Environment details
Consul version: 1.4.0
ArgoCD: 2.10.1

gcaracuel:
Pushed a fix for this on hashicorp/consul-k8s#3989 but do not looks like this is getting maintainers attention.

PR #3989, opened 2024-05-13
Fix ArgoCD hooks annotation location

gcaracuel:

ArgoCD hooks to be set to Job annotations and not at Job template spec level -> https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/#usage 

If set at current Job template spec level when Job is finished ArgoCD detects back again diff entering into a loop, if ArgoCD AutoSync is enabled this will be a loop of init jobs running forever.

fix: #3831 

* Ran helm bats tests successfully.

* Deploy this chart using ArgoCD Application. Without the fix after the sync it will enter in a diff sync loop. 

### Checklist ###
- [x] Tests added (Not needed)

### DIFF
--- a/.changelog/3989.txt
+++ b/.changelog/3989.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+helm: Fix ArgoCD hooks related annotation on init Job, they must be added at Job definition and not tempalte level.
+```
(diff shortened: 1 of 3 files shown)


gcaracuel:
Am I missing anything for this to be reviewed?

.changelog/3989.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+helm: Fix ArgoCD hooks related annotation on init Job, they must be added at Job definition and not tempalte level.
```
ndhanushkodi, line 2:
```suggestion
helm: Fix ArgoCD hooks related annotations on server-acl-init Job, they must be added at Job definition and not template level.
```

ndhanushkodi (commented):
Thanks for fixing this! Waiting for a test run and then I can officially approve. If in the meanwhile you don't mind making these small changes to the changelog that'd be great!

ndhanushkodi (approved):

Issue #1197, opened 2022-04-29
Expose imagePullPolicy


washtubs:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

#### Is your feature request related to a problem? Please describe.

We are deploying this as a subchart, and trying to prevent forking. It is possible but very difficult to modify imagePullPolicy with kustomize, and it would be nice if this was simply exposed in the values.yaml like it is for vault. 

#### Feature Description

Expose an imagePullPolicy attribute. A global attribute would be acceptable. Doesn't have to be one for the normal image and another for imageK8s, but I'll just point out that that's what vault does.

#### Use Case(s)

Changing the imagePullPolicy to Always in cases where we override the image to something with a fixed tag where the underlying image is updated in the registry. Not a great practice but still valid, and I'm sure there are plenty of other valid use cases.

#### Contributions

I would be happy to contribute.

david-yu:
Hi @washtubs that is a good feature request and I have also noticed that there is a imagePullPolicy with the vault Helm chart. Would you be willing to open up a PR that proposes how we could address this issue?

david-yu:
Re-opening as the changes had to be reverted prior to our GA release.

PR #3991, opened 2024-05-14
Adds ability to set the imagePullPolicy for all Consul images (consul…

missylbytes:
…, consul-dataplane, consul-k8s, consul-telemetry-collector)

- Adds the ability to set the imagePullPolicy globally for consul, consul-dataplane, consul-k8s, and consul-telemetry-collector

- Updated BATS
- Helm install

### Checklist ###
- [x] Tests added, but need to decide if bats tests need to be implemented everywhere the function is used 


david-yu:
Fixes https://github.com/hashicorp/consul-k8s/issues/1197

missylbytes:
- [ ] Need to add test for gateway deployment
- [ ] Need to add tests for any other containers not made with templates

charts/consul/templates/_helpers.tpl
```diff
@@ -632,7 +633,9 @@ Usage: {{ template "consul.dogstatsdAaddressInfo" . }}
 
 {{- define "consul.dogstatsdAaddressInfo" -}}
 {{- if (and .Values.global.metrics.datadog.enabled .Values.global.metrics.datadog.dogstatsd.enabled) }}
-        "dogstatsd_addr": "{{- if eq .Values.global.metrics.datadog.dogstatsd.socketTransportType "UDS" }}unix://{{ .Values.global.metrics.datadog.dogstatsd.dogstatsdAddr }}{{- else }}{{ .Values.global.metrics.datadog.dogstatsd.dogstatsdAddr | trimAll "\"" }}{{- if ne ( .Values.global.metrics.datadog.dogstatsd.dogstatsdPort | int ) 0 }}:{{ .Values.global.metrics.datadog.dogstatsd.dogstatsdPort | toString }}{{- end }}{{- end }}",{{- end }}
```
missylbytes, line 635: fix this, I think this is why everything is failing

missylbytes: fixed

charts/consul/templates/_helpers.tpl
```diff
@@ -682,4 +683,22 @@ Usage: {{ template "consul.versionInfo" }}
...
+    IfNotPresent
+    Always
+    Never
+    In the case of empty, see https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy for details
+
+Usage: {{ template "consul.imagePullPolicy" . }} TODO: melisa should we name this differently ?
+*/}}
+{{- define "consul.imagePullPolicy" -}}
+{{ if or (eq .Values.global.imagePullPolicy "IfNotPresent") (eq .Values.global.imagePullPolicy "Always") (eq .Values.global.imagePullPolicy "Never")}}imagePullPolicy: {{ .Values.global.imagePullPolicy }}
+{{ else if eq .Values.global.imagePullPolicy "" }}
```
jm96441n, line 700: do we want a default value for this?

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -97,9 +97,10 @@ func (w *MeshWebhook) consulDataplaneSidecar(namespace corev1.Namespace, pod cor
 	}
 
 	container := corev1.Container{
-		Name:      containerName,
-		Image:     w.ImageConsulDataplane,
-		Resources: resources,
+		Name:            containerName,
+		Image:           w.ImageConsulDataplane,
+		ImagePullPolicy: corev1.PullPolicy(w.GlobalImagePullPolicy),
```
jm96441n, line 102: is the idea that there's a global imagePullPolicy and that each stanza should be able to overwrite that with their own pull policy?

charts/consul/templates/_helpers.tpl
missylbytes: Yeah I was trying to decide, because kubernetes does its own thing with pull policies, and there isn't just one default (https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting )

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
missylbytes: Good question, I think the idea was to only have a global pull policy. I could see the usefulness of overwrite, but I don't think we have any plans to do that right now.

charts/consul/templates/_helpers.tpl
missylbytes: It does accept empty string though.

jm96441n (approved):
LGTM!

PR #4015, opened 2024-05-20
Prepare 1.3.5

missylbytes:
Prepare release for 1.3.5 (https://github.com/hashicorp/engineering-docs/blob/main/consul/consul-k8s/release-process.md#prepare-the-point-release-branch)

### Checklist ###
- [ ] Tests added


charts/consul/Chart.yaml
```diff
@@ -4,7 +4,7 @@
 apiVersion: v2
 name: consul
 version: 1.3.5
-appVersion: 1.17.5
+appVersion: 1.17.3
```
nathancoleman, line 7: This was a little confusing for me because there is a `v1.17.4` tag of hashicorp/consul; however, it seems this is because we didn't have our tag naming conventions in place at the time.

`v1.17.3` is the latest patch release available on homebrew and dockerhub, so I believe this is correct.

nathancoleman (approved):

PR #4016, opened 2024-05-21
security: update alpine base image to 3.19

zalimeni:
Resolves CVE-2024-0727.

### Changes proposed in this PR ###  
- Update `alpine:3.18` where used to `alpine:3.19` (currently inconsistent)

### How I've tested this PR ###
CI

### How I expect reviewers to test this PR ###
👀 

### Checklist ###
- [ ] Tests added
- [ ] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

PR #4022, opened 2024-05-21
chore: fix previous changelog for alpine upgrade

zalimeni:
The original change was made as a security fix, which was later determined to be unnecessary (both `alpine` 3.18 and 3.19 received the CVE fix in question). Recategorizing this as an improvement for the final changelog. The `alpine` upgrade is still worth doing, as is aligning all the base images in the Dockerfile on a single version.

I'll follow up w/ the same change on backports of https://github.com/hashicorp/consul-k8s/pull/4016.


jm96441n (approved):
minor comment: will this cause an update in the github release notes? should we update those as well (if we haven't yet)

zalimeni:
@jm96441n thanks for checking! I'll verify, but this shouldn't have made it into any patch release branches because they were already cut when it was opened - if I find one I'll update 👍🏻

PR #4028, opened 2024-05-21
Post release 1.3.5

missylbytes:
Puts release/1.3.x back into dev mode after 1.3.5 release

### DIFF
--- a/charts/consul/Chart.yaml
+++ b/charts/consul/Chart.yaml
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: consul
-version: 1.3.5-dev
+version: 1.3.6-dev
 appVersion: 1.17-dev
 kubeVersion: ">=1.22.0-0"
 description: Official HashiCorp Consul Chart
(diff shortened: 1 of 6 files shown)


charts/consul/values.yaml
```diff
@@ -86,7 +86,7 @@ global:
   # image that is used for functionality such as catalog sync.
   # This can be overridden per component.
   # @default: hashicorp/consul-k8s-control-plane:<latest version>
-  imageK8S: docker.mirror.hashicorp.services/hashicorppreview/consul-k8s-control-plane:1.3.5-dev
+  imageK8S: docker.mirror.hashicorp.services/hashicorppreview/consul-k8s-control-plane:1.3-dev
```
NiniOak, line 89: Should this be `consul-k8s-control-plane:1.3.6-dev`?

charts/consul/Chart.yaml
```diff
@@ -18,7 +18,7 @@ annotations:
     - name: consul
       image: docker.mirror.hashicorp.services/hashicorppreview/consul:1.17-dev
     - name: consul-k8s-control-plane
-      image: docker.mirror.hashicorp.services/hashicorppreview/consul-k8s-control-plane:1.3.5-dev
+      image: docker.mirror.hashicorp.services/hashicorppreview/consul-k8s-control-plane:1.3-dev
```
NiniOak, line 21: Should this be consul-k8s-control-plane:1.3.6-dev?

charts/consul/values.yaml
missylbytes: The 1.4.x branch uses just 1.4-dev. We publish both so I think that is what we would want to do? Just use the 1.3-dev

missylbytes: We publish both the 1.3-dev and 1.3.6-dev

NiniOak (approved):

PR #4040, opened 2024-05-23
build.yml: package consul-cni

david-yu:

- Only package `consul-cni` for amd64 and arm64 on linux 
- Added extra step to just package consul-cni, testing is only done for consul-k8s CLI
- For binary use `matrix.bin_name` instead of `matrix.pkg_name`
- Added skip_packaging to more artifacts to make things more explicit although likely not needed 

### Checklist ###
- [ ] Tests added


.github/workflows/build.yml
```diff
@@ -169,7 +169,7 @@ jobs:
           cp $GITHUB_WORKSPACE/LICENSE "$LICENSE_DIR/LICENSE.txt"
 
       - name: Package rpm and deb files
-        if: matrix.goos == 'linux' && matrix.component == 'cli' && matrix.skip_packaging != 'true'
+        if: matrix.goos == 'linux' && matrix.component != 'control-plane' && matrix.skip_packaging != 'true'
```
zalimeni, line 172: ~ Not a big deal, but here and below, I'd prefer we were explicit rather than inverting the match, in case other components are added in the future:
```yaml
contains(fromJSON('["cli", "control-plane/cni"]'), matrix.component)
```

That said, I don't think we need both checks at this point. I think the ideal thing might be to change `skip_packaging` to `package_linux` (opt-in), and set it for the matrix combos we want to package - or add `skip_packaging` to the `linux` `control-plane` values. Either way, at that point we could just remove the name match entirely, and have a clear value in the matrix that says whether a linux build is packaged.

@shore @david-yu WDYT?

zalimeni (approved):
Approving to unblock but have one comment on the matrix checks. Would prefer to get a 👍🏻 from RelEng as well in case I'm missing something.

.github/workflows/build.yml
david-yu: I can just add skip packaging manually to everything that needs to be skipped that might be easier.

.github/workflows/build.yml
```diff
@@ -179,25 +178,43 @@ jobs:
...
+          binary: "${{ matrix.component }}/dist/${{ matrix.bin_name }}"
           deb_depends: "openssl"
           rpm_depends: "openssl"
           config_dir: ".release/linux/package/"
 
+      - name: Package rpm and deb files for consul-cni plugin
+        if: matrix.goos == 'linux' && matrix.component == 'control-plane/cni' && matrix.skip_packaging != 'true'
+        uses: hashicorp/actions-packaging-linux@v1
+        with:
+          name: consul-cni{{ matrix.pkg_suffix }}
```
shore, line 190: Typo
```suggestion
          name: consul-cni${{ matrix.pkg_suffix }}
```

.github/workflows/build.yml
```diff
@@ -74,37 +74,36 @@ jobs:
...
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "linux", goarch: "arm64", component: "control-plane", pkg_name: "consul-k8s-control-plane", "bin_name": "consul-k8s-control-plane", gotags: "fips", env: "CGO_ENABLED=1 GOEXPERIMENT=boringcrypto CC=aarch64-linux-gnu-gcc", fips: "+fips1402", pkg_suffix: "-fips", "skip_packaging": "true" }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "windows", goarch: "amd64", component: "control-plane", pkg_name: "consul-k8s-control-plane", "bin_name": "consul-k8s-control-plane.exe", gotags: "fips", env: "CGO_ENABLED=1 GOEXPERIMENT=cngcrypto", fips: "+fips1402", "skip_packaging": "true"  }
 
           # consul-cni
-          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "freebsd", goarch: "386", component: "control-plane/cni", pkg_name: "consul-cni", "bin_name": "consul-cni" }
```
shore, line 94: Is this removal intended?  There are other freebsd/386 builds still in the matrix.

.github/workflows/build.yml
```diff
@@ -74,37 +74,36 @@ jobs:
...
           - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "freebsd", goarch: "amd64", component: "control-plane/cni", pkg_name: "consul-cni", "bin_name": "consul-cni" }
```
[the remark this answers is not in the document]
david-yu: That works.

.github/workflows/build.yml
david-yu: Yes the CNI plugin cannot run on freebsd as Blake noted. You can still use consul-k8s on freebsd other use cases that don't require the CNI plugin.

shore (approved):
👍

.github/workflows/build.yml
```diff
@@ -74,37 +74,35 @@ jobs:
...
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "solaris", goarch: "amd64", component: "control-plane/cni", pkg_name: "consul-cni", "bin_name": "consul-cni", "skip_packaging": "true"  }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "windows", goarch: "386", component: "control-plane/cni", pkg_name: "consul-cni", "bin_name": "consul-cni.exe", "skip_packaging": "true"  }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "windows", goarch: "amd64", component: "control-plane/cni", pkg_name: "consul-cni", "bin_name": "consul-cni.exe", "skip_packaging": "true"  }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "amd64", component: "control-plane/cni", pkg_name: "consul-cni", "bin_name": "consul-cni", "skip_packaging": "true"  }
+          - {go: "${{ needs.get-go-version.outputs.go-version }}", goos: "darwin", goarch: "arm64", component: "control-plane/cni", pkg_name: "consul-cni", "bin_name": "consul-cni", "skip_packaging": "true"  }
```
blake, line 102:
```suggestion
```

blake (changes requested):

.github/workflows/build.yml
david-yu: Done

blake (approved):

PR #4054, opened 2024-05-29
build: prepare for 1.6.0 dev

zalimeni:


NiniOak (approved):

control-plane/version/version.go
```diff
@@ -17,7 +17,7 @@ var (
 	//
 	// Version must conform to the format expected by
 	// github.com/hashicorp/go-version for tests to work.
-	Version = "1.5.0"
+	Version = "1.6.0"
```
tgross, line 20: @zalimeni just a heads up that I noticed this morning that the 1.4.x versions of `consul-cni` after 1.4.2 when this was introduced are all showing `1.5.0-dev` rather than the correct version. Not sure if that's something that should be addressed in this PR or is a missing step in the release checklist.

zalimeni: @tgross thanks for the heads up - I'm guessing a step was accidentally skipped when creating those patch releases. Having a look shortly, but it shouldn't impact this PR as `main` is ahead of those.

❓ just TBC, are you referring to released artifacts, or local builds from `release/1.4.x`? If the former, we may want to add a note to the changelog(s) that the version is incorrectly reported (I thought we had checks for this in CRT but perhaps there's a gap somewhere).

tgross: The ones downloaded from https://releases.hashicorp.com/consul-cni, unfortunately:

```
$ curl -sO https://releases.hashicorp.com/consul-cni/1.4.2/consul-cni_1.4.2_linux_amd64.zip

$ unzip consul-cni_1.4.2_linux_amd64.zip
Archive:  consul-cni_1.4.2_linux_amd64.zip
  inflating: consul-cni
  inflating: LICENSE.txt

$ ./consul-cni --version
CNI consul-cni plugin v1.5.0-dev
CNI protocol versions supported: 0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 1.0.0
```

I can file an issue for this somewhere if you'd like, but I noticed this PR while looking for an existing report.

zalimeni: Thanks @tgross - appreciate you noticing and reporting. If you want to drop a message in the Consul team Slack for 👀  and tag me, that would be great, and we can decide how to follow up from there 🙏🏻

flyinprogrammer: 1.4.3 was still released as 1.5.0-dev:

```shell
curl -L -o consul-cni.zip https://releases.hashicorp.com/consul-cni/1.4.3/consul-cni_1.4.3_linux_amd64.zip
unzip consul-cni.zip
./consul-cni --version
# CNI consul-cni plugin v1.5.0-dev
# CNI protocol versions supported: 0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 1.0.0
```

@zalimeni have we validated this is fixed for 1.6.0 ?

( @tgross 👋 miss yah! )

zalimeni: @flyinprogrammer thank you for the heads up! Unfortunately this was caught just as we were releasing 1.4.3, so we weren't able to address for that release, but are hoping to have this fixed in the next set of releases.

(For posterity: the issue was that we backported some changes to the `cni` submodule prior to `1.4.2` that instigated a bump to the [pinned](https://github.com/hashicorp/consul-k8s/pull/3829/files#diff-997b0ccf22eb73072c3ef0aa95ff188068f09d639ae34617b7fec113c5030429R6) `control-plane` submodule introduced recently to consume the common version string; that new SHA had 1.5.0-dev in the `version.go` file. I think the best approach to avoid this repeating in the future will be to move `consul-k8s` binaries to the `version/VERSION` embedded file scheme that `consul` uses, or introduce a `replace` for `control-plane` in `cni`'s `go.mod` to ensure we're always using the current revision's `version.go`.)

tgross: @flyinprogrammer 
> ( @tgross 👋 miss yah! )

Same!

zalimeni: Closing the loop: https://github.com/hashicorp/consul-k8s/pull/4091 this should address the versioning issues going forward. Thanks again for the report @tgross !

PR #4059, opened 2024-05-29
fix: deregistration should not do graceful shutdown if there is a new pod

ndhanushkodi:
The existing graceful shutdown logic checks for the pod with the same name and looks at its graceful shutdown time to mark the service instance as unhealthy in consul and delay its deletion. In the case of a statefulset rollout, a pod will go down and a new pod with the same name will come up, so we could run into a race condition where we try to mark the service instance as unhealthy thinking that the pod is terminating, but it's actually the new running pod, not the old terminating pod.

This change is to check whether the service instance we're proposing to deregister in consul should go through graceful shutdown or not. If the pod has a different UID from the pod UID used to register that service instance, then the service instance should be deregistered. 

- Unit tests
- I was not able to reproduce the race condition locally.

👀 

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
@@ -1047,10 +1047,11 @@ func (r *Controller) getGracefulShutdownAndUpdatePodCheck(ctx context.Context, a
 		return 0, fmt.Errorf("failed to get terminating pod %s/%s: %w", k8sNamespace, podName, err)
 	}
 
-	// In a statefulset rollout, pods can go down and a new pod will come back up with the same name but a different
-	// address. In that case, it's not the old pod that is gracefully shutting down; the old pod is gone and we
-	// should deregister that old instance from Consul.
-	if string(pod.UID) != svc.ServiceMeta[constants.MetaKeyPodUID] {
+	// In a statefulset rollout, pods can go down and a new pod will come back up with the same name but a different uid
+	// or node name. In older consul-k8s patches, the pod uid may not be set, so to account for that we also check if
+	// the node changed, since newer service instances should exist for the new node. In that case, it's not the old pod
+	// that is gracefully shutting down; the old pod is gone and we should deregister that old instance from Consul.
+	if string(pod.UID) != svc.ServiceMeta[constants.MetaKeyPodUID] || common.ConsulNodeNameFromK8sNode(pod.Spec.NodeName) != svc.Node {
 		return 0, nil
 	}
 
@@ -1064,7 +1065,7 @@ func (r *Controller) getGracefulShutdownAndUpdatePodCheck(ctx context.Context, a
 		// Update the health status of the service to critical so that we can drain inbound traffic.
 		// We don't need to handle the proxy service since that will be reconciled looping through all the service instances.
 		serviceRegistration := &api.CatalogRegistration{
-			Node:    common.ConsulNodeNameFromK8sNode(pod.Spec.NodeName),
+			Node:    svc.Node,
 			Address: pod.Status.HostIP,
 			// Service is nil since we are patching the health status
 			Check: &api.AgentCheck{
--- a/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
+++ b/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
@@ -1047,6 +1047,13 @@ func (r *Controller) getGracefulShutdownAndUpdatePodCheck(ctx context.Context, a
 		return 0, fmt.Errorf("failed to get terminating pod %s/%s: %w", k8sNamespace, podName, err)
 	}
 
+	// In a statefulset rollout, pods can go down and a new pod will come back up with the same name but a different
+	// address. In that case, it's not the old pod that is gracefully shutting down; the old pod is gone and we
+	// should deregister that old instance from Consul.
+	if string(pod.UID) != svc.ServiceMeta[constants.MetaKeyPodUID] {
+		return 0, nil
+	}
+
 	shutdownSeconds, err := r.getGracefulShutdownPeriodSecondsForPod(pod)
 	if err != nil {
 		r.Log.Error(err, "failed to get graceful shutdown period for pod", "name", pod, "k8sNamespace", k8sNamespace)
--- a/.changelog/4059.txt
+++ b/.changelog/4059.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+endpoints-controller: graceful shutdown logic should not run on a new pod with the same name. Fixes a case where statefulset rollouts could get stuck in graceful shutdown when the new pods come up.
+```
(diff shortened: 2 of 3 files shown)


control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -1047,6 +1047,13 @@ func (r *Controller) getGracefulShutdownAndUpdatePodCheck(ctx context.Context, a
 		return 0, fmt.Errorf("failed to get terminating pod %s/%s: %w", k8sNamespace, podName, err)
 	}
 
+	// In a statefulset rollout, pods can go down and a new pod will come back up with the same name but a different
+	// address. In that case, it's not the old pod that is gracefully shutting down; the old pod is gone and we
+	// should deregister that old instance from Consul.
+	if string(pod.UID) != svc.ServiceMeta[constants.MetaKeyPodUID] {
```
dhiaayachi, line 1053: I'm not sure I understand fully 🤔 
When the pod UID is different from the ServiceMeta stored in Consul, that means the Consul service instance has been updated with the new Pod information (Pod UID, address...) but based on this change we will deregister that service instance from Consul which means that the new Endpoint (associated with the new Pod) will no longer have a service instance associated with it in Consul.

This change will break the graceful shutdown cycle but I think it will make things worst for the user as this pod won't be addressable in the mesh anymore.

I think a better approach would be to just let the service instance in Consul and remove it from the `deregisterEndpointAddress` map so we won't try to do a graceful shutdown again.

dhiaayachi (commented):
👋 @ndhanushkodi 
Thank you for working on this! I left a comment about the approach, I'm not sure this resolve the issue as we saw it in the logs the other day but maybe I' missing something 😅. Let's chat about it!

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
DanStough: So we are theory for the existing failure mode is:
1. SS scales down. 
   1. We go into a reconcile loop where the endpoint is missing and ends up in a graceful shutdown backoff.
   1. The health status of the Consul service goes into critical.
1. SS scaled back up.
   1. We re-register the endpoint in Consul with the same ServiceID, because if it's in the endpoint list now, we always do a registration.
      1. I have some suspicions here that we are not fully overwriting the service registration with the new IP and/or health status.
   1. As reported, the Consul service somehow remains with a Critical status in the catalog.
      1. **Theory:** When we loop through the existing Consul services, we compare it with the IP-based map and go through graceful shutdown loop again. The old service instance isn't in the IP-based map, **_assuming_** the address changed, and it wasn't updated with the new service registration.
         1. This re-marks the service as critical, since this bit of the code uses the ServiceID, which is the same between the new and the old pod.
         1.  We also requeue again.
1. If we get a requeued backoff request.
   1. We get the endpoint resource, which probably looks very close to the the previous case when the SS scaled back up. So we end up with the service still in critical.

I guess I'm wondering how we end up back in this loop. Because if it's diverging IP addresses we would also need to fix that to get mesh working again.

As an aside, I'm realizing that as long as the pod is alive we never force the deregistration, even after the graceful period expires. The reason was that we don't necessarily know how much time has elapsed with each reconcile. Maybe this is fine, but then it doesn't really matter what you set the shutdown period to be since is will always be requeued. 😦

ndhanushkodi:
@DanStough @dhiaayachi re-requesting your review on this now that we know this should actually solve their problem

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
dhiaayachi: @DanStough 

the scenario that trigger the issue is as follow:
- A stateful set with, let's say, 2 pods is running
- Pod-0 get killed
- Kubernetes recreate the pod with the same name but in a different node and obviously with a different pod ip, let's called new-pod-0
- an event is queued to the endpoint_controller
- the endpoint controller read the endpoints and register all the service instances in the endpoints array (including the new endpoint associated with new-pod-0 
- The controller read the service instances from Consul and realize that there is a service instance associated with an unknown IP (old Pod-0), it mark it to be deleted
- graceful shutdown logic is ran and try to see if any pod is running with pod-0 name, it find new-pod-0
- decide to set the service instance health to critical (but set the wrong node, the one associated with new-pod-0) and re-queue the event after a while (the configured graceful shutdown time)
- This happen again and again, as new-pod-0 is not shutting down
- 
> As an aside, I'm realizing that as long as the pod is alive we never force the de-registration, even after the graceful period expires. The reason was that we don't necessarily know how much time has elapsed with each reconcile. Maybe this is fine, but then it doesn't really matter what you set the shutdown period to be since is will always be requeued. 😦

My understanding of the config is that it's not how long before we force shutdown but how long we need to recheck after as this config need to match the lifecycle config in kube that will force the pod to be killed.

dhiaayachi, line 1053: I'm wondering if there is any case where the `pod.UID` or `svc.ServiceMeta[constants.MetaKeyPodUID]` are not populated and if we should account for that? 🤔

dhiaayachi (approved):
Thank you for working on this @ndhanushkodi! I had a question, but nothing blocking.

Also I'm wondering if it will be more correct in:

https://github.com/hashicorp/consul-k8s/blob/f55172adcc4cc1d5ebfbc70d8d537517301f6272/control-plane/connect-inject/controllers/endpoints/endpoints_controller.go#L1067-L1068

to use `svc.Node` instead of ` common.ConsulNodeNameFromK8sNode(pod.Spec.NodeName)`

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
DanStough: >>I'm realizing that as long as the pod is alive we never force the deregistration, even after the graceful period expires.
>
>My understanding was the graceful shutdown is an endpoint in the proxy, and it gets called on a SIGTERM to the pod. Then after the countdown time passes, then the pod continues to finish shutting down. Maybe what you're saying is if the pod gets "stuck" alive somehow, we wouldn't deregister it?

I totally forgot that there is a [timeout in dataplane](https://github.com/hashicorp/consul-dataplane/blob/8455ba719f2eaa1c7309fd88e1cec5b9ea5d3f71/pkg/consuldp/lifecycle.go#L163-L165) that should trigger the pod to stop.  In the case it gets "stuck", the Kube graceful shutdown period expiring is probably sufficient to clean these up.

ndhanushkodi: Great call, I looked back to when we added pod.UID to service registrations and it was semi recent: https://github.com/hashicorp/consul-k8s/commit/1b6f700f17d8bc7976b6c04d92fbf2def7830e43 so older patch versions might not have it. 

I updated the check to check for either: pod uid changing, or the node name changing. So that way an upgrade from an older patch would detect node name changing even if pod uid is not yet there.

ndhanushkodi:
> to use svc.Node instead of  common.ConsulNodeNameFromK8sNode(pod.Spec.NodeName)

this also makes sense to me, since we're trying to update an existing registration. I think that would also have solved the issue of marking the wrong instance as unhealthy.

dhiaayachi (approved):
🚀

PR #4060, opened 2024-05-30
[NET-9420] clean up acl role policy for apigw

jm96441n:
previously all gateways shared the same acl role and policy so when deleting a gateway the policy/role did not need to be removed. With the changes to reduce the scope of gateway acls and make them a 1:1 mapping from role/policy to gateway we now need to clean up the role/policy when a gateway is removed.

- Wrote tests

Repro Steps:
1. on this branch run `make dev-docker` and for consul pull latest `main` and run `make dev-docker` to generate the images we'll be testing with
1. clone the following gist: https://gist.github.com/jm96441n/b92a80c2251c7a54d1feada320c5eac6
1. in the `start.sh` file of the gist update the `CONSUL_K8S_CHART_LOCATION` variable on line 5 to point to the location of the charts directory in consul-k8s on your machine
1. run the `start.sh` script, this creates your kind cluster, loads the consul-k8s and consul images into the cluster, installs consul, and sets up 2 api-gateways
1. run `kubectl port-forward service/consul-consul-ui 8501:443 -n consul` in a separate terminal window to port forward to consul
1. to copy the acl token needed to login run 
```sh
# macos
kubectl get --namespace consul secrets/consul-consul-bootstrap-acl-token --template={{.data.token}} | base64 -d | pbcopy

# linux
kubectl get --namespace consul secrets/consul-consul-bootstrap-acl-token --template={{.data.token}} | base64 -d | xsel -clipboard
```
7. open your browser and visit https://localhost:8501 and navigate through the untrusted cert warnings
8. login using the login token we copied previously
9. view the `policies` page and you should see an entry for each of the api-gateways
10. remove one of the gateways by running `kubectl delete -f gw2.yaml` and see the policy get removed (the same will happen for roles)
11. put the gateway back by running `kubectl apply -f gw2.yaml` and see the policy come back

read the code
run the above steps

To run the acceptance test:
1. run `kind create cluster`
2. build the control plane from this branch using `make dev-docker`
3. load the image into the cluster by running `kind load docker-image consul-k8s-control-plane:local`
4. cd into the `api-gateway` directory `cd ./acceptance/tests/api-gateway`
5. run `go test -consul-k8s-image=consul-k8s-control-plane:local -run TestAPIGateway_Lifecycle -v ./...` to run the test using the changes from this branch

### Checklist ###
- [X] Tests added


acceptance/framework/environment/environment.go
```diff
@@ -187,6 +191,7 @@ func (k kubernetesContext) ControllerRuntimeClient(t testutil.TestingTB) client.
 
 	client, err := client.New(config, client.Options{Scheme: s})
 	require.NoError(t, err)
+	logf.SetLogger(zap.New(zap.WriteTo(os.Stdout), zap.UseDevMode(true)))
```
jm96441n, line 194: need to rework this to not use zap, we're hitting the same issue listed here from the controller-runtime upgrade https://github.com/solo-io/gloo/issues/8548

sarahalsmiller (approved):
Thanks for grabbing that

PR #4061, opened 2024-05-30
remove catalogv2 implementation and tests

rboyer:

- Remove all direct plumbing related to the `resource-apis` and `v2tenancy` experiments
- Remove all remaining dead code pertaining to the remainder of Consul resources _(all experimental)_
- Remove all tests of the above

https://hashicorp.atlassian.net/browse/NET-9089


rboyer:
Pretty sure now I have had all of the flavors of acceptance test pass at least once, so I probably didn't over-delete anything this time.

Makefile
```diff
@@ -97,11 +99,32 @@ control-plane-fips-dev-docker: ## Build consul-k8s-control-plane FIPS dev Docker
 
 .PHONY: control-plane-test
 control-plane-test: ## Run go test for the control plane.
-	cd control-plane; go test ./...
+ifeq ("$(GOTESTSUM_PATH)","")
```
rboyer, line 102: I cribbed this conditional gotestsum running approach from the consul main repo to help when I was running and re-running the normal tests.

DanStough (changes requested):
My only blocking comment is deleting the rest of the v2 Namespace controller that uses the V1 APIs.

DanStough (approved):
:shipit: 🚀

zalimeni:
Removing labels for now but wondering whether we should backport this change to 1.4-1.6 to avoid discrepancies in backports.

PR #4066, opened 2024-06-03
[NET-9932] Always add NET_BIND_SERVICE capability to injected sidecar container

nathancoleman:
Consul-dataplane always requires the `NET_BIND_SERVICE` capability ([docs](https://developer.hashicorp.com/consul/docs/connect/dataplane#technical-constraints)). This is explicitly allowed under the `restricted-v2` `SecurityContextConstraints` (SCC) that ships with OpenShift 4.11+; however, the container needs to _request_ the capability in order for it to be granted. This change does that.

Worked w/ @natemollica-nm to verify that this fix allows the application sidecar running the `consul-dataplane` image to spin up successfully under the `restricted-v2` SCC where it would have previously logged the following error on startup.
```log
Defaulted container "consul-dataplane" out of: consul-dataplane, backend, consul-connect-inject-init (init)
[dumb-init] /usr/local/bin/consul-dataplane: Operation not permitted
```

Application sidecars should now be fully functional on OpenShift without creating a custom SCC.
The `Pod` containing the injected sidecar should get an annotation from OpenShift indicating that it's using the `restricted-v2` SCC.

See above

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.changelog/4066.txt
+++ b/.changelog/4066.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+connect-inject: add NET_BIND_SERVICE capability when injecting consul-dataplane sidecar
+```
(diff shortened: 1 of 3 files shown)


nathancoleman:
Note to self: This PR **cannot** be backported to any release that requires running Consul w/ the `anyuid` `SecurityContextConstraints`. If we do, then we get the following error due to the fact that `anyuid` does not allow the `NET_BIND_SERVICE` capability.



```shell
pods "backend-6c4cf7dc55-" is forbidden: unable to validate against any security context constraint:
| [provider anyuid: .containers[0].capabilities.add:                                                                                                                                                
│       Invalid value: "NET_BIND_SERVICE": capability may not be added
```

nathancoleman:
This change happened in https://github.com/hashicorp/consul-k8s/pull/4152

PR #4067, opened 2024-06-03
[NET-9500] Cleanup orphaned inline-certs and acl role/policy

jm96441n:
- Add a goroutine on start up that loops and tries to clean up inline certs that are no longer  referenced by any gateway and the original shared acl role/policy that is now specific to each gateway. Once all those are cleaned up the goroutine will exit and stop polling

Run/Write tests

#### Manual Testing:
1. clone the following gist: https://gist.github.com/jm96441n/b92a80c2251c7a54d1feada320c5eac6
2. run the start.sh script in the gist (this will install consul 1.18-ent so ensure you have an ent license setup)
3. port forward port `8501` to the consul server instance
3. once the cluster is up run the following in another terminal:
```
export CONSUL_HTTP_ADDR="127.0.0.1:8501"
export CONSUL_HTTP_TOKEN=$(kubectl get --namespace consul secrets/consul-consul-bootstrap-acl-token --template={{.data.token}} | base64 -d)
export CONSUL_HTTP_SSL=true
export CONSUL_HTTP_SSL_VERIFY=false
```
5. run `consul acl role list` and you will see the shared `"managed-gateway-acl-role"`
5. run `consul acl policy list` and you will see the shared `"api-gateway-token-policy"`
6. run `consul acl binding-rule list` and you will see the shared binding-rules referencing both `managed-gateway-acl-role`
7. run `consul config list -kind inline-certificate` and you will see an inline certificate listed
8. in the `consul_values.yaml` uncomment lines 3 and 4 and comment lines 4 and 5 (also make sure you build consul-k8s from this branch and have an up to date build of main of consul)
9. in the same terminal run `export CONSUL_K8S_CHARTS_LOCATION="$HOME/hashi/consul-k8s/charts/consul"` and replace the value being set with the location of the helm charts in your local copy of consul-k8s
10. run the following `helm upgrade --install consul $CONSUL_K8S_CHARTS_LOCATION -f ./consul_values.yaml -n consul --create-namespace --wait` to install the new version of consul-k8s that you built
11. force a re-reconciliation of your gateways (this should happen by default due to the change in dataplane version, if it does not you can delete and recreate the gateways)
12. run `consul acl role list` and you will not see the shared `"managed-gateway-acl-role"`
13. run `consul acl policy list` and you will not see the shared `"api-gateway-token-policy"`
14. . run `consul acl binding-rule list` and you will not see the shared binding-rules referencing `managed-gateway-acl-role`
15. run `consul config list -kind inline-certificate` and you will see no inline certificates listed

read the code
run the tests
do the above steps

### Checklist ###
- [ ] Tests added


control-plane/api-gateway/binding/cleanup.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		if err.Error() == "Unexpected response code: 401 (ACL support disabled)" {
+			return true, nil
+		}
+		return false, fmt.Errorf("failed to list binding rules: %w", err)
+	}
+
+	oldBindingRules := make(map[string]*api.ACLBindingRule)
+
+	// here we need to find binding rules with the old name that have a matching selector to the new gateway specific binding rule
```
jm96441n, line 74: so another direction we could've gone down here is to fetch all the gateways and then for each gateway match it to a binding rule and if we match each gateway to a new binding rule then we can safely delete all the binding rules, definitely open to push back to change this to utilize that heuristic for deleting binding rules rather than what we have here (and we already fetch all the gateways for the inline cert cleanup so it's not like we don't need that data anyways)

NiniOak (commented):
Code looks fine but the test steps don't seem correct. Happy to re-review it with you before approval

jm96441n:
> Code looks fine but the test steps don't seem correct. Happy to re-review it with you before approval

repro steps should be fixed now! sorry about that!

control-plane/api-gateway/binding/cleanup.go
```diff
@@ -0,0 +1,198 @@
...
+
+		if aclsCleanedUp && inlineCertsAllCleanedUp {
+			return
+		}
+	}
+}
+
+// cleanupACLRoleAndPolicy deletes the old shared gateway ACL role and policy if they exist.
+func (c Cleaner) cleanupACLRoleAndPolicy(client *api.Client) (bool, error) {
+	existingRules, _, err := client.ACL().BindingRuleList(c.AuthMethod, &api.QueryOptions{})
```
nathancoleman, line 64: Does this list for all namespaces?

nathancoleman: Discussed on Slack that the gateway approach described below is probably more scalable in all use cases and would query directly whatever namespace the gateway resided in.

control-plane/api-gateway/binding/cleanup.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/hashicorp/consul-k8s/control-plane/consul"
+	"github.com/hashicorp/consul/api"
+)
+
+const (
+	oldACLRoleName   = "managed-gateway-acl-role"
+	oldACLPolicyName = "api-gateway-token-policy"
+)
+
+var sleepTime = 60 * time.Second
```
nathancoleman, line 21: Pondering whether 60s here makes sense or if we should go with something like 10m intervals given that it's really just a cleanup job handling a long-term concern that isn't worth knocking on consul-server's door every 60s

control-plane/api-gateway/binding/cleanup.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+
+	certsToDelete := certSet.Difference(certsToKeep)
+	var mErr error
+	deletedCerts := 0
+	for cert := range certsToDelete.Iter() {
+		_, err := client.ConfigEntries().Delete(api.InlineCertificate, cert, &api.WriteOptions{})
+		if err != nil {
+			mErr = errors.Join(mErr, fmt.Errorf("failed to delete the inline cert %s: %w", cert, err))
```
nathancoleman, line 180:
```suggestion
			mErr = errors.Join(mErr, fmt.Errorf("failed to delete inline-certificate %s: %w", cert, err))
```

control-plane/api-gateway/binding/cleanup_test.go
```diff
@@ -0,0 +1,355 @@
...
+		aclRole                          *api.ACLRole
+		expectedDeletedACLRoleIDs        mapset.Set[string]
+		aclPolicy                        *api.ACLPolicy
+		expectedDeletedACLPolicyIDs      mapset.Set[string]
+		inlineCerts                      []*api.InlineCertificateConfigEntry
+		expxectedDeletedCertsName        mapset.Set[string]
+		apiGateways                      []*api.APIGatewayConfigEntry
+	}{
+		// add binding rules that match on selector and name to be cleaned up
+		"all old roles/policies/bindingrules and inline certs ge cleaned up": {
```
nathancoleman, line 37:
```suggestion
		"all old roles/policies/bindingrules and inline certs get cleaned up": {
```

control-plane/api-gateway/binding/cleanup.go
```diff
@@ -0,0 +1,198 @@
...
+		case <-ctx.Done():
+			return
+		case <-time.After(sleepTime):
+		}
+
+		client, err := consul.NewClientFromConnMgr(c.ConsulConfig, c.ServerMgr)
+		if err != nil {
+			c.Logger.Error(err, "failed to create Consul client")
+			continue
+		}
```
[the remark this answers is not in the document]
jm96441n: we do this for the cache code as well, everytime the cache syncs we re-create the client so I wanted to mirror that functionality

control-plane/api-gateway/binding/cleanup.go
jm96441n: we also always write the binding-rules into the default namespace and partition https://github.com/hashicorp/consul-k8s/blob/main/control-plane/api-gateway/cache/consul.go#L557

nathancoleman (approved):

PR #4091, opened 2024-06-07
[NET-9839] fix: add shared version submodule

zalimeni:
Introduce a new submodule for sharing versioning code across the binary-producing submodules in `consul-k8s` repo.

This is both an improvement in terms of consolidation, as well as a fix for incorrect `control-plane` submodule pinning in the `control-plane/cni` submodule that led to incorrect dev versions being used in 1.4.2 and 1.4.3 (the first releases that included a valid version of `control-plane/cni`. See [#4054 comments](https://github.com/hashicorp/consul-k8s/pull/4054#discussion_r1626171944) for more context.

I have a separate change to move us to using an embedded `VERSION` plaintext file like `consul`, but splitting that out from this change for brevity and to reduce risk before the upcoming release. I'll open that one for review as well and link to this PR and we can decide whether we think it's worthwhile.

Once merged, I plan to manually backport this change to `.x` branches and `release/1.5.0`, modifying the versions as needed.

### Changes proposed in this PR ###  
- Consolidate version management in `consul-k8s` binaries into a single shared (private) submodule
- Fix a bug in `prepare-` scripts leaving chart version blank when version is non-dev
- Unrelated cleanup in `go.mod`/`go.sum` due to running `go mod tidy` after module swaps

### How I've tested this PR ###
`control-plane` Docker in CI:
- https://github.com/hashicorp/consul-k8s/actions/runs/9421065328/job/25954781084#step:6:10 `1.6.0-dev`
- https://github.com/hashicorp/consul-k8s/actions/runs/9421065328/job/25954781381#step:7:10 `1.6.0-dev+fips1402`

Downloaded from CI run above:
```shell
~/Downloads/consul-k8s_1.6.0-dev_linux_amd64 
❯ docker run -it --rm --platform linux/amd64 -v "$(pwd):/home/test" -w "/home/test" ubuntu ./consul-k8s -version
v1.6.0-dev (cf59978+CHANGES)

~/Downloads/consul-cni_1.6.0-dev_linux_amd64   
❯ docker run -it --rm --platform linux/amd64 -v "$(pwd):/home/test" -w "/home/test" ubuntu ./consul-cni -version
CNI consul-cni plugin v1.6.0-dev (cf59978+CHANGES)
CNI protocol versions supported: 0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 1.0.0
```

Local:
```shell
❯ make version
1.6.0-dev

❯ cd control-plane/
~/workspace/consul-k8s/control-plane 
[...]

PR #4093, opened 2024-06-07
build: use VERSION file for setting version

zalimeni:
Adopt `consul` repo pattern for setting version to simplify management and build scripts.

Follow-up to https://github.com/hashicorp/consul-k8s/pull/4091

This PR will require manual backporting due to differences in versions and potential diffs in build scripts.

- Switch from parsed-Go approach to a simpler plaintext file a la `consul` (c.f. https://github.com/hashicorp/consul/pull/15631/files#diff-d8adf935a14f14d8de19314b10898d8815b13ee62dad994cd994eaefb5d1d8dfR1)

### Checklist ###
- [ ] Tests added


control-plane/build-support/functions/10-util.sh
```diff
@@ -108,70 +105,36 @@ function parse_version {
 	#   If the GOTAGS environment variable is present then it is used to determine which
 	#   version file to use for parsing.
 
-	local vfile="${1}/version/version.go"
+	local vfile="${1:-.}/version/VERSION"
```
zalimeni, line 108: Helps when running locally and sourcing scripts

control-plane/build-support/functions/10-util.sh
```diff
@@ -108,70 +105,36 @@ function parse_version {
...
-
-	# try to determine the version if we have build tags
-	for tag in "$GOTAGS"; do
-		for vfile in $(find "${1}/version" -name "version_*.go" 2>/dev/null | sort); do
-			if grep -q "// +build $tag" "${vfile}"; then
-				version_main=$(awk '$1 == "Version" && $2 == "=" { gsub(/"/, "", $3); print $3 }' <${vfile})
-				release_main=$(awk '$1 == "VersionPrerelease" && $2 == "=" { gsub(/"/, "", $3); print $3 }' <${vfile})
-			fi
-		done
-	done
```
zalimeni, line 151: This was `consul` copypasta by Mitchell and never used from what I can see (we don't have any files that match those globs)

control-plane/build-support/functions/10-util.sh
```diff
@@ -108,70 +105,34 @@ function parse_version {
...
 	# ensure the version file exists
 	if ! test -f "${vfile}"; then
 		err "Error - File not found: ${vfile}"
 		return 1
 	fi
 
-	local include_release="$2"
-	local use_git_env="$3"
-	local omit_version="$4"
-
```
zalimeni, line 122: These were always `true`, `false`, and unset, respectively (2 uses of `parse_version`, both in `40-publish.sh`)

control-plane/build-support/scripts/version.sh
```diff
@@ -1,14 +0,0 @@
-#!/usr/bin/env bash
```
zalimeni, line 1: This file was never used and is redundant with `make version`

control-plane/build-support/functions/10-util.sh
```diff
@@ -108,70 +105,34 @@ function parse_version {
...
+	release_main=$(cat ${vfile} | awk '{ split($0, arr, "-"); print arr[2]; }')
 
 	local version="${version_main}"
 	# override the version from source with the value of the GIT_DESCRIBE env var if present
 	if test -n "${git_version}"; then
 		version="${git_version}"
 	fi
 
-	local rel_ver=""
-	if is_set "${include_release}"; then
```
zalimeni, line 160: Always true, so removed as outer `if`

control-plane/build-support/functions/10-util.sh
```diff
@@ -108,70 +105,34 @@ function parse_version {
...
-		# Default to pre-release from the source
-		rel_ver="${release_main}"
-
-		# When no GIT_DESCRIBE env var is present and no release is in the source then we
-		# are definitely in dev mode
-		if test -z "${git_version}" -a -z "${rel_ver}" && is_set "${use_git_env}"; then
-			rel_ver="dev"
-		fi
```
zalimeni, line 168: `use_git_env` was always false, so removed

PR #4104, opened 2024-06-11
Backport: [NET-9606] Use namespace-local Secret for mounting TLS certificates into api-gateway Pod

nathancoleman:
> [!NOTE]
> This is a backport of #4100 targeting the consul-k8s 1.5.0 release

# Original PR Body

In order to mount a volume from a `Secret` containing TLS certificates into the API gateway `Pod`, those `Secrets` must exist in the same namespace as the API gateway `Pod`. This centralizes all referenced TLS certificates into a `Secret` in the same namespace as the API gateway `Pod` to make that possible. This `Secret` is managed in the same way as other API gateway resources - such as `Role`, `Service`, etc. - in that it has an owner reference pointing to the `Gateway` that's consuming it. That owner reference is checked before any operations take place on the `Secret` to ensure that we aren't overwriting any pre-existing resources that we don't own.

Create a `Gateway` that references a TLS certificate `Secret` from a namespace _other than the one containing the `Gateway`._ Verify that the certificate is correctly consumed by the `Gateway`, and Envoy ends up serving up the secret in its config.

Configurations for this kind of setup is available [here](https://github.com/nathancoleman/consul-lab/tree/main/k8s/api-gateway/basic).

See above

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/api-gateway/gatekeeper/gatekeeper_test.go
+++ b/control-plane/api-gateway/gatekeeper/gatekeeper_test.go
@@ -95,6 +95,7 @@ type resources struct {
 	deployments     []*appsv1.Deployment
 	roles           []*rbac.Role
 	roleBindings    []*rbac.RoleBinding
+	secrets         []*corev1.Secret
 	services        []*corev1.Service
 	serviceAccounts []*corev1.ServiceAccount
 }
@@ -146,6 +147,7 @@ func TestUpsert(t *testing.T) {
 					configureDeployment(name, namespace, labels, 3, nil, nil, "", "1"),
 				},
 				roles:           []*rbac.Role{},
+				secrets:         []*corev1.Secret{},
 				services:        []*corev1.Service{},
 				serviceAccounts: []*corev1.ServiceAccount{},
 			},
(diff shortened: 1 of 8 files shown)


control-plane/api-gateway/gatekeeper/gatekeeper_test.go
```diff
@@ -908,12 +939,90 @@ func TestUpsert(t *testing.T) {
...
 					configureRoleBinding(name, namespace, labels, "1"),
 				},
+				secrets:  []*corev1.Secret{},
 				services: []*corev1.Service{},
 				serviceAccounts: []*corev1.ServiceAccount{
 					configureServiceAccount(name, namespace, labels, "1"),
 				},
 			},
 		},
+		"create a new gateway with TLS certificate reference": {
```
jm96441n, line 949: minor: can we add a test where the secret is in a separate namespace just to document how that works

nathancoleman: This is the backport of the main PR that already merged. Okay with adding this test in a separate PR so we can keep everything lined up?

jm96441n (approved):
one minor comment about an additional test, otherwise LGTM

PR #4113, opened 2024-06-12
Update Error Message when Secret Reference is invalid

jm96441n:
- make the error message when the certificate reference kind/group is incorrect more descriptive. Previously the message was just `certificate type is not supported` which could point to an issue with the actual certificate itself _or_ the reference for it. This PR expands the error message to tell the user how to correct the issue

### Checklist ###
- [X] Tests added


control-plane/api-gateway/binding/validation.go
```diff
@@ -322,9 +323,18 @@ func validateJWT(gateway gwv1beta1.Gateway, listener gwv1beta1.Listener, resourc
 
 func validateCertificateRefs(gateway gwv1beta1.Gateway, refs []gwv1beta1.SecretObjectReference, resources *common.ResourceMap) error {
 	for _, cert := range refs {
+		var mErr error
 		// Verify that the reference has a group and kind that we support
-		if !common.NilOrEqual(cert.Group, "") || !common.NilOrEqual(cert.Kind, common.KindSecret) {
-			return errListenerInvalidCertificateRef_NotSupported
+		if !common.NilOrEqual(cert.Group, "") {
+			mErr = errors.Join(mErr, fmt.Errorf("group is not supported: %q, supported Groups are \"\"", common.DerefStringOr(cert.Group, "")))
```
sarahalsmiller, line 329: should this be supported Groups are %s, or is the only supported group always just ""

jm96441n: for now it's always `""` based on the conditional on the line above that triggers this

sarahalsmiller (approved):

PR #4127, opened 2024-06-13
ci: add conditional skip for docs changes

zalimeni:
[Similar to consul](https://github.com/hashicorp/consul/blob/e9c983f36107b044fcd889f738b3d5ba2b922e7a/.github/workflows/go-tests.yml#L33), skip build and tests if we're only changing docs.

This should help speed up releases and daily work quick-fixes.

`consul-dataplane` PR with testing notes: https://github.com/hashicorp/consul-dataplane/pull/560


DanStough (approved):

zalimeni:
See check here: https://github.com/hashicorp/consul-k8s/actions/runs/9505220792/job/26199606917#step:3:9

PR #4128, opened 2024-06-13
[NET-9446] chore: update CHANGELOG and tests for 1.5.0

zalimeni:
I'll follow up to set `"test-ce": false` on 1.4.x next.


DanStough (approved):
🙇‍♂️

PR #4152, opened 2024-06-26
Net 4414 remove anyuid openshift requirement

missylbytes:
This PR solves two problems:
- Adds the NET_BIND_SERVICE capability to the Consul Dataplane containers that are created. This work was done in another PR, but has been pulled into this one. [Original PR](https://github.com/hashicorp/consul-k8s/pull/4066)
- Removes the need on Openshift for using an SCC that allows anyuid. We now use the annotation provided by Openshift on the namespace to set the user and group IDs for our containers. This work was originally done in another PR, but has been pulled into this one and expanded to cover API gateways and use separate user/group IDs for the init + dataplane containers that do not conflict with application user/group IDs. [Original PR](https://github.com/hashicorp/consul-k8s/pull/3813)

Manual testing
Unit tests

Follow [this gist](https://gist.github.com/missylbytes/510ae4dda7e91f7e42d37c768a7f4b59) to setup and install Consul onto Openshift

### Checklist ###
- [x] Tests added


.changelog/3813.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+control-plane: Remove anyuid Security Context Constraints (SCC) requirement in OpenShift.
```
missylbytes, line 2: Need to also add release note for a NET_BIND_SERVICE

control-plane/api-gateway/gatekeeper/init.go
```diff
@@ -175,9 +178,36 @@ func initContainer(config common.HelmConfig, name, namespace string) (corev1.Con
...
+			Name: namespace,
+		}, ns)
+		if err != nil {
+			g.Log.Error(err, "error fetching namespace metadata for deployment")
+			return corev1.Container{}, fmt.Errorf("error getting namespace metadata for deployment: %s", err)
+		}
+
+		// We need to get the userID for the dataplane, we do not care about what is already defined on the pod
+		// for gateways, as there is no application container that could have taken a UID.
+		uid, err = ctrlCommon.GetDataplaneUID(*ns, corev1.Pod{})
```
missylbytes, line 197: Switch this to init container

sarahalsmiller:
Tested on open shift and saw the fix working, after Nathan and Melissa walked me through it. Reviewing the code portion now

control-plane/api-gateway/gatekeeper/gatekeeper_test.go
```diff
@@ -29,6 +29,13 @@ import (
 	"github.com/hashicorp/consul-k8s/control-plane/connect-inject/constants"
 )
 
+const (
+	designatedOpenShiftUIDRange       = "1000700000/100000"
```
sarahalsmiller, line 33: Is this always the same or does it need to be pulled in from the annotations?

control-plane/go.mod
```diff
@@ -164,4 +164,4 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-go 1.20
+go 1.21
```
sarahalsmiller, line 167: Does 1.22 break?

sarahalsmiller (approved):
Approved since I've tested it, but some non blocking questions

control-plane/api-gateway/gatekeeper/gatekeeper_test.go
missylbytes: In the real world it is pulled from the annotation, for this test we just explicitly set it in the annotation (line 949 for example) since there is no cluster

control-plane/go.mod
nathancoleman: I wouldn't expect it to. The `go <version>` directive here specifies the minimum Go version required to run the application, not necessarily the version that it will be built with since that's determined by what version of `go` is installed when you're building.

https://go.dev/ref/mod#go-mod-file-go

control-plane/connect-inject/webhook/redirect_traffic.go
```diff
@@ -19,16 +19,35 @@ import (
...
 //	ExcludeInboundPorts: prometheus, envoy stats, expose paths, checks and excluded pod annotations
 //	ExcludeOutboundPorts: pod annotations
 //	ExcludeOutboundCIDRs: pod annotations
 //	ExcludeUIDs: pod annotations
 func (w *MeshWebhook) iptablesConfigJSON(pod corev1.Pod, ns corev1.Namespace) (string, error) {
-	cfg := iptables.Config{
-		ProxyUserID: strconv.Itoa(sidecarUserAndGroupID),
+	cfg := iptables.Config{}
+
+	if !w.EnableOpenShift {
```
NiniOak, line 32: Should this be `if w.EnableOpenShift {..}`?
If I'm reading the code correctly, this block of code is attempting to get uids from ns to exclude from traffic redirection which is specific to OpenShift?

missylbytes: The else statement handles getting the uids from the namespace, not this block.

NiniOak (approved):

control-plane/go.mod
missylbytes: Also un-bumped go version, should be done in a separate PR

PR #4153, opened 2024-06-28
[NET-10247] Add Partition and Namespace blocks to external registration acl role/policy

jm96441n:
NOTE: Dependent on https://github.com/hashicorp/consul/pull/21382

- Add Partition and Namespace blocks to external registration acl role/policy
- Make `address` field optional in the CRD like it is for catalog registraton
- update license headers

- run the `start.sh` script in the following gist (ensure you have a local build of this branch and of the changes in the linked consul PR in consul-enterprise and have a valid consul license in an env variable named `CONSUL_LICENSE`
- shell into the `bender` service and run `apk add curl`
- run `curl zoidberg.virtual.consul` and see the request go through
- check the consul connect injector logs to see there are no errors
- view the `zoidberg-write-policy` to see the `namespace` block applied

### Checklist ###
- [X] Tests added

### DIFF
--- a/control-plane/catalog/registration/cache.go
+++ b/control-plane/catalog/registration/cache.go
@@ -4,39 +4,94 @@
 package registration
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"slices"
 	"strings"
 	"sync"
+	"text/template"
 
 	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/go-logr/logr"
 	"github.com/hashicorp/consul-k8s/control-plane/api/v1alpha1"
 	"github.com/hashicorp/consul-k8s/control-plane/consul"
 	capi "github.com/hashicorp/consul/api"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const NotInServiceMeshFilter = "ServiceMeta[\"managed-by\"] != \"consul-k8s-endpoints-controller\""
 
-type RegistrationCache struct {
-	ConsulClientConfig  *consul.Config
-	ConsulServerConnMgr consul.ServerConnectionManager
-	serviceMtx          *sync.Mutex
-	Services            map[string]*v1alpha1.Registration
-	synced              chan struct{}
-	UpdateChan          chan string
+func init() {
+	gatewayTpl = template.Must(template.New("root").Parse(strings.TrimSpace(gatewayRulesTpl)))
 }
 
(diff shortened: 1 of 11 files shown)


control-plane/catalog/registration/cache.go
```diff
@@ -77,20 +132,40 @@ func (c *RegistrationCache) run(ctx context.Context, log logr.Logger) {
...
 				log.Info("consul deregistered service", "svcName", svc)
 				c.UpdateChan <- svc
 			}
 
+			for _, svc := range servicesToAdd.ToSlice() {
+				registration := &v1alpha1.Registration{}
+
+				if err := c.k8sClient.Get(c.ctx, types.NamespacedName{Name: svc, Namespace: namespace}, registration); err != nil {
+					if !k8serrors.IsNotFound(err) {
+						log.Error(err, "unable to get registration", "svcName", svc)
```
nathancoleman, line 161: Probably helpful to include the namespace here as well

control-plane/catalog/registration/cache.go
```diff
@@ -148,38 +228,51 @@ func (c *RegistrationCache) updateTermGWACLRole(log logr.Logger, registration *v
...
+			log.Error(err, "error reading policy")
+			return err
 		}
-	} else {
-		policy = existingPolicy
-	}
 
-	var mErr error
-	for _, termGW := range termGWsToUpdate {
+		writeOpts := &capi.WriteOptions{}
```
nathancoleman, line 266: Do policies and roles have global or partition scope (as opposed to a particular namespace)?

control-plane/catalog/registration/cache.go
```diff
@@ -239,13 +332,17 @@ func (c *RegistrationCache) removeTermGWACLRole(log logr.Logger, registration *v
...
-	if err != nil {
-		return err
-	}
-
 	var mErr error
 	for _, termGW := range termGWsToUpdate {
+
+		queryOpts := &capi.QueryOptions{}
+
+		writeOpts := &capi.WriteOptions{}
```
nathancoleman, line 340: ~~What do we get by defining these as empty structs instead of continuing to pass `nil` below?~~

See above comment

control-plane/catalog/registration/cache.go
nathancoleman: Discussed on a call that it would be helpful to comment why these don't include a namespace. This is because historically all roles + policies are just created in the `default` namespace, and we don't want to re-evaluate that decision in the scope of this work.

nathancoleman (changes requested):
Just a couple of changes we discussed on a call; otherwise LGTM

nathancoleman (approved):

PR #4167, opened 2024-07-12
NET-10249 - add server acl init option to create ACL role for dns-proxy

dhiaayachi:
- Add `-dns-proxy` flag to the server-acl-init job
- When the flag is set to true render and create an acl policy and role associated to dns-proxy
- Bind that role to the dns-proxy service account

- Unit test for acl policy rendering
- manually using an eks cluster

### Checklist ###
- [x] Tests added


control-plane/subcommand/server-acl-init/rules.go
```diff
@@ -444,3 +444,24 @@ func (c *Command) renderRulesGeneric(tmpl string, data interface{}) (string, err
 
 	return buf.String(), nil
 }
+
+// acl = "write" is required when creating namespace with a default policy.
+// Attaching a default ACL policy to a namespace requires acl = "write" in the
+// namespace that the policy is defined in, which in our case is "default".
```
jmurret, line 450:
```suggestion

// dnsProxyRules defines the ACL policy for the `dns-proxy` service.  It defines the following:
// - read access to `node_prefix`.
// - read access to `service_prefix`.
// - will scope `partition` if provided.
// - namespaces except for the `default` are currently not supported.
```

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -222,6 +223,9 @@ func (c *Command) init() {
 	c.flags.BoolVar(&c.flagLogJSON, "log-json", false,
 		"Enable or disable JSON output format for logging.")
 
+	c.flags.BoolVar(&c.flagDNSProxy, "dns-proxy", false,
```
jmurret, line 226: [micro-nit] could put this in a const for this and the below 2 other uses.

control-plane/subcommand/server-acl-init/command.go
```diff
@@ -668,6 +672,21 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
+	if c.flagDNSProxy {
+		serviceAccountName := c.withPrefix("dns-proxy")
+
+		dnsProxyRules, err := c.dnsProxyRules()
+		if err != nil {
+			c.log.Error("Error templating inject rules", "err", err)
```
jmurret, line 680:
```suggestion
			c.log.Error("Error templating dns-proxy rules", "err", err)
```
could use const here as well

control-plane/subcommand/server-acl-init/rules_test.go
```diff
@@ -1013,6 +1013,101 @@ partition "part-1" {
 	}
 }
 
+// Test the inject rules with namespaces enabled or disabled.
+func TestDnsProxyRules(t *testing.T) {
```
jmurret, line 1017: 👏

jmurret (approved):
:shipit: looks great!  some minor nits.

PR #4168, opened 2024-07-12
Fix and re-enable catalog sync ingress acceptance test

ndhanushkodi:
- Un-Skip the catalog sync + ingress test
- Fix check and retry in case service event is picked up by controller first
- This test does not have to run on just EKS (nothing eks specific in it), so enabled it such that it should run on every PR.

- Ran acceptance tests locally

👀 

### Checklist ###
- [x] Tests added
~- [ ] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)~


acceptance/tests/sync/sync_catalog_test.go
```diff
@@ -86,15 +86,10 @@ func TestSyncCatalog(t *testing.T) {
...
 func TestSyncCatalogWithIngress(t *testing.T) {
-	t.Skip("TODO(fails): NET-8594")
-
 	cfg := suite.Config()
 	if cfg.EnableCNI {
 		t.Skipf("skipping because -enable-cni is set and sync catalog is already tested with regular tproxy")
 	}
-	if !cfg.UseEKS {
-		t.Skipf("skipping because -use-eks is not set and the ingress test only runs on EKS")
-	}
```
dhiaayachi, line 97: Is this not true anymore? 🤔

dhiaayachi (approved):
Thank you for working on this @ndhanushkodi 
I have a question about removing a skip otherwise LGTM!

acceptance/tests/sync/sync_catalog_test.go
ndhanushkodi: Yeah, this test isn't specific to EKS, as it works fine on Kind as well. I think originally the feature was built for folks who wanted to use it on EKS, but nothing about the feature acts on specific EKS annotations or relies on EKS itself. It doesn't even need an ingress controller deployed for the test itself, since the catalog sync only watches Ingress resources themselves.

PR #4184, opened 2024-07-16
Net 10114 consul is under resourced when deploying to crc local version of open shift

sarahalsmiller:
- Make resources in webhookCertManager configurable to resolve OOM error when installing consul on CRC
 
- Local install to crc from local helm chart

### Checklist ###
- [ ] Tests added


charts/consul/values.yaml
```diff
@@ -3480,6 +3480,33 @@ webhookCertManager:
...
+  #
+  # @recurse: false
+  # @type: map
+  resources:
+    requests:
+      memory: "50Mi"
+      cpu: "100m"
+    limits:
+      memory: "50Mi"
+      cpu: "200m"
```
sarahalsmiller, line 3508:
```suggestion
      cpu: "100m"
```

jm96441n:
do we need to do anything in code to set the defaults we previously had?

sarahalsmiller:
> do we need to do anything in code to set the defaults we previously had?

The chart has their original values if they're not set in the values.yml. If we're overriding the resources I'm not sure if the code should try and overwrite that since we don't seem to be doing that anywhere else the resources are configurable.

jm96441n (approved):

PR #4194, opened 2024-07-17
Add non-code files/directories to skip for acceptance tests

jm96441n:
- When non-code files are changed don't run unit tests/acceptance tests as all should already be passing

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.github/scripts/check_skip_ci.sh
+++ b/.github/scripts/check_skip_ci.sh
@@ -29,8 +29,6 @@ files_to_check=$(git diff --name-only "$(git merge-base origin/$SKIP_CHECK_BRANC
 # Define the directories to check
 skipped_directories=("assets" ".changelog/", "version", ".github/")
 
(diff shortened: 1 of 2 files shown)


.github/scripts/check_skip_ci.sh
```diff
@@ -16,35 +27,40 @@ set -euo pipefail
 files_to_check=$(git diff --name-only "$(git merge-base origin/$SKIP_CHECK_BRANCH HEAD~)"...HEAD)
 
 # Define the directories to check
-skipped_directories=("assets" ".changelog/")
+skipped_directories=("assets" ".changelog/", "version", ".github")
```
jmurret, line 30:
```suggestion
skipped_directories=("assets" ".changelog/", "version")
```
if we change workflows, this means that CI would not run to assert them

jm96441n: addressed

jm96441n:
If you check the CI run for commit `c64441a` you will see all the conditional skips applied

sarahalsmiller (approved):

Issue #3862, opened 2024-04-03
gateway-resources service account missing imagePullSecrets


VladAlexF:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

`consul-k8s/charts/consul/templates/gateway-resources-serviceaccount.yaml` is missing `imagePullSecrets`, which breaks the usage of private docker registries, as the Gateway Resources Job cannot pull the `consul-k8s-control-plane` image from private registries without these secrets.  
Note, other service accounts do include the `imagePullSecrets`, and therefore other pods can successfully pull from the private registry.

### Reproduction Steps

1. Run a helm install with the following `values.yaml` file:
```yaml
global:
  imagePullSecrets: 
    - name: private-registry-pull-secret
  imageConsulDataplane: <private-dockerhub-proxy-cache>.com/dockerhub/hashicorp/consul-dataplane:latest
  imageK8S: <private-dockerhub-proxy-cache>.com/dockerhub/hashicorp/consul-k8s-control-plane:latest
  image: <private-dockerhub-proxy-cache>.com/dockerhub/hashicorp/consul:latest
```
2. The `<release-name>-gateway-resources` job cannot launch containers, as it cannot pull the image from the private registry, due to missing imagePullSecrets on the service account the job uses. 

### Logs

The container cannot produce logs as it doesn't start, so kubernetes events for the pod from command `kubectl -n consul describe pod consul-gateway-resources-2fz5z` are provided:  
```
Events:
  Type     Reason     Age                    From               Message
  ----     ------     ----                   ----               -------
  Normal   Scheduled  53m                    default-scheduler  Successfully assigned consul/consul-gateway-resources-2fz5z to k8s05
  Normal   Pulling    52m (x4 over 53m)      kubelet            Pulling image "<private-registry>.com/dockerhub/hashicorp/consul-k8s-control-plane:latest"
[...]

PR #4210, opened 2024-07-25
NET-10548 Adds image pull secrets to the gateway cleanup and gateway resources job service account

missylbytes:
This fixes https://github.com/hashicorp/consul-k8s/issues/3862. When installing consul-k8s with a private image registry, the 'gateway-resources' and 'gateway-cleanup jobs could not pull the images as thier service accounts did not include the imagePull secrets

- Added bats test
- Manually tested with the following values and a private registry set up
*Do a helm install with the local chart from this branch*. If the jobs didn't have the required pull secrets, they would never spin up.
Also do a helm uninstall to see the cleanup-job run.
```
global:
  name: consul
  imageK8S: my-private-registry/consul-k8s-control-plane:latest
  imagePullSecrets:
  - name: consul-pull-secret
    namespace: consul
# This is useful to make sure it is always pulling the image, and not just getting it from another container on the cluster
  imagePullPolicy: "Always" 
```

Test on your own cluster with a private registry using the above values.

### Checklist ###
- [x] Tests added

### DIFF
--- a/.changelog/4210.txt
+++ b/.changelog/4210.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+helm: adds imagePullSecret to the gateway-resources job and the gateway-cleanup job, would fail before if the image was in a private registry
+```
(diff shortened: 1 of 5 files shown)


pawellegowski89:
#3862 the error has not been fixed. Version 1.19.2, chart 1.5.3 still affected.

PR #4212, opened 2024-07-26
NET-10124: Add metrics endpoint and labels to sync catalog deployment

NiniOak:
- Add metrics endpoints to scrape metrics from sync catalog deployment
- Add labels to monitor register, deregistered and errors syncing K8s services to Consul

### Checklist ###
- [ ] Tests added


control-plane/catalog/to-consul/syncer.go
```diff
@@ -17,6 +19,25 @@ import (
 	"github.com/hashicorp/go-hclog"
 )
 
+var SyncToConsulCounters = []prometheus.CounterDefinition{
+	{
+		Name: []string{"consul", "sync_catalog", "to_consul"},
```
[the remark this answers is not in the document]
NiniOak: Yeah, that's a good idea for clarity

control-plane/catalog/to-consul/syncer.go
```diff
@@ -433,14 +456,28 @@ func (s *ConsulSyncer) syncFull(ctx context.Context) {
...
 			"service-id", r.ServiceID,
 			"service-consul-namespace", r.Namespace)
+
 		_, err = consulClient.Catalog().Deregister(r, nil)
 		if err != nil {
+			// metric count for error syncing K8S services with Consul via calatog sync
+			labels := []metrics.Label{
+				{Name: "error", Value: err.Error()},
+			}
+			s.PrometheusSink.IncrCounterWithLabels([]string{"consul", "sync_catalog", "to_consul", "deregister", "error"}, 1, labels)
```
jm96441n, line 466: can we assign these slices to variables and reuse them throughout the code so if we change anything it's just changing it in one place?

control-plane/catalog/to-k8s/sink.go
```diff
@@ -19,6 +21,25 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+var SyncToK8sCounters = []prometheus.CounterDefinition{
+	{
+		Name: []string{"consul", "sync_catalog", "to_k8s"},
```
jm96441n, line 26: similar idea here, can we move these slices to variables and reuse them through the code and should we include "register" here as we use "deregister" for the opposite?

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -68,6 +73,12 @@ type Command struct {
 	flagK8SNSMirroringPrefix       string   // Prefix added to Consul namespaces created when mirroring
 	flagCrossNamespaceACLPolicy    string   // The name of the ACL policy to add to every created namespace if ACLs are enabled
 
+	// Metrics settings.
+	flagEnableMetrics        string
```
jm96441n, line 77: if we make this a bool instead of a string then we don't need to do any of the `isset` type stuff on parsing the value

jm96441n: one thing to keep in mind is that when we call it with `false` (which we probably don't need to here because just setting `-enable-metrics` will set the value to true and it will default to false) buuuut to call it with an explicit false we'd need to do `-enable-metrics=false`
https://pkg.go.dev/flag#hdr-Command_line_flag_syntax

control-plane/catalog/to-k8s/sink.go
```diff
@@ -232,7 +255,18 @@ func (s *K8SSink) Run(ch <-chan struct{}) {
...
+
+				// metric count for error syncing Consul services to K8s
+				labels := []metrics.Label{
+					{Name: "error", Value: err.Error()},
+				}
+				s.PrometheusSink.IncrCounterWithLabels([]string{"consul", "sync_catalog", "to_k8s", "deregister", "error"}, 1, labels)
+			}
+			// metric count for error syncing Consul services to k8s
+			labels := []metrics.Label{
+				{Name: "service_name", Value: name},
```
jm96441n, line 267: similar here with the `Name` field, can we extract these to constants since they're used in a few spots

control-plane/subcommand/common/metrics_util.go
```diff
@@ -0,0 +1,32 @@
+package common
+
+import "strconv"
+
+func GetScrapePort(v string) (int, bool) {
```
jm96441n, line 5: nit: can we rename these to be `ParseScrapePort` and `ParseScrapePath`, `Get` implies we're getting the value from somewhere (like a map, slice, or some external datasource) but really we're taking a value and parsing it

charts/consul/templates/sync-catalog-deployment.yaml
```diff
@@ -196,6 +203,18 @@ spec:
             {{- if .Values.syncCatalog.syncLoadBalancerEndpoints }}
             -sync-lb-services-endpoints=true \
             {{- end }}
+            {{- if (or (and (ne (.Values.syncCatalog.metrics.enabled | toString) "-") .Values.syncCatalog.metrics.enabled) (and (eq (.Values.syncCatalog.metrics.enabled | toString) "-") .Values.global.metrics.enabled)) }}
+            -enable-metrics=true \
+            {{- else }}
+            -enable-metrics=false \
```
jm96441n, line 209: so if we make this a boolean instead of a string we can get rid of the entire branch and just leave it unset

NiniOak: can you expand on this comment, not sure what you mean by `we can get rid of the entire branch`

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -259,6 +276,22 @@ func (c *Command) Run(args []string) int {
...
+	// Create the metrics sink
+	err = c.recordMetrics()
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("unable to create metrics sink: %s", err))
+		return 1
+	}
+
+	if c.prometheusSink == nil {
+		c.logger.Error("Prometheus sink not initialized, metrics cannot be displayed")
+		c.prometheusSink = &prometheus.PrometheusSink{}
```
NiniOak, line 292: I need this to not throw a nil pointer error. Happy to remove

control-plane/catalog/to-consul/syncer.go
```diff
@@ -17,6 +19,27 @@ import (
...
+	{
+		Name: append(baseName, "register", "error"),
+		Help: "Increments whenever a Consul api client returns an error for a catalog sync register request",
+	},
+	{
+		Name: append(baseName, "deregister", "error"),
+		Help: "Increments whenever a Consul api client returns an error for a catalog sync deregister request request",
+	},
+}
+
```
jm96441n, line 42: can we just move all these names to named variables so we can reuse them throughout the code? (also minor, but the way we're using append here typically isn't safe, it is here because the append calls always allocate a new backing array because the capacity of `baseName` is 3, but if the capacity was to the point that none of these appends cause a re-alloc it wouldn't be safe, eg: https://go.dev/play/p/8k4jvB6MTJS)
```suggestion
var (
     baseName = []string{"consul", "sync_catalog", "to_consul"}
     registerName = append(baseName, "register")
     deregisterName = append(baseName, "deregister")
     registerErrorName = append(baseName, "register", "error")
     deregisterErrorName = append(baseName, "deregister", 'error")
)

var SyncToConsulCounters = []prometheus.CounterDefinition{
	{
		Name: registerName,
		Help: "Increments for each service registered to Consul via catalog sync",
	},
	{
		Name: deregsiterName,
		Help: "Increments for each service deregistered from Consul via catalog sync",
	},
	{
		Name: registerErrorName,
		Help: "Increments whenever a Consul api client returns an error for a catalog sync register request",
	},
	{
		Name: deregisterErrorName,
		Help: "Increments whenever a Consul api client returns an error for a catalog sync deregister request request",
	},
}

```

control-plane/helper/test/test_util.go
```diff
@@ -43,6 +44,8 @@ type TestServerClient struct {
 	Watcher    consul.ServerConnectionManager
 }
 
+var TestPrometheusSink = &prometheus.PrometheusSink{}
```
jm96441n, line 47: why do we need this global here? can we create this in the `testSink` function and just use it there?

control-plane/subcommand/sync-catalog/command.go
jm96441n: will this condition ever not be nil? the default for a pointer field is `nil` and I don't see where we're setting this so I'd assume we always fall into this block right?

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -438,9 +486,57 @@ func (c *Command) validateFlags() error {
...
+	c.prometheusSink, err = prometheus.NewPrometheusSinkFrom(opts)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// authorizeMiddleware validates the token and returns http handler.
+func (c *Command) authorizeMiddleware() func(http.Handler) http.Handler {
```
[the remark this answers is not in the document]
NiniOak: This is just a placeholder for now

control-plane/subcommand/sync-catalog/command.go
NiniOak: It had some errors in testing so I put this there, I'll remove it cos it's not a condition I see us hitting in production.

jm96441n: oh nah I think it's a valid check, what I mean is there ever a case where this would actually be set? to me it seems like this will always be nil and need to be set

NiniOak: No it won't be nil, `recordMetrics()` returns `prometheus.PrometheusSink` which we assign to this field. So `c.prometheusSink` value will only be nil if `recordMetrics()` returns an error(as we just log create metrics sink error and continue running feature).

control-plane/catalog/to-consul/syncer.go
```diff
@@ -433,14 +464,28 @@ func (s *ConsulSyncer) syncFull(ctx context.Context) {
...
 				"node-name", r.Node,
 				"service-id", r.ServiceID,
 				"service-consul-namespace", r.Namespace,
 				"err", err)
+			continue
+		}
+
+		// metric count for deregistering k8s services from Consul
+		labels := []metrics.Label{
+			{Name: "id", Value: r.ServiceID},
```
nathancoleman, line 486: Wondering if we should go ahead and include the node name, service namespace, etc. in the labels here like we do for registrations. Thoughts?

control-plane/subcommand/common/metrics_util.go
```diff
@@ -0,0 +1,32 @@
...
+		return 0, false
+	}
+	return port, true
+}
+
+func GetScrapePath(v string) (string, bool) {
+	if v == "" {
+		return "", false
+	}
+	return v, true
```
nathancoleman, line 21: I think this is equivalent

```suggestion
	return v, v != ""
```

nathancoleman (commented):
Some thoughts just looking through the code. Spending some time this afternoon tinkering with the new metrics in Grafana. Exciting stuff, great work!

control-plane/catalog/to-consul/syncer.go
```diff
@@ -17,6 +19,33 @@ import (
...
+	registerName        = append(baseName, "register")
+	deregisterName      = append(baseName, "deregister")
+	registerErrorName   = append(baseName, "register", "error")
+	deregisterErrorName = append(baseName, "deregister", "error")
+)
+
+var SyncToConsulCounters = []prometheus.CounterDefinition{
+	{
+		Name: registerName,
+		Help: "Increments for each service registered to Consul via catalog sync",
```
nathancoleman, line 33: Are these metrics meant to count "service" or "service instance"? "Service" implies to me a higher level construct that can consist of multiple instances.

control-plane/catalog/to-consul/syncer.go
```diff
@@ -478,6 +529,19 @@ func (s *ConsulSyncer) syncFull(ctx context.Context) {
 				"service-name", r.Service.Service,
 				"consul-namespace-name", r.Service.Namespace,
 				"service", r.Service)
+
+			// metric count and service metadata syncing k8s services to Consul
+			labels := []metrics.Label{
+				{Name: "id", Value: r.Service.ID},
+				{Name: "service", Value: r.Service.Service},
+				{Name: "external_k8s_ref_name", Value: r.Service.Meta["external-k8s-ref-name"]},
```
nathancoleman, line 537: Together we noticed that this is coming back unset sometimes. We might want to handle this case more gracefully and/or determine why it's unset.

control-plane/catalog/to-consul/syncer.go
blake: Agreed, it sounds like this should be new service instance.

control-plane/catalog/to-consul/syncer.go
blake: @nathancoleman Is this the only K8s service that exhibits this behavior? If so, would it be a problem to ignore this service and not sync it to Consul?

nathancoleman: that's the only one I saw in my testing, but I don't fully understand what leads to this service not having the same metadata. can look into it a bit more

nathancoleman: After more digging, this case only happens when the `EndpointSlices` for a given k8s `Service` have no target ref. This can't generally be the case because a `Service` for everyday use must target some sort of backend like a `Pod`; however, the `kubernetes` service that is created in the default namespace doesn't have one since there is no `Pod` to represent the kube API server.

nathancoleman (approved):

control-plane/controllers/configentries/configentry_controller_test.go
```diff
@@ -1750,7 +1750,7 @@ func TestConfigEntryControllers_doesNotCreateUnownedConfigEntry(t *testing.T) {
 
 			// Change the config entry so protocol is https instead of http if test case says to
 			if c.makeDifferentFromConsul {
-				svcDefaults.Spec.Protocol = "https"
+				svcDefaults.Spec.Protocol = "http2"
```
NiniOak, line 1753: **NOTE**: This changes is not relevant to this PR, currently until tests fail with error 
`Unexpected response code: 500 (1 error occurred:* invalid value for protocol: https)` , this change fixes it. Need to investigate further as to why `https` is an invalid protocol for ServiceDefault config

nathancoleman (approved):

Issue #1486, opened 2022-09-07
Documentation - Transparent proxy + terminating gateway


mr-miles:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

Thanks for all the great work!  We are trying to get the transparent proxy going with mesh traffic only, and to get it to route successfully via a terminating gateway.  In the end we got it working but there are a few things that are undocumented or maybe even bugs, so I am noting them here for documentation updates.

### Reproduction Steps

1.  Terminating gateway will not work with transparent proxy unless dns is enabled in the helm chart.  This is because otherwise dns entries like <myservice>.virtual.consul do not resolve (where <myservice> is exposed on the terminating gateway), and this is needed for the envoy traffic filters to route properly.  AFAICS this is specific to terminating gateway services since these services do not have corresponding k8s resources / cluster IPs.  

2.  If you add a pod annotation specifying the service downstreams, then no transparent proxy envoy filters are created for those services.  This is particularly confusing if you have specified ALL the services this way, since in this case the outbound listener on 15001 isn't created at all!  I expected the transparent proxy setup to make the annotations redundant, rather than having to remove the same-datacenter ones.

david-yu:
Hi @mr-miles Could you provide details on how you are configuring the terminating Gateway via CRD? Btw with the new TProxy enhancements in Terminating Gateway you no longer need to register the service you are trying to dial with Consul and you would need to leverage some new configuration in ServiceDefaults instead: https://www.consul.io/docs/connect/config-entries/service-defaults#terminating-gateway-destination. I realize the docs are a little confusing and we probably need to update related Terminating GW docs elsewhere to show when to use this new config.

mr-miles:
Thanks @david-yu for the quick response  - I was not aware of that service-defaults usage.  And yes, currently I'm registering the destinations as services.

[...]

PR #4224, opened 2024-07-31
[NET-10567] Fix namespace normalization on external registration/ACL Setup for Terminating Gateways

jm96441n:
- Fix namespace normalization on external registration (when namespaces are enabled the empty namespace is converted to "default" for acl rules)
- Fix how acls are setup for terminating gateways -> moved from the registrations controller/cache to the terminating gateway controller to avoid race conditions with persisting the terminating gateway config entry (TODO: clean up extra policies when a terminating gateway is removed/services are removed and are not referenced by any other terminating gateways)
- Fixed bug in cache where values weren't being persisted with the correct key
- updated terminating gateway tests so they would catch this in the future

- wrote tests
- ran in kind cluster

- clone this branch and build the container using `make docker-dev`
- clone down the following gist: https://gist.github.com/jm96441n/1be432dd2b6b5ade7bc1b058539f6095
- modify the `CONSUL_K8S_CHART_LOCATION` variable in the `start.sh` file to point to the helm charts in your local version of consul-
- run the `start.sh` file (this requires you to have `kind` and `yq` on your machine, and you'll need to run `chmod +x ./start.sh`)
- after that completes run the following to set up your consul CLI to talk to consul:
```sh
export CONSUL_HTTP_ADDR="127.0.0.1:8501"
export CONSUL_HTTP_TOKEN=$(kubectl get --namespace consul secrets/consul-consul-bootstrap-acl-token --template={{.data.token}} | base64 -d)
export CONSUL_HTTP_SSL=true
export CONSUL_HTTP_SSL_VERIFY=false
```
- enable a port forward to the consul server on port 8501
- run `consul acl role list` and you will see both terminating gateways with the term gateway policy and the zoidberg and nibbler policies
- run `consul acl policy read -name zoidberg-write-policy` to see the policy include the namespace
- shell into the bender service and run `curl localhost:1234` to see the request to zoidberg go through, run `curl localhost:5678` to see the request to nibbler go through
- now we'll test that acl policies are removed correctly
- edit the `termgw.yaml` file to no longer reference the zoidberg service and apply the file
- run `consul acl role list` and see the first terminating gateway no longer references the zoidberg policy
- run `consul acl policy list` and you'll see the zoidberg policy still exists
- now edit the termgw2.yaml file to no longer reference the zoidberg service and apply that file
- now run `consul acl role list` and see that none of the terminating gateways reference that policy
- now run `consul acl policy list` and see that the zoidberg policy no longer exists because no gateway is referencing it

### Checklist ###
- [X] Tests added


acceptance/framework/k8s/deploy.go
```diff
@@ -165,10 +165,10 @@ func CheckStaticServerConnectionSuccessfulWithMessage(t *testing.T, options *k8s
 
 // CheckStaticServerConnectionSuccessful is just like CheckStaticServerConnection
 // but it always expects a successful connection.
-func CheckStaticServerConnectionSuccessful(t *testing.T, options *k8s.KubectlOptions, sourceApp string, curlArgs ...string) {
```
jm96441n, line 168: modified this argument to make it more clear what options it should be for (in multi-namespace scenarios)

acceptance/framework/helpers/helpers.go
```diff
@@ -358,3 +384,27 @@ func createCmdArgs(options *k8s.KubectlOptions) []string {
 	}
 	return cmdArgs
 }
+
+// WaitForInput starts a http server on a random port (which is output in the logs) and waits until you
+// issue a request to that endpoint to continue the tests. This is useful for debugging tests that require
+// inspecting the current state of a running cluster and you don't need to use long sleeps
+func WaitForInput(t *testing.T) {
```
jm96441n, line 391: this was super useful for debugging the state of the cluster during tests and not needing to put a random sleep time in the test

jm96441n: moved this into it's own PR https://github.com/hashicorp/consul-k8s/pull/4225

control-plane/catalog/registration/cache.go
```diff
@@ -231,9 +227,9 @@ func (c *RegistrationCache) updateTermGWACLRole(log logr.Logger, registration *v
 	var data bytes.Buffer
 	if err := gatewayTpl.Execute(&data, templateArgs{
 		EnablePartitions: c.partitionsEnabled,
-		Partition:        registration.Spec.Service.Partition,
+		Partition:        defaultIfEmpty(registration.Spec.Service.Partition),
```
jm96441n, line 230: these bits here fix the bug, the rest is just updating tests

acceptance/tests/config-entries/config_entries_namespaces_test.go
```diff
@@ -90,6 +90,10 @@ func TestControllerNamespaces(t *testing.T) {
 
 				"global.acls.manageSystemACLs": strconv.FormatBool(c.secure),
 				"global.tls.enabled":           strconv.FormatBool(c.secure),
+
```
jm96441n, line 93: this test exercises the terminating gateway path and those don't function correctly without this in your helm chart

control-plane/catalog/registration/cache.go
```diff
@@ -4,60 +4,20 @@
...
 	"github.com/hashicorp/consul-k8s/control-plane/consul"
 	capi "github.com/hashicorp/consul/api"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const NotInServiceMeshFilter = "ServiceMeta[\"managed-by\"] != \"consul-k8s-endpoints-controller\""
 
-func init() {
```
jm96441n, line 28: all this moved to the term gateway controller

control-plane/catalog/registration/cache.go
```diff
@@ -154,18 +117,28 @@ func (c *RegistrationCache) run(log logr.Logger, namespace string) {
...
 			for _, svc := range servicesToAdd.ToSlice() {
-				registration := &v1alpha1.Registration{}
+				log.Info("consul registered service", "svcName", svc)
+				registrationList := &v1alpha1.RegistrationList{}
+
+				if err := c.k8sClient.List(context.Background(), registrationList, client.MatchingFields{registrationByServiceNameIndex: svc}); err != nil {
+					log.Error(err, "error listing registrations", "svcName", svc)
+				}
 
-				if err := c.k8sClient.Get(c.ctx, types.NamespacedName{Name: svc, Namespace: namespace}, registration); err != nil {
```
jm96441n, line 159: the service name here doesn't necessarily map directly to the registration name

control-plane/catalog/registration/registrations_controller.go
```diff
@@ -44,12 +44,12 @@ type RegistrationsController struct {
 	Log    logr.Logger
 }
 
-// +kubebuilder:rbac:groups=consul.hashicorp.com,resources=servicerouters,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=consul.hashicorp.com,resources=servicerouters/status,verbs=get;update;patch
```
jm96441n, line 48: found this copy/pasta mistake

control-plane/catalog/registration/registrations_controller.go
```diff
@@ -44,12 +44,12 @@ type RegistrationsController struct {
...
+// +kubebuilder:rbac:groups=consul.hashicorp.com,resources=registration,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=consul.hashicorp.com,resources=registration/status,verbs=get;update;patch
 
 func (r *RegistrationsController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.V(1).WithValues("registration", req.NamespacedName)
-	log.Info("Reconciling Registaration")
```
jm96441n, line 52: fixing typo

control-plane/controllers/configentries/configentry_controller_test.go
```diff
@@ -543,48 +544,59 @@ func TestConfigEntryControllers_createsConfigEntry(t *testing.T) {
 	}
 
 	for _, c := range cases {
-		t.Run(c.kubeKind, func(t *testing.T) {
-			req := require.New(t)
-			ctx := context.Background()
+		for _, secure := range []bool{true, false} {
```
jm96441n, line 547: updated these tests to run in secure and non secure mode to ensure config entries are created when acls are enabled

acceptance/framework/helpers/helpers.go
```diff
@@ -382,26 +368,27 @@ func WaitForInput(t *testing.T) {
...
-		err = srv.Shutdown(context.Background())
+		err = r.Body.Close()
 		if err != nil {
-			t.Logf("error closing listener: %v", err)
+			t.Logf("error closing request body: %v", err)
+			err = nil
 		}
 
 		t.Log("input received, continuing test")
+		go func() {
```
jm96441n, line 386: this was hanging if we didn't do it async

control-plane/catalog/registration/registrations_controller.go
```diff
@@ -60,18 +60,6 @@ func (r *RegistrationsController) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	cachedRegistration, ok := r.Cache.get(registration.Spec.Service.Name)
```
jm96441n, line 63: moved this down so we handle deletion correctly

control-plane/catalog/registration/registrations_controller.go
```diff
@@ -143,48 +143,9 @@ func (r *RegistrationsController) handleRegistration(ctx context.Context, log lo
 		return result
 	}
 
-	if r.Cache.aclsEnabled() {
```
jm96441n, line 146: all the acl updates are done on the terminating gateway now to remove this controller racing with that one

acceptance/tests/fixtures/cases/terminating-gateway-namespaces/all-non-default/external-service-registration/external-service.yaml
```diff
@@ -5,7 +5,6 @@ apiVersion: consul.hashicorp.com/v1alpha1
 kind: Registration
 metadata:
   name: static-server-registration
```
missylbytes, line 7: Is this just because we don't care where the registration actually lives on the cluster?

control-plane/subcommand/inject-connect/v1controllers.go
```diff
@@ -254,8 +254,9 @@ func (c *Command) configureControllers(ctx context.Context, mgr manager.Manager,
 		ConfigEntryController: configEntryReconciler,
 		Client:                mgr.GetClient(),
 		Log:                   ctrl.Log.WithName("controller").WithName(apicommon.TerminatingGateway),
+		NamespacesEnabled:     c.flagEnableNamespaces,
```
missylbytes, line 257: Did we just not enable namespaces for terminating gateways before?

missylbytes: Oh I didn't realize we were moving functionality from the other controller.

acceptance/tests/fixtures/cases/terminating-gateway-namespaces/all-non-default/external-service-registration/external-service.yaml
jm96441n: yeah, you can have a service in your k8s cluster that is no part of your mesh and have a terminating gateway front it which is what we're doing here

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -4,33 +4,108 @@
...
+	EnableNamespaces bool
+}
+
+var (
+	servicePolicyTpl      *template.Template
+	servicePolicyRulesTpl = `
+{{- if .EnableNamespaces }}
+namespace "{{.Namespace}}" {
+{{- end }}
+  service "{{.ServiceName}}" {
```
jm96441n, line 61: this doesn't work for wildcard services, need to update the template here

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -41,6 +116,221 @@ func (r *TerminatingGatewayController) UpdateStatus(ctx context.Context, obj cli
...
+			terminatingGatewayRoleID = role.ID
+			break
+		}
+	}
+
+	if terminatingGatewayRoleID == "" {
+		return errors.New("terminating gateway role not found")
+	}
+
+	termGwRole, _, err := client.ACL().RoleRead(terminatingGatewayRoleID, nil)
```
nathancoleman, line 177: #nit feels like we might as well go with `terminatingGatewayRole` based on the variable names above

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -41,6 +116,221 @@ func (r *TerminatingGatewayController) UpdateStatus(ctx context.Context, obj cli
...
+	if err != nil {
+		return err
+	}
+
+	var termGWPolicy *capi.ACLRolePolicyLink
```
nathancoleman, line 182: #nit similarly here

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -41,6 +116,221 @@ func (r *TerminatingGatewayController) UpdateStatus(ctx context.Context, obj cli
...
+	}
+
+	roles, _, err := client.ACL().RoleList(nil)
+	if err != nil {
+		return err
+	}
+
+	terminatingGatewayRoleID := ""
+	for _, role := range roles {
+		if strings.Contains(role.Name, termGW.Name) {
```
jm96441n, line 167: this should be `HasSuffix(fmt.Sprintf("%s-acl-role", termGW.Name))`

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -41,6 +116,221 @@ func (r *TerminatingGatewayController) UpdateStatus(ctx context.Context, obj cli
...
+
+	termGwRole, _, err := client.ACL().RoleRead(terminatingGatewayRoleID, nil)
+	if err != nil {
+		return err
+	}
+
+	var termGWPolicy *capi.ACLRolePolicyLink
+
+	for _, policy := range termGwRole.Policies {
+		if strings.Contains(policy.Name, termGW.Name) {
```
jm96441n, line 185: this should be `HasSuffix(fmt.Sprintf("%s-policy", termGW.Name))`

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -41,6 +116,221 @@ func (r *TerminatingGatewayController) UpdateStatus(ctx context.Context, obj cli
...
+	var termGWPoliciesToRemove []*capi.ACLRolePolicyLink
+
+	existingTermGWPolicies := mapset.NewSet[string]()
+
+	for _, policy := range termGwRole.Policies {
+		existingTermGWPolicies.Add(policy.Name)
+	}
+
+	if termGW.ObjectMeta.DeletionTimestamp.IsZero() {
+		log.Info("terminating gateway is being modified")
```
nathancoleman, line 201: are these types of logs being intentionally left in for merge?

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -41,6 +116,221 @@ func (r *TerminatingGatewayController) UpdateStatus(ctx context.Context, obj cli
...
+		var data bytes.Buffer
+		if err := servicePolicyTpl.Execute(&data, templateArgs{
+			EnableNamespaces: r.NamespacesEnabled,
+			Namespace:        defaultIfEmpty(service.Namespace),
+			ServiceName:      service.Name,
+		}); err != nil {
+			// just panic if we can't compile the simple template
+			// as it means something else is going severly wrong.
+			panic(err)
+		}
```
nathancoleman, line 254: #nit I _think_ this could be moved down into the `if existingPolicy == nil` block

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -41,6 +116,221 @@ func (r *TerminatingGatewayController) UpdateStatus(ctx context.Context, obj cli
...
+			continue
+		}
+		if len(termGWList.Items) == 0 {
+			policiesToDelete = append(policiesToDelete, policy)
+		}
+	}
+
+	for _, policy := range policiesToDelete {
+		// don't delete the policy for the gateway itself
+		if strings.Contains(policy.Name, termGWName) {
```
nathancoleman, line 302: Should this be a suffix check instead?

missylbytes (approved):

.changelog/4224.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+terminating-gateways: Fix bug where namespace field was not being correctly set on acl policies when using the registration CRD and the registered service
+is in the default namespace and the namespace field is unset
+```
```
nathancoleman, line 4: I think we could simplify this. Careful if accepting this suggested change since the backtick patterns seem to throw it off.

```suggestion
```release-note:bug
terminating-gateways: Fix bug where namespace field was not correctly set on ACL policies if using the `Registration` CRD with the service's namespace unset.
```
```

File: acceptance/go.mod [hunk 24]
```diff
@@ -29,6 +29,11 @@ require (
 	sigs.k8s.io/gateway-api v0.7.1
 )
 
+replace (
+	github.com/hashicorp/consul-k8s/control-plane => ../control-plane
```
nathancoleman, line 33: Should this remain this way long term?

acceptance/tests/config-entries/config_entries_namespaces_test.go
```diff
@@ -90,6 +90,10 @@ func TestControllerNamespaces(t *testing.T) {
...
+				"terminatingGateways.enabled":              "true",
+				"terminatingGateways.gateways[0].name":     "terminating-gateway",
```
nathancoleman, line 95: 🧠  Curious how this works if `terminatingGateways` isn't already set to a list with >= 1 items (or is it?)

nathancoleman (commented):
Some thoughts after walking through the PR with @jm96441n on a call a few minutes ago

control-plane/controllers/configentries/terminatinggateway_controller.go
jm96441n: they should probably be downgraded to debug logs

acceptance/tests/config-entries/config_entries_namespaces_test.go
jm96441n: do you mean this specific string or do you mean how the code being tested behaves?

acceptance/go.mod
jm96441n: honestly I'm not sure, the acceptance tests should be pulling in the latest version of the code they're working with when testing, as of now if we add a new type to the control-plane we have to publish that first and then pull in the new version here, so the acceptance tests will always have to lag a bit against what is published

control-plane/controllers/configentries/terminatinggateway_controller.go
jm96441n: well the logr.Logger doesn't do debug logs sooo going to remove them

acceptance/tests/config-entries/config_entries_namespaces_test.go
nathancoleman: walked through on a call. setting this here isn't actually _updating_ something but instead uses magic to _create_ something behind the scenes

control-plane/controllers/configentries/terminatinggateway_controller.go
```diff
@@ -4,33 +4,122 @@
...
+}
+{{- end }}
+`
+
+	wildcardPolicyTpl      *template.Template
+	wildcardPolicyRulesTpl = `
+{{- if .EnableNamespaces }}
+namespace "{{.Namespace}}" {
+{{- end }}
+  service_preifx "" {
```
nathancoleman, line 75:
```suggestion
  service_prefix "" {
```

nathancoleman (approved):

natemollica-nm:
@jm96441n -- Awesome update to the TGW workflow! 

Can you ensure that this gets updated though to account for enabling Admin Partitions -- Testing this I had to manually (tproxy destinations method) update the ACL policy for one of my external services from:

```hcl
namespace "default" {
  service "example" {
    policy = "write"
  }
}
```

to 

```hcl
partition "default" {
  namespace "default" {
    node_prefix "" {
      policy = "read"
    }
    service "example" {
      policy    = "write"
      intention = "write"
    }
  }
}
```

Additionally, I'm positive the `node:read` and `intention:read` permissions are required for these policies (which is definitely an inconsistency in our docs):

* [Create a Terminating Gateway ACL Token: Requirements](https://developer.hashicorp.com/consul/docs/security/acl/tokens/create/create-a-terminating-gateway-token#requirements)
* Envoy Traffic Filter with Terminating Gateway | [Blake's Comment](https://github.com/hashicorp/consul/issues/12116#issuecomment-1019463753)
* [Document - Transparent Proxy + Terminating Gateway](https://github.com/hashicorp/consul-k8s/issues/1486)

PR #4225, opened 2024-08-01
Update acceptance test docs, add "WaitForInput" function

jm96441n:
- Minor updates to acceptance tests documentation
- Adds `WaitForInput` function to pause test execution at points and check on status of cluster

### Checklist ###
- [ ] Tests added

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -961,17 +961,23 @@ The tests are organized like this :
 ```shell
 demo $ tree -L 1 -d acceptance/tests
 acceptance/tests
+├── api-gateway
 ├── basic
 ├── cli
+├── cloud
 ├── config-entries
 ├── connect
 ├── consul-dns
+├── datadog
 ├── example
 ├── fixtures
 ├── ingress-gateway
 ├── metrics
 ├── partitions
 ├── peering
+├── sameness
+├── segments
+├── server
 ├── snapshot-agent
 ├── sync
 ├── terminating-gateway
@@ -1009,7 +1015,9 @@ $ kind create cluster --name=dc1 && kind create cluster --name=dc2
   `-consul-k8s-image=<your-custom-image>` && `-consul-image=<your-custom-image>`
 * You can set custom helm flags by modifying the test file directly in the respective directory.
 
-Finally, run the test like shown above:
+Finally, you have two options on how you can run your test:
+1. Take the following steps, this will run the test through to completion but not teardown any resources created by the test so you can inspect the state of the cluster
+at that point. You will be responsible for cleaning up the resources or deleting the cluster entirely when you're done.
 ```shell
 $ cd acceptance/tests
 $ go test -run Vault_WANFederationViaGateways ./vault/... -p 1 -timeout 2h -failfast -use-kind -no-cleanup-on-failure -kubecontext=kind-dc1 -secondary-kubecontext=kind-dc2 -enable-multi-cluster -debug-directory=/tmp/debug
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1018,6 +1026,22 @@ You can interact with the running kubernetes clusters now using `kubectl [COMMAN
 
 * `kind delete clusters --all` is helpful for cleanup!
 
+2. The other option is to use the helper method in the framework: `helpers.WaitForInput(t)` at the spot in your acceptance test where you would like to pause execution to inspect the cluster. This will pause the test execution until you execute a request to `localhost:8715` which tells the test to continue running, you can override the port value used by setting the `CONSUL_K8S_TEST_PAUSE_PORT` environment variable to a port of your choosing. When running the tests with the `-v` flag you will see a log output of the endpoint that the test is waiting on.
+
+```go
+
```
zalimeni, line 1032:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -1018,6 +1026,22 @@ You can interact with the running kubernetes clusters now using `kubectl [COMMAN
...
+import "github.com/hashicorp/consul-k8s/acceptance/framework/helpers"
+
+func TestSomeTest(t *testing.T) {
```
zalimeni, line 1035: ~ Looks like a mix of tabs and spaces here. Probably fixed by most IDEs on paste but could shore up now

CONTRIBUTING.md
```diff
@@ -1009,7 +1015,9 @@ $ kind create cluster --name=dc1 && kind create cluster --name=dc2
   `-consul-k8s-image=<your-custom-image>` && `-consul-image=<your-custom-image>`
 * You can set custom helm flags by modifying the test file directly in the respective directory.
 
-Finally, run the test like shown above:
+Finally, you have two options on how you can run your test:
+1. Take the following steps, this will run the test through to completion but not teardown any resources created by the test so you can inspect the state of the cluster
+at that point. You will be responsible for cleaning up the resources or deleting the cluster entirely when you're done.
```
zalimeni, line 1020: I think we should probably call out the specific flags that make this true - from what I can see, most of the command below is common instruction, we're just offering an alternative to `-no-cleanup-on-failure`, right?

acceptance/framework/helpers/helpers.go
```diff
@@ -358,3 +360,48 @@ func createCmdArgs(options *k8s.KubectlOptions) []string {
 	}
 	return cmdArgs
 }
+
+const DEFAULT_PAUSE_PORT = ":8517"
```
zalimeni, line 364: ~ Should we consider a higher port range for this than `8500`s, so that we're out of the "immediately likely next ports" that Consul/related tools might use in normal prod operations? e.g. `38501`

acceptance/framework/helpers/helpers.go
```diff
@@ -358,3 +360,48 @@ func createCmdArgs(options *k8s.KubectlOptions) []string {
...
+	envPort := os.Getenv("CONSUL_K8S_TEST_PAUSE_PORT")
+	listenerPort := DEFAULT_PAUSE_PORT
+
+	if envPort != "" {
+		listenerPort = ":" + envPort
+	}
+
+	mux := http.NewServeMux()
+	srv := &http.Server{
+		Addr:    listenerPort,
```
zalimeni, line 381: ~ Maybe make the default port a number/numeric string like the override for consistency? (could be referenced as a number somewhere else later)
```suggestion
	listenerPort := os.Getenv("CONSUL_K8S_TEST_PAUSE_PORT")
	if listenerPort == "" {
		listenerPort = DEFAULT_PAUSE_PORT
	}

	mux := http.NewServeMux()
	srv := &http.Server{
		Addr:    ":" + listenerPort,
```

CONTRIBUTING.md
jm96441n: we can call them out specifically, we have a list of the flags and what they do earlier in this doc as well https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#pre-requisites

and yeah the `WaitForInput` is similar to `--no-cleanup-on-failure` but allows you to inspect the cluster at different points of the setup rather than the state it's in when the test fails

zalimeni: Ok cool. For this doc change, my main thinking was this blurb implies now there's two divergent paths, but really, only the first one tells you what command to run.

So we should probably either have a single command without the flag and options 1 + 2 to use it or `WaitForInput`, or just have two similar commands, one per option. Does that sound right?

jm96441n: that makes sense, let me update this to show how that would work

jm96441n: updated the text a bit to show how the two different paths work

zalimeni: Nice, LGTM, thank you!

zalimeni (approved):
This seems like a really useful and low-touch QoL improvement for local dev - thanks for coming up w/ it, and adding docs to go with 👏🏻

Issue #4208, opened 2024-07-24
Helm chart's SCC should have volumes in alphabetical order


kbreit-insight:
### Overview of the Issue

The list of volumes at https://github.com/hashicorp/consul-k8s/blob/1b04a36d64e09410b3aae4f83b135b9fa3bb98ec/charts/consul/templates/cni-securitycontextconstraints.yaml#L42 should be in alphabetical order. The problem this causes is I am deploying Consul via helm chart in ArgoCD and it won't fully synchronize because it's expecting the `hostPath` to be at the bottom. However, Kubernetes appears to sort them as here's the current resource section:

```yaml
volumes:
- configMap
- downwardAPI
- emptyDir
- hostPath
- persistentVolumeClaim
- projected
- secret
```

### Reproduction Steps

Deploy Consul on OpenShift with the CNI enabled. Optionally, deploy via Argo but that should only show it's being applied in a different order.

### Expected behavior

I would expect the order to stay the same.

### Environment details
OpenShift 4.15 (Kubernetes 1.28.x)
Consul 1.19
Helm chart v1.5.1

zalimeni:
Thanks for the report, @kbreit-insight !

Based on some initial poking into OpenShift source, I think what you're suggesting makes sense - though the silent forced sorting in OpenShift is (AFAICT) an undocumented behavior, and makes use cases like Argo's diffing unnecessarily brittle. It seems there's an Argo [feature request](https://github.com/argoproj/argo-cd/issues/12033) to address cases like this, which could help in the future.

Side note: interestingly, the order `consul-k8s` is using is _almost_ based on OpenShift behavior, but seems a bit off:
* This code: https://github.com/openshift/kubernetes/blob/6892b57d65d25fb0588693bce3d338d8b8b1d2b4/openshift-kube-apiserver/admission/customresourcevalidation/securitycontextconstraints/defaults.go#L51 will by default add `hostPath` to the end of the array if it’s missing, and the `allowHostDirVolumePlugin` flag is set
  * That, combined w/ the [wrapping deserialize/serialize logic](https://github.com/openshift/kubernetes/blob/6892b57d65d25fb0588693bce3d338d8b8b1d2b4/openshift-kube-apiserver/admission/customresourcevalidation/securitycontextconstraints/defaulting_scc.go#L47) should normalize the data, forcing a sort on existing values.
[...]

PR #4227, opened 2024-08-02
[NET-10547] openshift: re-order SCC volume list for Argo sync

zalimeni:
Due to logic in OpenShift's admissions hook that force-reorders explicit and implicit entries in this list, our `SecurityContextConstraints` entries will never successfully sync via tools like ArgoCD, which expect an exact input and output match when diff'ing.

More details on the problem addressed by this change and potential future improvements to avoid it in the future can be found in https://github.com/hashicorp/consul-k8s/issues/4208 (see comments).

- Reorder SCC `volumes` entries to match the post-reconciliation order persisted by OpenShift

- Spun up a local OpenShift CRC cluster w/ ArgoCD (GitOps Operator), deployed original and modified charts to verify fix results in a fully synced `consul-cni` SCC.
- Tested on Azure using community ArgoCD operator, same results.

👀 

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.changelog/4227.txt
+++ b/.changelog/4227.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+openshift: order SecurityContextConstraint volumes alphabetically to match OpenShift behavior.
+This ensures that diff detection tools like ArgoCD consider the source and reconciled resources to be identical.
+```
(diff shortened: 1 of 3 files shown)


ndhanushkodi (approved):
🎉

PR #4244, opened 2024-08-12
bump kind and kubectl versions and update k8s support

NicoletaPopoviciu:
bump kind and kubectl versions, and update k8s versions tested

- bump kind and kubectl versions
- update k8s support

### Checklist ###
- [ ] Tests added

### DIFF
--- a/acceptance/ci-inputs/kind-inputs.yaml
+++ b/acceptance/ci-inputs/kind-inputs.yaml
@@ -1,6 +1,6 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
-kindVersion: v0.22.0
+kindVersion: v0.23.0
 kindNodeImage: kindest/node:v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245
-kubectlVersion: v1.27.1
+kubectlVersion: v1.30.2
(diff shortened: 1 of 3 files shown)


acceptance/ci-inputs/kind-inputs.yaml
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
-kindVersion: v0.22.0
+kindVersion: v0.23.0
 kindNodeImage: kindest/node:v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245
```
zalimeni, line 5: This will need to be bumped as well as it's what actually controls the K8s version - we can use the latest:

`kindest/node:1.30.2@sha256:ecfe5841b9bee4fe9690f49c118c33629fa345e3350a0c67a5a34482a99d6bba`

zalimeni (commented):
One more edit we should make, otherwise LGTM!

The acceptance tests that run for this PR should use the updated values automatically, we can observe them being updated here for the latest commit (note the `kindest/node` image version isn't updated yet): https://github.com/hashicorp/consul-k8s-workflows/actions/runs/10360967522/job/28680500788#step:1:141

NicoletaPopoviciu:
@zalimeni thank you ; committed your suggestion

zalimeni:
> @zalimeni thank you ; committed your suggestion. ? re backports - should this only be backported to 1.5.x?

Sorry, overlooked this while reviewing. I think yes to this change, then we'll want some targeted changes similar to it, but slightly different, for `release/1.4.x` and `release/1.1.x` (docs for those should not say "1.27-1.30" since LTS versions don't typically remove old support, just add new). The Jira ticket has a few notes on those versions, but happy to discuss as well.

PR #4255, opened 2024-08-15
remove all k8s services from consul

wangxinyi7:
- Support removing the registered K8S services from consul when user runs `consul sync-catalog -purge-k8s-services-from-node=k8s-sync`.

1. Start with disabling syncCatalog
2. Enabling syncCatalog: Confirmed from Consul UI that K8S services registered in Consul
3. Disabling syncCatalog: registered services still in the dashboard
4. Run `consul sync-catalog -purge-k8s-services-from-node=k8s-sync`
5. Confirmed from Consul UI that registered K8S services are de-registered in Consul

### Checklist ###
- [x] Tests added

[NET-10084]: https://hashicorp.atlassian.net/browse/NET-10084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


control-plane/subcommand/sync-catalog/command.go
```diff
@@ -251,6 +252,13 @@ func (c *Command) Run(args []string) int {
 	}
 	c.ready = true
 
+	if !c.flagToConsul && !c.flagToK8S {
+		if err := c.removeAllK8SServicesFromConsul(consulConfig); err != nil {
```
jm96441n, line 256: can we log the error here?

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -393,6 +401,22 @@ func (c *Command) Run(args []string) int {
...
 }
 
+// remove all k8s services from Consul.
+func (c *Command) removeAllK8SServicesFromConsul(consulConfig *consul.Config) error {
+	consulClient, err := consul.NewClientFromConnMgr(consulConfig, c.connMgr)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("unable to create Consul API client: %s", err))
+		return err
+	}
+	_, err = consulClient.Catalog().Deregister(&api.CatalogDeregistration{Node: c.flagConsulNodeName}, nil)
```
jm96441n, line 411: do we need to care about namespaces at all here?

jm96441n: do we need to be concerned about the blast radius for deregistering the entire node here? will that affect the consul service registration?

charts/consul/templates/gateway-resources-job.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
   annotations:
     "helm.sh/hook": post-install,post-upgrade
     "helm.sh/hook-weight": "0"
-    "helm.sh/hook-delete-policy": hook-succeeded
```
jm96441n, line 20: why did we need to update the gateway resources job here?

charts/consul/templates/sync-catalog-cleanup-k8s-svc-in-consul-job.yaml
```diff
@@ -0,0 +1,112 @@
...
+          - name: consul-ca-cert
+            mountPath: /consul/tls/ca
+            readOnly: true
+        {{- end }}
+        {{- end }}
+        command:
+          - consul-k8s-control-plane
+        args:
+          - sync-catalog
+          - -to-consul=false
```
jm96441n, line 99: are there any dangers reusing these rather than utilizing a new commend to deregister all services?

control-plane/subcommand/sync-catalog/command.go
wangxinyi7: Yup, adjusted the place to log the error

charts/consul/templates/gateway-resources-job.yaml
wangxinyi7: It was not deleted after failed during my test, but now, it should be gone

charts/consul/templates/sync-catalog-cleanup-k8s-svc-in-consul-job.yaml
wangxinyi7: Right, let me create a dedicated flag just in case of messing up other scenarios.

control-plane/subcommand/sync-catalog/command.go
wangxinyi7: 1. I distribute the de-register work into batches to avoid the traffic blaster
2. I added the `filter` flag to allow the user to control the de-register process in more granular level

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -393,6 +415,85 @@ func (c *Command) Run(args []string) int {
 	}
 }
 
+// remove all k8s services from Consul.
+func (c *Command) removeAllK8SServicesFromConsulNode(consulClient *api.Client, nodeName string) error {
+	node, _, err := consulClient.Catalog().NodeServiceList(nodeName, &api.QueryOptions{Filter: c.flagFilter})
+	if err != nil {
+		return err
+	}
```
blake, line 423: It might be more efficient to use [`Catalog().Services()`](https://pkg.go.dev/github.com/hashicorp/consul/api#Catalog.Services) to obtain the list of services when no user-defined filter is provided. Pass in a filter such as `Node=="<nodeName>"` to return a list of services that exist on that node. Count the number of logical services on that node and prompt for user confirmation before deleting the node. This will purge all of the instances associated with the previously returned logical services that exist on that node.

If the user provides their own filter, use `Catalog().NodeServiceList()` query the list of service *instances* that are registered against that node and that match the specified filter criteria.

wangxinyi7: Thanks for review! @blake 
I checked the API for [listing services](https://developer.hashicorp.com/consul/api-docs/catalog#list-services), it looks like the response is the structure of [map[string][]string](https://github.com/hashicorp/consul/blob/main/api/catalog.go#L212) whose keys are [service names](https://developer.hashicorp.com/consul/api-docs/catalog#sample-response-2). However, in our de-registration request, we need to have the information of [serviceID](https://developer.hashicorp.com/consul/api-docs/catalog#serviceid). So I think in our situation, [Catalog().NodeServiceList()](https://developer.hashicorp.com/consul/api-docs/catalog#list-services-for-node) suites better to our needs. Please point out something I overlooked here, thank you!

blake: You don't need the service ID when de-registering a node. You need to provide the node ID. All services on the node will then be removed. <https://developer.hashicorp.com/consul/api-docs/catalog#node-1>

> Specifies the ID of the node. If no other values are provided, this node, all its services, and all its checks are removed.

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -138,6 +143,10 @@ func (c *Command) init() {
 			"\"debug\", \"info\", \"warn\", and \"error\".")
 	c.flags.BoolVar(&c.flagLogJSON, "log-json", false,
 		"Enable or disable JSON output format for logging.")
+	c.flags.StringVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", "",
```
NiniOak, line 146: I'm wondering if this should be a `c.flags.BoolVar`, what kind of string are we expecting users to provide to purge k8s service? Or simply providing this flag will initiate a purge

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -393,6 +415,85 @@ func (c *Command) Run(args []string) int {
...
+						Node:      nodeName,
+						ServiceID: service.ID,
+					}, nil)
+					return err
+				}, maxRetries, retryDelay, c.logger)
+				if err != nil {
+					if len(errChan) == 0 {
+						errChan <- err
+					}
+				}
```
nathancoleman, line 469: We use an existing backoff-retry library elsewhere in the codebase that can handle this

```suggestion
				var b backoff.BackOff = backoff.NewConstantBackOff(200 * time.Millisecond)
				b = backoff.WithMaxRetries(b, 2)
				backoff.Retry(func() error {
					_, err := consulClient.Catalog().Deregister(&api.CatalogDeregistration{
						Node:      nodeName,
						ServiceID: service.ID,
					}, nil)
					return err
				}, b)
```

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -393,6 +415,85 @@ func (c *Command) Run(args []string) int {
...
+			}
+			c.UI.Info(fmt.Sprintf("Processed %v K8S services from %v", len(batch), nodeName))
+		}(services[i:end])
+		wg.Wait()
+	}
+
+	close(errChan)
+	if err = <-errChan; err != nil {
+		return err
+	}
```
nathancoleman, line 479: I find the `errgroup` package ([docs](https://pkg.go.dev/golang.org/x/sync/errgroup#Group)) super helpful for this kind of pattern. It's basically a `WaitGroup` plus error handling

```go
var eg errgroup.Group
for ... {
	eg.Go(func() error {
		err := doSomething()
		return err
	})
}

// eg.Wait() will wait for all of the goroutines above to complete
// or for one of them to return an error, whichever happens first
if err := eg.Wait(); err != nil {
	return err
}
```

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -393,6 +415,85 @@ func (c *Command) Run(args []string) int {
...
+	c.UI.Info("All K8S services were deregistered from Consul")
+	return nil
+}
+
+func retryOps(operation func() error, maxRetries int, retryDelay time.Duration, logger hclog.Logger) error {
```
nathancoleman, line 484: If we use `backoff.Retry` discussed above, then we no longer need this function

control-plane/subcommand/sync-catalog/command_test.go
```diff
@@ -571,6 +572,164 @@ func TestRun_ToConsulChangingFlags(t *testing.T) {
...
+
+	longRunningChan := runCommandAsynchronously(&cmd, []string{
+		"-addresses", "127.0.0.1",
+		"-http-port", strconv.Itoa(testClient.Cfg.HTTPPort),
+		"-consul-write-interval", "100ms",
+		"-add-k8s-namespace-suffix",
+	})
+	defer stopCommand(t, &cmd, longRunningChan)
+
+	// check that the name of the service is namespaced
```
nathancoleman, line 625:
```suggestion
	// check that the two K8s services have been synced into Consul
```

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -393,6 +415,85 @@ func (c *Command) Run(args []string) int {
...
+		input, _ := reader.ReadString('\n')
+		input = strings.TrimSpace(input)
+		if input == "y" || input == "Y" {
+			break
+		} else if input == "n" || input == "N" {
+			return nil
+		} else {
+			c.UI.Info("Invalid input. Please enter 'y' or 'n'.")
+		}
+	}
```
nathancoleman, line 445: I think this could be cleaner using a `strings.ToLower()` and `fmt.Scanln` ([example code](https://go.dev/play/p/gua5HUyPeEq))

nathancoleman (changes requested):
Great work! Just a few suggestions for simplifying things

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -138,6 +143,10 @@ func (c *Command) init() {
 			"\"debug\", \"info\", \"warn\", and \"error\".")
 	c.flags.BoolVar(&c.flagLogJSON, "log-json", false,
 		"Enable or disable JSON output format for logging.")
+	c.flags.StringVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", "",
+		"Purge all K8S services registered in Consul under the node name.")
+	c.flags.StringVar(&c.flagFilter, "filter", "",
+		"Specifies the expression used to filter the queries results for the node.")
```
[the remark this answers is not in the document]
wangxinyi7: Since we use the [List Services for Node](https://developer.hashicorp.com/consul/api-docs/catalog#list-services-for-node)API, the filter syntax is the same to that [API](https://developer.hashicorp.com/consul/api-docs/catalog#filtering-4), should I highlight this somewhere in the flag description?

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -393,6 +415,85 @@ func (c *Command) Run(args []string) int {
 	}
 }
 
+// remove all k8s services from Consul.
+func (c *Command) removeAllK8SServicesFromConsulNode(consulClient *api.Client, nodeName string) error {
```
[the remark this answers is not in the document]
wangxinyi7: right, good catch, updated

control-plane/subcommand/sync-catalog/command.go
wangxinyi7: Thank you!

control-plane/subcommand/sync-catalog/command.go
wangxinyi7: Yup, thanks! It looks more cleaner!

control-plane/subcommand/sync-catalog/command.go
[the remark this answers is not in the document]
wangxinyi7: I added the details in the flag explanation

control-plane/subcommand/sync-catalog/command.go
wangxinyi7: Thanks for the reference and the attached example. There is a little nuance here.

- The tasks are divided into batches and executed batch by batch in sequence.
- Since the tasks executed here are not critical, in each batch, they are allowed to fail and the `errChan` only collects the first failure.
- By the end of the whole tasks, it just highlights to the operator that not all of services were successfully de-registered and here is one example failure error.

This means, even though we introduced the errgroup, we still need to keep the `errChan` to collect the failure(The reason I didn't collect all failure is that the number of services is potentially a large number) and we won't gain many benefits for doing this.

nathancoleman: The `<Group>.Wait()` function does what you're describing for the channel I believe -- it returns the first non-nil error of any function that was run in the group ([docs](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Wait))

wangxinyi7: Yeah right, the tricky part here is that, <Group>.Go() execute the batches in parallel, but here we added work group to force the batches to be executed in sequence.

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -436,6 +459,73 @@ func (c *Command) Run(args []string) int {
...
+		}(services[i:end])
+		wg.Wait()
+	}
+
+	close(errChan)
+	if err = <-errChan; err != nil {
+		return err
+	}
+	c.UI.Info("All K8S services were deregistered from Consul")
+	return nil
```
nathancoleman, line 526: We discussed a bit outside of the PR that we _want_ to execute these requests in serial. Given that, I'm not sure why the `WaitGroup` and goroutine and error channel are necessary here. Why not just?

```suggestion
	var firstErr error
	for i := 0; i < len(services); i += batchSize {
		end := i + batchSize
		if end > len(services) {
			end = len(services)
		}
		
		batch := services[i:end]
		for _, service := range batch {
			var b backoff.BackOff = backoff.NewConstantBackOff(retryDelay)
			b = backoff.WithMaxRetries(b, uint64(maxRetries))
			err := backoff.Retry(func() error {
				_, err := consulClient.Catalog().Deregister(&api.CatalogDeregistration{
					Node:      c.flagPurgeK8SServicesFromNode,
					ServiceID: service.ID,
				}, nil)
				return err
			}, b)
			if err != nil {
				if firstErr == nil {
					firstErr = err
				}
			}
		}
		c.UI.Info(fmt.Sprintf("Processed %v K8S services from %v", len(batch), c.flagPurgeK8SServicesFromNode))
	}
	
	if firstErr != nil {
		return err
	}

	c.UI.Info("All K8s services were deregistered from Consul")
	return nil
```

nathancoleman (commented):
Some thoughts

control-plane/subcommand/sync-catalog/command.go
wangxinyi7: Updated the code to execute the batch in sequence but execute the tasks inside batch concurrently

control-plane/subcommand/sync-catalog/command.go
wangxinyi7: Updated the code to execute the batch in sequence but execute the tasks inside batch concurrently

NiniOak (approved):
PR looks good to me, well done 💯 
I'll wait from an approval from @nathancoleman before merging, as he had some comments earlier

nathancoleman (approved):

PR #4256, opened 2024-08-19
config entry: add validate_clusters to mesh config entry

ndhanushkodi:
- add validate_clusters to mesh config entry
- bump k8s dependencies in control-plane, cli, cni: controller-gen 0.14, k8s.io 0.29.8, controller-runtime 0.16
- not bumping k8s dependencies in acceptance because acceptance tests have PSP code
- go modules for 1.19.x line: 
[api/v1.29.4](https://github.com/hashicorp/consul/releases/tag/api%2Fv1.29.4)
[envoyextensions/v0.7.3](https://github.com/hashicorp/consul/releases/tag/envoyextensions%2Fv0.7.3)
[troubleshoot/v0.7.1](https://github.com/hashicorp/consul/releases/tag/troubleshoot%2Fv0.7.1)

- Unit tests
- End to end manual test
- 👀 

### Checklist ###
- [x] Tests added


control-plane/api/v1alpha1/mesh_types.go
```diff
@@ -60,6 +60,10 @@ type MeshSpec struct {
 	HTTP *MeshHTTPConfig `json:"http,omitempty"`
 	// Peering defines the peering configuration for the service mesh.
 	Peering *PeeringMeshConfig `json:"peering,omitempty"`
+	// ValidateClusters is false by default and configures whether Envoy proxies will validate clusters in a route. If
+	// set to true and any clusters in the route do not exist, the route table will not load. If set to false, the route
+	// table will load and routing to a non-existent cluster will result in a 404.
```
zalimeni, line 65: ~ Same comment as https://github.com/hashicorp/consul/pull/21587/files#r1722359714

zalimeni (approved):
LGTM pending `api` bump

Makefile
```diff
@@ -254,7 +254,7 @@ ifeq (, $(shell which controller-gen))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
```
ndhanushkodi, line 255: It turned out that when I ran make ctrl-generate and make ctrl-manifests probably one or both of those commands errored because I couldn't use the CRD when installing. I found via github issues that older versions of controller-gen can panic so bumped it here

charts/consul/templates/crd-controlplanerequestlimits.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
```
ndhanushkodi, line 3: Majority of these changes are whitespace-like things due to the controller gen bump

charts/consul/templates/crd-meshes.yaml
```diff
@@ -153,22 +151,31 @@ spec:
...
-                      not registered in Consul's catalog. If enabled, traffic will
-                      only be proxied to upstreams with service registrations in the
-                      catalog.
+                    description: |-
+                      MeshDestinationsOnly determines whether sidecar proxies operating in "transparent" mode can proxy traffic
+                      to IP addresses not registered in Consul's catalog. If enabled, traffic will only be proxied to upstreams
+                      with service registrations in the catalog.
                     type: boolean
                 type: object
+              validateClusters:
```
ndhanushkodi, line 160: This is the real change

control-plane/config/rbac/role.yaml
```diff
@@ -29,285 +29,19 @@ rules:
...
-  resources:
-  - servicesplitters/status
-  verbs:
-  - get
-  - patch
-  - update
-- apiGroups:
-  - consul.hashicorp.com
-  resources:
   - terminatinggateways
```
ndhanushkodi, line 45: This formatting is just consolidated now, you don't have to add the verbs for each resource type

Makefile
zalimeni: Sounds reasonable to bump this. ❓ when was the fix introduced? It looks like maybe `controller-gen` is intended for compatibility with a specific k8s.io version, so we may want to either upgrade to an earlier fix version, or bump our k8s.io dependencies as well: https://github.com/kubernetes-sigs/controller-tools/tree/main?tab=readme-ov-file#compatibility

(I’d lean toward bumping k8s.io but that could lead to a cascading bump of `controller-runtime` and other deps, which we’d ideally do with ample time to vet them)

ndhanushkodi: I see evidence in the convo here that controller-gen 0.14 could work, rather than bumping all the way to 0.16, do you think that's preferable, and we reduce the risk of things being super out of date?

zalimeni: Yeah I feel like the lowest viable / other-deps-aligned bump probably reduces risk now, then a fast follow ticket to get our dependencies updated fully in the next release (aligning w/ our K8s 1.30 support). WDYT?

ndhanushkodi: Yup that makes sense!

acceptance/go.sum
```diff
@@ -183,8 +184,8 @@ github.com/gruntwork-io/go-commons v0.8.0 h1:k/yypwrPqSeYHevLlEDmvmgQzcyTwrlZGRa
 github.com/gruntwork-io/go-commons v0.8.0/go.mod h1:gtp0yTtIBExIZp7vyIV9I0XQkVwiQZze678hvDXof78=
 github.com/gruntwork-io/terratest v0.46.7 h1:oqGPBBO87SEsvBYaA0R5xOq+Lm2Xc5dmFVfxEolfZeU=
 github.com/gruntwork-io/terratest v0.46.7/go.mod h1:6gI5MlLeyF+SLwqocA5GBzcTix+XiuxCy1BPwKuT+WM=
-github.com/hashicorp/consul/api v1.29.1 h1:UEwOjYJrd3lG1x5w7HxDRMGiAUPrb3f103EoeKuuEcc=
-github.com/hashicorp/consul/api v1.29.1/go.mod h1:lumfRkY/coLuqMICkI7Fh3ylMG31mQSRZyef2c5YvJI=
+github.com/hashicorp/consul/api v1.10.1-0.20240820053928-ed738a6f981b h1:t9wnJmPWqzUCQwm9ZZY6G9vezJ3GYMezW+5QLuuf8f4=
+github.com/hashicorp/consul/api v1.10.1-0.20240820053928-ed738a6f981b/go.mod h1:p0FnUpQUXtnXNdi00G9Axjx4GffX0DMAuvS28NHY2Y0=
```
zalimeni, line 188: This should be 1.29.2 now right?

acceptance/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/gruntwork-io/terratest v0.46.7
 	github.com/hashicorp/consul-k8s/control-plane v0.0.0-20240226161840-f3842c41cb2b
-	github.com/hashicorp/consul/api v1.29.1
+	github.com/hashicorp/consul/api v1.10.1-0.20240820053928-ed738a6f981b
```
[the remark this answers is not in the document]
ndhanushkodi: Yes eventually, but only after that tag is cut, for now, it's using a commit off of main (ed738). When releasing we're gonna need to grab the cut tag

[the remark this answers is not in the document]
zalimeni: Oh, sorry - the version in `cli` threw me [here](https://github.com/hashicorp/consul-k8s/pull/4256/files#diff-37a093b0dd13152e13290bd482db8bc68232329d6f3e9053f04c4e02eb910959R99), I thought it was the new tag. Ignore me! We're waiting on 1.29.**3**

acceptance/go.mod
```diff
@@ -30,10 +30,7 @@ require (
 )
 
 // replace these so we always use the latest version of the control-plane types
-replace (
-	github.com/hashicorp/consul-k8s/control-plane => ../control-plane
-	github.com/hashicorp/consul-k8s/version => ../version
-)
+replace github.com/hashicorp/consul-k8s/version => ../version
```
ndhanushkodi, line 33: If we don't do this, then acceptance cannot differ for k8s.io versions. I did bump consul-k8s control plane though

zalimeni: Can we not keep the replace for `control-plane` so we're always testing current for PRs that change it? Or was this made necessary for now by k8s.io / some other thing I'm maybe overlooking

zalimeni (commented):
One question but otherwise LGTM. Thanks for making sure this got back to `main` promptly 🙏🏻

acceptance/go.mod
ndhanushkodi: It's the k8s.io version unfortunately! It will keep trying to pull in k8s.io 0.29.8 if we follow exactly. So we will need a fairly fast follow to actually get acceptance on k8s.io 0.29.8 if we want to bring back this replace. 

I think since PSPs don't work after K8s 1.25 so it's mostly a matter of removing that config when bumping k8s.io but I just didn't want to add on to an already hairy yak shave which is why I chose to not do it right before the release.

zalimeni: Got it. I think that fast follow is important here so we can keep regression testing strong before 1.20 and Oct. patches. Thanks for being careful and explaining!

zalimeni (approved):

PR #4262, opened 2024-08-24
[release/1.1.x] config entry: add validate_clusters to mesh config entry

ndhanushkodi:
- add validate_clusters to mesh config entry
- k8s dependencies: controller-gen 0.14, k8s.io 0.28.13, controller-runtime 0.16
  - Our k8s compat matrix with Consul https://developer.hashicorp.com/consul/docs/k8s/compatibility says 1.23-1.28, and 1.23-1.25 still supports PSPs. So we need a version of k8s.io that also supports that. 0.29.x does not, but 0.28.x does
- go modules for 1.15.x line: 
[api/v1.21.3](https://github.com/hashicorp/consul/releases/tag/api%2Fv1.21.2)
[envoyextensions/v0.2.2](https://github.com/hashicorp/consul/releases/tag/envoyextensions%2Fv0.2.2)
[troubleshoot/v0.2.1](https://github.com/hashicorp/consul/releases/tag/troubleshoot%2Fv0.2.1)

### Checklist ###
- [x] Tests added


cli/go.mod
```diff
@@ -24,11 +24,11 @@ require (
...
-	k8s.io/api v0.29.0
+	k8s.io/api v0.29.8
 	k8s.io/apiextensions-apiserver v0.29.0
-	k8s.io/apimachinery v0.29.0
-	k8s.io/cli-runtime v0.29.0
-	k8s.io/client-go v0.29.0
+	k8s.io/apimachinery v0.29.8
+	k8s.io/cli-runtime v0.29.8
+	k8s.io/client-go v0.29.8
 	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
```
ndhanushkodi, line 32: This was already on the 0.29 line so I just bumped the minor.

charts/consul/templates/crd-exportedservices.yaml
```diff
@@ -4,16 +4,15 @@ apiVersion: apiextensions.k8s.io/v1
...
+    controller-gen.kubebuilder.io/version: v0.14.0
   name: exportedservices.consul.hashicorp.com
+spec:
   labels:
     app: {{ template "consul.name" . }}
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
     component: crd
-spec:
```
zalimeni, line 16: It looks like this is back to below `labels` in https://github.com/hashicorp/consul-k8s/blob/release/1.5.x/charts/consul/templates/crd-exportedservices-v1.yaml

Is it possible this is wrong? I don't see it in the spec [schema](https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.29.0/_definitions.json#/definitions/io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec) for 1.28.14:
```yaml
    "io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec": {
      "description": "CustomResourceDefinitionSpec describes how a user wants their resource to appear",
      "properties": {
        "conversion": {
          "$ref": "#/definitions/io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceConversion",
          "description": "conversion defines conversion settings for the CRD."
        },
        "group": {
          "description": "group is the API group of the defined custom resource. The custom resources are served under `/apis/<group>/...`. Must match the name of the CustomResourceDefinition (in the form `<names.plural>.<group>`).",
          "type": "string"
        },
        "names": {
          "$ref": "#/definitions/io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionNames",
          "description": "names specify the resource and kind names for the custom resource."
        },
        "preserveUnknownFields": {
          "description": "preserveUnknownFields indicates that object fields which are not specified in the OpenAPI schema should be preserved when persisting to storage. apiVersion, kind, metadata and known fields inside metadata are always preserved. This field is deprecated in favor of setting `x-preserve-unknown-fields` to true in `spec.versions[*].schema.openAPIV3Schema`. See https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning for details.",
          "type": "boolean"
        },
        "scope": {
          "description": "scope indicates whether the defined custom resource is cluster- or namespace-scoped. Allowed values are `Cluster` and `Namespaced`.",
          "type": "string"
        },
        "versions": {
          "description": "versions is the list of all API versions of the defined custom resource. Version names are used to compute the order in which served versions are listed in API discovery. If the version string is \"kube-like\", it will sort above non \"kube-like\" version strings, which are ordered lexicographically. \"Kube-like\" versions start with a \"v\", then are followed by a number (the major version), then optionally the string \"alpha\" or \"beta\" and another number (the minor version). These are sorted first by GA > beta > alpha (where GA is a version with no suffix such as beta or alpha), and then by comparing major version, then minor version. An example sorted list of versions: v10, v2, v1, v11beta2, v10beta3, v3beta1, v12alpha1, v11alpha2, foo1, foo10.",
          "items": {
            "$ref": "#/definitions/io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionVersion"
          },
          "type": "array"
        }
      },
      "required": [
        "group",
        "names",
        "scope",
        "versions"
      ],
      "type": "object"
    },
```

zalimeni: Same question for the other templates below

cli/go.mod
```diff
@@ -9,12 +9,12 @@ replace github.com/hashicorp/consul-k8s/version => ../version
...
 	github.com/bgentry/speakeasy v0.1.0
 	github.com/cenkalti/backoff v2.2.1+incompatible
-	github.com/fatih/color v1.16.0
+	github.com/fatih/color v1.17.0
 	github.com/google/go-cmp v0.6.0
 	github.com/hashicorp/consul-k8s/charts v0.0.0-00010101000000-000000000000
 	github.com/hashicorp/consul-k8s/version v0.0.0
-	github.com/hashicorp/consul/troubleshoot v0.3.1
-	github.com/hashicorp/go-hclog v1.5.0
+	github.com/hashicorp/consul/troubleshoot v0.2.1
```
zalimeni, line 16: Looks like this is an intentional downgrade, since it corresponds to Consul 1.15.x 👍🏻

cli/go.mod
zalimeni: Arguably correct since K8s 1.28 EOLs soon and this is for Consul 1.15 LTS

control-plane/connect-inject/controllers/peering/peering_acceptor_controller.go
```diff
@@ -266,7 +266,7 @@ func (r *AcceptorController) updateStatus(ctx context.Context, acceptorObjKey ty
 			return err
 		}
 		if acceptor.Status.LatestPeeringVersion == nil || *acceptor.Status.LatestPeeringVersion < peeringVersion {
-			acceptor.Status.LatestPeeringVersion = pointer.Uint64(peeringVersion)
+			acceptor.Status.LatestPeeringVersion = ptr.To(uint64(peeringVersion))
```
zalimeni, line 269: Similar question here - seems like `main` and https://github.com/hashicorp/consul-k8s/pull/4256 don't yet use `ptr`, do we need to sync those up?

zalimeni (commented):
Couple of questions about schema and deps updates, but otherwise looking good!

control-plane/cni/go.mod
```diff
@@ -9,24 +9,23 @@ require (
 	github.com/hashicorp/consul/sdk v0.13.1
 	github.com/hashicorp/go-hclog v1.2.2
 	github.com/stretchr/testify v1.8.4
-	k8s.io/api v0.28.9
-	k8s.io/apimachinery v0.28.9
-	k8s.io/client-go v0.28.9
+	k8s.io/api v0.29.8
+	k8s.io/apimachinery v0.29.8
+	k8s.io/client-go v0.29.8
```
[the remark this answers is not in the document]
ndhanushkodi: yep! I started from "backwards" so I decided not to merge 4256 and just do all of the module bumps needed surrounding the necessary ones (generate the crds and updating consul api)

control-plane/connect-inject/controllers/peering/peering_acceptor_controller.go
ndhanushkodi: yup, just stuck on 1.3.x still 😅

cli/go.mod
ndhanushkodi: yup!

control-plane/cni/go.mod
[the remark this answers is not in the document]
zalimeni: Ahh, cool cool. SGTM 👍🏻

control-plane/connect-inject/controllers/peering/peering_acceptor_controller.go
zalimeni: Thank you for this lovely yak shave 🙏🏻 🥲

charts/consul/templates/crd-exportedservices.yaml
ndhanushkodi: I fixed it! I tried syncing the makefile, and the copy-crds-to-chart hack script from 1.3.x to here, and then realized that the hack script does a very hacky thing and uses line numbers to insert labels into the CRD. I had to tweak the line number a bit (I think the license header or something like that was causing the issue)

zalimeni: Aha!! Nice find, glad we have an answer!

zalimeni: And we can probably come up with a nearly-equivalent bashism that avoids line number hacking. Will take a look tomorrow.

zalimeni: Tried doing that here: https://github.com/hashicorp/consul-k8s/pull/4268/files happy to punt to after these PRs close if you'd rather

zalimeni (approved):
Couple of non-blocking comments, but LGTM! Thank you for working through this and all the hairy history 🙏🏻

charts/consul/templates/crd-exportedservices.yaml
ndhanushkodi: Thank you!! That's great!! I skimmed and it looks great but will take a deeper look right after the k8s release!

Makefile
```diff
@@ -101,7 +114,8 @@ cni-plugin-lint:
 
 .PHONY: ctrl-generate
 ctrl-generate: get-controller-gen ## Run CRD code generation.
-	cd control-plane; $(CONTROLLER_GEN) object:headerFile="build-support/controller/boilerplate.go.txt" paths="./..."
```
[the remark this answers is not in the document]
ndhanushkodi: Makes sense, removing!

PR #4268, opened 2024-08-28
improve CRD template field insertion line targeting

zalimeni:
Match YAML elements instead of blindly matching line numbers.


ndhanushkodi (approved):
:shipit: thanks for making this change!!

PR #4273, opened 2024-08-29
prepare release 1.4.5

ndhanushkodi:
-
-

### Checklist ###
- [ ] Tests added

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -66,7 +66,7 @@ global:
   # image: "hashicorp/consul-enterprise:1.10.0-ent"
   # ```
   # @default: hashicorp/consul:<latest version>
-  image: docker.mirror.hashicorp.services/hashicorppreview/consul:1.18-dev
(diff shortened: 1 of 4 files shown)


dhiaayachi (approved):

charts/consul/values.yaml
```diff
@@ -66,7 +66,7 @@ global:
   # image: "hashicorp/consul-enterprise:1.10.0-ent"
   # ```
   # @default: hashicorp/consul:<latest version>
-  image: docker.mirror.hashicorp.services/hashicorppreview/consul:1.18-dev
```
zalimeni, line 69: ~ Feels like we could start freezing these after CE is out of support since the tag should be equivalent

zalimeni (approved):

charts/consul/values.yaml
ndhanushkodi: makes sense, like not use -dev and just freeze at the CE tag?

PR #4300, opened 2024-09-03
NET-10763 - Allow DNS Proxy to configured with an ACL token when manageSystemACLs is false

jmurret:
- add the ability for users to set the ACL token on dns-proxy similiar to how catalog sync exposes the ability to set the ACL token in a secret
-

CI & unit tests

👀 

### Checklist ###
- [x] Tests added

### DIFF
--- a/charts/consul/templates/dns-proxy-deployment.yaml
+++ b/charts/consul/templates/dns-proxy-deployment.yaml
@@ -1,5 +1,4 @@
 {{- if (or (and (ne (.Values.dns.proxy.enabled | toString) "-") .Values.dns.proxy.enabled) (and (eq (.Values.dns.proxy.enabled | toString) "-") .Values.global.enabled)) }}
-{{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
 {{ template "consul.validateRequiredCloudSecretsExist" . }}
 {{ template "consul.validateCloudSecretKeys" . }}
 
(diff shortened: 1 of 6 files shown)


charts/consul/templates/dns-proxy-deployment.yaml
```diff
@@ -1,5 +1,4 @@
 {{- if (or (and (ne (.Values.dns.proxy.enabled | toString) "-") .Values.dns.proxy.enabled) (and (eq (.Values.dns.proxy.enabled | toString) "-") .Values.global.enabled)) }}
-{{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
```
jmurret, line 2: this should not have been a requirement.

dhiaayachi: Shouldn't we have the opposite requirement? Is dns-proxy safe to use with connect-inject enabled?

charts/consul/values.yaml
```diff
@@ -1925,6 +1925,17 @@ dns:
     # Port number to be used by DNS proxy
     port: 53
 
+    # Refers to a Kubernetes secret that you have created that contains
+    # an ACL token for your Consul cluster which allows the dns proxy the correct
+    # permissions. This is only needed if ACLs are managed manually within the Consul cluster, i.e. `global.acls.manageSystemACLs` is `false`.
```
dhiaayachi, line 1930: Maybe we can add what are the correct permissions that dns-proxy needs in that case?

dhiaayachi (changes requested):
@jmurret Thank you for adding this. I had a question about the deployment requirement and a suggestion to document the type of permission the token needed.

charts/consul/templates/dns-proxy-deployment.yaml
jmurret: I can't think of any reason why it would be unsafe.  I think ultimately services would use the local dataplane or agent, but I can't think of a reason why it would be unsafe and require us to prevent it being configured.

dhiaayachi: The only scenario I can think of where I'm not sure is with tproxy on. I'm not sure how we handle DNS in that case and if it's even allowed to resolve DNS at all because all traffic will be handled through envoy and virtual ips but yes probably not a big deal.

PR #4345, opened 2024-09-19
[NET-11106] Make bats testing more reproducible

jm96441n:
- update the `bats-tests` make target so that you can pass in a specific test to run and to use the docker image we use in CI for reproducibility with CI runs
- Update version of bats, terraform, and go in the bats testing dockerfile

### Checklist ###
- [ ] Tests added

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,7 +24,7 @@ jobs:
       with:
         workflow: test.yml
         repo: hashicorp/consul-k8s-workflows
-        ref: main
+        ref: update-bats-tests
         token: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
         inputs: '{ "context":"${{ env.CONTEXT }}", "actor":"${{ github.actor }}", "repository":"${{ github.repository }}", "branch":"${{ env.BRANCH }}", "sha":"${{ env.SHA }}", "token":"${{ secrets.ELEVATED_GITHUB_TOKEN }}" }'
 
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,7 +24,7 @@ jobs:
       with:
         workflow: test.yml
         repo: hashicorp/consul-k8s-workflows
-        ref: update-bats-tests
+        ref: main
         token: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       with:
         workflow: test.yml
         repo: hashicorp/consul-k8s-workflows
-        ref: main
+        ref: update-bats-tests
```
jm96441n, line 27: until the PR against the workflows repo merges this PR will be in draft but is still fine for review

Makefile
```diff
@@ -30,9 +30,10 @@ generate-external-crds: ## Generate CRDs for externally defined CRDs and copy th
 	@cd ./control-plane/config/crd/external; \
 		kustomize build | yq --split-exp '.metadata.name + ".yaml"' --no-doc
 
+
```
jmurret, line 33:
```suggestion
```

Makefile
```diff
@@ -30,9 +30,10 @@ generate-external-crds: ## Generate CRDs for externally defined CRDs and copy th
...
 .PHONY: bats-tests
 bats-tests: ## Run Helm chart bats tests.
-	 bats --jobs 4 charts/consul/test/unit
+	docker run -it -v $(CURDIR):/consul-k8s hashicorpdev/consul-helm-test:latest bats --jobs 4 /consul-k8s/charts/consul/test/unit -f "$(TEST_NAME)"
```
jmurret, line 36:
```suggestion
	docker run -it -v $(CURDIR):/consul-k8s hashicorpdev/consul-helm-test:latest bats --jobs 4 /consul-k8s/charts/consul/test/unit
```

jmurret (changes requested):
one blocking change on the bats filter

Makefile
jmurret: I did not think about this deeply, I thought you had may be added this for debugging, but I do see you can pass `-f ""` and it will run the whole suite.  nice.

jm96441n: no worries! I figured while I was in here I'd make a small QoL improvement to make this easier to use

.github/workflows/pr.yml
jm96441n: updated because the workflow PR has merged

jmurret (approved):

PR #4377, opened 2024-10-01
[NET-11222] rollback command

jm96441n:
- rollback the `sync-catalog purgeK8sServicesFromNode` command so we can implement the automatic removal on helm upgrade/install
-

### Checklist ###
- [ ] Tests added


control-plane/subcommand/sync-catalog/command.go
```diff
@@ -156,8 +156,8 @@ func (c *Command) init() {
 			"\"debug\", \"info\", \"warn\", and \"error\".")
 	c.flags.BoolVar(&c.flagLogJSON, "log-json", false,
 		"Enable or disable JSON output format for logging.")
-	c.flags.StringVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", "",
-		"Specifies the name of the Consul node for which to deregister synced Kubernetes services.")
+	// c.flags.StringVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", "",
+	// "Specifies the name of the Consul node for which to deregister synced Kubernetes services.")
```
missylbytes, line 160: We should remove instead of just comment out, right?

jm96441n: yeah, we're going to be putting it back with some different behavior for  the next patch but it's probably better to just remove it now

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -156,8 +156,8 @@ func (c *Command) init() {
 			"\"debug\", \"info\", \"warn\", and \"error\".")
 	c.flags.BoolVar(&c.flagLogJSON, "log-json", false,
 		"Enable or disable JSON output format for logging.")
-	c.flags.StringVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", "",
-		"Specifies the name of the Consul node for which to deregister synced Kubernetes services.")
+	// c.flags.StringVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", "",
```
wangxinyi7: Do we want to completely remove the code or only comment out the code?
We have unused flag in the above and unused function in the bottom

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -156,8 +156,6 @@ func (c *Command) init() {
 			"\"debug\", \"info\", \"warn\", and \"error\".")
 	c.flags.BoolVar(&c.flagLogJSON, "log-json", false,
 		"Enable or disable JSON output format for logging.")
-	c.flags.StringVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", "",
-		"Specifies the name of the Consul node for which to deregister synced Kubernetes services.")
 	c.flags.StringVar(&c.flagFilter, "filter", "",
```
wangxinyi7, line 159: We also need to remove this flag if we want to rollback

sarahalsmiller (approved):

sophie-gairo (approved):

PR #4378, opened 2024-10-01
[NET-11297] Purge on disable

jm96441n:
- add new field to the helm chart to allow users to purge services from consul when catalog sync is disabled, users will have to opt into this to preserve existing behavior

### Checklist ###
- [X] Tests added


charts/consul/values.yaml
```diff
@@ -2090,6 +2090,10 @@ syncCatalog:
   # global.enabled.
   enabled: false
 
+  # True if you want to deregister all services in this cluster from consul that have been registered by the catalog sync when the `enabled` flag is set to false.
+  # @type: boolean
+  purgeServicesOnDisable: false
```
jm96441n, line 2095: this needs a better name

wangxinyi7: Also, as we discussed before, do we need to add more fields here to allow users to control the purging in more granular ways?

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -462,64 +458,15 @@ func (c *Command) Run(args []string) int {
 
 // remove all k8s services from Consul.
 func (c *Command) removeAllK8SServicesFromConsulNode(consulClient *api.Client) error {
-	node, _, err := consulClient.Catalog().NodeServiceList(c.flagPurgeK8SServicesFromNode, &api.QueryOptions{Filter: c.flagFilter})
+	_, err := consulClient.Catalog().Deregister(&api.CatalogDeregistration{
```
wangxinyi7, line 461: The concern for doing the de-registration in one shot is that JPMC reported that they have thousands of services and they hope the de-registration would not impact the consul performance, that's the context of implementing batch purging in the implementation.

jm96441n: so deregistering the node shouldn't have that effect, we can run some perf tests to check on this though

jm96441n:
> Also, as we discussed before, do we need to add more fields here to allow users to control the purging in more granular ways?
we don't, it's not really a requirement and since we're doing this on a disable of the catalog-sync we don't want to potentially leave behind services that might no longer exist

charts/consul/values.yaml
jm96441n: we don't, it's not really a requirement and since we're doing this on a disable of the catalog-sync we don't want to potentially leave behind services that might no longer exist. The main requirement is "remove services registered by catalog sync when catalog sync is disabled"

charts/consul/templates/sync-catalog-deployment.yaml
```diff
@@ -1,4 +1,5 @@
-{{- if (or (and (ne (.Values.syncCatalog.enabled | toString) "-") .Values.syncCatalog.enabled) (and (eq (.Values.syncCatalog.enabled | toString) "-") .Values.global.enabled)) }}
+{{ - $syncCatalogEnabled := (or (and (ne (.Values.syncCatalog.enabled | toString) "-") .Values.syncCatalog.enabled) (and (eq (.Values.syncCatalog.enabled | toString) "-") .Values.global.enabled)) }}
```
jm96441n, line 1: thinking about this more we should have an entirely separate job to handle this rather than re-use the sync-catalog job

wangxinyi7: Yes, right, totally agree, I think creating a brand new one makes more sense here

control-plane/subcommand/sync-catalog/command.go
```diff
@@ -156,8 +152,8 @@ func (c *Command) init() {
 			"\"debug\", \"info\", \"warn\", and \"error\".")
 	c.flags.BoolVar(&c.flagLogJSON, "log-json", false,
 		"Enable or disable JSON output format for logging.")
-	c.flags.StringVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", "",
-		"Specifies the name of the Consul node for which to deregister synced Kubernetes services.")
+	c.flags.BoolVar(&c.flagPurgeK8SServicesFromNode, "purge-k8s-services-from-node", false,
```
jmurret, line 155: 👏

charts/consul/values.yaml
jm96441n: this still needs a better name

jmurret: Just brain dumping other possibilities:
- `cleanUpServicesOnDisable`
- `cleanUpOnDisable` (might be my vote thus far 🤷‍♂️ )
- `deregisterServicesOnDisable`
- `deregisterOnDisable`

It feels like the name needs to be some version of `..OnDisable` because putting `disable` as a prefix makes it seem like you are disabling something additional.

jm96441n: yeah I don't like the prefix `disable` I was thinking maybe `cleanupNodeOnRemoval` since the `Disable` can be a bit of misnomer because this runs on uninstall of consul and when you disable `syncCatalog` whereas the `removal` can mean "removal of sync catalog" or "removal of consul"

charts/consul/templates/sync-catalog-cleanup-on-uninstall-job.yaml
```diff
@@ -0,0 +1,123 @@
...
+            -enable-metrics \
+            {{- end }}
+            {{- if .Values.syncCatalog.metrics.path }}
+            -metrics-path={{ .Values.syncCatalog.metrics.path }} \
+            {{- end }}
+            {{- if .Values.syncCatalog.metrics.port }}
+            -metrics-port={{ .Values.syncCatalog.metrics.port }} \
+            {{- end }}
+            -prometheus-retention-time={{ .Values.global.metrics.agentMetricsRetentionTime }} \
+            {{- if .Release.IsUpgrade }}
```
jmurret, line 101: I think this evaluates to false when it is a delete.  The [docs](https://helm.sh/docs/topics/charts/) say it is only true on helm upgrade and rollback

jmurret (commented):
Nice!  left some suggestions for helm value naming.  The only possible blocking questions are:
- related to using the `IsUpgrade` in the delete job
- we need BATS tests for these files.

charts/consul/templates/sync-catalog-cleanup-on-uninstall-job.yaml
jm96441n: good catch, yeah this shouldn't be here

jmurret (approved):
👏 :shipit: 🙏 Thank you!

sarahalsmiller (approved):

PR #4382, opened 2024-10-09
NO-JIRA: cleanup iamRoles as part of nightly test cleanup

NiniOak:
## PROBLEM
Acceptance test in CI are currently failing with error `LimitExceeded: Cannot exceed quota for RolesPerAccount: 1000`.

## SOLUTION
Update the `aws-acceptance-test-cleanup` script to include cleaning up iamroles created as part of test environment setup. 

Tested locally to ensure all exisiting iAMRoles created as part of previous acceptance test runs were cleaned up using this script

### Checklist ###
- [ ] Tests added

### DIFF
--- a/hack/aws-acceptance-test-cleanup/main.go
+++ b/hack/aws-acceptance-test-cleanup/main.go
@@ -769,13 +769,7 @@ func cleanupIAMRoles(ctx context.Context, iamClient *iam.IAM) error {
 	rolePrefix := "consul-k8s-"
 	err := iamClient.ListRolesPagesWithContext(ctx, &iam.ListRolesInput{},
 		func(page *iam.ListRolesOutput, lastPage bool) bool {
-			for _, role := range page.Roles {
-				roleName := aws.StringValue(role.RoleName)
-				if strings.HasPrefix(roleName, rolePrefix) {
-					rolesToDelete = append(rolesToDelete, role)
-				}
-			}
-			return !lastPage
+			return filterRolesWithPrefix(page, lastPage, &rolesToDelete, rolePrefix)
 		})
 	if err != nil {
 		return fmt.Errorf("failed to list roles: %v", err)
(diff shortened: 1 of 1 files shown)


hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -759,6 +764,68 @@ func destroyBackoff(ctx context.Context, resourceKind string, resourceID string,
...
+	err := iamClient.ListRolesPagesWithContext(ctx, &iam.ListRolesInput{},
+		func(page *iam.ListRolesOutput, lastPage bool) bool {
+			for _, role := range page.Roles {
+				roleName := aws.StringValue(role.RoleName)
+				if strings.HasPrefix(roleName, rolePrefix) {
+					rolesToDelete = append(rolesToDelete, role)
+				}
+			}
+			return !lastPage
+		})
```
jm96441n, line 779: could we extract this to a named function and pass that in to make this a little more readable?

hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -759,6 +764,68 @@ func destroyBackoff(ctx context.Context, resourceKind string, resourceID string,
...
+	err := iamClient.ListAttachedRolePoliciesPages(&iam.ListAttachedRolePoliciesInput{
+		RoleName: roleName,
+	}, func(page *iam.ListAttachedRolePoliciesOutput, lastPage bool) bool {
+		for _, policy := range page.AttachedPolicies {
+			_, err := iamClient.DetachRolePolicy(&iam.DetachRolePolicyInput{
+				RoleName:  roleName,
+				PolicyArn: policy.PolicyArn,
+			})
+			if err != nil {
+				fmt.Printf("Failed to detach policy %s from role %s: %v", *policy.PolicyArn, *roleName, err)
```
jm96441n, line 821: do we need to worry about the nil pointer deref here?

wangxinyi7: As we are not de-ref the `err` and just check the nullity of the var, I think it's OK. Please correct me if I were wrong.

And the PR looks all good from my side! Thank you for your great work Anita!

hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -759,6 +764,68 @@ func destroyBackoff(ctx context.Context, resourceKind string, resourceID string,
...
+	rolePrefix := "consul-k8s-"
+	err := iamClient.ListRolesPagesWithContext(ctx, &iam.ListRolesInput{},
+		func(page *iam.ListRolesOutput, lastPage bool) bool {
+			for _, role := range page.Roles {
+				roleName := aws.StringValue(role.RoleName)
+				if strings.HasPrefix(roleName, rolePrefix) {
+					rolesToDelete = append(rolesToDelete, role)
+				}
+			}
+			return !lastPage
```
wangxinyi7, line 778: Would be much helpful if you could add a one liner comment to the meaning of the `lastPage`

hack/aws-acceptance-test-cleanup/main.go
jm96441n: the issue is on `*roleName` and `*policy.PolicyArn` which could potentially be a nil pointer

wangxinyi7: They are coming from `page.AttachedPolicies` and `page.Roles` which are offered by the `aws-sdk-go` package, I checked the source, both are list of pointers. So maybe we could *trust* aws? 😂

jm96441n: it's probably better to not  trust that we're getting good intput and code defensively here, aws provides functions for dealing with pointers and we should probably use them here to avoid the potential nil derefs https://docs.aws.amazon.com/sdk-for-go/api/aws/#StringValue

hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -759,6 +764,111 @@ func destroyBackoff(ctx context.Context, resourceKind string, resourceID string,
...
+			fmt.Printf("Warning: PolicyArn is nil for a policy attached to: %s", *roleName)
+			continue
+		}
+
+		_, err := iamClient.DetachRolePolicy(&iam.DetachRolePolicyInput{
+			RoleName:  roleName,
+			PolicyArn: policy.PolicyArn,
+		})
+		if err != nil {
+			fmt.Printf("Failed to detach policy %s from role %s: %v\n", aws.StringValue(policy.PolicyArn), *roleName, err)
```
wangxinyi7, line 865: Sorry, since we are also de-ref `roleName`, do we also need the `aws.StringValue` to wrap it? The same applies to line 835.

hack/aws-acceptance-test-cleanup/main.go
NiniOak: Yeah, I saw that and used that function to deref those values. Thanks for pointing it out

hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -759,6 +764,111 @@ func destroyBackoff(ctx context.Context, resourceKind string, resourceID string,
...
+func cleanupIAMRoles(ctx context.Context, iamClient *iam.IAM) error {
+	var rolesToDelete []*iam.Role
+	rolePrefix := "consul-k8s-"
+	err := iamClient.ListRolesPagesWithContext(ctx, &iam.ListRolesInput{},
+		func(page *iam.ListRolesOutput, lastPage bool) bool {
+			return filterRolesWithPrefix(page, lastPage, &rolesToDelete, rolePrefix)
+		})
+	if err != nil {
+		return fmt.Errorf("failed to list roles: %v", err)
+	}
```
nathancoleman, line 776: #nit I think this and the `filterRolesWithPrefix` func get a little easier to understand with less dereferencing and whatnot if we just let the func that we're calling out to collect the roles itself.

```go
rolesToDelete, err := filterIAMRolesWithPrefix(ctx, iamClient, "consul-k8s-")
if err != nil {
    return fmt.Errorf("failed to list roles: %w", err)
}
```

```go
func filterIAMRolesWithPrefix(ctx context.Context, iamClient *iam.IAM, prefix string) ([]*iam.Role, error) {
	var roles []*iam.Role

	err := iamClient.ListRolesPagesWithContext(ctx, &iam.ListRolesInput{},
		func(page *iam.ListRolesOutput, lastPage bool) bool {
			for _, role := range page.Roles {
				name := aws.StringValue(role.RoleName)
				if strings.HasPrefix(name, prefix) {
					roles = append(roles, role)
				}
			}

			// can just always return true here since we always want to see all of the roles
			return true
		})
	if err != nil {
		return nil, err
	}

	return roles, nil
}
```

hack/aws-acceptance-test-cleanup/main.go
```diff
@@ -759,6 +764,111 @@ func destroyBackoff(ctx context.Context, resourceKind string, resourceID string,
...
+
+	if len(rolesToDelete) == 0 {
+		fmt.Println("Found no iamRoles to clean up")
+		return nil
+	}
```
nathancoleman, line 781: #nit might be useful to log how many roles we're attempting to clean up in the `else` case

nathancoleman (approved):
LGTM, just some nitpicks -- nothing to block merge

PR #4383, opened 2024-10-09
bump kubernetes version for AKS

NiniOak:
## PROBLEM
Nightly acceptance test on AKS is currently failing with error 
```Code="K8sVersionNotSupported" Message="Managed cluster consul-k8s-2814350665 is on version 1.27.16 which is not supported in this region``` due to [end of life support for Kubernetes version 1.27.](https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli)
## SOLUTION
Upgraded AKS kubernetes version to 1.28

### Checklist ###
- [ ] Tests added

### DIFF
--- a/charts/consul/test/terraform/aks/variables.tf
+++ b/charts/consul/test/terraform/aks/variables.tf
@@ -7,7 +7,7 @@ variable "location" {
 }
 
 variable "kubernetes_version" {
-  default     = "1.29"
+  default     = "1.28"
   description = "Kubernetes version supported on AKS"
 }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: charts/consul/test/terraform/aks/variables.tf [hunk 1]
```diff
@@ -7,7 +7,7 @@ variable "location" {
 }
 
 variable "kubernetes_version" {
-  default     = "1.27"
```
jm96441n, line 10: suuuper minor but would be cool in the future if there was an AKS api we could hit or something and grab the version of this so we don't have to do it manually but that's way out of scope for this

jm96441n (approved):

nathancoleman (approved):

PR #4384, opened 2024-10-10
Increase speed sync with k8s

v-kamerdinerov:
-
-

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/catalog/to-consul/syncer.go
+++ b/control-plane/catalog/to-consul/syncer.go
@@ -5,6 +5,7 @@ package catalog
 
 import (
 	"context"
+	"reflect"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


v-kamerdinerov:
Not done yet

blake:
Hi @v-kamerdinerov, the sync speed can be adjusted using the [`syncCatalog.consulWriteInterval`](https://developer.hashicorp.com/consul/docs/k8s/helm#v-synccatalog-consulwriteinterval) parameter in the Helm chart, or by passing the `-consul-write-interval` flag to `consul-k8s-control-plane sync-catalog`, if you happen to be manually starting the control plane process.

Is this suitable for your needs?

jukie:
> However, my fix also allows synchronization when - registration is new or changed
I may be mistaken but I though thtat syncer already reacts as soon as endpoints are updated by use of the Informer? `ConsulSyncPeriod` is useful to reconcile drift for example and ensure that periodic syncs happen but what are you looking to achieve with this?

v-kamerdinerov:
> > However, my fix also allows synchronization when - registration is new or changed
> 
> I may be mistaken but I though thtat syncer already reacts as soon as endpoints are updated by use of the Informer? `ConsulSyncPeriod` is useful to reconcile drift for example and ensure that periodic syncs happen but what are you looking to achieve with this?

In addition to registering all services periodically, now sync new services as they appear, allowing to increase the full sync interval and reduce the overall number of catalog register call to Consul servers.
This makes a difference for kube clusters with hundreds of instances without delaying updates.

PR #4432, opened 2024-11-26
[NET-11256] Add `gateway read` command to consul-k8s CLI

nathancoleman:
> [!NOTE]
> This PR is blocked by the bug fix in https://github.com/hashicorp/consul-k8s/pull/4434 which is preventing tests from passing

This PR adds a new command, `consul-k8s gateway read`, which collects the various components of configuration for a given Consul API Gateway – `Gateway`, attached `HTTPRoutes`, attached `TCPRoutes`, etc. – into a single yaml file.

It can output an archive named after the read gateway into the current directory, or it can output JSON to the terminal.

```shell
# Output a zip archive for my-gateway in the current namespace on your kubecontext
$ consul-k8s gateway read my-gateway --output archive

# Output JSON to the terminal for my-gateway in the <my-namespace> namespace
$ consul-k8s gateway read my-gateway --namespace my-namespace --output json
```

Create one or more Gateways in your Kube cluster. Example for this config is available [here](https://github.com/nathancoleman/consul-lab/tree/main/k8s/api-gateway/basic).

Run the commands above to inspect the various forms of output, making sure to use both the implicit namespace on your kubecontext and an explicit `--namespace <my-namespace>` arg.

See above

### Checklist ###
- [x] Tests added


cli/cmd/gateway/read/command.go
```diff
@@ -0,0 +1,299 @@
...
+
+// init establishes the flags for Command
+func (c *Command) init() {
+	c.set = flag.NewSets()
+
+	f := c.set.NewSet("Command Options")
+	f.StringVar(&flag.StringVar{
+		Name:    "namespace",
+		Target:  &c.flagGatewayNamespace,
+		Usage:   "The namespace of the Gateway to read",
```
jm96441n, line 61: should we notify that the default value for this is the default namespace

jm96441n: and should we clarify that it's the k8s namespace and not the consul namespace (in case they do some mapping of namespaces?)

cli/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/hashicorp/consul-k8s/cli
 
-go 1.22
+go 1.22.0
 
-toolchain go1.22.5
+toolchain go1.22.6
```
jm96441n, line 5: did we mean to bump the toolchain here? it's probably good that we do (and should probably be on the latest minor version tbh)

cli/cmd/gateway/read/command.go
```diff
@@ -0,0 +1,299 @@
...
+		c.UI.Output(err.Error(), terminal.WithErrorStyle())
+		return 1
+	}
+
+	return 0
+}
+
+func (c *Command) fetchCRDs() error {
+	// Fetch Gateway
+	var gateway gwv1beta1.Gateway
```
jm96441n, line 123: is it possible that someone is running the v1alpha1 of the gateway if they're on an old consul-k8s version? should we worry about that/care about that?

cli/go.mod
nathancoleman: I did not mean to -- I must be running a newer version of Go on my machine than what we use in the repo.

I'll bump it back down just to avoid sprawl in this PR

cli/cmd/gateway/read/command.go
nathancoleman: v1.2 which introduced `Gateway` to consul-k8s itself consumes `v1beta1.Gateway` [here](https://github.com/hashicorp/consul-k8s/blob/v1.2.9/control-plane/api-gateway/controllers/gateway_controller.go#L402), so customers would have to be using the old standalone consul-api-gateway project in order to be running the `v1alpha1.Gateway`.

Based on that, I don't think we need to support it here.

jm96441n (approved):

cli/cmd/gateway/read/command.go
nathancoleman: the default value is whatever namespace is set on the kubecontext, so it could be `default` or it could be something else

PR #4433, opened 2024-11-27
[NET-11256] Add `gateway list` command to consul-k8s CLI

nathancoleman:
> [!NOTE]
> This PR is blocked by the bug fix in https://github.com/hashicorp/consul-k8s/pull/4434 which is preventing tests from passing

This PR adds a new command, `consul-k8s gateway list`, which collects the various components of configuration for all Gateways in a given K8s cluster or namespace – `Gateway`, attached `HTTPRoutes`, attached `TCPRoutes`, etc. – into a single zip file containing one yaml file per Gateway.

It can output an archive named `gateways.zip` into the current directory, or it can output JSON to the terminal.

```shell
# Output a zip archive for all gateways in the current namespace on your kubecontext
$ consul-k8s gateway list --output archive

# Output JSON to the terminal for all gateways in the <my-namespace> namespace
$ consul-k8s gateway list --namespace my-namespace --output json

# Output JSON to the terminal for all gateways in the k8s cluster
$ consul-k8s gateway list --all-namespaces --output json
```

Create one or more Gateways in your Kube cluster. Example for this config is available [here](https://github.com/nathancoleman/consul-lab/tree/main/k8s/api-gateway/basic).

Run the commands above to inspect the various forms of output, making sure to use both the implicit namespace on your kubecontext and an explicit `--namespace <my-namespace>` arg.

See above

### Checklist ###
- [x] Tests added

### DIFF
--- a/cli/cmd/gateway/list/command.go
+++ b/cli/cmd/gateway/list/command.go
@@ -1,0 +1,312 @@
+package read
+
+import (
+	"archive/zip"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+
+	helmcli "helm.sh/helm/v3/pkg/cli"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+	"sigs.k8s.io/yaml"
+
+	"github.com/hashicorp/consul-k8s/cli/common"
+	"github.com/hashicorp/consul-k8s/cli/common/flag"
+	"github.com/hashicorp/consul-k8s/cli/common/terminal"
+)
+
+type Command struct {
+	*common.BaseCommand
+
+	kubernetes client.Client
+	restConfig *rest.Config
+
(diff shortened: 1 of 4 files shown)


cli/cmd/gateway/list/command.go
```diff
@@ -0,0 +1,312 @@
...
+	f.StringVar(&flag.StringVar{
+		Name:    "kubeconfig",
+		Aliases: []string{"c"},
+		Target:  &c.flagKubeConfig,
+		Usage:   "Set the path to a kubeconfig file.",
+	})
+	f.StringVar(&flag.StringVar{
+		Name:   "context",
+		Target: &c.flagKubeContext,
+		Usage:  "Set the Kubernetes context to use.",
```
NiniOak, line 88: Should there be Aliases for this flag?

NiniOak (approved):

cli/cmd/gateway/list/command.go
nathancoleman: I just made it consistent with the existing `consul-k8s proxy list` command which [doesn't specify an alias](https://github.com/hashicorp/consul-k8s/blob/acb1af85a012b3600480f0120a9d772cd4ad304f/cli/cmd/proxy/list/command.go#L89-L94). I'm not finding any other commands that specify an alias, so I think it makes sense not to have one here for consistency.

PR #4434, opened 2024-11-27
Fix duplicate key in connect-inject ACL policy

nathancoleman:
Stop including a duplicate `acl = "write"` rule at the root level of the connect-inject ACL policy when namespaces are not enabled. The security patch in https://github.com/hashicorp/consul/pull/21908 led to the connect-injector being unable to start up as its ACL policy is rejected by Consul.

1. Expanded the test matrix to cover all possible flag combinations w/ current policy output
2. Fixed policy generation
3. Updated the test assertions to account for the bug fix, verifying that `acl = "write"` never appears twice at the root level of the policy with any combination of flags

Review commit-by-commit observing the above 

### Checklist ###
- [x] Tests added

### DIFF
--- a/control-plane/subcommand/server-acl-init/rules_test.go
+++ b/control-plane/subcommand/server-acl-init/rules_test.go
@@ -883,7 +883,6 @@ func TestInjectRules(t *testing.T) {
   node_prefix "" {
     policy = "write"
   }
-    acl = "write"
     service_prefix "" {
       policy = "write"
       intentions = "write"
@@ -929,7 +928,6 @@ partition "part-1" {
     policy = "write"
   }
     policy = "write"
-    acl = "write"
     service_prefix "" {
       policy = "write"
       intentions = "write"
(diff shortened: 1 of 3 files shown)


control-plane/subcommand/server-acl-init/rules_test.go
```diff
@@ -917,7 +917,31 @@ func TestInjectRules(t *testing.T) {
...
+    policy = "write"
+    acl = "write"
+    service_prefix "" {
+      policy = "write"
+      intentions = "write"
+    }
+    identity_prefix "" {
+      policy = "write"
+      intentions = "write"
+    }
```
nathancoleman: You can see the bug here where `acl = "write"` winds up being included twice in the root of the policy whenever namespaces are disabled.

The later commit in this PR that adjusts the test assertions after the fix should indicate no scenario in which `acl` or any other type of rule is included twice at the same level.

sarahalsmiller (approved):

PR #4478, opened 2025-02-11
NET-12097 Support PodSecurityAdmission at the restricted level

nathancoleman:
> [!WARNING]
> - API gateways that bind to privileged ports must now use the pre-existing `mapPrivilegedContainerPorts` setting in `values.yaml` ([docs](https://developer.hashicorp.com/consul/docs/connect/gateways/api-gateway/tech-specs#privileged-container-ports)); otherwise, Envoy will fail to attach the listener when the gateway `Pod` is starting up.

-
-

### Checklist ###
- [x] Tests added


NiniOak (approved):
thanks for these changes

PR #4482, opened 2025-02-26
Fix ingress hostname extract logic for syncCatalog

oscar-dela:

In the current implementation, the control plane will register ingress service with it `ingress.Spec.Rules[*].Host` to consul. However it should be registered with `ingress.Status.LoadBalancer.Ingress[*].Hostname` which is when SyncLoadBalancerIPs is enabled, to make consul solve the service DNS to the cluster load balancer rather than a non-exists host entry.

1. Followed the contribution guide to run the tests.
2. Build the control-plane container, and deploy it with helm chart into my testing Kubernetes cluster to verify it works.

### Checklist ###
- [x] Tests added


dmeremyanin:
Hey @nitin-sachdev-29, sorry to ping you. I noticed you've been reviewing other PRs, and it would be greatly appreciated if you could take a look at this one when you have a chance. The change in this PR is really important for us as well. Thanks

control-plane/catalog/to-consul/resource.go
```diff
@@ -997,7 +997,7 @@ func (t *serviceIngressResource) Upsert(key string, raw interface{}) error {
 			}
 			hostName = ingress.Status.LoadBalancer.Ingress[0].IP
 		} else {
-			hostName = rule.Host
+			hostName = ingress.Status.LoadBalancer.Ingress[0].Hostname
```
dmeremyanin, line 1000: Wouldn't it make sense to add a separate if-branch with a config option, `SyncLoadBalancerHostnames` (similar to `SyncLoadBalancerIPs`), to maintain the current behavior? That way, people who need it can enable it, while ensuring we don't break anything.

PR #4519, opened 2025-04-13
fix: skip Ingress objects without HTTP rule to prevent segfault

dmeremyanin:

Avoids syncing Ingress resources that don't define an HTTP rule, which previously caused a segmentation fault. This aligns with the Kubernetes Ingress spec, where the HTTP rule is optional.

Reference:
https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressSpec

1. Unit tests have been updated to include a case where the Ingress resource lacks an HTTP rule.
2. Manual testing was performed by deploying the updated logic in a test cluster to ensure the fix works as expected in a live environment.

1. Review the updated test cases to confirm that the logic correctly skips Ingress resources without HTTP rules.
2. Ensure compatibility with the Kubernetes Ingress spec.

### Checklist ###
- [x] Tests added


dmeremyanin:
Hey @pajay-rao, sorry for the ping - I saw you've been contributing to this repo. If you have a moment, could you maybe take a quick look at the PR? Or if you know who would be the right person to review it, I'd really appreciate the guidance. Thanks!

PR #4576, opened 2025-05-15
Fixing multiple flaky test failures due to timeout, version updates, script changes

pajay-rao:
This PR fixes multiple small issues on GitHub workflows. 
Tests are flaky due to multiple issues including test timeouts, terraform version updates and terraform script updates.

This PR reduces the flaky failure of the tests but still there are issues which surface randomly.

Executions:
https://github.com/hashicorp/consul-k8s-workflows/actions/runs/15069649253/attempts/1

-
-

### Checklist ###
- [ ] Tests added


sreeram77 (dismissed):

charts/consul/test/terraform/aks/variables.tf
```diff
@@ -7,7 +7,7 @@ variable "location" {
 }
 
 variable "kubernetes_version" {
-  default     = "1.28"
+  default     = "1.31"
```
anandmukul93, line 10: is 1.32 not available for AKS ? just check and lets set this to a LTS version

anandmukul93 (commented):
Can you look at the version bumps for eks and is the time for retries going to be enough ?

.github/workflows/pr-cloud-accepance.yml
anandmukul93, line 1: I think we already have dispatcher workflow somewhere on raising PR. This will only be redundant.

anandmukul93 (commented):
Check the pr cloud acceptance workflow already fires the acceptance tests on consul-k8s-workflows via some test workflow. This will be redundant and wont be required.

PR #4717, opened 2025-08-25
CSL-11197: code for rollout restart

vdinesh4738:
This code only fixes part of JIRA related to rollout restart. Feature request needs to be raised for scaling down to 0 replicas

### DIFF
--- a/control-plane/api-gateway/gatekeeper/deployment.go
+++ b/control-plane/api-gateway/gatekeeper/deployment.go
@@ -59,7 +59,7 @@ func (g *Gatekeeper) upsertDeployment(ctx context.Context, gateway gwv1beta1.Gat
 	}
 
 	mutated := deployment.DeepCopy()
-	mutator := newDeploymentMutator(deployment, mutated, gcc, gateway, g.Client.Scheme())
+	mutator := newDeploymentMutator(deployment, mutated, existingDeployment, exists, gcc, gateway, g.Client.Scheme())
 
 	result, err := controllerutil.CreateOrUpdate(ctx, g.Client, mutated, mutator)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


control-plane/api-gateway/gatekeeper/deployment.go
```diff
@@ -222,10 +222,24 @@ func compareDeployments(a, b *appsv1.Deployment) bool {
 	return *b.Spec.Replicas == *a.Spec.Replicas
 }
 
-func newDeploymentMutator(deployment, mutated *appsv1.Deployment, gcc v1alpha1.GatewayClassConfig, gateway gwv1beta1.Gateway, scheme *runtime.Scheme) resourceMutator {
+func newDeploymentMutator(deployment, mutated, existingDeployment *appsv1.Deployment, deploymentExists bool, gcc v1alpha1.GatewayClassConfig, gateway gwv1beta1.Gateway, scheme *runtime.Scheme) resourceMutator {
 	return func() error {
 		mutated = mergeDeployments(gcc, deployment, mutated)
-		return ctrl.SetControllerReference(&gateway, mutated, scheme)
+		if err := ctrl.SetControllerReference(&gateway, mutated, scheme); err != nil {
```
sanikachavan5: even this won't give you the changes for multiple replicas right? as that is being handled by the mergeDeployments function right? Also, why can't we move these changes to the mergeDeployments, as it is kind of violating SRP of the function.

vdinesh4738: This is not for scaling to 0, that is the description

.changelog/4717.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+control-plane: fixes rollout restart for apigateway
```
sanikachavan5: api-gateway: resolve issue preventing rollout restarts from proceeding due to the gateway controller not detecting annotation changes in deployments made during rollout restarts.

.github/workflows/pr.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         repo: hashicorp/consul-k8s-workflows
         ref: main
         token: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
-        inputs: '{ "context":"${{ env.CONTEXT }}", "actor":"${{ github.actor }}", "repository":"${{ github.repository }}", "branch":"${{ env.BRANCH }}", "sha":"${{ env.SHA }}", "token":"${{ secrets.ELEVATED_GITHUB_TOKEN }}" }'
+        inputs: '{ "context":"${{ env.CONTEXT }}", "actor":"${{ github.actor }}", "repository":"${{ github.repository }}", "branch":"${{ env.BRANCH }}", "sha":"${{ env.SHA }}", "token":"${{ secrets.ELEVATED_GITHUB_TOKEN }}", "uniqueness_group": "pr-${{ github.event.pull_request.number }}" }'
```
sanikachavan5: Why did we need this?

control-plane/api-gateway/gatekeeper/deployment.go
```diff
@@ -162,11 +162,22 @@ func (g *Gatekeeper) deployment(gateway gwv1beta1.Gateway, gcc v1alpha1.GatewayC
...
 		b.Spec.Template = a.Spec.Template
 		b.Spec.Replicas = deploymentReplicas(gcc, a.Spec.Replicas)
 	}
+	if b.Spec.Template.Annotations == nil {
+		b.Spec.Template.Annotations = make(map[string]string)
+	}
+	if deploymentExists && existingDeployment.Spec.Template.Annotations != nil {
+		// Preserve existing annotations that are not set by the controller.
+		for k, v := range existingDeployment.Spec.Template.Annotations {
+			if _, ok := b.Spec.Template.Annotations[k]; !ok {
```
sanikachavan5, line 176: Use map.Copy

.github/workflows/pr.yml
vdinesh4738: there is a new field added in the consul-k8s-workflow, which is preventing it from running tests. There will be a pr from Mukul for the same to auto populate this value, but now this is required

sanikachavan5: Okay

control-plane/api-gateway/gatekeeper/deployment.go
vdinesh4738: I am copying the annotations which are additional, not to replace this. maps.Copy will replace

PR #4744, opened 2025-09-03
granularity for package level concurrency and rerun availability

anandmukul93:
- made each package re-runnable as a matrix job

### Checklist ###
- [ ] Tests added

## PCI review checklist

### DIFF
--- a/acceptance/ci-inputs/aks_acceptance_test_packages.yaml
+++ b/acceptance/ci-inputs/aks_acceptance_test_packages.yaml
@@ -2,6 +2,20 @@
 # SPDX-License-Identifier: MPL-2.0
 
 # Cloud package is not included in test suite as it is triggered from a non consul-k8s repo and requires HCP credentials
-- {runner: 0, test-packages: "connect peering snapshot-agent wan-federation"}
-- {runner: 1, test-packages: "consul-dns example partitions metrics sync"}
-- {runner: 2, test-packages: "basic cli config-entries api-gateway ingress-gateway terminating-gateway vault server"}
+- {runner: 0, test-packages: "connect"}
+- {runner: 1, test-packages: "peering"}
+- {runner: 2, test-packages: "snapshot-agent"}
+- {runner: 3, test-packages: "wan-federation"}
+- {runner: 4, test-packages: "consul-dns"}
+- {runner: 5, test-packages: "example"}
+- {runner: 6, test-packages: "partitions"}
+- {runner: 7, test-packages: "metrics"}
+- {runner: 8, test-packages: "sync"}
+- {runner: 9, test-packages: "basic"}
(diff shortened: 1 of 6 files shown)


.github/workflows/pr-cloud-accepance.yml
```diff
@@ -27,6 +22,6 @@ jobs:
         with:
           workflow: cloud.yml
           repo: hashicorp/consul-k8s-workflows
-          ref: main
+          ref: "mukul/granularity-reporting-fix"
```
Vikramarjuna: please fix this

anandmukul93: the workflows-k8s need to be merged and one more run of this workflow needs to be checked then this will pass. then we can merge this.

anandmukul93: i ll push then with main as now if i push it will be error anyways.

acceptance/ci-inputs/aks_acceptance_test_packages.yaml
```diff
@@ -2,6 +2,20 @@
...
 
 # Cloud package is not included in test suite as it is triggered from a non consul-k8s repo and requires HCP credentials
-- {runner: 0, test-packages: "connect peering snapshot-agent wan-federation"}
-- {runner: 1, test-packages: "consul-dns example partitions metrics sync"}
-- {runner: 2, test-packages: "basic cli config-entries api-gateway ingress-gateway terminating-gateway vault server"}
\ No newline at end of file
+- {runner: 0, test-packages: "connect"}
+- {runner: 1, test-packages: "peering"}
+- {runner: 2, test-packages: "snapshot-agent"}
+- {runner: 3, test-packages: "wan-federation"}
```
curtbushko, line 8: **suggestion:** #team-pss is not going to like this. You are basically 17x the number of github runners we use.

anandmukul93: We have limited the max parallel to 4 for the matrix job that uses these . earlier the matrix had cardinality of 3 and no parallelization limits so it was capped to 3. The tests are a time-taking bunch and take a lot of time with timeout set to 2 hours per package(including failure reruns). We needed this change as this was impacting the pace at which we did our release and got stuck at a lot of times on reruns with each rerun taking a huge amount of time.

curtbushko: Thanks for the context. I am aware of some of the flakiness and time consumption of the acceptance tests on multiple clouds. Before the handoff, we never gated releases based on the cloud acceptance tests. If the Kind acceptance tests passed then we were good to release.

The cloud tests take too long to run and often fail due to reasons outside of our consul-k8s code changes.

PR #4745, opened 2025-09-04
Use privileged dataplane entrypoint for consul-dataplane

Vikramarjuna:
> [!NOTE]
> This PR has a dependency on https://github.com/hashicorp/consul-dataplane/pull/652

The current implementation enables NET_BIND_SERVICE for all the uses cases of data plane (envoy). We do not need this capability to be added to the pod when there privileged ports are not used.

We have generated the new data plane image as part of <PR-TODO> which can take the command to invoke the privileged envoy binary which can support the NET_BIND_SERVICE capability.

The current set of changes uses that, privileged, entry point when privileged ports are used for the use case. Although the most common use case is Ingress gateways which historically supported binding to privileged ports, for backward compatibility, the support is extended to other use cases: API Gateway and Mesh Gateway where user can configure the port.

All other use cases of data plane now use the "unprivileged", default, entry point which does not require the NET_BIND_SERVICE capability at runtime.

- Verified functionality for service mesh with ingress, API, and mesh gateways using both privileged and unprivileged ports to verify that it uses the appropriate entry point.
- Verified the function or service mesh with terminating gateway and regular service to service communication where unprivileged entry point.

see above.

### Checklist ###
- [ ] Tests added

## PCI review checklist

### DIFF
--- a/charts/consul/templates/mesh-gateway-deployment.yaml
+++ b/charts/consul/templates/mesh-gateway-deployment.yaml
@@ -196,27 +196,23 @@ spec:
       - name: mesh-gateway
         image: {{ .Values.global.imageConsulDataplane | quote }}
         {{ template "consul.imagePullPolicy" . }}
-        {{ if not .Values.meshGateway.hostNetwork}}
         securityContext:
-          {{- if lt (int .Values.meshGateway.containerPort) 1024 }}
-          allowPrivilegeEscalation: true
-          runAsNonRoot: false
-          capabilities:
-            drop:
-              - ALL
-            add:
-              - NET_BIND_SERVICE
-          {{- else }}
           allowPrivilegeEscalation: false
-          runAsNonRoot: true
-          capabilities:
-            drop:
-              - ALL
-          {{- end }}
           readOnlyRootFilesystem: true
+          runAsNonRoot: true
           seccompProfile:
             type: RuntimeDefault
-        {{- end }}
+          {{- if or (not .Values.meshGateway.hostNetwork) (lt (int .Values.meshGateway.containerPort) 1024) }}
+          capabilities:
+            {{- if not .Values.meshGateway.hostNetwork }}
+            drop:
+              - ALL
+            {{- end }}
+            {{- if lt (int .Values.meshGateway.containerPort) 1024 }}
+            add:
+              - NET_BIND_SERVICE
+            {{- end }}
+          {{- end }}
         {{- if .Values.meshGateway.resources }}
         resources:
             {{- if eq (typeOf .Values.meshGateway.resources) "string" }}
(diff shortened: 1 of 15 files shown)


control-plane/api-gateway/gatekeeper/dataplane_test.go
```diff
@@ -0,0 +1,343 @@
...
+			gateway: gwv1beta1.Gateway{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-gateway",
+					Namespace: "default",
+				},
+				Spec: gwv1beta1.GatewaySpec{
+					Listeners: []gwv1beta1.Listener{
+						{
+							Name: "http",
+							Port: 8080, // Non-privileged port
```
yaminisridaran: @Vikramarjuna can we please use constant instead of hardcode value

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
```diff
@@ -265,15 +265,6 @@ func (w *MeshWebhook) consulDataplaneSidecar(namespace corev1.Namespace, pod cor
 		RunAsGroup:               ptr.To(group),
 		RunAsNonRoot:             ptr.To(true),
 		AllowPrivilegeEscalation: ptr.To(false),
-		SeccompProfile: &corev1.SeccompProfile{
```
anandmukul93, line 268: was the requirement also for seccomp Profile or only net bind capabability ? Is there any need to define this ?

control-plane/api-gateway/gatekeeper/dataplane_test.go
Vikramarjuna: Addressed

control-plane/connect-inject/webhook/consul_dataplane_sidecar.go
Vikramarjuna: No, it was removed as part of refactoring. Reverted the change. Added the SeccompProfile config back.

srahul3 (dismissed):
✅ LGTM

yaminisridaran (approved):
Looks Good to me

charts/consul/templates/mesh-gateway-deployment.yaml
```diff
@@ -196,19 +196,27 @@ spec:
...
+        {{ if not .Values.meshGateway.hostNetwork}}
         securityContext:
-          allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: true
-          runAsNonRoot: true
-          seccompProfile:
-            type: RuntimeDefault
+          {{- if lt (int .Values.meshGateway.containerPort) 1024 }}
+          allowPrivilegeEscalation: true
+          runAsNonRoot: false
```
anandmukul93: Can you check if this should be true and runAsUser is 100

Vikramarjuna: fixed the code. runAsNonRoot is set to true always

anandmukul93 (approved):
LGTM

pajay-rao (approved):

pajay-rao (approved):

PR #4771, opened 2025-09-12
DO NOT MERGE: under test

vdinesh4738:


.changelog/4771.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:test
```
anandmukul93, line 1: will this file be in the final pr ? 
we dont add release notes for ci/cd , lint, etc related changes.

.github/workflows/pr.yml
```diff
@@ -2,35 +2,90 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Check file changes
+        id: filter
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            only_meta:
+              - 'charts/consul/**'
+              - '.changelog/**'
```
anandmukul93, line 30: on charts changes tests do run as chart changes include impacting changes on control plane. Can we do a branch naming convention based skip and use that as test PR without firing these workflows. Applicable only if these changes will be pushed to main.

PR #4776, opened 2025-09-17
feat: add dual stack support for Consul configuration

nitin-sachdev-29:
- feat: add dual stack support for Consul configuration
-

### Checklist ###
- [ ] Tests added

## PCI review checklist

### DIFF
--- a/charts/consul/templates/client-daemonset.yaml
+++ b/charts/consul/templates/client-daemonset.yaml
@@ -291,7 +291,7 @@ spec:
               exec /usr/local/bin/docker-entrypoint.sh consul agent \
                 -node="${NODE}" \
                 -advertise="${ADVERTISE_IP}" \
-                {{- if .Values.dualStack.defaultEnabled }}
+                {{- if .Values.global.dualStack.defaultEnabled }}
                 -bind="[::]" \
                 -client="[::]" \
                 {{- else }}
@@ -490,7 +490,7 @@ spec:
                   {{- if .Values.global.tls.enabled }}
                   curl \
                     -k \
-                    {{- if .Values.dualStack.defaultEnabled }}
+                    {{- if .Values.global.dualStack.defaultEnabled }}
                     https://[::1]:8501/v1/status/leader \
                     {{- else }}
                     https://127.0.0.1:8501/v1/status/leader \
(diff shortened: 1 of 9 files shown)


charts/consul/templates/client-daemonset.yaml
```diff
@@ -291,8 +291,13 @@ spec:
               exec /usr/local/bin/docker-entrypoint.sh consul agent \
                 -node="${NODE}" \
                 -advertise="${ADVERTISE_IP}" \
+                {{- if .Values.dualStack.defaultEnabled }}
+                -bind="::" \
+                -client="::" \
```
anandmukul93: we should wrap this in []

charts/consul/templates/dns-proxy-deployment.yaml
```diff
@@ -126,7 +126,11 @@ spec:
           command:
             - consul-dataplane
           args:
+            {{- if .Values.dualStack.defaultEnabled }}
+            - -consul-dns-bind-addr="::"
+            {{- else }}
```
anandmukul93, line 131: same here []

pajay-rao: for just ipv6 address its not required. if it has port its needed.

anandmukul93: found out that standard with ports should be as below.
If some library handles the full ipv6 addr + port generation i am fine with that otherwise we should wrap it right ?
But lets not get blocked on this for now. 

https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers

anandmukul93 (dismissed):
LGTM as seems bind addr is handled inside consul along with ports. @pajay-rao Am i right from what we might have understood?

pajay-rao (approved):

charts/consul/values.yaml
```diff
@@ -3784,3 +3784,6 @@ telemetryCollector:
   # it could be used to configure custom consul parameters.
   # @type: map
   extraEnvironmentVars: {}
+
+dualStack:
```
pajay-rao: need to move dualStack section inside
global:
   network:
      dualStack:
          enabled: false

pajay-rao (approved):

PR #4783, opened 2025-09-22
CSL-10969: Add `-output` flag to proxy stats command. 

LordAbhishek:
- Added a new flag `-output` to proxy stats command. With this flag, command will write the proxy stats of a pod to a json file in current working dir.
- For now, supported output format is only `archive`.
- `proxy stats <podname> -output archive`


cli/cmd/proxy/stats/command.go
```diff
@@ -135,15 +159,24 @@ func (c *StatsCommand) Run(args []string) int {
 		RestConfig: c.restConfig,
 	}
 
+	if c.flagOutputFormat == "archive" {
+		proxyStatsFilePath := filepath.Join("proxy", "proxy-stats-"+c.flagPod+".json")
```
sanikachavan5: use fmt.Sprintf instead

LordAbhishek: updated

.changelog/4783.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+cli: add new flag -output to proxy stats cmd, this will enable user to write output to json file using -output flag.
```
sriramr98: nitpick: Since it's a public repo, please fix grammer

Suggestion:

cli: Added new flag `-output` to proxy stats command enabling the user to write output to json file

cli/cmd/proxy/stats/command.go
```diff
@@ -50,22 +63,29 @@ func (c *StatsCommand) init() {
 
 	f := c.set.NewSet("Command Options")
 	f.StringVar(&flag.StringVar{
-		Name:    "namespace",
+		Name:    flagNameNamespace,
 		Target:  &c.flagNamespace,
 		Usage:   "The namespace where the target Pod can be found.",
 		Aliases: []string{"n"},
 	})
+	f.StringVar(&flag.StringVar{
```
sriramr98, line 76: My understanding so far is that this can just be a boolean instead of a string only accepting `archive` as an input. A better flag can be `--output <true|false>`. Is there any reason why it's a specific string `archive` ? Are there plan in the future for extending this?

Even if we have plans of extending this, outputFormat should be an actual format like `json` or `hcl`. I'm not clear about the usage of `archive` here.

cli/cmd/proxy/stats/command.go
```diff
@@ -170,6 +204,48 @@ func (c *StatsCommand) getEnvoyStats(pf common.PortForwarder) (string, error) {
...
+	if err := os.MkdirAll(filepath.Dir(proxyStatsFilePath), 0755); err != nil {
+		return fmt.Errorf("error creating proxy stats output directory: %w", err)
+	}
+
+	var statsJson interface{}
+	err = json.Unmarshal(stats, &statsJson)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling JSON proxy stats output: %w", err)
+	}
+	marshaled, err := json.MarshalIndent(statsJson, "", "\t")
```
sriramr98: Why are we performing `Unmarshal` and `Marshal` back to back? Is it only for indentation / prettify purposes?

cli/cmd/proxy/stats/command.go
```diff
@@ -170,6 +204,48 @@ func (c *StatsCommand) getEnvoyStats(pf common.PortForwarder) (string, error) {
...
+	if err != nil {
+		return fmt.Errorf("error converting proxy stats output to json: %w", err)
+	}
+
+	if err := os.WriteFile(proxyStatsFilePath, marshaled, 0644); err != nil {
```
sriramr98: Extract 0644 to a constant and give a descriptive variable name. This is called a `Magic Number` and isn't good for readability.

cli/cmd/proxy/stats/command_test.go
```diff
@@ -37,6 +44,18 @@ func TestFlagParsing(t *testing.T) {
 			args: []string{"-namespace", "notaname"},
 			out:  1,
 		},
+		"Valid podName passed, Invalid arg value(so it can fail at validation)": {
```
sriramr98: Better test name is - `Valid podName passed, -output invalid value, should fail`

cli/cmd/proxy/stats/command_test.go
```diff
@@ -158,3 +177,145 @@ func startHttpServer(port int) *http.Server {
...
+
+			c := setupCommand(new(bytes.Buffer))
+			c.kubernetes = fake.NewSimpleClientset(&v1.PodList{Items: tc.pods})
+			c.flagNamespace = tc.namespace
+
+			for _, pod := range tc.pods {
+				c.flagPod = pod.Name
+				mpf := &MockPortForwarder{}
+
+				expectedFilePath := filepath.Join(tempDir, "proxy", "proxy-stats-"+pod.Name+".json")
```
sriramr98: Use `fmt.Sprintf` instead of `+`

sriramr98 (changes requested):

.changelog/4783.txt
LordAbhishek: Updated.

cli/cmd/proxy/stats/command.go
LordAbhishek: Basically, the requirement was to extend 3 proxy command(proxy list, stats & read) to support "writing output to a file". 
Other commands have existing outputFormats and they are already supporting json/table to output the data in specified format over terminal. 

So, to save the output, either we could have another flag like -save=true/archive=true to write the file in specified format by -output flag.. which i guess would be increasing code complexity. 
1. to avoid code complexity, i just went to use `archive` argument in output format which will write the output ONLY in JSON format. 
2. this will be consistent across all other proxy commands.

cli/cmd/proxy/stats/command.go
LordAbhishek: yes, it is only for indentation & prettify purpose. (otherwise json was wrote in  a single line in the file).

cli/cmd/proxy/stats/command.go
LordAbhishek: Updated.

cli/cmd/proxy/stats/command_test.go
LordAbhishek: Updated.

cli/cmd/proxy/stats/command_test.go
LordAbhishek: Updated.

cli/cmd/proxy/stats/command.go
sriramr98: In that case, you can just use `json.Indent` ( [ref](https://pkg.go.dev/encoding/json#Indent) )

sriramr98 (changes requested):

cli/cmd/proxy/stats/command.go
LordAbhishek: Updated.

sriramr98 (approved):

PR #4788, opened 2025-09-22
Add the -capture flag to proxy log command. 

LordAbhishek:
- Added -capture flag to proxy log command. 
- This will enable user to capture log for given duration and write them to a file in current working dir. 
- If -update-level flag is provided along with -capture flag, it will set given new log level, capture the log for given duration and reset the log level to existing one post log capture.
- `consul-k8s proxy log <podname> -capture 1m -update-level debug`

#### Signal interrupt handling:
- Here, with existing base command setup, if signal interrupt is received during log capture, it would exit the terminal at an instant, so log levels won't be reset back to existing levels. 
- So, to handle the signal interrupt while capturing log, I have modified main.go, commands.go, common/base.go which will handle signal interrupts with the help of a `cleanupReqAndCompleted` channel for any cli command in consul-k8s.  
- With the new setup of common/base.go and main.go, it will reset back the log levels back to previous one, basically main will wait unitll reset.


.changelog/4788.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+cli: add new `-capture` flag to proxy loglevel cmd. Also updated main.go, command.go and common/base.go to support graceful handle signal interrupts for any commands
```
sriramr98: This a public changelog which will be published. Need a more professional changelog. This looks more like a commit message than a changelog

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
+
+	// cleanup is required to ensure that if new log level set,
+	// should be reset back to existing log level after log capture
+	// even if user interrupts the command during log capture.
+	select {
+	case <-l.CleanupReq:
+	default:
+	}
+	l.CleanupReq <- true
+	defer func() {
```
sriramr98, line 342: Why two separate defer functions?

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
+	err = l.captureLogs()
+	if err != nil {
+		l.UI.Output(err.Error(), terminal.WithErrorStyle())
+	}
+	return err
+}
+
+func (l *LogLevelCommand) fetchExistingLogLevels(adminPorts map[string]int) (map[string]LoggerConfig, error) {
+	l.UI.Output(fmt.Sprintf("Fetching existing log levels..."))
+	existingLogger := make(map[string]LoggerConfig, 0)
```
sriramr98: Why are we creating a new map here only to be overriden at line 359?

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
+	newLogLevels := l.level
+	l.level = ""
```
sriramr98: Why are we updating it here just to override it again at line 360?

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
+	existingLogger, err := l.fetchOrSetLogLevels(adminPorts)
```
sriramr98: The function `fetchExistingLogLevels` denotes it's only fetching the existing level, but calling `fetchOrSetLogLevels` creates a side effects which is not a good coding practise.

sriramr98: Either change the function name `fetchExistingLogLevels` to something that more describes what the function does or move the `set` to outside the function

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
 	return nil
 }
 
+// fetchPodLogs - captures the logs from the given pod for the given duration and writes it to a file
+func (l *LogLevelCommand) fetchPodLogs(ctx context.Context) error {
+	duration, err := time.ParseDuration(l.capture)
+	if err != nil {
+		return fmt.Errorf("error parsing capture's duration")
+	}
+	durationChn := time.After(duration)
```
sriramr98: Create the `durationChn` in line 445. This is because we cann k8s api in line 423 which is an API call and will have some delay. This time will be consumed in the duration and the channel will fire earlier.

For example, if the duration is 0.5 seconds, but the api takes 0.6 seconds, then the select statement will immediately execute and won't wait for that 0.5 seconds.

Here's an example code to understand why this isn't an optimal implementation

```
package main

import "time"

func main() {

	start := time.Now()
	duration := time.Duration(5 * time.Second)
	durationCh := time.After(duration)
	println(duration.String())

	time.Sleep(3 * time.Second)

	startSelect := time.Now()
	select {
	case <-durationCh:
		println("Duration elapsed")
		println("Elapsed time since select:", time.Since(startSelect).String())
		println("Total elapsed time:", time.Since(start).String())

	}

}
```

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
+	proxyLogFilePath := filepath.Join("proxy", fmt.Sprintf("proxy-log-%s.log", l.podName))
+
+	// metadata of log capture
+	l.UI.Output("Pod Name:             %s", pod.Name)
+	l.UI.Output("Namespace:            %s", pod.Namespace)
+	l.UI.Output("Log Capture Duration: %s", l.capture)
+	l.UI.Output("Log File Path:        %s", proxyLogFilePath)
+
+	select {
+	case <-durationChn:
```
sriramr98, line 448: It's better to use `time.Since` here instead of creating a variable in line no 420

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
+		logs, err := l.getLogFunc(ctx, pod, podLogOptions)
+		if err != nil {
+			return err
+		}
+		// Create file path and directory for storing logs
+		// NOTE: currently it is writing log file in cwd /proxy only. Also, log file contents will be overwritten if
+		// the command is run multiple times for the same pod name or if file already exists.
+		if err := os.MkdirAll(filepath.Dir(proxyLogFilePath), 0755); err != nil {
```
sriramr98: Avoid magic numbers. Move 0755 to a constant with a descriptive variable name

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
+			return fmt.Errorf("error creating directory for log file: %w", err)
+		}
+		if err := os.WriteFile(proxyLogFilePath, logs, 0644); err != nil {
```
sriramr98: Avoid magic numbers. Move 0644 to a constant with a descriptive variable name

cli/cmd/proxy/loglevel/command.go
```diff
@@ -261,19 +290,195 @@ func (l *LogLevelCommand) fetchOrSetLogLevels(adminPorts map[string]int) error {
...
+	l.outputLevels(currentLogger)
+
+	err = l.captureLogs()
+	if err != nil {
+		l.UI.Output(err.Error(), terminal.WithErrorStyle())
+	}
+	return err
+}
+
+func (l *LogLevelCommand) fetchExistingLogLevels(adminPorts map[string]int) (map[string]LoggerConfig, error) {
```
sriramr98: The logic of this function is very confusing. Can you add unit tests specific to this function to ensure the implementation works as expected?

sriramr98 (changes requested):

.changelog/4788.txt
LordAbhishek: Updated

cli/cmd/proxy/loglevel/command.go
LordAbhishek: removed

cli/cmd/proxy/loglevel/command.go
LordAbhishek: removed

cli/cmd/proxy/loglevel/command.go
LordAbhishek: removed

cli/cmd/proxy/loglevel/command.go
LordAbhishek: removed `fetchExistingLogLevels`.
updated `fetchOrSetLogLevels` to accept 2 parameters - `fetchOrSetLogLevels(adminPorts map[string]int, logLevel string)`. 

Previously, `fetchOrSetLogLevels` was designed to work as both getter and setter & it was controlled by `l.level` flag.
Currently, we could control the behaviour of it with the help of `logLevel` parameter, so that we could use it whereever possible without modifying `l.level` (no side-effects).

Let me know if i need to modify this one.. i also tried to seperate the get and set functions.. that was also working perfectly fine but was it was repeating one. so removed that one and modified `fetchOrSetLogLevels`.

cli/cmd/proxy/loglevel/command.go
LordAbhishek: updated

cli/cmd/proxy/loglevel/command.go
LordAbhishek: keeping this one for now, 
select will not wait for anything else other than channels, isn't it ? 
so not sure how can we use  time.since here

```
start := time.now()
select {
case time.since(start) == l.capture: (this is boolean, select case: wont wait).
....
}
```
this is what i am able to think of wrt to time.since.. not sure if there is any other way.

cli/cmd/proxy/loglevel/command.go
LordAbhishek: updated

cli/cmd/proxy/loglevel/command.go
LordAbhishek: updated

cli/cmd/proxy/loglevel/command.go
LordAbhishek: REMOVED.

sriramr98 (approved):

sriramr98 (dismissed):

sreeram77 (dismissed):

kswap (approved):
Going ahead and unblocking @LordAbhishek with the merge, IT is already been approved by other reviewers

PR #4790, opened 2025-09-22
Update the status command to output the consul client & deployments status as well along with existing ones.

LordAbhishek:
- Updated the status command to output the consul client & deployments status in tabular format, along with the existing status.
- Updated the consul server status to output in tabular format in terminal, too.


cli/cmd/status/status.go
```diff
@@ -122,11 +124,10 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 
-	if err := c.checkConsulServers(namespace); err != nil {
-		c.UI.Output("Unable to check Kubernetes cluster for Consul servers: %v", err)
-		return 1
+	err = c.checkConsulComponentsStatus(namespace)
+	if err != nil {
+		return 1 // Only for testing
```
sanikachavan5: unable to infer from the command, is this for local testing? If so, do we need to keep it?

LordAbhishek: checkConsulComponentsStatus is doing the main core work along with printing the errors. 
The purpose of returning error is only in unit tests, to check if command run() is exiting with correct error codes. 

If we dont output error code, the unit test at line 165 will will not be able to assert if the checkConsulComponentsStatus returned any error or it passed successfully.

sriramr98: We don't need the comment since the method expects an integer return value anyways :D

cli/cmd/status/status.go
```diff
@@ -217,23 +218,117 @@ func validEvent(events []release.HookEvent) bool {
 	return false
 }
 
-// checkConsulServers prints the status of Consul servers if they
-// are expected to be found in the Kubernetes cluster. It does not check for
-// server status if they are not running within the Kubernetes cluster.
+// checkConsulComponentsStatus lists status of different consul components
```
sriramr98: A more detailed comment will be great

cli/cmd/status/status.go
```diff
@@ -217,23 +218,117 @@ func validEvent(events []release.HookEvent) bool {
...
+	} else {
+		c.UI.Output("No Consul Client found in Kubernetes cluster.")
+	}
+	return nil
+}
+
+// checkConsulServers list the Consul Servers and their ready status (number of pods ready/desired),
 func (c *Command) checkConsulServers(namespace string) error {
-	servers, err := c.kubernetes.AppsV1().StatefulSets(namespace).List(c.Ctx, metav1.ListOptions{LabelSelector: "app=consul,chart=consul-helm,component=server"})
+	servers, err := c.kubernetes.AppsV1().StatefulSets(namespace).List(c.Ctx, metav1.ListOptions{})
```
sriramr98: Is this a breaking change? The behaviour of this flow will change as we're removing the label selector filters

cli/cmd/status/status.go
```diff
@@ -217,23 +218,117 @@ func validEvent(events []release.HookEvent) bool {
...
+	}
+	c.UI.Output("Consul Deployment status: ", terminal.WithHeaderStyle())
+	if err = c.checkConsulDeployments(namespace); err != nil {
+		c.UI.Output("Unable to list Consul deployments: %s", err, terminal.WithWarningStyle())
+	}
+	return err
+}
+
+// checkConsulClients list the Consul Clients and their ready status (number of pods ready/desired)
+func (c *Command) checkConsulClients(namespace string) error {
```
sriramr98: `printConsulClients` might be a more apt name

cli/cmd/status/status.go
```diff
@@ -217,23 +218,117 @@ func validEvent(events []release.HookEvent) bool {
...
+		}
+		c.UI.Table(tbl)
+	} else {
+		c.UI.Output("No Consul Client found in Kubernetes cluster.")
+	}
+	return nil
+}
+
+// checkConsulServers list the Consul Servers and their ready status (number of pods ready/desired),
 func (c *Command) checkConsulServers(namespace string) error {
```
sriramr98: `printConsulServers` might be a more apt name

cli/cmd/status/status.go
```diff
@@ -217,23 +218,117 @@ func validEvent(events []release.HookEvent) bool {
...
 		}
+		c.UI.Table(tbl)
+	} else {
+		c.UI.Output("No Consul Server found in Kubernetes cluster.")
 	}
+	return nil
+}
 
+// checkConsulDeployments list the Consul Deployed Deployments and their ready status (number of pods ready/desired),
+func (c *Command) checkConsulDeployments(namespace string) error {
```
sriramr98: `printConsulDeployments` might be a more apt name

sriramr98: It would be better if this returns `tbl` so that unit tests can just use go native structs for assertions. 
The line `c.UI.Table(tbl)` can be moved to the caller

cli/cmd/status/status_test.go
```diff
@@ -48,17 +48,102 @@ func TestCheckConsulServers(t *testing.T) {
...
+				err := createServers("consul-server", namespace, int32(tc.desired), int32(tc.healthy), c.kubernetes)
+				require.NoError(t, err)
+			}
+
+			// Verify that the correct server statuses are seen.
+			err := c.checkConsulServers(namespace)
 			require.NoError(t, err)
 
+			actual := buf.String()
+			if tc.desired > 0 {
```
sriramr98, line 61: For testing, it'll be better if `checkConsulServers` returns an instance of `terminal.Table` and it can be used for assertions directly. Asserting the stdout as string is not a good idea.

sriramr98 (changes requested):

cli/cmd/status/status.go
LordAbhishek: Updated `checkConsulDeployments` to `getConsulDeoploymentTable` which returns tbl to caller and then caller calls  `printComponentStatus` for printing component status.
and in testing directly using `getConsulDeoploymentTable` to assert the table instance instead of buffer.

LordAbhishek: updated

cli/cmd/status/status_test.go
LordAbhishek: updated.

cli/cmd/status/status.go
LordAbhishek: updated to `getConsulServerTable`

cli/cmd/status/status.go
LordAbhishek: updated to `getConsulClientsTable`

cli/cmd/status/status.go
LordAbhishek: Thanks for catching this, added label selector for both consul servers and clients.

cli/cmd/status/status.go
LordAbhishek: updated.

sriramr98 (dismissed):

cli/cmd/status/status.go
```diff
@@ -217,24 +218,127 @@ func validEvent(events []release.HookEvent) bool {
...
+// getConsulClientsTable returns the table instance with the Consul Clients
+// and their ready status (number of pods ready/desired)
+func (c *Command) getConsulClientsTable(namespace string) (*terminal.Table, error) {
+	clients, err := c.kubernetes.AppsV1().DaemonSets(namespace).List(c.Ctx, metav1.ListOptions{LabelSelector: "app=consul,chart=consul-helm,component=client"})
+	if err != nil {
+		return nil, err
+	}
+	var tbl *terminal.Table
+	if len(clients.Items) != 0 {
+		tbl = terminal.NewTable("NAME", "READY", "AGE", "CONTAINERS", "IMAGES")
```
sriramr98: FOR FUTURE : This can be constants scoped to the package so that they can be reused in tests. No need to hold the PR for this now.

cli/cmd/status/status_test.go
```diff
@@ -48,17 +48,126 @@ func TestCheckConsulServers(t *testing.T) {
...
+			// Verify that the correct clients statuses are seen.
+			tbl, err := c.getConsulClientsTable(namespace)
+			require.NoError(t, err)
+
+			if tc.desired > 0 {
+				require.NotNil(t, tbl)
+				expectedHeaders := []string{"NAME", "READY", "AGE", "CONTAINERS", "IMAGES"}
+				assert.Equal(t, expectedHeaders, tbl.Headers)
+
+				require.Len(t, tbl.Rows, 1)
```
sriramr98, line 113: How come the test case `3 clients expected, 3 healthy` is passing when clearly 3 is epxected, but we assert for 1?

cli/cmd/status/status_test.go
```diff
@@ -48,17 +48,126 @@ func TestCheckConsulServers(t *testing.T) {
...
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			buf := new(bytes.Buffer)
+			c := getInitializedCommand(t, buf)
+			c.kubernetes = fake.NewSimpleClientset()
+
+			// Deploy clients if needed.
+			if name != "No deployments" {
```
sriramr98: For Future: This is not extendable code. Ideally the testCase should have a boolean parameter which defines if a deployment should be created incase the scope changes in future.

cli/cmd/status/status_test.go
```diff
@@ -48,17 +48,126 @@ func TestCheckConsulServers(t *testing.T) {
...
 
-			actual := buf.String()
-			if tc.desired != 0 {
-				require.Contains(t, actual, fmt.Sprintf("Consul servers healthy %d/%d", tc.healthy, tc.desired))
+			if tc.desired > 0 {
+				require.NotNil(t, tbl)
+				expectedHeaders := []string{"NAME", "READY", "AGE", "CONTAINERS", "IMAGES"}
+				assert.Equal(t, expectedHeaders, tbl.Headers)
+
+				require.Len(t, tbl.Rows, 1)
```
sriramr98, line 161: Same issue as above. I assume that it's 1 deployment, 3 instances, 3 healthy.

sriramr98 (changes requested):

cli/cmd/status/status.go
```diff
@@ -217,24 +218,127 @@ func validEvent(events []release.HookEvent) bool {
...
 		}
 	}
+	return tbl, nil
+}
 
-	return nil
+// getConsulDeploymentsTable returns the table instance with the Consul Deployed Deployments
+// and their ready status (number of pods ready/desired),
+func (c *Command) getConsulDeploymentsTable(namespace string) (*terminal.Table, error) {
+	deployments, err := c.kubernetes.AppsV1().Deployments(namespace).List(c.Ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
LordAbhishek: updated..

cli/cmd/status/status.go
LordAbhishek: updates this one too

cli/cmd/status/status_test.go
LordAbhishek: i am creating only one daemonset at line 100 and for that 3 replica is expected and 3 healthy..
so in table, that will be in a single row... (asserting the number of rows{daemonsets})

for expected vs healthy, i am asserting at line 117.

cli/cmd/status/status_test.go
LordAbhishek: updated..

cli/cmd/status/status_test.go
LordAbhishek: this one is also updated...

sriramr98 (approved):

PR #4800, opened 2025-09-26
Add `consul-k8s debug` command to consul-k8s cli.

LordAbhishek:
### Changes proposed in this PR

Added `consul-k8s debug` command to consul-k8s cli. 
The command will capture helm config, CRDs and its resources, consul-injected sidecar pods, consul-components(statefulsets, daemonsets, deployments, sidecars) pod logs and Envoy admin endpoint data for all the proxies(gateways and sidecars) and create archive of them.

#### Signal interrupt handling:
Use-case: Whenever debug run receives signal interrupt, it has to cleanup all the captured info. 
So, with existing base-command setup, if signal interrupt is received during debug run, it would exit the terminal at an instant(see main.go), so cleanup is not possible.
So, to handle the signal interrupt while capturing log, I have modified main.go, commands.go, common/base.go which will handle signal interrupts with the help of `cleanupReqAndCompleted` channel for any cli command in consul-k8s.
Another use-case: PR: https://github.com/hashicorp/consul-k8s/pull/4788

`consul-k8s debug -help`

> Capture debugging information from a Consul deployment on Kubernetes.
> 
> Usage: Consul-k8s debug [flags]
> 
> Command Options:
> 
>   -archive
>       Whether to archive the output debug directory to a .tar.gz. This is
>       aliased as "-a". The default is true.
> 
>   -capture=<string>
>       List of components to capture. Supported values are: all, helm, crds,
>       sidecar, pods, proxy. (e.g. -capture pods -capture proxy). This is
>       aliased as "-c". The default is all.
> 
>   -duration=<duration>
>       Capture logs of consul cluster for a given duration. (e.g., '5m' for the
>       logs of next 5 minutes). This is aliased as "-d". The default is 5m0s.
> 
>   -namespace=<string>
>       Namespace where the target Pod can be found. This is aliased as "-n".
>       The default is consul.
> 
>   -output=<string>
>       Filename of the debug output archive. This is aliased as "-o". The
>       default is consul-debug-2025-10-15T15-59-42+0530.
> 
>   -since=<duration>
>       Time duration from which to start capturing logs from consul cluster.
>       (e.g., '5m' for the logs of last 5 minutes). This is aliased as "-s".
> 
> Global Options:
> 
>   -kubeconfig=<string>
>       Set the path to kubeconfig file. This is aliased as "-kc".
> 
>   -kubecontext=<string>
>       Set the Kubernetes context to use.


anandmukul93:
I am not particularly convinced by the class/structure design and lack of DRY principles. everywhere i see namespaces being fetched separately, no context manager for each use case via enumerations, all public and private and utility methods has no segregation.  Was there any design done for any of this.

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,901 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to create compressed temp archive: %s", err)
+	}
+
+	g := gzip.NewWriter(f)
+	t := tar.NewWriter(g)
+
+	tempName = f.Name()
+
+	cleanup := func(err error) (string, error) {
```
yaminisridaran, line 888: can you please explain the logic here with Newwriter and cleanup.

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+		if err != nil {
+			err = multierror.Append(errs, fmt.Errorf("Unable to list pods for Consul Server- '%s': %v\n", server.Name, err))
+		}
+		workloads = append(workloads, workload{server.Name, "statefulsets", podsList})
+	}
+	// daemonset
+	for _, client := range l.components.clientList.Items {
+		podsList, err := l.getPodsForWorkload(client.Spec.Selector)
+		if err != nil {
+			err = multierror.Append(errs, fmt.Errorf("Unable to list pods for Consul Clients- '%s': %v\n", client.Name, err))
```
yaminisridaran: If we are not able to list the pods, arent we required to break out from this loop?

yaminisridaran (commented):
Provide few review comments

cli/cmd/debug/debug.go
LordAbhishek: gzip.newwriter is used to walk the "already created dir during debug run" and for each file it _compresses in gzip format_ and write it to temporary file `f`.  and 

tar.newwriter creates  _tar archive_  as files are added to the tar.Writer (using t.WriteHeader and io.Copy), they are formatted into the tar archive format and then passed to the gzip.Writer to be compressed. The overall effect is the creation of a gzipped tar archive (.tar.gz).

**cleanup function:** at line 872, we are trying to cleanup tar newwriter instance which we created at 821 and in case of error, it will invoke this cleanup fxn to retry the cleanup.

cli/cmd/debug/debugPodLogs.go
LordAbhishek: the loop ranges over the clients `clientList.Items` and **for each client**, we have to to get pods lists.. via `.getPodsForWorkload(client.Spec.Selector)` so.. if we arn't able to get pods list for a particular client/server/deployments, we still have to loop over for other clients... so no need to break the loop...

cli/main.go
```diff
@@ -42,6 +51,7 @@ func main() {
 		<-ch
 		// Any cleanups, such as cancelling contexts
 		cancel()
+		<-cleanupReqAndCompleted // by default this will be false, so this will proceed to exit, but if a command requires cleanup, it will wait here(emply channel)
```
sreeram77:
```suggestion
		<-cleanupReqAndCompleted // by default this will be false, so this will proceed to exit, but if a command requires cleanup, it will wait here(empty channel)
```

cli/main.go
```diff
@@ -30,7 +30,16 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	basecmd, commands := initializeCommands(ctx, log)
+	// Below channel is required to handle cleanup on signal interrupt
+	// - By default, main assumes that no command requires cleanup, so it sends false to the channel
+	// - If a command requires cleanup, it should read from this channel BEFORE context cancellation ,
+	// so channel will be empty and "signal handler goroutine" in main will wait unitll command sends true to the channel at line- 54
```
sreeram77:
```suggestion
	// so channel will be empty and "signal handler goroutine" in main will wait until command sends true to the channel at line- 54
```
Line - 54 is relative and can change in future.

cli/common/base.go
```diff
@@ -22,6 +22,9 @@ type BaseCommand struct {
 
 	// UI is used to write to the CLI.
 	UI terminal.UI
+
+	// CleanupConfirmation is the channel to update if cleanup is required and send confirmation that cleanup is done to main
```
sreeram77:
```suggestion
	// CleanupReqAndCompleted is the channel to update if cleanup is required and send confirmation that cleanup is done to main
```

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,927 @@
...
+	noHelmReleaseFound = "No helm release found"
+	noCRDsFound        = "No consul CRDs found in the cluster"
+	noSidecarPodsFound = "No consul injected sidecar pods found in all namespace"
+	noProxiesFound     = "No envoy proxy pods found in the cluster"
+	noPodsFound        = "No pods found to capture log"
+)
+
+var (
+	// Predefined errors - helpful for comparing error types
+	notFoundError        = errors.New("not found")
```
sreeram77: Use go naming convention for naming error variables - https://go.dev/wiki/Errors#naming

sreeram77 (changes requested):

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+			return signalInterruptError
+		}
+	}
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// =======================================================
```
sreeram77: What is this for?

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+
+// getWorkloadLogs - fetches logs 'of each containers' 'of each pods' 'of each workload items'.
+// write log status to logCaptureAudit file and errors to logCaptureErrors file.
+func (l *LogCapture) getWorkloadLogs() error {
+
+	// create logCaptureAudit file for each container logs collection
+	auditFilePath := filepath.Join(l.output, "logs", "logCaptureAudit.txt")
```
sreeram77: Please use a const for the file name.

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+	if err := os.MkdirAll(filepath.Dir(auditFilePath), dirPerm); err != nil {
+		return fmt.Errorf("error creating logCaptureAudit directory: %v", err)
+	}
+	auditFile, err := os.OpenFile(auditFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, filePerm)
+	if err != nil {
+		return fmt.Errorf("error creating logCaptureAudit file: %v", err)
+	}
+	w := tabwriter.NewWriter(auditFile, 1, 3, 2, ' ', 0)
+	fmt.Fprintln(w, "WORKLOAD-KIND\tWORKLOAD-NAME\tPOD-NAME\tCONTAINER-NAME\tSTATUS\tDETAILS")
```
sreeram77: Please move this to a const

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+	defer auditFile.Close()
+	defer w.Flush()
+
+	// log collection
+	resultsChan := make(chan logCollectionResult, l.totalContainers)
+	var wg sync.WaitGroup
+
+	l.logCollector(&wg, resultsChan)
+	go func() {
+		wg.Wait()
```
sreeram77, line 294: Why is the wait inside a go routine? This will not pause the main routine for collecting the logs.

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+		if err := os.WriteFile(errorFilePath, errorContent, filePerm); err != nil {
+			return fmt.Errorf("error writing log capture errors to file: %v\n Collected Errors:\n%v", err, errorContent)
+		}
+		return oneOrMoreErrorOccured
+	}
+	return nil
+}
+
+// getContainerLogAndUpdateResult - spawns goroutine to fetch logs for a container in a pod and write its status to results channel
+func (l *LogCapture) getContainerLogAndUpdateResult(wg *sync.WaitGroup, resultsChan chan<- logCollectionResult, cd containerData) {
```
sreeram77: Why is the WaitGroup passed as a param? This looks like an anti-pattern. We can just use:
`
var wg sync.WaitGroup

wg.Go(func() {
// Collect the data
})

wg.Wait()

`

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+	sem := make(chan struct{}, 10) // Buffered Channel Semaphore: limit to 10 concurrent goroutines
+	wg.Add(1)
+	sem <- struct{}{} // aquire semaphore {blocks if full}
```
sreeram77:
```suggestion
	sem <- struct{}{} // acquire semaphore (blocks if full)
```

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+}
+
+// resultCollectorAndAuditor - collects results & errors of each resource (from logCollector) and writes to audit & error file resp.
+func (l *LogCapture) resultCollectorAndAuditor(w *tabwriter.Writer, resultsChan <-chan logCollectionResult) error {
+	var logCaptureErrors *multierror.Error
+	var tabWriterMutex sync.Mutex
+	var auditWriteErrOnce sync.Once
+
+ReadLoop:
```
sreeram77: Why are we using a GOTO? It's not recommended to use GOTO. Please use a for loop.

cli/cmd/debug/debugPodLogs.go
sreeram77:
```suggestion
	sem <- struct{}{} // acquire semaphore (blocks if full)
```

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+func (l *LogCapture) fetchLogs(namespace, podName string, podLogOptions *corev1.PodLogOptions) (io.ReadCloser, error) {
+	podLogRequest := l.kubernetes.CoreV1().Pods(namespace).GetLogs(podName, podLogOptions)
+	podLogStream, err := podLogRequest.Stream(l.ctx)
+	if err != nil {
+		return nil, fmt.Errorf("error getting log stream: %v", err)
+	}
+	return podLogStream, nil
+}
+
+// =======================================================
```
sreeram77: What is this comment for?

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,387 @@
...
+func (l *LogCapture) outputLogCaptureMetadata() {
+	l.UI.Output(fmt.Sprintf(" - Total Pods:        %d", l.totalPods))
+	l.UI.Output(fmt.Sprintf(" - Total Containers:  %d", l.totalContainers))
+	if l.since != 0 {
+		l.UI.Output(fmt.Sprintf(" - Since:             %s", l.since))
```
sreeram77, line 420: Why do we need this space padding here?

cli/cmd/debug/debugProxyData.go
```diff
@@ -0,0 +1,350 @@
...
+		proxyType = mappedType
+	} else if proxyPod.Labels["api-gateway.consul.hashicorp.com/managed"] == "true" {
+		// Special case for deprecated API Gateway.
+		proxyType = "API Gateway(Depricated)"
+	}
+	return proxyType
+}
+
+func (e *EnvoyProxyCapture) writeEnvoyProxyPodList() error {
+	type podDataType map[string]map[string]string
```
sreeram77: Why do we require a 3 level map here? Can we use map[string]any?

sreeram77 (changes requested):

cli/cmd/debug/debug.go
LordAbhishek: updated.

cli/cmd/debug/debugPodLogs.go
LordAbhishek: Correct this will not pause the main routine for log collection. 

Basically, `logCollector` will collect log and push the collection status (result/error) to `resultChan` and  `resultCollectorAndAuditor` will be symultaneously process the results from `resultChan`. 
So, `resultCollectorAndAuditor` will only terminate once the `resultChan` is closed.  

Now, `resultChan` will not be closed untill `logCollector` finishes and `logCollector` internally spawns multiple goroutine to collect logs and only way to know if all goroutines are finished is by adding the wait group to each goroutine. 
So, thats why passing waitgroup in all goroutines, and creating a seperate goroutine which will be waiting till all log collection goroutine finishes, so that it closes the `resultChan`. 


I have updated it use worker pool pattern for controlled goroutines and few channels to track containers and result.

`// 

    // initialize channels
	l.resultsChan = make(chan logCollectionResult, l.totalContainers)
	l.containersChan = make(chan containerData, l.totalContainers)

	// initialize worker pool of log collector - getContainerLogAndUpdateResult
	var wg sync.WaitGroup
	numWorkers := int(math.Min(10, float64(l.totalContainers)))
	for i := 0; i < numWorkers; i++ {
		wg.Go(func() {
			l.getContainerLogAndUpdateResult()
		})
	}

	// fetch all containers of the workload and push it to containersChan
	l.pushWorkloadContainers()
	close(l.containersChan)

	// seperate goroutine to close resultsChan
	// as soon as all logCollector workers are done.
	go func() {
		wg.Wait()
		close(l.resultsChan)
	}()

	// resultCollectorAndAuditor-
	// read from resultsChan and write to audit file and error file.
	// Please Note: this function is blocking and will return only when
	// resultsChan is closed and all results are read from it or SIGINT is received.
	return l.resultCollectorAndAuditor(w)`

cli/cmd/debug/debugPodLogs.go
LordAbhishek: removed GOTO.

cli/cmd/debug/debugPodLogs.go
LordAbhishek: Updated it.

cli/cmd/debug/debugPodLogs.go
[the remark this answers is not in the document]
LordAbhishek: updated

LordAbhishek: updated.

LordAbhishek: updated

cli/cmd/debug/debugPodLogs.go
LordAbhishek: removed. was used as seperator.

cli/cmd/debug/debugPodLogs.go
LordAbhishek: to output in proper format.

cli/cmd/debug/debugProxyData.go
LordAbhishek: Yes, we require 3 level maps there, (you can see below example) but we can also use map[string]any. 
That will give us flexibility to have anytype of data, not just only string at runtime. but at the same time, we also lose the compile time type safety.

Currently keeping it. We will discuss and change it if needed. 

lets say we have 3 proxy pods, 2 sidecars and 1 api-gateway pod
json structure before writing to file will be:
```
{
	  "sidecars": [
        {
	      "pod1Name": {
	        "ready": "1/1",
            "status": "Running",
				...
	      }
	    },
	    {
	      "pod2Name": {
	        "ready": "1/1",
            "status": "Running",
				...
	      }
	    }
	  ],
	  "api-gateways": [
	    {
	      "pod3Name": {
	        "ready": "1/1",
            "status": "Running",
				...
	      }
	    }
	  ]
	}
```

cli/common/base.go
LordAbhishek: Updated.

cli/main.go
LordAbhishek: Removed "line - 54" part.

cli/main.go
LordAbhishek: Updated.

cli/cmd/debug/debugPodLogs.go
LordAbhishek: removed.

cli/cmd/debug/debugPodLogs.go
LordAbhishek: updated.

cli/cmd/debug/debugPodLogs.go
LordAbhishek: updated.

cli/cmd/debug/debugPodLogs.go
sreeram77: Please define proper format.

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,967 @@
...
+	Timestamp string   `json:"timestamp"`
+}
+
+// captureTask defines a single capture task to be performed by the debugger
+// including its name, target identifier, the function to call to perform the capture
+type captureTask struct {
+	name        string
+	target      string
+	captureFxn  func() error
+	notFoundMsg string
```
anandmukul93: this could be a bunch of errors . if this is error , name this as error. 
Notfound doesn't make sense if there are other types of errors. 
could be a better multierror type.

.changelog/4800.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+cli: add consul-k8s debug command to consul-k8s cli
```
sreeram77: Please describe briefly what the debug command does. 
eg: cli: add debug command to collect xyz logs from abc.

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,967 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/consul-k8s/cli/common"
```
sreeram77: The imports are not formatted properly. The general grouping is: 
std library

external

internal

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,967 @@
...
+
+var (
+	// Predefined errors - helpful for comparing returned errors
+	errNotFound        = errors.New("not found")
+	errSignalInterrupt = errors.New("signal interrupt received")
+
+	// ErrMultipleErrorsOccuredAndWritten is returned to stricly indicate that one or more errors occurred
+	// during a single capture task and they (multiple errors) are written successfully to debug bundle,
+	// In case if errors are not written, whole error would be printed on terminal
+	errMultipleErrorsOccuredAndWritten = errors.New(fmt.Sprint("one or more errors occurred during capture for this target",
```
sreeram77: fmt.Sprint is not required here. The text can be a single string.

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,967 @@
...
+	c.set = flag.NewSets()
+
+	f := c.set.NewSet("Command Options")
+	defaultOutputFilename := fmt.Sprintf("consul-debug-%v", time.Now().Format(timeDateFormat))
+
+	f.DurationVar(&flag.DurationVar{
+		Name:    flagDuration,
+		Target:  &c.duration,
+		Default: debugDuration,
+		Usage:   "To capture the logs of consul cluster for the a given duration",
```
sreeram77:
```suggestion
		Usage:   "Capture the logs of consul cluster for a given duration",
```

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,967 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func (c *DebugCommand) preChecks() error {
+	// Ensure the output directory can be created (have write permissions and it does not already exist
+	if _, err := os.Stat(c.output); os.IsNotExist(err) {
+		err := os.MkdirAll(c.output, 0755)
```
anandmukul93, line 308: this is more like prepare with check and also cleanup in case of partial failures. Keep the scope to this and rename according. 

rename to prepareForDebug

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,967 @@
...
+		Target:  &c.duration,
+		Default: debugDuration,
+		Usage:   "To capture the logs of consul cluster for the a given duration",
+		Aliases: []string{"d"},
+	})
+	f.DurationVar(&flag.DurationVar{
+		Name:    flagSince,
+		Target:  &c.since,
+		Default: 0,
+		Usage:   "The time duration since when to capture logs from pods",
```
sreeram77:
```suggestion
		Usage:   "The time duration from which to start capturing logs from pods",
```

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,967 @@
...
+	if err := c.set.Parse(args); err != nil {
+		c.UI.Output("Error parsing arguments: %v", err.Error(), terminal.WithErrorStyle())
+		return 1
+	}
+	// Validate the command line flags.
+	if err := c.validateFlags(); err != nil {
+		c.UI.Output("Invalid argument: %v", err.Error(), terminal.WithErrorStyle())
+		return 1
+	}
+	// Checks if cwd have write permissions and
```
sreeram77: Is this supposed to be cmd?

cli/cmd/debug/debug.go
```diff
@@ -0,0 +1,967 @@
...
+	if err != nil {
+		return fmt.Errorf("error writing index.json: %s", err)
+	}
+	return nil
+}
+
+// captureTarget returns true if the target capture type is enabled.
+// (Otherwords, is the target given in capture flag in command line?)
+func (c *DebugCommand) captureTarget(target string) bool {
+	if c.capture == nil || slices.Contains(c.capture, "all") || slices.Contains(c.capture, target) || target == "index" {
```
sreeram77: A switch case with fallthrough can make this easier to read.

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,416 @@
...
+	workloads = append(workloads, workload{"sidecar", "sidecars", proxyPodList})
+
+	l.workloads = workloads
+	return errs
+}
+
+// pushWorkloadContainers - pushes all containers of all pods of all workload items to containersChan
+func (l *LogCapture) pushWorkloadContainers() {
+	for _, workload := range l.workloads {
+		if len(workload.podsList.Items) == 0 {
```
sreeram77: podsList is a pointer. Please add a nil check.

cli/cmd/debug/debugPodLogs.go
```diff
@@ -0,0 +1,416 @@
...
+			l.resultsChan <- logCollectionResult{
+				StatusLine: fmt.Sprintf("%s\t%s\t%s\t%s\t%s\t%s", workload.kind, workload.name, "", "", "No Pods Found", "No Pods Found"),
+			}
+			continue
+		}
+		for _, pod := range workload.podsList.Items {
```
sreeram77, line 178: podsList is a pointer. Please add a nil check.

cli/cmd/debug/debugPodLogs.go
sreeram77: Is this for the output alignment?

cli/commands.go
```diff
@@ -6,6 +6,8 @@ package main
 import (
 	"context"
 
+	debug "github.com/hashicorp/consul-k8s/cli/cmd/debug"
```
sreeram77: Please group the imports

LordAbhishek: updated

cli/cmd/debug/debugPodLogs.go
LordAbhishek: updated

cli/cmd/debug/debugPodLogs.go
LordAbhishek: updated

cli/cmd/debug/debug.go
LordAbhishek: renamed it.

for cleanup case,
currenlty, cleanup happens only when signal interrupt is recieved for current debug run (signalInterrupt are handled by `runCapture`). 
otherwise, debug will run completely either with or without errors. 

Not sure if i understood correctly for cleanup thing, please let me know if i didn't.

.changelog/4800.txt
LordAbhishek: updated it.
cli: Add `debug` command to collect configs, logs, and other data from a Consul on Kubernetes deployment in a archive/dir.

cli/cmd/debug/debug.go
LordAbhishek: This one was to output the not found error. **Removed this one.**
So initially, all the `captureFunctions` returns  the generic `errNotFound` if they do not find any resource to capture. 
`runCapture` which is caller of each captureFunction, handles errors returned by them. 
Since all function returns the same `errNotFound`, _we need to output different contextual "not found" errors to terminal output for different capturesFunctions_, so thats why i had this `notFoundMsg`.. which will be printed when a captureFunction returns `errNotFound`. 

I have updated it.. now each captureFunction will return errors(if any) (including contextual not found errors).
`runCapture` handles all errors accordingly.

cli/cmd/debug/debug.go
LordAbhishek: updated for all files

cli/cmd/debug/debug.go
LordAbhishek: updated.

cli/cmd/debug/debug.go
LordAbhishek: updated

cli/cmd/debug/debug.go
LordAbhishek: updated

cli/cmd/debug/debug.go
LordAbhishek: it is current working directory. (CWD). i have removed short form from comments.

cli/cmd/debug/debug.go
LordAbhishek: i have removed `c.capture == nil || slices.Contains(c.capture, "all")`, they are already handled in `validateFlags`

now its simple 2 if conditions.

cli/cmd/debug/debugPodLogs.go
LordAbhishek: Yes.. this is for output alignment.

cli/cmd/debug/debugProxyData.go
LordAbhishek: this one is updated to use struct instead of maps.

sreeram77 (approved):

PR #4809, opened 2025-09-29
fix tests in api-gw connect peering packages

shashankNandigama:

Updated endpoints controller to use podIP from endpoints object instead of pod spec. This change is to resolve a racing scenario with current implementation. As endpoint and pod objects are updated independently, in a few cases endpoint is getting updated before pod leaving the pod IP to be empty when the controller gets triggered. By using pod IP from endpoint, it will have pod IP even if pod spec is not updated with pod IP.

Updated k8s dump collector used in testing to include init containers' info which helps in debugging issues.

Updated TestConnectInject_ProxyLifecycleShutdown to make sure that test pod is cleaned up by scaling down the deployment. This was causing an issue with cleanup because a new pod was getting created by the time tests are complete.

Updated consul-dns tests to use unprivileged port for consul-dns-proxy pod for testing. Another test dedicated to privileged port will be added as a follow-up.

Updates in endpoints controller were tested by building control-plane image on local and using that while installing consul helm chart on a kubernetes cluster. Then two services were created to test consul services registration and also validated pod to pod communication. Similarly tested gateway's service registration by creating an ingress-gateway object.

### Checklist ###
- [ ] Tests added

## PCI review checklist

### DIFF
--- a/acceptance/tests/api-gateway/api_gateway_kitchen_sink_test.go
+++ b/acceptance/tests/api-gateway/api_gateway_kitchen_sink_test.go
@@ -98,9 +98,11 @@ func TestAPIGateway_KitchenSink(t *testing.T) {
 		fixturePath += "-ent"
 	}
 
-	counter := &retry.Counter{Count: 60, Wait: 10 * time.Second}
-	retry.RunWith(counter, t, func(r *retry.R) {
-		out, err = k8s.RunKubectlAndGetOutputE(t, ctx.KubectlOptions(r), "apply", "-k", fixturePath)
+	// Use more frequent retries for resource creation
+	applyCounter := &retry.Counter{Count: 30, Wait: 5 * time.Second}
+	logger.Log(t, "applying api-gateway resources")
+	retry.RunWith(applyCounter, t, func(r *retry.R) {
+		out, err = k8s.RunKubectlAndGetOutputE(r, ctx.KubectlOptions(r), "apply", "-k", fixturePath)
 		require.NoError(r, err, out)
 	})
 
@@ -141,13 +143,16 @@ func TestAPIGateway_KitchenSink(t *testing.T) {
 	)
 
 	logger.Log(t, "waiting for gateway and httproute to be ready")
-	counter = &retry.Counter{Count: 6, Wait: 5 * time.Minute}
-	retry.RunWith(counter, t, func(r *retry.R) {
+
+	// Waiting for gateway to be ready.
+	gatewayCounter := &retry.Counter{Count: 30, Wait: 10 * time.Second}
+	logger.Log(t, "waiting for gateway to be ready")
+	retry.RunWith(gatewayCounter, t, func(r *retry.R) {
 		var gateway gwv1beta1.Gateway
 		err = k8sClient.Get(context.Background(), types.NamespacedName{Name: "gateway", Namespace: "default"}, &gateway)
 		require.NoError(r, err)
 
-		//CHECK TO MAKE SURE EVERYTHING WAS SET UP CORECTLY BEFORE RUNNING TESTS
+		//CHECK TO MAKE SURE EVERYTHING WAS SET UP CORRECTLY BEFORE RUNNING TESTS
 		require.Len(r, gateway.Finalizers, 1)
 		require.EqualValues(r, gatewayFinalizer, gateway.Finalizers[0])
 
@@ -155,13 +160,6 @@ func TestAPIGateway_KitchenSink(t *testing.T) {
 		checkStatusCondition(r, gateway.Status.Conditions, trueCondition("Accepted", "Accepted"))
 		checkStatusCondition(r, gateway.Status.Conditions, trueCondition("ConsulAccepted", "Accepted"))
 		require.Len(r, gateway.Status.Listeners, 2)
-
-		// http route checks
-		counter = &retry.Counter{Count: 6, Wait: 5 * time.Minute}
-		retry.RunWith(counter, t, func(r *retry.R) {
-			err = k8sClient.Get(context.Background(), types.NamespacedName{Name: "http-route", Namespace: "default"}, &httpRoute)
-			require.NoError(r, err)
-		})
 
 		require.EqualValues(r, int32(1), gateway.Status.Listeners[0].AttachedRoutes)
 		checkStatusCondition(r, gateway.Status.Listeners[0].Conditions, trueCondition("Accepted", "Accepted"))
@@ -172,6 +170,14 @@ func TestAPIGateway_KitchenSink(t *testing.T) {
 		require.Len(r, gateway.Status.Addresses, 2)
 		// now we know we have an address, set it so we can use it
 		gatewayAddress = gateway.Status.Addresses[0].Value
+	})
+
+	// Then, wait for HTTP route to be ready with its own retry loop
+	httpRouteCounter := &retry.Counter{Count: 50, Wait: 2 * time.Second}
+	logger.Log(t, "waiting for http route to be ready")
+	retry.RunWith(httpRouteCounter, t, func(r *retry.R) {
+		err = k8sClient.Get(context.Background(), types.NamespacedName{Name: "http-route", Namespace: "default"}, &httpRoute)
+		require.NoError(r, err)
 
 		// check our finalizers
 		require.Len(r, httpRoute.Finalizers, 1)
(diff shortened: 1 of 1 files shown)


acceptance/framework/k8s/debug.go
```diff
@@ -61,6 +62,36 @@ func WritePodsDebugInfoIfFailed(t *testing.T, kubectlOptions *k8s.KubectlOptions
 			logFilename := filepath.Join(testDebugDirectory, fmt.Sprintf("%s.log", pod.Name))
 			require.NoError(t, os.WriteFile(logFilename, []byte(logs), 0600))
 
+			if strings.Contains(pod.Name, "static-server") || strings.Contains(pod.Name, "static-client") {
```
anandmukul93: why name ? 
if tomorrow there is a different name used in some other test again this would require modification. is there a mechanism using something else like service pods , annotation on pods (service, gateways, etc) that we can add using a enumeration based rule function in the test framework,. can we avoid literal names. 
Also, shouldn't we get info for all init containers if present on spec anyways.

.changelog/4809.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:improvement
+control-plane: added retries in endpoints controller to avoid racing scenarios
+tests: updated tests in api-gateway, connect and peering packages to improve dump collection and cleanup
```
anandmukul93: **routine additions of unit tests, bug fixes in existing tests, or internal improvements to the testing process are typically not included in the main release notes.**

Anything relevant for usage of product to users like - performance-tests, upgrade-tests, migration tests, stability tests etc might be present in release notes.

acceptance/framework/consul/helm_cluster.go
```diff
@@ -157,7 +157,7 @@ func (h *HelmCluster) Create(t *testing.T) {
 	}
 	// Retry the install in case previous tests have not finished cleaning up.
 	retry.RunWith(&retry.Counter{Wait: 2 * time.Second, Count: 30}, t, func(r *retry.R) {
-		err := helm.InstallE(r, h.helmOptions, chartName, h.releaseName)
+		err := helm.UpgradeE(r, h.helmOptions, chartName, h.releaseName)
```
anandmukul93, line 160: We cannot do upgradeE i believe as the thought must be an expectation that a fresh cluster should either be installed or not. upgrade might cause some existing statefulsets to misbehave with persistent state information across upgrades.

acceptance/framework/k8s/debug.go
```diff
@@ -61,6 +62,36 @@ func WritePodsDebugInfoIfFailed(t *testing.T, kubectlOptions *k8s.KubectlOptions
...
+					}
+
+					// Write init container logs or err to file name <pod.Name>-<container.Name>-init.log
+					initLogFilename := filepath.Join(testDebugDirectory, fmt.Sprintf("%s-%s-init.log", pod.Name, container.Name))
+					require.NoError(t, os.WriteFile(initLogFilename, []byte(initLogs), 0600))
+
+					// If this init container restarted, get logs from previous instance
+					if pod.Status.ContainerStatuses != nil {
+						for _, status := range pod.Status.InitContainerStatuses {
+							if status.Name == container.Name && status.RestartCount > 0 {
```
anandmukul93: We already have this check here that makes it more apt for only logging failed init containers.

acceptance/tests/api-gateway/api_gateway_kitchen_sink_test.go
```diff
@@ -65,6 +65,12 @@ func TestAPIGateway_KitchenSink(t *testing.T) {
 	// Override the default proxy config settings for this test
 	consulClient, _ := consulCluster.SetupConsulClient(t, true, serverReleaseName)
 	logger.Log(t, "have consul client")
+
+	retry.Run(t, func(r *retry.R) {
```
anandmukul93, line 69: why do we need this peer check here as 1 ? 
is this test a single server setup ?

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -216,10 +216,38 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				}
 
 				// Skip registration if node information is incomplete to prevent duplicate registrations.
-				if pod.Spec.NodeName == "" || pod.Status.PodIP == "" || pod.Status.HostIP == "" {
+				// Retry for 5 seconds because the pod IP may not be immediately available after pod creation.
+				// retries := 5
+				// for range retries {
+				// 	if pod.Spec.NodeName == "" || pod.Status.PodIP == "" || pod.Status.HostIP == "" {
```
anandmukul93: why are we having so many comments here. Cleanup

anandmukul93: Add a single comment stating what is expected, why we chose this approach, why not retry probably also in a succint way would be useful

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -354,15 +382,15 @@ func parseLocality(node corev1.Node) *api.Locality {
 
 // registerGateway creates Consul registrations for the Connect Gateways and registers them with Consul.
 // It also upserts a Kubernetes health check for the service based on whether the endpoint address is ready.
-func (r *Controller) registerGateway(apiClient *api.Client, pod corev1.Pod, serviceEndpoints corev1.Endpoints, healthStatus string) error {
+func (r *Controller) registerGateway(apiClient *api.Client, pod corev1.Pod, podIP string, serviceEndpoints corev1.Endpoints, healthStatus string) error {
```
anandmukul93, line 357: are there no corresponding tests for this ? as we chaned the contract

acceptance/tests/api-gateway/api_gateway_kitchen_sink_test.go
shashankNandigama: It is a single server setup, this retry block is to make sure that the server is up before creating other app deployments which need containers injection.

acceptance/framework/consul/helm_cluster.go
shashankNandigama: This change is to address the issue:
`Error: INSTALLATION FAILED: cannot re-use a name that is still in use`
This consul helm installation is done in each acceptance test separately, so there will be a new consul cluster for each test. If helm install is not successful it doesn't delete any resources created, so in retries we get the error that cannot use release name that already exists. With helm.UpgradeE(helm upgrade --install), during the retries it will install the missing resources.

.changelog/4809.txt
shashankNandigama: Updated this, please check the new changelog

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
shashankNandigama: Cleaned up those comments

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
shashankNandigama: yes, this method doesn't have tests as of now

acceptance/framework/k8s/debug.go
shashankNandigama: Updated to capture logs of all failed init containers. Please review again

acceptance/framework/k8s/debug.go
shashankNandigama: Updated to capture logs of all failed init containers only.

acceptance/framework/consul/helm_cluster.go
anandmukul93: Ok got it

acceptance/tests/consul-dns/consul_dns_partitions_test.go
```diff
@@ -293,6 +293,9 @@ func setupClustersAndStaticService(t *testing.T, cfg *config.TestConfig, default
 		"dns.enabled":           "true",
 		"dns.proxy.enabled":     "true",
 		"dns.enableRedirection": strconv.FormatBool(cfg.EnableTransparentProxy),
+
+		// Configure DNS proxy to use a non-privileged port to work with K8s 1.30+
+		"dns.proxy.port": "8053",
```
anandmukul93, line 298: what was the earlier portrange used ? 
Can you share the code pointer where that is set ?

shashankNandigama: It was using 53 earlier. That is coming from default values of consul helm chart. https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/values.yaml#L1977

anandmukul93 (approved):
LGTM 
@yaminisridaran 
 1- raise and  fix backport and remove 1.22 changes into older branches 
 2 - add integration tests for 1.22 in main branch in a separate ticket for privileged ports in DNS and other resources like connect-injector etc. wherever priviledged port is being added in chart templates. 
 Create a story around it and tackle each test one by one in tasks. 
 3 - Design part - avoid if else in every flow, make the privileged ports tests as a separate consolidated utility in one file and use it everywhere as required.

PR #4887, opened 2025-10-16
Fix flaky test in sameness package

shashankNandigama:

In acceptance tests of sameness package 4 consul clusters are created and cluster peering is established among them. There is an intermittent issue where the peering acceptor object is not getting ready before the peering acceptor token is copied from one cluster to another which is causing the test to fail. Updated test to check for peering acceptor readiness and the secret to be present before we attempt to copy the token. It also recreates the peering acceptor if it is stuck without getting ready.

### Checklist ###
- [ ] Tests added

## PCI review checklist


krutibaraiya (approved):
LGTM

PR #4946, opened 2025-11-13
Support file-based envoy access logs via proxy-defaults for API gateway and sidecar proxies

anilvpatel:
- Fetch the global proxy-defaults configuration from the Consul server.

- When proxy-defaults has access logs enabled with a file-based sink, automatically inject a new emptyDir volume and corresponding volume mount into the consul-dataplane container.

- This ensures both API Gateway and sidecar proxies receive the required volume when file-based Envoy access logs are configured.

- Set up a local kind Kubernetes cluster.
- Built the local control-plane image:
````GOOS=linux make dev-docker````
- Loaded the built image into the kind cluster:
````kind load docker-image consul-k8s-control-plane:local````
- Applied the following Helm values while following the tutorial:
https://developer.hashicorp.com/consul/tutorials/get-started-kubernetes/kubernetes-gs-ingress
````
  image: hashicorp/consul:1.22.0
  imageK8S: consul-k8s-control-plane:local
  imageConsulDataplane: hashicorp/consul-dataplane:1.9
````
- Verified that the consul-dataplane pods for both sidecars and API Gateway received the expected volume and mount when file-based access logs were enabled.

- Build the control-plane image locally using make dev-docker.
- Load it into a test Kubernetes cluster (kind or equivalent).
- Enable file-based access logging in proxy-defaults.
- Confirm that a new volume and volumeMount are added to:
  - API Gateway dataplane pods
  - Sidecar proxy dataplane pods
### Checklist ###
- [x ] Tests added
- [ x] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)

### CODE REVIEW

**Review by santoshpulluri — COMMENTED** [review]
Changes looks good to me. 
Please give me a day to validate this in local setup.

File: control-plane/connect-inject/webhook/container_volume.go [hunk 1]
```diff
@@ -21,3 +24,19 @@ func (w *MeshWebhook) containerVolume() corev1.Volume {
 		},
 	}
 }
+
+func accessLogVolume() corev1.Volume {
```
sanikachavan5, line 28: I can see these are duplicated across multiple files, and the only differentiating factor is the name of the log volume. So can we accept that as a parameter, and then reuse these functions at all places? Maybe have a accesLogVolume_utils.go (thinking out loud)

control-plane/connect-inject/webhook/mesh_webhook.go
```diff
@@ -770,3 +778,44 @@ func sliceContains(slice []string, entry string) bool {
 	}
 	return false
 }
+
+// Fetches the global proxy-defaults config from Consul and checks if access logs are enabled.
+// If enabled and of file type, it returns the access log path to be used for creating a volume mount.
+func (w *MeshWebhook) additionalAccessLogVolumeMount(pod *corev1.Pod) error {
```
sanikachavan5: NIT
```suggestion
func (w *MeshWebhook) mountAdditionalAccessLogVolume(pod *corev1.Pod) error {
```

sanikachavan5 (changes requested):

control-plane/connect-inject/webhook/container_volume.go
anilvpatel: Both the access volume is for different package and it would be good if we keep separate to add additional feature scope based on their respective use-case in package.

sanikachavan5 (approved):

anilvpatel:
Duplicate PR: closing now 
Merged PR: https://github.com/hashicorp/consul-k8s/pull/5008

Issue #4894, opened 2025-10-20
Critical Safety Issues in API Gateway Controller


vorbidan:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

1. Gateway Controller Processes Non-Consul Gateways & Deletes Resources Without Provenance Validation.
The controller reconciles every Gateway resource in the cluster, regardless of gateway class.

https://github.com/hashicorp/consul-k8s/blob/e8fac3f2118f7b741ec2380fea36a21105d44b5e/control-plane/api-gateway/controllers/gateway_controller.go#L432-L433

2. Intentional Processing of Non-Consul Gateways
The code intentionally processes gateways not controlled by consul with explicit comments justifying this dangerous behavior.

https://github.com/hashicorp/consul-k8s/blob/e8fac3f2118f7b741ec2380fea36a21105d44b5e/control-plane/api-gateway/controllers/gateway_controller.go#L145-L147

3. Dangerous Deletion Logic Flow
When a non-consul gateway is processed:

https://github.com/hashicorp/consul-k8s/blob/e8fac3f2118f7b741ec2380fea36a21105d44b5e/control-plane/api-gateway/binding/binder.go#L98-L105

The gateway is treated as "Deleted",  and then a K8s resource is deleted.
Gatekeeper deletes resources based on name and namespace check only - no provenance validation, here is the example how Deployment is being deleted:

https://github.com/hashicorp/consul-k8s/blob/e8fac3f2118f7b741ec2380fea36a21105d44b5e/control-plane/api-gateway/gatekeeper/deployment.go#L81-L85

Other k8s resources, which do not "belong" to this controller are also deleted in a similar manner.

### Reproduction Steps

1. Deploy any non-consul gateway controller (kgateway)
2. Create a Gateway with their gateway class
3. Deploy consul-k8s with API Gateway controller
4. Watch consul delete resources it didn't create

### Logs

```
[...]

PR #4952, opened 2025-11-14
fix(api-gateway): prevent gateway controller from processing non-consul gateways

vorbidan:

Add early return check in gateway controller Reconcile method to skip
gateways that don't belong to the consul controller. This prevents the
controller from processing and potentially deleting resources for
gateways managed by other gateway controllers in multi-controller
environments.

The fix checks the GatewayClass.Spec.ControllerName against the consul
controller name (consul.hashicorp.com/gateway-controller) and returns
early if there's no match or if the GatewayClass is missing.

This follows the same pattern already used in the GatewayClass controller
and prevents unauthorized resource deletion that could occur when multiple
gateway controllers are deployed in the same cluster.

Fixes: Gateway controller reconciliation of non-consul gateways

Addresses Issue #4894

Added comprehensive unit tests (TestGatewayControllerReconcile) to validate:

- Gateways with non-consul GatewayClasses return early without processing
- Gateways with missing GatewayClasses return early without processing
- Gateways with consul GatewayClasses continue past the early return check

Locally built a docker image, deployed consul to k8s cluster via Helm chart to validate functionality

Run unit tests, build image locally, deploy to k8s with helm chart.

### Checklist ###
- [X] Tests added

## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.

### DIFF
--- a/.changelog/4952.txt
+++ b/.changelog/4952.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api-gateway: Fixed isuue when gateway controller processes non-consul gateways & deletes resources without provenance validation
+```
--- a/control-plane/api-gateway/controllers/gateway_controller.go
+++ b/control-plane/api-gateway/controllers/gateway_controller.go
@@ -94,6 +94,19 @@ func (r *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
+	// Early return if this gateway doesn't belong to the consul controller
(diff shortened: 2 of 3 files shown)


vdinesh4738:
Sorry for the delay; Why can't we watch for resources only linked to consul controller instead of early return at Reconcile?

vdinesh4738:
Also to continue the discussion, we now look for deletion stamp of the gw before deletion and thus deletion of other gw can't be done. But yes, currently consul tries to reconcile all the GWs and fixing it at watch should be good option.

PR #5053, opened 2025-12-23
fixed consul-dns, vault test cases

shani34:
Test cases fixes 
1. fixes related to dns 
- GKE-aware DNS handling: The code now detects GKE and uses stubDomains to extend DNS safely without breaking kube-dns behavior.
-  Clean config backups: Metadata like resourceVersion is stripped before restore, preventing conflicts during cleanup.
- Correct DNS restarts: The right DNS deployment (coredns, kube-dns, or rke2-coredns) is detected and restarted to avoid stale DNS.

2. Fixes related to Vault
 -  Explicit pod CIDR allocation: ip_allocation_policy.cluster_ipv4_cidr_block assigns a deterministic, non-overlapping pod IP range (10.100.0.0/14 split per cluster), which is critical for multi-cluster communication.
- Expanded firewall source ranges: source_ranges now allows traffic not only from the other cluster’s pod CIDR but also from its VPC subnet CIDR, ensuring pod-to-pod and node-level traffic works reliably across clusters.
- Targeted firewall scoping: target_tags restricts these firewall rules to only the Consul/K8s nodes, improving security while still enabling required cross-cluster traffic.


.github/workflows/pr-cloud.yaml
```diff
@@ -0,0 +1,26 @@
+name: pr-cloud-acceptance
```
shashankNandigama: this workflow was removed as public cloud workflows were inconsistent, let's not add it until all those issues are resolved. I think this was needed for GKE tests, but please remove this workflow for now. Let's add it back after all cloud tests issues are resolved.

shashankNandigama (changes requested):

.github/workflows/pr-cloud.yaml
shani34: Done

shashankNandigama (approved):
LGTM

PR #5062, opened 2026-01-02
CSL-13047: fix for segmentation fault or SIGSEGV with consul-cni network plugin's code on 1.9.1 

vdinesh4738:
Provided fix for the segmentation fault or SIGSEGV with consul-cni network plugin's code, which is common across the platform till date on Multus cni extension.

Background:
As per jira, when we enable cni and Multus, and then deploy the application with the api-resource 'network-attachment-definition' annotation like  'k8s.v1.cni.cncf.io/networks': '[{"namespace": "consul", "name":"consul-consul-cni"}]' is attached to the application pod, the application pod fails to come up at pod init with error in 

Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create pod network sandbox. (more detailed in the jira: `https://hashicorp.atlassian.net/browse/CSL-13047`

Root cause:
Creation of client with unavailable Kubeconfig or with the wrong Kubeconfig file name causing nil pointer dereference.

Fix:
Created the function to looks/fetch the latest Kubeconfig file and creates client.

Validation:
With the cni and Multus enabled: we update the iptables in the pod network ns.
With this fix : App is up and running and the iptables are updated in the pod network ns.


Surabhi-1605 (dismissed):

bharath-k1999 (dismissed):

control-plane/cni/main.go
```diff
@@ -274,19 +276,58 @@ func main() {
...
+	pattern := stable + "-*"
+	matches, err := filepath.Glob(pattern)
+	if err != nil {
+		return "", fmt.Errorf("glob failed for %s: %w", pattern, err)
+	}
+	if len(matches) == 0 {
+		return "", fmt.Errorf("no kubeconfig found at %s or %s-*", stable, stable)
+	}
+
+	// Pick newest by mtime (bounded + deterministic)
```
LordAbhishek: Can we use liner search here, instead of sorting ?

control-plane/cni/main.go
```diff
@@ -274,19 +276,58 @@ func main() {
...
+	sort.Slice(matches, func(i, j int) bool {
```
santoshpulluri: Don't we need to handle errors ?
```

sort.Slice(matches, func(i, j int) bool {
    fi, errI := os.Stat(matches[i])
    fj, errJ := os.Stat(matches[j])
    if errI != nil || errJ != nil {
        return false  // or handle appropriately
    }
    return fi.ModTime().After(fj.ModTime())
})
```

control-plane/cni/main.go
```diff
@@ -274,19 +276,58 @@ func main() {
 	skel.PluginMain(c.cmdAdd, cmdCheck, cmdDel, cniv.All, bv.BuildString("consul-cni"))
 }
 
+func resolveKubeconfigPath(dir, base string) (string, error) {
+	stable := filepath.Join(dir, base)
+	if fi, err := os.Stat(stable); err == nil && !fi.IsDir() {
```
santoshpulluri, line 282: Can you please add comments here appropriately !

Like a comment for explaining the case when we return and when we continue

control-plane/cni/main_test.go
```diff
@@ -194,6 +200,225 @@ func Test_cmdAdd(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			dir := t.TempDir()
+			tc.setup(t, dir)
+
+			got, err := resolveKubeconfigPath(dir, "kubeconfig")
+
+			if tc.expectError {
+				if err == nil {
+					t.Fatalf("expected error, got nil (path=%s)", got)
+					return
```
santoshpulluri, line 333: Whether `return` statement is required here !

control-plane/cni/main_test.go
```diff
@@ -194,6 +200,225 @@ func Test_cmdAdd(t *testing.T) {
...
+	path := filepath.Join(dir, "kubeconfig")
+
+	if err := os.WriteFile(path, []byte(data), 0600); err != nil {
+		t.Fatalf("failed to write kubeconfig: %v", err)
+	}
+
+	return path
+}
+
+func TestResolveKubeconfigPath(t *testing.T) {
```
santoshpulluri, line 246: Please add a comment/description for the newly introduced functions

LordAbhishek (approved):

santoshpulluri (approved):
LGTM

PR #5146, opened 2026-02-26
Fix eks terraform script

LordAbhishek:
**Existing Issue:** 
Currently eks workflows are failing at infrastructure provisioning in cloud nightly main tests.
Failing Workflow: https://github.com/hashicorp/consul-k8s-workflows/actions/runs/22421960130/job/64921986795 

```
Run # # On eks, we're setting the github run id instead of build URL because label values
╷
│ Error: Unsupported argument
│ 
│   on .terraform/modules/vpc/main.tf line 1030, in resource "aws_eip" "nat":
│ 1030:   vpc = true
│ 
│ An argument named "vpc" is not expected here.
╵
Error: Process completed with exit code 1.
```

**Fix eks terraform script**
* removed depricated versions, updated them with latest supported version.
* added storage class for consul servers pvcs

**Workflow with updated/fixed terraform script:**
https://github.com/hashicorp/consul-k8s-workflows/actions/runs/22436582955/job/64968467657 
(terraform apply step is successfully completed).


LordAbhishek:
Please let me know if this change needs changelog file.
I have added it by mistake and just want to confirm before removing it.

charts/consul/test/terraform/eks/main.tf
```diff
@@ -192,20 +276,32 @@ resource "aws_vpc_peering_connection_accepter" "peer" {
...
 }
 
-# Add routes that so traffic going from VPC 0 to VPC 1 is routed through the peering connection.
-resource "aws_route" "peering0" {
-  # We have 2 route tables to add a route to, the public and private route tables.
-  count                     = var.cluster_count > 1 ? 2 : 0
-  route_table_id            = [module.vpc[0].public_route_table_ids[0], module.vpc[0].private_route_table_ids[0]][count.index]
+# Add routes to all route tables in VPC 0 to route traffic to VPC 1 through the peering connection.
+resource "aws_route" "peering_private_0" {
+  count                     = var.cluster_count > 1 ? length(module.vpc[0].private_route_table_ids) : 0
```
anandmukul93: Direct Private Subnet Communication - By adding private subnet peering routes, you're enabling a direct path between worker nodes across VPCs that wouldn't exist in production if you're only using public subnet access.

Consul Mesh Behavior Divergence - In production, if Consul is configured to work through public subnet endpoints (NAT gateways, load balancers), the mesh would:

Route through public subnets
Potentially use different network paths
Have different latency/failure characteristics
Testing Gap - Your test environment would validate:

✅ Direct private-to-private node communication
✅ Public subnet routes (also present)
❌ But NOT the actual production traffic pattern
What This Means:

- Tests might pass because nodes can communicate via the "shortcut" private subnet route
- But in production, if private peering isn't allowed, the same tests could fail
- You're not validating the actual network path traffic will take

charts/consul/test/terraform/eks/main.tf
```diff
@@ -143,28 +180,75 @@ data "aws_eks_cluster_auth" "cluster" {
...
+  volume_binding_mode = "WaitForFirstConsumer"
+}
+
 # The following resources are only applied when cluster_count=2 to set up vpc peering and the appropriate routes and
 # security groups so traffic between VPCs is allowed. There is validation to ensure cluster_count can be 1 or 2.
 
 # Each EKS cluster needs to allow ingress traffic from the other VPC.
 resource "aws_security_group_rule" "allowingressfrom1-0" {
-  count             = var.cluster_count > 1 ? 1 : 0
+  count             = var.cluster_count > 1 ? length(module.vpc[1].private_subnets_cidr_blocks) : 0
```
anandmukul93: Same here see below comment on private subnet peering access. it doesnt simulate the general use case consul is built for in production.

charts/consul/test/terraform/eks/main.tf
LordAbhishek: removed private subnet route.

charts/consul/test/terraform/eks/main.tf
LordAbhishek: updated it as well.

charts/consul/test/terraform/eks/main.tf
```diff
@@ -91,6 +103,31 @@ module "eks" {
   tags = var.tags
 }
 
+resource "aws_ebs_encryption_by_default" "enable" {
+  enabled = true
```
santoshpulluri: @LordAbhishek  Why is it needed ? What is the impact & scope of adding this ?

santoshpulluri: quick googling mentioned this "`enables aws_ebs_encryption_by_default, which is an account-and-region-wide setting rather than a fixture-local change. That has a broader blast radius than the gp3 StorageClass resources and is the stronger review point.`"

LordAbhishek: This is needed to meet security compliance requirements from security team, whereby we need to encrypt all EBS volumes across all hashicorp accounts & regions, otherwise clusters/infra get auto deleted. 

We can remove the `aws_ebs_encryption_by_default` as we already create encrypted volumes from Storage class.

santoshpulluri (approved):
LGTM.

PR #5150, opened 2026-02-26
Add 3 additional sub testcases within peering gateway test

LordAbhishek:
**Add additional test case to peering_gateway.go.**

**This test case checks if api_gateway is able to route it request to another service which is present in another cluster. Both clusters are peered.**

**Existing Base validates:** if api_gateway reaches static-server across peers when both clusters use Local mesh_gateway mode via proxyDefaults.

**Additional cases:**
1. All services in both clusters use Remote mesh_gateway mode.
2. Only api_gateway uses Local mesh_gateway mode.
3. Only api_gateway uses Remote mesh_gateway mode.


LordAbhishek:
Please let me know if this change need changelog file to be added.

LordAbhishek:
https://github.com/hashicorp/consul-k8s-workflows/actions/runs/22461900895/job/65059168991
Peering test cases passed.

acceptance/tests/peering/peering_gateway_test.go
```diff
@@ -319,4 +319,45 @@ func TestPeering_Gateway(t *testing.T) {
...
+
+	// Test 2: only api_gateway uses Local mesh_gateway mode.
+	// proxyDefaults are already Remote for all services; override api_gateway with service-defaults.
+	// service-defaults takes precedence over proxy-defaults.
+	logger.Log(t, "patching api_gateway to set mesh_gateway mode to local")
+	out, err = k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "apply", "-f", "../fixtures/cases/api-gateways//service-defaults/gateway-local.yaml")
+	require.NoError(t, err, out)
+	helpers.Cleanup(t, cfg.NoCleanupOnFailure, cfg.NoCleanup, func() {
+		k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "delete", "-f", "../fixtures/cases/api-gateways//service-defaults/gateway-local.yaml")
+	})
```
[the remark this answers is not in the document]
LordAbhishek: fixed

acceptance/tests/peering/peering_gateway_test.go
```diff
@@ -319,4 +319,45 @@ func TestPeering_Gateway(t *testing.T) {
...
+	// Test 3: only api_gateway uses Remote mesh_gateway mode.
+	// Set proxyDefaults to Local for both clusters, then override api_gateway with service-defaults.
+	logger.Log(t, "patching api_gateway to set mesh_gateway mode to remote")
+	k8s.RunKubectl(t, staticClientPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"local"}}}`, "--type=merge")
+	k8s.RunKubectl(t, staticServerPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"local"}}}`, "--type=merge")
+	out, err = k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "apply", "-f", "../fixtures/cases/api-gateways//service-defaults/gateway-remote.yaml")
+	require.NoError(t, err, out)
+	helpers.Cleanup(t, cfg.NoCleanupOnFailure, cfg.NoCleanup, func() {
+		k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "delete", "-f", "../fixtures/cases/api-gateways//service-defaults/gateway-remote.yaml")
+	})
```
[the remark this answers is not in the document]
LordAbhishek: fixed

acceptance/tests/fixtures/cases/api-gateways/service-defaults/gateway-local.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: consul.hashicorp.com/v1alpha1
+kind: ServiceDefaults
+metadata:
```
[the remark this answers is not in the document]
LordAbhishek: added

acceptance/tests/fixtures/cases/api-gateways/service-defaults/gateway-remote.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: consul.hashicorp.com/v1alpha1
+kind: ServiceDefaults
+metadata:
```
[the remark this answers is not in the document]
LordAbhishek: added

acceptance/tests/peering/peering_gateway_test.go
```diff
@@ -319,4 +319,45 @@ func TestPeering_Gateway(t *testing.T) {
...
+	// when both clusters use Local mesh_gateway mode via proxyDefaults.
+	//
+	// Additional cases:
+	// 1. All services in both clusters use Remote mesh_gateway mode.
+	// 2. Only api_gateway uses Local mesh_gateway mode.
+	// 3. Only api_gateway uses Remote mesh_gateway mode.
+
+	// Test 1: all services in both clusters use Remote mesh_gateway mode.
+	logger.Log(t, "patching all services to set mesh_gateway mode to remote in both clusters")
+	k8s.RunKubectl(t, staticClientPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"remote"}}}`, "--type=merge")
```
santoshpulluri: @LordAbhishek 
Thanks for adding coverage for the additional mesh gateway mode combinations. The scenarios themselves are useful, but I have a concern about how they are modeled in the test today.

The title says adding additional sub testcases, but they are currently implemented as sequential mutations inside one long test rather than isolated subtests. That makes the new coverage stateful and order-dependent:

1. the first added case patches proxy-defaults in both clusters to remote
2. the second case relies on that mutated state and then applies a service-defaults override
3. the third case flips proxy-defaults back to local and applies a different override

Because of that, a partial failure or delayed reconcile in one case can affect the next one, which will make failures harder to diagnose and may introduce flakiness.

I think this would be much safer if each scenario were split into its own t.Run block with explicit setup/reset for the config it depends on. 

```
t.Run("all-services-remote", func(t *testing.T) {
    logger.Log(t, "patching all services to set mesh_gateway mode to remote in both clusters")
    k8s.RunKubectl(t, staticClientPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"remote"}}}`, "--type=merge")
    k8s.RunKubectl(t, staticServerPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"remote"}}}`, "--type=merge")

    // wait until config is effective
    k8s.CheckStaticServerConnectionSuccessful(t, staticClientOpts, staticClientName, targetAddress)
})
```

NOTE:  we should wait for config propagation after patching proxy-defaults or applying service-defaults before asserting connectivity, otherwise a successful request may still be exercising stale config rather than the intended mode.

LordAbhishek: @santoshpulluri Thanks for reviewing. I have isolated all the test cases. 
Regarding config propagation (and using Stale config), `CheckStaticServerConnectionSuccessful` internally retries for 15 minutes if connection fails. So, we do not have to worry about stale config or delayed reconcile.

acceptance/tests/peering/peering_gateway_test.go
```diff
@@ -317,6 +317,68 @@ func TestPeering_Gateway(t *testing.T) {
...
+	t.Run("api-gateway-local", func(t *testing.T) {
+		logger.Log(t, "patching all services to set mesh_gateway mode to remote in both clusters")
+		k8s.RunKubectl(t, staticClientPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"remote"}}}`, "--type=merge")
+		k8s.RunKubectl(t, staticServerPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"remote"}}}`, "--type=merge")
+		// Override api_gateway to use Local mesh gateway mode via service-defaults.
+		// service-defaults takes precedence over proxy-defaults.
+		logger.Log(t, "patching api_gateway to set mesh_gateway mode to local")
+		out, err := k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "apply", "-f", "../fixtures/cases/api-gateways/service-defaults/gateway-local.yaml")
+		require.NoError(t, err, out)
+		helpers.Cleanup(t, cfg.NoCleanupOnFailure, cfg.NoCleanup, func() {
```
santoshpulluri, line 357: @LordAbhishek 
The `gateway` ServiceDefaults override is currently removed via `helpers.Cleanup(...),` which looks parent-test scoped rather than subtest scoped. Since these subtests mutate shared cluster state, could we switch the cleanup here to `t.Cleanup(...) i`nside each subtest (and optionally delete before apply as well)? That would ensure the override is removed immediately after each subtest and avoid state leaking across cases.

The recommendations is to use 
```

_, _ = k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "delete", "-f", "../fixtures/cases/api-gateways/service-defaults/gateway-local.yaml")
out, err := k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "apply", "-f", "../fixtures/cases/api-gateways/service-defaults/gateway-local.yaml")
require.NoError(t, err, out)
t.Cleanup(func() {
	_, _ = k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "delete", "-f", "../fixtures/cases/api-gateways/service-defaults/gateway-local.yaml")
})
```

acceptance/tests/peering/peering_gateway_test.go
```diff
@@ -317,6 +317,68 @@ func TestPeering_Gateway(t *testing.T) {
...
+	t.Run("api-gateway-remote", func(t *testing.T) {
+		logger.Log(t, "patching all services to set mesh_gateway mode to local in both clusters")
+		k8s.RunKubectl(t, staticClientPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"local"}}}`, "--type=merge")
+		k8s.RunKubectl(t, staticServerPeerClusterContext.KubectlOptions(t), "patch", "proxydefaults", "global", "-p", `{"spec":{"meshGateway":{"mode":"local"}}}`, "--type=merge")
+		// Override api_gateway to use Local mesh gateway mode via service-defaults.
+		// service-defaults takes precedence over proxy-defaults.
+		logger.Log(t, "patching api_gateway to set mesh_gateway mode to remote")
+		out, err := k8s.RunKubectlAndGetOutputE(t, staticClientOpts, "apply", "-f", "../fixtures/cases/api-gateways/service-defaults/gateway-remote.yaml")
+		require.NoError(t, err, out)
+		helpers.Cleanup(t, cfg.NoCleanupOnFailure, cfg.NoCleanup, func() {
```
santoshpulluri, line 377: Same here.

acceptance/tests/peering/peering_gateway_test.go
LordAbhishek: @santoshpulluri ,

1. For the cleanup, `helpers.Cleanup(..., cleanupFxn)` is already at subtest scoped. 
It modifies the `cleanupFxn` according to `config Flags` and `Signal Interrupt` handler internally and passes it to `t.Cleanup()`. [ref](https://github.com/hashicorp/consul-k8s/blob/5ce7a4018e4e7c1229206dd8f8cc4977c33734ca/acceptance/framework/helpers/helpers.go#L98)
Direct usage of `t.Cleanup` will remove signal interrupt handler and config flags. I guess there is no need to remove `helpers.Cleanup`.
2. I will update the PR with explicit deletion of ServiceDefaults configs before applying.

PR #5167, opened 2026-03-05
Openshift Acceptance tests run with 2 clusters

Manishakumari-hc:
* PR with openshift connect acceptance test run changes.
* Triggers k8s-workflows cloud.yml wherein aks,eks,gks are commented for faster execution of openshift test cases.


acceptance/framework/connhelper/connect_helper.go
```diff
@@ -441,6 +447,11 @@ func (c *ConnectHelper) helmValues() map[string]string {
 		"dns.enableRedirection":        "true",
 	}
 
+	// On OpenShift, disable managing Gateway API CRDs since they already exist
```
vdinesh4738, line 450: this is not true on 4.18; Did you check this?

PR #5186, opened 2026-03-25
[Deprecate Ingress Gateway]sds support for api-gateway

bharath-k1999:
- added sds support for api-gateway at default, listner and service(ROUTE) level overrides
- New crd routetlssdsfilter is introduced to add sds override at route/service level

### Checklist ###
- [x] Tests added

## PCI review checklist


santoshpulluri:
@bharath-k1999 Can you please confirm if the Title and Description are accurate ? Its bit confusing to understand the purpose of this PR.

control-plane/api-gateway/common/tls_sds.go
```diff
@@ -0,0 +1,70 @@
...
+			clusterName = strings.TrimSpace(string(value))
+			clusterSet = true
+		}
+		if value, ok := tls.Options[gwv1.AnnotationKey(TLSSDSCertResourceAnnotationKey)]; ok {
+			certResource = strings.TrimSpace(string(value))
+			certSet = true
+		}
+	}
+
+	configured := clusterSet || certSet
```
sanikachavan5: Can we adjust SDS resolution to match the RFC’s source-precedence behavior?

The RFC says that for each listener:

- if both SDS keys are set in listener `tls.options`, those listener values are used
- otherwise, gateway annotations are used as defaults
- if neither source sets SDS, listener SDS is unset

The current implementation appears to merge the two sources field-by-field. For example, if gateway annotations provide `gateway-cluster + gateway-cert` and listener `tls.options` only provides `listener-cluster`, the resolved config becomes `listener-cluster + gateway-cert`.

That seems different from the RFC. Should partial listener `tls.options` fall back entirely to gateway annotations, or be rejected as incomplete listener SDS config? Either way, I think we should avoid producing a mixed listener/gateway SDS pair and add a test for this case.

sanikachavan5 (changes requested):

control-plane/api-gateway/common/tls_sds.go
sanikachavan5: https://github.com/hashicorp/consul-k8s/blob/a8d71db066effd485028b9b9ae669b351f42bbfb/control-plane/api-gateway/common/tls_sds.go#L24-L35

bharath-k1999: Thanks for calling this out — I’ve updated SDS resolution so we no longer produce mixed listener/gateway pairs.

### What was happening before
`ResolveListenerTLSSDSConfig` resolved SDS keys independently across listener `tls.Options` and gateway annotations.  
That allowed mixed results like `listener-cluster + gateway-cert` when only one listener SDS key was set.

### What was changed
In `tls_sds.go`, resolution now follows strict source precedence:

1. If listener `tls.Options` sets **both** SDS keys (`cluster_name` + `cert_resource`), use listener values exclusively.
2. If listener `tls.Options` sets **only one** SDS key, treat it as **invalid/incomplete** (no gateway fallback).
3. If listener sets **no** SDS keys, use gateway annotations as defaults.

So resolved SDS config always comes from a single source.

### Test coverage added
- `tls_sds_test.go`
  - listener has both keys → listener wins
  - listener has only `cluster_name` → invalid (even if gateway is complete)
  - listener has only `cert_resource` → invalid (even if gateway is complete)
  - listener has no SDS keys + gateway has both → valid via gateway
  - gateway has partial SDS only → incomplete signal

- `binding/validation_test.go`
  - added `TestValidateTLS` cases ensuring partial listener SDS returns `errListenerTLSSDSIncomplete`
  - validated success paths for full listener SDS and gateway-default SDS

### Result
Partial listener SDS is now surfaced as a clear validation error instead of being silently completed from gateway annotations.

control-plane/api-gateway/common/tls_sds.go
```diff
@@ -17,42 +17,86 @@ type EffectiveTLSSDSConfig struct {
...
 	if tls != nil {
-		if value, ok := tls.Options[gwv1.AnnotationKey(TLSSDSClusterNameAnnotationKey)]; ok {
-			clusterName = strings.TrimSpace(string(value))
-			clusterSet = true
+		listenerCluster, listenerClusterSet := listenerOptionValue(tls, TLSSDSClusterNameAnnotationKey)
+		listenerCert, listenerCertSet := listenerOptionValue(tls, TLSSDSCertResourceAnnotationKey)
+
+		if listenerClusterSet && listenerCertSet {
+			// Both keys present in the listener → use listener values exclusively.
+			if listenerCluster == "" || listenerCert == "" {
```
sanikachavan5, line 42: isn't this same as L59? why double check, this should be your else if block right? and gateway annotations should be else

sanikachavan5 (changes requested):

control-plane/api-gateway/common/tls_sds.go
bharath-k1999: I refactored this into `if / else if` so the precedence flow is explicit:

- `if listenerClusterSet && listenerCertSet`: listener is the selected source, and we validate both values are non-empty.
- `else if listenerClusterSet || listenerCertSet`: partial listener SDS is invalid, with no gateway fallback.
- `else`: listener has no SDS keys, so we evaluate gateway annotations next.

I kept the non-empty check in the first branch intentionally, since key presence and value validity are separate concerns (a key can be present with an empty value).

control-plane/api-gateway/common/resources.go
```diff
@@ -242,6 +244,51 @@ func (s *ResourceMap) ReferenceCountGateway(gateway gwv1.Gateway) {
...
+		key := IndexedNamespacedNameWithDefault(parent.Name, parent.Namespace, route.Namespace)
+		consulKey := NormalizeMeta(s.toConsulReference(api.APIGateway, key))
+
+		gateway, ok := s.gateways[consulKey]
+		if !ok {
+			continue
+		}
+
+		for _, listener := range gateway.Spec.Listeners {
+			if parent.SectionName != nil && listener.Name != *parent.SectionName {
```
sanikachavan5: Should this use the listeners the route actually binds to?

When `sectionName` is omitted, this walks all TLS listeners on the parent Gateway. But route binding still narrows that down by protocol/hostname/etc., so we could treat inheritance as ambiguous even when only one listener would actually accept the route. Was that intentional?

sanikachavan5 (changes requested):

charts/consul/templates/crd-routetlssdsfilters.yaml
```diff
@@ -0,0 +1,113 @@
...
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: RouteTLSSDSFilterSpec defines the desired state of RouteTLSSDSFilter.
+            properties:
+              sds:
+                description: SDS allows configuring TLS certificate from an SDS service.
+                properties:
+                  certResource:
```
sanikachavan5, line 64: since we are anyways rejecting empty or missing certResource, shouldn't this field be marked as required?

control-plane/api-gateway/common/resources.go
bharath-k1999: agreed, inheritance should use the listeners the route can actually bind to.

I patched `InheritedTLSSDSClusterForHTTPRoute` in `control-plane/api-gateway/common/resources.go` to only consider listeners that are compatible with HTTPRoute binding semantics (section match, HTTP/HTTPS protocol, allowed route kind, namespace policy, hostname compatibility), and to ignore non-terminate TLS listeners.

I also added regression coverage in `control-plane/api-gateway/common/resources_test.go`:
- `TestInheritedTLSSDSClusterForHTTPRoute_OmittedSectionUsesCompatibleListener`
- `TestInheritedTLSSDSClusterForHTTPRoute_IgnoresNonTerminateTLSListener`

Both tests pass in a non-cached run.

sanikachavan5 (approved):

PR #5223, opened 2026-04-14
Support multiport services in service mesh on Kubernetes

shashankNandigama:
Update endpoints controller to support multi-port service registrations.

Tested this on kind clusters in the combinations of following scenarios:

1. transparent-proxy enabled and disabled
2. Single DC same partition and cross-partition
3. mesh gateway modes: local, remote, none

In all the above 12 scenarios, a multi-port service (response) and a client service (hello) are created either in same kind cluster or different cluster according to the scenario. Connectivity from hello service to all 3 ports of response service is validated. Also multi-port service catalog registration and envoy configs of both services are verified to be as expected.

### Checklist ###
- [x] Tests added

## PCI review checklist


control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -1648,6 +1839,310 @@ func getMultiPortIdx(pod corev1.Pod, serviceEndpoints corev1.Endpoints) int {
...
+
+	if idx := getMultiPortIdx(pod, serviceEndpoints); idx >= 0 && idx < portCount {
+		return []int{idx}
+	}
+
+	return nil
+}
+
+func servicePortNameForConsul(serviceEndpoints corev1.Endpoints, token string, port int, idx int, seenNames map[string]int) string {
+	name := ""
```
sanikachavan5, line 2114:
```suggestion
	portName := ""
```

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -1648,6 +1839,310 @@ func getMultiPortIdx(pod corev1.Pod, serviceEndpoints corev1.Endpoints) int {
...
+	for _, subset := range serviceEndpoints.Subsets {
+		for _, endpointPort := range subset.Ports {
+			if int(endpointPort.Port) == port && endpointPort.Name != "" {
+				name = endpointPort.Name
+				break
+			}
+		}
+		if name != "" {
```
sanikachavan5, line 2122: do we need the port for just 1 subset?

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
```diff
@@ -430,22 +431,12 @@ func annotationProxyConfigMap(pod corev1.Pod) (map[string]any, error) {
 // createServiceRegistrations creates the service and proxy service instance registrations with the information from the
 // Pod.
 func (r *Controller) createServiceRegistrations(pod corev1.Pod, podIP string, serviceEndpoints corev1.Endpoints, healthStatus string) (*api.CatalogRegistration, *api.CatalogRegistration, error) {
-	// If a port is specified, then we determine the value of that port
-	// and register that port for the host service.
+
```
sanikachavan5, line 434: add helm tests

sanikachavan5 (changes requested):

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
shashankNandigama: no, we need to check for port in all subsets but the port for which the function is called will be present in only one of them. So we return just after finding port in any subset. This function gets the defined port name from endpoint which will be used in service registration.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
shashankNandigama: updated var name.

control-plane/connect-inject/controllers/endpoints/endpoints_controller.go
shashankNandigama: There are no changes in helm templates, so updating helm tests may not be required.

LakshmiNarayananDesikan (approved):

sanikachavan5 (approved):

PR #5252, opened 2026-04-24
fix: correct FIPS Consul version check in connect-init (CSL-13179)

Surabhi-1605:
## Summary

Fixes misleading WARN messages in the `consul-connect-inject-init` init container logs that appeared even when running a fully FIPS-compliant setup (all three images as FIPS variants).

**Jira:** CSL-13179
**Reproduced on:** consul-k8s 1.7.6-fips1402 through 1.8.8-fips1402 with consul-enterprise-fips 1.20.12/1.21.11

## Root Cause

The FIPS check in `connect-init/command.go` had **three bugs** that caused both warnings to always fire on every pod startup, regardless of whether Consul was a FIPS build or not:

1. **Wrong endpoint** — called `/v1/agent/version` which does not exist in Consul. Always returns a 404 error.
2. **Nil map bug** — `var versionInfo map[string]interface{}` is a nil map passed by value to `Raw().Query()`. The HTTP response body was never unmarshalled into it, so it was always empty.
3. **Missing else guard** — the second warning (`A non-FIPS version of Consul was detected`) fired unconditionally, even when the first call had already failed. Both warnings always appeared together.

## Fix

Read `state.DataplaneFeatures[DATAPLANE_FEATURES_FIPS]` directly instead of making a new API call.

The `consul-server-connection-manager` (v0.1.12, used across all affected release branches) **already calls `GetSupportedDataplaneFeatures`** via gRPC during watcher initialization, auto-injects the ACL token via its unary interceptor, and stores the parsed result in `state.DataplaneFeatures`. This means:
- No new gRPC call is needed
- Works correctly with ACLs enabled (token is handled by the watcher interceptor)
- `DATAPLANE_FEATURES_FIPS` is `true` only when the Consul server was built with the `fips` build tag — the authoritative server-side FIPS status

Reference: [consul-server-connection-manager watcher.go#L507-L522](https://github.com/hashicorp/consul-server-connection-manager/blob/v0.1.12/discovery/watcher.go#L507-L522)

## Backports

- #5253 → `release/1.7.x` (consul-k8s 1.7.6-fips1402, original reporter's version)
- #5254 → `release/1.8.x` (consul-k8s 1.8.8-fips1402, Abhineet's reproduced version)
- #5255 → `release/1.9.x` (latest supported)

## Testing

- `go build ./subcommand/connect-init/...` ✅
- `go vet ./subcommand/connect-init/...` ✅
- FIPS consul server (`DATAPLANE_FEATURES_FIPS=true`): no warnings ✅
- Non-FIPS consul server (`DATAPLANE_FEATURES_FIPS=false`): correct single warning ✅
- Features map nil (connection failure): unable-to-verify warning ✅

## Files Changed

- `.changelog/5252.txt`


control-plane/subcommand/connect-init/command.go
```diff
@@ -164,13 +164,13 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 	if version.IsFIPS() {
-		// make sure we are also using FIPS Consul
-		var versionInfo map[string]interface{}
-		_, err := consulClient.Raw().Query("/v1/agent/version", versionInfo, nil)
-		if err != nil {
```
vdinesh4738, line 170: _, err := consulClient.Raw().Query("/v1/agent/version", &versionInfo, nil)
if err != nil {
    log.Fatal(err)
}


You should probably try this. once

vdinesh4738 (changes requested):

.changelog/5252.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+connect-init: fix incorrect FIPS Consul version check that caused misleading WARN messages in the `consul-connect-inject-init` init container logs even when a fully FIPS-compliant setup was used. The original check called a non-existent API endpoint (`/v1/agent/version`) and used a nil map that was never populated, causing both FIPS warnings to fire unconditionally on every pod startup. The fix reads `state.DataplaneFeatures["DATAPLANE_FEATURES_FIPS"]` which is already populated by `consul-server-connection-manager` during initialization.
```
vdinesh4738: please update the change log **"original check called a non-existent API endpoint"** is redundant

vdinesh4738 (approved):

vdinesh4738 (changes requested):
please resolve the comment

vdinesh4738 (dismissed):
LGTM

Manishakumari-hc (dismissed):

Manishakumari-hc (approved):

vdinesh4738 (approved):

PR #5253, opened 2026-04-24
fix: correct FIPS Consul version check in connect-init (CSL-13179) [backport 1.7.x]

Surabhi-1605:
## Summary

Backport of #5252 → `release/1.7.x` (consul-k8s 1.7.x, includes 1.7.6-fips1402 — original reporter's version).

**Jira:** CSL-13179

## What was broken

The FIPS check in `connect-init/command.go` had three bugs causing both WARN messages to fire unconditionally on every pod startup:
1. Called `/v1/agent/version` — endpoint does not exist in Consul (404)
2. `var versionInfo map[string]interface{}` is nil/passed by value — response never populated
3. No `else` guard — 'non-FIPS detected' warning fired even when the API call failed

## Fix

Read `state.DataplaneFeatures["DATAPLANE_FEATURES_FIPS"]` — already populated by `consul-server-connection-manager v0.1.12` (used on this branch) during watcher initialization with ACL token auto-injected via the watcher's gRPC unary interceptor. Zero extra network calls.

## Files Changed
- `.changelog/_5252.txt`

### DIFF
--- a/.changelog/_5252.txt
+++ b/.changelog/_5252.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+connect-init: fix incorrect FIPS Consul version check that caused misleading WARN messages in the `consul-connect-inject-init` init container logs even when a fully FIPS-compliant setup was used.
+```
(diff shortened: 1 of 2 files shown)


santoshpulluri:
Consul K8s 1.7 is desupported.
cc @Surabhi-1605

PR #5254, opened 2026-04-24
fix: correct FIPS Consul version check in connect-init (CSL-13179) [backport 1.8.x]

Surabhi-1605:
## Summary

Backport of #5252 → `release/1.8.x` (consul-k8s 1.8.x, includes **1.8.8-fips1402** — the exact version Abhineet confirmed still broken on 2026-03-13).

**Jira:** CSL-13179

## What was broken

The FIPS check in `connect-init/command.go` had three bugs causing both WARN messages to fire unconditionally on every pod startup:
1. Called `/v1/agent/version` — endpoint does not exist in Consul (404)
2. `var versionInfo map[string]interface{}` is nil/passed by value — response never populated
3. No `else` guard — 'non-FIPS detected' warning fired even when the API call failed

## Fix

Read `state.DataplaneFeatures["DATAPLANE_FEATURES_FIPS"]` — already populated by `consul-server-connection-manager v0.1.12` (used on this branch) during watcher initialization with ACL token auto-injected via the watcher's gRPC unary interceptor. Zero extra network calls.

## Files Changed
- `.changelog/_5252.txt`

### DIFF
--- a/control-plane/subcommand/connect-init/command.go
+++ b/control-plane/subcommand/connect-init/command.go
@@ -164,13 +164,11 @@ func (c *Command) Run(args []string) int {
 		return 1
 	}
 	if version.IsFIPS() {
-		// Verify the Consul server is also a FIPS build. The consul-server-connection-manager
-		// already fetches supported dataplane features during initialization and stores them
-		// in state.DataplaneFeatures, so we read from there directly instead of making an
-		// additional gRPC call.
-		if state.DataplaneFeatures == nil {
-			c.logger.Warn("This is a FIPS build of consul-k8s, which should be used with FIPS Consul. Unable to verify FIPS Consul while setting up Consul API client.")
-		} else if !state.DataplaneFeatures["DATAPLANE_FEATURES_FIPS"] {
+		var versionInfo map[string]interface{}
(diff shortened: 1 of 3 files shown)


control-plane/subcommand/connect-init/command.go
```diff
@@ -164,13 +164,13 @@ func (c *Command) Run(args []string) int {
...
 	if version.IsFIPS() {
-		// make sure we are also using FIPS Consul
-		var versionInfo map[string]interface{}
-		_, err := consulClient.Raw().Query("/v1/agent/version", versionInfo, nil)
-		if err != nil {
+		// Verify the Consul server is also a FIPS build. The consul-server-connection-manager
+		// already fetches supported dataplane features during initialization and stores them
+		// in state.DataplaneFeatures, so we read from there directly instead of making an
+		// additional gRPC call.
+		if state.DataplaneFeatures == nil {
```
vdinesh4738: please do changes as asked in main pr.

vdinesh4738 (changes requested):

vdinesh4738 (approved):

PR #5270, opened 2026-04-27
**Do-not-merge** - Fix proxy lifecycle drain check to use Envoy local listener

santoshpulluri:
### Summary
This change updates the proxy lifecycle shutdown acceptance test to include a request to Envoy's local listener on localhost:1234 during the post-delete grace-period check.

### Why
When transparent proxy is enabled, iptables or ip6tables redirect rules may be removed during pod termination before the pod is fully gone. In that situation, a request to the service address can fail because traffic interception has already been torn down, even if Envoy is still serving locally.

The goal of this change is to make the test less dependent on transparent-proxy redirection during shutdown.

### What Changed
The shutdown test now seeds the curl command with localhost:1234 and adds comments explaining that the drain check should not rely solely on iptables-based interception during pod termination.

### Impact
This change narrows one source of ambiguity in the shutdown test by checking Envoy's local listener directly as part of the post-delete probe.

**It may help in cases where:**

the service-address request fails because transparent-proxy redirection has already been removed
Envoy is still running locally during the grace period
It does not address cases where:

Envoy's local listener on localhost:1234 is already unavailable
the failure is due to Envoy having already stopped accepting connections


panman90 (dismissed):

santoshpulluri:
Not Required anymore

PR #5306, opened 2026-05-12
Fix TestServiceResource_addIngress test

paras-gupta2:
- Fix failing test by manually triggering ingress upsert for fake clients, the fake Kubernetes client doesn't emit watch events for nested informers, so we directly call the ingress resource upsert handler to simulate what the informer would do in production.
- Refactor test to eliminate repetitive ingress definitions


shashankNandigama (approved):
LGTM

PR #5327, opened 2026-05-18
Fix boolean and string conversion issue

shani34:
** The Issue**
When users initiate an installation or upgrade using the consul-k8s CLI and provide a custom Helm values.yaml file, the CLI parses these values using internal Go structs (specifically in cli/helm/values.go) before rendering the Helm charts.
Several fields in the Helm chart theoretically support both string representations of booleans (e.g., "true") and native YAML booleans (e.g., true). However, the CLI structs had strictly typed these fields (such as server.enabled, client.enabled, dns.enableRedirection, ui.enabled, metrics.defaultEnabled) as string or bool. When users supplied raw boolean values in their values.yaml that conflicted with the strict struct type, the CLI's YAML unmarshaler threw an error, entirely failing the install or upgrade process. 

** The Fix**
To support the same type flexibility that native Helm does, the data types for these strict fields in cli/helm/values.go were changed to interface{}.
By using interface{}, the Go YAML parser can defer the strict type enforcement. It correctly unmarshals the input whether the user provides a boolean (true) or a string ("true"), safely passing the value downstream to the Helm execution layer which internally knows how to cast or process the values appropriately. 

How We Verified the Fix
To prove the fix completely resolves the issue, we functionally verified it locally using a kind cluster and the newly compiled CLI binary.
 Built the CLI: We compiled the updated code containing the interface{} changes using make cli-dev.
 Setup Test Values: We created a test-values.yaml file specifically leveraging native booleans: server:
```
 enabled: true
client:
enabled: false
dns:
enabled: true
 enableRedirection: false
ui:
enabled: true
metrics:
enabled: true 
```
 End-to-End Installation Test: We ran ./cli/bin/consul-k8s install -f test-values.yaml -wait=false against a fresh kind cluster.
 Result: The CLI successfully processed the file without throwing any unmarshal errors. It accurately mapped the overrides and installed the Consul pods (consul-server-0, consul-connect-injector, etc.) into the cluster.

Upgrade Verification
1. Setup Upgrade Values: We created a modified upgrade-values.yaml with updated boolean fields to enable the client and redirect DNS: server:
```
 enabled: true
client:
 enabled: true
 dns:
 enabled: true
 enableRedirection: true
ui:
 enabled: true
   metrics:
     enabled: false
```

End-to-End Upgrade Test: We ran ./cli/bin/consul-k8s upgrade -f upgrade-values.yaml -auto-approve -wait=false.


vdinesh4738:
Did you try consul-k8s cli installation with this helm?

vdinesh4738 (approved):
LGTM

PR #5335, opened 2026-05-20
add enterprise gating while registering services with multiple ports

shashankNandigama:
On clusters with consul CE while registering services with multiple container ports, send only the default port for service registration as multi-port services is an consul enterprise-only feature. Sending list of ports in consul catalog register api results in failure of connect-proxy service registration which can happen when a pod has more than 1 container port in the cluster using consul CE server.

-
-

### Checklist ###
- [x] Tests added

## PCI review checklist

### DIFF
--- a/control-plane/subcommand/inject-connect/v1controllers.go
+++ b/control-plane/subcommand/inject-connect/v1controllers.go
@@ -74,10 +74,19 @@ func (c *Command) configureControllers(ctx context.Context, mgr manager.Manager,
 		DefaultPrometheusScrapePath: c.flagDefaultPrometheusScrapePath,
 	}
 
-	isEnterpriseDistribution, err := consul.IsEnterpriseDistribution(consulConfig, watcher)
-	if err != nil {
-		setupLog.Info("Unable to validate Consul enterprise license for enterprise feature gating; multi-port service registrations will remain disabled until a valid license is detected", "error", err)
-		isEnterpriseDistribution = false
+	// If partitions are enabled, we know this is Consul Enterprise,
+	// so skip the license check. The license endpoint requires operator:read
+	// which is not allowed in partition-scoped ACL policies.
+	var isEnterpriseDistribution bool
+	if c.flagEnablePartitions {
+		isEnterpriseDistribution = true
+	} else {
+		var err error
+		isEnterpriseDistribution, err = consul.IsEnterpriseDistribution(consulConfig, watcher)
+		if err != nil {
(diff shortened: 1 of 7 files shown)


sanikachavan5 (approved):

sanikachavan5 (dismissed):

control-plane/subcommand/inject-connect/v1controllers.go
```diff
@@ -110,12 +127,6 @@ func (c *Command) configureControllers(ctx context.Context, mgr manager.Manager,
 		return err
 	}
 
-	isEnterpriseDistribution, err := consul.IsEnterpriseDistribution(consulConfig, watcher)
-	if err != nil {
```
anandmukul93, line 114: does this impact api gateway hpa scaling . why do we need to remove this while setting up controllers. have we moved this logic to hpa reconciler for gateway then ?

control-plane/subcommand/inject-connect/v1controllers.go
```diff
@@ -73,6 +73,22 @@ func (c *Command) configureControllers(ctx context.Context, mgr manager.Manager,
...
+	// so skip the license check. The license endpoint requires operator:read
+	// which is not allowed in partition-scoped ACL policies.
+	var isEnterpriseDistribution bool
+	if c.flagEnablePartitions {
+		isEnterpriseDistribution = true
+	} else {
+		var err error
+		isEnterpriseDistribution, err = consul.IsEnterpriseDistribution(consulConfig, watcher)
+		if err != nil {
+			setupLog.Info("Unable to validate Consul enterprise license for enterprise feature gating; multi-port service registrations will remain disabled until a valid license is detected", "error", err)
```
anandmukul93, line 87: warn log level

anandmukul93:
@sanikachavan5 does this impact hpa scaling as i see he has removed your logic for scaling feature. Can we refactor this to create a feature gating utility which handles gated features registry and exposes methods to check gating and gating limits (if we want to give some usage limit in CE vs ent), it should return appropritate error. 
log level should be warn for such errors.

control-plane/subcommand/inject-connect/v1controllers.go
shashankNandigama: it is not removed, just moved to before setting up endpoints controller as we need this flag there too. Same is used for gateway hpa controller.

anandmukul93 (approved):
LGTM

PR #5233, opened 2026-04-15
gateway networking templates and license

vdinesh4738:
new modified templates, added third party license and updated readme

PR #5181, opened 2026-03-23
custom crd feature

vdinesh4738:
Changes included:
1. Added custom controller under "consul.hashicorp.com" API group.
   (Impact: 
   1. New controller(custom controller) introduction to support the gateway under the above API group to support TCP under ocp 4.19+
   2. Old Controller (stable controller) is upgraded to package 1.5.1 of API gateway.networking.k8s.io
   3. Both controllers can be used across all K8s flavours.
   )
2. Gateway Reconcile to ONLY reconcile the consul controlled gateway --> introduced labels, and controller name check.
 ( Previously, we used to reconcile all the GWs, with this enhancement  we only watch for the consul controlled api gw at two stages one at watch -- with labels and secondary check in the reconcile with the controller Name)
4. Replaced Pod Security Admission controller inplace of the pod security policies.
5. Gateway Scaling at the run time from kubectl instead of updating the gateway class config only from the helm upgrade.

PR #5370, opened 2026-05-26
Dual GW Arch compatibility

vdinesh4738:
This PR will add the support for Dual Gateway - Standard API Gateway and the Consul API Gateway to support the customers on 1.8.x to upgrade their platforms which manage the Gateway Crds under the stable version, for example: ocp 4.18 to 4.19. 

This will unblock the customers and continue their operations.


santoshpulluri:
@vdinesh4738 Please update the PR description with relevant details.

charts/consul/templates/crd-referencegrants-external.yaml
```diff
@@ -8,10 +24,11 @@ metadata:
...
   labels:
-    app: {{ template "consul.name" . }}
-    chart: {{ template "consul.chart" . }}
-    heritage: {{ .Release.Service }}
-    release: {{ .Release.Name }}
+    # app: {{ template "consul.name" . }}
+    # chart: {{ template "consul.chart" . }}
+    # heritage: {{ .Release.Service }}
+    # release: {{ .Release.Name }}
+    name: consul-v1-refrencegrantss
```
santoshpulluri: Typo remove the extra 's' at the end.

acceptance/tests/partitions/partitions_gateway_test.go
```diff
@@ -343,6 +344,27 @@ func TestPartitions_Gateway(t *testing.T) {
...
+		out, err = k8s.RunKubectlAndGetOutputE(
+			t,
+			secondaryPartitionClusterStaticServerOpts,
+			"get", "httproute", "http-route", "-o", "yaml",
+		)
+		require.NoError(t, err)
+		logger.Logf(t, "HTTPRoute:\n%s", out)
+
+		logger.Log(t, "waiting for routing to stabilize")
+		time.Sleep(60 * time.Second)
```
santoshpulluri, line 366: [Suggestion] Replace this bare sleep with an active poll on the HTTPRoute's reconciliation status. The sleep adds a fixed 60s to every run regardless of how fast the controller actually reconciles, which compounds the already long runtime of the acceptance suite.

The HTTPRoute's Status.Parents field is populated by the controller once it has processed the patched spec — that's the correct signal to wait on rather than a wall-clock delay:

```go
// Wait for the HTTPRoute to be reconciled (parent status set by controller) before asserting connection behavior.
logger.Log(t, "waiting for route to be reconciled")
retry.RunWith(&retry.Counter{Count: 60, Wait: 5 * time.Second}, t, func(r *retry.R) {
    var route gwv1beta1.HTTPRoute
    err := k8sClient.Get(context.Background(), types.NamespacedName{Name: "http-route", Namespace: staticServerNamespace}, &route)
    require.NoError(r, err)
    require.NotEmpty(r, route.Status.Parents, "route should have at least one parent status")
})
```

This keeps the same 5-minute worst-case budget but exits as soon as the controller is done, saving time on the happy path. The k8sClient variable is already in scope from the gateway address check above.

CHANGELOG.md
santoshpulluri, line 1: This change is nothing to do with this PR. Please revert the changes.

Makefile
santoshpulluri, line 1: [Blocker] This target has been switched from docker buildx to podman, which introduces two problems:

1. Inconsistency — the rest of the Makefile still uses Docker.
control-plane-dev-skaffold, control-plane-dev-docker-multi-arch, and control-plane-fips-dev-docker all use docker/docker buildx. Any developer or CI runner without Podman installed will now get podman: command not found on make dev-docker while every other image target works fine.

2. The dev-docker parent target silently breaks.
The caller immediately does:
```go
dev-docker: control-plane-dev-docker
    docker tag '$(DEV_IMAGE)' 'consul-k8s-control-plane:local'

```
Podman and Docker maintain separate image stores. An image built with podman build is not visible to docker tag, so make dev-docker will either fail with Error: No such image or tag the wrong (stale) image with no obvious indication of why.

Root cause: The original intent was to drop --load from docker buildx build. With docker buildx, --load is required to copy the image from the BuildKit cache back into the local Docker image store. Without it, the image is built into the cache and immediately discarded — Podman was likely introduced to side-step that behaviour, but it trades one problem for two worse ones.

Suggested fix: Use plain docker build (not buildx), which always writes to the local Docker image store without needing --load:
```go
control-plane-dev-docker:
    @$(SHELL) $(CURDIR)/control-plane/build-support/scripts/build-local.sh --os linux --arch $(GOARCH)
    @docker build \
       --no-cache \
       --platform=$(GOOS)/$(GOARCH) \
       -t '$(DEV_IMAGE)' \
       --target=dev \
       --build-arg GOLANG_VERSION=$(GOLANG_VERSION) \
       --build-arg TARGETARCH=$(GOARCH) \
       --build-arg GIT_COMMIT=$(GIT_COMMIT) \
       --build-arg GIT_DIRTY=$(GIT_DIRTY) \
       --build-arg GIT_DESCRIBE=$(GIT_DESCRIBE) \
       -f $(CURDIR)/control-plane/Dockerfile $(CURDIR)/control-plane
```

If Podman support is intentional, it should be a separate PR that migrates all Makefile targets together, with explicit team sign-off. It should not be bundled into a gateway feature backport.

charts/consul/templates/generate-manifests-pvc.yaml
santoshpulluri, line 1: **generate-manifests-pvc.yaml** — and related generate-manifests-job.yaml / generate-manifests-apply-job.yaml

**[Blocker]** This PVC design has several operational problems that will affect production users:

**1. Permanent storage leak after `helm uninstall`**

```yaml
"helm.sh/resource-policy": keep
```

This annotation prevents Helm from ever deleting the PVC — including on `helm uninstall`. Every uninstall leaves a 1Gi PVC behind in the namespace with no automated cleanup. The cleanup kubectl command was explicitly commented out in the apply job, meaning there is currently **no path to remove this resource except a manual `kubectl delete pvc`**. This must be documented at minimum, and ideally a `pre-delete` hook should clean it up.

**2. Silent hang on clusters without a default StorageClass**

The PVC has no `storageClassName`, so on clusters without a default StorageClass (common in OCP, air-gapped, or custom environments) the PVC will stay `Pending` indefinitely. The `pre-upgrade` hook will then time out with a confusing error. At minimum, expose `storageClassName` as a Helm value:

```yaml
# values.yaml
global:
  generateManifests: false
  generateManifestsStorageClass: ""   # uses cluster default if empty
```

**3. Stale data risk between upgrades**

The PVC persists across upgrades but its contents are never explicitly cleared before the next `generate-manifests` job writes to it. If a prior upgrade left partial files (e.g. due to a mid-upgrade failure), those stale artifacts will be picked up by the subsequent `apply-manifests` job.

**4. `bitnami/kubectl:latest` is an unpinned floating tag**

```yaml
image: bitnami/kubectl:latest
imagePullPolicy: IfNotPresent
```

Pin this to a specific version (e.g. `bitnami/kubectl:1.35.2`) to avoid unintended breakage when the image is refreshed. `IfNotPresent` combined with `latest` means the first pull locks in whatever version happened to be `latest` at that time — a maintenance hazard.

**Broader suggestion:** The PVC-as-inter-hook-data-bus pattern is unusual for Helm charts. A simpler alternative would be to combine the generate and apply steps into a single job (write files to an `emptyDir` volume), eliminating the PVC entirely and all the persistence problems it brings.

charts/consul/templates/generate-manifests-apply-job.yaml
```diff
@@ -0,0 +1,87 @@
...
+        "openshift.io/scc": "restricted-v2"
+        {{- end }}
+        
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ template "consul.fullname" . }}-apply-sa
+      {{- if not .Values.global.openshift.enabled }}
+      containers:
+      - name: apply
+        image: bitnami/kubectl:latest
```
santoshpulluri, line 45: We should use Pinned version instead of pulling the latest ?

charts/consul/templates/crd-httproutes-external.yaml
```diff
@@ -8,10 +24,11 @@ metadata:
     gateway.networking.k8s.io/channel: experimental
   creationTimestamp: null
   labels:
-    app: {{ template "consul.name" . }}
-    chart: {{ template "consul.chart" . }}
-    heritage: {{ .Release.Service }}
-    release: {{ .Release.Name }}
+    # app: {{ template "consul.name" . }}
+    # chart: {{ template "consul.chart" . }}
```
santoshpulluri: What is the point of keeping commented block ?

charts/consul/templates/crd-referencegrants-external.yaml
```diff
@@ -8,10 +24,11 @@ metadata:
     gateway.networking.k8s.io/channel: experimental
   creationTimestamp: null
   labels:
-    app: {{ template "consul.name" . }}
-    chart: {{ template "consul.chart" . }}
-    heritage: {{ .Release.Service }}
-    release: {{ .Release.Name }}
+    # app: {{ template "consul.name" . }}
```
santoshpulluri: Same here, no point in keeping the commented block

charts/consul/templates/generate-manifests-podsecuritypolicy.yaml
```diff
@@ -0,0 +1,36 @@
...
+kind: PodSecurityPolicy
+metadata:
+  name: {{ template "consul.fullname" . }}-generate-manifests
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: enerate-manifests
```
santoshpulluri: Typo Error: 'g' is missing for generate-mafiests

acceptance/tests/api-gateway/api_gateway_tenancy_test.go
```diff
@@ -131,8 +132,19 @@ func TestAPIGateway_Tenancy(t *testing.T) {
 			k8sClient := ctx.ControllerRuntimeClient(t)
 			consulClient, _ := consulCluster.SetupConsulClient(t, c.secure)
 
+			// log the gateway and route for debugging purposes
+			var gateway gwv1.Gateway
+			err := k8sClient.Get(context.Background(), types.NamespacedName{Name: "gateway", Namespace: gatewayNamespace}, &gateway)
+			require.NoError(t, err)
+			logger.Logf(t, "gateway: %+v", gateway)
```
santoshpulluri, line 139: We shouldn't be using verbose logging !

charts/consul/templates/consul-helm-values-configmap.yaml
santoshpulluri, line 1: [Blocker] The gate condition is inconsistent with every other template in this feature set, and creates a real runtime failure.

Problem 1 — Wrong gate

This ConfigMap is gated on:

`{{- if .Values.terminatingGateways.enabled }}`

All 10 other generate-manifests-* templates are gated on:

`{{- if and .Values.connectInject.enabled .Values.global.generateManifests }}`

The [TerminatingGatewayController](vscode-file://vscode-app/Users/santoshgupthapulluri/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) calls [GetHelmValues()](vscode-file://vscode-app/Users/santoshgupthapulluri/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) on every reconcile and returns a hard error if the ConfigMap is missing. With the current gate, any install where generateManifests=true but terminatingGateways.enabled=false will result in the controller failing every reconcile loop with [configmap "helm-values" not found](vscode-file://vscode-app/Users/santoshgupthapulluri/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). The gate should be:

`{{- if and .Values.connectInject.enabled .Values.global.generateManifests }}`

Problem 2 — Name is not release-scoped

```
metadata:
  name: helm-values

```
Every other resource in this chart uses [{{ template "consul.fullname" . }}](vscode-file://vscode-app/Users/santoshgupthapulluri/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) as a prefix. This hardcoded name means two Consul releases in the same namespace will silently overwrite each other's ConfigMap. Should be:

`  name: {{ template "consul.fullname" . }}-helm-values`

Note: [GetHelmValues()](vscode-file://vscode-app/Users/santoshgupthapulluri/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [helmvalues/helpers.go](vscode-file://vscode-app/Users/santoshgupthapulluri/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) also hardcodes the lookup name to ["helm-values"](vscode-file://vscode-app/Users/santoshgupthapulluri/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and will need to be updated to match.

acceptance/tests/partitions/partitions_gateway_test.go
vdinesh4738: We can have this as part of the enhancements, since this is same in main. As part of enhancement, we can fix it in all release/s and main branch

acceptance/tests/partitions/partitions_gateway_test.go
```diff
@@ -343,6 +344,27 @@ func TestPartitions_Gateway(t *testing.T) {
 		logger.Log(t, "patching route to target server")
 		k8s.RunKubectl(t, secondaryPartitionClusterStaticServerOpts, "patch", "httproute", "http-route", "-p", `{"spec":{"rules":[{"backendRefs":[{"group":"consul.hashicorp.com","kind":"MeshService","name":"mesh-service","port":80}]}]}}`, "--type=merge")
 
+		logger.Log(t, "logging gateway + route for debugging")
+
+		out, err := k8s.RunKubectlAndGetOutputE(
+			t,
```
santoshpulluri, line 350: We should conditionally dump the logs by detecting failures. Otherwise, it will unnecessarily push the logs in both failure and successful scenarios.

charts/consul/templates/generate-manifests-pvc.yaml
vdinesh4738: please go through the other manifests using this object.

Yes, I agree the PVC is dangling in similar to PVC of consul. So ATM, when helm uninstall consul is executed, it leaves the PVC of consul connect injector as well. Which should be handled for both. I agree it should be deleted with the helm uninstall, but this will not leave  10 PVC if you perform "helm uninstall 10 times".

StorageClass by default uses the default storage class in the cluster, OCP any platform by default have the storage class, unless we want to use a specific storage class type. Then should be treated for all the releases.

charts/consul/templates/crd-httproutes-external.yaml
vdinesh4738: removed in main, somehow missed

.github/workflows/weekly-acceptance-1-1-x.yml
santoshpulluri, line 1: Consul currently supports only 2.0, 1.9 and 1.8 release branches. We should simply delete the older release branch related workflows and create the missing new release branch workflows. cc @LordAbhishek

charts/consul/templates/consul-helm-values-configmap.yaml
vdinesh4738: Not handled in this feature, delete file

acceptance/tests/partitions/partitions_gateway_test.go
vdinesh4738: Should be handled across the main and other release branches, should be an enhancement

charts/consul/templates/generate-manifests-apply-job.yaml
vdinesh4738: Will be taken as enhancement, it is present across main and other releases

control-plane/cni/main.go
```diff
@@ -55,6 +55,9 @@ const (
 	// annotationRedirectTraffic stores iptables.Config information so that the CNI plugin can use it to apply
 	// iptables rules.
 	annotationRedirectTraffic = "consul.hashicorp.com/redirect-traffic-config"
+
+	// annotationDualStack stores if pod need to run in dualstack mode
+	annotationDualStack = "consul.hashicorp.com/dual-stack"
```
santoshpulluri: Dual Stack not supported for 2.0

santoshpulluri (changes requested):
@vdinesh4738 

While reconciling this backport against the three source PRs (#5181, #5233, #5341), I confirmed the file counts and the dual-GW custom CRDs / OCP job / gateway-resources are all inline. Two areas from #5181 are not carried over, and I want to confirm this is intentional rather than an accidental drop:

GW Horizontal Pod Autoscaler (HPA) / autoscaling — gatekeeper/scaling.go + scaling_test.go and their wiring in gateway_controller.go are absent. The backport keeps only the existing min/max replica clamp in deploymentReplicas; the consul.hashicorp.com/hpa-* annotation handling, ParseScalingAnnotations, DetermineScalingMode/ReconcileScaling, and the api_gateway_scaling_test.go acceptance test are not present (grep for ParseScalingAnnotations|hpa-enabled|autoscalingv2 returns zero hits on this branch).

Dual-stack (IPv4/IPv6) networking — ConsulDualStackEnvVar, constants.Getv4orv6Str(...), the IPv6 bind addresses (::1/::), and the related flags (-envoy-admin-bind-address, -xds-bind-addr, -consul-dns-bind-addr, -graceful-addr) are not in the backport (zero hits for ConsulDualStackEnvVar|DualStack).

My understanding is that HPA-based GW autoscaling and dual-stack support are 2.0-only and planned/deferred for a later 1.8 release, so excluding them here is by design. Could you confirm that's the case? If so, no action needed — I just want it explicitly acknowledged so it's not a silent gap. Everything else in the backport reconciles cleanly with the source PRs.

control-plane/subcommand/inject-connect/command.go
```diff
@@ -247,6 +257,12 @@ func (c *Command) init() {
 	c.flagSet.BoolVar(&c.flagLogJSON, "log-json", false,
 		"Enable or disable JSON output format for logging.")
 
+	// enable TCP watch
+	c.flagSet.BoolVar(&c.flagEnableTCPRoute, "enabe-tcp-route", false, "Enables TCP Watch under gateway.networkings.k8s.io API Group")
```
santoshpulluri, line 261: nit: typo error.

missing l for enable in  "`enabe-tcp-route`"

santoshpulluri (dismissed):
The backport changes are inline with original changs. IPV6 support and auto-scaling are limited to only 2.0 release only.

santoshpulluri (dismissed):

santoshpulluri (approved):

PR #5402, opened 2026-06-10
Fix for the consul-injector-webhook issue during the OCP upgrade causing upgrade halt or cluster inaccessible

vdinesh4738:
Issue:
On upgrading the ocp to 4.18 to 4.19 and 4.19+, we encounter the consul-connect-injector web hook failing causing the ocp upgrade halt or making the cluster inaccessible, mainly when the schedule of the upgrade is delayed by some time.

Fix: To exclude the OpenShift namespace

### DIFF
--- a/.changelog/5402.txt
+++ b/.changelog/5402.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+helm: Update the namespace selector of consul-connect-inject webhook to be ignored incase of openshift 
+```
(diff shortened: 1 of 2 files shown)


.changelog/5402.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+helm: Update the namespace selector of consul-connect-inject webhook to be ignored incase of openshift 
+```
```
santoshpulluri, line 3: the selector change itself looks good.

one request before approval: could you please rewrite the changelog details ?

Something like this would read more clearly:

`helm: Exclude OpenShift cluster-monitoring namespaces from the consul-connect-inject webhook namespace selector by default`

santoshpulluri (approved):

PR #5398, opened 2026-06-09
security: install Uptycs EDR agent on EKS acceptance clusters

Surabhi-1605:
## Summary

Installs the **Uptycs EDR agent** (k8sosquery DaemonSet + kubequery) on every ephemeral EKS acceptance cluster so the worker nodes are covered by the security baseline, resolving **HC-COMPUTE-010 / SECVULN-44200**.

## Changes (`charts/consul/test/terraform/eks`)

- Add `hashicorp/helm` provider with per-cluster aliases (`cluster_0` / `cluster_1`) using EKS token auth.
- Add `helm_release` resources: `uptycs_0` / `kubequery_0` / `uptycs_1` / `kubequery_1`.
- Add variables: `uptycs_owner`, `uptycs_enroll_secret`, `uptycs_webhook_ca_bundle`, `uptycs_webhook_tls_crt`, `uptycs_webhook_tls_key`.
- Add `k8sosquery-values.yaml` (identical to AKS) and `kubequery-values.yaml` (AKS values with `cloudProvider: EKS`).

## Notes

The companion workflow change is hashicorp/consul-k8s-workflows#128.

> The branch name still references `hc-security-base`; the implementation has moved to the Uptycs EDR approach to match the merged GKE/AKS precedent. Branch left as-is to avoid opening a new PR.

PR #5409, opened 2026-06-11
ci(eks): install hc-security-base host package for HC-COMPUTE-010

Surabhi-1605:
## What
Adds an **additive, toggle-gated** path so EKS acceptance worker nodes carry the `hc-security-base` Debian package required by **HC-COMPUTE-010 / SECVULN-44200**.

When `enable_security_baseline=true` (default **false**):
- worker nodes launch from a Canonical **Ubuntu EKS-optimized AMI**, and
- `hc-security-base` is installed at boot from internal Artifactory (mirrors `ami-builder/scripts/packer-install_meta_deb.sh`).

## Why
The merged Uptycs EDR DaemonSet (PR #5398) satisfies the Kubernetes posture control (HC-COMPUTE-011-K8) but does **not** register a package in the host DB. The Wiz/Uptycs scanner for HC-COMPUTE-010 checks `dpkg -l | grep hc-security-(base|pre-uptycs|wasatch)`, so the VMs stayed flagged after merge. This adds the host package **on top of** the DaemonSet (both are kept).

## Safety
- Default **OFF** → zero change to current CI until enabled via the workflow toggle.
- Uptycs `helm_release` resources untouched.
- Verified: `terraform fmt`/`validate` clean, `bash -n` on template OK, console OFF-path = default Amazon Linux node group (no AMI lookup, no launch template).

Enabled by companion PR in consul-k8s-workflows.

### DIFF
--- a/charts/consul/test/terraform/eks/main.tf
+++ b/charts/consul/test/terraform/eks/main.tf
@@ -84,9 +84,10 @@ locals {
   }
 
   # User-data that installs hc-security-base from internal Artifactory at boot.
+  # Credentials are base64-encoded so they embed safely regardless of content.
   hc_security_base_pre_userdata = var.enable_security_baseline ? templatefile("${path.module}/templates/install-hc-security-base.sh.tpl", {
-    afy_user     = var.afy_user
-    afy_password = var.afy_password
+    afy_user_b64     = base64encode(var.afy_user)
+    afy_password_b64 = base64encode(var.afy_password)
   }) : ""
 
   # Hardened node group: Canonical Ubuntu EKS AMI + hc-security-base install.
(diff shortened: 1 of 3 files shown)


LordAbhishek:
Hi @Surabhi-1605,
Please test the changes by:
1. running a single acceptance test on EKS in the nightly workflow (`consul-k8s-ci` account) to verify if the clusters created as part of the acceptance test stills gets caught up by the wiz or not, or
2. creating simple cluster from the terraform file with this change in your AWS personal sandbox account and if the change doesn't work, we as well get the same SECVULN after 48h. (i guess, this personal account cluster will not be caught in wiz)
These are the 2 ways i find to verify the changes.

PR #5491, opened 2026-07-13
argocd fix

vdinesh4738:
Issue: Sev 1 from the customer since ArgoCD sync is hanging infinitely after PVC creation with VolumeBinding Mode set to "WaitForFirstConsumer".

this Fix requires a storageClass with attribute VolumeBinding to "immediate" created. ( Should be documented).

introduced the options/flags:

generateManifestsStorageClassName: <pass the storageClassName (defaults standard)>
generateManifestsStorage: <pass the storage required ( default is 1Gi)>

### DIFF
--- a/charts/consul/values.yaml
+++ b/charts/consul/values.yaml
@@ -800,7 +800,7 @@ global:
   
   generateManifests: false
 
-  # Uses the provided storage class name for the generated manifests. If not provided, the default storage class will be used.
+  # Uses the provided storage class name for the generated manifests- recommended to create SC with volumebinding as Immediate. If not provided, the default storage class will be used.
   generateManifestsStorageClassName: "standard-imm"
 
   # Storage requirement for generateManifests pvc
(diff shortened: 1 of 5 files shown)


charts/consul/values.yaml
```diff
@@ -800,6 +800,12 @@ global:
   
   generateManifests: false
 
+  # Uses the provided storage class name for the generated manifests- recommended to create SC with volumebinding as Immediate. If not provided, the default storage class will be used.
+  generateManifestsStorageClassName: "standard-imm"
```
shashankNandigama: Can we leave this value unset so that a Kubernetes default storageClass is used if user doesn't set this? If there is a storageClass with another name and users expect that to be used by default, this can cause issue if they don't pass it in helm values.

vdinesh4738: got it

shashankNandigama (approved):
LGTM

PR #5516, opened 2026-07-23
api-gateway: fix ext_authz filter update propagation

bharath-k1999:
- **Fix [`httpRouteRulesEqual()`](control-plane/api-gateway/common/diff.go) to include `ExtAuthz` in its equality check.**
  The function compared every `api.HTTPFilters` field (`RetryFilter`, `TimeoutFilter`, `JWT`) but omitted `ExtAuthz`. Any reconciler cycle that only changed `Filters.ExtAuthz` (e.g. swapping an `HTTPRoute`'s `ExtensionRef` from `ext-authz-enabled` → `ext-authz-disabled`) would see `EntriesEqual → true` and silently skip the Consul write, leaving stale `ext_authz` enforcement on the data plane. A new `extAuthzFiltersEqual` helper is wired in via `bothNilOrEqualFunc`, exactly mirroring the `JWT` filter pattern. The same fix is applied to `api-gateway-custom` for consistency.
- **Add unit tests** (`TestEntriesEqual_HTTPRoute_ExtAuthz`) covering nil/nil, enabled/enabled, disabled/disabled, enabled↔disabled, and nil↔enabled cases — the enabled-vs-disabled case is the direct regression test.
- **Extend the `TestAPIGateway_ExtAuthz_OIDC` acceptance test** with a two-step filter-toggle regression scenario: patch the `HTTPRoute` to `ext-authz-disabled` → assert Consul `ExtAuthz.Enabled=false` and unauthenticated traffic returns `200`; patch back to `ext-authz-enabled` → assert `ExtAuthz.Enabled=true` and unauthenticated traffic is denied (`302`). New `RouteAuthFilter` fixture CRDs (`ext-authz-enabled` / `ext-authz-disabled`) are added to support this.
- **Remove duplicate function definitions** in [`translation.go`](control-plane/api-gateway/common/translation.go) (`routeExtAuthzFromAnnotations`, `gatewayExtAuthz`) that were dead code left by a prior merge conflict.

- **Unit:** `go test ./control-plane/api-gateway/common/... ./control-plane/api-gateway-custom/common/...` passes including the new `TestEntriesEqual_HTTPRoute_ExtAuthz` suite.
- **Acceptance:** `TestAPIGateway_ExtAuthz_OIDC` (Consul Enterprise, requires a license) now exercises the two-step filter-toggle regression. On an unfixed binary the Consul assertion loop times out; on the fixed binary both steps pass cleanly.

- Review [`diff.go`](control-plane/api-gateway/common/diff.go) to confirm `ExtAuthz` is now included in `httpRouteRulesEqual` using the same `bothNilOrEqualFunc` pattern as the other filters.
- Run `go test ./control-plane/api-gateway/common/... -run TestEntriesEqual_HTTPRoute_ExtAuthz -v` locally (no cluster required) to exercise the regression unit tests.
- Optionally run `TestAPIGateway_ExtAuthz_OIDC` against an Enterprise cluster to validate the end-to-end filter-toggle behaviour.

### Checklist ###
- [x] Tests added

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.
  > Bug fix: the diff layer in the API-gateway reconciler omitted `ExtAuthz` from its equality check, causing in-place `RouteAuthFilter` changes to never be written to Consul. The fix is minimal (one helper function + one `&&` clause) and has no user-visible API or schema changes.

  > Not applicable — the change is contained to the reconciler diff logic and can be reverted by simply reverting this PR with no additional migration steps.

  > This fix restores correct enforcement of the `ext_authz` filter (an AuthZ security control) on `HTTPRoute` updates. Without this fix, a user who intended to re-enable per-route authentication by patching an `HTTPRoute` would see no enforcement change on the data plane, creating a security gap. The fix closes that gap; it does not add or remove any logging pipeline or access control method.


anandmukul93:
@bharath-k1999 
so the changes are only to add Filters.ExtAuthz understanding and the reconciler  now as well understands difference between enabled disabled and nil and work in each case ?

anandmukul93 (dismissed):
LGTM

control-plane/api-gateway/common/diff.go
```diff
@@ -257,7 +257,11 @@ func (e entryComparator) httpRouteRulesEqual(a, b api.HTTPRouteRule) bool {
 		bothNilOrEqualFunc(a.Filters.RetryFilter, b.Filters.RetryFilter, e.retryFiltersEqual) &&
 		bothNilOrEqualFunc(a.Filters.TimeoutFilter, b.Filters.TimeoutFilter, e.timeoutFiltersEqual) &&
 		bothNilOrEqualFunc(a.Filters.JWT, b.Filters.JWT, e.jwtFiltersEqual) &&
-		slices.EqualFunc(a.Filters.ExtProc, b.Filters.ExtProc, e.extProcFiltersEqual)
```
pajay-rao, line 260: slices.EqualFunc(a.Filters.ExtProc, b.Filters.ExtProc, e.extProcFiltersEqual) is getting removed. 
It is needed

pajay-rao (changes requested):

control-plane/api-gateway/common/diff.go
bharath-k1999: looks like it got removed during merge conflicts

pajay-rao (dismissed):

bharath-k1999:
> @bharath-k1999 so the changes are only to add Filters.ExtAuthz understanding and the reconciler now as well understands difference between enabled disabled and nil and work in each case ?

>ExtAuthz was the only one missed previously. All other filters in httpRouteRulesEqual were already covered.


Gateway ExtAuthz block | Route rule ExtAuthz | What Envoy does
-- | -- | --
nil (omitted) | nil | ✅ check runs (gateway default = enabled)
nil (omitted) | &{Enabled: true} | ✅ check runs (route overrides, same result)
nil (omitted) | &{Enabled: false} | ✅ check skipped (route overrides gateway default)
&{Enabled: true} | nil | ✅ check runs (inherits gateway)
&{Enabled: true} | &{Enabled: false} | ✅ check skipped (route overrides)
&{Enabled: false} | nil | ✅ check skipped (inherits gateway)
&{Enabled: false} | &{Enabled: true} | ✅ check runs (route overrides)

pajay-rao (approved):

PR #5527, opened 2026-07-28
fix(helm): global.trustedCAs certificates are never trusted (missing OpenSSL hash symlinks)

shani34:

Fixes custom CAs supplied via `global.trustedCAs` are not actually trusted.

The chart writes the `global.trustedCAs` certificates into `/trusted-cas` and points
`SSL_CERT_DIR` at that directory. However, **OpenSSL only discovers CAs in a certificate
directory by their subject hash** (`<hash>.N`); the plain `.pem` files are ignored. The
certificates were therefore present but never trusted.

- **Create OpenSSL subject-hash symlinks (`<hash>.N`)** for every CA written to
  `/trusted-cas`, in both `server-statefulset.yaml` and
  `telemetry-collector-deployment.yaml`. This is the core fix.

- **Copy the Vault-injected server CA** (`/vault/secrets/serverca.crt`) into
  `/trusted-cas` when `global.secretsBackend.vault.enabled` is set. Without this the
  Vault-issued CA remains untrusted, which is the reported reproduction path
  ("custom CA w/ Vault as a backend"). This mirrors the workaround the reporting
  customer applied to their forked chart.

- **Allocate hash suffixes sequentially** (`.0`, `.1`, `.2` …) rather than always `.0`.
  Certificates that share a subject share a subject hash, so a fixed `.0` would
  overwrite and silently untrust the older CA — precisely what happens during a CA
  rotation, when the old and new CA are both listed.

- **Guard the whole block on `command -v openssl`.** The
  `consul-telemetry-collector` image does not ship openssl, and the container command
  runs under `sh -ec`, so an unguarded `openssl` call terminates container startup
  before `exec`.

- **Clear stale symlinks before rebuilding.** `/trusted-cas` is an `emptyDir` that
  survives container restarts, so links would otherwise accumulate on each restart.

- Logic is consolidated into a single `consul.trustedCAsRehash` helper in
  `_helpers.tpl` rather than duplicated across the two templates.

Helm unit tests only assert the **rendered string**, so they cannot catch a missing
binary or a non-functional symlink. Testing was therefore done by extracting the
actual `helm template` output and executing it inside the real container images
(`hashicorp/consul` and `hashicorp/consul-telemetry-collector:0.0.2`).

Trust verified end-to-end with purpose-built CAs — including **two distinct CAs
sharing a subject** — and leaf certificates signed by each:

| Check | Result |
| --- | --- |
| `openssl verify -CApath`, standard CA | OK (failed before the fix) |
| `openssl verify -CApath`, colliding-subject CA | OK |
| `openssl verify -CApath`, Vault-injected CA | OK |
| Telemetry collector container startup | exit 0, reaches `exec` |
| Re-run simulating container restart | byte-identical, idempotent |
| Five repeated runs | link count stable |
| Malformed certificate present | skipped; valid certs still linked; no crash |
| Empty `/trusted-cas` | no crash |
| Go `x509.SystemCertPool()` before vs after symlinks | 145 = 145, no duplicate loading |
| Chart unit test suite | 2220/2220 pass |
| `helm lint` | pass (pre-existing PodDisruptionBudget warnings only) |

Non-breaking verification:

- The **full rendered chart was diffed against `main`** across five configurations
  (defaults; telemetry collector; TLS + ACLs; connect inject + mesh gateway;
  metrics + UI). Output is **byte-identical in every case** — users who do not set
  `global.trustedCAs` see no change, so there is no pod spec churn or rolling
  restart on upgrade.
- With `global.trustedCAs` set, the rendered diff is **insertions only**; no existing
  line is removed or modified.
- Consul is a Go binary and Go reads *every* file in `SSL_CERT_DIR`, so the added
  symlinks were checked for double-loading. Go skips same-directory symlinks and the
  loaded certificate count is unchanged — the change is inert for Go consumers and
  benefits OpenSSL-based ones.

1. Deploy with `global.trustedCAs` set to a custom CA, and with
   `global.secretsBackend.vault.enabled=true` plus `server.serverCert.secretName` set.
2. Exec into the `consul` container and confirm the hashed symlinks exist alongside
   the `.pem` files, including one for the Vault CA:
   `ls -l /trusted-cas`
3. Confirm the CA is actually trusted:
   `openssl verify -CApath /trusted-cas <cert-signed-by-that-CA>`
4. Confirm no regression to the collector: deploy with
   `telemetryCollector.enabled=true` together with `global.trustedCAs` and check the
   pod reaches Running rather than CrashLoopBackOff.
5. Optionally, list two CAs that share a subject (e.g. an old and new revision of the
   same root) and confirm both receive links (`<hash>.0` and `<hash>.1`).

### Checklist ###
- [x] Tests added

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

  Reverting the pull request is sufficient. The change is confined to Helm template
  rendering with no persisted state; reverting and upgrading restores the previous
  container command exactly, and `/trusted-cas` is an `emptyDir` recreated per pod.

- [x] If applicable, I've documented the impact of any changes to security controls.

  This change affects **certificate trust establishment**. It is strictly additive:
  CAs that the operator explicitly supplied via `global.trustedCAs` (and, under Vault,
  the Consul server CA) become discoverable to OpenSSL, which is the documented intent
  of the setting. No CA is trusted that the operator did not already configure, the
  system trust store at `/etc/ssl/certs` remains first in `SSL_CERT_DIR`, and no
  verification is disabled or relaxed. The collision fix additionally prevents a CA
  the operator *did* configure from being silently dropped.


sanikachavan5 (approved):
Looks good, but pipeline is failing

Manishakumari-hc (approved):
LGTM. The testing result is good as added in description

PR #5562, opened 2026-08-07
Consul CNI DaemonSet Toleration Helm Value Enablement

binilvrgs:
### Issue

The Consul CNI DaemonSet in `charts/consul/templates/cni-daemonset.yaml` has its `tolerations` block hard-coded with no Helm value to override it:

```yaml
tolerations:
  - key: CriticalAddonsOnly
    operator: Exists
  - effect: NoExecute
    operator: Exists
```

There is no `connectInject.cni.tolerations` value in `values.yaml`, so operators have no way to customise the toleration list at deploy time.

---

### Customer Impact

Bank of America (BoFA / FRB-783) runs Consul on OpenShift, which enforces a strict 3-tier node topology:

- **Master nodes** — tainted `node-role.kubernetes.io/master:NoSchedule`
- **Infra nodes** — tainted `node-role.kubernetes.io/infra:NoSchedule`
- **Worker nodes** — no taint

An internal GIS policy requires operators to explicitly control the toleration list. Without `connectInject.cni.tolerations`, BoFA has no compliant mechanism to guarantee CNI stays off master/infra nodes across chart upgrades.

---

### Root Cause

Historically the tolerations block contained a `NoSchedule: Exists` wildcard (no key) which did schedule CNI on master/infra nodes. That entry has since been removed from the current file. However, there is still no Helm value exposed, leaving operators with no formal control or audit trail over the toleration list.

---

### Requirement

Expose `connectInject.cni.tolerations` as a Helm value so operators can supply their own toleration list. The default must remain unchanged — operators who do not set the value get exactly the current behaviour.

---

### Fix

**1. `charts/consul/templates/cni-daemonset.yaml`**

```diff
+     {{- if .Values.connectInject.cni.tolerations }}
+     tolerations:
+       {{ tpl .Values.connectInject.cni.tolerations . | nindent 8 | trim }}
+     {{- else }}
      tolerations:
        - key: CriticalAddonsOnly
          operator: Exists
        - effect: NoExecute
          operator: Exists
+     {{- end }}
```

**2. `charts/consul/values.yaml`**

```yaml
# Toleration settings for the CNI installer DaemonSet pods.
# This should be a multi-line string matching the Tolerations
# array in a Pod spec. If not set, defaults to tolerating
# CriticalAddonsOnly and NoExecute taints.
# @type: string
tolerations: null
```

**3. `charts/consul/test/unit/cni-daemonset.bats`** — added unit tests covering default behaviour, custom value, and null override.

### TLDR

Any tolerations defined in values.yaml will replace the default CNI DaemonSet tolerations. If no tolerations are specified in values.yaml, the default tolerations will be applied.

Example:

If values.yaml contains:

```yaml
connectInject:
  cni:
    tolerations: |
      - key: node-role.kubernetes.io/worker
        operator: Exists
        effect: NoSchedule
```

Then the effective tolerations applied to daemonset/consul-cni under spec.template.spec.tolerations will be:

```json
[
  {
    "key": "node-role.kubernetes.io/worker",
    "operator": "Exists",
    "effect": "NoSchedule"
  }
]
```


charts/consul/templates/cni-daemonset.yaml
```diff
@@ -43,12 +43,19 @@ spec:
       # consul-cni only runs on linux operating systems
       nodeSelector:
         kubernetes.io/os: linux
+
+      {{- if .Values.connectInject.cni.tolerations }}
+      tolerations:
+        {{ tpl .Values.connectInject.cni.tolerations . | nindent 8 | trim }}
```
vdinesh4738, line 48: tpl is not necessary here, because we are expecting it to be yaml only. toYaml should be good enough

vdinesh4738 (changes requested):

charts/consul/templates/cni-daemonset.yaml
binilvrgs:
```yaml
# Configures the automatic Connect sidecar injector.
connectInject:
  # Configures consul-cni plugin for Consul Service mesh services
  cni:
    # Toleration settings for the CNI installer DaemonSet pods.
    # This should be a multi-line string matching the Tolerations
    # array in a Pod spec. If not set, defaults to tolerating
    # CriticalAddonsOnly and NoExecute taints.
    # @type: string
    tolerations: null
    
# Server, when enabled, configures a server cluster to run. This should
# be disabled if you plan on connecting to a Consul cluster external to
# the Kube cluster.
server:
  # Toleration settings for server pods. This
  # should be a multi-line string matching the
  # [Tolerations](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/)
  # array in a Pod spec.
  tolerations: ""

  # @type: map
  extraLabels: null
```

Here `tolerations` field is defined as `@type: string` and expecting a multi-line string. not a yaml   `# @type: map`

Also in multiple places, example in `server.tolerations` default value is set as empty string, "". that's why we used **tpl** expecting a string, `toYaml ` is proffered in case of  `# @type: map` like in the field `extraLabels`

Also I can see, in client-daemonset, **tpl** is used

image


connect-inject-deployment
image


* tpl is used with tolerations (in 19 places), by expecting a multi-line string

TDLR

```
@type: string  →  use tpl
@type: map     →  use toYaml
```

vdinesh4738: Can you please add the example in the values.yaml; 

Are you suggesting that we will be passing the the values as below in the values.yaml

```
tolerations: |
  - key: dd
    operator: Exists
  - effect: bb
    operator: Exists
```
if this is the case it is fine. Please provide the example in the values.yaml or else it would be confusing.

vdinesh4738 (changes requested):

charts/consul/templates/cni-daemonset.yaml
binilvrgs: Yes, exactly. The tolerations should be specified in values.yaml using the following format (as a mule-line string):

```yaml
connectInject:
  cni:
    tolerations: |
      - key: dd
        operator: Exists
      - effect: bb
        operator: Exists
```

This is the expected format and is consistent with how tolerations are defined in multiple other places across the repository. I have updated the description to include this example for clarity.

vdinesh4738 (approved):

vdinesh4738:
@binilvrgs Have tried the below test:

Consul deployed with cni dameonset. Then try helm upgrade with your changes.

hc-github-team-consul-core:
:mega: Hi @binilvrgs! a backport is missing for this PR [5562] for versions [2.0.x] please perform the backport manually and add the following snippet to your backport PR description:
```
<details>
	<summary> Overview of commits </summary>
		- <<backport commit 1>>
		- <<backport commit 2>>
		...
</details>
```

PR #5597, opened 2026-08-14
Add TLS annotations for  API Gateway Auto-DNS Registration & Zero-Touch TLS Termination

bharath-k1999:
# PR: API Gateway Auto-DNS Registration & Zero-Touch TLS Termination

> **K8s implementation**: consul-k8s PR #5597  
> **Server implementation**: consul PR #23647

---

## Summary

This PR adds Kubernetes-side support for **operator-free HTTPS termination** on
API Gateways. A single annotation on the `Gateway` resource opts into
Consul-managed TLS — no `Secret`, no `CertificateRef`, no manual CA rotation.

The Consul server (PR #23647) issues a Connect leaf certificate whose DNS SANs
include a wildcard covering every service registered under the gateway.
consul-k8s translates the annotation into the `APIGatewayConfigEntry`, and the
dataplane uses that cert to terminate downstream HTTPS automatically.

---

## Problem

Previously, enabling HTTPS on a Consul API Gateway required the operator to:

1. Generate a TLS certificate and private key.
2. Store them as a Kubernetes `Secret`.
3. Reference the `Secret` via `spec.listeners[*].tls.certificateRefs`.
4. Rotate that certificate on expiry.

This created operational overhead even when the workload was fully inside the
Consul service mesh, where Connect already provides mutual TLS between services.

---

### Annotation → Config Entry

Add the annotation to a `Gateway` manifest:

```yaml
annotations:
  consul.hashicorp.com/tls-enabled: "true"
```

No `certificateRefs` are needed on the listener. The translator converts this
into `APIGatewayConfigEntry.TLS.Enabled = true`, which signals Consul to use
the gateway's auto-issued Connect leaf certificate for downstream TLS
termination.

### DNS Auto-Registration

The Consul server registers each gateway service under a special DNS zone:

```
<service>.api-gateway.<domain>
```

For Community Edition the domain is `consul`, producing
`static-server.api-gateway.consul`. In Consul Enterprise with namespace
mirroring the domain is `<ns>.consul`, producing
`static-server.api-gateway.<ns>.consul`.

CoreDNS is configured to forward `.consul` queries to the Consul DNS service,
so pods inside the cluster resolve these names without any extra DNS
infrastructure.

### Wildcard SANs on the Leaf Certificate

The Connect leaf certificate issued to the gateway carries wildcard DNS SANs:

| Deployment | SAN 1 | SAN 2 |
|---|---|---|
| CE (default ns, default partition) | `*.api-gateway.consul` | `*.api-gateway.<dc>.consul` |
| Ent (non-default ns, default partition) | `*.api-gateway.<ns>.consul` | `*.api-gateway.<ns>.<dc>.consul` |

Any service hostname derived from the wildcard — e.g.
`static-server.api-gateway.consul` — is covered by the cert without issuing a
dedicated per-service certificate.

### End-to-End Traffic Flow

```
curl (client pod)
  │  HTTPS  static-server.api-gateway.consul:8443
  ▼
API Gateway pod  ← terminates HTTPS with Connect leaf cert
  │  mTLS (Envoy → Envoy)
  ▼
static-server sidecar → static-server app
```

The client verifies the gateway cert against the Consul Connect CA root. No
`-k` / `--insecure` flag, no operator-supplied certificate.

---

### Kubernetes Controller (`control-plane/api-gateway/common/`)

| File | Change |
|---|---|
| [`translation.go`](../control-plane/api-gateway/common/translation.go) | `gatewayTLS()` reads `consul.hashicorp.com/tls-enabled: "true"` and returns `api.GatewayTLSConfig{Enabled: true}`, which is set on `APIGatewayConfigEntry.TLS` |
| [`diff.go`](../control-plane/api-gateway/common/diff.go) | `apiGatewaysEqual()` now compares `TLS.Enabled` and `ExtAuthz` so that adding or removing the annotation triggers a Consul config entry write |
| [`diff.go`](../control-plane/api-gateway/common/diff.go) | New `apiGatewayExtAuthzEqual()` helper for nil-safe `ExtAuthz` comparison |
| [`constants.go`](../control-plane/api-gateway/common/constants.go) | Defines `AnnotationTLSEnabled` and `TLSEnabledValue` constants |

The same `diff.go` fix is applied to
[`api-gateway-custom/common/diff.go`](../control-plane/api-gateway-custom/common/diff.go)
to keep the two gateway controller paths in sync.

### Unit Tests

| File | Tests added |
|---|---|
| [`translation_test.go`](../control-plane/api-gateway/common/translation_test.go) | `TestTranslator_ToAPIGateway_GatewayLevelTLS` — 3 table cases: annotation present, absent, and unrecognised value |

### Acceptance Test (`acceptance/tests/api-gateway/`)

`TestAPIGateway_AutoTLS` in
[`api_gateway_auto_tls_test.go`](../acceptance/tests/api-gateway/api_gateway_auto_tls_test.go)
is a single end-to-end test that covers the full feature path. When run with
`-enable-enterprise` it activates enterprise-specific SAN assertions
automatically — no separate test function is needed.

**Assertions (in order):**

1. `APIGatewayConfigEntry.TLS.Enabled = true` is set in Consul after apply.
2. `Gateway` status condition `Accepted=true`.
3. `HTTPRoute` status condition `Accepted=true` with one parent ref.
4. Sanity HTTPS via raw gateway IP with `-k` (baseline).
5. Leaf certificate is fetched via `openssl s_client` from inside the cluster.
6. **CE mode** — cert carries `*.api-gateway.consul` + `*.api-gateway.<dc>.consul`.  
   **Ent mode** — cert carries `*.api-gateway.<ns>.consul` + `*.api-gateway.<ns>.<dc>.consul`.
7. `x509.VerifyHostname` confirms the wildcard covers the derived hostname.
8. Leaf cert chains to the Consul Connect CA root (no operator cert required).
9. DNS resolution: `static-server.api-gateway.consul` resolves to the gateway IP.
10. **CA-verified HTTPS** using `--cacert /tmp/consul-ca.crt` — no `-k`.
11. Annotation round-trip regression: remove annotation → `TLS.Enabled=false`; re-add → `TLS.Enabled=true`.
12. HTTPS still works after annotation round-trip.

### Test Fixtures (`acceptance/tests/fixtures/cases/api-gateways/auto-tls/`)

| File | Description |
|---|---|
| `gateway.yaml` | `Gateway` with `tls-enabled: "true"`, HTTPS listener, **no** `certificateRefs` |
| `gatewayclass.yaml` | `GatewayClass` pointing to `consul.hashicorp.com/gateway-controller` |
| `gatewayclassconfig.yaml` | `GatewayClassConfig` with `mapPrivilegedContainerPorts: 8000` |
| `httproute.yaml` | `HTTPRoute` routing to `static-server:80` |
| `kustomization.yaml` | Includes `static-server-inject` fixture (connect-inject enabled backend) |

---

### Prerequisites

- Docker Desktop (arm64 / amd64)
- `kind` ≥ v0.23
- `kubectl`
- `go` ≥ 1.21
- Both repositories cloned:
  - `consul-k8s` on branch `bharath/auto-dns-registration`
  - `consul` on branch `bharath/api-gateway-auto-dns`

### 1 — Build images

```bash
# consul server (from hashicorp/consul repo)
cd /path/to/consul
GOARCH=arm64 make linux
docker buildx build \
  --platform linux/arm64 \
  -f build-support/docker/Consul-Dev-Multiarch.dockerfile \
  -t consul:local \
  ./pkg/bin/

# consul-k8s control-plane (from this repo)
cd /path/to/consul-k8s
GOOS=linux GOARCH=arm64 make dev-docker
# produces consul-k8s-control-plane:local

# consul-dataplane — retag an existing build
docker tag hashicorp/consul-dataplane:2.0-arm64 hashicorp/consul-dataplane:local
```

### 2 — Create kind cluster and load images

```bash
kind create cluster \
  --config=acceptance/framework/environment/kind/kind.config \
  --name dc1 \
  --image kindest/node:v1.30.2@sha256:ecfe5841b9bee4fe9690f49c118c33629fa345e3350a0c67a5a34482a99d6bba

kind load docker-image --name dc1 consul:local
kind load docker-image --name dc1 hashicorp/consul-dataplane:local
kind load docker-image --name dc1 consul-k8s-control-plane:local
```

### 3 — Run the acceptance test (CE)

```bash
cd acceptance/tests/api-gateway

go test -v -timeout 30m -run TestAPIGateway_AutoTLS \
  -consul-image consul:local \
  -consul-k8s-image consul-k8s-control-plane:local \
  -consul-dataplane-image hashicorp/consul-dataplane:local \
  -use-kind \
  -no-cleanup-on-failure \
  .
```

Expected output (trimmed):

```
=== RUN   TestAPIGateway_AutoTLS
    CE mode: verifying *.api-gateway.consul and *.api-gateway.<dc>.consul SANs
    leaf cert carries CE wildcard SANs ✓
    derived service hostname: static-server.api-gateway.consul
    leaf cert wildcard SAN covers derived hostname ✓
    leaf cert chains to Consul Connect CA ✓
    DNS resolution of static-server.api-gateway.consul → gateway ✓
    CA-verified HTTPS to https://static-server.api-gateway.consul:8443 (no -k)
    CA-verified HTTPS to static-server.api-gateway.consul ✓
    regression: removing tls-enabled annotation → expect TLS.Enabled=false
    regression: re-adding tls-enabled annotation → expect TLS.Enabled=true
--- PASS: TestAPIGateway_AutoTLS (166.16s)
```

### 4 — Run with Enterprise (Ent SANs)

```bash
export CONSUL_ENT_LICENSE="<your-license>"

go test -v -timeout 30m -run TestAPIGateway_AutoTLS \
  -consul-image consul-enterprise:local \
  -consul-k8s-image consul-k8s-control-plane:local \
  -consul-dataplane-image hashicorp/consul-dataplane:local \
  -use-kind \
  -enable-enterprise \
  -no-cleanup-on-failure \
  .
```

With `-enable-enterprise` the test automatically:
- Adds `global.enableConsulNamespaces=true` and
  `connectInject.consulNamespaces.mirroringK8S=true` to the Helm install.
- Calls `verifyEntSANs` instead of `verifyCESANs`, asserting
  `*.api-gateway.<ns>.consul` and `*.api-gateway.<ns>.<dc>.consul`.

### Helm values used by the test

| Key | Value |
|---|---|
| `connectInject.enabled` | `true` |
| `global.acls.manageSystemACLs` | `true` |
| `global.tls.enabled` | `true` |
| `global.logLevel` | `trace` |
| `dns.enabled` | `true` |
| `global.enableConsulNamespaces` *(Ent only)* | `true` |
| `connectInject.consulNamespaces.mirroringK8S` *(Ent only)* | `true` |

### Minimum Gateway manifest

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: my-gateway
  annotations:
    consul.hashicorp.com/tls-enabled: "true"   # ← only addition required
spec:
  gatewayClassName: consul
  listeners:
  - name: https
    protocol: HTTPS
    port: 8443
    tls:
      mode: Terminate
    allowedRoutes:
      namespaces:
        from: All
```

> `certificateRefs` is deliberately omitted. Consul supplies the cert
> automatically via the Connect leaf certificate mechanism.


control-plane/api-gateway-custom/controllers/gateway_controller.go
```diff
@@ -430,6 +430,21 @@ func SetupGatewayControllerWithManager(ctx context.Context, mgr ctrl.Manager, co
...
+	// control-plane/api-gateway/controllers/gateway_controller.go: the
+	// ComponentLabel is set on generated resources, not on the user-authored
+	// Gateway CR, so on its own it drops every Gateway event. Metadata alone is
+	// not a safe discriminator here either, since a newly created Gateway has
+	// no Consul annotation yet. Admit anything naming a GatewayClass and let
+	// Reconcile do the authoritative ControllerName ownership check.
+	gwPredicate = predicate.Or(
+		gwPredicate,
+		common.ConsulAnnotationPredicate(),
+		predicate.NewPredicateFuncs(func(o client.Object) bool {
```
sanikachavan5: predicate.NewPredicateFuncs(func(o client.Object) bool {
    gw, ok := o.(*gwv1.Gateway)
    return ok && gw.Spec.GatewayClassName != ""
})
gatewayClassName is a required field in Gateway API (gateway_types.go:70 — no omitempty, CRD-enforced), so it is never empty on an admitted object. And For(&gwv1.Gateway{}) only ever delivers *gwv1.Gateway, so the type assertion always succeeds. This clause returns true for every event.
That collapses predicate.Or(gwPredicate, ConsulAnnotationPredicate(), <always-true>) to true, which means:
- the ComponentLabel selector at :438 is dead;
- ConsulAnnotationPredicate is dead — ~460 lines of new predicates.go + predicates_test.go across both trees that cannot affect behavior. The tests pass only because they invoke the predicate directly, never through the wiring.
The underlying bug and your diagnosis are correct. Make the fix say what it actually does:
// ComponentLabel is only applied to resources we generate (Deployment/Service/Pod),
// never to the user-authored Gateway CR, so filtering the root watch on it drops
// every Gateway event. Reconcile performs the authoritative ownership check via
// gatewayClass.Spec.ControllerName, so admit all Gateways here.
builder.For(&gwv1.Gateway{})
Identical behavior, ~460 fewer lines. If ConsulAnnotationPredicate is wanted for a future watch, land it in the PR that uses it — don't merge it dead.

control-plane/api-gateway/controllers/gateway_controller.go
```diff
@@ -440,6 +440,31 @@ func SetupGatewayControllerWithManager(ctx context.Context,
...
+	// during reconciliation), so filtering on metadata alone deadlocks it --
+	// the annotation can only appear after a reconcile that the filter itself
+	// prevents. Admit anything that names a GatewayClass; Reconcile already
+	// performs the authoritative ownership check by comparing
+	// gatewayClass.Spec.ControllerName against GatewayClassControllerName and
+	// returns early for gateways this controller does not own.
+	gwPredicate = predicate.Or(
+		gwPredicate,
+		common.ConsulAnnotationPredicate(),
+		predicate.NewPredicateFuncs(func(o client.Object) bool {
```
sanikachavan5: same as above

acceptance/framework/k8s/deploy.go
```diff
@@ -130,14 +130,24 @@ func CheckStaticServerConnectionMultipleFailureMessages(t *testing.T, options *k
...
 	retry.RunWith(retrier, t, func(r *retry.R) {
 		output, err := RunKubectlAndGetOutputE(r, options, args...)
 		if expectSuccess {
+			if err != nil {
+				// Fast failure: if the pod or container does not exist or has already terminated,
+				// fail immediately instead of retrying for the full timeout.
+				errStr := err.Error()
+				if strings.Contains(errStr, "pods \""+sourceApp+"\" not found") ||
+					strings.Contains(errStr, "container not found") ||
+					strings.Contains(errStr, "not found") && strings.Contains(errStr, "NotFound") {
```
sanikachavan5: && binds tighter than ||, so this is A || B || (C && D). That's likely the intent, but it doesn't read that way, add explicit parentheses. Also A is fully subsumed by C && D, and C alone matches any NotFound error, including unrelated ones. Collapse the redundancy.

Broader concern: this string-matches kubectl stderr to detect a typed condition. If a kubectl version rewords the message, the fast-fail silently stops firing and the helper reverts to hanging for the full timeout — reintroducing the exact flake this is meant to fix, but harder to spot next time.

sanikachavan5 (changes requested):

control-plane/api-gateway-custom/controllers/gateway_controller.go
bharath-k1999: Verified each point: GatewayClassName is +required with no omitempty and CRD-enforced non-empty, so that clause is always true; For(&gwv1.Gateway{}) only delivers gwv1.Gateway, so the type assertion can't fail; predicate.Or therefore short-circuits and both the ComponentLabel selector and ConsulAnnotationPredicate were unreachable.

I also confirmed empirically why the label selector was the bug rather than the fix. ComponentLabel is only written by common.LabelsForGateway(), and every caller is in gatekeeper/ (Deployment, Service, Role, ServiceAccount, Secret, RoleBinding). On a live cluster all Gateway CRs show LABELS <none> while the generated Deployment carries component=api-gateway — so the selector matched zero Gateways and dropped 100% of root-watch events. Reconciles were only firing incidentally via the secondary Owns/Watches sources, which is exactly why annotation-only and spec-only edits were silently lost.

Applied your suggestion verbatim, in both trees, and deleted predicates.go + predicates_test.go from both (net −524/+17). podPredicate is retained — Pods do carry the label, so filtering there is correct.

control-plane/api-gateway/controllers/gateway_controller.go
bharath-k1999: added as suggested

acceptance/framework/k8s/deploy.go
bharath-k1999: removed that fast failure block

sanikachavan5 (dismissed):

sanikachavan5 (dismissed):

